PR #133, opened 2022-06-06
DID SDK

pfrazee:
Adds `@adx/did-sdk` with support for did:key, did:web, and did:ion. Includes:

- Tools to resolve and get fields from a DID
- Tools to create (and modify) did:key and did:ion
- A simple did:web server


did-sdk/test/did-web.ts
```diff
@@ -0,0 +1,77 @@
...
+    t.deepEqual(didRes.getPublicKey('assertionMethod'), ed25519Key)
+    t.deepEqual(didRes.getPublicKey('authentication'), ed25519Key)
+    t.deepEqual(didRes.getPublicKey('capabilityInvocation'), ed25519Key)
+    t.deepEqual(didRes.getPublicKey('capabilityDelegation'), ed25519Key)
+    t.deepEqual(didRes.listServices(), [service])
+    t.deepEqual(didRes.getService(service.type), service)
+  }
+})
+
+test('Resolve throws on malformed did:webs', async (t) => {
```
dholms, line 73: test suite is sometimes hanging & timing out on this one. other times it works. not sure why.

package.json
```diff
@@ -12,6 +12,7 @@
   "workspaces": [
     "common",
     "cli",
+    "did-sdk",
```
dholms, line 15: added this to the monorepo so you can install all dependencies with `yarn` from root. added a bunch to the root yarn.lock

did-sdk/src/ion/keypairs.ts
```diff
@@ -0,0 +1,35 @@
+import { LdKeyPairInstance } from '@transmute/ld-key-pair'
+import { Ed25519KeyPair } from '@transmute/did-key-ed25519'
+import { Secp256k1KeyPair } from '@transmute/did-key-secp256k1'
+
+export type KeyType = 'ed25519' | 'secp256k1'
+export type generateFromSeedOptions = { secureRandom: () => Buffer }
```
dholms, line 6: will the `Buffer` type work alright in browsers? does this allow for Uint8array/array buffer to be returned?

did-sdk/README.md
```diff
@@ -0,0 +1,49 @@
...
+console.log(did.getService('SomeService')) // => ServiceEndpoint
+
+// create a did
+const didKey = await didSdk.key.create('ed25519', {
+  secureRandom: () => crypto.randomBytes(32)
+})
+const didIon = await didSdk.ion.create(
+  {
+    services: [{
+      id: Buffer.from('#service1', 'utf8').toString('base64'),
```
dholms, line 25: Can this id be anything? or is it convention that it's the base64 version of the service name? if so, maybe we move that into the library code?

did-sdk/README.md
```diff
@@ -0,0 +1,49 @@
...
+      type: 'SomeService',
+      serviceEndpoint: 'https://example.com'
+    }]
+  },
+  {
+    keyType: 'secp256k1',
+    secureRandom: () => crypto.randomBytes(32),
```
dholms, line 32: with webcrypto in node & browsers, I've half a mind to just default this to `webcrypto.getRandomValues()` with the option to override

did-sdk/src/web/server.ts
```diff
@@ -0,0 +1,62 @@
...
+  constructor(public port: number) {
+    this._server = http
+      .createServer((req, res) => this._onRequest(req, res))
+      .listen(port)
+    this.whenReady = new Promise((resolve) => {
+      this._server.on('listening', () => resolve())
+    })
+  }
+
+  _onRequest(req: http.IncomingMessage, res: http.ServerResponse) {
```
dholms, line 20: ah a man of culture i see. writing your own request handler 🕴️

did-sdk/.eslintrc
```diff
@@ -0,0 +1,19 @@
+{
```
dholms, line 1: need to figure out how to move this all to repo root so we can stop duplicating it in every package

did-sdk/src/ion/ion.ts
```diff
@@ -0,0 +1,340 @@
...
+  ionResolveEndpoint = RESOLVE_ENDPOINT,
+): Promise<ReadOnlyDidDocAPI> {
+  const res = await fetch(ionResolveEndpoint + '/' + didUri)
+  if (res.status >= 400) throw new Error('Not Found')
+  const resObj = (await res.json()) as ResolveResponse
+  if (!resObj?.didDocument) throw new Error('Not found')
+  return new ReadOnlyDidDocAPI(resObj.didDocument, resObj.didDocumentMetadata)
+}
+
+export async function create(
```
dholms, line 55: Do we have a way to do the signature from another location? 

(putting aside the key management issues of this for a minute, let's just say that a ledger or TPM can sign valid ION updates), could we generate a keypair on the hardware device, format the DID doc, and then send it over to the hardware device for signing?

did-sdk/src/key/key.ts
```diff
@@ -0,0 +1,120 @@
...
+  state: DidKeySerializedState,
+): Promise<KeyDidDocAPI> {
+  const did = new KeyDidDocAPI()
+  await did.hydrate(state)
+  return did
+}
+
+export class KeyDidDocAPI extends WritableDidDocAPI {
+  id = ''
+  keys: DidKey[] = []
```
dholms, line 52: is there ever a world in which a `did:key` has more than one key? seems to go against the nature of it 🤔 

or is this just for consistency's sake?

did-sdk/src/ion/ion.ts
```diff
@@ -0,0 +1,340 @@
...
+    return op
+  }
+
+  async update(params: UpdateParams, options: KeyParams): Promise<OpUpdate> {
+    this.assertActive()
+    const op: OpUpdate = {
+      operation: 'update',
+      content: params,
+      previous: this.getPreviousOperation('update'),
+      update: await generateKeyPair(options.keyType, options),
```
dholms, line 226: oh every update you set a new update keypair?

did-sdk/src/ion/ion.ts
```diff
@@ -0,0 +1,340 @@
...
+    }
+    if (!op.previous.update) {
+      throw new Error('Update key not found on previous ops')
+    }
+    const reqBody = await IonSdk.IonRequest.createUpdateRequest({
+      didSuffix: this.getSuffix(),
+      signer: IonSdk.LocalSigner.create(op.previous.update.privateJwk),
```
dholms, line 233: where is this private key getting persisted to? It looks like it's just memory in the tests

did-sdk/src/key/key.ts
```diff
@@ -0,0 +1,120 @@
+import * as key from '@transmute/did-key.js'
+import { DIDDocument } from 'did-resolver'
+import { ReadOnlyDidDocAPI, WritableDidDocAPI } from '../did-documents.js'
+
+export type generateFromSeedOptions = { secureRandom: () => Buffer }
+export type generateFromRandomOptions = {
+  kty: 'EC' | 'OKP'
+  crvOrSize: 'P-256' | 'P-384' | 'P-521'
+}
+export type KeyType =
```
dholms, line 10: do we want to support NIST curves here? I always go back & forth on it. but webcrypto & a lot of device TPMs only support RSA + NIST curves. 

I'm using NIST P-256 in AWAKE for instance

did-sdk/.eslintrc
AaronGoldman: Are you referring to cascading
https://eslint.org/docs/user-guide/configuring/configuration-files#cascading-and-hierarchy

dholms: yes but i think it runs into some hiccups in monorepos cause it stops looking after a package.json

but i haven't really looked into this yet. not enough friction to really warrant it 😅

did-sdk/test/did-web.ts
pfrazee: This is on your machine? Is it definitely this test?

did-sdk/src/ion/keypairs.ts
pfrazee: We'll need to do a pass on all of this for browsers. I figured I'd wait till I have an environment using it so I can do the testing, but I'll see about switching to WebCrypto now

did-sdk/README.md
pfrazee: This is a really bizarre requirement that's enforced by the did-ion code and I have to see if it's a bug in their code or if the service IDs really do need to be base64 encoded.

did-sdk/src/web/server.ts
pfrazee: Heh indeed

did-sdk/src/ion/ion.ts
pfrazee: Generally speaking that should be possible; we'll just need to write more custom code.

did-sdk/src/key/key.ts
pfrazee: This is because the module we're using generates both the ed25519 and the derived x25519.

did-sdk/src/ion/ion.ts
pfrazee: Assuming their SDK is correct, yeah you rotate the signing keypair with *every* update.

did-sdk/src/ion/ion.ts
pfrazee: This is the `serialize()` API which did:key and did:ion provide. You're supposed to store the output of those methods and then you can load them with `inst()`.

For did:ion, it seems you want to store a local copy of the operations. It might be sufficient to only store the latest keypairs but I haven't explored that.

did-sdk/test/did-web.ts
dholms: yeah on my machine. hmm not sure 🤔

did-sdk/README.md
dholms: huh strange 🤔 

might make sense to move it into the library code then so that a developer doesn't have to see the strangeness

did-sdk/src/key/key.ts
dholms: ahhh got it

did-sdk/src/key/key.ts
pfrazee: I decided to see what the deal is with webcrypto curves and came across [this interesting exchange](https://news.ycombinator.com/item?id=19749019). I know tptacek isn't the end-authority on all crypto but he tends to know what he's talking about. The other thing I found is unconfirmed statements that Microsoft has been the blocker on 25519.

We might not really have a choice *not* to use the NIST curves. We need the widest support possible (TPMs) and it's probably safer to be using WebCrypto than js/wasm versions of crypto.

dholms: that's sorta my sense as well. TPM with NIST P-256 is definitely better than browser storage with ed25519. So we certainly shouldn't be dogmatic about not using NIST curves cause then we lose out on most available TPMs.

in the browser, the benefit of webcrypto is:
- you can store keys as "unexportable" in IDB. so an attack can do operations with them but not actually steal the key material
- operations are faster cause they're at the system level instead of javascript runtime which means you get some hardware acceleration on things like sha256

downside is obviously the downside of NIST over Ed25519:
- longer public keys
- not constant time
- possibility of a backdoor

did-sdk/README.md
pfrazee: 👍 Done

did-sdk/README.md
pfrazee: 🤦 It was the `#` that was choking it up. That "base64" validation is basically making sure the developer (me!) isn't putting values into the ID which can't be used in the fragment section of a URL correctly. Fixed by removing the pound sign.

did-sdk/README.md
```diff
@@ -0,0 +1,49 @@
...
+const did = await didSdk.resolve('did:web:example.com')
+console.log(did.getURI()) // => 'did:web:example.com'
+console.log(did.getController()) // => string
+console.log(did.listPublicKeys()) // => VerificationMethod[]
+console.log(did.getPublicKey('assertionMethod')) // => VerificationMethod
+console.log(did.listServices()) // => ServiceEndpoint[]
+console.log(did.getService('SomeService')) // => ServiceEndpoint
+
+// create a did
+const didKey = await didSdk.key.create('ed25519', {
```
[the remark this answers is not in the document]
pfrazee: I think this comment is on an outdated version. I have the readme showing `create('ed25519')` right now.

did-sdk/docs/classes/key.KeyDidDocAPI.md
```diff
@@ -0,0 +1,291 @@
...
+
+`VerificationMethod`
+
+#### Inherited from
+
+WritableDidDocAPI.getPublicKey
+
+#### Defined in
+
+[src/did-documents.ts:37](https://github.com/bluesky-social/bluesky-prototype/blob/05593da/did-sdk/src/did-documents.ts#L37)
```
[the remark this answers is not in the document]
pfrazee: Ah hmm, I'll see if `typedoc` supports that.

did-sdk/src/did-documents.ts
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  listServices(): ServiceEndpoint[] {
+    return this.didDoc.service || []
+  }
+
+  getService(type: string): ServiceEndpoint {
+    if (!this.didDoc.service || this.didDoc.service.length === 0)
+      throw new Error('Service not found')
+    const service = this.didDoc.service.find((s) => s.type === type)
```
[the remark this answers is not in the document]
pfrazee: We could add an optional type filter on `listServices()`, but generally dunno what else we would do than return the first match.

did-sdk/test/did-web.ts
dholms: cant replicate this anymore. think we're good 🤷‍♂️

PR #134, opened 2022-06-09
Browser wallet

dholms:
Instructions on how to run: https://github.com/bluesky-social/adx/blob/9d09d6910892bea8facc48e426818c7321036e1c/auth-lobby/README.md


auth-lobby/src/env.ts
```diff
@@ -0,0 +1,9 @@
+if (typeof process.env.REACT_APP_RELAY_HOST !== 'string') {
+  throw new Error('ENV: No relay host provided')
+}
+export const RELAY_HOST = process.env.REACT_APP_RELAY_HOST
+// @TODO: this is temporary for dev purposes
+if (typeof process.env.REACT_APP_ROOT_USER !== 'string') {
```
pfrazee, line 6: Is this check correct? Shouldn't it be looking at RELAY_HOST?

auth/src/browser-store.ts
```diff
@@ -0,0 +1,89 @@
...
+import * as builders from './builders.js'
+
+// @TODO: Stand in since we don't have actual root keys
+// ---------------
+const PRIV_KEY =
+  'aghhYelL58I8vwVWt7W5cs+Y4uYYWQma6jX0MjkCt7cOdiciXp5JJcOtoWHCgljZhLFV/h0E9zdeYXoSZ0rCgA=='
+
+const isRootKey = (): boolean => {
+  console.log(process.env.REACT_APP_ROOT_KEY)
+  return process.env.REACT_APP_ROOT_KEY === 'true'
```
pfrazee, line 12: Is there any way this can be an API that's exposed by the auth lib, rather than the auth lib directly referenceing process.env?

auth-lobby/src/env.ts
pfrazee: May just be that the error below is wrong

dholms: ah yup just the error is wrong. thx!

auth/src/browser-store.ts
dholms: ah yeah you're right. that's an artifact of refactoring this out of the app layer. I'll handle env vars in the react app 👍

PR #139, opened 2022-06-15
Crypto lib with public key compression

dholms:
This PR splits out our crypto functions into their own library.

Creates wrappers around NIST P-256 curves for both ECDH & ECDSA as well as a wrapper around AES-256 keys for symmetric encryption with AES-GCM.

Importantly, it correctly serializes NIST P-256 curves to DIDs which are meant to be represented as their 33byte _compressed_ public keys.

Integrates the changes into AWAKE


pfrazee (approved):
LGTM

PR #143, opened 2022-06-16
Dev Env

dholms:
Utility to spin up a customizable dev environment with one script.

Currently:
- personal data server(s)
- did:web server
- key manager
- did network
- auth lobby
- example app

This PR also adds in a simple custodial key management server & makes the did:web server persistent.

To configure the dev environment, update `dev-env/.env` with the ports that you want services running on. Any vars left blank will not be run, some vars take multiple comma-separate ports for spinning up more than one instance (for a PDS or auth lobby for instance)

_TODO: thread the .env through the various services so they know who to talk to_


did-sdk/src/web/web.ts
```diff
@@ -2,17 +2,25 @@ import { DIDDocument, parse } from 'did-resolver'
 import fetch from 'node-fetch'
 import { ReadOnlyDidDocAPI } from '../did-documents.js'
 
-const DOC_PATH = '/.well-known/did.json'
+export const DOC_PATH = '/.well-known/did.json'
+
+async function get(url: string, timeout = 5000): Promise<any> {
```
dholms, line 7: tests were failing because this request was just hanging so I added a timeout.

Feel like 5s is pretty reasonable, but lmk if you think it should be different

@pfrazee

pfrazee: Yeah that's fine

crypto/test/key-compression.ts
```diff
@@ -15,7 +15,7 @@ test('public key compression', async (t) => {
 
 test('works consistently', async (t) => {
   const pubkeys = []
-  for (let i = 0; i < 100; i++) {
+  for (let i = 0; i < 10000; i++) {
```
pfrazee, line 18: is this secretly a token miner

dev-env/src/key-manager/did.ts
```diff
@@ -0,0 +1,35 @@
...
+    type: 'AdxIdentityProvider',
+    serviceEndpoint: idProvider,
+  }
+  return {
+    '@context': [
+      'https://www.w3.org/ns/did/v1',
+      'https://w3id.org/security/suites/ecdsa-2019/v1',
+    ],
+    id: did,
+    controller: 'did:example:123456', // TODO what controller to put here?
```
pfrazee, line 27: I think we just leave `controller` empty. I'm, like, 75% sure the controller is about recovery.

did-sdk/src/index.ts
```diff
@@ -18,15 +17,10 @@ export async function resolve(didUri: string): Promise<ReadOnlyDidDocAPI> {
   throw new Error(`Unsupported did method (${didMethod}) ${didUri}`)
 }
 
-export async function createDidWebServer(port = 9999): Promise<DidWebServer> {
```
pfrazee, line 21: Did you mean to remove this? That's fine if so, you just have other code in this PR that still seems to use it.

pfrazee (approved):
👍Left a couple small comments but merge whenever you're ready.

crypto/test/key-compression.ts
dholms: hehe 

it probably doesn't _need_ to be that high. I did have an error in it from not left padding the y coord which would only pop up once every ~256 keys. So I figured why not throw 10000 at it to make sure 😅 
I'll drop it down to just 1000 for now

dev-env/src/key-manager/did.ts
dholms: ah cool 👍

did-sdk/src/index.ts
dholms: I did. I switched it out for a static method. I guess I forgot to rebuild so never got an error. oops

PR #144, opened 2022-06-16
Style the auth-lobby UI

pfrazee:
A little aesthetic verve for the auth lobby. Nothing final, but a little cleaner.

image

image

image


dholms (approved):
This is awesome 🙌

PR #148, opened 2022-06-28
API package

pfrazee:
TODOs:

- [x] API Client
- [ ] API Server
- [x] Schemas library (records, views)
- [x] Localization
- [ ] Tests


common/test/uri.ts
```diff
@@ -0,0 +1,322 @@
+import test from 'ava'
+import { AdxUri } from '../src/network/uri.js'
+
+test('Parses valid ADX URIs', (t) => {
+  //                 input   host    path    query   hash
+  type AdxUriTest = [string, string, string, string, string]
+  const TESTS: AdxUriTest[] = [
+    ['foo.com', 'foo.com', '', '', ''],
+    ['adx://foo.com', 'foo.com', '', '', ''],
+    ['adx://foo.com/', 'foo.com', '/', '', ''],
```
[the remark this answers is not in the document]
pfrazee: Whoops, checking into it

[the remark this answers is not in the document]
pfrazee: Fix pushed

api/src/index.ts
```diff
@@ -0,0 +1,159 @@
+import * as didSdk from '@adxp/did-sdk'
+import * as nameSdk from './names.js'
+import * as ucan from 'ucans'
+import { Keypair } from '../common/types.js'
+import * as t from './types.js'
+
+export class AdxClient {
```
dholms, line 7: Does AdxClient need to be a class? ie will it encapsulate some state? Otherwise maybe we just make it functions. that way when calling, you don't have to instantiate every time
```ts
const client = new AdxClient()
client.someMethod(...)
```

dholms:
Vverall this looks great to me :thumbsup: 
AdxUri class is much needed & very thorough
Api classes seem reasonable & nice to work with. i like the split of AdxClient & AdxRepoClient
I’m :100:% in favor of going in on “repo” over “user”

Didn't get to test out the REPL because of some TS compile issues. But the gist of it looks right to me

dev-env/src/key-manager/index.ts
```diff
@@ -1,15 +1,18 @@
...
-import axios from 'axios'
 import * as uint8arrays from 'uint8arrays'
 import * as crypto from '@adxp/crypto'
 
 import KeyManagerDb from './db.js'
 import { formatDidWeb } from './did.js'
 
-const DID_SERVER = 'localhost:2582'
-
 export const runServer = (db: KeyManagerDb, port: number): http.Server => {
```
dholms, line 16: I think every service's `runServer` method is a bit different. 
Some don't return anything.
Some return the `http.Server`
Some console.log the "listening" message
The DidWebServer returns a class that let's you interact with the server without making http calls.
(this incongruity is mostly my fault 😅)

What's your general preference on these? & we can work to make them sync up

pfrazee: I got them all mostly equivalent, there's just one (the did:web server I think) that isn't. We can do another pass later to unify more

pfrazee: One thing I did do was move all the console.logs outside of the APIs

dholms: Ah I guess i missed the moving the `console.logs` out.

Yeah i refactored the DidWebServer last week, but left it as a class you can interact with without http for the tests/etc. Wasn't sure if we'd want that when working in dev-env 🤷‍♂️

dholms:
for posterity's sake: https://hackmd.io/DXxwgcGZTt6tX-r_wGyD5A

dholms:
Let's get this bad boy in. Some of it will be useful for my data structure work & this PR is gonna blow up into a monstrosity if we don't merge it soon

dholms (approved):
Let's get this bad boy in. Some of it will be useful for my data structure work & this PR is gonna blow up into a monstrosity if we don't merge it soon

PR #151, opened 2022-07-07
Did doc diff

AaronGoldman:
Create the did doc diff library 
https://docs.google.com/document/d/1pT8FR6WyJNVVky1ed9wt_5M47LjeVxeFya5MUrx5_WI/
 
Example identity tick
```json
    {
      "tid": "3j55-hlo-v4g4-22",
      "diffs": {
        "3j55-hih-g24c-22": { "a": 1, "b": 2, "c": 3 },
        "3j55-hiu-hcxs-22": {
          "prev": "3j55-hih-g24c-22",
          "patches": [["put", ["d"], 4]],
          "key": "did:key:zDnaeYgdnK7nVH2xNQENrBRJbdCQ1KKjxt5sbykiUGLAh46Ez",
          "sig": "z58kG77wksPLi8185y6fsJG3NidUjYVBzBKVqSG3T5NQKwrNu9d6aL2UorPRzre4VCvKUASuec53jain2LRaMCtmZ",
        },
        "3j55-hk2-pbqk-22": {
          "prev": "3j55-hiu-hcxs-22",
          "patches": [["del", ["b"]]],
          "key": "did:key:zDnaeYgdnK7nVH2xNQENrBRJbdCQ1KKjxt5sbykiUGLAh46Ez",
          "sig": "z45gdspAmr9W1tC2KCBrnugdQSXP7fbTUvxz8yM1QnPPNLTKHs2PWzDYgmJe8RPLhZKrfpVFEsvaCMHkhLkFGfxRo",
        },
        "3j55-hki-vs6w-22": {
          "prev": "3j55-hk2-pbqk-22",
          "patches": [["put", ["e", "ea"], "each"]],
          "key": "did:key:zDnaeYgdnK7nVH2xNQENrBRJbdCQ1KKjxt5sbykiUGLAh46Ez",
          "sig": "z66WkmrsuVNiqBT3mhjF8G5MNcgpcKbCvXYinqFL73xYkwbSRvNNE7Mj8ak9gEN5FCkamWgsa82GrG984sJxcJdw8",
        },
        "3j55-hl6-3ctt-22": {
          "prev": "3j55-hk2-pbqk-22",
          "patches": [["put", ["Don't", "put", "me"], "in"]],
          "key": "did:key:zDnaeYgdnK7nVH2xNQENrBRJbdCQ1KKjxt5sbykiUGLAh46Ez",
          "sig": "z5oHy2qCSKR9Z9hXuha6c7qEpUWMzGNeZcj8D5kDAxsWP6P8FN4kmSWHcoLv4q5umiyMWcUk5CePGeh51khRGEgGL",
        }
      },
      "key": "did:key:zDnaesnjCLbxSB2EiJgTZUYCBqHTkuDJDdLbSmaAn2bpqXtuY",
      "sig": "z4Y3LwMCQcmgFvyGfMmSCa5Qr7vCr7tUUDWnRBQqzZsDpSCKP3ZB1WnQTUVBivruavBsCprhjJ5p4WBR1RTyeYqUd"
    }
```
  This makes
```json
    {
    "id": "did:aic:zwulnkrxrjkobowd",
    "a": 1,
    "c": 3,
    "d": 4,
    "e": { "ea": "each" },
    }
```
Dose not yet authorize the key?


dholms:
Gonna give a bunch of feedback just cause in early PRs it's important to be in sync on code styling/patterns/etc. Saves a lot of headaches later on so.... sorry for all the nits 😛

did-doc-diff/src/index.ts
```diff
@@ -0,0 +1,217 @@
...
+    {
+    id: 'did:aic:zwulnkrxrjkobowd',
+    a: 1,
+    c: 3,
+    d: 4,
+    e: { ea: 'each' },
+    }
+*/
+
+export async function tick_to_diffs(tick: {
```
dholms, line 51: do camelCase instead of snake_case. for function names & variable names. so
`tickToDiffs`

did-doc-diff/src/index.ts
```diff
@@ -0,0 +1,217 @@
...
+  sig: string
+  key: string
+  [index: string]: any
```
dholms, line 54: `any` completely blows open the Typescript type system. one `any` can have cascading effects throughout the codebase. It's sometimes useful in isolated contexts where the type system is punishing you. But it really should be isolated or massaged back into a type before it encounters anything.

Even a pretty open type like 
```ts
type Value = string | number | undefined | Record<string, Value> | Array<Value>
```
protects you a lot more than `any`

did-doc-diff/src/index.ts
```diff
@@ -0,0 +1,217 @@
+import * as crypto from '@adxp/crypto'
+import { TID, util } from '@adxp/common'
```
dholms, line 2: make sure this is in the `package.json`

did-doc-diff/src/index.ts
```diff
@@ -0,0 +1,217 @@
...
+  const base32 = uint8arrays.toString(data, 'base32')
+  const sort_order_invariant_base32 = base32.split('').map((c) => {
+    return S32_CHAR[B32_CHAR.indexOf(c)]
+  })
+  return sort_order_invariant_base32.join('')
+}
+
+async function pid(data: Uint8Array) {
+  // return the pid120 a 24 char pid
+  const hash = s32encode((await sha256.digest(data)).digest)
```
dholms, line 147: this works. but we also have a `sha256` in `@adxp/crypto` that would prevent the extra dependency

did-doc-diff/src/index.ts
```diff
@@ -0,0 +1,217 @@
+import * as crypto from '@adxp/crypto'
+import { TID, util } from '@adxp/common'
+import { sha256 } from 'multiformats/hashes/sha2'
+import * as uint8arrays from 'uint8arrays'
+
+/*
+  Example identity tick
```
dholms, line 7: This is very useful to have here 👍

did-doc-diff/src/index.ts
```diff
@@ -0,0 +1,217 @@
...
+    (doc['adx/account_keys'] && doc['adx/account_keys'].includes(didkey)) ||
+      (doc['adx/recovery_keys'] && doc['adx/recovery_keys'].includes(didkey)),
+  )
+  console.log('authorise_key:', valid)
+  return true
+}
+
+export async function diffs_to_did_doc(diffs: { [index: string]: any }) {
+  if (typeof diffs !== 'object') {
+    return undefined
```
dholms, line 97: `undefined`/`null` are weird in JS/TS. they _kinda_ mean the same thing.

my general rule of thumb is: use `undefined` when something just hasn't been set. use `null` when you're explicitly giving a null result

So I use `undefined` for optional fn parameters or optional values in a record. I use `null` for return values or when I want sometime to _explicitly_ acknowledge the fact that they're not providing a value

did-doc-diff/src/index.ts
```diff
@@ -0,0 +1,217 @@
...
+  },
+  key: crypto.EcdsaKeypair,
+) {
+  if (doc.key !== key.did()) {
+    throw 'passed in secret key and did do not match'
+  }
+  doc['sig'] = ''
+  const data = uint8arrays.fromString(JSON.stringify(doc)) // todo: canonicalise json/cbor?
+  const sig = await key.sign(data)
+  doc['sig'] = 'z' + uint8arrays.toString(sig, 'base58btc')
```
dholms, line 193: smart. i'm realizing now that I have some non-multibase sigs floating around in the codebase. would probably be good to make those base58 w the multibase in front

did-doc-diff/test/index.ts
```diff
@@ -0,0 +1,303 @@
...
+    privateKey:
+      '2EPbyvKQKUaUPMF7Mm94FjEzvs5tsLWfesyc97W1dqYeeZFEG2axhDPdH5LDfLF28yNjJu26wYXiSypZtaMWCajYu4hHgUvHhygV8kHzQkvszbL6BrvGP8GsYpeWBBWBxx4q816yVUBjzFH21mkekFnHF5fWAXDUN72rii66kR81rkaxCTacSMJu2XUuz',
+  },
+  {
+    exportable: true,
+    encoding: 'base58btc',
+  },
+)
+
+const key_tick = await crypto.EcdsaKeypair.import(
```
dholms, line 26: you're gonna hate me. i changed this import to JWK 😅

did-doc-diff/src/index.ts
```diff
@@ -0,0 +1,217 @@
...
+export async function sign(
+  doc: {
+    sig: string
+    key: string
+    [index: string]: any
+  },
+  key: crypto.EcdsaKeypair,
+) {
+  if (doc.key !== key.did()) {
+    throw 'passed in secret key and did do not match'
```
dholms, line 188: wrap this in an error:
```ts
throw new Error('passed in secret key and did do not match')
```

Javascript stupidly lets you throw whatever you want. But we should try to only throw errors

did-doc-diff/src/index.ts
```diff
@@ -0,0 +1,217 @@
...
+  [index: string]: any
+}) {
+  if (await validate_sig(tick)) {
+    return tick.diffs
+  } else {
+    return undefined
+  }
+}
+
+export async function tick_from_diffs(
```
dholms, line 63: `this` (the execution context of a fn) is another strangity of JS.

How `this` is handled varies between how a function is declared. It often doesn't matter, but occasionally it does.

I nearly always use arrow functions in my code because I've found that `this` is more predictable in that context.

The equivalent here is:
```ts
export const tickFromDiffs = async (
  diffs: { [index: string]: any },
  key: crypto.EcdsaKeypair,
): Promise<ReturnType> => {
  ...
}
```

Different people have different opinions on this. You can read about it here:
https://dmitripavlutin.com/differences-between-arrow-and-regular-functions/

dholms: It's also good to annotate what the return type of a function should be 👍

did-doc-diff/src/index.ts
```diff
@@ -0,0 +1,217 @@
...
+  for (const p of patches) {
+    const [op, path, value] = p
+    let node = doc
+    for (const segment of path.slice(0, -1)) {
+      if (op === 'put' && !(segment in node)) {
+        node[segment] = {}
+      }
+      node = node[segment]
+    }
+    if (op === 'put') {
```
dholms, line 169: I'd give some feedback if invalid `op`. maybe throw an error?

did-doc-diff/src/index.ts
```diff
@@ -0,0 +1,217 @@
...
+
+  // add the id feild
+  doc['id'] =
+    'did:aic:' +
+    (await pid(uint8arrays.fromString(JSON.stringify(diffs[last]))))
+
+  for (const tid of tids.slice(1)) {
+    const diff = diffs[tid]
+    if (!(await validate_sig(diff))) {
+      console.log('Warning ignoreing: bad sig')
```
dholms, line 114: If there's a bad sig in it, maybe we should freak out on the whole thing not just skip over it?

did-doc-diff/src/index.ts
```diff
@@ -0,0 +1,217 @@
...
+}
+
+export async function tick_from_diffs(
+  diffs: { [index: string]: any },
+  key: crypto.EcdsaKeypair,
+) {
+  // This function dose not validate the diffs remember to do that first
+  const tids = Object.keys(diffs).sort()
+  if (tids.length < 1) {
+    return undefined // ToDo: handel the case where aic has no diffs
```
dholms, line 70: usually do `@TODO` for this.

Some ide tools help highlight those for you, but just generally it's helpful to have one label for every todo to make it easily searchable

did-doc-diff/src/index.ts
```diff
@@ -0,0 +1,217 @@
...
+  }
+
+  const tick = {
+    tid: TID.next().formatted(),
+    diffs: diffs,
+    id:
```
dholms, line 76: is `id` in the tick? or just in the doc?

did-doc-diff/package.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "@adxp/did-doc-diff",
```
dholms, line 2: this is fine for now 👍 

im guessing we'll want to add more to this pacakge than just diffs. so something like: `@adxp/did-aic`

did-doc-diff/test/index.ts
```diff
@@ -0,0 +1,303 @@
...
+      key: key_account.did(),
+      sig: '',
+    },
+    key_account,
+  )
+  // console.log(diff)
+  t.assert(await validate_sig(diff))
+})
+
+test('tid next', async (t) => {
```
dholms, line 229: this is a good test to have, but maybe we put it in the tid test file in `common`?

did-doc-diff/src/index.ts
```diff
@@ -0,0 +1,217 @@
...
+    key: key.did(),
+    sig: '',
+  }
+  const signed_tick = sign(tick, key)
+  return signed_tick
+}
+
+function authorise_key(doc: { [index: string]: any }, didkey: string) {
+  const valid = Boolean(
+    (doc['adx/account_keys'] && doc['adx/account_keys'].includes(didkey)) ||
```
dholms, line 88: I think there are specific fields for stuff like this in the DID doc spec, but that might be a paul question 🧐

did-doc-diff/src/index.ts
```diff
@@ -0,0 +1,217 @@
...
+    }
+  }
+  const head = S32_CHAR[31 - twos]
+  const tail = hash.slice(twos, twos + 15) // 23 for pid120; 15 for pid80 // Are we ok with 80 bits?
+  return head + tail
+}
+
+function patch(doc: { [index: string]: any }, patches: Array<any>) {
+  for (const p of patches) {
+    const [op, path, value] = p
```
dholms, line 161: are there any constraints we should put on values?

and do we allow more complicated patches that aren't just `put`/`del`? such as `append` for adding something to a list?

dholms:
Alrighty I know I blew you up with a million comments. It's mostly style/formatting stuff that should be pretty quick to adopt. And now is the time for us to all sync up on stuff like this.

The meat of this looks good to me. The format of patches/diffs makes sense to me & the logic all seems sound 👍 

I'd say the main style thing is: don't forgo the type system! typescript unfortunately makes it too easy to turn it off by using `any`, but we should reserve that for particular cases & not use it generally.

We probably need to be a bit more rigorous with what sorts of values we allow in the DID doc. Do we want to actually allow arbitrary puts/dels or should we be forcing the user to conform to the DID doc spec? 

I also need to go back to the PID doc you put together. I'm still not quite sure why they're necessary here but I could be convinced.

lmk if you have any questions or anything!

did-doc-diff/src/index.ts
```diff
@@ -0,0 +1,217 @@
...
+function s32encode(data: Uint8Array): string {
+  // this is gross so many alocatons and funtion calls :(
+  const base32 = uint8arrays.toString(data, 'base32')
+  const sort_order_invariant_base32 = base32.split('').map((c) => {
+    return S32_CHAR[B32_CHAR.indexOf(c)]
+  })
+  return sort_order_invariant_base32.join('')
+}
+
+async function pid(data: Uint8Array) {
```
dholms, line 145: a test on this function would be nice too 👍

did-doc-diff/src/index.ts
AaronGoldman: diffs["oldest tid"] is the initial data for the did doc and could be any JSON object.
diffs["other tid"].patches[*][2] is is the value being put and is any JSON Value.

Is there a canonical typescript why to say json Values?

did-doc-diff/src/index.ts
AaronGoldman: the `did:aic:whatevs` could be calculated from the diffs but I include it in the tick signed by AIC.
This serves both the purpose that ticks can be searched for the ID and if the diffs are `{}` there is no initial state so we get a natural way to say as of tid this did is unknown to AIC

did-doc-diff/src/index.ts
```diff
@@ -0,0 +1,217 @@
...
+    doc = patch(doc, patches)
+    last = tid
+  }
+  return doc
+}
+
+const S32_CHAR = '234567abcdefghijklmnopqrstuvwxyz'
+const B32_CHAR = 'abcdefghijklmnopqrstuvwxyz234567'
+
+function s32encode(data: Uint8Array): string {
```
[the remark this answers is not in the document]
AaronGoldman: That one is for Number this one is for Uint8Array.
not sure how much of this should be moved to common

s32encode(data: number): string
s32encode(data: Uint8Array): string

did-doc-diff/src/index.ts
```diff
@@ -0,0 +1,217 @@
...
+    return S32_CHAR[B32_CHAR.indexOf(c)]
+  })
+  return sort_order_invariant_base32.join('')
+}
+
+async function pid(data: Uint8Array) {
+  // return the pid120 a 24 char pid
+  const hash = s32encode((await sha256.digest(data)).digest)
+  let twos = 0
+  for (twos = 0; twos < 32; twos++) {
```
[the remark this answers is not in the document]
AaronGoldman: I use the twos after the loop ends and goes out of scope.
if var preferred with function scope or let but at the hire scope.

did-doc-diff/src/index.ts
AaronGoldman: now if you want to append you put the whole list again.
I only allow navigating down objects 
```json
{"a": [{"b": "c"}]}
```
```json
["put", ["a", 0, "b"], "not c"]
```
will get mad you would need to do put the whole list
```json
[ "put", ["a"], [{"b": "not c"}] ]
```
we can loosen this restriction late but for now made the code a little simpler for the type checker

did-doc-diff/src/index.ts
AaronGoldman: https://github.com/multiformats/multicodec/blob/master/table.csv

I see the multicodec for public keys and secret keys but not for the signatures.
I wanted to have multicodec inside the multibase but fine. 
The did:key kinda gives away the signature type anyway.

did-doc-diff/test/index.ts
AaronGoldman: Yes I planed to move it but was useing to test that I could import from module to module I can move it now.

aic/src/index.ts
```diff
@@ -0,0 +1,404 @@
...
+type Diffs = { [index: string]: Document | Diff } // the first diff is arbatrary JSON object initial value
+type Tick = {
+  tid: TidString
+  did: string
+  diffs: Diffs
+  key: DidKeyString
+  sig: Signature
+}
+
+type Document = { [key: string]: Value }
```
pfrazee, line 79: We've been sticking most of the types for a module in the `./types.ts` file. You import them as-needed internally, but in your index.ts I always export all of them (`export * from './types'`) because consumers of the APIs often need to reference the types.

did-doc-diff/src/index.ts
pfrazee: I _think_ in the case that you have an actually "open" value the ideal is to use `unknown`. That can then get cast via runtime type-checking.

did-doc-diff/src/index.ts
pfrazee: Yeah there are fields in DID documents that should be used here, unless we're intentionally going off the DID doc spec for internal storage?

did-doc-diff/src/index.ts
dholms: yup^^

did-doc-diff/src/index.ts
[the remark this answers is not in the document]
dholms: maybe:

s32encodeInt(data: number): string
s32encode(data: Uint8Array): string

ideally `s32encodeInt` even calls `s32encode` so we make sure they're consistent

did-doc-diff/src/index.ts
dholms: yea im not sure there is a multicodec for signatures 🤔 
having the public key that goes with it should work

packages/server/src/routes/aic/index.ts
```diff
@@ -0,0 +1,125 @@
+import express from 'express'
+import tid from './tid'
+import { TID } from '@adxp/common'
+import { update_tick } from '@adxp/aic'
+import * as crypto from '@adxp/crypto'
+import { z } from 'zod'
+
+// Note: do not use the dev/test key in production
+//       pull the prod key from the secret store
+export const CONSORTIUM_KEYPAIR = crypto.EcdsaKeypair.import(
```
dholms, line 10: hey I'm pretty sure we want the AIC server outside of the PDS. Most folks that run a PDS do not want to run AIC. And we'll probably even run them as separate infrastructure.

also in terms of startup from the cli: I think we'll want that in the `dev-env` package but not the `cli` package

did-doc-diff/src/index.ts
dholms: @AaronGoldman 
bump^^ we really do want to standardize on camelCase in code

AaronGoldman: yup I will move them all to camelCase

did-doc-diff/src/index.ts
AaronGoldman: I can use unknown for what comes from the http post 
then cast as I check it before passing to functions that expect a Value

did-doc-diff/src/index.ts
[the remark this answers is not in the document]
AaronGoldman: for now I think it is fine to keep it separate so aic is stand alone

did-doc-diff/src/index.ts
AaronGoldman: thanks, used  it

did-doc-diff/test/index.ts
AaronGoldman: updates to use JWT 😛

packages/server/src/index.ts
```diff
@@ -25,6 +25,7 @@ const run = async () => {
   }
 
   if (dbLoc) {
+    console.log('Database:', dbLoc)
```
dholms, line 28: 🔪

packages/aic/src/doc/signature.ts
```diff
@@ -0,0 +1,49 @@
...
+      typeof doc.key !== 'string' ||
+      doc.key.length < 1
+  ) {
+      return false
+  }
+
+  const doc_copy = JSON.parse(JSON.stringify(doc))
+  if (doc_copy.sig[0] !== 'z') {
+      // check the multibase https://github.com/multiformats/multibase
+      throw "signatures must be 'z'+ base58btc" // maby just retuen false?
```
dholms, line 40: wrap in error

packages/aic/src/low_pid.ts
```diff
@@ -0,0 +1,28 @@
+import { pid } from './pid'
+
+const run = async () => {
```
dholms, line 3: I don't think we want this in `src`

I made a separate `bench` command in `common` that runs through jest. may be good to model it off of

packages/aic/src/types.ts
```diff
@@ -0,0 +1,41 @@
...
+    | number
+    | string
+    | Value[]
+    | { [key: string]: Value }
+    | undefined
+
+export type TidString = string
+export type Signature = string
+export type DidKeyString = string
+export type Patch = ['put' | 'del', (string | number)[], Value]
```
dholms, line 19: Similar to our MST encoding convo: do we want to avoid heterogenous arrays/tuples?

Should we prefer something like
```ts
type Patch = {
  op: 'put' | del'
  key: string[]
  val: Value
}
```

packages/aic/src/types.ts
```diff
@@ -0,0 +1,41 @@
...
+export type Diff = {
```
dholms, line 20: Does a diff need to include _its_ TID?

Cause ticks don't stick around, right?

packages/aic/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "lint": "eslint . --ext .ts,.tsx",
+    "lint:fix": "yarn lint --fix",
+    "verify": "run-p prettier lint",
+    "verify:fix": "yarn prettier:fix && yarn lint:fix",
+    "build": "node ./build.js",
+    "postbuild": "tsc --build tsconfig.build.json",
+    "low": "node dist/low_pid.js"
+  },
+  "dependencies": {
+    "@adxp/common": "*",
```
dholms, line 20: remove `common` if we're not using

looks like we're keeping `crypto` in here for now? I like that for now 👍 
Once we break up the `common` package a bit, we might be able to pull in just the TID part without the cruft

packages/aic/src/index.ts
```diff
@@ -0,0 +1,42 @@
+import * as crypto from '@adxp/crypto'
+import { database, Database } from './db'
+import { server } from './server'
+
+const run = async () => {
+  // create or open db
+  const dbLoc = 'adx.sqlite'
```
dholms, line 7: we may want this to be more configurable (similar to the PDS) where I can pass in which ever database instance I like - an in-memory DB for instance

packages/aic/src/db.ts
```diff
@@ -0,0 +1,69 @@
...
+    tickForDid: ((did: string) => Promise<{ did: string; tid: string; tick: string } | undefined>),
+    putTickForDid: ((
+        did:string,
+        tid:string,  // the new tid must me larger then the previous tid
+        prevTid:string,  // gard: if the prevTid has changed the tick from the db is stale
+        tick:string  // the tick is a json string
+    ) => Promise<void>),
+}
+
+export const database = async (location: string): Promise<Database> =>  {
```
dholms, line 20: Bigger tabs than elsewhere in the repo. Think prettier should fix this for you 🤔 
I can help get that set up if you like.

I also ran `yarn lint` & it popped up some linting errors that should be fixed (nothing major).

I may need to take a look again at how these tools interact w one another & if we can just go all in on one 🤔 

It's nice to have them setup with auto-formatting (similar to go) so you just don't ever have to think about it

packages/aic/src/document.ts
```diff
@@ -0,0 +1,318 @@
...
+  // check that the diff.patches[*][2]
+  // the recovey key is only authorised to update account or recovery keys
+  // only changes 'adx/account_keys' or 'adx/recovery_keys'
+  return patches.every((patch: Patch) => {
+    const pathSegment = patch[1][0]
+    return ( pathSegment === 'adx/account_keys' || pathSegment === 'adx/recovery_keys' )
+  })
+}
+
+export const tickToDidDoc = async (tick: Tick, consortiumDid: string, key: Asymmetric, asOf?: TidString,): Promise<Document> => {
```
dholms, line 219: Reading this function signature (& others in this file), it gives me the impression that it _always_ returns a Document. Either need to return `Promise<Document | Error>` or just throw the error.

Coming from Go, I always wanted to return errors. But I know use throw/catch much more often. It prevents you from screwing up error checking (cause the error will always bubble up & not get swallowed)

packages/aic/src/routes.ts
```diff
@@ -0,0 +1,164 @@
...
+      },
+      await consortiumCrypto(),
+    ),
+  )
+})
+
+router.get(
+  '/:pid([234567abcdefghijklmnopqrstuvwxyz]{16})/',
+  async function (req, res) {
+    // AIC git
```
dholms, line 78: get

packages/aic/src/routes.ts
```diff
@@ -0,0 +1,164 @@
...
+    const row = await res.locals.db.tickForDid(did) // retreve latest tick
+    const prev_tick = row ? JSON.parse(row.tick) : null
+    const signedDoc = await updateTick(
+      did,
+      tid(),
+      null, // candidateDiff: this is a get no updates
+      prev_tick,
+      await consortiumCrypto(),
+    )
+    res.type('json').send(signedDoc)
```
dholms, line 92: We need to send a proper status code if there's an error. Even just a 500. But right now we're sending a `200` with an error msg

packages/aic/src/routes.ts
```diff
@@ -0,0 +1,164 @@
...
+      prev_tick, // tick from db for did:aic
+      await consortiumCrypto(), // server's aic key
+    )
+
+    if ('error' in newTick) {
+      // error return from updateTick
+      // type ErrorMessage = { error: string; cause?: ErrorMessage; [index: string]: Value }
+      // if there is an error property it is an error don't store in DB
+      // also send the last tick that is still valid
+      res.status(200)
```
dholms, line 122: should this be a 200?

& this is sorta my point on error handling.

`if ('error' in newTick) {` _works_, but it's not the most idiomatic way of doing it. & if it's not captured by either the type system or throw/catches, I can see errors slipping by & causing issues

packages/aic/src/db.ts
dholms, line 20: haha re: the convo we just had, this seems like a nice place for a class 😉

packages/aic/src/document.ts
```diff
@@ -0,0 +1,318 @@
...
+  did: string,
+  tid: TidString,
+  candidateDiff: Document | null,
+  key: Asymmetric,
+): Promise<Tick | ErrorMessage> => {
+  if (candidateDiff === null) {
+    // since candidateDiff is null
+    // this is a request for a fresh tick of a non-exixtant did
+    // return a signed affirmation that the did dose not exist
+    return (await sign(
```
dholms, line 93: I'm not sure we need to be signing error messages.

I understand the appeal for ticks. But for errors, isn't this handled by HTTPS?

packages/aic/src/document.ts
```diff
@@ -0,0 +1,318 @@
...
+    return registerNewDid(did, tid, candidateDiff, key)    
+  }
+
+  // since stored_tick is not null this is an update
+  // of a registered did:aic
+  if (!(await validateSig(stored_tick, key))) {
+    // this means the consortium data base has been corupted
+    return { error: 'bad signature: stored tick' }
+  }
+  if (stored_tick.key != key.did()) {
```
dholms, line 66: kind of a nit cause I don't think it matters here, but I'd get in the habit of always doing `!==` for equality checks.

dholms:
General comments:

- the logic all seems sound here 👍 
- needs some cleanup. clearing out old comments, logs, etc. there's a lot of misspelled words in comments. I know there are spell checkers in vscode that may be able to help out so it isn't any more overhead (i kinda wanna do this as well)
- I think we may want more DID document semantics in here. Should users be able to do arbitrary updates if they break the DID document format? My gut says no 🤔 
- errors are very important in this context & im worried they're going to get eaten the way we're handling it right now. Would feel more comfortable encapsulating them in a tuple/either format (like go or haskell) or sticking with the idiomatic JS throw/catch model. They should require an explicit check (or make you _explicitly_ ignore them like Go).
- status codes are important & we should return error codes when a request is rejected for some reason.
- tests could be cleaned up a bit. a big reason I like tests is for documentation purposes & I have an easier time understanding the code than I do the tests here

packages/aic/src/index.ts
AaronGoldman: thanks, though I had put that back to an ENV var

packages/aic/src/document.ts
AaronGoldman: With the throw pattern we would still need to build an error JSON object at the end to return to the client.
I was using throw for errors for the server to deal with and a JSON Document for the Errors that go to the client.

What is the pattern you like to keep these clear what we are doing?

packages/aic/src/types.ts
AaronGoldman: The diff dose not know its own tid at the time it is signed.
The tid is assigned by the consortium when it puts the diff in the order.

packages/aic/src/document.ts
AaronGoldman: The real problem here is that it is not really an error.

It is an affirmative statement that the DID is not registered.
This is useful as a cacheable value that the DID is not registered but if the DID has never been updated in any way then it is still in the initial state.
That means that it can be validated by just using the hash of the initial state.

If I want to keep my ID private until I need to do the first update then the only info a validator needs to know that they got the current version of the doc it that there are no updates which this message proofs.

PR #152, opened 2022-07-08
Tooling refactor + @ucans/core integration

dholms:
Please forgive this monstrosity of a PR 👹 

As I was pulling ucans core in, I was running into tooling issues & decided to revamp everything.

Whole new monorepo setup + build tools + testing tools.

Alrighty things that happened:
- switched out `tsc` builds for `esbuild` this bundles everything as one package and will be significantly better in web. esbuild is fast, handles both cjs and esm, and can output as cjs/esm/umd.
- switched out Ava for Jest. Jest gives more in-depth feedback and has better TS integration so doesn't freak out on a tsc error
- switched out module types to cjs. This was necessary for Jest. ESM is a lot newer and has some rough edges that were causing headaches. also after beating my head against an ESM module this last week, I think we should try not to subject developers that our using our tools to that until the ecosystem is moer mature
- switched out `react-scripts` for Vite. This is faster, runs on esbuild, more easily configurable, etc
- moved most of the config to the root: eslint, prettier, etc. I also moved the root jest config to the root as well as the root tsconfig. each module imports & extends the jest config + tsconfig
- used a `composite` tsconfig with references to each package in it. This means that packages _directly_ reference the typescript code of the packages that they rely on. So for instance `@adxp/auth` directly builds in the `src/*.ts` of `@adxp/crypto` instead of building in the already compiled `dist/*.js`. Better bundling & no more building submodules and restarting your TS server!
- added in Lerna which helps with monorepo management. More shared dependencies, and build tools for buliding the entire repo. Now when a user downloads the repo and runs `yarn` in root, it installs all dependencies & builds all packages.
- This PR also includes the initial work that was being down before i launched into the land of sleeky smooth yaks: it integrates an alpha version of the new `@ucans/core` library. We don't rely on any of the crypto from the original library right now, only the crypto we bring ourselves. It's totally pluggable so we can add in did resolution methods or new crypto at will. I also _only_ pull this library into `@adxp/crypto` and `@adxp/auth` and then build + export abstractions around it. All of our auth semantics are directly in the auth library so developers don't have to worry about them or understand them. This keeps downstream developers from having to mess with the nitty gritty of UCANs which can be overwhelming at first.


pfrazee:
Superb work. This is going to save us a lot of work in the long-run. Well done!

pfrazee (approved):

PR #158, opened 2022-08-09
Repo Rework

dholms:
Reworking Repo implementation to use some of our new concepts
- integrating MST
- updating repo & collection semantics
- new layout for ucan capabilities as `repo/namespace/dataset/record`
- improved diffing
- safe commits on repo
- batch writes on repos

---

A few design decisions that I'd like feedback on:

## Mutable Repo

The MST is immutable. So to do a mutation, you must do:
`mst = await mst.add(...)`
instead of
`await mst.add(...)`

This has served me very nicely and works quite well because, by and large, it won't be exposed to developers
However, I decided to make the **repo stateful & mutable**.
So `await repo.getCollection('bsky/posts').addRecord(...)`, adds the record to the blockstore, adds it to the datastructure, formats the commit, updates the head, and returns the `TID` of the record.

We _may_ even want it to automatically push to the server.

A big part of the reason I did this is that at Fission, we split out the "edit" and "save/publish" functionality & app developers _consistently_ forgot to save/publish. We got so many bug reports about this. And even Fission devs were forgetting to do it.

As a result of this, I also implemented a `safeCommit` function to ensure that races don't end up in one commit getting clobbered.

## Collections API

I decided to mimic the API that @pfrazee put together for interacting with the server.
- The repo lets you do "repo level" actions and batch writes
- To do puts/deletes/posts/gets, you interact with the collection
(You can see how these semantics play out in `repo.test.ts`

I went back and forth on if a collection should be a separate concept from repo.
ie
`repo.getCollection('bsky/posts').addRecord({})`
vs
`repo.addRecord('bsky/posts', {})`

I still don't have a super strong opinion on this. Any strong thoughts one way or the other? There's nothing stopping us from having both available I suppose. But I like the simplicity of "one right way to do something"

## Collection-naming/UCAN semantics
We need a nomenclature for referring to collections.  I like Aaron's breakdown of `collection = namespace/dataset`.

I decided to make _every_ collection a `namespace/dataset` (so no un-namespaced collections & no deeply nested collections). I like how straightforward but flexible this is.

The UCAN on the root of the repo now has the `issuer` as the application key & the `audience` as the root DID. It originally just required a UCAN with the audience as the application key. I like the way that this closes the loop, but it does introduce some extra size. I briefly chatted with Phillip (from Fission) about this & he generally agreed with my sentiments here, but I'll likely chat a bit more with the Fission team because I know they're thinking of sticking their UCANs _on_ the actual data structure as well.

## Sync protocol
I pulled out any "helper information" stored at the root (new cids, semantic deltas). This slows down the sync protocol a decent bit (it has to diff the MST for _every_ commit). We'll likely want to add it back in but as a separate index _alongside_ the repo.

## Other information on the root
I was _nearly_ tempted to have the root simply be the root of the MST. However, we need to point back to the prev commit & to the auth token for the tx in that node. So I ended up sticking with a repo root node _above_ the MST root. 

I also included the repo's DID in there as well. Including the DID seems like a nice way to close the loop: "The PDS says this is the repo for this DID & the repo says it's the repo for this DID". That being said, it seems silly to replicate the DID in _every_ repo root even though it isn't changing often (ever?). But referencing it by CID may actually be _longer_ than just plaintext.

I'm trying to think if there is any other info/metadata we may want at the root 🤔 

Closes https://github.com/bluesky-social/adx/issues/88


packages/auth/src/capabilities.ts
```diff
@@ -1,28 +1,38 @@
...
 import * as ucan from './ucans/index'
 
 export const writeCap = (
   did: string,
+  namespace?: string,
   collection?: string,
-  schema?: string,
   record?: string,
 ): ucan.Capability => {
   let resource = did
```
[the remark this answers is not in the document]
dholms: I think for capabilities, we should always encode them using the DID. It has a nice semantic simplicity: "I have the authority to write to this repo because I have a token originating from the root DID of the repo". Doesn't require going out to identity providers.

Agreed on needing good terminology here. I like `collection = namespace + dataset` 👍

packages/auth/src/auth-store.ts
```diff
@@ -91,6 +97,51 @@ export class AuthStore implements Signer {
       .build()
   }
 
+  async createUcanForCaps(
+    audience: string,
+    caps: ucan.Capability[],
+    lifetime = MONTH_IN_SEC,
```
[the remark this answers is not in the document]
dholms: Just needed a timeframe. And 1 month felt like a good amount of time before expiry 🤷‍♂️ 

We can totally do longer if we like

packages/common/tests/repo.test.ts
```diff
@@ -1,127 +1,71 @@
...
-    const commit = await repo.getCommit()
-    const verified = await auth.didPlugins.verifySignature(
-      repo.did,
-      commit.root.bytes,
-      commit.sig,
-    )
-    expect(verified).toBeTruthy()
-  })
+  it('does basic operations', async () => {
+    const collection = repo.getCollection('bsky/posts')
```
dholms, line 21: this test has basic usage of a repo.

You get a collection & then do operations on that collection. These then call back to the repo to do a safe commit

packages/common/src/repo/repo.ts
```diff
@@ -0,0 +1,455 @@
...
+    }
+    return new Collection(this, parts[0], parts[1])
+  }
+
+  // The repo is mutable & things can change while you perform an operation
+  // Ensure that the root of the repo has not changed so that you don't get local branching
+  async safeCommit(
+    mutation: (data: DataStore) => Promise<DataStore>,
+  ): Promise<void> {
+    if (this.authStore === null) {
```
pfrazee, line 145: Do you need to wrap this in a mutex? Line 149 applies the update to `this.data` but it's not updated until after a lot of async at 169, which I suspect could lead two `safeCommit()` calls to clobber each other.

dholms: I generally try to stay clear of mutexes. The `DataStore` is immutable so the mutation doesn't mutate the passed in object at all, it just returns a new object with the mutation applied. So it does all the async work but doesn't mutate the _actual repo_ until it's done the check to ensure it's in the same state it thought it was.

Remembering that I meant to write an async test for this to make sure `safeCommit` is working so i'll add that in

pfrazee: > The DataStore is immutable so the mutation doesn't mutate the passed in object at all, it just returns a new object with the mutation applied.

I could be wrong here -- but I think that's actually the problem. Two safeCommits running at the same time will apply to the same starting state rather than running sequentially, causing one to clobber the other.

pfrazee: Turns out I missed the commit guard at https://github.com/bluesky-social/adx/pull/158/files#diff-167949500561ff7a2113519f05ddcd4d6005dac965e67c9b292116d617b6e0d0R165-R167, this is fine

PR #161, opened 2022-08-19
fix: Fixed requests api path

1grzyb1:
I noticed that after API changes requests paths were not updated and because of that tests in `indexer.test.ts` were failing.
What do you think about putting those paths in common to avoid those issues in the future?


dholms:
Hey @1grzyb1 thanks for the changes + PR. We're currently in the process of revamping the protocol & PDS. The changes that I made to auth & the repo (along with the changes Paul made in speccing out the v1 api) broke the PDS, and instead of applying an intermediary fix, I left it until I could do a proper refactor. 

That's currently underway over here: https://github.com/bluesky-social/adx/pull/162

I probably would have pulled these changes in a couple months ago when I punted on it. But I think doing so now will just cause headaches when I merge the PDS revamp in. I'll close this in the mean time, but feel free to carry on the discussion if you like.

Thanks again though for taking initiative on this 😄 

> What do you think about putting those paths in common to avoid those issues in the future?
Yes! I'm a big fan of this. And in fact, we're pulling types out into their own package as well, that way we have a type safe API.

PR #162, opened 2022-08-19
PDS revamp

dholms:
# PDS Revamp ⚡ 

- Switch DB library to TypeORM instead of Knex (moar types!)
- Move [schemas](https://github.com/bluesky-social/working-schemas) to monorepo `microblog` package
- Add indexing layer to DB with tables for each record schema in the starting set
- Create view queries for each view schema in the starting set
- Add `repo` & `data` routes for [v1 API](https://hackmd.io/DXxwgcGZTt6tX-r_wGyD5A)
  - one change is that I included routes for single writes, instead of just batch writes - create/edit/delete a record. Let me know if you disagree with this inclusion & think we should only have the one route for mutations
- Remove old server routes
- Added a `MicroblogClient` that uses the `AdxClient` but gives semantic & typed methods
- Tests for CRUD operations & computed views

### On indexing
We index records instead of maintaining & updating computed views. 
_Every_ record is indexed in the `records` table, with its `did`, `collection`, `tid`, and `raw` record JSON. 
Every record that passes validation for its given collection (that we have the schema for) is indexed according to that schema. This allows us to calculate views over the data.
When requesting an update, a client can _optionally_ ask for validation (defaults to `true`). In this case, the server will perform schema validation on the record _prior to_ writing it to the repo or indexing it.

### Some supporting changes
- `did:example` allows _any_ key to sign on it's behalf (for dev purposes)
- repo validation logic improved so that a root signature can be signed by the root DID
- return dataset/namespace from AdxUris - _all_ adx uris should be of the form `adx://did/namespace/dataset/tid`

### Not included:
- sync/federation
- did:aic (currently using did:example for root dids)
- blobs / embedded media
- server auth (currently the client just sends their DID in the authorization header)
- notifications


packages/cli/src/bin.ts
```diff
@@ -2,7 +2,7 @@
 
 import minimist from 'minimist'
 import chalk from 'chalk'
-import './commands/index.js'
+// import './commands/index.js'
```
dholms, line 5: removed this so cli builds fine, we'll probably want to deprecate this package

packages/common/src/repo/repo.ts
```diff
@@ -76,8 +80,8 @@ export class Repo {
   }
 
   static async load(blockstore: IpldStore, cid: CID, authStore?: AuthStore) {
-    const commit = await blockstore.get(cid, schema.commit)
-    const root = await blockstore.get(commit.root, schema.repoRoot)
+    const commit = await blockstore.get(cid, def.commit)
```
dholms, line 83: changed all previously used `schema`s (to describe `zod` object shapes) to `def` so we don't overload the term "schema"

packages/microblog/src/client.ts
```diff
@@ -0,0 +1,284 @@
...
+          tag,
+        },
+        subject: did,
+        createdAt: new Date().toISOString(),
+        ...hardcode,
+      })
+    return uri
+  }
+
+  async acceptBadge(badgeUri: AdxUri): Promise<AdxUri> {
```
dholms, line 166: This is the profile mutation that would probs make sense as a procedure 🤔

packages/server/src/db/index.ts
```diff
@@ -1,412 +1,212 @@
...
-  AccountInfo,
-  TimelinePost,
-  Follow,
-} from '@adxp/common'
-import knex from 'knex'
-import { CID } from 'multiformats'
-import * as schema from './schema'
-import { KnexDB } from './types'
-import { ServerError } from '../error'
+import { Badge, Follow, Like, Post, Profile, Repost } from '@adxp/microblog'
```
dholms, line 1: this diff wasn't getting rendered but it's one of the most important files

packages/server/src/db/views/postThread.ts
```diff
@@ -0,0 +1,155 @@
...
+): PostThreadView.Post => {
+  return {
+    uri: row.uri,
+    author: {
+      did: row.authorDid,
+      name: row.authorName,
+      displayName: row.authorDisplayName || undefined,
+    },
+    record: JSON.parse(row.rawRecord),
+    parent: parent ? { ...parent } : undefined,
```
dholms, line 138: There's a tricky thing here where a `Post` contains references to both the children & the parent which creates a circular reference in the JSON that can't be stringified.

So we clone the parent to break that circular reference.

Eliminating the circular reference means that you can't step into children & then back to the parent without losing information. Not sure the best way to deal with this except for reconstructing the references on the client.

Or maybe we just don't care about the children of a parent & if we need that info we can grab that view.

Not sure if this comment makes sense, lmk if I should clarify 😅

packages/server/tests/views.test.ts
```diff
@@ -0,0 +1,266 @@
...
+    await dan.followUser(bob.did)
+  })
+
+  it('makes some posts', async () => {
+    const alice0 = await alice.createPost(posts.alice[0])
+    const bob0 = await bob.createPost(posts.bob[0])
+    const carol0 = await carol.createPost(posts.carol[0])
+    const dan0 = await dan.createPost(posts.dan[0])
+    const dan1 = await dan.createPost(posts.dan[1], [
+      {
```
dholms, line 64: not in love with this way of creating user mentions, but also don't see a great alternative. thoughts?

packages/microblog/src/client.ts
```diff
@@ -0,0 +1,284 @@
...
+        description,
+        ...hardcode,
+      })
+    return uri
+  }
+
+  async createPost(
+    text: string,
+    entities?: Post.Entity[],
+    hardcode: Partial<Post.Record> = {},
```
pfrazee, line 65: What's this `hardcode` pattern?

packages/microblog/src/client.ts
pfrazee: Yeah makes sense

packages/server/src/db/index.ts
```diff
@@ -1,412 +1,212 @@
...
-  ): Promise<void> {
-    await this.db('posts').where({ tid, author, namespace }).del()
-  }
-
-  async postCount(author: string): Promise<number> {
-    const res = await this.db('posts').count('*').where({ author })
-    const count = (res[0] || {})['count(*)']
-    if (typeof count !== 'number') {
-      throw new ServerError(500, 'Unable to retrieve post count')
+    const table = this.db.getRepository(RepoRoot)
```
pfrazee, line 80: lol bad luck that TypeORM calls those "repositories"

packages/server/src/db/views/postThread.ts
pfrazee: Yeah that makes sense. The way I was handling this (IIRC) is that I only set `parent` for ancestors of the target post, and `children` for descendants. Does that make sense?

pfrazee (approved):
Excellent work as always. Didn't get in-depth with anything but I scanned every line, and it's all well-organized and fits with my understanding. Well done!

packages/microblog/src/client.ts
dholms: There could probably be another term for it. But it essentially lets you "hardcode" certain attributes on records that aren't covered by the semantic interface. So for instance, providing a different `createdAt`.

I'm not totally sure it's necessary. I added it in when I was going to create records with deterministic timestamps but then I ended up not pursuing that idea.

packages/server/src/db/index.ts
dholms: haha I know 😅

packages/microblog/src/client.ts
pfrazee: Yeah makes sense. Okay up to you.

dholms: decided to ditch this pattern. I think it adds unneeded complexity & if you need to drop down into hardcoding certain record values, you can always do it by just using the AdxClient

PR #163, opened 2022-09-02
Misc AIC maintenance

devinivy:
I performed a bit of maintenance on the AIC service, building on top of the work in #151.  Here's what you'll find in here:
 - Added missing eslint-config-prettier dep to fix an issue with running the lint commands.
 - Fixed type- and lint-related errors/warnings.
 - Some misc. localized refactorings.
 - Based all error-handling on throwing errors.
 - Removed boilerplate around turning JWKs into crypto helpers, and made the consortium's crypto helpers available in route handlers.
 - Moved the route to get a tick to `GET /:pid/tick`, and repurposed `GET /:pid` for getting the DID doc itself.
 - Updated the tests to account for all of the above :)

Due to some prettier whitespace reformatting, I'd recommend reviewing [with whitespace hidden](https://github.com/bluesky-social/adx/pull/163/files?&w=1) 😏


packages/aic/src/document.ts
```diff
@@ -57,50 +57,51 @@ makes
...
   candidateDiff: Document | null,
   key: Asymmetric,
-): Promise<Tick | ErrorMessage> => {
+): Promise<Tick> => {
   if (candidateDiff === null) {
     // since candidateDiff is null
-    // this is a request for a fresh tick of a non-exixtant did
-    // return a signed affirmation that the did dose not exist
+    // this is a request for a fresh tick of a non-existent did
+    // return a signed affirmation that the did does not exist
```
AaronGoldman, line 104: Since we changed the errors to throws this is not an error

We should change the API from 
```
{
        tid: tid,
        did: did,
        error: 'not found',
        key: key.did(),
        sig: '',
      },
```
To
```
{
        tid: tid,
        did: did,
        diffs: {},
        key: key.did(),
        sig: '',
      },
```
It's not an error it's a affirmative claim that there are no stored diffs even the origin diff

PR #166, opened 2022-09-12
PLC server

dholms:


pfrazee:
Eugh github just lost my comment. Basically was just saying, make sure to build some flexibility into the oplog around services and key material, in case we need to support more of them or add some attributes (eg services in did documents have an `id` and `type` field).

Somewhere we'll need a method which outputs the DID Document format. That can be here or in our did-sdk, whichever you think that makes sense.

PR #168, opened 2022-09-13
Remove deprecated packages: Auth-Lobby, Awake, Example-App, WS-Relay

pfrazee:
- Moving to a new key-management scheme deprecates Auth-Lobby, Awake, and WS-Relay.
- Removing Example App because it's no longer needed.

Some updates still required:

- Auth library still has some code related to the lobby and awake proofs.
- Dev Env needs to be brought up-to-date.
- CLI needs to be brought up-to-date or deprecated.


dholms (approved):
Looks good 👍 

Cleared out a couple of other things as well

PR #173, opened 2022-09-13
Lock uint8arrays

dholms:
`uint8arrays` included a breaking change in `3.1.0` that used buffers under the hood instead of uint8arrays.

This locks uint8arrays at `3.0.0` for both out libraries and the ucans lib.

Had to change a couple interfaces for the upgraded ucans lib


packages/dev-env/package.json
```diff
@@ -25,13 +25,10 @@
...
     "get-port": "^6.1.2",
     "level": "^8.0.0",
-    "sqlite3": "^5.0.11",
-    "ucans": "^0.9.1",
-    "uint8arrays": "^3.0.0"
+    "sqlite3": "^5.0.11"
   },
   "devDependencies": {
     "@types/express": "^4.17.13",
-    "@types/inquirer": "^8.2.1",
```
dholms, line 34: I saw you removed `inquirer` so I also removed the types. That good?

pfrazee: Yep!

pfrazee (approved):

PR #174, opened 2022-09-13
Additional documentation updates

pfrazee:
This PR updates our documentation by:

- Removing old information from the readme
- Reorganizing docs into the `/docs` folder
- Adding some initial specs
  - ADX URI
  - HTTP Routes
  - Name Resolution
 
For the HTTP Routes I used the Swagger format and converted it to markdown. We should evaluate whether the Swagger format would be useful for other forms of code-generation to help the codebase align with the spec at all times.


dholms (commented):
A few ideas/notes:

- Remove this line from the README:
> This README still contains the instructions for that demo.
- Add `js` to the code block in name-resolution.md so that we get some highlighting
- `schema-id` is discussed in `adx-uri.md` but isn't in the URL scheme and doesn't link to anything (is this a future doc?)

pfrazee:
Yep done.

PR #175, opened 2022-09-13
Add docs generator and improve http-routes.md generation

pfrazee:
- Add a `yarn build` to the docs directory to generate docs.
- Improve the generated `http-routes.md` generation. Still not perfectly happy with it, but it's better than before.


dholms (approved):
Nice the generated docs look good & +1 on generation

Altho question: if we’re going to this trouble, why not just use Swagger? I feel like it ends up giving nicer docs than we can get out of markdown :thinking_face: 

Also just curious, are all the `.def` files from somewhere in particular?

pfrazee:
> Why not just use Swagger?

What do you mean? I'm using swagger/openapi definitions, but just trying to generate a markdown spec we can put in our repo. Is there a better option?

> Also just curious, are all the .def files from somewhere in particular?

Yeah that comes from https://github.com/Mermade/widdershins

dholms:
Ah sorry I meant Swagger UI: https://swagger.io/tools/swagger-ui/ which is a nice friendly UI that you basically get for free from the yaml.

I can see the benefit to this living in the repo as well though.

pfrazee:
Oh yeah eventually we should look at that. Going to merge for now.

PR #177, opened 2022-09-15
Add NSID package

pfrazee:
- Adds `@adxp/nsid`

### DIFF
--- a/packages/nsid/package.json
+++ b/packages/nsid/package.json
@@ -1,0 +1,19 @@
+{
+  "name": "@adxp/nsid",
+  "version": "0.0.1",
+  "main": "src/index.ts",
+  "scripts": {
+    "test": "jest",
+    "prettier": "prettier --check src/",
+    "prettier:fix": "prettier --write src/",
+    "lint": "eslint . --ext .ts,.tsx",
+    "lint:fix": "yarn lint --fix",
+    "verify": "run-p prettier lint",
+    "verify:fix": "yarn prettier:fix && yarn lint:fix",
+    "build": "esbuild src/index.ts --define:process.env.NODE_ENV=\\\"production\\\" --bundle --platform=node --sourcemap --outfile=dist/index.js",
+    "postbuild" : "tsc --build tsconfig.build.json"
+  },
+  "license": "MIT",
+  "dependencies": {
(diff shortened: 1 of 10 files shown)


packages/nsid/src/index.ts
```diff
@@ -0,0 +1,70 @@
...
+*/
+
+const SEGMENT_RE = /^[a-zA-Z]([a-zA-Z0-9-])*$/
+
+export class NSID {
+  segments: string[] = []
+
+  constructor(nsid: string) {
+    const segments = nsid.split('.')
+    if (segments.length <= 1) {
```
dholms, line 22: shouldn't this need to be `< 3`?
Since an authority needs to have 2 parts?

packages/nsid/src/index.ts
```diff
@@ -0,0 +1,70 @@
...
+  get name() {
+    return this.segments.at(this.segments.length - 1)
+  }
+
+  toString() {
+    return this.segments.join('.')
+  }
+}
+
+export function parse(nsid: string): NSID {
```
dholms, line 54: nit, but elsewhere in the codebase, fns like these are static methods on the class

dholms (approved):
looks good 👍 

just two small things

packages/nsid/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "@adxp/nsid",
```
dholms, line 2: not saying this is wrong, but it does bring up the question for me of:
how granular do we want our packages to be?

I don't have a super strong opinion. But for instance, TIDs & AdxUris are both in `@adxp/common` so that may be a good home for NSIDs as well?

pfrazee: Is it module madness to put them all in their own packages? I'm wondering if they should be -- at least nsid and adxuri

packages/nsid/src/index.ts
pfrazee: Done

packages/nsid/src/index.ts
pfrazee: Yeah that would make sense, I don't imagine anybody should be using TLDs as their authority

PR #180, opened 2022-09-19
New XRPC-based Server and API implementations

pfrazee:
- Rewrites server and api packages to use XRPC via the codegen cli
- Adds `/schemas` to the top level of the repo
- Updates adxuri, auth, and repo to use nsid collections

Leaves one failing test in the server, something to do with the getPostThread view missing a reply


pfrazee:
Ready for review, depends on https://github.com/bluesky-social/adx/pull/179

packages/api/src/index.ts
```diff
@@ -1,404 +1,817 @@
...
-export enum PdsEndpoint {
-  GetDid,
-  Account,
-  Session,
-  Repo,
-  RepoCollection,
-  RepoRecord,
-  View,
-}
+/**
```
dholms, line 1: whoa this entire package is generated? neato

packages/api/src/index.ts
```diff
@@ -1,404 +1,817 @@
...
+* GENERATED CODE - DO NOT MODIFY
+* Created Tue Sep 20 2022
+*/
+import {
+  Client as XrpcClient,
+  ServiceClient as XrpcServiceClient,
+} from '@adxp/xrpc'
+import { methodSchemas, recordSchemas } from './schemas'
```
dholms, line 9: looks like `recordSchemas` isn't used here

packages/auth/src/auth-store.ts
```diff
@@ -115,7 +115,6 @@ export class AuthStore implements Signer {
     lifetime = MONTH_IN_SEC,
   ): Promise<ucan.Ucan> {
     const proofs: CapWithProof[] = []
-    const encodedTokens = new Set()
```
dholms, line 118: was this just bad logic on my part 🤔 

the point of the logic here was to find the vaguest proof for each cape but to only encode each proof once so `encodedTokens` was meant to prevent duplication. but it looks like i either didn't finish it or it got merge-conflicted at some point.

the current code will work, but may result in some unnecessarily large UCANs. probably worth sticking a TODO in

packages/auth/src/semantics.ts
```diff
@@ -7,16 +7,14 @@ Resource name: 'adx'
...
     adx://did:example:userDid/*
-- Permission to write to particular application namespace: 
-    adx://did:example:userDid/namespace/*
-- Permission to write objects within a particular collection (namespace + dataset): 
-    adx://did:example:userDid/namespace/post/*
+- Permission to write to particular application collection: 
+    adx://did:example:userDid/com.foo.post/*
 - Permission to create a single interaction on user's behalf: 
-    adx://did:example:userDid/namespace/post/234567abcdefg
+    adx://did:example:userDid/com.foo.post/234567abcdefg
```
dholms, line 13: The change in semantics here means that we can't delegate a UCAN to for `com.foo.*`

this is probably okay for now, I can come back and fix this up if we decide to stick with UCANs for auth

packages/api/src/index.ts
pfrazee: Yeah boiiiiii

packages/server/src/db/records/badge.ts
```diff
@@ -14,9 +14,8 @@ import { UserDid } from '../user-dids'
 import schemas from '../schemas'
 import { collectionToTableName } from '../util'
 
-const schemaId = 'blueskyweb.xyz:Badge'
-const collection = 'bsky/badges'
-const tableName = collectionToTableName(collection)
+const type = 'todo.social.badge'
```
dholms, line 17: this is a nit, but I kinda want a better word here than `type`. maybe `collection`? even `schemaType` or `collectionType`?

dholms: butttt im very glad to have united the two diff concepts of `colleciton` & `schemaId` & I like the NSID syntax better than both of the previous versions

dholms (approved):
this is badass 💪 

couple of small nits.

and I think there's some issues with the UCAN semantics, but we can iron those out later if we decide to keep them

packages/auth/src/auth-store.ts
dholms: added in TODO myself because this will be an easy thing to forget about

packages/api/src/index.ts
pfrazee: Yeah that's a todo

packages/auth/src/semantics.ts
pfrazee: That occurred to me while I was making the changes. I decided not to do anything yet, but I think we can definitely add the `*` semantic in the future.

packages/server/src/db/records/badge.ts
pfrazee: I currently think of it as "record type" and "type" is just short for that, though I also think of "collection" as being equivalent so in this case I'd be fine with `collection`

PR #181, opened 2022-09-20
Did PLC

dholms:
## did:plc (DID Placeholder) Server

This PR implements the did:plc server & client.

The server accepts and verifies the following operations to create & augment a DID document
- create
- rotate signing key
- rotate recovery key
- update username
- update ATP PDS

The DID itself is the base32-encoded sha256 of the genesis operation, truncated to 24 chars:
`did:plc${base32Encode(sha256(signedCreateOperation)).slice(0, 24)}`

Each operation has a pointer to the CID of the preceding operation.

The server exists solely for ordering operations. It also allows for historical rewrites by the recovery key for a 72 hour window. This is to provide recovery of the DID during an adversarial situation in which the signing key is leaked or the signing key is delegated to a bad actor that tries to commandeer the DID.

The current state of the DID document can be deterministically recreated by replaying all operations that have occurred in order.

The server provides 4 routes:
- `POST /:did` (for submitting an create/update operation)
- `GET /:did` (for retrieving the current DID document)
- `GET /log/:did` (for retrieving the log of all operations)
- `GET /data/:did` (for retrieving the current state of the constituent DID document data, _not_ in DID document form)


packages/plc/src/server/db.ts
```diff
@@ -0,0 +1,164 @@
...
+
+  static async memory(): Promise<SafeDB> {
+    return SafeDB.sqlite(':memory:')
+  }
+
+  async close(): Promise<void> {
+    await this.db.destroy()
+  }
+
+  async transaction<T>(fn: (manager: EntityManager) => Promise<T>): Promise<T> {
```
dholms, line 122: Okay so Sqlite only has one connection at a time & neither sqlite nor typeorm implement some sort of tx queue/pool: https://github.com/typeorm/typeorm/issues/307

We actually have two separate consistency problems here:
- the normal reason for using sql txs: on recovery operations we nullify some operations & add a new operations. these need to happen atomically
- our DB needs to be _strongly_ consistent. An new operation should not be added to the DB unless we know _for sure_ that no other operations have happened in the meantime. Otherwise we fork the DID

The first requires implementing a tx queue around writes.
The second requires a mutex on individual DIDs.

To avoid the complexity of competing mutexes, I simply implemented a mutex on txs. & do both the read and the write within that tx. 

I broke out the more costly validation step to be a simple read outside of the tx and just do a quick check within the tx that the prev value matches the expected value

packages/plc/README.md
```diff
@@ -0,0 +1,473 @@
+# did:plc
```
pfrazee, line 1: Need to update this bad boy

pfrazee (approved):
Looking great. I didn't subject this to _intense_ scrutiny but everything checks out based on our conversations. Great work!

PR #183, opened 2022-09-20
Unify schemas under lexicon package

pfrazee:
- Removes 'schemas' package
- Adds 'lexicon' package
- Moves schema behaviors and types into lexicon package
- Renames xrpc-cli to lex-cli
- Unifies schema definitions under lexicon


packages/lex-cli/src/codegen/client.ts
```diff
@@ -28,9 +27,9 @@ export async function genClientApi(schemas: Schema[]): Promise<GeneratedAPI> {
   const api: GeneratedAPI = { files: [] }
   const nsidTree = schemasToNsidTree(schemas)
   for (const schema of schemas) {
-    if ('xrpc' in schema) {
+    if (schema.type === 'query' || schema.type === 'procedure') {
```
dholms, line 30: nice i like this better 👌

pfrazee: Yeah this unification really cleaned up some things

dholms (approved):
looks good 🙌 
i like this unification a lot

PR #185, opened 2022-09-21
Rename the generated-code folder in the server package

pfrazee:
Renamed the `xrpc` folder to `lexicon` which I think is a little clearer


dholms (approved):
schweet. makes sense 👌

PR #187, opened 2022-09-23
JWT auth

dholms:
Added simple JWT auth to the server. 

- Encodes the user's `did` as the `sub` (subject) & signs w `HS256`
- Added passwords to users table in DB, hashed w bcrypt
- stuck auth on all of the current authed routes
- fleshed out `createSession` for logging in & receiving a jwt & `getSession` for getting current user info with an authed request
- no expiration date on JWTs yet. not sure what a reasonable time frame would be 🤔 a week? a month?


packages/lex-cli/src/codegen/client.ts
```diff
@@ -146,6 +146,23 @@ const indexTs = (project: Project, schemas: Schema[], nsidTree: NsidNS[]) =>
         ].join('\n'),
       )
 
+    //= setHeader(key: string, value: string): void {
+    //=   this.xrpc.setHeader(key, value)
+    //= }
+    const setHeaderMethod = serviceClientCls.addMethod({
```
dholms, line 152: new method. make sure i didn't do anything silly 😛

packages/lex-cli/src/codegen/client.ts
```diff
@@ -327,10 +344,14 @@ function genRecordCls(file: SourceFile, schema: RecordSchema) {
       name: 'record',
       type: `${typeModule}.Record`,
     })
+    method.addParameter({
+      name: 'headers?',
```
dholms, line 348: added optional `headers` param on record methods

packages/lex-cli/src/codegen/client.ts
pfrazee: Nailed it

packages/lex-cli/src/codegen/client.ts
pfrazee: 👍

packages/server/src/db/index.ts
```diff
@@ -1,12 +1,13 @@
+import { DataSource } from 'typeorm'
+import * as bcrypt from 'bcryptjs'
```
pfrazee, line 2: Why use a JS implementation when node has a native one?

pfrazee (approved):
Looks good, just one question about the bcrypt library

packages/server/src/db/index.ts
dholms: oh node has native scrypt 🤔 
https://nodejs.org/api/crypto.html#crypto_crypto_scrypt_password_salt_keylen_options_callback

pfrazee: oh my b. Is bcrypt that much better? Just gonna suck resources to do it in js

dholms: no looks like scrypt is actually better. ill switch it over 👌

PR #188, opened 2022-09-23
Various fixes related to integration with client software

pfrazee:
- Factors out `@adxp/uri`. I need that code in the client and don't want some of the nodejs-specific stuff that comes with common.
- Refactors xrpc to use the native fetch function (you were right, no more axios).
  - Also makes it possible to replace the fetch function, which is needed in React Native environments
- Alters the `indexedAt` and `receivedAt` timestamp construction in the database so that it's now handled by our code, to ensure the ISO-8601 encoding is used.
- Various other small fixes


packages/server/src/db/index.ts
```diff
@@ -161,6 +161,9 @@ export class Database {
     }
     record.raw = JSON.stringify(obj)
 
+    record.indexedAt = new Date().toISOString()
+    record.receivedAt = record.indexedAt
```
dholms, line 165: what's the difference here?
is it like "createdAt" vs "updatedAt"?

pfrazee: Yeah and so obviously this logic is wrong -- discussing in slack

dholms (approved):
looks good 💪

PR #189, opened 2022-09-26
Did sdk revamp

dholms:
Revamping the DID SDK 👤 

- Ended up using more of the `did-resolver` library & their resolution system
- Resolves `did:web` & `did:plc`
- included a `resolveAtpData` that resolves a DID into a format that we our system understands better (ensuring existence & formatting of signing/recovery keys, username & service)

This does require fairly rigidly formatted DIDs & makes some assumptions about how the doc will look. We'll always receive DIDs of this format from did:plc. I think it's alright to impose this structure for now on did:webs. & we can add more validation logic & relax it a bit in the future


pfrazee (approved):
LGTM

PR #190, opened 2022-09-26
Pds Did integration

dholms:
Integrating PLC into PDS

- generating PLC dids for users on signup
- accepting email & invite code for sign up as well
- remove did:test


pfrazee:
👍 LGTM!

PR #192, opened 2022-09-27
More client integration updates

pfrazee:
- Updates dev-env to use new auth system
- Updates `todo.adx.createSession()` to include some account information in the response
- Adds a new method `todo.adx.getAccountsConfig()` to describe a service's configuration for creating new accounts
- Various bugfixes

### DIFF
--- a/packages/api/src/types/todo/social/profile.ts
+++ b/packages/api/src/types/todo/social/profile.ts
@@ -1,6 +1,5 @@
 /**
 * GENERATED CODE - DO NOT MODIFY
-* Created Tue Sep 27 2022
 */
 export interface Record {
   displayName: string;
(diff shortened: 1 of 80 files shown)


packages/api/src/types/todo/social/profile.ts
```diff
@@ -1,6 +1,6 @@
 /**
 * GENERATED CODE - DO NOT MODIFY
-* Created Thu Sep 22 2022
+* Created Tue Sep 27 2022
```
dholms, line 3: probs worth regenerating these in this PR to wipe out dates

dholms (approved):
looks good 👍

packages/api/src/types/todo/social/profile.ts
pfrazee: Done

PR #193, opened 2022-09-27
Update createAccount to accept email and inviteCode and allocate DIDs

pfrazee:
Moving us closer to the real account creation/management. Notice that this moves did-allocation to the server @dholms


dholms:
Closed in favor of https://github.com/bluesky-social/adx/pull/190

PR #194, opened 2022-09-28
Improved XRPC error protocol

pfrazee:
- Adds a Lexicon spec doc
- Adds an application-level error code protocol to XRPC
- Updates Lexicon to define the errors
- Updates lex-cli to generate error types for the schema's declared errors
- Modifies the server with a test case of the new error behavior


docs/specs/lexicon.md
```diff
@@ -0,0 +1,111 @@
...
+Lexicon is a schemas document format used to define [XRPC](./xrpc.md) methods and [ATP Repository](./adx/repo.md) record types. Every Lexicon schema is written in JSON and follows the interface specified below. The schemas are identified using [NSIDs](./nsid.md) which are then used to identify the methods or record types they describe.
+
+## Interface
+
+```typescript
+interface LexiconDoc {
+  lexicon: 1
+  id: string // an NSID
+  type: 'query' | 'procedure' | 'record'
+  revision?: number
```
dholms, line 12: we should have a convo about schema revisions at some point. both how they're handled in the code & the conventions we want to push around them.

imm worried that a `revision` field might encourage people to think about versioning the wrong way

packages/server/src/api/todo/adx/account.ts
```diff
@@ -33,14 +34,17 @@ export default function (server: Server) {
     const cfg = util.getConfig(res)
 
     if (username.startsWith('did:')) {
-      throw new InvalidRequestError(
-        'Cannot register a username that starts with `did:`',
-      )
+      return {
+        status: 400,
+        error: 'InvalidUsername',
+        message: 'Cannot register a username that starts with `did:`',
```
dholms, line 40: they types have this as
```
interface XrpcError {
  name: string
  description?: string
}
```

should this be `description` instead of `message`?

dholms: actually could we just get the error types in the server's lexicon? like how it's in the clients?

packages/server/tests/account.test.ts
```diff
@@ -33,6 +36,24 @@ describe('auth', () => {
...
+        {
+          username: 'did:bad-username.test',
+          did: 'bad.test',
+          password: 'asdf',
+        },
+      )
+      throw new Error('Didnt throw')
+    } catch (e: any) {
+      expect(
+        e instanceof TodoAdxCreateAccount.InvalidUsernameError,
```
dholms, line 52: hell yea

dholms (approved):
looks good.
just the note `description` which i _think_ should be `message` 🧐 

& maybe getting error classes into the server if you think it's worthwhile

docs/specs/lexicon.md
pfrazee: Okay sgtm

packages/server/src/api/todo/adx/account.ts
pfrazee: You're confusing the Lexicon schema document with the protocol shapes. In the lex schema document, it's `{name, description}` where the description is purely for documentation purposes. The error objects that get returned are `{error, message}`

packages/server/tests/account.test.ts
pfrazee: We certainly _could_ get the error types in the server lexicon, but what I did instead was type the response so that `error` has to be one of the declared error names. That means you'll get a type error on invalid ones, and the autocomplete will help you out. I figured that was simpler than having to import the error classes, but we can do both.

dholms: ah got it. think string types are fine for now 👍

packages/server/src/api/todo/adx/account.ts
dholms: ahhh got it 👍

PR #195, opened 2022-09-28
Plc auth lib integration

dholms:
I split this out into it's own PR cause it was sort of its own branch of work.

This works did:plc resolution into our auth lib.

tbh I'm not particularly thrilled with how this turned out, but I'm not sure of our other option. 

The main change in DX is that we can't just call verification code from the auth lib, we have to instantiate a `Verifier` that encapsulates some DID resolution & verification logic. This is because resolving dids takes parameters (`timeout` & `plcUrl`) which means those have to pass through to any function that may require DID resolution.

I briefly thought about ripping out the UCAN stuff, but we'd end up building something with a similar shape anyway, so I ended up leaving it in for now to avoid having to refactor repo code.


pfrazee:
👍 LGTM!

PR #196, opened 2022-09-28
Add better cursor pagination behavior to getFeed

pfrazee:
- Adds a `cursor` field to items returned by `todo.social.getFeed`
- Internally switches to ordering by indexedAt (needs to eventually be rtime probably)

This change basically makes it unambiguous what to pass into the `before` param when using getFeed. I'd suggest using this pattern or similar any time we need pagination in APIs.


dholms (approved):
looks good 👍

PR #197, opened 2022-09-28
Add PLC server to dev-env

dholms:
Adds PLC server to dev-env.

Only allows one at a time & makes it a requirement to start up any PDSs


pfrazee:
LGTM!

PR #200, opened 2022-09-29
Notifications

dholms:
Notifications 🔔 

Adds a notifications table that maintains a list of all notifications relevant to a user.

Adds views for
- listing notifications
- getting unread notification count
- telling the server that the user has seen notifications after a certain time.

Handles notification `reason`s similarly to entity `type`s: it's an unconstrained field, but there is a comment in the schema suggesting what expected values are.

For each notification, returns:
- uri of the object that caused the notif
- the author of the object
- the record that caused the notif
- the reason (like, repost, mention, etc)
- the reasonSubject if applicable (for instance the uri of the post that was liked)
- if the notif has been read
- indexed time of the object that caused the notif


packages/server/src/api/todo/social/getFeed.ts
```diff
@@ -36,6 +36,7 @@ export default function (server: Server) {
           )
           .leftJoin(User, 'author', 'author.did = post.creator')
           .where('user.did = :did', { did: requester })
+          .andWhere('post.creator != :requester', { requester })
```
dholms, line 39: we weren't returning a user's posts in their feed, but we were returning them if they were reposted, so this strips those out as well. can change this if we like

packages/server/src/api/todo/social/postNotificationsSeen.ts
```diff
@@ -0,0 +1,32 @@
+import { Server } from '../../../lexicon'
+import { AuthRequiredError, InvalidRequestError } from '@adxp/xrpc-server'
+import * as util from '../../../util'
+import { User } from '../../../db/user'
+
+export default function (server: Server) {
+  server.todo.social.postNotificationsSeen(async (_params, input, req, res) => {
+    const { seenAt } = input.body
```
dholms, line 8: this allows any time to be set as `seenAt` i thought maybe doing the `Max` of `seenAt` & the currently registered time in the DB, but then thought there might be a usecase for "rewinding" notifications 🤔

packages/server/src/db/index.ts
```diff
@@ -181,12 +190,19 @@ export class Database {
...
   }
 
   async deleteRecord(uri: AdxUri) {
     const table = this.findTableForCollection(uri.collection)
     const recordTable = this.db.getRepository(AdxRecord)
-    await Promise.all([table.delete(uri), recordTable.delete(uri.toString())])
+    await Promise.all([
+      table.delete(uri),
+      recordTable.delete(uri.toString()),
+      this.notifications.deleteForRecord(uri),
```
dholms, line 204: removes notifications for deleted objects

packages/server/src/api/todo/social/getFeed.ts
pfrazee: Actually I think we need user's posts in their feed, but not the retweets? It can be confusing not to see your own posts

pfrazee (approved):
Looking good. Can merge as is, but I wonder if the `reason` token ought to be an NSID-based constant? Like `app.bsky.likeEvent`. Would need a new Lexicon doc type probably.

packages/server/src/api/todo/social/getFeed.ts
dholms: Made an issue. i'll get to it soonish: https://github.com/bluesky-social/adx/issues/201

PR #202, opened 2022-09-30
Feed bugfixes

pfrazee:
Mainly fixes an issue where the `repostedBy` information would be incorrectly populated on non-reposts


dholms (approved):
nice looks good 👍

PR #204, opened 2022-10-03
Invite codes

dholms:


packages/server/src/db/invite-codes.ts
```diff
@@ -0,0 +1,31 @@
...
+  @Column('varchar')
+  createdBy: string
+}
+
+@Entity({ name: 'invite_codes_uses' })
+export class InviteCodeUses {
+  @PrimaryColumn('varchar')
+  code: string
+
+  @PrimaryColumn('varchar')
```
pfrazee, line 26: Hey was just doing some early review and noticed you've got multiple PrimaryColumns here, is that correct?

dholms: here it is yup.
We have the `InviteCode` table where the code is unique. And then the `InviteCodeUse` table where each code can have more than one use, but only one for each user. So the primary key is code + user_that_used_that_code

pfrazee: Ah I was unclear on whether you _could_ have multiple primary columns on a table. It looks like TypeORM turns that into a composite primary key.

dholms: ah yupp it does^^

packages/server/src/api/todo/adx/account.ts
```diff
@@ -67,6 +97,14 @@ export default function (server: Server) {
 
     await db.registerUser(email, username, did, password)
 
+    if (config.inviteRequired && inviteCode) {
+      const inviteCodeUse = new InviteCodeUse()
+      inviteCodeUse.code = inviteCode
+      inviteCodeUse.usedBy = did
+      inviteCodeUse.usedAt = new Date().toISOString()
+      await db.db.getRepository(InviteCodeUse).insert(inviteCodeUse)
```
devinivy, line 105: I think a double-use could technically occur with the invite code between the `useCount` check and this insert.  I don't think it needs to hold-up the PR, though— probably could take a pass at this and similar issues separately from this work.

packages/server/src/auth.ts
```diff
@@ -45,6 +50,18 @@ export class ServerAuth {
     return authorized === did
   }
 
+  verifyAdmin(req: express.Request): boolean {
+    const header = req.headers.authorization || ''
+    if (!header.startsWith(BASIC)) return false
+    const token = header.slice(BASIC.length)
+    const [username, password] = uint8arrays
+      .toString(uint8arrays.fromString(token, 'base64url'), 'utf8')
```
devinivy, line 58: I believe basic auth typically calls for "regular" base64 encoding rather than url-safe base64.  Just a nit, certainly not too hung-up on it :)

packages/server/tests/account.test.ts
```diff
@@ -4,36 +4,99 @@ import AdxApi, {
...
+          inviteCode: 'fake-invite',
+        },
+      )
+      throw new Error('Didnt throw')
+    } catch (e) {
+      console.log(e)
+      expect(
+        e instanceof TodoAdxCreateAccount.InvalidInviteCodeError,
+      ).toBeTruthy()
+    }
```
devinivy, line 90: Just a headsup, I think there's a (arguably) more convenient way to write this—something along the lines of:
```js
await expect(createAccount()).rejects.toThrow(TodoAdxCreateAccount.InvalidInviteCodeError);
```

devinivy (approved):
Had a few thoughts, but this looks real nice to me!

packages/server/src/db/invite-codes.ts
```diff
@@ -0,0 +1,31 @@
...
+  usedBy: string
+
+  @PrimaryColumn('varchar')
```
devinivy, line 29: Is there a reason `usedAt` oughtta be part of the primary key?  It seems like it might be fair to limit each entry's identifier to just `(code, usedBy)`.

packages/server/src/auth.ts
dholms: ah you're right. & i think padded too. used to using base64url elsewhere in our code haha

packages/server/src/db/invite-codes.ts
dholms: oh nope you're totally right about that. thanks for the catch

packages/server/tests/account.test.ts
dholms: yup good looks 👌 we do that elsewhere in the code but for some reason didn't make it in here. (& you can tell i copy/pasted tests 😛)

packages/server/src/api/todo/adx/account.ts
dholms: yup i thought of that as well. I'm banking on requests being sparse enough that that isn't an issue 😅 

there's serveral spots where writes should be transactional. and a couple like this where we need to make sure writes are consistent with the current DB value (updating the repo root for instance to prevent branched histories)

pfrazee: I'm okay with leaving the bug if we leave a TODO, maybe create a tracking issue the areas we know transactionality is an issue

pfrazee (approved):
👍 LGTM

packages/server/src/api/todo/adx/account.ts
dholms: good idea

PR #205, opened 2022-10-04
Support firehose feed algorithm

devinivy:
A few things going on here!  The main goal is to support a "firehose" algorithm as a choice for the user's home feed.  This feed would include all posts and reposts in it.

First, `getFeed` was split into `getAuthorFeed` and `getHomeFeed`.  The author feed includes all posts and reposts by some specific author, always in reverse-chronological order.  The home feed is based on the user's followees, and includes different posts depending on the chosen `algorithm` parameter: we support `'firehose'` and `'reverse-chronological'` (the default) as options.

I ended-up unifying the queries and presentation a bit between `getAuthorFeed` and `getHomeFeed`, which I hope will make it easier to see where they are similar versus where they differ (generally just in the result set of FeedItems).  In particular, the home feed now uses a subquery to find (re-)posts by followees, rather than starting from the users table and joining over to posts, followers, etc.

The majority of the lines of code changed are from codegen, so hopefully it wont be quite as big a lift to review as it appears at first glance!


packages/api/tsconfig.json
```diff
@@ -7,6 +7,6 @@
   "include": ["./src"],
   "references": [
     { "path": "../xrpc/tsconfig.build.json" },
-    { "path": "../xrpc-cli/tsconfig.build.json" },
+    { "path": "../xrpc-server/tsconfig.build.json" }
```
devinivy, line 10: I was seeing an error here in my IDE coming from typescript.  I _think_ this is the right change (xrpc-cli no longer exists), but if not feel free to tweak it or I can revert this change.

devinivy: Ahh got it, that's clear 👍 
```suggestion
    { "path": "../lex-cli/tsconfig.build.json" }
```

pfrazee:
👍 LGTM!

dholms (approved):
looks excellent 🙌

PR #208, opened 2022-10-04
Fixes for handling of posts vs. reposts in feeds

### LINKED ISSUE #201: Add users own posts to their feed
Users expect to see their own posts in their feed
But not reposts of their own posts

devinivy:
There are a few fixes in here to home and author feeds:
 - A post and its repost are now distinguished from each other, and can appear together in the same feed.  This was fixed by adjusting the `groupBy()` clause.  This also repaired a flaky test, where `repostedBy` was sometimes not appearing when it should have.
 - Reposts are are ordered in the feed based on their indexed time, rather than that of their post.  The FeedItem `cursor` values are updated to reflect this.
 - Users' own posts (but not reposts) now appear in their feed. Resolves #201

### DIFF
--- a/packages/server/src/api/todo/social/getAuthorFeed.ts
+++ b/packages/server/src/api/todo/social/getAuthorFeed.ts
@@ -8,6 +8,7 @@ import {
   queryPostsWithReposts,
   queryResultToFeedItem,
 } from './util'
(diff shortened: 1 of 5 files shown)


pfrazee:
Nice, skim lgtm, @dholms ?

dholms (approved):
looks great 🎉

PR #209, opened 2022-10-05
Logging

dholms:
Added logging to server & various packages ✍️ 

Using [pino](https://github.com/pinojs/pino) for structured logging and `[pino-http](https://github.com/pinojs/pino-http)` to implement some logging middleware for our server.

We have a root logger in `@adxp/common` that then allows you to create child loggers off of it, for each subsystem (repo, xrpc, etc).

There are a few env variables that control logging output:
- `LOG_ENABLED`:  enables logging (default: false)
- `LOG_SYSTEMS`: takes a space-or-comma-separated list & enables logging various subsystems. if unset, all are enabled. for instance the value `repo did-sdk` would enable logging around repos & the did-sdk but not xrpc, http, auth, etc.
- `LOG_LEVEL`: the minimum level of log to be outputted. (default: 'info')
- `LOG_DESTINATION`:  takes file path or a numerical file descriptor: `test.log` or `1` (for stdout). Default to stdout.

Pino is compatible with all sorts of logging service, i've held off on that for the time being, but it'll be quick to add in.

---

I tried to give us enough logging for a good picture into the going ons of the server without overdoing it.
- http reqs
- repo mutations
- did resolutions
- xrpc errors
- internal server errors (caught by xrpc or express)

---

Tests by default log to `test.log` in the relevant package folder.
This logs can be viewed by running `yarn test:log` which pipes the output through `pino-pretty` to give a nicer view.

### Other things that slipped in here:
I ended up buffing up our `locals` (that be pass around on `res` like db/blockstore/etc) a bit. Moving them to their own file instead of util & adding a type to keep us honest.


pfrazee (approved):
Nice. Very lightly skimmed so I know what's up, lgtm

packages/common/src/logger.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export const subsystemLogger = (name: string): pino.Logger => {
+  if (subsystems[name]) return subsystems[name]
+  const subsystemEnabled =
+    enabled && (allSystemsEnabled || enabledSystems.indexOf(name) > -1)
+
+  // can't disable child loggers, so we just set their level to fatal to effectively turn them off
+  subsystems[name] = rootLogger.child(
+    { name },
+    { level: subsystemEnabled ? level : 'fatal' },
```
devinivy, line 33: I think you may be able to use `'silent'` for this.

packages/server/src/api/todo/adx/account.ts
```diff
@@ -31,7 +31,7 @@ export default function (server: Server) {
 
   server.todo.adx.createAccount(async (_params, input, _req, res) => {
     const { email, username, password, inviteCode } = input.body
-    const { db, blockstore, auth, config, keypair } = util.getLocals(res)
+    const { db, blockstore, auth, config, keypair, logger } = locals.get(res)
```
devinivy, line 34: Codifying `locals` like this feels like a nice improvement 👍

packages/server/src/api/todo/adx/repo.ts
```diff
@@ -120,6 +120,7 @@ export default function (server: Server) {
     try {
       await repoDiff.processDiff(db, repo, diff)
     } catch (err) {
+      logger.info({ did, err }, 'failed to index batch write')
```
devinivy, line 123: Not sure if we want to get to this level of granularity just yet, but this is the kind of thing that may fit nicely under `logger.warn()` or `logger.error()`.  On the flipside of that, there may be some logs that could be downgraded from info the `logger.debug()`.  Just a general thing, probably no need to worry about it in this PR which is already a huge life improvement.

packages/server/src/auth.ts
```diff
@@ -67,4 +64,21 @@ export class ServerAuth {
...
+  if (!token.startsWith(BASIC)) return null
+  const b64 = token.slice(BASIC.length)
+  let parsed: string[]
+  try {
+    parsed = uint8arrays
+      .toString(uint8arrays.fromString(b64, 'base64pad'), 'utf8')
+      .split(':')
+  } catch (err) {
+    return null
+  }
```
devinivy, line 79: What's the error that could happen here?

Also, a small/unrelated thing, but I wonder if it may be interesting to consider incorporating something like [bounce](https://hapi.dev/module/bounce/) to distinguish unrecoverable node/"system" errors (e.g. TypeError, SyntaxError) from operational ones.

devinivy (approved):
I'm a fan 👍 👍

packages/server/src/auth.ts
dholms: Invalid base64 throws here. `abc%%|%|%|%|%` would throw for instance. and since i'm using in the log serializer, it's probably best to catch it & not let it bubble up

dholms: ah nice I like the look of something like `bounce` for sure

packages/server/src/api/todo/adx/repo.ts
dholms: good call. i changed these to `warn` & I dropped some of the duplicate ones in `db` fns down to `debug` (I think those were the main culprit)

packages/common/src/logger.ts
dholms: oh nice great find 👍

PR #210, opened 2022-10-05
Cleanup

dholms:
A couple quick cleanup things:

- removing unused network code in `@adxp/common` (the only relevant part of that, adx-uris is in it's own package now)
- renaming did-sdk to the more descriptive did-resolver


devinivy (approved):
Looking good 👍

PR #216, opened 2022-10-06
Forgot password methods

devinivy:
Added two methods here to service password resets:
#### **Input**

- `email` - the email of the user who is requesting a password reset.

#### **Output**

None, i.e. an empty object `{}`.

If an account for `email` exists, an email is sent to `email` containing a short-lived JWT (15mins) which includes the user's `did`, and indicates that it permits `did` to reset their password. The JWT will be signed using a secret concatenated with the hash of the user's current password.  This way, token validation only succeeds if their password has not changed, making it effectively single-use.

#### **Errors**

None. If the email doesn't exist then an email isn't sent. It’s common practice to feign success even if there’s no existing account, so that it can’t be used as a way to check if a given user has an account with the service.

#### **Input**

- `token` - this is the token described in `requestAccountPasswordReset`'s output: a JWT indicating that the user `did` can change their current password.
- `password` - the user's updated password.

#### **Output**

None, i.e. an empty object `{}`.

If the `token` can be verified using a secret concatenated with the hash of `did`'s current password, then the user's password is reset.

#### **Errors**

- `ExpiredToken` - the user took too long to use their password reset token.
- `InvalidToken` - the token was not able to be verified, most likely because the user has changed their password since the token was originally requested.


packages/lex-cli/src/codegen/server.ts
```diff
@@ -164,8 +164,10 @@ function genNamespaceCls(file: SourceFile, ns: NsidNS) {
     })
     method.setBodyText(
       [
-        `/** @ts-ignore */`,
-        `return this.server.xrpc.method('${schema.id}', handler)`,
+        // Placing schema on separate line, since the following one was being formatted
+        // into multiple lines and causing the ts-ignore to ignore the wrong line.
+        `const schema = '${schema.id}' // @ts-ignore`,
```
devinivy, line 169: This was an incidental change that I found I had to make to avoid type errors.  I tried for the alternate path of ensuring the generated file would not format the following line onto multiple lines when it got too long.  But I wasn't able to easily identify where that formatting was happening— I think it may be something implicit in ts-morph.

packages/server/src/db/user.ts
```diff
@@ -9,7 +9,7 @@ export class User {
   @Index()
   username: string
 
-  @Column('varchar')
+  @Column({ type: 'varchar', collation: 'nocase' })
```
devinivy, line 12: Made the password field case-insensitive.  This will help with uniqueness if we ever enforce that.  But more importantly here, it helps with user lookups by email, acknowledging that the user may not type in their password with the same casing every time.

packages/server/src/api/todo/adx/account/index.ts
```diff
@@ -1,11 +1,14 @@
...
-import * as locals from '../../../locals'
+import * as locals from '../../../../locals'
 import { Repo } from '@adxp/repo'
 import { PlcClient } from '@adxp/plc'
-import { InviteCode, InviteCodeUse } from '../../../db/invite-codes'
+import { InviteCode, InviteCodeUse } from '../../../../db/invite-codes'
+import passwordReset from './password-reset'
 
 export default function (server: Server) {
+  passwordReset(server)
```
dholms, line 10: nit, but im inclined to either
- not nest these directories & just have `password-reset` at the same level as `account` (similar to `invites`)
- or more aggressively nest directories & have a separate file for every xrpc method

leaning towards the former

packages/server/src/api/todo/adx/account/password-reset.ts
```diff
@@ -0,0 +1,105 @@
...
+
+const PASSWORD_RESET_SCOPE = 'todo.adx.resetAccountPassword'
+
+export default function (server: Server) {
+  server.todo.adx.requestAccountPasswordReset(
+    async (_params, input, _req, res) => {
+      const { db, mailer, config } = locals.get(res)
+      const { email } = input.body
+
+      // @TODO should multiple users be able to have the same email?
```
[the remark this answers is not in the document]
dholms: that's a good point... hmmm probably not? most services don't allow email re-use & several things (this being an example) get easier when you don't

packages/server/src/api/todo/adx/account/password-reset.ts
```diff
@@ -0,0 +1,105 @@
...
+        body: {},
+      }
+    },
+  )
+
+  server.todo.adx.resetAccountPassword(async (_params, input, _req, res) => {
+    const { db, config } = locals.get(res)
+    const { token, password } = input.body
+
+    const tokenBody = jwt.decode(token)
```
dholms, line 45: `verify` also does `decode` so this code could probably be simplified just a bit

`const payload = jwt.verify(...)`

dholms: ohhh wait you need the `did` from the decoded payload to get the signing key for verification

ignore prev message^^ 😛

packages/server/src/api/todo/adx/account/password-reset.ts
```diff
@@ -0,0 +1,105 @@
...
+      encoding: 'application/json',
+      body: {},
+    }
+  })
+}
+
+const getSigningKey = (user: User, config: ServerConfig) =>
+  `${config.jwtSecret}::${user.password}`
+
+type ErrorResponse = {
```
dholms, line 85: no need to do for this PR, but this type is relevant to other xrpc methods as well. & maybe we could go so far as to codegen error creators 👀

packages/server/src/db/user.ts
dholms: I'm guessing you meant "email" here not "password". But nice I like it 👌

packages/server/src/index.ts
```diff
@@ -36,29 +39,45 @@ const runServer = (
...
+      ? createTransport(config.emailSmtpUrl)
+      : createTransport({ jsonTransport: true })
+
+  const mailer = new ServerMailer(mailTransport, config)
+
   const app = express()
   app.use(cors())
   app.use(loggerMiddleware)
 
+  const locals: Locals = {
```
dholms, line 53: good call on pulling this out from the middleware 👍 

since we overwrite the logger in the middleware (since that one _is_ request specific), can we not set it here?

`Omit<Locals, 'logger'>`

packages/server/src/api/todo/adx/account/password-reset.ts
devinivy: Yeah, this was an interesting one.  The main reason I had to write this type was to make it easier to write the type `ErrorResponse & { error: 'SomeSpecificError' }` for those error creators.  Otherwise typescript wouldn't infer that the error creators were always returning an allowed `error` value for the handler.  On the upside, it means the handlers are typed specifically enough to know exactly the permitted errors per the schema :)

devinivy: (P.S. that would be pretty awesome!)

packages/server/src/index.ts
devinivy: Can do!  Is the idea to omit `logger` from `app.locals` entirely, or just to change how we piece these types together for `app.locals` versus `res.locals`?

dholms: ohh i see the point of that now

okay never mind, leave this as is, but maybe add a quick comment next to the middleware part (`logger: req.log`) noting that the logger is request specific.

dholms (approved):
looking good 🎉 

couple of small nits. the only outstanding thing is making sure this works with the build script

packages/server/src/mailer/index.ts
```diff
@@ -0,0 +1,56 @@
...
+    return await this.transporter.sendMail({
+      ...mailOpts,
+      from: mailOpts.from ?? this.config.emailNoReplyAddress,
+      html,
+    })
+  }
+
+  private compile(name) {
+    return this.handlebars.compile(
+      fs.readFileSync(`${__dirname}/templates/${name}.hbs`).toString(),
```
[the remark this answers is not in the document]
devinivy: Thanks for catching that!

packages/server/src/db/user.ts
devinivy: Yep!  I've edited the comment to avoid further confusion 🙃

packages/server/src/api/todo/adx/account/password-reset.ts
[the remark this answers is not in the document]
devinivy: Addressed here: https://github.com/bluesky-social/adx/pull/217

dholms:
looks great 👍

PR #217, opened 2022-10-06
Enforce single user per email, usernames case-insensitive

devinivy:
Addresses https://github.com/bluesky-social/adx/pull/216#discussion_r989244589.  This enforces unique emails per user, and makes usernames case-insensitive.


packages/server/src/api/todo/adx/account.ts
```diff
@@ -81,12 +81,22 @@ export default function (server: Server) {
       isTestUser = true
     }
 
-    // verify username is available
-    const found = await db.getUser(username)
-    if (found !== null) {
+    // verify username and email are available.
+
+    // @TODO consider pushing this to the db, and checking for a
```
dholms, line 86: yeah that's exactly why i held off on it :/

dholms (approved):
good looks 👍 

thanks for doin this

PR #218, opened 2022-10-06
Fix dev-env build, account for server's hbs templates

devinivy:
Fixing a regressing to the dev-env's build introduced in #216.


dholms (approved):
nice 👍

PR #220, opened 2022-10-06
Use Kysely for PLC DB

dholms:
This started out as just checking out Kysely to get a feel for it in a smaller context. 
It's much better than TypeORM. 
- Better typing in queries
- Transactions that _actually_ work & don't require a mutex
- Dismisses the whole ORM abstraction, which I think is more honest & probably better tooling for server admins

We're planning to switch over TypeORM to Kysely on the server, so this is now an _actual_ PR & not just a testing ground 😛


packages/plc/src/server/db.ts
```diff
@@ -1,44 +1,59 @@
...
 import * as t from '../lib/types'
 import { ServerError } from './error'
 
-export class Database {
-  db: SafeDB
+interface OperationsTable {
+  did: string
+  operation: string
+  cid: string
+  nullified: 0 | 1
```
dholms, line 13: better-sqlite3 forces you to be honest about sqlite values & doesn't let you use booleans 😛

packages/plc/src/server/db.ts
```diff
@@ -1,44 +1,61 @@
...
```
devinivy, line 13: If we want to wrangle this quirk using types, we may be able to use an enum.  E.g. something along the lines of:
```ts
enum DbBool {
  true = 1,
  false = 0,
}
```

devinivy (approved):
Big yup! 👍

packages/plc/src/server/db.ts
dholms: im not too worried about it just yet. if we're feeling friction from it in the future, then totally 👌

Issue #211, opened 2022-10-05
Testing


dholms:
Some spots that I noticed need testing. Feel free to add to it if you see anything else

### pds
- [x] (@devinivy) generally, just going through views.test.ts & ensuring we’re testing the whole view for each. I shied away from fixtures on this, but it might not be the worst idea 🤔  
- [x] (@devinivy) views.test.ts could probably be separated out into several different test suites. & in each one we test the view several different times. for instance, getting a timeline from each user’s perspective or likedBy on several different posts
- [x] (@devinivy) failure test for a bad password
  https://github.com/bluesky-social/atproto/pull/418
- [x] (@dholms) handles races (several updates to a repo that happen at the same time. these should be linearized & not stomp each other) 
  https://github.com/bluesky-social/atproto/pull/245

### plc
- [x] (@dholms) handles races (really doesn’t matter who wins if they’re happening at basically the same time, but it does need to definitively resolve one way or the other & not fork history)
- [x] (@dholms) does not allow didKey types that we don’t support
- [x] (@dholms) handles service endpoins/usernames with & without `https://`. [related](https://github.com/bluesky-social/adx/issues/221)
    https://github.com/bluesky-social/adx/pull/226

### crypto
- [x] (@dholms) serializes & deserializes keys bytes <-> did:keys
    https://github.com/bluesky-social/adx/pull/225

PR #222, opened 2022-10-07
Seeds for testing server views

devinivy:
Taking a cue from #211, this starts breaking-up the view tests into their own suites.  The main work done here is to a. allow them to share a common data seed, and 2. to make it convenient to reference data loaded from that seed when writing tests.  Otherwise the tests themselves were unaltered, though there will be a followup PR (now #223) to demo how this might look with snapshot testing added into the mix.


dholms (approved):
nice this looks great. already so much more understandable 🙌

PR #225, opened 2022-10-07
Tests for did conversion

dholms:
Added tests to confirm we transform between NIST P-256 public keys & did:keys correctly. Based on test vectors from the w3c


devinivy (approved):
👍

PR #226, opened 2022-10-07
Plc tests & cleanup

dholms:
Takes care of the tests from https://github.com/bluesky-social/adx/issues/211
- ensuring a valid did:key on key rotate ops (had to implement)
- a test around race conditions (& a bug that i sussed out on how we were returning ops)
- handling username/pds with or without a protocol handler (had to implement)

Adds logging with pino

Fixes some build issues around switching to keysely
- nohoist better-sqlite3 & mark it external in build script (I had to wipe out my node_modules & reinstall for this to work, you may have to do the same)
- fix up a couple of small things in tests for server/did-resolver to bring them up to date (mainly creating tables)

resolves https://github.com/bluesky-social/adx/issues/221

### DIFF
--- a/packages/plc/tests/server.test.ts
+++ b/packages/plc/tests/server.test.ts
@@ -4,6 +4,7 @@ import * as document from '../src/lib/document'
 import getPort from 'get-port'
 import * as util from './util'
 import { cidForData } from '@adxp/common'
+import { AxiosError } from 'axios'
 
 const USE_TEST_SERVER = true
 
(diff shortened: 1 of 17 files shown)


packages/plc/tests/server.test.ts
```diff
@@ -82,6 +83,15 @@ describe('PLC server', () => {
     expect(doc.atpPds).toEqual(atpPds)
   })
 
+  it('does not allow key types that we do not support', async () => {
```
devinivy, line 86: 👍 👍

devinivy (approved):
Looking good, some nice tightening-up in here.

PR #228, opened 2022-10-10
Fix PLC server usage in dev-env

devinivy:
The dev-env wasn't working as of the recent introduction of better-sqlite3.  Included here are the fixes!
 - Create the PLC tables when bootstrapping the PLC server in dev-env.
 - Ensure better-sqlite3 bindings are external to the dev-env build.


pfrazee:
Appreciate this!

PR #231, opened 2022-10-11
Handling updates

dholms:
Editing records (especially posts & badges) is tricky and you need good UX to prevent breaking user expectations. This provides a stop gap until we iron out that UX:
- disables edits for the time being 
- future-proofs our schemas & DB tables by including a `cid` (the "version") along with the `uri` when referencing a post
- takes a step in the direction of https://github.com/bluesky-social/adx/issues/224 by not requiring the `recordKey` to be a TID

Schema updates:
- changed `type` param to `collection` (& made the sugary codegen methods support this as well)
- changed `tid` param to `recordKey`
- create schema returns a cid as well
- anything with a subject (that references a URI not a DID) also includes a subjectCid field
- posts that reference a parent also include a parentCid field
- if a post has a parent, it _must_ include the root & parentCid. the side effect of this (which we'd already assumed in the query logic but not reflected in the schema) is that if you're commenting directly on the root post, you see the root post uri as both `root` & `parent`


packages/server/src/api/todo/adx/repo.ts
```diff
@@ -95,6 +94,10 @@ export default function (server: Server) {
       throw new AuthRequiredError()
     }
     const tx = input.body
+    const update = tx.writes.findIndex((write) => write.action === 'update')
+    if (update > -1) {
```
devinivy, line 98: This looks perfect, just wanted to advertise a similar option, which can be a little more intuitive without the fun `-1` business 😆 
```suggestion
    const hasUpdate = tx.writes.some((write) => write.action === 'update')
    if (hasUpdate) {
```

packages/server/src/db/records/like.ts
```diff
@@ -19,8 +22,10 @@ export const createTable = async (db: Kysely<PartialDB>): Promise<void> => {
   await db.schema
     .createTable(tableName)
     .addColumn('uri', 'varchar', (col) => col.primaryKey())
+    .addColumn('cid', 'varchar', (col) => col.notNull())
     .addColumn('creator', 'varchar', (col) => col.notNull())
     .addColumn('subject', 'varchar', (col) => col.notNull())
+    .addColumn('subjectCid', 'varchar', (col) => col.notNull())
```
devinivy, line 28: Not intended for this PR, but it might be time we sprinkled some column indexes in here.  E.g. we want to make it quick to query liked-by, which would be supported by an index `(subject, subjectCid)`.

packages/server/test.env
```diff
@@ -1,2 +1,2 @@
-LOG_ENABLED=true
+LOG_ENABLED=false
```
devinivy, line 1: Just want to note that I have been enjoying the ability to go through test log output, thanks for sneaking that in during #209!

packages/server/tests/_util.ts
```diff
@@ -99,6 +100,7 @@ export const forSnapshot = (obj: unknown) => {
   const records = { [kTake]: 'record' }
   const collections = { [kTake]: 'collection' }
   const users = { [kTake]: 'user' }
+  const cids = { [kTake]: 'cids' }
```
devinivy, line 103: 🤤 nice call!

devinivy (approved):
I know some small changes may be on their way, but everything here is looking solid to me.  This was a lot of work! Good stuff 🙏

packages/server/src/api/todo/adx/repo.ts
dholms: oh much better 👍

packages/server/src/db/records/like.ts
dholms: agreed 👍

packages/server/test.env
dholms: 🕺 
for sure! glad it's been helpful. I haven't used it much but when ive needed it, it's been great

packages/server/tests/seeds/client.ts
```diff
@@ -71,8 +94,7 @@ export class SeedClient {
     this.profiles[by] = {
       displayName,
       description,
-      uriRaw: res.uri,
-      uri: new AdxUri(res.uri),
+      ref: new Reference(res.uri, res.cid),
```
devinivy, line 97: 👍 👍

devinivy (approved):

PR #233, opened 2022-10-12
Run PDS using postgres

devinivy:
A handful of things going on here!

First, some Dockerized helpers for working with postgres have been added to `packages/pg/`.  The main idea is that there's a script `with-test-db.sh` that allows running arbitrary commands with a single-use postgres database available.  Backing this up is Docker and a `docker-compose.yaml` file which additionally contains a service for a persistent postgres database (i.e. for use in development).  There is more info in `packages/pg/README.md`!

Second, the PDS has been updated to support postgres in addition to sqlite.  In order to run the PDS tests with postgres, you can run `yarn workspace @adxp/server test:pg`.  Here's a summary of changes that needed to be made to support this:
 - The home and author feed queries needed to be rewritten: sqlite allows the not-well-defined group-by clause and postgres doesn't allow this.  I rewrote it to use a union between posts and reposts, which actually felt a little cleaner anyway.  It also fixed a bug!
 - The postgres driver by default selects big integers as js strings.  I added a little snippet to select them as js integers in order to align the behavior with sqlite.  The drawback is that if the integer gets too large, js wont be able to represent it, but I don't think this is an imminent concern for us.
 - Postgres treats column references as case-insensitive unless they are quoted.  Kysely will do this quoting for us, as long as we stick to `db.dynamic.ref('table.column')` rather than e.g. writing them raw `` sql`table.column` ``.  You can always put refs inside raw sql statements, so this doesn't turn out to be a limitation :)
 - Primary key constraint names needed to be unique for postgres, so I prefixed them with their table name.  This also matches the names that are being generated by kysely (or perhaps by pg itself).
 - Tests need to specify a unique [pg schema](https://www.postgresql.org/docs/current/ddl-schemas.html) to run in.  This is how we maintain test suite parallelism: all tests use the same postgres database, but segment their tables into separate schemas.

The dev-env hasn't been touched yet, so it's still running on sqlite.  But as some followup work, we could certainly wire it up to pg.  I think that's it!  Will add some comments below too, to call a few minor things out.


packages/server/src/db/tables/invite.ts
```diff
@@ -37,8 +37,9 @@ export const createTable = async (db: Kysely<PartialDB>): Promise<void> => {
   await db.schema
     .createTable(supportingTableName)
     .addColumn('code', 'varchar', (col) => col.notNull())
-    .addColumn('usedBy', 'integer', (col) => col.notNull())
-    .addColumn('usedAt', 'integer', (col) => col.notNull())
-    .addPrimaryKeyConstraint('primary_key', ['code', 'usedBy'])
+    .addColumn('usedBy', 'varchar', (col) => col.notNull())
+    .addColumn('usedAt', 'varchar', (col) => col.notNull())
```
devinivy, line 41: I think these were intended to be `varchar` to begin with.  Kind of concerning that sqlite didn't complain about it!

dholms: 😱

packages/server/src/api/todo/adx/account.ts
```diff
@@ -49,7 +50,7 @@ export default function (server: Server) {
         .select([
           'invite.disabled as disabled',
           'invite.availableUses as availableUses',
-          sql<number>`count(code_use.usedBy)`.as('useCount'),
+          sql<number>`count(${ref('code_use.usedBy')})`.as('useCount'),
```
dholms, line 53: is `ref` only required here because the column name is case-sensitive?

we can always change column names to snake case, wouldn't be a big deal 🧐

devinivy: Yep, I believe that is the case.  I have gotten sort of used to this kind of thing from working with objection, so I don't have a strong opinion on it either way!

PR #238, opened 2022-10-13
User search

devinivy:
- Adding two new methods to support user search:
   - `app.bsky.getUsersTypeahead` - meant to be quick and return minimal user info to show search suggestions to users as they type.  Not paginated, but does accept a `limit`.
   - `app.bsky.getUsersSearch` - similar to `getUsersTypeahead`, but returns more complete user/profile information, and is fully paginated.
 - Both fuzzy search on `username` and `displayName` using postgres trigrams.
 - For sqlite we use a simpler approach just to give some reasonable results.


pfrazee:
Nice! This looks great to me.

packages/server/src/api/app/bsky/util/search.ts
```diff
@@ -0,0 +1,158 @@
...
+    .split(/\s+/)
+    .filter(Boolean)
+    .slice(0, 3)
+
+  if (!safeWords.length) {
+    // Return no results. This could happen with weird input like ' % _ '.
+    return db.db.selectFrom('user').where(sql`1 = 0`)
+  }
+
+  // We'll ensured there's a space before each word in both textForMatch and in safeWords,
```
devinivy, line 103:
```suggestion
  // We'll ensure there's a space before each word in both textForMatch and in safeWords,
```

packages/server/src/api/app/bsky/util/search.ts
```diff
@@ -0,0 +1,158 @@
...
+  const { ref } = db.db.dynamic
+  const { term, limit, before } = opts
+
+  // Performing matching by word using "strict word similarity" operator.
+  // The more characters the user gives us, the more we can ratchet down
+  // the distance threshold for matching.
+  const threshold = term.length < 3 ? 0.9 : 0.8
+  const cursor = before !== undefined ? unpackCursor(before) : undefined
+
+  // Matching user accounts based on username
```
devinivy, line 19: I think these queries are alright, but seeing some room for improvement in how they are executed.  Peeled that off into an issue we can poke at later: https://github.com/bluesky-social/adx/issues/239

packages/server/src/api/app/bsky/util/search.ts
```diff
@@ -0,0 +1,158 @@
...
+// E.g. { distance: .94827, name: 'pfrazee' } -> '[0.94827,"pfrazee"]'
+export const packCursor = (row: { distance: number; name: string }): string => {
+  const { distance, name } = row
+  return JSON.stringify([distance, name])
+}
+
+export const unpackCursor = (
+  before: string,
+): { distance: number; name: string } => {
+  const result = JSON.parse(before) /// @TODO bourne
```
dholms, line 136: what does this TODO mean?

is this a Jason Bourne joke? 😆

dholms (approved):
looks great to me as well 👌

packages/server/src/api/app/bsky/util/search.ts
devinivy: Ah thanks for calling this out, I nearly forgot about it.  I think at the end of the day it actually is a Jason Bourne joke haha!  I was thinking of this library which I usually use for JSON parsing untrusted inputs, which can lead to proto poisoning https://github.com/hapijs/bourne.  This one is equally good too: https://github.com/fastify/secure-json-parse.

Because of the way the object is being used I know we aren't subject to proto poisoning, but it can be a nice practice to just use something that handles this so that you don't have to really think about it.

devinivy: I pulled bourne in for the parsing, also helped handle more types of bad cursors (e.g. if it's not JSON at all).

PR #242, opened 2022-10-17
Unify pagination interface across all methods

### LINKED ISSUE #227: Use opaque `cursor` property for all paginated PDS views
While the author feed and home feed use a `cursor` property based on the posts' `createdAt` dates, the other PDS views don't use this convention yet.  We shoud ensure all the paginated views use this convention.  See also https://github.com/bluesky-social/adx/pull/223#discussion_r991370949

devinivy:
This provides a consistent pagination interface across all XRPC methods.  Each paginated method has an optional `cursor` property in its output which can be used with the `before` (or `after`) input in order to fetch the next page of records.  In some cases such as the feeds and user search, the `cursor` property was removed from each of the individual records in the output.  Resolves #227


packages/lex-cli/src/codegen/client.ts
```diff
@@ -306,7 +306,7 @@ function genRecordCls(file: SourceFile, schema: RecordSchema) {
     const method = cls.addMethod({
       isAsync: true,
       name: 'list',
-      returnType: `Promise<{records: ({uri: string, value: ${typeModule}.Record})[]}>`,
+      returnType: `Promise<{cursor?: string, records: ({uri: string, value: ${typeModule}.Record})[]}>`,
```
devinivy, line 309: This change was needed so that the convenience `list()` methods (e.g. `client.app.bsky.post.list()`) supported the cursor output.

packages/server/src/api/app/bsky/getLikedBy.ts
```diff
@@ -48,6 +48,8 @@ export default function (server: Server) {
         encoding: 'application/json',
         body: {
           uri,
+          cid,
```
devinivy, line 51: I saw that `cid` is in the output per the schema but was never being used, so I added it here.  If this wasn't the intent, just give a shout!

packages/server/src/api/app/bsky/getRepostedBy.ts
```diff
@@ -48,7 +48,9 @@ export default function (server: Server) {
         encoding: 'application/json',
         body: {
           uri,
+          cid,
```
devinivy, line 51: This was added for the same reason as https://github.com/bluesky-social/adx/pull/242/files#r997401639

packages/server/src/api/com/atproto/repo.ts
```diff
@@ -57,10 +57,15 @@ export default function (server: Server) {
...
     )
 
+    const lastRecord = records.at(-1)
+    const lastUri = lastRecord && new AdxUri(lastRecord?.uri)
+
     return {
       encoding: 'application/json',
       body: {
         records,
+        // Paginate with `after` by default, paginate with `before` when using `reverse`.
```
devinivy, line 67: Just a note that this is the opposite convention as many of the views since we default here to an ascending sort, but the views descending sort.

packages/xrpc/src/util.ts
```diff
@@ -44,7 +44,9 @@ export function constructMethodCallUri(
       if (!paramSchema) {
         throw new Error(`Invalid query parameter: ${key}`)
       }
-      uri.searchParams.set(key, encodeQueryParam(paramSchema.type, value))
+      if (value !== undefined) {
```
devinivy, line 47: Without this change, if you pass `undefined` as a parameter it would make it into the query as `?param=undefined`.

lexicons/bsky.app/getRepostedBy.json
```diff
@@ -15,6 +15,8 @@
       "required": ["uri", "repostedBy"],
       "properties": {
         "uri": {"type": "string"},
+        "cid": {"type": "string"},
```
devinivy, line 18: I added `cid` here to unify it with `getLikedBy`.   If that wasn't the intention just give a shout 👍

packages/server/src/api/com/atproto/repo.ts
dholms: I wonder if we should flip these semantics just to keep it consistent 🤔 

similar to most other queries, you'd normally want to ask for most recent objects not first-created objects

packages/server/tests/views/author-feed.test.ts
```diff
@@ -110,28 +92,34 @@ describe('pds author feed views', () => {
...
+          before: cursor,
+          limit: 2,
+        },
+        undefined,
+        { headers: sc.getHeaders(dan) },
+      )
+      return res.data
+    }
+
+    const paginatedAll = await paginateAll(paginator)
```
dholms, line 109: nice i like this better then just testing a single slice

packages/server/tests/views/home-feed.test.ts
```diff
@@ -62,9 +61,6 @@ describe('pds home feed views', () => {
 
     expect(forSnapshot(aliceFeed.data.feed)).toMatchSnapshot()
     aliceFeed.data.feed.forEach(expectOriginatorFollowedBy(alice))
-    expect(getCursors(aliceFeed.data.feed)).toEqual(
```
dholms, line 65: is there a reason these tests are taken out here & not on views like likes/follow/reposts?

packages/xrpc/src/util.ts
dholms: nice 👌

dholms (approved):
a couple small questions, but overall looks good 👍 

do you mind making this change for the `getBadgeMembers` view that I just merged as well?

packages/server/tests/views/home-feed.test.ts
devinivy: The reason I removed these is because the value used for pagination doesn't exist on individual feed items anymore.  The users search view is another one like this.  While it would be awesome to test the property explicitly like this, my thought/hope is that the snapshots technically still cover this for us.

packages/server/src/api/com/atproto/repo.ts
devinivy: Got it!  So the default is descending and `reverse` will flip it to ascending.

devinivy:
@dholms should be all set for another peek 👍

dholms:
LGTM 👍

PR #243, opened 2022-10-17
Schema `keyed` hint

dholms:
Added a `keyed` value to lexicon record schemas.

Currently there are two options:
- `"tid"`
- `"literal:${keyName}"`, for instance `"literal:self"`

This is an optional field, and the expected behavior is to default to `"tid"` if not supplied.

Collections at the repository level can have arbitrary keys, but this offers the ability to give a PDS a "hint" for what key they should store a newly created post under.

For instance,
- if I ask my PDS to create a post, it should store it in the posts collection under a new `tid` 
- if I ask my PDS to create a profile, it should store it in the profiles collection under the `self` key

---
I made a couple of codegen changes as well.
- writing the schemas in `schemas.ts` in an dict as well as a list, that way we can easily get certain params (such as `keyed` from the schema)
- also generating an enum-like object for getting the NSID of a schema


packages/common/src/tid.ts
```diff
@@ -32,6 +32,10 @@ export class TID {
     return TID.fromTime(timestamp, clockid)
   }
 
+  static nextStr(): string {
```
dholms, line 35: tempted to make this whole thing not a class. not really giving us anything anymore

pfrazee:
I'm good w/this. Can we change keyed to key? (Why keyed?)

Issue #207, opened 2022-10-04
Transactional DB operations


dholms:
There are several places where DB operations should be transactional:
- repo write operations have several pieces
  - indexing raw record
  - indexing record (if we understand schema)
  - processing notifications (if we understand schema)
  - and setting repo root
- creating a user & setting their repo root
- others? 🤔 

There are a couple places where writes need to ensure consistency with the current DB value:
- repo root updates to ensure we don't fork history
- invite code uses to ensure they aren't double-spent

dholms:
fixed: https://github.com/bluesky-social/atproto/pull/245

PR #245, opened 2022-10-18
PDS transactions

devinivy:
Sweeping the PDS to ensure the methods are atomic where necessary.  This will be easiest to review while [ignoring whitespace](https://github.com/bluesky-social/adx/pull/245/files?&w=1).

Addresses #207


packages/server/src/api/com/atproto/account.ts
```diff
@@ -81,63 +62,111 @@ export default function (server: Server) {
...
+          throw new InvalidRequestError(
+            'No invite code provided',
+            'InvalidInviteCode',
+          )
+        }
 
-    // @TODO consider pushing this to the db, and checking for a
-    // uniqueness error during registerUser(). Main blocker to doing
-    // that now is that we need to create a did prior to registration.
+        await dbTxn.lockTable('invite_code_use')
```
dholms, line 74: not thrilled with throwing a manual lock on the table... it's better than using SERIALIZABLE txs, and I couldn't find another way of doing this. lmk if you ahve any other ideas

devinivy: Scratch that.  Selecting the invite code `FOR UPDATE` (even though it's not going to be updated) is simpler than both of those options, and gives you the per-invite granularity 🙃

devinivy: Here are some options for this: https://github.com/bluesky-social/atproto/pull/253

packages/server/src/api/com/atproto/repo.ts
```diff
@@ -117,30 +118,49 @@ export default function (server: Server) {
...
+        }),
+      )
+
+      const repo = await RepoStructure.load(blockstore, currRoot)
+      await repo
+        .stageUpdate(cidWriteOps)
+        .createCommit(authStore, async (prev, curr) => {
+          const success = await db.updateRepoRoot(did, curr, prev)
+          if (!success) {
+            throw new Error('Could not update repo root')
```
devinivy, line 159: We might consider turning this into a 409 Conflict for HTTP-land.

packages/server/src/api/com/atproto/repo.ts
```diff
@@ -162,37 +182,57 @@ export default function (server: Server) {
...
+        .stageUpdate({
+          action: 'create',
+          collection,
+          rkey,
+          cid,
+        })
+        .createCommit(authStore, async (prev, curr) => {
+          const success = await txn.updateRepoRoot(did, curr, prev)
+          if (!success) {
+            throw new Error('Could not update repo root')
```
devinivy, line 230: Same here, we might consider turning this into a 409 for HTTP-land.

packages/server/src/api/app/bsky/updateProfile.ts
```diff
@@ -3,115 +3,134 @@ import { AuthRequiredError, InvalidRequestError } from '@atproto/xrpc-server'
...
-    // @TODO transactionalize
-    await Promise.all([
-      recordQuery,
-      profileQuery,
-      delBadgesQuery,
-      addBadgesQuery,
-      db.updateRepoRoot(requester, repo.cid),
-    ])
+    const { profileCid, updated } = await db.transaction(async (txnDb) => {
+      const currRoot = await txnDb.getRepoRoot(requester, true)
```
devinivy, line 22: I believe the `forUpdate` here has fixed the race I mentioned before, where two concurrent profile updates could clobber each other. 😎👍

devinivy (commented):
Left a few thoughts, but all the substantive stuff is looking real solid to me.  Headsup that I wont be able to mark this with an actual approval since I opened the PR 🙃

packages/server/src/api/com/atproto/repo.ts
devinivy: (And there's one more below.)

packages/server/package.json
```diff
@@ -9,7 +9,7 @@
     "postbuild": "tsc --build tsconfig.build.json",
     "start": "node dist/bin.js",
     "test": "jest",
-    "test:pg": "../pg/with-test-db.sh jest",
+    "test:pg": "../pg/with-test-db.sh yarn test",
```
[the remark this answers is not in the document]
dholms: no issues. but when I want to drill in on a certain test file & was hardcoding it in the `test` command & this just makes it so I don't have to hardcode twice 😛

packages/server/src/db/index.ts
```diff
@@ -338,6 +399,12 @@ export class Database {
     }
     return found
   }
+
+  async lockTable(tableName: string): Promise<void> {
+    if (this.dialect === 'pg') {
+      await sql`LOCK TABLE ${sql.ref(tableName)} IN SHARE MODE`.execute(this.db)
+    }
+  }
```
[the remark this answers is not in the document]
dholms: ah good looks 👌

packages/server/src/db/records/badgeAccept.ts
```diff
@@ -93,7 +93,7 @@ const insertFn =
 const deleteFn =
   (db: Kysely<PartialDB>) =>
   async (uri: AtUri): Promise<void> => {
-    await db.deleteFrom(tableName).where('uri', '=', uri.toString())
+    await db.deleteFrom(tableName).where('uri', '=', uri.toString()).execute
```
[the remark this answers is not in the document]
dholms: 😱

packages/server/package.json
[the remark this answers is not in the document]
devinivy: Ahhh got it!  I just hadn't run into it since I've been tacking the file onto the end of the command, e.g.:
```
yarn workspace @atproto/server test:pg packages/server/tests/views/profile.test.ts
```

packages/server/tests/crud.test.ts
```diff
@@ -171,6 +172,59 @@ describe('crud operations', () => {
...
+      for (const uri of uris) {
+        await expect(
+          aliceClient.com.atproto.repoGetRecord({
+            user: alice.did,
+            collection: uri.collection,
+            rkey: uri.rkey,
+          }),
+        ).rejects.toThrow(Error)
+      }
+      uris.map
```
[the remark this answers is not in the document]
dholms: nope good to nix 👌

packages/server/src/api/com/atproto/sync.ts
```diff
@@ -34,43 +33,36 @@ export default function (server: Server) {
...
-
-    // check to see if we have their username in DB, for indexed queries
-    // const haveUsername = await db.isDidRegistered(did)
-    // if (!haveUsername) {
-    //   const username = await service.getUsernameFromDidNetwork(did)
-    //   if (username) {
-    //     const [name, host] = username.split('@')
-    //     await db.registerDid(name, did, host)
-    //   }
+    // // we don't need auth here because the auth is on the data structure 😎
```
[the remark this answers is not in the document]
dholms: ah i forgot i committed up the commenting. yeah i'll just wipe this out & throw an error

packages/server/src/db/index.ts
```diff
@@ -115,23 +131,39 @@ export class Database {
...
-  async getRepoRoot(did: string): Promise<CID | null> {
-    const found = await this.db
+  async getRepoRoot(did: string, forUpdate?: boolean): Promise<CID | null> {
+    let builder = this.db
       .selectFrom('repo_root')
       .selectAll()
       .where('did', '=', did)
-      .executeTakeFirst()
+    if (forUpdate && this.dialect !== 'sqlite') {
+      // SELECT FOR UPDATE is not supported by sqlite, but sqlite txs are SERIALIZABLE so we don't actually need it
```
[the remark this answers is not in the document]
dholms: ah that's a good point 👍

packages/server/src/api/app/bsky/updateProfile.ts
dholms: 💪

packages/server/src/api/com/atproto/repo.ts
dholms: Still this feels like a special enough case (& encountered in several spots) that we may want a special error type to handle it?

packages/server/src/db/index.ts
```diff
@@ -146,7 +146,7 @@ export class Database {
       .selectFrom('repo_root')
       .selectAll()
       .where('did', '=', did)
-    if (forUpdate && this.dialect !== 'sqlite') {
+    if (forUpdate && this.dialect !== 'sqlite' && this.isTransaction) {
```
devinivy, line 149: This seems reasonable to me, though in sqlite this could be run outside of a transaction in which case no real locking would happen.  If that matters to us, could do something like this:
```js
if (forUpdate) {
  this.assertTransaction()
  if (this.dialect !== 'sqlite') {
    builder = builder.forUpdate()
  }
}
```

packages/server/package.json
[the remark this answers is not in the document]
dholms: ah that's not a bad idea

packages/server/src/api/com/atproto/repo.ts
devinivy: That makes a lot of sense.  Barring a special error type, maybe it's the kind of thing we'd want to log.

dholms: yeah that's a good idea

dholms (approved):
_pats self on back_

PR #247, opened 2022-10-19
Immutable repo structure

dholms:
This PR introduces a new class: `RepoStructure` which is an immutable version of our previous `Repo`.
It also does not maintain any internal notion of and `auth.AuthStore` or `auth.Verifier`.

`RepoStructure` allows for 2 primary operations:
- `stageUpdate`: stages updates in memory as a list of "operations" which include `action` type, `collection`, `rkey`, & `cid`
- `createCommit`: apply those updates & create a valid commit

`createCommit` takes an optional callback param with both the `prev` & `curr` commit hashes. This lets us reimplement `safeCommit` but also exposes enough information to maintain atomic consistency in SQL txs for repo root updates on the server. The callback can inform `createCommit` if the write was successful or provide with a new CID to "rebase" the proposed commit on top of.

I also split some of the utility/verification methods into functions in supporting files.

`Repo` is still around with a similar interface, but it wraps around & makes use of `RepoStructure` under the hood. Frankly, I'm not sure we'll even be using `Repo` on the server, but it's a bit more dev-friendly & could be useful in other contexts so I kept it around 🤷‍♂️


packages/repo/src/structure.ts
```diff
@@ -0,0 +1,289 @@
...
+  async getFullHistory(): Promise<Uint8Array> {
+    return this.getDiffCar(null)
+  }
+
+  private async openCar(
+    fn: (car: BlockWriter) => Promise<void>,
+  ): Promise<Uint8Array> {
+    const { writer, out } = CarWriter.create([this.cid])
+    await fn(writer)
+    writer.close()
```
devinivy, line 229: I wonder if this could cause some kind of leak without a little extra love:
```suggestion
    try {
      await fn(writer)
    } finally {
      writer.close()
    }
```

devinivy (approved):
A lot of good things happening here!

packages/repo/src/blockstore/ipld-store.ts
```diff
@@ -70,7 +71,20 @@ export abstract class IpldStore {
...
+  async loadCar(buf: Uint8Array): Promise<CID> {
+    const car = await CarReader.fromBytes(buf)
+    const roots = await car.getRoots()
+    if (roots.length !== 1) {
+      throw new Error(`Expected one root, got ${roots.length}`)
+    }
+    const rootCid = roots[0]
+    for await (const block of car.blocks()) {
+      await this.putBytes(block.cid, block.bytes)
+    }
```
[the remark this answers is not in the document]
dholms: yup good looks 👌

packages/repo/src/structure.ts
dholms: ah excellent. thanks for the catch 🙏

PR #249, opened 2022-10-20
Setup PLC server with postgres and migrations

devinivy:
The PLC server now supports the postgres dialect.  Also sets-up a system for managing incremental db migrations.  Some of this will be a little easier to review while ignoring whitespace.

### DIFF
--- a/packages/plc/build.js
+++ b/packages/plc/build.js
@@ -13,10 +13,11 @@ require('esbuild')
     platform: 'node',
     assetNames: 'src/static',
     external: [
-      './node_modules/better-sqlite3/*',
-      '../../node_modules/better-sqlite3/*',
+      'better-sqlite3',
       '../../node_modules/level/*',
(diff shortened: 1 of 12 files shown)


packages/plc/build.js
```diff
@@ -13,10 +13,11 @@ require('esbuild')
     platform: 'node',
     assetNames: 'src/static',
     external: [
-      './node_modules/better-sqlite3/*',
-      '../../node_modules/better-sqlite3/*',
+      'better-sqlite3',
```
devinivy, line 16: I couldn't find any downside to this change (e.g. dev-env still seems happy), and it played nicer for me when consuming the build from outside the monorepo.

dholms: ah nice. that's probably fine. my whole method with these was guess & check anyway 😅 so id say if it builds & runs you're good

dholms (approved):
yup this is looking good 👌

PR #253, opened 2022-10-21
Transactional invite code use, username/email creation

devinivy:
Picking-up from https://github.com/bluesky-social/atproto/pull/245#discussion_r1001288568, I wanted to throw out a couple options for avoiding a. double-spends of invite codes w/o a table lock and b. multiple users claiming the same username or email.

For the first one, I think we can replace the table lock for invite_code_use with a row lock on the invite_code itself.  That should be a bit more granular, allowing different invite codes to be used in parallel without blocking each other.  I did this using a `SELECT FOR UPDATE` on postgres.  On sqlite this isn't needed since the transactions are already serialized.

The test for races for username was relying on the invite_code_use table lock.  I changed the test to show what happens when the users trying to claim the same username aren't using the same invite code, and as expected it exposed a race.  I think enforcing uniqueness of username/email is best enforced at the database level using a uniqueness constraint.  So I added that constraint into the db, and when creating the user added an `ON CONFLICT DO NOTHING`.  We can detect whether the insert was successful by checking if there was any result of the insert (this was just a bit more convenient to code around versus surfacing the a uniqueness constraint error).

Thanks for surfacing these issues @dholms 🙏 🙏  All good to take these changes as-is, not take them at all, or adapt them— mostly just wanted to get these options out there and expose the username race!

### DIFF
--- a/packages/server/src/api/com/atproto/account.ts
+++ b/packages/server/src/api/com/atproto/account.ts
@@ -6,7 +6,7 @@ import { PlcClient } from '@atproto/plc'
 import * as uint8arrays from 'uint8arrays'
 import { Server } from '../../../lexicon'
 import * as locals from '../../../locals'
-import { countAll, keys, selectValues, vals } from '../../../db/util'
+import { countAll } from '../../../db/util'
 import { UserAlreadyExistsError } from '../../../db'
 
 export default function (server: Server) {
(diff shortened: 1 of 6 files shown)


packages/server/src/api/com/atproto/account.ts
```diff
@@ -71,32 +71,36 @@ export default function (server: Server) {
...
-              validInviteQuery(dbTxn, inviteCode),
-            ),
-          )
-          .returning('code')
+        const invite = await dbTxn.db
+          .selectFrom('invite_code')
+          .selectAll()
+          .where('code', '=', inviteCode)
+          // Lock invite code to avoid duplicate use
+          .if(dbTxn.dialect === 'pg', (qb) => qb.forUpdate())
```
dholms, line 79: oh cool didnt know you could inline "if"s like this

dholms (approved):
okay nice this looks excellent 👌

PR #254, opened 2022-10-21
Setup PDS server with migrations

devinivy:
Following the approach taken in #249, this adds the same system for db migrations to the PDS as used in PLC.  The `KyselyWithDialect` thing from #249 to access the db dialect in migrations didn't work the way I hoped it would.  So I got rid of that in both PDS and PLC, and instead went a different direction to allow that to happen.


packages/server/src/api/app/bsky/updateProfile.ts
```diff
@@ -87,11 +87,13 @@ export default function (server: Server) {
...
-      .where('profileUri', '=', uri.toString())
-      .where('badgeUri', 'in', toDelete)
-      .execute()
+    const delBadgesQuery = toDelete.length
+      ? db.db
+          .deleteFrom('app_bsky_profile_badge')
+          .where('profileUri', '=', uri.toString())
+          .where('badgeUri', 'in', toDelete)
+          .execute()
+      : null
```
devinivy, line 96: Not sure how/why I didn't see this issue until now, but on postgres an empty `toDelete` list was causing a syntax error.

dholms (approved):
sweet this looks good to me 👌 

i'd pull the changes from https://github.com/bluesky-social/atproto/pull/257 into here as well specifically re: marking better-sqlite `external` & running migrations in `server/bin.ts`

PR #256, opened 2022-10-23
Basic CI

harlantwood:
Hello again & congrats on the atproto release!  

I noticed that CI, which I originally added in https://github.com/bluesky-social/atproto/pull/120, went away in a11c1754a2aa5126e2662cb1e92813913a52bc51 -- so this commit adds it again, but more simply now: just calling the top level monorepo scripts: `build`, `test`, `verify`.

As of this writing, `verify` and `test` are failing on my branch: https://github.com/harlantwood/adx/actions/runs/3306753316 

My suggestion is to merge this PR first, as it just gives visibility into the current state, and then merge fixes for those issues; but happy to rebase if you prefer to do fixes first.

PR #257, opened 2022-10-23
server: fix startup by adding better-sqlite3 to externals, create tables

zhuowei:


packages/server/src/bin.ts
```diff
@@ -28,12 +29,20 @@ const run = async () => {
     blockstore = new MemoryBlockstore()
   }
 
+  let createDbTables = false;
+
   if (cfg.databaseLocation) {
+    createDbTables = !fs.statSync(cfg.databaseLocation, {throwIfNoEntry: false});
```
devinivy, line 35: Just a headsup that the linter will ask it to look more like this:
```suggestion
    createDbTables = !fs.statSync(cfg.databaseLocation, { throwIfNoEntry: false });
```

With #256 it should start getting flagged in CI, though.

devinivy (approved):
Thanks for the contribution!  Just a headsup that once I merge this I will incorporate it with #254, which will replace the `createDbTables` check with a more robust system for tracking migrations.

PR #258, opened 2022-10-23
sync: fix syncing empty repo not using specified plc url param

zhuowei:


dholms:
Thank you for the thoughtful PRs. You correctly scoped out that the sync methods are not working at present & both related PRs are reasonable fixed.

However, there are some open questions/issues that will require additional work as well:
- tracking which repos are from users _on_ the server vs off
- and related^^ if the current RepoRoot is the authoritative HEAD or just the "most recently known"
- should the PDS accept repo pushes from repos/collections it is not subscribed to?
- without user-held keys at present, this route is meaningless for repos managed by the server

We are planning on doing a larger piece of work related to this in the coming weeks, and I was actually intending to delete the current sync code & simply return an error this week.

Again, thank you for the PRs, but I'll be closing for now 🙏

PR #260, opened 2022-10-24
Update createAccount to accept email and inviteCode and allocate DIDs

laMoradaPostrera:
sign up a registry cobosfrancoinc colombia villavicencio


laMoradaPostrera:
hello blue how is it going with green and white only my bet just in case to guess
________________________________

PR #263, opened 2022-10-24
Recovery key

dholms:
- add recovery key to server config
- allow user-supplied recovery keys on account creation
- testing that user DIDs get created correctly
- some refactor of account.test.ts to remove duplicate code


packages/server/src/config.ts
```diff
@@ -47,6 +48,8 @@ export class ServerConfig {
 
     const didPlcUrl = process.env.DID_PLC_URL || 'http://localhost:2582'
     const serverDid = process.env.SERVER_DID || ''
+    const recoveryKey = process.env.RECOVERY_KEY || ''
```
devinivy, line 51: Is it fair to say this is required?  If so, we might want to eagerly fail if it's missing.

devinivy (approved):
Looks good!

packages/server/src/config.ts
dholms: yeah that's a good point 👍

PR #265, opened 2022-10-24
PDS cleanup

dholms:
- removing unneeded TODOs
- throwing errors on unimplemented methods
- removing unused code

Optional buddy PR: https://github.com/bluesky-social/atproto/pull/266


packages/server/src/locals.ts
```diff
@@ -97,7 +97,6 @@ export const plcClient = (res: Response): plc.PlcClient => {
 
 export const getAuthstore = (res: Response, did: string): AuthStore => {
   const { auth, keypair } = get(res)
-  // @TODO check that we can sign on behalf of this DID
```
dholms, line 100: this TODO is unnecessary because we can just attempt the signature & it'll just fail if not allowed

packages/server/src/routes/index.ts
```diff
@@ -1,7 +0,0 @@
-import express from 'express'
-import WellKnown from './well-known'
-const router = express.Router()
-
-router.use('/.well-known', WellKnown)
-
-export default router
```
dholms, line 7: this router wasn't even getting hooked up anywhere

also this is a did:web route so it doesn't really make sense to server did:plc off of it

packages/server/src/subscriptions.ts
```diff
@@ -1,62 +0,0 @@
-import { Repo } from '@atproto/repo'
-import Database from './db'
-
-// @TODO revamp all subscriber logic
```
dholms, line 4: this will all get rewritten at some point so 🔪

devinivy (approved):
👍

PR #267, opened 2022-10-25
Secp256k1 support

dholms:
Adds secp256k1 support to 
- crypto lib
- did:plc server
- did resolution


packages/crypto/src/did.ts
```diff
@@ -38,6 +41,8 @@ export const formatDidKey = (jwtAlg: string, keyBytes: Uint8Array): string => {
   }
   if (jwtAlg === 'ES256') {
     keyBytes = p256.compressPubkey(keyBytes)
+  } else if (jwtAlg === 'ES256K') {
+    keyBytes = secp.compressPubkey(keyBytes)
```
devinivy, line 45: We could use the new JWT algorithm consts here.

dholms: oh good find thx 🙏

dholms: fixing in the other locations too 👌

devinivy (approved):
Looks great 👍

PR #268, opened 2022-10-25
Update resolver.ts

dmw94:


pfrazee:
If you submit a PR like this again (no description, no useful changes) we'll ban you from the org. Do not waste our time.

PR #272, opened 2022-10-26
Miscellaneous dep and build fixes

devinivy:
Just a little of this and that:
 - Changing the leveldb externals for builds to use the same convention we started using for sqlite in #249, which is a little simpler to configure and works equally well or better.
 - Add a build script to the repo package in order to list leveldb externals.
 - Add some missing dependencies to packages, e.g. auth depends on did-resolver, so did-resolver has been added to auth's package.json.
 - Support config overrides when using pds's `ServerConfig.readEnv()`.

### DIFF
--- a/packages/auth/package.json
+++ b/packages/auth/package.json
@@ -16,6 +16,7 @@
   },
   "dependencies": {
     "@atproto/crypto": "*",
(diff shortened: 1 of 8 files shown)


dholms (approved):
sweet 👌

Issue #274, opened 2022-10-26
Yarn build error on better-sqlite3


brh28:
When executing `yarn` from the root directory, I'm getting the following error:

`error /home/benhindman/Workspace/atproto/packages/pds/node_modules/better-sqlite3, /home/benhindman/Workspace/atproto/packages/plc/node_modules/better-sqlite3: Command failed.`
...
```
gyp ERR! UNCAUGHT EXCEPTION 
gyp ERR! stack TypeError: Cannot assign to read only property 'cflags' of object '#<Object>'
gyp ERR! stack     at createConfigFile (/home/benhindman/Workspace/atproto/node_modules/@npmcli/run-script/node_modules/node-gyp/lib/configure.js:117:21)
gyp ERR! stack     at /home/benhindman/Workspace/atproto/node_modules/@npmcli/run-script/node_modules/node-gyp/lib/configure.js:84:9
gyp ERR! stack     at FSReqCallback.oncomplete (node:fs:189:23)
gyp ERR! System Linux 5.15.0-52-generic
gyp ERR! command "/home/benhindman/.nvm/versions/node/v19.0.0/bin/node" "/home/benhindman/Workspace/atproto/node_modules/.bin/node-gyp" "rebuild" "--release"
gyp ERR! cwd /home/benhindman/Workspace/atproto/packages/pds/node_modules/better-sqlite3
```

Any troubleshooting tips would be appreciated.

OS: Ubuntu 22.04.1
node: v19.0.0
yarn: 1.22.19

brh28:
Switched to node v18. The build now completes without failing. However, when I run`yarn start` from the `dev-env` directory, I get a `Error: Cannot find module 'better-sqlite3'` so still not sure if the build is running as expected.  

I do believe better-sqlite3 is compatible with my system. At least, I am able to install with `yarn add better-sqlite3` in an empty directory.

BTW, I seeing a console message about upgrading to node v19. Not sure why that would be there

brh28:
Manually added better-sqlite3 in the dev-env dir and the PDS server does startup correctly now. Not sure if I'm going to run into issues down the road though

devinivy:
I am seeing the same thing.  I think I just bonked that a couple hours ago in #272— will take a closer look at it before long.

brh28:
Sounds good. Thanks for the help!

dholms:
yeah these build scripts are pretty error prone, fixed it up here: https://github.com/bluesky-social/atproto/pull/275

[...]

PR #275, opened 2022-10-26
Fix a couple build issues

dholms:
Two things:
- external dependency issue in: https://github.com/bluesky-social/atproto/issues/274
  - the problem here being that `dev-env` doesn't install `better-sqlite3` itself & so needs to reference the location where it is _actually_ installed
- we weren't supply the required env variables in server/bin. For this simple setup it just sets the recovery did to the signing key

### DIFF
--- a/packages/dev-env/build.js
+++ b/packages/dev-env/build.js
@@ -8,9 +8,8 @@ require('esbuild')
     outdir: 'dist',
     platform: 'node',
     external: [
-      '../plc/node_modules/better-sqlite3/*',
-      '../pds/node_modules/better-sqlite3/*',
-      '../../node_modules/classic-level/*',
+      'better-sqlite3',
+      'classic-level',
       // Referenced in pg driver, but optional and we don't use it
       'pg-native',
     ],
--- a/packages/dev-env/build.js
+++ b/packages/dev-env/build.js
@@ -8,8 +8,9 @@ require('esbuild')
     outdir: 'dist',
     platform: 'node',
     external: [
-      'better-sqlite3',
-      'classic-level',
+      '../plc/node_modules/better-sqlite3/*',
+      '../pds/node_modules/better-sqlite3/*',
+      '../../node_modules/classic-level/*',
       // Referenced in pg driver, but optional and we don't use it
       'pg-native',
     ],
--- a/packages/pds/src/bin.ts
+++ b/packages/pds/src/bin.ts
@@ -20,6 +20,9 @@ const run = async () => {
   let blockstore: IpldStore
   let db: Database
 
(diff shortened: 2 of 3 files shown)


dholms:
wait actually... we might be able to hoist `better-sqlite3` 🤔 

gonna try something out

packages/dev-env/build.js
```diff
@@ -8,8 +8,9 @@ require('esbuild')
     outdir: 'dist',
     platform: 'node',
     external: [
-      'better-sqlite3',
-      'classic-level',
+      '../plc/node_modules/better-sqlite3/*',
+      '../pds/node_modules/better-sqlite3/*',
+      '../../node_modules/classic-level/*',
```
devinivy, line 13: Yupyup, thanks for fixing-up my slop.  The externals config in a monorepo really messes with me!

I think it's already clear to you, but what I'm seeing is that this change seems to alter how `better-sqlite3` is referenced within the build, though it doesn't actually include or exclude anything from the build.

packages/pds/src/bin.ts
```diff
@@ -20,6 +20,9 @@ const run = async () => {
   let blockstore: IpldStore
   let db: Database
 
+  const keypair = await crypto.EcdsaKeypair.create()
+  process.env.RECOVERY_KEY = keypair.did()
+
   const cfg = ServerConfig.readEnv()
```
devinivy, line 26: I've altered `readEnv()` so that it can receive overrides.  So I think this would also work:
```suggestion
  const keypair = await crypto.EcdsaKeypair.create()

  const cfg = ServerConfig.readEnv({ recoveryKey: keypair })
```

devinivy (approved):
🥂

packages/pds/src/bin.ts
dholms: oh nice that's much cleaner 👌

packages/dev-env/build.js
dholms: heh no worries i've beat my head against this so many times.

i actually think you're change is more the direction we should be going. still poking around the hoisting but seems to be working out

dholms:
kk @devinivy check out that change actually.

taking better-sqlite3 out of `nohoist` it's now at package root, only installed once, we don't have to balance multiple references to it or point to binaries from other packages

don't quite remember why i was nohoisting in the past. the old sqlite library may have required it 🧐 or it may have just been mistakenly added in at a lowpoint of tooling refactor delirium 😛

dholms:
oh this will require wiping out your root node_modules & reinstalled btw. otherwise it doesn't bother with the hoisting

devinivy (approved):
Oh yeah, I like that a lot.  Good thought 💎

PR #276, opened 2022-10-26
SQL Blockstore

dholms:


packages/pds/src/db/tables/record.ts
```diff
@@ -4,7 +4,6 @@ export interface Record {
   did: string
   collection: string
   rkey: string
-  raw: string
   receivedAt: string
   indexedAt: string
```
dholms, line 8: we have `indexedAt` in a lot of places:
- blockstore
- this record table
- & the structure table

wondering if we should simplify this or leave as is 🤔

packages/pds/src/db/migrations/20221021T162202001Z-init.ts
```diff
@@ -32,6 +33,11 @@ export async function up(db: Kysely<unknown>, dialect: Dialect): Promise<void> {
       if (!err?.detail?.includes?.('(pg_trgm) already exists')) throw err
     }
   }
+
+  // Postgres uses the type `bytea` for variable length bytes
+  // Kysely doesn't have this as a "supported data type", but it is 😛
+  const binaryDatatype = dialect === 'sqlite' ? 'blob' : ('bytea' as any)
```
devinivy, line 39: Small peas, but if you want to avoid the `any`, I think this is how kysely supports types it may not know about:
```suggestion
  const binaryDatatype = dialect === 'sqlite' ? 'blob' : sql`bytea`
```

packages/pds/src/db/migrations/20221021T162202001Z-init.ts
```diff
@@ -77,10 +83,18 @@ export async function up(db: Kysely<unknown>, dialect: Dialect): Promise<void> {
...
     .addColumn('rkey', 'varchar', (col) => col.notNull())
-    .addColumn('raw', 'text', (col) => col.notNull())
     .addColumn('receivedAt', 'varchar', (col) => col.notNull())
     .addColumn('indexedAt', 'varchar', (col) => col.notNull())
     .execute()
+  // Ipld Blocks
+  await db.schema
+    .createTable(ipldBlockTable)
+    .addColumn('cid', 'varchar', (col) => col.primaryKey())
+    .addColumn('did', 'varchar', (col) => col.notNull())
```
devinivy, line 93: If two dids attempt to store the same cid, could we get into trouble?  Or perhaps there's some reason that is guaranteed not to happen.

devinivy: Unless there's a very clear reason why that shouldn't happen, I might suggest making this a compound key `(did, cid)`, and then add an index on `cid`.  Or if we want to avoid the possible duplicate storage of `content`, breaking the did relationship out into a separate table.

packages/pds/src/db/tables/record.ts
devinivy: I am totally pickin' up what you're putting down.  Just one idea— maybe we keep it on the blockstore and on the secondary indexes, but remove it from the record table.  My thought there is that the secondary index tables technically could end-up combining data from multiple records, and we'd want to keep the indexed time of the latest data.  The record table, on the other hand, is pretty much guaranteed to reference a single block.

packages/pds/src/db/migrations/20221021T162202001Z-init.ts
dholms: ah excellent 🙏

packages/pds/src/db/tables/record.ts
dholms: seems reasonable to me 👌

packages/pds/src/db/migrations/20221021T162202001Z-init.ts
dholms: ugh you're right...

I'd like for `cid` to be the primary key here, it just sits better with me. So I think i'll break out the did-cid relationship 🤔

packages/pds/src/sql-blockstore.ts
```diff
@@ -0,0 +1,49 @@
...
+    return found.content
+  }
+
+  async putBytes(cid: CID, bytes: Uint8Array): Promise<void> {
+    const row = {
+      cid: cid.toString(),
+      did: this.did,
+      size: bytes.length,
+      content: bytes,
+      indexedAt: new Date().toISOString(),
```
devinivy, line 35: Random thought here: it would be kind of nice if rows written in the same transaction had the same `indexedAt` times.  If we want to achieve that with the blockstore, the the constructor for `SqlBlockstore` could received a date, and if we're in a transaction then that date would be used for `indexedAt`.  In all other cases, it would work as it currently does.

packages/pds/src/db/migrations/20221021T162202001Z-init.ts
devinivy: Sounds like a plan 👍

packages/pds/src/sql-blockstore.ts
devinivy: May also be a little funky.  Honestly no strong feelings on it, especially if we're nixing the `indexedAt` column on the records table anyway.

dholms: yeah that's not a bad idea. could apply that idea to other transactional operations as well 🤔

packages/pds/src/sql-blockstore.ts
```diff
@@ -27,18 +31,21 @@ export class SqlBlockstore extends IpldStore {
...
   async putBytes(cid: CID, bytes: Uint8Array): Promise<void> {
-    const row = {
-      cid: cid.toString(),
-      did: this.did,
-      size: bytes.length,
-      content: bytes,
-      indexedAt: new Date().toISOString(),
-    }
-    await this.db.db
+    const insertBlock = this.db.db
```
devinivy, line 34: Only if you're feelin it.
```suggestion
    this.db.assertTransaction()
    const insertBlock = this.db.db
```

devinivy (approved):
Real nice!

packages/pds/src/sql-blockstore.ts
dholms: hmmm yeah that probably makes sense 🤔 

if for some reason we find we need to do this non-transactionally, we can loosen the constraint

PR #278, opened 2022-10-28
Name Resolution

dholms:
- splits user table into two tables
  - users for info about users with accounts on the pds
  - user_dids for a simple mapping of username <-> did for any user encountered on the network
- remove test username registry
- resolve name queries off of user table
- add `availableUserDomains` config var
- add `serverDid` config var
- updated tests to use properly formatted usernames

Still to do:
Username validation beyond what exists. What constraints do we want to put on this? length? valid chars? all lower case?


packages/pds/src/db/index.ts
```diff
@@ -186,13 +187,16 @@ export class Database {
...
+    let query = this.db
+      .selectFrom('user')
+      .innerJoin('user_did', 'user_did.username', 'user.username')
+      .selectAll()
     if (usernameOrDid.startsWith('did:')) {
       query = query.where('did', '=', usernameOrDid)
     } else {
       query = query.where(
-        sql`lower(username)`,
+        sql`lower(user.username)`,
```
dholms, line 199: this lines giving me hell. works fine with sqlite but not pg

with "username" it gives the error
`column reference \"username\" is ambiguous`

with "user.username" it gives the error
`syntax error at or near \".\""`

packages/pds/src/db/index.ts
```diff
@@ -216,74 +221,57 @@ export class Database {
...
       .executeTakeFirst()
     if (!inserted) {
       throw new UserAlreadyExistsError()
     }
-    log.info({ username, email, tempDid }, 'pre-registered user')
+    log.info({ username, email }, 'registered user')
   }
 
-  async postRegisterUser(tempDid: string, did: string) {
+  async registerUserDid(username: string, did: string) {
```
devinivy, line 245: This feels so much better.

packages/pds/src/api/com/atproto/account.ts
```diff
@@ -40,7 +31,6 @@ export default function (server: Server) {
     // In order to perform the significant db updates ahead of
     // registering the did, we will use a temp invalid did. Once everything
     // goes well and a fresh did is registered, we'll replace the temp values.
-    const tempDid = uint8arrays.toString(randomBytes(16), 'base32')
```
devinivy, line 43: 💎

packages/pds/src/api/com/atproto/account.ts
```diff
@@ -52,17 +42,17 @@ export default function (server: Server) {
...
-    let isTestUser = false
-    if (username.endsWith('.test')) {
-      if (!config.debugMode || !config.testNameRegistry) {
-        throw new InvalidRequestError(
-          'Cannot register a test user if debug mode is not enabled',
-        )
-      }
-      isTestUser = true
+    const supportedUsername = config.availableUserDomains.some((host) =>
+      username.toLowerCase().endsWith(host),
```
devinivy, line 46:
```suggestion
      username.toLowerCase().endsWith(`.${host}`),
```

devinivy (approved):
Looks good, this feels much more solid.

packages/pds/src/api/com/atproto/account.ts
devinivy: Oops, ignore this ☝️ I see the hosts already start with a dot.

Perhaps the alternate fix is just to validate that in `ServerConfig`.  The spirit of the suggestion is that we don't want to end-up in a situation where accounts are only allowed from `mydomain.com`, and someone is able to sign-up with a `notmydomain.com` account.

packages/pds/src/config.ts
```diff
@@ -82,16 +90,26 @@ export class ServerConfig {
...
       emailNoReplyAddress,
       ...overrides,
     })
+
+    const invalidDomain = cfg.availableUserDomains.find(
+      (domain) => domain.length > 0 && domain.startsWith('.'),
+    )
+    if (invalidDomain) {
+      throw new Error(`Invalid domain: ${invalidDomain}`)
+    }
```
devinivy, line 110: What if this validation happened in `ServerConfig`'s constructor?  The only upside there is that no matter how you create the `ServerConfig`, it will be validated.

Sort of related, I think that the tests `availableUserDomains: ['.test']` would fail this validation though we want it to succeed.  Perhaps we want something like this:
```suggestion
    const invalidDomain = cfg.availableUserDomains.find(
      (domain) => !domain.startsWith('.'),
    )
    if (invalidDomain) {
      throw new Error(`Invalid domain: ${invalidDomain}`)
    }
```

PR #280, opened 2022-10-28
Misc HTTP error fixes

### LINKED ISSUE #273: PDS gives internal error rather than 404 when hitting unknown route
When hitting a route on the PDS outside of `/xrpc/*` (e.g. `curl http://localhost:3000/x`), the server serves a 500 due to a runtime TypeError.  Instead, I would expect to see a 404.  Here's that TypeError:
```
TypeError: res.status is not a function
    at handler2 (/atproto/packages/pds/dist/index.js:96487:7)
    at newFn2 (/atproto/packages/pds/dist/index.js:68772:20)
    at Layer2.handle [as handle_request] (/atproto/packages/pds/dist/index.js:574:9)
    at trim_prefix (/atproto/packages/pds/dist/index.js:1416:17)
    at /atproto/packages/pds/dist/index.js:1389:13
    at Function.process_params (/atproto/packages/pds/dist/index.js:1424:16)
    at next (/atproto/packages/pds/dist/index.js:1383:15)
    at /atproto/packages/pds/dist/index.js:1602:19
    at next (/atproto/packages/pds/dist/index.js:1376:18)
    at Function.handle (/atproto/packages/pds/dist/index.js:1319:7)
```

A secondary thing here— it may be nice to get sourcemaps going so that these errors are a little easier to track down.  Happy to peel that off into a separate issue if it sounds like it might be desirable.

devinivy:
A few little http fixes in here:
 - Avoid double-decoding route params in PLC.
 - Avoid including error details from unexpected errors.
 - Do not reflect user input into HTML output on PLC, move to JSON error responses on PLC.
 - Check if headers are already sent in http error handlers.
 - Fix an issue where 404s were turning into 500s on the PDS, resolves #273.

### DIFF
--- a/packages/pds/tests/server.test.ts
+++ b/packages/pds/tests/server.test.ts
@@ -1,0 +1,44 @@
+import { CloseFn, runTestServer, TestServerInfo } from './_util'
+import axios, { AxiosError } from 'axios'
+
+describe('server', () => {
+  let server: TestServerInfo
+  let close: CloseFn
+
+  beforeAll(async () => {
+    server = await runTestServer({
+      dbPostgresSchema: 'server',
+    })
+    close = server.close
+  })
+
+  afterAll(async () => {
+    await close()
+  })
+
+  it('preserves 404s.', async () => {
+    const promise = axios.get(`${server.url}/unknown`)
+    await expect(promise).rejects.toThrow('failed with status code 404')
+  })
+
+  it('turns unknown errors into 500s.', async () => {
+    // Sneak a new, failing route in before the server's error handler
+    const errorHandler = server.app._router.stack.pop()
+    server.app.get('/oops', () => {
+      throw new Error('Oops!')
+    })
+    server.app._router.stack.push(errorHandler)
+
+    const promise = axios.get(`${server.url}/oops`)
+    await expect(promise).rejects.toThrow('failed with status code 500')
+    try {
+      await promise
+    } catch (err: unknown) {
+      const axiosError = err as AxiosError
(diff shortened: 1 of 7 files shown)


packages/pds/tests/server.test.ts
```diff
@@ -0,0 +1,44 @@
...
+    await close()
+  })
+
+  it('preserves 404s.', async () => {
+    const promise = axios.get(`${server.url}/unknown`)
+    await expect(promise).rejects.toThrow('failed with status code 404')
+  })
+
+  it('turns unknown errors into 500s.', async () => {
+    // Sneak a new, failing route in before the server's error handler
```
dholms, line 25: not quite sure im following this one 🧐

packages/plc/src/server/error.ts
```diff
@@ -1,22 +1,22 @@
-import { Request, Response, NextFunction } from 'express'
+import { ErrorRequestHandler } from 'express'
```
dholms, line 1: oh nice didnt know about this type

dholms (approved):
this looks good. thanks for tackling this 👌 

just the one question but that's more for me & doesn't need to hold this up

packages/pds/tests/server.test.ts
devinivy: If I add a route to `server.app` without this rigamarole, it will be placed as middlware _after_ the app's error handler.  so any errors from the route wouldn't land in the error handler.  I suppose I could just create a new express app here intead and test the error handler in isolation.  I'll do that!

PR #281, opened 2022-10-28
Add public url config to pds for setting did service endpoint

devinivy:
The PDS's `origin` is enough to locate the service relative to the machine/network it's running on, but for the purposes of setting service endpoints in a did doc, the PDS also needs to know its public URL if it has one.  This ensures the public URL makes its way into the did doc when present.


dholms (approved):
nice good catch 👌 

i should integrate this into https://github.com/bluesky-social/atproto/pull/278
in the places where i setup the server's DID

PR #282, opened 2022-10-28
Username validation

dholms:
Username validation:
- length >= 2 & <=20
- only lowercase letters, numbers & hyphens
- can’t start with `did:`
- must end in available domain

Accept usernames & emails in any case & then `toLower` before storing in DB


packages/pds/src/db/migrations/20221021T162202001Z-init.ts
```diff
@@ -43,23 +43,11 @@ export async function up(db: Kysely<unknown>, dialect: Dialect): Promise<void> {
   await db.schema
     .createTable(userTable)
     .addColumn('username', 'varchar', (col) => col.primaryKey())
-    .addColumn('email', 'varchar', (col) => col.notNull())
+    .addColumn('email', 'varchar', (col) => col.notNull().unique())
     .addColumn('password', 'varchar', (col) => col.notNull())
     .addColumn('lastSeenNotifs', 'varchar', (col) => col.notNull())
     .addColumn('createdAt', 'varchar', (col) => col.notNull())
     .execute()
-  await db.schema
```
dholms, line 51: im good to delete these right? or are they serving another purpose?

devinivy: To be honest part of me likes enforcing uniqueness this way, but I agree that these are redundant with using application code to enforce usernames and emails be stored in lowercase.  @pfrazee do you see any issue with "forgetting" the casing of usernames?

packages/pds/src/api/com/atproto/account.ts
```diff
@@ -162,3 +145,37 @@ export default function (server: Server) {
...
+  } else if (front.length > 20) {
+    throw new InvalidRequestError('Username too long', 'InvalidUsername')
+  }
+  const validChars = /^[a-zA-Z0-9-]*$/.test(front)
+  if (!validChars) {
+    throw new InvalidRequestError(
+      'Invalid characters in username',
+      'InvalidUsername',
+    )
+  }
```
devinivy, line 180: I wonder if it might be worth doing full domain validation here.  There are still plenty of ways to create invalid domains, including some esoteric ones such as placing a hyphen in the 3rd and 4th positions (conflicts with punycode) 🙃

devinivy (approved):
Lookin good! 👍 Any idea what might have caused the build to have that weird run where it didn't complete?

packages/pds/src/db/migrations/20221021T162202001Z-init.ts
dholms: im not opposed to leaving them in as an extra sanity check

devinivy: I'm not too hung-up on it, either way is cool!

packages/pds/src/api/com/atproto/account.ts
dholms: do you have a good canonical way of doing full domain validation? most of what im seeing online are esoteric regexes. which im not sure handle the esoteric cases

dholms:
hmmm i don't know... looks like a problem with the initial migration 🤔

packages/pds/src/api/com/atproto/account.ts
devinivy: Ah good question, I guess it is a little bit of a crapshoot.  Two that come to mind are these, though I see how we might not want to reach for them:
https://www.npmjs.com/package/is-valid-domain
https://github.com/hapijs/address/blob/d67f9ba02378155184b76ec34362a53410e92167/lib/domain.js

It's possible that this is more of a theoretical issue than a practical one at the moment.

dholms: yea i came across that first one & it's deprecated (altho paul is a contributor on it 👀)

im fine with pullin in this hapi lib. seems simple/straightforward enough & handles the strange cases

dholms:
okay...

that test issue was a dumb copy-paste error in the migration code 😛 

i also brought in a bunch of reserved usernames from a listing i found online that seemed sensible (we can always make that more lenient) along with some atp/bsky specific usernames

I also pulled in that hapi package which seems to be working quite nicely

devinivy (approved):
This feels real solid! 🪨

PR #286, opened 2022-10-31
Session refresh token for PDS

devinivy:
Here we implement refresh tokens for the PDS.  When the users obtains a session, they receive to tokens: a short-lived access token and a longer-lived refresh token.

The access token is not revocable but the refresh token is revocable, and the user may revoke their session using `com.atproto.deleteSession()` e.g. during logout.

When the access token expires, `com.atproto.refreshSession()` may be used to obtain a new session containing a fresh access token and refresh token.  The refresh token used to perform the session refresh is revoked.

The overall effect is that users can stay logged-in indefinitely without providing their username and password so long as they have some activity within the window of the long-lived duration of a refresh token.

### DIFF
--- a/packages/api/src/schemas.ts
+++ b/packages/api/src/schemas.ts
@@ -39,9 +39,12 @@ export const methodSchemaDict: Record<string, MethodSchema> = {
       encoding: 'application/json',
       schema: {
         type: 'object',
-        required: ['jwt', 'username', 'did'],
+        required: ['accessJwt', 'refreshJwt', 'username', 'did'],
         properties: {
-          jwt: {
+          accessJwt: {
+            type: 'string',
+          },
+          refreshJwt: {
             type: 'string',
           },
           username: {
@@ -128,9 +131,12 @@ export const methodSchemaDict: Record<string, MethodSchema> = {
       encoding: 'application/json',
       schema: {
         type: 'object',
-        required: ['jwt', 'name', 'did'],
+        required: ['accessJwt', 'refreshJwt', 'name', 'did'],
         properties: {
-          jwt: {
+          accessJwt: {
+            type: 'string',
+          },
+          refreshJwt: {
             type: 'string',
           },
           name: {
@@ -168,15 +174,8 @@ export const methodSchemaDict: Record<string, MethodSchema> = {
     id: 'com.atproto.deleteSession',
     type: 'procedure',
     description: 'Delete the current session.',
-    parameters: {},
-    input: {
-      encoding: '',
-      schema: {
-        $defs: {},
-      },
-    },
     output: {
-      encoding: '',
+      encoding: 'application/json',
       schema: {
         $defs: {},
       },
@@ -240,6 +239,34 @@ export const methodSchemaDict: Record<string, MethodSchema> = {
         type: 'object',
         required: ['name', 'did'],
         properties: {
+          name: {
+            type: 'string',
+          },
+          did: {
+            type: 'string',
+          },
+        },
+        $defs: {},
+      },
+    },
+  },
+  'com.atproto.refreshSession': {
+    lexicon: 1,
+    id: 'com.atproto.refreshSession',
+    type: 'procedure',
+    description: 'Refresh an authentication session.',
+    output: {
+      encoding: 'application/json',
+      schema: {
+        type: 'object',
+        required: ['accessJwt', 'refreshJwt', 'name', 'did'],
+        properties: {
+          accessJwt: {
+            type: 'string',
+          },
+          refreshJwt: {
+            type: 'string',
+          },
           name: {
             type: 'string',
           },
@@ -2397,7 +2424,7 @@ export const methodSchemaDict: Record<string, MethodSchema> = {
             type: 'array',
             items: {
               type: 'object',
-              required: ['did', 'name', 'createdAt', 'indexedAt'],
+              required: ['did', 'name'],
               properties: {
                 did: {
                   type: 'string',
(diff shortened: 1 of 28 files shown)


packages/api/src/schemas.ts
```diff
@@ -2397,7 +2424,7 @@ export const methodSchemaDict: Record<string, MethodSchema> = {
             type: 'array',
             items: {
               type: 'object',
-              required: ['did', 'name', 'createdAt', 'indexedAt'],
+              required: ['did', 'name'],
```
devinivy, line 2427: Not sure the cause of this change— I feel like it may have been leftover from #278.

packages/pds/tests/auth.test.ts
```diff
@@ -0,0 +1,165 @@
...
+
+  it('provides valid access and refresh token on account creation.', async () => {
+    const account = await createAccount({
+      username: 'alice.test',
+      email: 'alice@test.com',
+      password: 'password',
+    })
+    // Valid access token
+    const sessionInfo = await getSession(account.accessJwt)
+    expect(sessionInfo).toEqual({ did: account.did, name: account.username }) // @TODO standardize on name or username?
```
devinivy, line 58: I think this called-out a little inconsistency in the API: should we consider unifying on `name` or `username`?

dholms: yeah I support 👍 

- "username" is the domain-style mutable id for a user. we should use this one
- "name" doesn't refer to anything in particular. could be displayName, username, pet name, first part of username before the host. confusing word that we should avoid in code
- "user" we've been using this to refer to `usernameOrDid`, but we may want a more descriptive term here cause it requires some context to understand

packages/api/src/schemas.ts
dholms: yup we get those values from the `profile` record which is not guaranteed to be there. so must be not required

packages/pds/src/auth.ts
```diff
@@ -13,6 +14,19 @@ export type ServerAuthOpts = {
   didResolver: DidResolver
   adminPass: string
 }
+
+export enum AuthScopes {
```
dholms, line 18: cool i like this 👍

packages/pds/src/api/com/atproto/session.ts
```diff
@@ -1,15 +1,12 @@
...
 
 export default function (server: Server) {
   server.com.atproto.getSession(async (_params, _input, req, res) => {
     const { db, auth } = locals.get(res)
-    const did = auth.getUserDid(req)
-    if (!did) {
-      throw new AuthRequiredError()
-    }
-
+    const did = auth.getUserDidOrThrow(req)
```
dholms, line 10: 👍

packages/pds/src/auth.ts
```diff
@@ -26,36 +42,56 @@ export class ServerAuth {
...
-  getUserDid(req: express.Request): string | null {
-    const header = req.headers.authorization || ''
-    if (!header.startsWith(BEARER)) return null
-    const token = header.slice(BEARER.length)
-    const payload = jwt.verify(token, this._secret)
+  createRefreshToken(did: string, expiresIn?: string | number) {
+    const payload = {
+      scope: AuthScopes.Refresh,
+      sub: did,
+      jti: uint8arrays.toString(crypto.randomBytes(32), 'base64'),
```
devinivy, line 63: @dholms do you feel this is appropriate for generating a unique id?  Typically I would just reach for a uuid, but I saw they aren't used anywhere else, so I reached for this instead.

packages/pds/src/api/com/atproto/session.ts
```diff
@@ -37,14 +34,97 @@ export default function (server: Server) {
...
+    const token = auth.getToken(req)
+    if (!token) {
+      throw new AuthRequiredError()
+    }
+    const lastRefreshId = auth.verifyToken(token, AuthScopes.Refresh).jti
+    if (!lastRefreshId) {
+      throw new Error('Unexpected missing refresh token id')
+    }
+
+    // If the token was already revoked, that's alright.
```
dholms: if it's alright if the token is already revoked, then should we be throwing if not found?

dholms: seems fine to make this idempotent & if the session is already deleted, just returning 200

dholms (approved):
sweet this all seems great to me 👌

packages/pds/src/auth.ts
dholms: yup this seems fine to me 👍 
32 random bytes is enough to not worry about collisions

packages/pds/src/api/com/atproto/session.ts
devinivy: In the latest commit I removed that erroneous comment (good catch!)  It was there because I originally wrote it to be idempotent, but then balked once I realized that I was already throwing during token verification when the token was expired.  I figured either both cases should throw, or both cases should not throw.

But I like that a lot more— I will take a pass at this to ignore both token expiration and revocation 👍

dholms: makes sense 👍

packages/pds/tests/auth.test.ts
AaronGoldman: Why username rather than handle?

@handle is clear signal that this is not meant as a real names policy and that a user could have many @handles that point to the same DID

dholms: actually I really like this. handle is an increasingly common term that shouldn't be too disorienting for users. & it is much more descriptive than "username"

dholms: updated 😁 
https://github.com/bluesky-social/atproto/issues/287

PR #288, opened 2022-11-01
Session management in PDS client

devinivy:
With refresh tokens implemented in #286, this adds a new service client that will automatically manage session credentials including the token refresh lifecycle.

Some specs:
 - On session refresh, session credentials are set.
 - A single concurrent session refresh can happen at a time.
 - On session deletion, session credentials are unset.
 - On session or account creation, session credentials are set.
 - When any other request takes place, if it fails due to an expired token then credentials are refreshed and the request is retried.
 - If a user sets auth headers directly during a request, then the request is passed through without any automatic session management.
 - Applications can hydrate existing credentials onto the service client using `client.sessionManager.set()`, and can subscribe to session changes using `client.sessionManager.on('session', ...)`.


pfrazee:
Great work on this. Very clean, and I'm glad you addressed potential races.

Feel free to move this into the xrpc package or leave it as is, per our slack discussion.

devinivy:
This is now all good for review!

@pfrazee I dug into the implementation being in api vs. xrpc, and ended-up leaving things as-is.  The `SessionClient` has the `app.bsky` and `com.atproto` schemas added to it, but the underlying `SessionServiceClient` does not, and could be used on its own.  I thought this was an acceptable tradeoff for now: if someone wanted to create a client with `com.atproto` methods for session management but not `app.bsky` methods, they could do so using `SessionServiceClient` 👍 I am sure we'll further refine this over time!

pfrazee:
Okay looks good to me!

packages/pds/src/api/app/bsky/getUsersTypeahead.ts
```diff
@@ -1,4 +1,4 @@
-import { QueryParams } from '@atproto/api/src/types/app/bsky/getUsersTypeahead'
+import { AppBskyGetUsersTypeahead as Method } from '@atproto/api'
```
dholms, line 1: we should import these from the server types (pds/src/lexicon) instead of the api package

packages/xrpc/src/client.ts
```diff
@@ -129,7 +133,7 @@ async function defaultFetchHandler(
   httpUri: string,
   httpMethod: string,
   httpHeaders: Headers,
-  httpReqBody: any,
+  httpReqBody: unknown,
```
dholms, line 136: bless you 🙏

dholms (approved):
yup this looks really great 👍

PR #289, opened 2022-11-01
Account declaration & invites

dholms:
This adds in 3 record types:
- `declaration`
- `invite`
- `inviteAccept`

It also updates the `follow` record to include both the subject did & a reference to the subject's declaration.

On account creation, it creates the account's declaration file & sets it to `user`

--- 
Next up (in order):
- removing badges
- adding in scene creation methods
- testing for groups & invites
- scene logic - for what gets reposted & getting that content into feeds


lexicons/bsky.app/declaration.json
```diff
@@ -0,0 +1,16 @@
+{
+  "lexicon": 1,
+  "id": "app.bsky.declaration",
+  "type": "record",
+  "key": "literal:self",
+  "record": {
+    "type": "object",
+    "required": ["actorType"],
+    "properties": {
+      "actorType": {
```
dholms, line 10: It feels a bit strange that this record is so sparse but I'm not sure if it's bad. Every account of the same type will have the same CID for their declaration.

Is there any other info we'd want to include in here?

One option would be to include `did` on the declaration as well & then a follow could be _just_ on the declaration CID. The tradeoff being that it makes follow records a decent bit more opaque which is why I didn't go with that option.

lexicons/bsky.app/invite.json
```diff
@@ -0,0 +1,22 @@
+{
+  "lexicon": 1,
+  "id": "app.bsky.invite",
+  "type": "record",
+  "key": "tid",
+  "record": {
+    "type": "object",
+    "required": ["group", "subject", "createdAt"],
+    "properties": {
+      "group": {"type": "string"},
```
dholms, line 10: Since the invite is created _by_ the group, we don't need to include the declaration CID as a check. 

However, you see that we do include the declaration CID on `inviteAccept`.

lexicons/bsky.app/declaration.json
pfrazee: Sparseness doesnt bother me. This is kind of a special record. If we think of other things that need this kind of behavior, we can add them.

We could include `did` but how would that change follow records? You'd still need the did of the repo in there, right?

devinivy: I personally like how you have it as-is: it feels symmetrical with how we treat uri/cid pairs on likes, reposts, etc.

lexicons/bsky.app/declaration.json
```diff
@@ -0,0 +1,16 @@
+{
+  "lexicon": 1,
+  "id": "app.bsky.declaration",
```
devinivy, line 3: I think that a `"description"` may be especially useful here for self-documentation, particularly because the term "declaration" is quite generic.  Alternately, we might consider a more verbose name like `accountDeclaration`.
```suggestion
  "id": "app.bsky.declaration",
  "description": "Context for an account that is considered intrinsic to it. A declaration should be treated as immutable.",
```

packages/pds/src/api/app/bsky/getUserFollows.ts
```diff
@@ -19,11 +19,11 @@ export default function (server: Server) {
       let followsReq = db.db
         .selectFrom('app_bsky_follow as follow')
         .where('follow.creator', '=', creator.did)
-        .innerJoin('user_did as subject', 'subject.did', 'follow.subject')
+        .innerJoin('user_did as subject', 'subject.did', 'follow.subjectDid')
```
devinivy, line 22: You are a saint for going through all of these.

devinivy (approved):
That came together quick, and is looking great to me 👍

lexicons/bsky.app/declaration.json
```diff
@@ -0,0 +1,16 @@
...
+  "type": "record",
+  "key": "literal:self",
```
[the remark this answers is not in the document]
dholms: yup pretty happy with this one 👌

lexicons/bsky.app/declaration.json
dholms: if you're aware of the declared repo, you actually _don't_ need the `did` to track the follow.

but I realized after posting this comment that you _do_ need the `did` to make the subscription to the correct repo & if you don't have it, then the CID is just this random hash you can't easily resolve

Cool i'll leave as is. It is a pretty special record so I think it's fine for it to be unique in a couple of ways (sparse & same CID for many repos)

packages/pds/src/api/app/bsky/getUserFollows.ts
dholms: actually this was another big win for kysely, I got compiler errors that told me exactly where these all were 😎

lexicons/bsky.app/declaration.json
dholms: good point 👍 
especially important on a schema like this, but as paul noted, we should be adding descriptions/comments to all of our schemas

devinivy (approved):

devinivy (approved):
D'oh, sorry I missed those!

dholms:
the problem with leaving testing for a later PR 😛

PR #291, opened 2022-11-01
Remove badges

dholms:


pfrazee:
oh no the badgers!

Just to reiterate to outsiders: this is also temporary. The idea isn't gone, we just want to avoid shipping code before it's in use.

devinivy (approved):
Looks right!

PR #292, opened 2022-11-02
Lexicon tokens

pfrazee:
- Adds lexicon tokens to the `lexicon` package
- Adds lexicon tokens to markdown and code generated by lex-cli
- Adds actorUser and actorScene schemas
- Does NOT add the resulting code generation to make merging easier

Adds lexicon "token" schema types, which are essentially declarations of the NSID (and some accompanying metadata) with no validation behaviors. They serve a purpose akin to exported constants. To wit, the code generated by tokens looks like this:

```typescript
export const APP_BSKY = {
  ActorScene: 'app.bsky.actorScene',
  ActorUser: 'app.bsky.actorUser',
}
```

The intent of Lexicon tokens is to be used as values in records and methods. For instance, the two tokens above are used in an upcoming "declaration record" type as intended values of the `actorType` field.

### DIFF
--- a/packages/lex-cli/src/codegen/server.ts
+++ b/packages/lex-cli/src/codegen/server.ts
@@ -1,10 +1,22 @@
-import { IndentationText, Project, SourceFile } from 'ts-morph'
+import {
+  IndentationText,
+  Project,
+  SourceFile,
+  VariableDeclarationKind,
+} from 'ts-morph'
 import { Schema, MethodSchema, RecordSchema } from '@atproto/lexicon'
 import { NSID } from '@atproto/nsid'
 import * as jsonSchemaToTs from 'json-schema-to-typescript'
 import { gen, schemasTs } from './common'
 import { GeneratedAPI } from '../types'
-import { schemasToNsidTree, NsidNS, toCamelCase, toTitleCase } from './util'
+import {
+  schemasToNsidTree,
+  NsidNS,
+  schemasToNsidTokens,
+  toCamelCase,
+  toTitleCase,
+  toScreamingSnakeCase,
+} from './util'
 
 export async function genServerApi(schemas: Schema[]): Promise<GeneratedAPI> {
   const project = new Project({
@@ -13,6 +25,7 @@ export async function genServerApi(schemas: Schema[]): Promise<GeneratedAPI> {
   })
   const api: GeneratedAPI = { files: [] }
   const nsidTree = schemasToNsidTree(schemas)
+  const nsidTokens = schemasToNsidTokens(schemas)
   for (const schema of schemas) {
     if (schema.type === 'query' || schema.type === 'procedure') {
       api.files.push(await methodSchemaTs(project, schema))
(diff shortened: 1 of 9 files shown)


pfrazee:
I originally tried to have the code generate an enum rather than a constant object, but ts-morph was fighting me (possibly a bug in their code generator?). I figured an object would work just as well.

packages/lex-cli/src/codegen/server.ts
```diff
@@ -60,6 +78,30 @@ const indexTs = (project: Project, schemas: Schema[], nsidTree: NsidNS[]) =>
         .setNamespaceImport(toTitleCase(schema.id))
     }
 
+    // generate token enums
+    for (const nsidAuthority in nsidTokens) {
+      // export const {THE_AUTHORITY} = {
+      //  {Name}: "{authority.the.name}"
+      // }
+      file.addVariableStatement({
```
devinivy, line 86: I believe the type of `{THE_AUTHORITY}` is probably going to be of type `{ {Name}: string }`.  If an `as const` is added to the end then it would take its precise type, with the exact key-value pairs.  Could then go a step further and get a union of all the values, which might be a useful type, as `typeof {AUTHORITY}[keyof typeof {AUTHORITY}]`.

devinivy (approved):
Looks great to me, not to mention super useful 🙏 💪

packages/lex-cli/src/mdgen/index.ts
```diff
@@ -48,7 +50,9 @@ export async function process(outFilePath: string, schemas: Schema[]) {
 async function genMdLines(schemas: Schema[]): Promise<StringTree> {
   let xprcMethods: StringTree = []
   let recordTypes: StringTree = []
+  let tokenTypes: StringTree = []
   for (const schema of schemas) {
+    console.log(schema.id)
```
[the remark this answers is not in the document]
pfrazee: Haha I _knew_ this would get flagged. I actually left it intentionally so the command tells you what schemas got ingested.

packages/lex-cli/src/codegen/server.ts
pfrazee: Good call, I'll add the `as const`. Can you show a specific example of the latter suggestion though? I get the gist of it but I'm unsure what exactly you're wanting to construct.

devinivy: Yeah, sure thing!  Consider this type:
```ts
const Animal = {
  Dog: 'app.bsky.animalDog',
  Cat: 'app.bsky.animalCat',
} as const
```

If someone wanted to accept one of the app.bsky nsids for an animal as an argument then they'd want the type `'app.bsky.animalDog' | 'app.bsky.animalCat'`, which could be constructed from `Animal` as:
```ts
const AnimalNsid = typeof Animal[keyof typeof Animal]
//               = 'app.bsky.animalDog' | 'app.bsky.animalCat'
```

packages/lex-cli/src/mdgen/index.ts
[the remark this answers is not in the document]
devinivy: Forgot I was in a CLI for a minute 🙃

packages/lex-cli/src/codegen/server.ts
pfrazee: Argh ts-morph is fighting me on adding `as const` too. How important is this?

packages/lex-cli/src/codegen/util.ts
```diff
@@ -47,3 +66,8 @@ export function toCamelCase(v: string): string {
   v = v.replace(/[.-]([a-z])/gi, (_, g) => g.toUpperCase()) // uppercase any dash or dot segments
   return v.replace(/[.-]/g, '') // remove lefover dashes or dots
 }
+
+export function toScreamingSnakeCase(v: string): string {
```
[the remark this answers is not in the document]
dholms: the inspiration for fission's best selling swag: https://fission.codes/blog/screaming-snake-case/

packages/lex-cli/src/codegen/server.ts
devinivy: I don't think it's too important!  Still should be quite useful without it 🙏

pfrazee: Oh yeah I see what you're saying. Yeah I think we'll just have to live without this for now.

packages/lex-cli/src/codegen/util.ts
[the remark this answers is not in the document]
pfrazee: To be fair, that is _extremely good_ swag

PR #293, opened 2022-11-02
Fix typo in PLC docs to point to correct key for recovery

j0ev:
Reading through this doc today and I think this a typo, or at least does not seem to follow from the preceding paragraphs.

### DIFF
--- a/packages/plc/README.md
+++ b/packages/plc/README.md
@@ -69,7 +69,7 @@ This is to be used in adversarial situations in which a user's `signingKey` leak
 
 In a situation such as this, the `recoveryKey` may be used to rotate both the `signingKey` and `recoveryKey`.
 
-If a user wishes to recovery from this situation, they sign a new operation rotating the `signingKey` to a key that they hold and set the `prev` of that operation to point to the most recent pre-attack operation.
+If a user wishes to recover from this situation, they sign a new operation rotating the `signingKey` to a key that they hold and set the `prev` of that operation to point to the most recent pre-attack operation.
 
 ## Example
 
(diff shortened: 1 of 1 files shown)


packages/plc/README.md
```diff
@@ -67,7 +67,7 @@ The PLC server provides a 72hr window during which the `recoveryKey` can "rewrit
 
 This is to be used in adversarial situations in which a user's `signingKey` leaks or is being held by some custodian who turns out to be a bad actor.
 
-In a situation such as this, the `signingKey` may be used to rotate both the `signingKey` and `recoveryKey`.
+In a situation such as this, the `recoveryKey` may be used to rotate both the `signingKey` and `recoveryKey`.
 
 If a user wishes to recovery from this situation, they sign a new operation rotating the `signingKey` to a key that they hold and set the `prev` of that operation to point to the most recent pre-attack operation.
```
AaronGoldman, line 72: recovery -> recover

May be worth noting that this is not considered duplicity on the part of the PLC server.

If the history "rewrite" set the `prev` to the attack operation this would not leave the PLC server in the position of having forked the history but the attack operation would live in the operations history forever. This is clearly a DOS risk if the attack operation was deliberately obnoxiously large.

By cutting out the attack operation we reduce the power of the attacker to affect the DID Doc history but we technically have forked the history and other implementations of the `did:plc:` client SHOULD NOT penalize the plc server as duplicitous locally or start gossiping about the duplicity to other Home Servers.

devinivy (approved):
Looks good, thanks!

PR #294, opened 2022-11-02
Fix app.bsky.post entity def to be more intuitive

devinivy:
Just addressing a little naming funk in the post entity.  In the docs you'll see this, where it's easy to miss that `Entity` actually represents an array of entities:
```ts
export type Entity = {
  index: TextSlice;
  type: string;
  value: string;
}[];
...
entities?: Entity;
```

With this update the types will look more like this:
```ts
export interface Entity {
  index: TextSlice;
  type: string;
  value: string;
};
...
entities?: Entity[];
```


pfrazee:
Excellent, yeah. I have no idea why I did it that way originally.

PR #300, opened 2022-11-02
Rename getHomeFeed to getTimeline to improve some semantic clarity

pfrazee:
- Renames `app.bsky.getHomeFeed` to `app.bsky.getTimeline`

We decided to do this to help distinguish between two separate concepts: timeline as the user's "consuming" view and feed as a user's "publishing" view.

Note that getTimeline's output object still uses the terms `feed` and `feedItem`. If we want to fix that, probably the right way would be to use a more generic term like `entries` or `items`.


dholms (approved):
schweet 👌

PR #301, opened 2022-11-03
Allow duplicate block by the same did w/ sql blockstore

devinivy:
Fixes a bug where the `SqlBlockstore` would hit a uniqueness error inserting the same block multiple times for the same did.  Also added a test for prevent regression.


packages/pds/tests/sql-blockstore.test.ts
```diff
@@ -0,0 +1,51 @@
+import SqlBlockstore from '../src/sql-blockstore'
+import { CloseFn, runTestServer, TestServerInfo } from './_util'
+import * as locals from '../src/locals'
+
+describe('server', () => {
```
devinivy, line 5:
```suggestion
describe('sql blockstore', () => {
```

dholms (approved):
excellent 👌

Issue #287, opened 2022-11-01
Standardize on "username" in schemas


dholms:
We should standardize on "username" when referring to a user's username & not use shortened ids like "name"

https://github.com/bluesky-social/atproto/pull/286#discussion_r1009670150

PR #303, opened 2022-11-03
Username -> Handle

dholms:
We decided to standardize on "handle" to describe the mutable domain-based identifier for users.

We prefer for a couple reasons:
- "username" is too vague. there are multiple types of names in the system: display names, possibly pet names or nick names
- handles identify "repos" which may or may not actually be users

I also updated the PLC data model to reflect this change along with the did-resolver library.

Closes https://github.com/bluesky-social/atproto/issues/287


pfrazee:
Thanks for hitting this. Tedious but worth it.

PR #304, opened 2022-11-03
Scene creation & membership

dholms:
- scene creation
- assertions & confirmations for scene creator & scene member
- auth for allowing the scene owner (currently just tracked in DB not repo) to make writes to the scene
- extends getProfile method to return profile info for both users & scenes


packages/handle/tests/handle.test.ts
```diff
@@ -17,7 +17,7 @@ describe('handle validation', () => {
     expect(check('john.bsky.io')).toThrow('Not a supported handle domain')
     expect(check('john.com')).toThrow('Not a supported handle domain')
     expect(check('j.test')).toThrow('Handle too short')
-    expect(check('jayromy-johnber123456.test')).toThrow('Handle too long')
+    expect(check('jaymome-johnber123456.test')).toThrow('Handle too long')
```
dholms, line 20: forgot i did this haha. canonical spelling 😛

packages/pds/src/api/app/bsky/actor/createScene.ts
```diff
@@ -0,0 +1,233 @@
...
+      const sceneCtx = repoUtil.mutationContext(dbTxn, did, now)
+      const sceneRepo = await RepoStructure.create(
+        sceneCtx.blockstore,
+        did,
+        sceneAuth,
+      )
+
+      const [sceneDeclaration, creatorAssert, memberAssert] = await Promise.all(
+        [
+          repoUtil.prepareCreate(
```
dholms, line 110: have some thoughts around some better abstractions & moving a lot of this into the repo lib

main goal being we want to prevent multiple roundtrips to the DB (which rn we're doing quite a few)

packages/pds/src/api/app/bsky/actor/getProfile.ts
```diff
@@ -18,38 +18,58 @@ export default function (server: Server) {
       const { ref } = db.db.dynamic
 
       const queryRes = await db.db
-        .selectFrom('user_did')
-        .where(userWhereClause(user))
+        .selectFrom('did_handle')
+        .where(actorWhereClause(actor))
+        .innerJoin(
```
dholms, line 23: inner join because we just want to ignore any repos that don't have a declaration

the thinking here is that a declaration should be required (in our ecosystem) & you're not a "valid `app.bsky` repo" without one

packages/uri/src/index.ts
```diff
@@ -35,6 +35,13 @@ export class AtUri {
     this.searchParams = parsed.searchParams
   }
 
+  static make(handleOrDid: string, collection?: string, rkey?: string) {
```
dholms, line 38: lil convenience method

packages/pds/tests/seeds/client.ts
```diff
@@ -197,7 +206,71 @@ export class SeedClient {
     return repost
   }
 
-  userRef(did: string): UserRef {
+  async createScene(creator: string, handle: string) {
+    const res = await this.client.app.bsky.actor.createScene(
+      {
+        handle,
+      },
+      {
```
dholms, line 214: would be nice to make the types a bit friendlier here, but that's for another day

pfrazee:
Looking great

packages/pds/tests/seeds/client.ts
devinivy: Just curious— where's the unfriendly type here?

packages/pds/tests/seeds/client.ts
```diff
@@ -197,7 +206,71 @@ export class SeedClient {
...
+        headers: this.getHeaders(creator),
+        encoding: 'application/json',
```
devinivy, line 216: Is it possible we can nix this `encoding`?  I think it's intended to be the default when there's an input object.

packages/pds/src/api/app/bsky/actor/getProfile.ts
devinivy: Makes sense.  Maybe some day we could offer a method to answer the question "does my repo look alright? if no, what's wrong?"

packages/pds/tests/seeds/client.ts
dholms: the same that you pointed it. it doesn't allow partial opts. so you can either provide _both_ headers & encoding or _neither_

packages/pds/tests/seeds/client.ts
dholms: merp https://github.com/bluesky-social/atproto/pull/304#discussion_r1015839741

packages/pds/tests/seeds/client.ts
devinivy: Ahh, got it!

lexicons/app/bsky/actor/createScene.json
```diff
@@ -0,0 +1,33 @@
...
+      "properties": {
+        "handle": { "type": "string" },
+        "did": { "type": "string" },
+        "declarationCid": { "type": "string" }
+      }
+    }
+  },
+  "errors": [
+    {"name": "HandleNotAvailable"}
+  ]
```
devinivy, line 32: Could add `InvalidHandle` here if we're feeling it, not sure that it's really critical though.

devinivy (approved):
I think you `did_handle` this very nicely 👍

packages/pds/src/api/app/bsky/actor/getProfile.ts
dholms: yeah that'd be good fer sure^^

lexicons/app/bsky/actor/createScene.json
dholms: yeah good catch 👍

packages/pds/src/api/app/bsky/actor/createScene.ts
```diff
@@ -0,0 +1,233 @@
...
+  server.app.bsky.actor.createScene(async (_params, input, req, res) => {
+    const { db, auth, config, keypair, logger } = locals.get(res)
+    const { recoveryKey } = input.body
+
+    const requester = auth.getUserDid(req)
+    if (!requester) {
+      throw new AuthRequiredError()
+    }
+
+    const handle = input.body.handle.toLowerCase()
```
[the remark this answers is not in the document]
dholms: hmmm right. it's out here because it needs to be normalized before running through validation logic & we need to return the normalized version

It could be normalized in `ensureValid` but that's kinda overloading it & I feel like it'd be too easy to drop the normalized version. 

Maybe a new method `normalizeAndEnsureValid` 🤔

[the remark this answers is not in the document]
devinivy: At the end of the day it's no biggie :) Main goal in my mind is to couple things closely enough to ensure no uppers make their way into the db, since it's an assumption we make about how the handles are stored.  One easy solution to that which wouldn't involve moving things around is to add an assertion inside `preregisterDid()`.

PR #305, opened 2022-11-03
Handle package

dholms:
Broke some of the logic for handle validation out into it's own library (similar to nsid & uri pacakges)


devinivy (approved):
There it is!

packages/pds/src/api/com/atproto/account.ts
```diff
@@ -35,7 +35,7 @@ export default function (server: Server) {
     const email = input.body.email.toLowerCase()
 
     // throws if not
-    ensureValidHandle(handle, config.availableUserDomains)
+    handleLib.ensureValid(handle, config.availableUserDomains)
```
devinivy, line 38: Ah, looking at the test failures, I suppose it's going to want to end-up looking something like this:
```suggestion
    try {
      handleLib.ensureValid(handle, config.availableUserDomains)
    } catch (err) {
      if (err instanceof InvalidHandleError) {
        throw InvalidRequestError(err.message, 'InvalidHandle')
      }
      throw err
    }
```

dholms: ugh yup

PR #308, opened 2022-11-04
Multibase tag for the did:plc hash

AaronGoldman:
https://github.com/multiformats/multibase/blob/master/multibase.csv

| encoding | code | description                           |
|-|-|-|
| base32    | b    | rfc4648 case-insensitive - no padding |

To be able to change the base in the future we may wish to tag with the multibase prefix.

also consider multihash prefix 0x1218 to be able to update the hash algorithm and prefix length 
```TS
// https://github.com/multiformats/multihash
// 0x12 SHA-256, 24 byte prefix length
prefix = new Uint8Array([0x12, 24])
const hashB32 = uint8arrays.toString( prefix + hashOfGenesis, 'base32')
```
0x12 SHA-256
0x18 prefix length 24 bytes
https://github.com/multiformats/multihash

### DIFF
--- a/packages/plc/src/lib/document.ts
+++ b/packages/plc/src/lib/document.ts
@@ -129,7 +129,10 @@ export const hashAndFindDid = async (op: t.CreateOp, truncate = 24) => {
   const hashOfGenesis = await crypto.sha256(cbor.encode(op))
   const hashB32 = uint8arrays.toString(hashOfGenesis, 'base32')
   const truncated = hashB32.slice(0, truncate)
-  return `did:plc:${truncated}`
+  // https://github.com/multiformats/multibase/blob/master/multibase.csv
+  // | encoding | code | description                           |
+  // | base32   | b    | rfc4648 case-insensitive - no padding |
+  return `did:plc:b${truncated}`
 }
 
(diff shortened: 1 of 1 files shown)


AaronGoldman:
Not intended as allowing any other hash algorithm or prefix length just providing future proofing if it is needed.
Adding options would still require a spec update just not ambiguity.

dmcorrales:
LGTM

dholms:
I've gone back & forth on this a couple times since reading it.

Generally very +1 on multiformats/multicodecs. But think I wanna hold off on this because:
- PLC is gonna be pretty tightly specced so there isn't much need for self-describing data
- PLC _should_ be temporary. So ideally we deprecate the whole system before we have any concern about upgrading the hash algo
- This lengthens the DID identifier. 24 chars is a decent balance between "lengthy enough for the security we need" & "still kinda parseable for a dev looking at it or shareable by users in extreme circumstances". If we're adding an extra char, I think I'd opt for 5 more bits of hash strength than a multibase char

dholms:
Gonna close for now. but feel free to keep the discussion going & we can always reassess

PR #309, opened 2022-11-04
Healthchecks for PDS and PLC

devinivy:
Adding a service healthcheck endpoint to PDS and PLC, plus some miscellaneous related updates:
 - Add `GET /health` endpoint to plc and pds services, which can be used as a healthcheck.
 - Unify conventions in plc with the better-established ones from pds.
 - Remove necessity of choosing a port for the server using `get-port`: node http will happily do this itself :)
 - Add `version` config to pds and plc so that they can report their version.  In practice this will be set by the build/deploy process.


dholms (approved):
✨

pfrazee:
This LGTM though we should consider putting the route under XRPC? I'm concerned about colliding with routes that other server software (which wants to implement XRPC alongside normal http stuff) might have

dholms:
Hmmm actually would it be worth having a defined XRPC method around this? This could actually be a useful method to call on other servers you're interacting with to ensure they're online & check their software version.

Generally my thought is: if any actor you don't control may be calling this, make it an XRPC method

devinivy:
Sounds like we will consider adding a healthcheck xrpc method to the pds once we have server-to-server interactions with the pds.

PR #318, opened 2022-11-08
Add app.bsky.actor.getSuggestions

pfrazee:
- Adds `app.bsky.actor.getSuggestions` as a means to get suggested follows


lexicons/app/bsky/actor/getSuggestions.json
```diff
@@ -0,0 +1,44 @@
...
+        "cursor": {"type": "string"},
+        "actors": {
+          "type": "array",
+          "items": {
+            "type": "object",
+            "required": ["did", "handle", "actorType"],
+            "properties": {
+              "did": {"type": "string"},
+              "handle": {"type": "string"},
+              "actorType": {"type": "string"},
```
dholms, line 25: I did the whole enum thing in the `getProfile` view: https://github.com/bluesky-social/atproto/blob/main/lexicons/app/bsky/actor/getProfile.json#L16

I guess it doesn't really matter because it's all strings in the end 😛 but wondering if we should decide on one or the other.

Would be a useful place for schema fragments

dholms (approved):
looks good 💪

lexicons/app/bsky/actor/getSuggestions.json
pfrazee: Yeah that occurred to me as well. I decided we'll probably take another pass over schemas before long and to just do the simple thing for now.

PR #319, opened 2022-11-08
Move from likes to votes

### LINKED ISSUE #248: app.bsky.dislike
It seems the bluesky protocol is aiming to cover behaviors that can reasonably be presumed to exist in any social media app as they are currently designed. While not ubiquitous in the absolute sense, 'dislike' is common and useful.

This is a request to add `app.bsky.dislike` to the base bsky lexicon records library -- similar to bluesky's existing [`like` record type](https://atproto.com/lexicons/bsky-app#appbskylike).

This could also be added in related constructs. For example, [`AppBskyGetHomeFeedFeedItem`](https://atproto.com/lexicons/bsky-app#appbskygethomefeed) in the docs contains `likeCount: number;` -- so `dislikeCount: number;` could be added in parallel.

Happy to do a PR if so desired.

Thank you for your attention! 🙏

Edit: Or there could be an `upvote`/`downvote` mechanism alongside `like`.

devinivy:
Implements up/downvote functionality in lieu of likes. Will likely follow this up with some additional procedures for voting actions.  Resolves #248


packages/pds/src/api/app/bsky/feed/getVotes.ts
```diff
@@ -0,0 +1,80 @@
...
+import { InvalidRequestError } from '@atproto/xrpc-server'
+
+export default function (server: Server) {
+  server.app.bsky.feed.getVotes(
+    async (params: GetVotes.QueryParams, _input, _req, res) => {
+      const { uri, limit, before, cid, direction } = params
+      const { db } = locals.get(res)
+      const { ref } = db.db.dynamic
+
+      if (
```
dholms, line 14: does schema validation handle this?

packages/pds/src/db/records/vote.ts
```diff
@@ -89,21 +92,25 @@ const notifsForRecord = (
   if (!matchesSchema(obj)) {
     throw new Error(`Record does not match schema: ${type}`)
   }
+  if (obj.direction === 'down') {
+    // No notifications for downvotes
```
dholms, line 96: I actually wonder if we _should_ return downvote notifs from the server & leave it up to the client if they want to display them 🤔 

Gives the option to filter them out, aggregate them into a summary, or just ruin the user's day by informing them of every downvote 😛

packages/pds/tests/views/notifications.test.ts
```diff
@@ -36,7 +36,7 @@ describe('pds notification views', () => {
       { headers: sc.getHeaders(alice) },
     )
 
-    expect(notifCount.data.count).toBe(12)
+    expect(notifCount.data.count).toBe(10)
```
dholms, line 39: kinda a pain that these numbers keep changing. maybe we could have the seed client track them?

(not for this PR, just throwing out a thought)

dholms (approved):
couple small qs. but looking good 🚀

packages/pds/src/api/app/bsky/feed/getVotes.ts
dholms: kk sweet 👍

packages/pds/src/db/records/vote.ts
devinivy: @pfrazee any feels on this?  Do you want downvote notifications from the backend and allow clients to filter them as needed?  I think it could still mess with unread counts, and maybe create some new cases to handle around setting the latest unseen time.

PR #325, opened 2022-11-08
Implement `setVote()` method

devinivy:
Implements `app.bsky.feed.setVote()` so that we can transactionally switch between the three vote states that a user may have for a post: up, down, or none.  A switch from an upvote to a downvote, for example, requires deleting then adding a record in the user's repo.

I decided to mirror `post.myState` in the output from `setVote()`, so it looks something like `{ upvote?: uri, downvote?: uri }`.

### DIFF
--- a/packages/pds/src/api/app/bsky/feed/setVote.ts
+++ b/packages/pds/src/api/app/bsky/feed/setVote.ts
@@ -1,0 +1,87 @@
+import { RepoStructure } from '@atproto/repo'
+import { AtUri } from '@atproto/uri'
+import { InvalidRequestError } from '@atproto/xrpc-server'
+import * as schema from '../../../../lexicon/schemas'
+import { Server } from '../../../../lexicon'
+import * as locals from '../../../../locals'
+import * as repoUtil from '../../../../util/repo'
+import { TID } from '@atproto/common'
+
+export default function (server: Server) {
+  server.app.bsky.feed.setVote(async (_params, input, req, res) => {
+    const { subject, direction } = input.body
+    const { auth, db, logger } = locals.get(res)
+
+    const requester = auth.getUserDidOrThrow(req)
+    const authStore = await locals.getAuthstore(res, requester)
+
+    const created = await db.transaction(async (dbTxn) => {
+      const currRoot = await dbTxn.getRepoRoot(requester, true)
+      if (!currRoot) {
+        throw new InvalidRequestError(
+          `${requester} is not a registered repo on this server`,
+        )
+      }
+
+      const now = new Date().toISOString()
+      const ctx = repoUtil.mutationContext(dbTxn, requester, now)
+      const repo = await RepoStructure.load(ctx.blockstore, currRoot)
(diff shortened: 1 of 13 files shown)


pfrazee:
Wunderbar, danke sehen

packages/pds/src/api/app/bsky/feed/setVote.ts
```diff
@@ -0,0 +1,87 @@
...
+      const repo = await RepoStructure.load(ctx.blockstore, currRoot)
+
+      const existingVotes = await dbTxn.db
+        .selectFrom('vote')
+        .select('uri')
+        .where('creator', '=', requester)
+        .where('subject', '=', subject.uri)
+        .execute()
+
+      const deleteExistingVotes = existingVotes.map(({ uri }) =>
```
dholms, line 37: if the existing vote already matches the intended outcome, should we just make this a no-op?

dholms: I think i'm leaning towards yes on this 🤔 no sense doing multiple mutations & publishing events when you can just stay put

dholms (approved):
looks good 👍 nailed the repo mutation

just the idempotence question

packages/pds/src/api/app/bsky/feed/setVote.ts
```diff
@@ -0,0 +1,87 @@
...
+        .createCommit(authStore, async (prev, curr) => {
+          const success = await dbTxn.updateRepoRoot(requester, curr, prev, now)
+          if (!success) {
+            logger.error({ did: ctx.did, curr, prev }, 'repo update failed')
+            throw new Error('Could not update repo root')
+          }
+          return null
+        })
+
+      await Promise.resolve([
```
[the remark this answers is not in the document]
devinivy: Oh man, oops!  This is just mistake.
```suggestion
      await Promise.all([
```

The way it's written it would literally have no effect aside from waiting a microtick or two 🙈

packages/pds/src/api/app/bsky/feed/setVote.ts
devinivy: Good call, I'll handle that case 👍

packages/pds/src/api/app/bsky/feed/setVote.ts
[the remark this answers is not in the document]
dholms: sweet 👍 i'll let you do the commit so we dont get in eachotehrs way

packages/pds/src/api/app/bsky/feed/setVote.ts
devinivy: @dholms should be all set now.  Added a test to ensure idempotence in that case 👍 Going to merge once I see CI turn green, but still open to making changes if you end-up reviewing the update.

dholms: yup looks good 👍

PR #327, opened 2022-11-08
Scene trending

dholms:
This adds a message queue abstraction to the database for handling side effects of created records.

It is currently used for 
- notifications
- scene processing

For scene processing, we maintain two tables: scene_member_count & scene_votes_on_post. And we publish a new trending record when a post has atleast 2 upvotes from a scene & the ratio of upvotes is >=20% of the scene


packages/pds/src/db/message-queue/index.ts
```diff
@@ -0,0 +1,276 @@
...
+  async catchup(): Promise<void> {
+    let builder = this.db.db
+      .selectFrom('message_queue')
+      .where('read', '=', 0)
+      .selectAll()
+    if (this.db.dialect !== 'sqlite') {
+      builder = builder.forUpdate()
+    }
+    const res = await builder.execute()
+    await Promise.all(res.map((row) => this.process(row.id)))
```
pfrazee, line 49: Is there any concern with races if we dont await each `process()` call individually?

packages/pds/src/db/message-queue/tables/messageQueue.ts
```diff
@@ -0,0 +1,12 @@
+import { Generated } from 'kysely'
+
+export interface MessageQueue {
+  id: Generated<number>
+  message: string
+  read: 0 | 1
```
pfrazee, line 6: This is small and completely fine w/you ignoring it, but a separate "MessageQueueCursors" table might be better for two reasons -- it allows multiple separate processors (may not need it but who knows) and it would eliminate N numbers for N events, saving ... some ... space on disk

packages/pds/src/db/message-queue/messages.ts
```diff
@@ -0,0 +1,97 @@
+export type AddMember = {
```
pfrazee, line 1: We should open an issue or something to track how we want to evolve the semantics of these events potentially into lexicon, but I'm +1 on not worrying about it for now

packages/pds/src/db/message-queue/index.ts
```diff
@@ -0,0 +1,276 @@
...
+import { sql } from 'kysely'
+import { dbLogger as log } from '../../logger'
+import { SceneVotesOnPost } from './tables/sceneVotesOnPost'
+import { AuthStore } from '@atproto/auth'
+import * as repoUtil from '../../util/repo'
+import * as schema from '../../lexicon/schemas'
+import { TID } from '@atproto/common'
+import { MessageQueue } from '../types'
+
+export class SqlMessageQueue implements MessageQueue {
```
pfrazee, line 21: I'm being a jerk by not checking myself, but just make sure you're using Kafka's semantics here, both for API and the expected fields. I'd suggest copying the exact fields of messages in Kafka, even if they don't apply to the table system, just so we don't get caught up with awkward abstractions over the differences in the future

packages/pds/src/db/message-queue/index.ts
```diff
@@ -0,0 +1,276 @@
...
+        'votes.count as voteCount',
+        'votes.postedTrending as postedTrending',
+      ])
+      .execute()
+
+    const now = new Date().toISOString()
+    await Promise.all(
+      state.map(async (scene) => {
+        if (scene.postedTrending) return
+        const ratio = scene.voteCount / scene.memberCount
```
pfrazee, line 226: Probably want to sanity check against a divide by zero here. Something is very wrong if that happens but you know

pfrazee (commented):
Yeah this is looking great. Don't let my comments come off as negativity, this is feeling very clean and getting solid value for the amount of effort.

packages/pds/src/db/message-queue/index.ts
dholms: idt we should rely on the application for serialization. this whole section should be considered async processing. I've tried to write the db transactions in such a way that they have the necessary locks in place to not race

packages/pds/src/db/message-queue/messages.ts
dholms: 👍

packages/pds/src/db/message-queue/index.ts
dholms: good catch 👌

packages/pds/src/db/message-queue/tables/messageQueue.ts
dholms: this should still allow for multiple separate processors. we lock rows with `SELECT FOR UPDATE` that keep separate processors from stepping on each other's toes

pfrazee: I'm thinking 2 processors that aren't coordinated with each other, eg maintaining their own cursors. But it's not a real concern tbh

packages/pds/tests/seeds/basic.ts
```diff
@@ -42,6 +84,9 @@ export default async (sc: SeedClient) => {
   await sc.vote('down', carol, sc.posts[alice][1].ref)
   await sc.vote('up', carol, sc.posts[alice][2].ref)
   await sc.vote('up', dan, sc.posts[alice][1].ref)
+  await sc.vote('up', alice, sc.posts[carol][0].ref)
+  await sc.vote('up', bob, sc.posts[carol][0].ref)
+  mq && (await mq.process())
```
dholms, line 89: here, we want to wait till the queue is done processing before moving on

this gives us a predictable order for our snapshots

packages/pds/src/db/message-queue/index.ts
```diff
@@ -0,0 +1,288 @@
...
+    await this.ensureCursor()
+    const keepGoing = await this.db.transaction(
+      async (dbTxn): Promise<boolean> => {
+        let builder = dbTxn.db
+          .selectFrom('message_queue_cursor as cursor')
+          .innerJoin('message_queue', 'message_queue.id', 'cursor.cursor')
+          .selectAll()
+        if (this.db.dialect !== 'sqlite') {
+          builder = builder.forUpdate()
+        }
```
devinivy: You could add a `skipLocked()` here so that if there's contention for the cursor, one consumer gets it and the other just bails rather than wait for it to open up.

packages/pds/src/db/message-queue/index.ts
```diff
@@ -0,0 +1,288 @@
...
+
+        const res = await builder.executeTakeFirst()
+        // all caught up
+        if (!res) return false
+
+        const message: Message = JSON.parse(res.message)
+        await this.handleMessage(dbTxn, message)
+        await dbTxn.db
+          .updateTable('message_queue_cursor')
+          .set({ cursor: sql`cursor + 1` })
```
devinivy: Only question I have about this is if we can assume the incrementing id is guaranteed to not leave any gaps.  Otherwise I think the cursor can get stuck.

packages/pds/src/db/message-queue/index.ts
```diff
@@ -0,0 +1,288 @@
...
+      case 'add_upvote':
+        return this.handleAddUpvote(db, message)
+      case 'remove_upvote':
+        return this.handleRemoveUpvote(db, message)
+      case 'create_notification':
+        return this.handleCreateNotification(db, message)
+      case 'delete_notifications':
+        return this.handleDeleteNotifications(db, message)
+    }
+  }
```
devinivy: Eventually I think it might be nice to colocate these handlers with related code (e.g.`handleAddMember()` living closer to member methods) and allow other areas of the app to hook-in to the message queue.  Thinking of the message queue as more of a pull-based transport, rather than a vertical slice of the app.  Zero problem with it as-is today, though, just thinkin!

devinivy (approved):
Had a few thoughts, but so many things I'm digging about this.  Real impressive 

Approving since I think it's essentially all set, but this comment could probably use attention: https://github.com/bluesky-social/atproto/pull/327/files/908c1e829a64f6a9e7372d4aed8da6eab22b1c6a#diff-4c1f0ad0cea9ddfe9a8ab78478a07d53a598be698252db06be359db1c77eda21

packages/pds/src/db/index.ts
```diff
@@ -58,7 +58,7 @@ export class Database {
       confirmation: Confirmation.makePlugin(db),
       profile: Profile.makePlugin(db),
     }
-    this.notifications = notificationPlugin(db)
+    // this.notifications = notificationPlugin(db)
```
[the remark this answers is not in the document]
dholms: good catch. nix it 👌

packages/pds/src/db/index.ts
```diff
@@ -353,10 +373,11 @@ export class Database {
       throw new Error('Expected indexed URI to contain a record key')
     }
     await this.db.insertInto('record').values(record).execute()
+
     const table = this.findTableForCollection(uri.collection)
-    await table.insert(uri, cid, obj, timestamp)
-    const notifs = table.notifsForRecord(uri, cid, obj)
-    await this.notifications.process(notifs)
+    const events = await table.insert(uri, cid, obj, timestamp)
+    this.messageQueue && (await this.messageQueue.send(this, events))
```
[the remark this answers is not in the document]
dholms: nice 👌

packages/pds/src/db/migrations/20221021T162202001Z-init.ts
```diff
@@ -249,9 +257,43 @@ export async function up(db: Kysely<unknown>, dialect: Dialect): Promise<void> {
...
+    .addColumn('did', 'varchar', (col) => col.primaryKey())
+    .addColumn('count', 'integer', (col) => col.notNull())
+    .execute()
+  await db.schema
+    .createTable(sceneVotesOnPostTable)
+    .addColumn('did', 'varchar', (col) => col.notNull())
+    .addColumn('subject', 'varchar', (col) => col.notNull())
+    .addColumn('count', 'integer', (col) => col.notNull())
+    .addColumn('postedTrending', 'int2', (col) => col.notNull())
+    .addPrimaryKeyConstraint('primary_key', ['did', 'subject'])
```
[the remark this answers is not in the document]
dholms: oh nice good looks. got lazy haha

packages/pds/src/db/message-queue/index.ts
dholms: i think we actually want to wait for it to open up. we basically ping the consumer on every event push to try & process exactly 1 event. So it should wait until the cursor is freed up & then give it a go

packages/pds/src/db/message-queue/index.ts
dholms: oh that's a good point. i _think_ so? worth a look 🧐

dholms: it doesn't guarantee it :notlikethis:

dholms: kk this should fix it
https://github.com/bluesky-social/atproto/pull/327/commits/e4464814e4f03b50ce7aec963b24ca45803f67cd

(also added an actual SQL LIMIT in the next commit)

packages/pds/src/db/message-queue/index.ts
devinivy: Ahh, okay got it.  How does `keepGoing` play into that?

packages/pds/src/db/message-queue/index.ts
dholms: yup makes sense 👍

packages/pds/src/index.ts
```diff
@@ -37,6 +38,11 @@ const runServer = (
     didResolver,
   })
 
+  const messageQueue = new SqlMessageQueue('pds', db, (did: string) => {
+    return auth.verifier.loadAuthStore(keypair, [], did)
+  })
+  db.setMessageQueue(messageQueue)
```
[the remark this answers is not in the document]
dholms: yeah both of these crossed my mind as well 🤔 

we may want something like `RecordIndexer`. This kinda has to do with that fact that the db has _a lot_ of functionality jammed into it. probably needs to be rethought & broken up a bit

packages/pds/src/db/message-queue/index.ts
dholms: ah not sure if you saw, but `keepGoing` actually got ripped out.

Yeah beforehand, i was throwing way to many handlers at this & they'd all just hang out at the lock with their hands open until the DB got back to them & said "sorry buddy no cursor for you"

packages/pds/src/db/message-queue/index.ts
```diff
@@ -0,0 +1,316 @@
...
+
+      const res = await builder.executeTakeFirst()
+      // all caught up
+      if (!res) return
+
+      const message: Message = JSON.parse(res.message)
+      await this.handleMessage(dbTxn, message)
+      await dbTxn.db
+        .updateTable('message_queue_cursor')
+        .set({ cursor: sql`cursor + 1` })
```
devinivy, line 109: I think by incrementing by one here we can cause messages to get processed multiple times if there's a gap in the ids.  Pretty sure we'll want to catch the cursor up to whichever message was just processed (plus 1).

packages/pds/src/db/message-queue/index.ts
devinivy: Ahhh right on!

packages/pds/src/db/message-queue/index.ts
dholms: a true hero you are 🙏

devinivy (approved):
💯

PR #332, opened 2022-11-09
Add 'member' to getProfile state output, and add some dev-env mock data

pfrazee:
- Adds `member` to the `myState` output of `getProfile`. The value is the URI of the confirmation record.
- Adds some scene mock data to the dev environment

### DIFF
--- a/packages/dev-env/src/mock/index.ts
+++ b/packages/dev-env/src/mock/index.ts
@@ -2,7 +2,7 @@ import { DevEnv } from '../index'
 import { ServerType } from '../types'
 import { genServerCfg } from '../util'
 import { AtUri } from '@atproto/uri'
-import { ServiceClient } from '@atproto/api'
+import { ServiceClient, APP_BSKY_GRAPH } from '@atproto/api'
 import { postTexts, replyTexts } from './data'
 
 // NOTE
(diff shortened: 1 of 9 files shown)


packages/dev-env/src/mock/index.ts
```diff
@@ -121,6 +121,65 @@ export async function generateMockSetup(env: DevEnv) {
...
+      {
+        assertion: APP_BSKY_GRAPH.AssertMember,
+        subject: {
+          did: bob.did,
+          declarationCid: bob.declarationCid,
+        },
+        createdAt: new Date().toISOString(),
+      },
+    )
+    bob.api.app.bsky.graph.confirmation.create(
```
devinivy, line 140: I think we want these `await`s, but feel free to ignore if it was intentional as-is.
```suggestion
    await bob.api.app.bsky.graph.confirmation.create(
```

packages/dev-env/src/mock/index.ts
```diff
@@ -121,6 +121,65 @@ export async function generateMockSetup(env: DevEnv) {
...
+      {
+        assertion: APP_BSKY_GRAPH.AssertMember,
+        subject: {
+          did: carla.did,
+          declarationCid: carla.declarationCid,
+        },
+        createdAt: new Date().toISOString(),
+      },
+    )
+    carla.api.app.bsky.graph.confirmation.create(
```
devinivy, line 167:
```suggestion
    await carla.api.app.bsky.graph.confirmation.create(
```

devinivy (approved):
Excellent!

pfrazee:
Got it, thanks for the tweaks. (Holy cow that's a useful github feature)

PR #335, opened 2022-11-09
Include scene trends in feeds

devinivy:
This adds trends into the author feed and timeline.  Those schemas were modified by adding `trendedBy` (a la `repostedBy`) to the feed item and `actorType` to all actor references within the feed items.


pfrazee:
lgtm

packages/pds/tests/seeds/client.ts
```diff
@@ -219,12 +233,14 @@ export class SeedClient {
         encoding: 'application/json',
       },
     )
+    this.dids[handle] = res.data.did
```
devinivy, line 236:
```suggestion
```

packages/pds/tests/views/notifications.test.ts
```diff
@@ -70,13 +70,16 @@ describe('pds notification views', () => {
   })
 
   it('paginates', async () => {
+    // @TODO there is a small bug where notifications generated with the same
```
dholms, line 73: yeah :/ this is actually an issue with all our pagination i believe :/

dholms (approved):
awesome thanks for tackling this 🙌

packages/pds/tests/views/notifications.test.ts
devinivy: I think you're right.  The exception is the actor search, whose cursor is a pair `(score, did)`.

PR #336, opened 2022-11-10
Fix codegen signature for the record delete methods

pfrazee:
Fixes a minor bug in our generated code for the API client


devinivy (approved):
Nice catch 🙏

PR #340, opened 2022-11-11
Add app.bsky.graph.getAssertions()

pfrazee:
- Adds `app.bsky.graph.getAssertions()`
  - Can select by author and/or subject
  - Can filter by confirmation status and assertion type

I need this to check on pending invites. Built it to be fairly generic.


dholms (approved):
lookin solid to me 👌

PR #341, opened 2022-11-11
Allow updating scene profiles

dholms:
Handle authorization for `updateProfile` similarly to CRUD: request can come from the account itself or from an owning account


packages/api/src/client/index.ts
```diff
@@ -1141,7 +1141,7 @@ export class AssertionRecord {
   }
 
   async delete(
-    params: Omit<ComAtprotoRepoDeleteRecord.InputSchema, 'collection'>,
+    params: Omit<ComAtprotoRepoDeleteRecord.QueryParams, 'collection'>,
```
devinivy, line 1144: I think you may need to rebuild lex-cli to grab Paul's recent fix.  I _think_ the way this was written is correct.

devinivy (approved):
That's it!

PR #342, opened 2022-11-11
Fix test snapshots post-merge

devinivy:
Just reconciling some test snapshots from #340 and #341.


dholms (approved):
yup 👍

devinivy:
Tests are intermittently failing due to pagination in the pds assertion views.  It's the same pagination we see in a few places, where items paginated by timestamp that occur at the same time may not appear in a well-defined order.  So I'm merging this as an incremental improvement :)

PR #349, opened 2022-11-14
Give 'HandleNotAvailable' error on reserved handles

pfrazee:
- Updates handles package to include a reserved handle error
- Updates PDS to respond with HandleNotAvailable with this kind of error

Need this to give better error feedback in the client

### DIFF
--- a/packages/handle/src/index.ts
+++ b/packages/handle/src/index.ts
@@ -24,7 +24,7 @@ export const ensureValid = (
   }
 
   if (reservedSubdomains[front]) {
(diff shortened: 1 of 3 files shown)


devinivy (approved):
Looks good!  We might also consider adding this to the list of documented error types in `lexicons/app/bsky/actor/createScene.json` and `lexicons/com/atproto/account/create.json`.

dholms (approved):
👍

pfrazee:
I think the "not available" error works fine here, but we do want to duplicate the error codes into the createScene method at some point

Issue #344, opened 2022-11-11
Ambiguous timestamp pagination in pds


devinivy:
There are a few views that paginate by timestamp, causing intermittent failing tests.  The issue is that items with the same timestamp may not appear in a consistent order.

This can be fixed by ordering by some identifier after ordering by timestamp, similar to the actor search views: https://github.com/bluesky-social/atproto/blob/0d98291c42dbbd77482b71e7aca47cc3af4ea463/packages/pds/src/api/app/bsky/util/search.ts#L161-L165

Referenced in a `@TODO` here: https://github.com/bluesky-social/atproto/blob/5b4e2b9ee33e43fbbe72a80e97d18310fd9fd248/packages/pds/tests/views/notifications.test.ts#L73-L75

dholms:
Fixed by: https://github.com/bluesky-social/atproto/pull/365

PR #351, opened 2022-11-14
Stagable blockstore

dholms:
This cleans up a few things surrounding repo mutations:

### Stageable blockstore
The blockstore now "stages" new blocks instead of "putting" them. this means holding them in memory until it is "saved" which them persists them to the actual blockstore.
This allows us to push all the new blocks to the DB in one operation instead of multiple round trips.
It also lets us sync a repo without permanently storing any new blocks until we verify that the repo updates are valid.

### Remove mutable repo
We have an immutable `RepoStructure` class and a mutable `Repo` class. I removed the latter & renamed `RepoStructure -> Repo`. We weren't using `Repo`, it was extra code & I think it encourages bad practices.
  
### Split out repo mutation logic into "utils/repo"
We had a lot of replicated logic around repo mutations. This logic has been split out into utils/repo.
Each mutation involves three steps now:
- preparing the updates: verifying them, setting $type to collection name, determining uri & cid
- writing to repo
- passing along to indexed

### Remove LevelDB (mostly)
Removed `level` as a dependency for repo & PDS. 

It is still a dependency for the did:web implementation in did-resolver. I think we should remove this for the time being, and can do so in a follow up PR. If we add it back in at some point, it should live in `dev-env`.


packages/pds/src/api/app/bsky/graph/getAssertions.ts
```diff
@@ -96,6 +96,7 @@ export default function (server: Server) {
         limit,
         before,
         by: ref('assertion.createdAt'),
+        secondaryOrder: ref('assertion.uri'),
```
dholms, line 99: fix for ordering issue for now. need a longer term fix that takes it into account w cursor: https://github.com/bluesky-social/atproto/issues/344

packages/repo/src/repo.ts
```diff
@@ -1,240 +1,311 @@
...
+    this.cid = params.cid
+    this.stagedWrites = params.stagedWrites
   }
 
   static async create(
     blockstore: IpldStore,
     did: string,
     authStore: auth.AuthStore,
-    verifier?: auth.Verifier,
+    initialRecords: RecordCreateOp[] = [],
```
dholms, line 53: can have records in first commit now

packages/pds/src/util/repo.ts
```diff
@@ -1,76 +1,179 @@
...
+    )
   }
+  const writeOps = writes.map((write) => write.op)
+  const repo = await Repo.load(blockstore, currRoot)
+  await repo
+    .stageUpdate(writeOps)
+    .createCommit(authStore, async (prev, curr) => {
+      const success = await dbTxn.updateRepoRoot(did, curr, prev, now)
+      if (!success) {
+        throw new Error('Repo root update failed, could not linearize')
```
dholms, line 59: This shouldn't ever be reached because of the SELECT FOR UPDATE lock. But we perform the check & throw an internal server error just in case

packages/pds/src/sql-blockstore.ts
```diff
@@ -21,36 +27,58 @@ export class SqlBlockstore extends IpldStore {
...
 
-  async putBytes(cid: CID, bytes: Uint8Array): Promise<void> {
+  async saveStaged(): Promise<void> {
     this.db.assertTransaction()
-    const insertBlock = this.db.db
-      .insertInto('ipld_block')
-      .values({
+    const blocks: IpldBlock[] = []
+    const creators: IpldBlockCreator[] = []
+    // const promises: Promise<any>[] = []
```
devinivy, line 49:
```suggestion
```

packages/pds/src/sql-blockstore.ts
```diff
@@ -21,36 +27,58 @@ export class SqlBlockstore extends IpldStore {
...
+        did: this.did,
+      })
+    }
+    const insertBlocks = this.db.db
+      .insertInto('ipld_block')
+      .values(blocks)
       .onConflict((oc) => oc.doNothing())
       .execute()
-    const insertBlockOwner = this.db.db
+    const insertCreators = this.db.db
```
devinivy, line 68: Feels good to see creators tracked only for inserts to the blockstore that are likely to live under a real commit for them.

packages/pds/src/sql-blockstore.ts
```diff
@@ -1,17 +1,23 @@
 import { IpldStore } from '@atproto/repo'
 import { CID } from 'multiformats/cid'
 import Database from './db'
+import { IpldBlock } from './db/tables/ipld-block'
+import { IpldBlockCreator } from './db/tables/ipld-block-creator'
 
 export class SqlBlockstore extends IpldStore {
+  staged: Map<string, Uint8Array>
```
devinivy, line 8: Would be really neat if multiformats/cid gave `CID`s reference equality, and you could meaningfully write `Map<CID, Uint8Array>`.

packages/repo/src/repo.ts
devinivy: 🤤 👍

packages/repo/tests/_util.ts
```diff
@@ -79,70 +81,102 @@ export type RepoData = Record<string, CollectionData>
...
-): Promise<RepoData> => {
+): Promise<{ repo: Repo; data: RepoData }> => {
   const { adds = 0, updates = 0, deletes = 0 } = params
   const repoData: RepoData = {}
+  const writes: RecordWriteOp[] = []
   for (const collName of testCollections) {
     const collData = prevData[collName]
     const shuffled = shuffle(Object.entries(collData))
-    const coll = await repo.getCollection(collName)
+    // const coll = await repo.getCollection(collName)
```
devinivy, line 127:
```suggestion
```

packages/repo/tests/_util.ts
```diff
@@ -79,70 +81,102 @@ export type RepoData = Record<string, CollectionData>
...
 
     for (let i = 0; i < adds; i++) {
       const object = generateObject()
-      const { rkey } = await coll.createRecord(object)
+      const rkey = TID.nextStr()
+      // const { rkey } = await coll.createRecord(object)
```
devinivy, line 132:
```suggestion
```

packages/repo/src/mst/mst.ts
```diff
@@ -180,16 +180,16 @@ export class MST implements DataStore {
   // -------------------
 
   // Persist the MST to the blockstore
-  async save(): Promise<CID> {
+  async stage(): Promise<CID> {
```
devinivy, line 183: What's the right way to think about this name change in the MST?  Is the comment above "persist the MST..." accurate?

Is the idea is that the MST itself doesn't really have the ability to commit, since that will happen in coordination with the repo?

packages/repo/src/blockstore/memory-blockstore.ts
```diff
@@ -2,43 +2,63 @@ import { CID } from 'multiformats/cid'
...
+    if (fromBlocks) return fromBlocks
+    throw new Error(`Not found: ${cid.toString()}`)
   }
 
-  async putBytes(k: CID, v: Uint8Array): Promise<void> {
-    this.map.set(k.toString(), v)
+  // async putBytes(k: CID, v: Uint8Array): Promise<void> {
+  //   this.map.set(k.toString(), v)
+  // }
+
```
devinivy, line 25:
```suggestion
```

packages/repo/src/blockstore/ipld-store.ts
```diff
@@ -10,12 +10,14 @@ import { CarReader } from '@ipld/car/reader'
 export abstract class IpldStore {
   abstract has(cid: CID): Promise<boolean>
   abstract getBytes(cid: CID): Promise<Uint8Array>
-  abstract putBytes(cid: CID, bytes: Uint8Array): Promise<void>
+  abstract stageBytes(cid: CID, bytes: Uint8Array): Promise<void>
```
devinivy, line 13: More of a "nice to have", but I do wonder if this interface could be pared-back even more for implementers.  For example, `IpldStore` could implement at least a default `stageBytes()` and `clearStaged()` using a `Map`.  It could also coordinate the stage lookups implementers need to do in `getBytes()`, and let implementers just worry about persistence.  I don't think this is necessarily worth digging into right now, but wanted to share the thought!

packages/repo/tests/repo.test.ts
```diff
@@ -26,34 +28,60 @@ describe('Repo', () => {
...
-    const updatedObj = util.generateObject()
-    await collection.updateRecord(rkey, updatedObj)
-    got = await collection.getRecord(rkey)
-    expect(got).toEqual(updatedObj)
-
-    await collection.deleteRecord(rkey)
-    got = await collection.getRecord(rkey)
+    // const collection
+    // const collection = repo.getCollection('com.example.posts')
+
```
devinivy, line 33:
```suggestion
```

devinivy (approved):
Love the direction this has been heading.  Looking forward to working with this in the pds!

packages/pds/src/api/com/atproto/repo.ts
```diff
@@ -205,52 +182,24 @@ export default function (server: Server) {
...
-        logger.warn(
-          { uri: uri.toString(), err, validate },
-          'failed to index new record',
-        )
-        if (validate) {
-          throw new InvalidRequestError(`Could not index record: ${err}`)
-        }
-      }
+    const now = new Date().toISOString()
+    const write = await repoUtil.prepareCreate(did, {
```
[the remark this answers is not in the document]
dholms: aw yess 👌

packages/pds/src/sql-blockstore.ts
dholms: woops good find 🔪

packages/pds/src/sql-blockstore.ts
dholms: yeah i've had a rough time with CID equality in the past :/ you can even see with CidSet that I have to convert them to strings to get them deduped.

packages/repo/src/blockstore/memory-blockstore.ts
dholms: lol really did a bad job cleaning up comments on this one 😅 😅

packages/repo/src/mst/mst.ts
dholms: Yup exactly. There are actually a couple stages to the block saving process here:
- node in MST is edited: we don't hash anything, just maintain in-memory references (this actually gives us ~100x speedup)
- MST is `staged`: we're reasonably confident this is the final state of the MST after some set of mutations. Traverse it & stage any changed nodes to the blockstore
- blockstore is saved: a valid commit was produced so the MST nodes are now canonical & we can save them to the permanent store

packages/repo/src/blockstore/ipld-store.ts
dholms: yeah i actually thought about this too 🤔 

eh i'll just throw it together real quick, should be fast

packages/repo/src/mst/mst.ts
devinivy: Cool cool that all tracks, thanks!

PR #352, opened 2022-11-15
Support shallow builds and sourcemaps

devinivy:
I took a pass over all the packages and ensure they support "shallow" builds, meaning that they do not bundle their dependencies.  This will be important if/when any of these packages are published, and is also important for being able to instrument the pds and plc.  I also started generating sourcemaps with the builds.

Part of supporting a shallow build involves setting package.json `main` to the `dist/` rather than the typescript files.  I support this as a flag to our build scripts, `--update-main-to-dist`.  The `--update-main-to-dist` is performed separately from the build because when you build all the packages together, they rely on `main` pointing to the typescript file in order to pass the `tsc` check.

Going to be easiest to peek at this one [ignoring whitespace](https://github.com/bluesky-social/atproto/pull/352/files?w=1).


packages/api/package.json
```diff
@@ -1,11 +1,12 @@
 {
   "name": "@atproto/api",
   "version": "0.0.1",
-  "main": "src/index.ts",
+  "main": "./dist/index.js",
```
devinivy, line 4:
```suggestion
  "main": "src/index.ts",
```

packages/api/package.json
```diff
@@ -15,6 +16,8 @@
   },
   "devDependencies": {
     "@atproto/lex-cli": "*",
-    "@atproto/pds": "*"
+    "@atproto/pds": "*",
+    "@npmcli/package-json": "^3.0.0",
```
dholms, line 20: don't hate me for this, but what do you think of putting these in the monorepo root instead of each individual pacakge

dholms: part of me also just wants to move the build script to the monorepo root as well & reuse it 🤔

packages/api/build.js
```diff
@@ -0,0 +1,22 @@
+const pkgJson = require('@npmcli/package-json')
+const { nodeExternalsPlugin } = require('esbuild-node-externals')
+
+const buildShallow =
+  process.argv.includes('--shallow') || process.env.ATP_BUILD_SHALLOW === 'true'
+
+if (process.argv.includes('--update-main-to-dist')) {
```
dholms, line 7: this works for now. but we'll probably want a build + publish script that does this, publishes & then reverts. or some fancier copy files around stuff

devinivy: Totally agree.  For now this is little more than a stepping stone to unravel some things and enable the app to be instrumented.

packages/api/package.json
devinivy: Might be easier to chat about this, but main thoughts I'm having are:
 - I lean towards being honest with dependencies, and declaring them wherever they are needed.  So if the build script lives in `api/` and needs these deps, I would lean towards keeping them here.  But if the build script using these deps lives elsewhere, I'd tend to move them there.
 - Yarn is deduping and hoisting these dependencies for us, so we shouldn't have dupes lying around.
 - Maybe a good middle-ground would be to make a package for build tooling.  That would give us a chance to centralize the deps and logic, and allow us to make exceptions for packages that need some special build treatment.
 - I just realized esbuild is installed in the root, so I created a little inconsistency here at best.

dholms: alright cool that makes sense 👍 
Yeah my general rule of thumb was to keep honest about dependencies but group most dev dependencies in the monorepo root. Makes it easier to track & switch out without leaving cruft hanging around.
I dig the reasoning tho. I like the idea of splitting out dev tooling into it's own package eventually though. Fine with it as it for now & we can always readdress at some point

dholms (approved):
🔨 🏗️

packages/api/package.json
devinivy: That makes sense.  It sounds like a decent way to force you to keep all the packages in sync with each other 👍

PR #353, opened 2022-11-16
Use public hostname in invite codes

devinivy:
A fix to make invite codes a little more secure and a little prettier.


dholms (approved):
nice 👌

PR #357, opened 2022-11-17
Remove Level

dholms:
Remove all references to level

The big one left was the did-web server in the did-resolution module. I stripped out the persistent option for that and moved it to tests. Can maybe find a better home in dev-env for something eventually


devinivy (approved):
Yep!

Issue #350, opened 2022-11-14
Incorrect notifications


dholms:
From Paul:
>  in the dev env I toggled following somebody and afterwards got a notification about them upvoting a post? Not sure why that would happen

PR #358, opened 2022-11-17
MessageQueue fix

dholms:
Message queue was getting behind because for each batch of messages added (could be 1 or multiple) we were only prompting the receiver to process 1.

This fixes that & prompts `processNext` for the number of messages sent.

We also add in a sanity check interval where we process any unprocessed messages every 3 min.

Closes https://github.com/bluesky-social/atproto/issues/350

### DIFF
--- a/packages/pds/src/db/message-queue/index.ts
+++ b/packages/pds/src/db/message-queue/index.ts
@@ -25,7 +25,11 @@ export class SqlMessageQueue implements MessageQueue {
     private name: string,
     private db: Database,
     private getAuthStore: GetAuthStoreFn,
-  ) {}
+  ) {
+    this.ensureCaughtUp().catch((err) =>
+      log.error({ err }, 'error catching queue up on startup'),
+    )
+  }
 
   async send(tx: Database, messages: Message | Message[]): Promise<void> {
     const msgArray = Array.isArray(messages) ? messages : [messages]
@@ -37,8 +41,11 @@ export class SqlMessageQueue implements MessageQueue {
     }))
 
     await tx.db.insertInto('message_queue').values(values).execute()
-
-    this.processNext()
+    for (let i = 0; i < msgArray.length; i++) {
+      this.processNext().catch((err) => {
+        log.error({ err }, 'error processing message')
+      })
+    }
   }
 
   private async ensureCursor(): Promise<void> {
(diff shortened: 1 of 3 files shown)


packages/pds/src/db/message-queue/index.ts
```diff
@@ -51,6 +56,15 @@ export class SqlMessageQueue implements MessageQueue {
     this.cursorExists = true
   }
 
+  async ensureCaughtUp(): Promise<void> {
+    try {
+      await this.processAll()
+    } catch (err) {
+      log.error({ err }, 'error ensuring queue is up to date')
+    }
+    setTimeout(this.ensureCaughtUp, 60000) // 1 min
```
devinivy, line 65: I think that it _might_ need to be like this, to avoid `this` not being bound when the callback is called:
```suggestion
    setTimeout(() => this.ensureCaughtUp, 60000) // 1 min
```

packages/pds/src/db/message-queue/index.ts
```diff
@@ -51,6 +56,15 @@ export class SqlMessageQueue implements MessageQueue {
...
+  }
+
```
devinivy, line 67: Would be nice to have something like this so that we can teardown the message queue— otherwise processes running the queue wont naturally exit, and they'll need to be torn down with a bit of force.  We might have a couple similar cases like this dangling around, too.  (Pretty sure jest even notes them at the end of test runs.)
```suggestion
    this.ensureCaughtUpTimeout = setTimeout(this.ensureCaughtUp, 60000) // 1 min
  }

 destroy() {
   clearTimeout(this.ensureCaughtUpTimeout)
 }
```

packages/pds/src/db/message-queue/index.ts
dholms: ah shit you're totally right

packages/pds/src/db/message-queue/index.ts
```diff
@@ -37,8 +39,11 @@ export class SqlMessageQueue implements MessageQueue {
     }))
 
     await tx.db.insertInto('message_queue').values(values).execute()
-
-    this.processNext()
+    for (let i = 0; i < msgArray.length; i++) {
+      this.processNext().catch((err) => {
```
devinivy, line 43: Right now this initiates a handful of queries which will end-up pending.  Each one might take-up a connection from the connection pool while they wait.  It's a minor thing, but if we serialize those ourselves rather than leaning on the connection pool and db locks, it could be fruitful.  That might look like `this.processNext(n)`, and allowing `processNext()` to internally loop `n` times.

dholms: took this to chat, but just for posterity's sake: we're nearly always doing one event at a time here, max 2. so this isn't the big bottleneck. 

we'll keep an eye on the message queue tho once we get into connection pool scaling issues

devinivy (approved):

PR #362, opened 2022-11-18
Wait for servers to close at end of test suite

dholms:
The open handles warning we were getting was actually from the callback from shutting down the server. We wrap that in a promise to get everything shutdown nicely & throw an error if encountered.

Also snuck in a chance to just tail the logs (defaults to 50 logs) instead of piping it all through pino-prettier


devinivy (approved):
Delightful 💞

Issue #271, opened 2022-10-26
Sporadic CRUD failure in CI


dholms:
CRUD tests are sporadically failing in CI.

Last I saw it was on racing deletes.

It doesn't happen all the time, doesn't happen on reruns and can't replicate locally.

dholms:
Got it here: https://github.com/bluesky-social/atproto/actions/runs/3332711017/jobs/5513897190

dholms:
wait i just got this for the first time locally. so this seems to be an actual issue & not CI

Still on racing deletes

devinivy:
I poked at this a bit and found that the source of the error is hitting this condition in the mst's `delete()` method: https://github.com/bluesky-social/atproto/blob/70c525f80626b681caaaf80561e25cea5e62f960/packages/repo/src/mst/mst.ts#L330

dholms:
hmmm seems like the most likely thing is that a delete misshaping the MST somehow and either making the key unfindable deleting multiple keys

Sounds like this is probably my domain then.

Oh actually! A way to narrow this down more: serialize the requests to check if it's an issue with races or with the MST impl

pfrazee:
$10 says serializing the requests affects timing and hides the bug again

dholms:
yeah i have tests for serialized deletes on the MST & it passes fine 🤔

PR #364, opened 2022-11-21
Fix a few MST bugs

dholms:
Failing deletes were actually caused by 2 issues:

### Trimming top
One stemmed from the case where a delete leaves the root of the MST pointing to only one node which is a tree. This makes it impossible to tell the layer of the root without further recursing down the tree which breaks some of our logic.

This PR includes 2 fixes:
- on stage, trim the top of the MST if it is a tree pointing to only a tree
- on load, if you need the layer of a tree, recurse until you find it. This shouldn't ever need to be done after taking into account the first fix, but it's good to have around for completeness

### Outdated pointers
The second issue was the case where a key is added that is 2 layers higher than the current highest key. We have to add some empty nodes to preserve the structure. These nodes weren't getting marked as having outdated pointers so when the tree was persisted to the blockstore, they weren't calling for their children to update & were getting persisted with bad pointers

Closes https://github.com/bluesky-social/atproto/issues/271

### DIFF
--- a/packages/repo/src/mst/mst.ts
+++ b/packages/repo/src/mst/mst.ts
@@ -169,27 +169,57 @@ export class MST implements DataStore {
   // If we have neither a hint nor a leaf, we throw an error
   // We could recurse to find the layer, but it isn't necessary for any of our current operations
   async getLayer(): Promise<number> {
+    this.layer = await this.attemptGetLayer()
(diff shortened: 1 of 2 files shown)


devinivy (approved):
Pumped to have that taken care of, looks good!

devinivy (approved):
🦾

PR #365, opened 2022-11-22
Well-defined pagination in PDS

devinivy:
Make keyset cursors a tuple in the PDS in order to avoid ambiguous pagination, e.g. when paginating by datetime when datetimes need not be unique.  In these cases, we now first order by a datetime, and then in the case of a tie order by some unique field such as a URI.

### DIFF
--- a/packages/pds/src/api/app/bsky/feed/getRepostedBy.ts
+++ b/packages/pds/src/api/app/bsky/feed/getRepostedBy.ts
@@ -1,7 +1,7 @@
 import { Server } from '../../../../lexicon'
 import * as GetRepostedBy from '../../../../lexicon/types/app/bsky/feed/getRepostedBy'
 import * as locals from '../../../../locals'
-import { Keyset, paginate } from '../../../../db/util'
+import { paginate, TimeCidKeyset } from '../../../../db/pagination'
 import { getDeclarationSimple } from '../util'
 
 export default function (server: Server) {
@@ -22,7 +22,7 @@ export default function (server: Server) {
           'did_handle.actorType as actorType',
           'did_handle.handle as handle',
           'profile.displayName as displayName',
-          'repost.uri as uri',
+          'repost.cid as cid',
           'repost.createdAt as createdAt',
           'repost.indexedAt as indexedAt',
         ])
@@ -31,7 +31,10 @@ export default function (server: Server) {
         builder = builder.where('repost.subjectCid', '=', cid)
       }
 
-      const keyset = new Keyset(ref('repost.createdAt'), ref('repost.uri'))
+      const keyset = new TimeCidKeyset(
+        ref('repost.createdAt'),
+        ref('repost.cid'),
+      )
       builder = paginate(builder, {
         limit,
         before,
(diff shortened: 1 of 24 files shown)


packages/pds/src/db/util.ts
```diff
@@ -19,22 +21,87 @@ export const actorWhereClause = (actor: string) => {
...
+    }
+  }
+  packFromResult(results: T | T[]): string | undefined {
+    const result = Array.isArray(results) ? results.at(-1) : results
+    if (result === undefined) return
+    return this.pack(this.cursorFromResult(result))
+  }
+  pack(cursor?: Cursor): string | undefined {
+    if (cursor === undefined) return
+    return JSON.stringify([cursor.primary, cursor.secondary])
```
dholms, line 73: hmm maybe strange for the JSON schema to say this type is "string" when it's stringified JSON. Wondering if it may make more sense to separate it with some delimiter 🧐

packages/pds/src/db/util.ts
```diff
@@ -19,22 +21,87 @@ export const actorWhereClause = (actor: string) => {
...
-    .if(opts.before !== undefined, (q) => q.where(opts.by, '<', opts.before))
+    .orderBy(opts.keyset.primary, 'desc')
+    .orderBy(opts.keyset.secondary, 'desc')
+    .if(keysetCondition !== undefined, (qb) =>
+      keysetCondition ? qb.where(keysetCondition) : qb,
+    )
+}
+
+type DefaultRow = { createdAt: string; uri: string }
+export abstract class Keyset<T = DefaultRow> {
```
dholms, line 56: could these just be params on the constructor instead of abstract values?

packages/pds/src/api/app/bsky/feed/getRepostedBy.ts
```diff
@@ -30,14 +31,14 @@ export default function (server: Server) {
         builder = builder.where('repost.subjectCid', '=', cid)
       }
 
+      const keyset = new RepostsKeyset()
```
dholms, line 34: think it could be nice to be able to tell what the cursor is composed of inline. would be pretty straightforward if the primary/secondary vals were passed through the constructor 🤔

dholms:
overall really like the direction this is heading 👌

packages/pds/src/db/util.ts
devinivy: My thinking is that:
 - For consumers the cursor is an opaque value, they shouldn't really muck around in there or try to build one themselves.
 - I reached for a json tuple just because it's a foolproof way to stick two things together and then take them apart without worrying about what characters are allowed in each part.

Admittedly it's a bit ugly-looking, but it has the properties I was going for.  Definitely not married to it!

packages/pds/src/db/util.ts
devinivy: Yeah, let's do it.  I think this ended-up a little funky because of some different things I was trying out along the way.

dholms: nice. i like how this cleaned up 👍

devinivy:
@dholms this should be ready for another look.  I ended-up going a bit further with it and unifying the user search pagination as well.

dholms (approved):
nice this looks great 💪 

hah hadn't considered how switching to CIDs wouldn't help with test ordering 😅

PR #367, opened 2022-11-23
Use integers in lexicons

devinivy:
Includes a fix for lexicons that were using json schema `number` where `integer` would be more appropriate.


dholms (approved):
excellent 👍 

lol @ codegen

PR #375, opened 2022-11-28
Add NotFoundError to getPostThread

pfrazee:
Adds an explicit NotFound error so I can handle it better in the client.

### DIFF
--- a/packages/pds/src/api/app/bsky/feed/getPostThread.ts
+++ b/packages/pds/src/api/app/bsky/feed/getPostThread.ts
@@ -24,7 +24,7 @@ export default function (server: Server) {
         .executeTakeFirst()
 
       if (!queryRes) {
-        throw new InvalidRequestError(`Post not found: ${uri}`)
+        throw new InvalidRequestError(`Post not found: ${uri}`, 'NotFound')
       }
 
       const thread = rowToPost(queryRes)
--- a/lexicons/app/bsky/feed/getPostThread.json
+++ b/lexicons/app/bsky/feed/getPostThread.json
@@ -120,5 +120,8 @@
       "type": "string",
       "not": {"enum": ["app.bsky.system.actorUser", "app.bsky.system.actorScene"]}
     }
-  }
+  },
+  "errors": [
+    {"name": "NotFound"}
+  ]
 }
--- a/packages/api/src/client/schemas.ts
+++ b/packages/api/src/client/schemas.ts
@@ -2243,6 +2243,11 @@ export const methodSchemaDict: Record<string, MethodSchema> = {
         },
       },
     },
+    errors: [
+      {
+        name: 'NotFound',
+      },
+    ],
   },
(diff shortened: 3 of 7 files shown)


packages/pds/src/api/app/bsky/feed/getPostThread.ts
```diff
@@ -24,7 +24,7 @@ export default function (server: Server) {
         .executeTakeFirst()
 
       if (!queryRes) {
-        throw new InvalidRequestError(`Post not found: ${uri}`)
+        throw new InvalidRequestError(`Post not found: ${uri}`, 'NotFound')
```
devinivy, line 27: There's a case to be made that this could be a 404, but I can see it either way.  Mostly just want to surface it to ensure that option was considered 👍

devinivy (approved):
Left a little thought/question, but this looks great!

packages/pds/src/api/app/bsky/feed/getPostThread.ts
pfrazee: A `404` in XRPC means the XRPC method isn't implemented.

devinivy: Ah, I thought those might have been turned into 501s.

devinivy: Did a little poking around and didn't get to the bottom of it, but I did see that 404 is used by the xrpc client as `XRPCNotSupported`, I think when an unsupported status code is returned from the server.  501s are definitely in the mix too in some capacity as `MethodNotImplemented` on the server.  Anyways, I'm not too hung-up on it—this looks good!

dholms: hah i was actually thinking this exact same thing before i got to this comment. i think partially owing to devin & i chatting earlier about trying to keep xrpc as expressive as http as possible.

Losing out on 404s is pretty unfortunate. They're like the most popular status code 😅 and 501s more accurately describe the error.

i wonder if in xrpc-server we should have a "fallback handler" that sends a 501 for any request to `/xrpc/...` that doesn't match a method

dholms: anyway outside of the scope of this PR. but i think it's worth thinking on a bit 🤔

dholms (approved):
👍

packages/pds/src/api/app/bsky/feed/getPostThread.ts
dholms: ah i actually just realized that we do that exact thing already
501 - XRPC supported but not this method
404 - XRPC not supported

mer i see why that's important but still feels a shame :/

pfrazee: Yeah you just have to think about how HTTP servers are going to behave generally. If the server gives a 404, it's saying it doesn't have xrpc routes.

Issue #366, opened 2022-11-23
Auth checks must happen before any other work


whyrusleeping:
Currently we are validating schemas (potentially expensive work) before checking auth. This opens us up to additional DoS vectors that we shouldn't allow.

PR #376, opened 2022-11-28
Enforce max payload size for PLC and PDS

devinivy:
This enforces a max payload size of 100kb for JSON requests to PLC and the PDS.  This involved a little rejiggering of xrpc-server so that it could rely on the consumer providing payload parsing rather than implementing it itself.

### DIFF
--- a/packages/pds/src/index.ts
+++ b/packages/pds/src/index.ts
@@ -51,6 +51,7 @@ const runServer = (
   const mailer = new ServerMailer(mailTransport, config)
 
   const app = express()
+  app.use(express.json({ limit: '100kb' }))
   app.use(cors())
   app.use(loggerMiddleware)
 
--- a/packages/plc/src/server/index.ts
+++ b/packages/plc/src/server/index.ts
@@ -18,7 +18,7 @@ export type App = express.Application
 export const server = (db: Database, port?: number, _version?: string) => {
   const version = _version || '0.0.0'
   const app = express()
-  app.use(express.json())
+  app.use(express.json({ limit: '100kb' }))
   app.use(cors())
 
   app.use(loggerMiddleware)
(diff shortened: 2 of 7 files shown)


dholms (commented):
code looks great 👍 

a bit nervous about moving body processing the express layer & requiring implicit knowledge of what xrpc requires, although i like giving the consumer more control over how the payload is handled 🤔 

I'm wondering if it may make more sense to stick this middleware on the xrpc handler instead & allow the config to be passed into the constructor

devinivy:
First, I am into that— let's do it 👍 

Before I lose them, some additional thoughts I've had around this!
 - I was wary of having xrpc-server inherit the configuration for express's separate json, raw, text parsers and possibly future urlencoded and multipart parsers.
   - On second thought, I don't think xrpc-server will need to support very much configuration for those parsers, so it feels reasonable when I look at this again.
 - I was a little concerned about a server registering these parsers multiple times, e.g. within xrpc-server and then also on other parts of the server with their own independent purposes.  In short, not totally comfy with how middleware might compose.  I tried to mitigate the worst of the "implicit knowledge" by having the server error with a 500 and instructive message when it failed to parse a request body of a particular type.  I think in practice this would probably be "good enough" for our purposes today, though I agree it doesn't feel very smooth.
   - Still not pumped about the middleware composition question, but I think it's a comparable tradeoff to the implicit knowledge consumers would need to setup their own body parsing, so I think it's a wash.
 - I think moving the parsing back into xrpc will help us when it comes to coordinating auth and payload parsing in #366.
   - I also think there could be some radically different solutions to #366, but moving parsing into xrpc-server remains the best way for us to leave our options open for dealing with this in the near-term.

dholms (approved):
awesome 🙌 

thx again for back & forth on this 🙏

PR #377, opened 2022-11-29
Standardize limit schema across lexicons

### LINKED ISSUE #370: Add defaults and upper-values to limit value in all API queries
At some point we lost the default `limit` values in APIs, leading the methods to fetch way more than they should. Two things need to happen:

- [ ] Make sure a default value is set if no `limit` is provided
- [ ] Clamp the `limit` to a reasonable range to avoid overly resource-taxing queries

devinivy:
Most lexicons were missing a default value for limits used in pagination.  This standardizes all limits on min: 1, max: 100, default: 50.  We can tweak that, of course— just nabbed the values that were being used for the one query that already had a min/max/default.

Closes #370

### DIFF
--- a/lexicons/app/bsky/actor/getSuggestions.json
+++ b/lexicons/app/bsky/actor/getSuggestions.json
@@ -6,7 +6,7 @@
   "parameters": {
     "type": "object",
     "properties": {
-      "limit": {"type": "integer", "maximum": 100},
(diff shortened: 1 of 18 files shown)


dholms (approved):
sweet 👌

pfrazee:
Thanks!

PR #378, opened 2022-11-29
Image processing interface

devinivy:
This outlines an interface for image processors, and gives a reference implementation backed by [sharp](https://sharp.pixelplumbing.com).

**Edit**
This now includes:
 - an image processing service that can be used for serving resized images.  Comes with a cache and generic storage, all streaming.
 - a helper class for building signed URLs for imgproxy and the pds's own image service.


packages/pds/src/image/server.ts
```diff
@@ -0,0 +1,105 @@
...
+          // @NOTE sharp does emit this in time to be set as a header
+          .once('info', (info) => res.setHeader('content-length', info.size))
+          .pipe(res)
+      )
+    } catch (err: unknown) {
+      if (err instanceof BadPathError) {
+        return next(new InvalidRequestError(err.message))
+      }
+      if (err instanceof BlobNotFoundError) {
+        return next(new XRPCError(404, 'Image not found', 'NotFound'))
```
dholms, line 51: Can't use 404 for XRPC error code 🙃

packages/pds/src/image/sharp.ts
```diff
@@ -0,0 +1,79 @@
+import { Readable } from 'stream'
+import sharp from 'sharp'
+import {
+  formatsToMimes,
+  forwardStreamErrors,
+  ImageProcessor,
+  Options,
+} from './util'
+
+export class SharpImageProcessor implements ImageProcessor {
```
dholms, line 10: since this doesn't wrap any sort of state, thoughts on splitting it up & just making it two functions?

dholms (approved):
lookin good 👍

packages/pds/src/image/sharp.ts
devinivy: Yeah for sure, good call 👍 originally was thinking `ImageProcessor` would be a pluggable interface that would have to be passed into the server, but I think it's safe to say that's become a non-concern.

packages/pds/src/image/server.ts
devinivy: Was thinking it could be acceptable here since this is a non-xrpc service, and I believe it matches imgproxy's behavior.  One awkward part is that I'm using XRPC errors, but it's only because they make for a nice http error interface—maybe I oughtta switch them to [http-errors](https://github.com/jshttp/http-errors) to avoid the mixed signals.

pfrazee: Huh this actually brings up an interesting problem for XRPC -- if it doesn't support 404s as standard behaviors, it might have problems with proxies that are expecting 404s to indicate not found.

I wonder if the solution here is to use 404s in XRPC but differentiate between "lack of support" and "resource not found" by the response body? If the response body is a valid XRPC error object, we know it's the latter. Not sure why that didn't occur to me before!

devinivy: I have a feeling we'll eventually need to grapple with it one way or another, and I like that idea 👍

pfrazee:
Looks great!

packages/pds/src/image/server.ts
dholms: yeah @pfrazee we actually briefly riffed on that yesterday^^
404 in HTTP already kinda means two things: "route not implemented" & "resource not found" & we're actually at an advantage with XRPC because we do have structured responses/errors

PR #379, opened 2022-11-30
Lexicon rework

pfrazee:
Lexicon has been working wonderfully for us thus far, but we've run into some challenges that required a rework of the format.

Historically Lexicon has embedded JSON-Schema definitions in itself to define constraints on records and XRPC params/inputs/outputs. We then generated code from the lexicons which would provide us with good static types, and which would carry the original JSON-Schema definitions for runtime validation. As we used this format, two particular challenges emerged:

1. Code generation in stricter languages such as Go struggled to generate code from the JSON-Schemas due to the flexibility of the latter. It was simply too easy to write JSON-Schemas which didn't map cleanly to Go's type system. A more constrained language would solve this by ensuring that the Lexicon docs never deviated from what can be represented in a stricter type system.
2. JSON-Schema lacks all of the features we need; specifically, the introduction of "blobs" into ATP created awkward points where we needed to be able to write custom constraints.

A third issue with Lexicon was how we defined the hash-fragment pointers. We originally defined them as JSON-Pointers which was A. brittle as non-breaking refactors to a schema would break the pointers, and B. very visually noisy, which discouraged code-sharing.

With this refactor, we are making three core changes:

1. Moving from JSON-Schema to a custom format which is similar but has ATP-specific constraints. Importantly this custom format does not allow embedded objects and uses order-dependent fallthroughs on variadic types, both of which match programming languages' type systems more closely.
2. Reorganizing definitions in Lexicons to all fall under the "defs" object, with the "main" key representing the primary definition.  (You can think of it as similar to a module that exports definitions.)
3. Updating the hash-fragment spec to always point to the keys of the "defs" object. If no hash-fragment is present, the URI points to "/defs/main".

Tasks:

- Lexicon
  - [x] New types and validation
  - [x] Records, procedures, and queries can only be the main def
  - [x] Date-time type. ADDED: "datetime" type
  - [x] Empty object type. ADDED: "unknown" type
  - [x] Blobs in objects and arrays
- Codegen
  - [x] Client
  - [x] Server
  - [x] Tokens
  - [x] External-file refs
  - [x] knownValues hints
  - [x] Generate comments
- [x] Update all lexicons
- [x] Update all code to work with new lexicon system

After this PR, we should plan to refactor the lexicons for definition re-use, and consider adding the shorthand tooling.

Some gotchas to be aware of:

- The generated code no longer creates a listing of record vs method schemas (just one listing) so code that previously depended on that assumption could break. The better solution now is it create a `Lexicons` (from the lexicon dep) and then call `getDefOrThrow` and specify the desired definition type (`["record"]` or `["query", "procedure"]`).


dholms:
haven't fully reviewed but wow this is gonna be really nice to work with

pfrazee:
@dholms @devinivy Be sure to take a close look at changes to the PDS package. I tried to keep my changes as minimal as possible; you may want to make tweaks after this merges.

lexicons/app/bsky/graph/assertion.json
```diff
@@ -1,22 +1,27 @@
...
+      "type": "record",
+      "key": "tid",
+      "record": {
         "type": "object",
-        "required": ["did", "declarationCid"],
+        "required": ["assertion", "subject", "createdAt"],
         "properties": {
-          "did": {"type": "string"},
-          "declarationCid": {"type": "string"}
+          "assertion": {"type": "string"},
```
devinivy, line 12: Do we want to drop some `knownValues` here using the assertion tokens that were setup?

lexicons/app/bsky/graph/getAssertions.json
```diff
@@ -71,20 +71,10 @@
       "properties": {
         "cid": {"type": "string"},
         "actorType": {
-          "oneOf": [
-            {"$ref": "#/defs/actorKnown"},
-            {"$ref": "#/defs/actorUnknown"}
-          ]
+          "type": "string", 
+          "knownValues": ["app.bsky.system.actorUser", "app.bsky.system.actorScene"]
```
devinivy, line 75: Still haven't totally wrapped my head around the `token` story— would it make sense to say that the `type` of this field is `token`?  Generally sort of wondering—how (if at all) do we tie together the existence of these tokens with the validation of this property?

lexicons/com/atproto/repo/getRecord.json
```diff
@@ -1,27 +1,27 @@
 {
   "lexicon": 1,
   "id": "com.atproto.repo.getRecord",
-  "type": "query",
-  "description": "Fetch a record.",
-  "parameters": {
-    "type": "object",
-    "required": ["user", "collection", "rkey"],
```
devinivy, line 8: Current feeling is that I think I'll miss "required" for params.  I don't recall what the considerations there were, though.

lexicons/app/bsky/graph/assertion.json
pfrazee: That would make sense. I'd suggest we have a separate PR after this where we do a refactor of the lexicons, to reuse some definitions and fix bits like that.

lexicons/app/bsky/graph/getAssertions.json
pfrazee: Ah, no the token type is just for declaring empty NSIDs, not for consuming tokens. But that's an easy misunderstanding and we might need to think about clarity.

lexicons/com/atproto/repo/getRecord.json
pfrazee: Bleh. (Not you, the situation.) Parameters are awkward because they're almost exactly like objects but they're not, so we can't reuse object definitions but it's hard to deviate from them too. The loss of "required" was me making the least awkward choice for the schema language.

I'm open to ideas for re-introducing that information.

packages/api/src/client/types/app/bsky/system/declaration.ts
```diff
@@ -1,12 +1,7 @@
...
 * GENERATED CODE - DO NOT MODIFY
 */
-export type ActorKnown =
-  | 'app.bsky.system.actorUser'
-  | 'app.bsky.system.actorScene'
-export type ActorUnknown = string
-
 export interface Record {
-  actorType: ActorKnown | ActorUnknown;
+  actorType: string;
```
devinivy, line 5: There's a nice TS trick we can use to get the `knownValues` into the type system.  The types validate the same way, but the allowed values are preserved and can appear e.g. through intellisense:
```js
  actorType: 'app.bsky.system.actorUser' | 'app.bsky.system.actorScene' | (string & {});
```

packages/api/src/client/types/com/atproto/repo/putRecord.ts
```diff
@@ -5,38 +5,25 @@ import { Headers, XRPCError } from '@atproto/xrpc'
...
-export interface CallOptions {
-  headers?: Headers;
-  qp?: QueryParams;
-  encoding: 'application/json';
-}
-
 export interface InputSchema {
-  /**
-   * The DID of the repo.
-   */
```
devinivy, line 17: Not a big deal, but I was a fan of this documentation appearing inline during development :) We can always add it back later if we'd like—happy to peel off an issue if it's of interest.

packages/api/src/client/types/app/bsky/graph/assertCreator.ts
```diff
@@ -0,0 +1,4 @@
+/**
+* GENERATED CODE - DO NOT MODIFY
+*/
+export const MAIN = 'app.bsky.graph.assertCreator#main'
```
devinivy, line 4: How/where do we imagine this constant being used?  It would be sort of nice if the string values used for the tokens appeared as a const somewhere, too!

General thought bouncing around my brain is that some deeper connection between the existence of a token and the string value we use to represent them might help crystallize lexicon's concept of tokens a bit better.  Also recognize I may just not have the right feel for them yet.

packages/api/src/client/types/com/atproto/repo/putRecord.ts
pfrazee: I'll add that to the PR tasks, probably not too hard to do in the code generation.

packages/api/src/client/types/app/bsky/graph/assertCreator.ts
pfrazee: That's sort of a wonky side-effect of what I was mainly thinking about, which is non-main tokens having a const exported. But the main usage of tokens in our codegen is in the main index.ts, where they're enumerated in dicts.

Open to ideas for better approaches to tokens. The trick is, they still need to ultimately act as strings, and we can't do a closed set so enums are a bit awkward to use.

lexicons/com/atproto/repo/getRecord.json
devinivy: I don't think object is so far off, especially now that you don't allow sub-objects in `object` unless they are behind a reference.  We sort of have the same issue with json, since it could contain a ref to something it can't support, such as a `video`.  My gut would be to treat them as objects.

Side note—the fact that arrays are allowed inside object also seems alright to me.  While params don't support sub-objects in any standard way, they do support arrays: `x=one&x=two` is `x: ['one', 'two']` (can see this reflected e.g. in [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams)).

packages/lex-cli/src/util.ts
```diff
@@ -145,6 +110,26 @@ export function applyFileDiff(diff: FileDiff[]) {
...
+    }
+    return true
+  } else {
+    for (const k in node) {
+      if (k === '_errors') {
+        continue
+      }
+      if (printZodError(node[k], `${path}/${k}`)) {
+        // return true
+      }
```
devinivy, line 127: Choose your own adventure!  (Or ignore lol—I wasn't totally sure what the intent was here.)
```suggestion
      if (printZodError(node[k], `${path}/${k}`)) {
        return true
      }
```

```suggestion
      printZodError(node[k], `${path}/${k}`)
```

lexicons/com/atproto/repo/getRecord.json
pfrazee: 99% you know this, but just to clarify a bit --

"Lexicon refs" and "data links" are two different things. Lexrefs are essentially links in lexicon definitions, while data links are links in values. Objects _do_ allow sub-objects, but lexicon object definitions _dont_ allow inline object definitions -- you have to use a lexref. So: more accurately, LexParams are LexObject subsets that don't allow sub-LexObjects.

I suppose a simple solution would be to mirror the object-definition structure accordingly:

```js
"parameters": {
  "type": "params",
  "required": ["foo"],
  "params": { // or "properties"
    "foo": {"type": "string"},
    "bar": {"type": "string"},
  }
}
```

It's more verbose, but it's also more consistent and it gives us a place to put the required-param info.

packages/lex-cli/src/util.ts
pfrazee: Haha nice, yeah obviously I forgot to cleanup after deciding how to do that algorithm. What I settled on is _okay_ but could be much better. If you ever get frustrated with the output of lex-cli for errors in lexicons, you know where to look.

packages/api/src/client/types/app/bsky/system/declaration.ts
devinivy: Typescript will upcast `'my-string' | string` to `string`, and you lose the information in the type system that `'my-string'` is a special value.  Here's an example:
Screen Shot 2022-12-01 at 11 25 37 AM
Screen Shot 2022-12-01 at 11 25 25 AM

pfrazee: Interesting! Okay I'll do that when I add the feature to codegen.

packages/lexicon/src/lexicons.ts
```diff
@@ -0,0 +1,203 @@
...
+    yield [`lex:${doc.id}#${defId}`, doc.defs[defId]]
+    if (defId === 'main') {
+      yield [`lex:${doc.id}`, doc.defs[defId]]
+    }
+  }
+}
+
+// WARNING
+// this method mutates objects
+// -prf
```
devinivy, line 182: Appreciate these warnings :)

pfrazee: Yeah :) you gotta track that shit

lexicons/com/atproto/repo/getRecord.json
devinivy: Ah yeah sorry, I did get a little turned around there in a couple ways.  Ignore the `video` example!  I was also looking at the spec which shows this for `object`:
```js
properties: Record<string, LexRef | LexArray | LexPrimitive | LexRef[]>
```

I see now that the blob variants are also allowed there, though ✨ 

The main thought I was having is that objects can already contain refs to things that they can't represent, and that is the same issue we have with params.  I am realizing now that you've spec'ed that issue away for objects, though.  So the issue that remains is just that parameters could contain a ref to an object, and that's not allowed.  I think the proposal you tossed out seems very reasonable 👍 👍

packages/api/src/client/types/app/bsky/actor/getSuggestions.ts
```diff
@@ -49,3 +31,28 @@ export function toKnownErr(e: any) {
...
+  follow?: string;
+  [k: string]: unknown;
+}
+
+export interface Declaration {
+  cid: string;
+  actorType:
+    | 'app.bsky.system.actorUser'
+    | 'app.bsky.system.actorScene'
+    | (string & {});
```
devinivy, line 56: Glorious ❤️

packages/api/src/client/types/app/bsky/system/declaration.ts
pfrazee: Done!

packages/api/src/client/types/com/atproto/repo/batchWrite.ts
```diff
@@ -5,54 +5,49 @@ import { Headers, XRPCError } from '@atproto/xrpc'
...
 }
 
 export function toKnownErr(e: any) {
   if (e instanceof XRPCError) {
   }
   return e
 }
+
+export interface Create {
+  action: string;
```
devinivy, line 33: I think we want this to reflect the const listed in the schema `action: 'create'` .

pfrazee: Fixed

packages/xrpc-server/tests/errors.test.ts
```diff
@@ -45,33 +46,33 @@ describe('Procedures', () => {
...
       throw new Error('Didnt throw')
-    } catch (e: any) {
+    } catch (e) {
       expect(e instanceof XRPCError).toBeTruthy()
-      expect(e.success).toBeFalsy()
-      expect(e.error).toBe('InvalidRequest')
-      expect(e.message).toBe('Invalid Request')
+      expect((e as XRPCError).success).toBeFalsy()
+      expect((e as XRPCError).error).toBe('InvalidRequest')
+      expect((e as XRPCError).message).toBe('Invalid Request')
```
devinivy, line 75: Not suggesting any changes around this here/now, but I realized recently that node's `assert()` hooks nicely into the type system so that these pesky `as XRPCError`s can be removed.  Looks something like this:
```ts
import assert from 'assert'
// ...
      assert(e instanceof XRPCError) // TS now knows e is an XRPCError
      expect(e.success).toBeFalsy()
```

packages/xrpc-server/src/types.ts
```diff
@@ -3,7 +3,8 @@ import { isHttpError } from 'http-errors'
 import zod from 'zod'
 import { ResponseType, ResponseTypeStrings } from '@atproto/xrpc'
 
-export type Params = typeof express.request['query']
+export type UndecodedParams = typeof express.request['query']
```
devinivy, line 6: I just took a closer look at this type, and I think in the future we may want to switch query param parsers to guarantee much simpler output.  I believe there could be objects kicking around in here, which I'd advocate for nipping in the bud at the parsing step.  Poor `decodeQueryParam` can't assume anything about the value, when in actually we can do much better, guaranteeing `string | string[]`.  I could handle this in my pending xrpc-server work if we want to make that change.

devinivy (approved):
As @dholms said, it's clear you had that in you just waiting to come out.  Love how this turned-out, great stuff! 💪

packages/xrpc-server/tests/errors.test.ts
pfrazee: Nice, good to know

packages/xrpc-server/src/types.ts
pfrazee: Sure whatever you think is best

packages/pds/src/api/app/bsky/feed/getAuthorFeed.ts
```diff
@@ -19,6 +19,10 @@ export default function (server: Server) {
         throw new AuthRequiredError()
       }
 
+      if (!author) {
```
dholms, line 22: since we updated `parameters` to allow `required`. shouldn't this be handled in xrpc-server instead?

pfrazee: Yeah there's a bunch of checks that I added prior to reworking params that I haven't removed because I dont wanna do it.

devinivy: I'm happy to clean those up once this lands 👍

pfrazee: 1 sec I'll go through and remove them

pfrazee: Done

dholms: kk coolio 👍 

wanted to make sure those weren't required anymore. & yeah i think it's best to do it in this PR or shortly after otherwise we'll have that cruft hanging around for a while lol 😅

dholms (approved):
truly beautiful 🤌

PR #382, opened 2022-12-02
Lexicon tweaks: encoding and union $type discriminators

pfrazee:
Two more Lexicon updates to accommodate stricter programming languages:

- LexXrpcBody `.encoding` is now always a `string`. If multiple encodings must be specified, use a comma-separated list.
- Union discriminators
  - The `union` type now only supports object-values with a `$type` field.
  - The `$type` field must specify the lexicon ID being used.
  - The `union` def now has a `closed` field which, if true, will fail validation if a `$type` is provided which does not match the enumerated refs.

### DIFF
--- a/packages/lexicon/src/types.ts
+++ b/packages/lexicon/src/types.ts
@@ -173,7 +173,7 @@ export type LexXrpcParameters = z.infer<typeof lexXrpcParameters>
 
 export const lexXrpcBody = z.object({
   description: z.string().optional(),
-  encoding: z.union([z.string(), z.string().array()]),
+  encoding: z.string(),
   schema: z.union([lexRefVariant, lexObject]).optional(),
 })
 export type LexXrpcBody = z.infer<typeof lexXrpcBody>
--- a/packages/lexicon/src/types.ts
+++ b/packages/lexicon/src/types.ts
@@ -82,6 +82,7 @@ export const lexRefUnion = z.object({
   type: z.literal('union'),
   description: z.string().optional(),
   refs: z.string().array(),
+  closed: z.boolean().optional(),
 })
 export type LexRefUnion = z.infer<typeof lexRefUnion>
 
@@ -285,6 +286,14 @@ export function hasProp<K extends PropertyKey>(
   prop: K,
 ): data is Record<K, unknown> {
   return prop in data
+}
+
+export const discriminatedObject = z.object({ $type: z.string() })
+export type DiscriminatedObject = z.infer<typeof discriminatedObject>
+export function isDiscriminatedObject(
+  value: unknown,
+): value is DiscriminatedObject {
+  return discriminatedObject.safeParse(value).success
 }
 
 export class LexiconDocMalformedError extends Error {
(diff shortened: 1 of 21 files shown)


whyrusleeping:
Shouldnt media on feed.getPostThread also have a type field?
```diff
diff --git a/lexicons/app/bsky/feed/getPostThread.json b/lexicons/app/bsky/feed/getPostThread.json
index 0bf365d..8922767 100644
--- a/lexicons/app/bsky/feed/getPostThread.json
+++ b/lexicons/app/bsky/feed/getPostThread.json
@@ -33,6 +33,7 @@
       "type": "object",
       "required": ["uri", "cid", "author", "record", "replyCount", "repostCount", "upvoteCount", "downvoteCount", "indexedAt"],
       "properties": {
+        "type": {"type": "string", "const": "post"},
         "uri": {"type": "string"},
         "cid": {"type": "string"},
         "author": {"type": "ref", "ref": "app.bsky.actor.ref#withInfo"},
```

pfrazee:
@whyrusleeping The discriminator used by values is the `$type` field and is not specified in the lexicons

whyrusleeping:
oooooh, so its implicit now? i see

packages/lexicon/src/types.ts
```diff
@@ -287,6 +288,14 @@ export function hasProp<K extends PropertyKey>(
   return prop in data
 }
 
+export const discriminatedObject = z.object({ $type: z.string() })
```
devinivy, line 291: If I am reading this right, I think we want this type to also allow unknown keys.

E.g. right now the type is:
```ts
type DiscriminatedObject = {
    $type: string;
}
```

And I think we want:
```ts
type DiscriminatedObject = {
    $type: string;
    [x: string]: unknown;
}
```

If that's right, that could by done by tacking on a `.catchall(z.unknown())`.

packages/pds/src/lexicon/types/app/bsky/feed/embed.ts
```diff
@@ -5,7 +5,7 @@ import * as AppBskyActorRef from '../actor/ref'
 
 /** A list embeds in a post or document. */
 export interface Main {
-  items?: (Media | Record | External | Unknown)[];
+  items?: (Media | Record | External | { $type: string })[];
```
devinivy, line 8: Do we need to allow unknown keys on `{ $type: string }`?

devinivy (approved):
Had a couple questions, but it could be based on a misunderstanding I'm having.  In general this looks solid!

packages/lexicon/src/types.ts
pfrazee: Oh -- you're right, but not for there; it's an issue in the codegen. I'll fix that.

packages/pds/src/lexicon/types/app/bsky/feed/embed.ts
pfrazee: Yeah here's where I need to fix that.

pfrazee: Fixed

pfrazee:
@whyrusleeping correct!

devinivy (approved):

whyrusleeping:
@pfrazee last question for clarity, the union `$type` will *always* be fully qualified?

pfrazee:
It's always a string of `{nsid}[#hash]` with possibly `lex:` in front, but never just `#hash`

pfrazee:
Actually, edit: It may be undefined. You only need to set `$type` when using a union.

dholms (approved):
sweet this seems reasonable to me 👌

PR #383, opened 2022-12-05
File uploads & profile image support

dholms:
Blobs!

A user creates a blob in their repo in a two step process:
- uploading the blob
- referencing the blob by CID with a record in their repo

On upload, we store some information about a blob including:
- size
- mimetype
- height/width (if image)

When a blob is referenced by a CID, we track that reference in the DB for later use in sync.

This PR also hooks up the image processing from https://github.com/bluesky-social/atproto/pull/378
Which includes 4 new config variables:
- imgUriSalt: salt for generating img uris
- imgUriKey: key for generating img uris
- imgUriEndpoint (optional): if provided, prefixes img uris with this endpoint. if not provided, then mounts image processing to the server & defaults to `cfg.publicUrl + /image`
- blobCacheLocation (optional): only used if doing own image processing & is the location of an on-disk cache of processed images


packages/xrpc-server/src/server.ts
```diff
@@ -109,6 +115,12 @@ export class Server {
       } catch (e) {
         throw new InvalidRequestError(String(e))
       }
+
+      // @TODO move to middleware
```
devinivy, line 119: I believe it's the `raw()` middleware that does this 👍 Should jibe nicely with the reorganization in #384.

devinivy (commented):
I know this is in progress, but just did a quick sweep of review.  Pumped for this, it's looking solid!

packages/pds/src/repo/blobs.ts
```diff
@@ -0,0 +1,165 @@
...
+      throwInvalid(
+        `Referenced image width is too large. Expected: ${blob.constraints.maxWidth}. Got: ${found.width}`,
+      )
+    }
+  }
+}
+
+const acceptedMime = (mime: string, accepted: string[]): boolean => {
+  if (accepted.indexOf('*/*') > -1) return true
+  return accepted.indexOf(mime) > -1
```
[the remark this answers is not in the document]
dholms: oh i forgot this exists 🙏

packages/pds/src/repo/blobs.ts
```diff
@@ -0,0 +1,165 @@
...
+
+export const addUntetheredBlob = async (
+  dbTxn: Database,
+  blobs: BlobStore,
+  mimeType: string,
+  bytes: Uint8Array,
+): Promise<CID> => {
+  const tempKey = await blobs.putTempBytes(bytes)
+  // @TODO calcualte cid with chunking
+  const cid = await cidForData(bytes)
```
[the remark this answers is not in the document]
dholms: yup yup totally 👍

packages/pds/src/repo/blobs.ts
```diff
@@ -0,0 +1,165 @@
...
+    throwInvalid(
+      `Referenced Mimetype is not accepted. Expected: ${blob.constraints.accept}, Got: ${blob.mimeType}`,
+    )
+  }
+  if (blob.constraints.type === 'image') {
+    if (!blob.mimeType.startsWith('image')) {
+      throwInvalid(`Expected an image, got ${blob.mimeType}`)
+    }
+    if (
+      blob.constraints.maxHeight &&
```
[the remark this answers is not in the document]
dholms: ah yup we'll probably want that as well

packages/xrpc-server/src/server.ts
devinivy: Maybe you already gave this a whirl, but I think you will need to set its [`type` option](https://expressjs.com/en/api.html#express.raw) for our use-case.

packages/pds/src/api/com/atproto/data.ts
```diff
@@ -0,0 +1,27 @@
...
+  server.com.atproto.data.uploadFile({
+    auth: ServerAuth.verifier,
+    handler: async ({ input, req, res }) => {
+      const { db, blobstore } = locals.get(res)
+
+      const cid = await repo.addUntetheredBlobStream(
+        db,
+        blobstore,
+        input.encoding,
+        req,
```
dholms, line 16: the bytes don't get read off the req in middleware, so we can just send that readstream directly to the blob processor.

We may want to make this more explicit in xrpc-server, & pipe them into a stream that we stick on to `input`. Thoughts?

devinivy: Yepyep, that sounds good to me.

packages/pds/tests/_util.ts
```diff
@@ -44,6 +51,8 @@ export const runTestServer = async (
     'https://pds.public.url',
   )
 
+  const blobstoreLoc = '/tmp/' + randomStr(5, 'base32')
```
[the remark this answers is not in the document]
dholms: ah excellent 🙏

packages/xrpc-server/src/util.ts
```diff
@@ -91,9 +92,19 @@ export function validateInput(
...
   }
 
+  // if middleware already got the body, we pass that along as input
+  // otherwise, we pipe it into a readable stream
+  let body
+  if (req.complete) {
+    body = req.body
+  } else {
+    body = new stream.PassThrough()
+    req.pipe(body)
```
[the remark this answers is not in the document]
dholms: nice catch. fixed this on my stream pass 👌

packages/crypto/src/sha.ts
```diff
@@ -10,3 +12,17 @@ export const sha256 = async (
   const hash = await mf.sha256.digest(bytes)
   return hash.digest
 }
+
+export const sha256Stream = async (stream: Readable): Promise<Uint8Array> => {
+  const hash = crypto.createHash('sha256')
```
devinivy, line 17: Here's what this might look like like using `stream.promises.pipeline()`:
```ts
  const hash = crypto.createHash('sha256')
  await pipeline(stream, hash)
  return hash.read()
```

devinivy (approved):
I know it's still technically in draft mode, but ✔️✔️✔️ this is killer!

packages/crypto/src/sha.ts
dholms: so i actually tried that, but it was getting backed up, blocking the stream, and then timing out 🤔 
so i ended up throwing it out & doing it this way lol

packages/pds/tests/image/server.test.ts
```diff
@@ -1,24 +1,36 @@
...
-import {
-  BlobDiskCache,
-  BlobDiskStorage,
-  ImageProcessingServer,
-} from '../../src/image/server'
+import { BlobDiskCache, ImageProcessingServer } from '../../src/image/server'
+import { DiskBlobStore } from '../../src'
+import { cidForData } from '@atproto/common'
+import { CID } from 'multiformats/cid'
+import { runTestServer } from '../_util'
```
devinivy, line 12: I don't think this is used anywhere.
```suggestion
```

devinivy (approved):
Biiiig yup!

packages/crypto/src/sha.ts
devinivy: Oh, funky!

devinivy (approved):

pfrazee:
Great fuckin work bois. Looks good on my end. 

One update I will probably request in a followup is to add the avatar cid to `app.bsky.actor.ref#withInfo` since I pretty much always need to render avatars any time that is relevant (and how awesome is it we have just one lexicon type to modify there).

PR #384, opened 2022-12-05
Auth in XRPC server

### LINKED ISSUE #366: Auth checks must happen before any other work

devinivy:
This introduces the idea of auth into xrpc server, largely to resolve #366 (particularly in the context of blob support).  The core changes are:
 - xrpc-server allows you to configure auth for each xrpc method.
 - xrpc-server routing has been updated so that each method gets its own route with its own specialized middleware stack.
 - xrpc-server error handling has been reorganized so that it's all handled by a single middleware.
 - the interface for an xrpc handler has been updated to take a single `context` argument containing `params`, `input`, `auth`, `req`, `res`.
 - lex-cli codegen for the server supports configuring auth on each method, and tying the type of the handler to the result of the auth verifier.
 - lex-cli codegen is now more in-line with the style of the project config.  In particular, this means that lots of colons were removed.
 - Auth is required now on the entirety of the `app.bsky` namespace on the pds.

Definitely want to review [ignoring whitespace](https://github.com/bluesky-social/atproto/pull/384/files?w=1).  The core changes are in `packages/lex-cli/src/codegen/`, `packages/xrpc-server/src/`, and `packages/pds/src/auth.ts`.


packages/lex-cli/src/codegen/common.ts
```diff
@@ -4,10 +4,11 @@ import prettier from 'prettier'
 import { GeneratedFile } from '../types'
 
 const PRETTIER_OPTS = {
-  parser: 'babel',
+  parser: 'babel-ts',
```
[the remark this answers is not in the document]
dholms: yay to semicolon removals

pfrazee (approved):
Looks great to me!

packages/xrpc-server/src/types.ts
```diff
@@ -12,6 +12,12 @@ export const handlerInput = zod.object({
 })
 export type HandlerInput = zod.infer<typeof handlerInput>
 
+export const handlerAuth = zod.object({
+  credentials: zod.any(),
```
dholms, line 16: could we use generics here so we actually know what types we're getting out of the auth fn?

dholms (approved):
just the question on generics, but besdies that, this is looking really great 👌

packages/xrpc-server/src/types.ts
devinivy: In the handlers inside the pds, each handler's `ctx.auth` is properly typed.  That level of specificity is provided by lex-cli's typings rather than xrpc-server's.  That's the same way the handler `input` and `params` typings work, too.  Do you think that covers it?

dholms: oh yup that's good 👍

PR #389, opened 2022-12-08
Fix S3 blob store, issue putting object from a stream

devinivy:
I was seeing an issue with the `putObject()` calls within the S3 blob store.  The issue is that the client library needs to know the length of the stream in advance.  Sounds like the way to do this is to use `Upload` from `@aws-sdk/lib-storage`.

Refs https://github.com/aws/aws-sdk-js-v3/issues/2348 https://github.com/aws/aws-sdk-js-v3/issues/1382

Error was from `S3BlobStore.putTemp (/atproto/packages/aws/src/s3.ts:39:5)`:
```
NotImplemented: A header you provided implies functionality that is not implemented
```

### DIFF
--- a/packages/aws/package.json
+++ b/packages/aws/package.json
@@ -18,6 +18,7 @@
     "@atproto/crypto": "*",
     "@aws-sdk/client-kms": "^3.196.0",
     "@aws-sdk/client-s3": "^3.224.0",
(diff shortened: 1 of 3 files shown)


dholms (approved):
nice good find 🙇

PR #392, opened 2022-12-08
Add 'avatar' to the outputs of views

pfrazee:
This PR adds `avatar` URLs to the output of most views.

As we've discussed, this is creating a _lot_ of signed URIs and we need to look at special-casing avatars


dholms (approved):
looks good 💪

PR #395, opened 2022-12-08
Verify content type of files

dholms:
There was a concern that a user could supply a bad mimetype & screw up bookkeeping for future uploads of the same bytes.

This PR:
- checks the magic bytes at the start of a file to try to sus out content type. If it finds something, it stores that. If not, it uses the user offered mimetype
- we also attempt to find image dimensions on every blob (since we can't guarantee the content type yet), but simply return null if we find that we can't process the blob as an image

### DIFF
--- a/packages/pds/src/repo/blobs.ts
+++ b/packages/pds/src/repo/blobs.ts
@@ -9,6 +9,7 @@ import { BlobRef, PreparedWrites } from './types'
 import { Blob as BlobTable } from '../db/tables/blob'
 import * as img from '../image'
 import { sha256Stream } from '@atproto/crypto'
+import { fileTypeFromStream } from 'file-type'
 
 export const addUntetheredBlob = async (
   dbTxn: Database,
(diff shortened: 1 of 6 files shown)


packages/pds/src/repo/blobs.ts
```diff
@@ -9,28 +9,20 @@ import { BlobRef, PreparedWrites } from './types'
...
-  }
-
-  const [tempKey, size, sha256, imgInfo] = await Promise.all([
+  const [tempKey, size, sha256, imgInfo, fileType] = await Promise.all([
     blobstore.putTemp(cloneStream(blobStream)),
     streamSize(cloneStream(blobStream)),
     sha256Stream(cloneStream(blobStream)),
-    maybeGetImgInfo(cloneStream(blobStream)),
+    img.maybeGetInfo(cloneStream(blobStream)),
+    fileTypeFromStream(blobStream),
```
dholms, line 25: so this is actually the only way i could get it to work 🙃 
cloning each stream except for the file type one 

i'm not super pleased with this, but it does work & im honestly not sure what else to do right now 😅

devinivy (approved):
Killer 💪

PR #396, opened 2022-12-08
Support compression in requests to xrpc-server

devinivy:
Allow raw payloads to be sent compressed to xrpc-server.

### DIFF
--- a/packages/pds/src/index.ts
+++ b/packages/pds/src/index.ts
@@ -102,9 +102,9 @@ const runServer = (
 
   const apiServer = API({
     payload: {
-      jsonLimit: '100kb',
-      textLimit: '100kb',
-      rawLimit: '5mb',
+      jsonLimit: 100 * 1024, // 100kb
+      textLimit: 100 * 1024, // 100kb
+      blobLimit: 5 * 1024 * 1024, // 5mb
     },
   })
   app.use(health.router)
--- a/packages/xrpc-server/src/util.ts
+++ b/packages/xrpc-server/src/util.ts
@@ -1,6 +1,9 @@
+import { Readable } from 'stream'
+import { createDeflate, createGunzip, Deflate, Gunzip } from 'zlib'
 import express from 'express'
-import { Lexicons, LexXrpcProcedure, LexXrpcQuery } from '@atproto/lexicon'
 import mime from 'mime-types'
+import { Lexicons, LexXrpcProcedure, LexXrpcQuery } from '@atproto/lexicon'
+import { forwardStreamErrors } from '@atproto/common'
 import {
   UndecodedParams,
   Params,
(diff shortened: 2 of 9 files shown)


packages/pds/src/index.ts
```diff
@@ -102,9 +102,9 @@ const runServer = (
 
   const apiServer = API({
     payload: {
-      jsonLimit: '100kb',
-      textLimit: '100kb',
-      rawLimit: '5mb',
+      jsonLimit: 100 * 1024, // 100kb
+      textLimit: 100 * 1024, // 100kb
+      blobLimit: 5 * 1024 * 1024, // 5mb
```
dholms, line 107: awesome 🙌

packages/xrpc-server/src/util.ts
```diff
@@ -182,3 +187,33 @@ export function processBodyAsBytes(req: express.Request): Promise<Uint8Array> {
...
+  req: express.Request,
+  maxSize: number | undefined,
+): Readable {
+  const contentEncoding = req.headers['content-encoding']
+  const contentLength = req.headers['content-length']
+
+  if (
+    maxSize !== undefined &&
+    contentLength &&
+    parseInt(contentLength, 10) > maxSize
```
dholms, line 201: doesn't need to hold up this PR, but a bad client could send up fake content-lengths right? is there something we can stick on the stream to bail out & throw an error if it goes over the purported length?

dholms (approved):
awesome 💪 😎

packages/xrpc-server/src/util.ts
devinivy: Yep yep, you're dead on.  There is a way we could turn that into an error.  There's a similarish case here where a client could send up bad data by saying it's gzipped, but it's actually not— that would cause an error to be emitted from the gzip decoder stream that should turn into a 400.  The issue I am a bit stuck on is that there's not a great way for these errors to surface properly right now— I believe it would become an unhandled exception.  So I can get that error going, but I think there's a bigger question here about how to catch errors off the stream and turn them into appropriate responses.

devinivy: I will at minimum peel this off into an issue prior to merging 👍

dholms: yeah i think we'd want to emit the error on the stream that we pass to the event handler 🤔

devinivy: This didn't end-up being too bad 👍 Should be handled now!

packages/xrpc-server/src/util.ts
```diff
@@ -203,17 +204,37 @@ function decodeBodyStream(
...
+  if (maxSize !== undefined) {
+    const maxSizeChecker = new MaxSizeChecker(maxSize)
+    forwardStreamErrors(stream, maxSizeChecker)
+    stream = stream.pipe(maxSizeChecker)
+  }
+
+  return stream
+}
+
+class MaxSizeChecker extends Transform {
```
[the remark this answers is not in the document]
devinivy: Sure thing!

dholms (approved):
:shipit:

pfrazee:
Nice

PR #397, opened 2022-12-09
Banners

dholms:
Adding in banners, handled in a similar way to avatars.

Took the time to make a couple changes to lexicon codegen namely:
- renaming `lexicons` -> `schemas`
- adding in a `schemaDict` that is keyed similarly to ids
- adding in `lexicons` var which is just the `Lexicons` object that you can actually run validation against

This let me simplify a bit of the code that was consuming it & not have to copy/paste blob constraints in the repo processor.

Note: ended up stripping out the post media changed & will get those up in a subsequent PR shortly


lexicons/app/bsky/actor/profile.json
```diff
@@ -23,6 +23,13 @@
             "maxWidth": 500,
             "maxHeight": 500,
             "maxSize": 300000
+          },
+          "banner": {
+            "type": "image",
+            "accept": ["image/png", "image/jpeg"],
+            "maxWidth": 1500,
+            "maxHeight": 500,
+            "maxSize": 1000000
```
pfrazee, line 32: Tbh 400 or 500kb should be plenty here. 300kb is arguably enough.

lexicons/app/bsky/feed/post.json
```diff
@@ -15,6 +15,11 @@
             "items": {"type": "ref", "ref": "#entity"}
           },
           "reply": {"type": "ref", "ref": "#replyRef"},
+          "media": {
```
pfrazee, line 18: We should chat through this, originally (IIRC) this was what `embeds` were meant to be for

dholms: oh it totally was. i yoloed this too quick lol

pfrazee:
Looks great

devinivy (approved):
Oh yeah!

PR #403, opened 2022-12-12
Streaming: decouple consumers from the message queue

devinivy:
This decouples the PDS stream consumers from the message queue by introducing a `listen()` interface on the queue, and a `Consumer` interface that can listen to certain events from the queue.


packages/pds/src/db/message-queue/index.ts
```diff
@@ -135,195 +130,18 @@ export class SqlMessageQueue implements MessageQueue {
...
-            count: 1,
-            postedTrending: 0 as 0 | 1,
-          })),
-        )
-        .execute()
+    for (const listener of listeners) {
+      try {
+        await listener({ db, message }) // @TODO returned side effects
+      } catch (err) {
+        log.error({ message, err }, `unable to handle event: ${message.type}`)
```
dholms, line 142: future work: we should have an "unhandled messages" container of some sort

packages/pds/src/stream/types.ts
```diff
@@ -0,0 +1,25 @@
...
+  type: T
+  [s: string]: unknown
+}
+
+export type Listener<M extends MessageOfType = MessageOfType> = (ctx: {
+  db: Database
+  message: M
+}) => Promise<void | MessageOfType[]>
+
+export type Listenable<M extends MessageOfType = MessageOfType> =
```
dholms, line 13: i see you went to some extra ends to differentiate Listeners & Consumers, but it looks like we're only using Consumers at this point. Just out of curiosity, what do you have in mind for non-consuming listeners?

dholms (approved):
this all makes sense. consumers as an unit makes sense & will keep the processing much more understandable 👌

packages/pds/src/stream/types.ts
dholms: & kinda wondering if we can simplify some of the `Listenable` song & dance (checking two locations, etc). but not overly worried about it, cause I can tell you have a plan w it 👌

devinivy: TBH I was a little bit in library-writing mindset, and some of this was about conveniences that we don't actually need!  To sort of split the difference I am going to keep the `Listenable` interface separate from the `Consumer` class, but nix the ability to pass a listener callback as a listenable.  That should be a little simpler :)

PR #404, opened 2022-12-12
Streaming: support multi-stage processing

devinivy:
This lightly introduces the idea of multi-stage stream processing in the PDS by allowing queue consumers to return effects in the form of additional messages to be processed.


packages/pds/src/stream/consumers/add-upvote.ts
```diff
@@ -34,74 +25,12 @@ export default class extends Consumer<AddUpvote> {
             did: scene,
             subject: message.subject,
             count: 1,
-            postedTrending: 0 as 0 | 1,
+            postedTrending: 0 as const,
```
dholms, line 28: ah nice one forgot about this 👌

PR #405, opened 2022-12-12
A couple lexicon-related fixes

### LINKED ISSUE #391: Lex-cli generates wrong type for encoding '*/*'
It treats that as a constant instead of an open string.

pfrazee:
- Fix the generated procedure call signature when an encoding is `*/*` (close #391)
- Fix the named error generation for `updateProfile()`
- Add a test for named error generation

EDIT:
- Add more named blob errors and make the blob error messages more human-friendly
  - The "friendlier" descriptions are so that I can just display that error directly. The other option would be to make XRPC errors more sophisticated so that they can carry data like `blobSize` and `maxBlobSize`, enabling the client to render its own error message.

### DIFF
--- a/lexicons/app/bsky/actor/updateProfile.json
+++ b/lexicons/app/bsky/actor/updateProfile.json
@@ -49,7 +49,10 @@
         }
       },
       "errors": [
-        {"name": "InvalidBlob"}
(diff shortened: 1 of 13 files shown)


devinivy (approved):
Killer 👍

PR #406, opened 2022-12-12
Fix size/url for profile banner

devinivy:
Small fix to profile banner size: height and width were mixed-up! 🙃


dholms (approved):
lg ✨ 

sorry everyone 😛

PR #407, opened 2022-12-12
Downgrade file-type dep version, fix build

devinivy:
The file-type package has become ESM-only, which throws a wrench in the build.  I downgraded to the most recent non-ESM version of the package.  Need to take a closer look at why the build completes successfully despite technically being broken (i.e node wont be able to run it).


pfrazee:
lgtm

PR #409, opened 2022-12-13
Update example code in xrpc and xrpc-server READMEs

snarfed:
* addSchema => addLexicon
* update lexicons for new defs.main structure
* update server ping method implementation to match XRPCHandler type signature

Hi all! First off, I know this is all under active development on your side. Hope you don't mind this PR. Should hopefully be straightforward and non-controversial, but if triaging it is more pain that it's worth, feel free to ignore it.

I've been writing my own Lexicon/XRPC implementation, https://github.com/snarfed/lexrpc , purely for fun, as part of a bigger project. I tested it against the xrpc and xrpc-server packages in this repo today, and got it working, but the example code in those packages' READMEs needed these tweaks.

### DIFF
--- a/packages/xrpc-server/README.md
+++ b/packages/xrpc-server/README.md
@@ -4,26 +4,37 @@
 
 ```typescript
 import * as xrpc from '@atproto/xrpc-server'
+import express from 'express'
 
 // create xrpc server
 const server = xrpc.createServer([{
-    xrpc: 1,
+    lexicon: 1,
     id: 'io.example.ping',
-    type: 'query',
-    parameters: { message: { type: 'string' } },
-    output: {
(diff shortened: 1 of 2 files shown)


dholms:
Hey! 👋 

Great to see that you're writing an implementation of lexicon/xrpc! & hope the churn on our side isn't making it too difficult for you 😅  

As I'm sure you noticed, we just wrapped up a pretty massive rewrite of lexicon, but the dust is mostly settled & it should be more stable from here on out. Docs have lagged behind a bit & we'll need to bring those up to date.

Your suggestions here all check out, so I'll get them approved & merged

dholms (approved):
lgtm 👍

PR #413, opened 2022-12-14
Test rebuilding pds timeline indexes from repos

devinivy:
This is an early proof of concept to show that we can rebuild PDS indexes directly from repositories.


dholms (approved):
nicee this is pretty cool 👍

PR #414, opened 2022-12-14
Streaming: decouple producer side more, refactor db class

devinivy:
A few things going on here.  This started by me trying to decouple the message queue producer side from other functionality of the app.  In particular, the circular dependency between `Database` and `SqlMessageQueue` was of interest.

The main source of the little tangle there is that record indexing functionality (e.g. `indexRecord()`, `deleteRecord()`) lived on the database class, and they needed to work with the queue.  So I started by peeling record indexing methods off of the database and into their own class that would be instanced with `db` and `messageQueue`.

That lead me to peel off other methods that lived on the database class that related to actor and repo functionality.  I felt like the record, actor, and repo functionality needed to be reorganized into some common type of "thing" so I introduced the idea of services which are containers for business logic/transactions from different domains in the application.  This left the database class focused on just generic "database stuff" like setting-up connections and migrations, and nothing application-specific, which felt pretty good.

After setting-up a record, actor, and repo service from the functionality that previously lived directly on the database, I went another step and unified repo processing transactions from src/repo/ into the new repo service.  And that's pretty much where things stand!

This work is in a complete state, but throwing it up as a draft just to get feedback on the approach, since it's a bit opinionated and affects how code would be organized moving forward.


packages/pds/src/db/database-schema.ts
```diff
@@ -21,10 +21,10 @@ import * as trend from './tables/trend'
 import * as follow from './tables/follow'
 import * as blob from './tables/blob'
 import * as repoBlob from './tables/repo-blob'
-import * as messageQueue from './message-queue/tables/message-queue'
-import * as messageQueueCursor from './message-queue/tables/message-queue-cursor'
-import * as sceneMemberCount from './message-queue/tables/scene-member-count'
-import * as sceneVotesOnPost from './message-queue/tables/scene-votes-on-post'
+import * as messageQueue from './tables/message-queue'
```
dholms, line 24: 👍  yea had been meaning to move these

packages/pds/src/services/record.ts
```diff
@@ -0,0 +1,181 @@
...
+import Database from '../db'
+import * as Declaration from '../db/records/declaration'
+import * as Post from '../db/records/post'
+import * as Vote from '../db/records/vote'
+import * as Repost from '../db/records/repost'
+import * as Trend from '../db/records/trend'
+import * as Follow from '../db/records/follow'
+import * as Assertion from '../db/records/assertion'
+import * as Confirmation from '../db/records/confirmation'
+import * as Profile from '../db/records/profile'
```
dholms, line 15: may make sense for all this code to live in `services/record`

packages/pds/src/db/index.ts
```diff
@@ -2,62 +2,18 @@ import assert from 'assert'
...
-import { User } from './tables/user'
 import { dummyDialect } from './util'
 import * as migrations from './migrations'
 import { CtxMigrationProvider } from './migrations/provider'
-import { DidHandle } from './tables/did-handle'
-import { Record as DeclarationRecord } from '../lexicon/types/app/bsky/system/declaration'
-import { APP_BSKY_GRAPH } from '../lexicon'
-import { MessageQueue } from './types'
 
 export class Database {
```
dholms, line 10: With this class so stripped down now & most application logic moved to the services, would it make sense to have this `extend Kysely<DatabaseSchema>`?

This would keep us from having to do the `db.db` nonsense that we do everywhere.

Really this class serves to:
- give static methods for creation
- simplify the tx flow a bit & manage some state about if we're in a tx or not
- give migration helpers

I think we could just ditch the `close` fn in favor of Kysely's `destroy`

packages/pds/src/api/app/bsky/feed/setVote.ts
```diff
@@ -12,13 +12,14 @@ export default function (server: Server) {
...
       const { subject, direction } = input.body
-      const { db } = locals.get(res)
+      const { db, services } = locals.get(res)
 
       const requester = auth.credentials.did
       const authStore = await locals.getAuthstore(res, requester)
       const now = new Date().toISOString()
 
       const voteUri = await db.transaction(async (dbTxn) => {
+        const repoTxn = services.repo.using(dbTxn)
```
dholms, line 22: I don't think this is bad, but just thinking through possible footguns. I feel like it's very easy to forget to do `.using(tx)` & accidentally fire off non-tx queries from within a tx.

Almost wonder if we should not pass around services that are hooked up to the DB & always make the route instantiate them (either with a non-tx db or with a dbTxn).

Nearly all of our logic is held in txns anyway so i don't think this would increase boilerplate code much

dholms:
very very much support this reorg 👍 

The `Database` class was getting clogged up & has needed this for a while & this is coming together very cleanly

packages/pds/src/index.ts
```diff
@@ -19,9 +19,10 @@ import { ServerConfig, ServerConfigValues } from './config'
 import { Locals } from './locals'
 import { ServerMailer } from './mailer'
 import { createTransport } from 'nodemailer'
-import SqlMessageQueue from './db/message-queue'
+import SqlMessageQueue from './stream/message-queue'
```
dholms, line 22: just a nit, but I kinda want a different name for this than `stream` that is more descriptive of its function. `stream` just makes me think of data streams not event processing

packages/pds/src/db/index.ts
devinivy: Love this idea, and sadly striking out on it right now.  I'd like to came back to it, though.  The issue is that I'm not coming up with a straightforward way to get an instance of our extended class for transactions.  I.e. hard to upgrade the `txn` that comes out of kysely here: `this.transaction().execute((txn) => {`, which would mess with some things like checking the db dialect in a transaction, asserting that the instance is a transaction, etc.

I think it's figureoutable, just might be a bit more involved than I can pull off with the other stuff going on in this particular PR.

dholms: cool cool no worries, it's not pressing 👌

devinivy:
@dholms I think that feedback should be addressed now.

packages/pds/src/api/com/atproto/password-reset.ts
```diff
@@ -42,7 +42,7 @@ export default function (server: Server) {
       return createInvalidTokenError('Malformed token')
     }
 
-    const user = await db.getUser(did)
+    const user = await services.actor(db).getUser(did)
```
dholms, line 45: really like how this reads

dholms (approved):
yup this is excellent 🙏 really liking how this is all shaping up

PR #421, opened 2022-12-15
Message queue blocking and graceful teardown

devinivy:
Fixes a bad case with the message queue, where processing gets backed-up and ties-up db connections.  This is achieved by limiting concurrency of `processNext()` in-process rather than lean on database locks to do the same job: without this many pending `processNext()`s could each be waiting and holding onto an item in the db connection pool, which would start to affect the xrpc side of things.

A nice side-effect is that we also get a much more graceful `destroy()` of the message queue, where we can stop accepting new processing and drain any pending items.


dholms (approved):
nice i like it 👍 

im also not opposed to taking this the direction you brought up a little bit back of ditching in-process hints for "processing next" & just polling instead

devinivy:
@dholms Cool!  I think having this little structure in here is good either way since it gives us a way to manage and introspect in-flight work.  I have been thinking about alternative mechanisms of keeping the queue moving, and if we ever move to use `notify` for postgres, I expect this would be useful there too.

I am also pretty happy with how things currently function in sqlite with its single concurrent connection, where the queue in the test is being consumed quite eagerly but still making room for other queries!

PR #423, opened 2022-12-15
Refactor feeds & post threads

dholms:
Refactor feeds (`getTimeline` & `getAuthorFeed`) and post threads (`getPostThread`)

Factor out a new `app.bsky.feed.post#view` schema that is used by all three xrpc schemas. As well as a `app.bsky.feed.feedViewPost` schema used by both `getTimeline` & `getAuthorFeed`.

This builds on @devinivy's work with splitting out database logic to services, introduces a new feed service & breaks up the queries from one monolithic query to "data" & "presentation" queries.

This will cause breaking changes to the client. We should discuss how we want to go about deploying that & coordinate to do so. Probably hold off on merging until we're ready.

⚠️  Note: I tried my darndest to compare jest snapshots & make sure I didn't do anything too screwy. I found some bugs and fixed them, but these are roouughhh to compare. We'll want to test this in staging to make sure it works as expected.


pfrazee:
SGTM, but might bikeshed the new names a bit. `feed.item` and `feed.myState` -- should we get more specific? Is there some word we can settle on that's better than "presented" for these kinds of things? Maybe "view?" `feed.viewItem`? Not sure

dholms:
good spot for bikeshedding lol

yeah I don't hate `viewItem` 🤔

Maybe `interactionState` instead of `myState`? or do we want to tip off in the name that it is meant for views

pfrazee:
Yeah maybe it's

- `app.bsky.feed.viewItem`
- `app.bsky.feed.viewItem#interactionState`

And then do we want to s/presented/view/ in the other stuff? `app.bsky.embed.image#view` ?

dholms:
I'm good with that 👍 

That was actually how i originally did this, but then broke them into separate schemas because, for instance, getPostThread uses `myState` & not `feed.item`

devinivy:
If we're coming-up with a scheme for naming lexicons presented through xrpc, two considerations I'd throw in the mix:
 - A given lexicon might not just have one "view."  It's easy to imagine, for example there being a "list view" and "detail view" for the same entity.
 - It may not always be obvious which lexicon to tack a view onto.  I suppose we already have an illustration of that with `app.bsky.embed.image#presented` (presenting `app.bsky.embed.image`) versus `app.bsky.feed.item` (which is its own thing, not a part of any records).

I also like the idea of having a way to call out which things are views versus stuff that may appear directly in a record 👍

devinivy:
Another extremely bikesheddy/intrusive thought that I'd be happy to let go of— do we feel "state" is pulling its weight in the names `myState`/`interactionState`?  It's just one of those generic terms like "data" that I tend to take a second look at when I use it to name something.

devinivy:
Alright, alright sorry for all the posting 😅   But I did some thinking and wanted to followup with something more concrete.

The github gql API carves out personalized info with the prefix `viewer`, e.g. `repository.viewerHasStarred`.  I think that is pretty good.  Also keeps it "third-person enough" that writing code feels natural.  E.g. as the coder, you don't have to pretend to be the user by writing something like `item.my.repost`.  Instead it might be something like `item.viewerRepost` or `item.viewer.repost`.

Beyond that, I like Paul's suggestion above pretty well since it has conventions for both views that appear as a variant of a record schema and views in their own lexicon schema.  Just might recommend taking into account when naming a def `#view` that down the road there may also be a `#viewDetail`.

pfrazee:
"viewer" is definitely a better semantic than "my"

pfrazee:
👍 Yeah looks great. I see what you mean on the snapshot diffs.

After we merge, I'll test on local, then coordinate with devin to test on staging and then deploy the app update concurrently with prod.

packages/pds/src/api/app/bsky/feed/getPostThread.ts
```diff
@@ -53,169 +50,134 @@ export default function (server: Server, ctx: AppContext) {
...
+    actors,
+    posts,
+    embeds,
+  )
+
+  let parent
+  if (threadData.parent) {
+    if (threadData.parent instanceof ParentNotFoundError) {
+      parent = {
+        $type: 'app.bsky.feed.getPostThread#notFoundPost',
```
devinivy, line 71: If forget, are there codegen constants kicking around for these that we can use?

packages/pds/src/api/app/bsky/feed/getTimeline.ts
```diff
@@ -21,167 +14,49 @@ export default function (server: Server, ctx: AppContext) {
...
-      let postsQb = db
-        .selectFrom('post')
-        .select([
-          sql<FeedItemType>`${'post'}`.as('type'),
-          'uri as postUri',
-          'cid as postCid',
-          'creator as originatorDid',
-          'indexedAt as cursor',
-        ])
+      const feedService = ctx.services.feed(ctx.db, ctx.imgUriBuilder)
```
devinivy, line 21: 👍 nice to see the new service parameter fitting in without too much trouble.

packages/pds/src/api/app/bsky/feed/getTimeline.ts
```diff
@@ -21,167 +14,49 @@ export default function (server: Server, ctx: AppContext) {
...
       feedItemsQb = paginate(feedItemsQb, {
         limit,
         before,
         keyset,
       })
-
-      const queryRes = await feedItemsQb.execute()
-      const feed = await composeFeed(db, ctx.imgUriBuilder, queryRes)
+      const feedItems: FeedRow[] = await feedItemsQb.execute()
+      const feed = await composeFeed(feedService, feedItems, requester)
```
devinivy, line 53: I think this is the way to do it 👍 much nicer.

packages/pds/src/api/app/bsky/feed/getPostThread.ts
```diff
@@ -1,50 +1,47 @@
...
-        )
-      }
-
+      const thread = composeThread(
+        threadData,
+        feedService,
+        posts,
+        actors,
+        embeds,
+      )
```
devinivy, line 44: It's kind of a pain but I think this is the way to do it—much nicer 👍

packages/pds/src/services/feed/index.ts
```diff
@@ -0,0 +1,256 @@
...
+import { View as PostView } from '../../lexicon/types/app/bsky/feed/post'
+import { ActorViewMap, FeedEmbeds, PostInfoMap, FeedItemType } from './types'
+
+export * from './types'
+
+export class FeedService {
+  constructor(public db: Database, public imgUriBuilder: ImageUriBuilder) {}
+
+  static creator() {
+    return (db: Database, imgUriBuilder: ImageUriBuilder) =>
```
devinivy, line 16: If you wanted to, you could probably hoist this up like so:
```suggestion
  static creator(imgUriBuilder: ImageUriBuilder) {
    return (db: Database) =>
```

devinivy (approved):
I dig!

packages/pds/src/api/app/bsky/feed/getTimeline.ts
dholms: yea really slick 👍

packages/pds/src/api/app/bsky/feed/getPostThread.ts
dholms: I don't think so. They haven't made it into the genned `ids` map yup. but we should have that in mind. i'll peel off an issue

packages/pds/src/services/feed/index.ts
dholms: oh i like that 👍 
didn't see til not that you'd done that with messageQueue & blobstore

packages/pds/src/api/app/bsky/feed/getPostThread.ts
dholms: yeah agreed. was a bit of a pain to rewrite, but much nicer on the other side

PR #424, opened 2022-12-16
App context

dholms:
Make the PDS a class that we can instantiate.

Switch out `Locals` for `AppContext` which we pass around through methods with closures.

Snuck a couple other changes in here that I'll point out in comments


packages/pds/src/db/database-schema.ts
```diff
@@ -54,4 +55,6 @@ export type DatabaseSchema = user.PartialDB &
   sceneMemberCount.PartialDB &
   sceneVotesOnPost.PartialDB
 
+export type DatabaseSchema = Kysely<DatabaseSchemaType>
```
dholms, line 58: We use this enough that I just decided to export it as the `Kysely` wrapped type.

I had to introduce (the somewhat sloppily named 😛 ) `DatabaseSchemaType` for a few cases that the underlying type

packages/pds/src/api/com/atproto/util/auth.ts
```diff
@@ -1,24 +0,0 @@
-import { RefreshToken } from '../../../../auth'
-import Database from '../../../../db'
-
-export const grantRefreshToken = async (
```
dholms, line 4: snuck these into their own auth service

packages/pds/src/context.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  get keypair(): DidableKey {
+    return this.opts.keypair
+  }
+
+  get auth(): auth.ServerAuth {
+    return this.opts.auth
+  }
+
+  get accessVerifier() {
```
dholms, line 43: One nice thing about ctx, is sticking smart getters on them like this 👌

packages/pds/src/bin.ts
```diff
@@ -45,8 +45,8 @@ const run = async () => {
     blobstore = new MemoryBlobStore()
   }
 
-  const { listener } = server(db, blobstore, keypair, cfg)
-  listener.on('listening', () => {
+  const pds = PDS.create({ db, blobstore, keypair, cfg })
+  pds.server.on('listening', () => {
```
dholms, line 49: Any thoughts on making `PDS.create` async & only returning after this is listening?

devinivy: Listening is just such an effectful thing— I think it can be useful to have a way to create the server without listening, mostly for testing or using it "headlessly" as a library rather than a web service.  To take an example, in hapi there's `initialize()` which revs everything up and ensures the server is complete/healthy, and then `start()` which does all of that but additionally listens.  Drawing that analogy roughly over to the pds, you might have something like:

E.g.
```js
const app = PDS.create(/* ... */) // Creates app inertly
await app.initialize() // Ensures db connections look good, message queue might start processing
const { db, services } = app
await services(db).actor().getUser('x') // This works now
await app.start() // App is accepting traffic over http, could also call without initialize() first
```

dholms: ah nice i quite like the sound of that 👌

packages/pds/src/db/database-schema.ts
```diff
@@ -25,8 +25,9 @@ import * as messageQueue from './tables/message-queue'
 import * as messageQueueCursor from './tables/message-queue-cursor'
 import * as sceneMemberCount from './tables/scene-member-count'
 import * as sceneVotesOnPost from './tables/scene-votes-on-post'
+import { Kysely } from 'kysely'
```
devinivy, line 28: A nit (you're welcome to ignore ✨ )— in the pds external dependencies seem to generally be imported above local dependencies by convention.

devinivy (approved):
Super pumped about this 💎 💎

packages/pds/src/index.ts
```diff
@@ -6,117 +6,144 @@ import 'express-async-errors'
...
+  }
 
-  app.use((req, res, next) => {
-    const reqLocals: Locals = {
-      ...locals,
-      logger: req.log, // This logger is request-specific
+  async destroy(force = false): Promise<void> {
+    await this.ctx.messageQueue.destroy()
+    if (this.server) {
+      const terminator = createHttpTerminator({
```
devinivy, line 139: I think that it's possible the terminator needs to be created prior to accepting connections, but I'm not positive— just opened an issue to ask https://github.com/gajus/http-terminator/issues/42

packages/pds/src/index.ts
```diff
@@ -6,117 +6,144 @@ import 'express-async-errors'
...
+    const server = this.app.listen(this.ctx.cfg.port)
+    this.server = server
+    return new Promise((resolve, reject) => {
+      server.on('listening', () => {
+        resolve(server)
+      })
+      server.on('error', (err) => {
+        reject(err)
+      })
+    })
```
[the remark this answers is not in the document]
dholms: oh nice was not aware of this 👌

packages/pds/src/index.ts
```diff
@@ -6,117 +6,143 @@ import 'express-async-errors'
...
+  }
 
-  app.use((req, res, next) => {
-    const reqLocals: Locals = {
-      ...locals,
-      logger: req.log, // This logger is request-specific
+  async destroy(): Promise<void> {
+    await this.ctx.messageQueue.destroy()
+    if (this.terminator) {
+      await this.terminator.terminate
```
devinivy, line 142:
```suggestion
      await this.terminator.terminate()
```

dholms: 😱

PR #428, opened 2022-12-16
Support concurrent event processing by topics in the pds

devinivy:
This is in pretty good shape, but I am leaving it as a draft for the time being since I am able to write a stress test that cause some minor correctness issues— not quite ready for prime time.

 - Occasional deadlocks on postgres when stressed.  I think these are acceptable, just need to ensure that the cursor doesn't move forward in this case, any maybe eagerly retry.
 - I've seen a message get process multiple times in on postgres as well (only in the stress test).  This seems to occur only for very first messages, so it's probably a race related to `ensureCursor()` to tidy-up.
 - I've seen the test complete without processing all messages (only in the stress test).  As far as I can tell the cursors still land in a consistent state, the `processNext()` hints just didn't cause everything to get processed.  I don't think this is much of an issue since that is what `ensureCaughtUp()` anticipates, anyway.  But we do need to write the test to be aware of this.

This is still good to be reviewed 👍


devinivy:
This is quite stale, so I'm closing this.  I will keep the branch around and we can re-open if/when this is better finalized.

PR #429, opened 2022-12-19
app.bsky.feed.post lexicon: add textSlice description

snarfed:
Hi again, all! Really minor suggestion, I was working with this lexicon today and wanted to double check my understanding of the textSlice endpoint handling. Feel free to correct me if I'm wrong!


pfrazee:
Hey snarf! That's accurate. SMH at myself for not doing length instead. Thanks for the PR!

PR #435, opened 2022-12-21
Create PDS with ServerConfig rather than ServerConfigValues

devinivy:
Just a small change to simplify the API of `PDS.create()`, and make it long-term compatible with `ServerConfig.fromEnv()`.


dholms (approved):
bueno 👍

PR #436, opened 2022-12-21
Smarter repo storage

dholms:
We're moving away from a simple blockstore for repo storage. This is the first step in making the storage layer smarter & giving it more responsibility. Repo storage now has a concept of commits, repo history & the current HEAD of the repo.


packages/pds/src/services/repo/index.ts
```diff
@@ -129,24 +76,16 @@ export class RepoService {
...
-      .stageUpdate(writeOps)
-      .createCommit(authStore, async (prev, curr) => {
-        const success = await this.updateRepoRoot(did, curr, prev, now)
-        if (!success) {
-          throw new Error('Repo root update failed, could not linearize')
-        }
-        return null
-      })
+    const repo = await Repo.load(storage, currRoot)
+    const updated = await repo.stageUpdate(writeOps).createCommit(authStore)
```
devinivy, line 88: 💎

packages/repo/src/util.ts
```diff
@@ -17,61 +17,39 @@ export const ucanForOperation = async (
...
-  return null
-}
-
 export const getWriteOpLog = async (
-  blockstore: IpldStore,
-  earliest: CID | null,
+  storage: RepoStorage,
   latest: CID,
+  earliest: CID | null,
 ): Promise<RecordWriteOp[][]> => {
```
devinivy, line 24: I take it this is what bonked the pds sync test 😆

packages/repo/src/verify.ts
```diff
@@ -52,11 +51,15 @@ export const verifyUpdates = async (
       nextRepo.commit.sig,
     )
     if (!validSig) {
-      throw new Error(`Invalid signature on commit: ${nextRepo.cid.toString()}`)
+      throw new RepoVerificationError(
```
devinivy, line 54: I like carving out the new error type for these 👍

packages/repo/src/util.ts
dholms: haha no it was actually the recursive sql query which returned
latest -> earliest when i expected the other way

packages/pds/src/sql-repo-storage.ts
```diff
@@ -0,0 +1,200 @@
...
+      }
+    }
+    const found = await builder.executeTakeFirst()
+    return found ? CID.parse(found.root) : null
+  }
+
+  async getSavedBytes(cid: CID): Promise<Uint8Array | null> {
+    const found = await this.db.db
+      .selectFrom('ipld_block')
+      .where('cid', '=', cid.toString())
```
devinivy, line 36: Should `getSavedBytes()` and `hasSavedBytes()` be restricted by `this.did`?

packages/pds/src/sql-repo-storage.ts
```diff
@@ -0,0 +1,200 @@
...
+      .values({
+        cid: cid.toString(),
+        did: this.did,
+      })
+      .onConflict((oc) => oc.doNothing())
+      .execute()
+    await Promise.all([insertBlock, insertCreator])
+  }
+
+  async putMany(toPut: Map<string, Uint8Array>): Promise<void> {
```
devinivy, line 74: This will hit a limit on the size of `toPut` based on the pg/sqlite max parameter counts.  If it's a concern, a common workaround is to perform batches of inserts.

packages/pds/src/sql-repo-storage.ts
```diff
@@ -0,0 +1,200 @@
...
+      })
+      .onConflict((oc) => oc.doNothing())
+      .execute()
+    await Promise.all([
+      insertBlocks,
+      insertCommit,
+      updateRoot,
+      insertCommitHistory,
+    ])
+    this.clearStaged()
```
devinivy, line 135: I assume it's cool for the blocks to remain staged even if this promise rejects, just flagging for a second look.

packages/pds/src/sql-repo-storage.ts
```diff
@@ -0,0 +1,200 @@
...
+              .if(earliest !== null, (qb) =>
+                // @ts-ignore
+                qb.where('commit.commit', '!=', earliest?.toString() as string),
+              ),
+          ),
+      )
+      .selectFrom('ancestor')
+      .select('commit')
+      .execute()
+    return res.map((row) => CID.parse(row.commit)).reverse()
```
devinivy, line 192: This is pretty slick!

packages/pds/src/sql-repo-storage.ts
```diff
@@ -0,0 +1,200 @@
...
+  }
+
+  async destroySaved(): Promise<void> {
+    throw new Error('Destruction of SQL repo storage not allowed at runtime')
+  }
```
devinivy, line 197: I think we may actually need this fairly soon.  I could still imagine it not living on this class, though.

devinivy (approved):
Very nice, these interfaces are just getting more and more intuitive.

packages/pds/src/sql-repo-storage.ts
dholms: yeah that's a good idea

packages/pds/src/sql-repo-storage.ts
dholms: yeah good point. we shouldn't be hitting that ever but doesn't hurt to add in

packages/pds/src/sql-repo-storage.ts
dholms: yeah i think you'd want them too, because you may want to retry
that being said, my next PR takes a diff approach & removes staged blocks entirely

packages/pds/src/sql-repo-storage.ts
```diff
@@ -0,0 +1,200 @@
...
+  private getTimestamp(): string {
+    return this.timestamp || new Date().toISOString()
+  }
+
+  async getCommitPath(
+    latest: CID,
+    earliest: CID | null,
+  ): Promise<CID[] | null> {
+    const res = await this.db.db
+      .withRecursive('ancestor(commit, prev)', (cte) =>
```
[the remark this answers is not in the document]
dholms: my first every recursive query 🕺 
kinda fun to put together & simple enough that im not too worried about it

devinivy (approved):

PR #437, opened 2022-12-21
Api package fixes

pfrazee:
- Updates the API package to properly publish
- Bumps `@atproto/api` to 0.0.2

We'll need to make these updates to all the packages but I'm leaving that for now.

### DIFF
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -7,6 +7,9 @@
     "build": "node ./build.js",
     "postbuild": "tsc --build tsconfig.build.json",
     "update-main-to-dist": "node ./build.js --update-main-to-dist",
+    "update-main-to-src": "node ./build.js --update-main-to-src",
+    "prepublish": "npm run update-main-to-dist && npm run build",
+    "postpublish": "npm run update-main-to-src",
     "test": "jest"
   },
   "license": "MIT",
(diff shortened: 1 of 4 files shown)


packages/api/package.json
```diff
@@ -1,12 +1,15 @@
...
   "name": "@atproto/api",
-  "version": "0.0.1",
+  "version": "0.0.2",
   "main": "src/index.ts",
   "scripts": {
     "codegen": "lex gen-api ./src/client ../../lexicons/com/atproto/*/* ../../lexicons/app/bsky/*/*",
     "build": "node ./build.js",
     "postbuild": "tsc --build tsconfig.build.json",
     "update-main-to-dist": "node ./build.js --update-main-to-dist",
+    "update-main-to-src": "node ./build.js --update-main-to-src",
```
dholms, line 10: we may need to rethink how these scripts work. we probably don't need to rebuild on `postpublish` & `prepublish` doesn't build types so those could get out of sync with the built code

don't think it needs to be polished yet, but maybe we open an issue so we don't lose track

dholms (approved):
yup 👍

devinivy (approved):

packages/api/package.json
pfrazee: Good point, I'll open that issue

Issue #438, opened 2022-12-21
Finalize the publish flows for packages


pfrazee:
https://github.com/bluesky-social/atproto/pull/437 lays out the steps needed on each package. As @dholms points out, we also need to rethink the scripts that update the "main" field in package.json since it currently runs a buildstep (incorrectly).

frank-dspeed:
you do not need to care for packaging most best would be if you could avoid packaging at all that would save me from the ts babel step to clean your builds. 

my advice would be use directly .js and then tsconfig allowJs checkJs true as also if needed which should not often be the case type antotate via JSDOC ```/** @type {} */ ```

## Benefits
I can more easy depend on your core project and your users will also benefit from that i am in fact at present 
packaging your code as so called stealify language modules 

a new module package component format based on cutting edge web tech. 

would suggest usage of better-doc as incremental step to get the jsdoc definitions relativ automated and then later hand optimize:

https://github.com/SoftwareBrothers/better-docs

PR #439, opened 2022-12-22
Api package fixes 2

pfrazee:
It _looks_ like running build in prepublish caused the d.ts files to get stripped, so I went ahead and separated out the update-pkg script. We could probably move that script into the root of the package but it's safer here when it's working off __dirname (no risk of using cwd and getting the wrong package json).

Related to https://github.com/bluesky-social/atproto/issues/438

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -21,6 +21,7 @@
     "jest.config.js",
     "babel.config.js",
     "build.js"
+    "update-pkg.js"
   ],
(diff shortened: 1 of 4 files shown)


devinivy (approved):
Yep!

devinivy (approved):

dholms (approved):
nice 👍

PR #440, opened 2022-12-22
Moderation action for actor takedown

devinivy:
This is the crudest, server-level moderation control: banning an actor from the server.  It takes the form of a soft delete on the actor's profile and posts.  Moderation actions are modeled here so that they are logged and reversible: there a new method to take a moderation action, as well as an action to reverse a previous moderation action.  Also found a little bug in xrpc-server that's fixed here.


lexicons/app/bsky/administration/moderationAction.json
```diff
@@ -0,0 +1,21 @@
+{
+  "lexicon": 1,
+  "id": "app.bsky.administration.moderationAction",
```
pfrazee, line 3: Here comes the bikeshed!

Any reason we can't shorten to `.admin.` ?

lexicons/app/bsky/administration/moderationAction.json
```diff
@@ -0,0 +1,21 @@
...
+  "defs": {
+    "view": {
+      "type": "object",
+      "required": ["id", "action", "subject", "rationale", "createdBy", "createdAt"],
+      "properties": {
+        "id": {"type": "integer"},
+        "action": {"type": "string", "knownValues": ["takedown"]},
```
pfrazee, line 10: Do we want to use nsids for the actions?

lexicons/app/bsky/administration/moderationAction.json
```diff
@@ -0,0 +1,21 @@
...
+        "subject": {"type": "union", "refs": ["app.bsky.actor.ref"]},
+        "rationale": {"type": "string"},
```
pfrazee, line 12: Boy I'm going to bikeshed all over this. Why rationale vs reason?

lexicons/app/bsky/administration/moderationAction.json
```diff
@@ -0,0 +1,21 @@
...
+        "createdBy": {"type": "string"},
+        "createdAt": {"type": "string"},
+        "reversedBy": {"type": "string"},
+        "reversedAt": {"type": "string"},
+        "reversedRationale": {"type": "string"}
```
pfrazee, line 17: I'm the worst. Any way we could put these reversed fields into a subobject that's optional, but then requires all the inner fields?

lexicons/app/bsky/administration/takeModerationAction.json
```diff
@@ -0,0 +1,30 @@
+{
+  "lexicon": 1,
+  "id": "app.bsky.administration.takeModerationAction",
```
pfrazee, line 3: maybe just `moderate()`?

pfrazee (commented):
Excellent start, so glad you dove into this like you did. I'm sorry for bikeshedding all over it (especially given the move-fast prerogative).

From here the two things I'm inclined to sort out are some mechanics and then the next capabilities we need. For example:

- What other kinds of actions do we expect than "takedown"
- (naturally) is "takedown" the right term for this action
- What other admin routes do we need? (I'm gathering lots of requirements as I work on the screen, things like "enumerate server accounts")
- How are "admin" permissions being handled?

So I figure we can discuss these next week.

lexicons/app/bsky/administration/moderationAction.json
devinivy: Sounds good 👍

lexicons/app/bsky/administration/moderationAction.json
devinivy: Sure!  P.S. the other actions I imagined in the nearterm are a. flag (noted as questionable content/behavior, could be used as a paper trail and/or to help admins implement a policy such as three strikes) and b. acknowledge (explicit check and non-action, i.e. in response to a report).

lexicons/app/bsky/administration/moderationAction.json
devinivy: It was just the word in my head for this, all good changing it.

lexicons/app/bsky/administration/moderationAction.json
devinivy: That makes a lot of sense.  What do you think of `reversal: { createdAt, createdBy, reason }`?

lexicons/app/bsky/administration/takeModerationAction.json
devinivy: My only concern would be that it becomes more concise but maybe trickier to think/talk about without referencing the "moderation action."  E.g. what would the name of the reversal method be, and what would be the "thing" that you're passing an id for to the reversal method?

AaronGoldman:
You are mixing the concepts of labeling and actioning. This will likely be harder to disentangle later.
If I want to flag something as `spam` a user may want it sent to their client or not. If I flag something a `DMCA` the server should not send it to any client. 

There are a number of softer or harder actions that will eventually be wanted and the PDS should be able to make its own policy. If the admins using the moderation tool are adding labels, then the PDS configuration can be changed. If the admins using the moderation tool are directly taking actions, then it is very difficult to change policy often involving error prone database migrations.

Both PDSs and Client will want to apply policy based on the labels but if there is no separation of concerns between safety labels and moderation actions then there can be only one policy. You don't know why an action is it just is.

Also, worth thinking about how you want to handle the same label being asserted and retracted multiple times. AI labels it, agent retracts label on appeal, admin re-labels it overriding the agent. Is this just two distinct label events or should there be a reference chain back to the beginning? The first label event having a null supersedes field. The second retraction label event having the CID of the first event in its supersedes field. The third having the CID of the retraction in its supersedes field.

AaronGoldman:
You are mixing the concepts of labeling and actioning. This will likely be harder to disentangle the latter.
If I want to flag something as `spam` a user may want it sent to their client or not. If I flag something a `DMCA` the server should not send it to any client. 

There are a number of softer or harder actions that will eventually be wanted and the PDS should be able to make its own policy. If the admins using the moderation tool are adding labels, then the PDS configuration can be changed. If the admins using the moderation tool are directly taking actions, then it is very difficult to change policy often involving error prone database migrations.

Both PDSs and Client will want to apply policy based on the labels but if there is no separation of concerns between safety labels and moderation actions then there can be only one policy. You don't know why an action is it just is.

Also, worth thinking about how you want to handle the same label being asserted and retracted multiple times. AI labels it, agent retracts label on appeal, admin re-labels it overriding the agent. Is this just two distinct label events or should there be a reference chain back to the beginning? The first label event having a null supersedes field. The second label event having a null supersedes field.

devinivy:
I totally see what you're getting at 👍 The intent here isn't to do any labeling, it's really just to keep a log of any moderation actions that are taken by the server admin.  This is meant to be a pretty coarse-grained tool so that if you're running a server you have the ability to takedown content, and there's a history of the actions that you've taken.  I expect there will need to be other mechanisms that allow for more subjectivity/fuzziness and less tied to a single server, but the hope here is really just to give a simple tool so that a server admin isn't forced to host e.g. illegal content.

lexicons/app/bsky/admin/takeModerationAction.json
```diff
@@ -0,0 +1,30 @@
...
+      "type": "procedure",
+      "description": "Take a moderation action on an actor.",
+      "input": {
+        "encoding": "application/json",
+        "schema": {
+          "type": "object",
+          "required": ["action", "subject", "reason", "createdBy"],
+          "properties": {
+            "action": {"type": "string", "knownValues": ["app.bsky.admin.actionTakedown"]},
+            "subject": {"type": "union", "refs": ["app.bsky.actor.ref"]},
```
dholms, line 15: I don't think the declaration CID is actually relevant to the takedown. 

Like if I've banned a certain user, & they change their repo declaration to a "scene repo", I probably still want it banned

packages/pds/src/api/app/bsky/actor/createScene.ts
```diff
@@ -9,7 +9,7 @@ import AppContext from '../../../../context'
 
 export default function (server: Server, ctx: AppContext) {
   server.app.bsky.actor.createScene({
-    auth: ctx.accessVerifier,
+    auth: ctx.accessVerifierCheckTakedown,
```
dholms, line 12: should we be checking this on every procedure? for instance CRUD routes

packages/pds/src/api/app/bsky/admin/reverseModerationAction.ts
```diff
@@ -0,0 +1,50 @@
...
+        ) {
+          await adminTxn.reverseTakedownActorByDid({ did: result.subjectDid })
+        }
+
+        return result
+      })
+
+      return {
+        encoding: 'application/json',
+        body: adminService.formatModActionView(moderationAction),
```
dholms, line 46: enjoying this new pattern of services, makes routes very quick to grok

packages/pds/src/api/app/bsky/actor/createScene.ts
dholms: wait actually, why is this necessary? looks like refresh tokens get revoked & they can't create a new session 🤔

packages/pds/src/api/app/bsky/feed/getTimeline.ts
```diff
@@ -27,18 +27,27 @@ export default function (server: Server, ctx: AppContext) {
 
       const postsQb = feedService
         .selectPostQb()
-        .where('creator', '=', requester)
-        .orWhere('creator', 'in', followingIdsSubquery)
+        .where((qb) =>
```
dholms, line 30: just curios, what is the benefit of this syntax?

packages/pds/src/api/app/bsky/actor/createScene.ts
devinivy: In my opinion it's not necessary on read routes, but I'm open to it.  It's a plus that the access tokens are stateless are don't require a db roundtrip to verify.  The irrevocability of the access tokens are intended to be balanced-out by them also being short-lived.

packages/pds/src/api/app/bsky/feed/getTimeline.ts
devinivy: Good q— that puts parens around the sql clause.  Without the parens, the "or" condition can match and cause a result to be included even if a required condition within `selectPostQb()` is not matched.

packages/pds/src/services/actor.ts
```diff
@@ -14,10 +15,14 @@ export class ActorService {
...
 
-  async getUser(handleOrDid: string): Promise<(User & DidHandle) | null> {
+  async getUser(
+    handleOrDid: string,
+    includeSoftDeleted = false,
+  ): Promise<(User & DidHandle) | null> {
     let query = this.db.db
       .selectFrom('user')
       .innerJoin('did_handle', 'did_handle.handle', 'user.handle')
+      .if(!includeSoftDeleted, (qb) => qb.where(actorNotSoftDeletedClause()))
```
dholms, line 25: nice 👌

packages/pds/src/api/app/bsky/feed/getTimeline.ts
dholms: ah got it got it. thx 👌

lexicons/app/bsky/admin/takeModerationAction.json
```diff
@@ -0,0 +1,30 @@
+{
+  "lexicon": 1,
+  "id": "app.bsky.admin.takeModerationAction",
```
dholms, line 3: one more bikeshed: arguably these should be in the `atproto` namespace 🤔 

Like server takedowns seem general, but as we introduce concepts of zones/labels/etc, those seem more bsky

devinivy: Sounds good.  In that case I think we just need to ensure we're happy moving forward making the method sufficiently generalized, i.e. that we can apply a takedown to any record (versus e.g. being specific to posts).

That might encourage us to write other methods in a generalized fashion, too, and that seems totally reasonable but comes with its drawbacks.  For example, if we have a method that we want to use to list all posts that have been moderated, that method would probably receive a collection name as a parameter, and return raw post records rather than the views of post records that app.bsky provides.

My inclination was to start with specialized tooling for posts and work backwards into more generalized tooling, but not at all married to that approach!

packages/pds/src/api/app/bsky/actor/createScene.ts
devinivy: Oops, I think I responded to an old version of your comment!  All that still applies, but to answer more specifically—

I applied the takedown check on any routes that cause writes.  The idea is that a. writes are less common than reads, and we can afford the db roundtrip to verify write requests; and b. we want to _immediately_ disallow the user from being able to put up content, rather than the 15min grace period a fresh access token would give them.

lexicons/app/bsky/admin/takeModerationAction.json
devinivy: Thinking a little more, I suppose we could put actor and record takedown procedures in com.atproto and still place moderation views in app.bsky as needed.  It would still be app.bsky's job to observe those moderation actions and apply them to its indexes, but that seems reasonable enough.

One snag still coming to mind— there could be a case where a moderation action of a record has some meaning but isn't implemented.  E.g. if we allow moderating any record in com.atproto, and you takedown something very specific like a follow or vote record, there's a good chance that app.bsky doesn't actually implement that takedown in its views, and it has no practical effect.

packages/pds/src/services/feed/index.ts
```diff
@@ -108,6 +121,8 @@ export class FeedService {
       .selectFrom('post')
       .where('post.uri', 'in', postUris)
       .innerJoin('ipld_block', 'ipld_block.cid', 'post.cid')
+      .innerJoin('did_handle', 'did_handle.did', 'post.creator')
+      .where(actorNotSoftDeletedClause())
```
devinivy, line 125: Reminder to self: take a look at this— possible we want to keep this logic out of `getPostViews()` and relegate it entirely to `selectPostQb()`, etc.

packages/pds/src/api/app/bsky/actor/createScene.ts
dholms: ah got it. yeah that makes sense to me 👍

packages/pds/src/services/feed/index.ts
devinivy: Keeping this, as it covers taken-down replies on the feed views.  Added a note in the code to make that clear.

lexicons/app/bsky/admin/takeModerationAction.json
devinivy: This ended-up being the call 👍 the server-level moderation actions are based around repos and records via com.atproto.

dholms (approved):
LGTM 💪

PR #441, opened 2022-12-23
Repo mutation diffs

dholms:
This adjusts the repo storage abstraction so that it understands commits.

Instead of having a staging section that then get's persisted, the application code is in charge of collecting new blocks for a commit & storage "applies a commit", saving the blocks, and associating them with some commit CID

I also snuck a change in here to the sql repo store which does some optimistic caching of repo blocks that are likely to change. This removes at least 3 (but often more) round trips to the DB on each mutation.

### DIFF
--- a/packages/repo/bench/repo.bench.ts
+++ b/packages/repo/bench/repo.bench.ts
@@ -23,9 +23,10 @@ describe('Repo Benchmarks', () => {
       if (i % 500 === 0) {
         console.log(i)
       }
-      await repo
-        .stageUpdate({
-          action: 'create',
+
+      await repo.applyCommit(
+        {
+          action: 'create' as const,
           collection: 'app.bsky.post',
           rkey: TID.nextStr(),
           value: {
@@ -38,8 +39,9 @@ describe('Repo Benchmarks', () => {
             },
             createdAt: new Date().toISOString(),
           },
-        })
-        .createCommit(authStore)
+        },
+        authStore,
+      )
     }
 
     console.log('SIZE: ', await blockstore.sizeInBytes())
--- a/packages/repo/bench/repo.bench.ts
+++ b/packages/repo/bench/repo.bench.ts
@@ -1,6 +1,6 @@
 import * as auth from '@atproto/auth'
 import { TID } from '@atproto/common'
-import { MemoryBlockstore, Repo } from '../src'
+import { MemoryBlockstore, Repo, WriteOpAction } from '../src'
 import * as util from '../tests/_util'
 
 describe('Repo Benchmarks', () => {
@@ -26,7 +26,7 @@ describe('Repo Benchmarks', () => {
 
       await repo.applyCommit(
         {
-          action: 'create' as const,
+          action: WriteOpAction.Create,
           collection: 'app.bsky.post',
           rkey: TID.nextStr(),
           value: {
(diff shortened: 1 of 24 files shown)


packages/repo/bench/repo.bench.ts
```diff
@@ -23,9 +23,10 @@ describe('Repo Benchmarks', () => {
       if (i % 500 === 0) {
         console.log(i)
       }
-      await repo
-        .stageUpdate({
-          action: 'create',
+
+      await repo.applyCommit(
+        {
+          action: 'create' as const,
```
devinivy, line 29: It's small peas, but we could eventually consider getting an enum in here to skirt the need for `as const`.

packages/repo/src/sync.ts
```diff
@@ -4,12 +4,14 @@ import { RepoStorage } from './storage'
...
 import Repo from './repo'
 import * as verify from './verify'
+import * as util from './util'
 
 export const loadRepoFromCar = async (
   carBytes: Uint8Array,
   storage: RepoStorage,
   verifier: auth.Verifier,
 ): Promise<Repo> => {
+  const car = await util.readCar(carBytes)
```
devinivy, line 14: I don't see `car` used anywhere— is this a way to verify the bytes are for a car?  If so, might just be worth dropping a comment.

dholms: ah nope just missed it on cleanup 👌

packages/repo/src/util.ts
```diff
@@ -1,9 +1,12 @@
 import { CID } from 'multiformats/cid'
 import * as auth from '@atproto/auth'
+import { def } from '@atproto/common'
 import Repo from './repo'
 import { DataDiff, MST } from './mst'
 import { RepoStorage } from './storage'
 import { DataStore, RecordWriteOp } from './types'
+import BlockMap from './block-map'
+import { CarReader } from '@ipld/car/reader'
```
devinivy, line 9: No biggie here, just pulling this dep up since we seem to put external deps at the top of the file by convention.
```suggestion
import { CID } from 'multiformats/cid'
import { CarReader } from '@ipld/car/reader'
import * as auth from '@atproto/auth'
import { def } from '@atproto/common'
import Repo from './repo'
import { DataDiff, MST } from './mst'
import { RepoStorage } from './storage'
import { DataStore, RecordWriteOp } from './types'
import BlockMap from './block-map'
```

packages/repo/src/util.ts
```diff
@@ -17,6 +20,25 @@ export const ucanForOperation = async (
...
+  bytes: Uint8Array,
+): Promise<{ root: CID; blocks: BlockMap }> => {
+  const car = await CarReader.fromBytes(bytes)
+  const roots = await car.getRoots()
+  if (roots.length !== 1) {
+    throw new Error(`Expected one root, got ${roots.length}`)
+  }
+  const root = roots[0]
+  const blocks = new BlockMap()
+  for await (const block of car.blocks()) {
```
devinivy, line 33: I don't think this was needed 👍 
```suggestion
  for (const block of car.blocks()) {
```

devinivy (approved):
Looks good— nice idea on that little optimization! 🥂

packages/repo/src/util.ts
dholms: yup 👍

packages/repo/src/util.ts
dholms: hah opposite actually. needed it inside the for loop as well 👌

packages/repo/bench/repo.bench.ts
dholms: ended up just getting it in here 👌

devinivy (approved):

PR #444, opened 2022-12-27
Repo checkouts

dholms:
This started out as introducing repo checkouts to the API but spun out into several related things that improve our storage & sync layer:
- ripped out auth lib (we weren't actually making use of this & it was just adding extra song and dance to all of repo operations & sync verification)
- hash blocks as they come in from car file to ensure the bytes match the purported CID
- better error handling
- split out repo storage class into the naive `ReadableBlockstore` and the higher level `RepoStorage`
- alongside that, introduce `ReadableRepo` that allows for loading a repo off of the naive blockstore
- enabled by both of those, we can remove `temp` from `RepoStorage` and introduce a new `SyncStorage` class on top of two readable blockstores. this is flexible such that you can handle sync in memory or with an on-disk cache
- improve the sync & verification methods to make use of all of this
- _AND FINALLY_ add in repo checkouts
- also added `getCommitPath` as a bonus

_Note: this one does not yet include blobs for diffs/checkouts_


packages/repo/src/util.ts
```diff
@@ -14,17 +16,30 @@ import {
...
-): Promise<string> => {
-  const diff = await prevData.diff(newData)
-  const neededCaps = diff.neededCapabilities(rootDid)
-  const ucanForOp = await authStore.createUcanForCaps(rootDid, neededCaps, 30)
-  return auth.encodeUcan(ucanForOp)
+export async function* verifyIncomingCarBlocks(
+  car: AsyncIterable<CarBlock>,
+): AsyncIterable<CarBlock> {
+  for await (const block of car) {
+    await verifyCidForCborBytes(block.cid, block.bytes)
```
dholms, line 26: note to self: this should take an arbitrary codec not require cbor

packages/repo/src/storage/memory-blockstore.ts
```diff
@@ -48,56 +62,41 @@ export class MemoryBlockstore extends RepoStorage {
...
-      if (earliest && curr.equals(earliest)) {
-        return path.reverse()
-      }
-      const root = await this.get(commit.root, def.repoRoot)
+      const commit = await this.readObj(curr, def.commit)
+      const root = await this.readObj(commit.root, def.repoRoot)
       if (!earliest && root.prev === null) {
         return path.reverse()
+      } else if (earliest && root.prev.equals(earliest)) {
+        return path.reverse()
```
dholms, line 70: I'd like to get some opinions on how `getCommitPath` should be ordered.

Instinct says it should be in increasing order, ie: [commitX, commitY, commitZ] (and that's how it is right now). This makes sense because, when processing commits, you want you'd probably want to do them in order

However, I've structured it similarly to `array.slice` where the first param is _inclusive_ & the second is _exclusive_
This is because if you ask for a commit path form the current commitZ to commitX, you should already have knowledge of commitX & you don't actually need it, you only need commitZ & commitY. 
And in sense, it would suggest that the you'd receive back [commitZ, commitY]

You can feel this tension on display in `pds/sync.test.ts` where I ask for `earliest: commit[2], latest: commit[15]`, but then have to compare to `commit.slice(3, 16)`. Feels weird 🤔 

You'll also notice that in `getCommitPath` on both memory-blockstore & sql-repo-storage, we have to reverse the path when returning because the path is by nature constructed from getting a reference latest and then working backwards to the earliest.

Any thoughts?

packages/repo/src/mst/mst.ts
```diff
@@ -200,31 +209,17 @@ export class MST implements DataStore {
   // -------------------
 
   // Return the necessary blocks to persist the MST to repo storage
-  // If the topmost tree only has one entry and it's a subtree, we can eliminate the topmost tree
```
dholms, line 203: From my last MST bughunt adventures: we were trimming the top of the tree before saving, but i've moved that to the delete operation itself since it was screwing up our diff bookkeeping (& it's actually just more intuitive to do it on the operation anyway)

packages/repo/src/verify.ts
```diff
@@ -1,52 +1,132 @@
...
+  newCids: CidSet
+}
+
+export const verifyCheckout = async (
+  storage: ReadableBlockstore,
+  root: CID,
+  didResolver: DidResolver,
+): Promise<VerifiedCheckout> => {
+  const repo = await ReadableRepo.load(storage, root)
+  const validSig = await didResolver.verifySignature(
```
dholms, line 24: I will say, I don't love having to thread the `didResolver` through all the sync & verify methods. But it's either that or doing the resolution beforehand & threading through the approved signingKey.
It also doesn't take into account the fact that a user's signing key can change yet. (& god forbid their DID changes, we just error out on that for now)

packages/pds/src/api/com/atproto/sync.ts
```diff
@@ -1,9 +1,9 @@
 import { Server } from '../../../lexicon'
 import { InvalidRequestError } from '@atproto/xrpc-server'
-import { def as common } from '@atproto/common'
 import { Repo } from '@atproto/repo'
 import SqlRepoStorage from '../../../sql-repo-storage'
 import AppContext from '../../../context'
+import { CID } from 'multiformats/cid'
 
 export default function (server: Server, ctx: AppContext) {
   server.com.atproto.sync.getRoot(async ({ params }) => {
```
dholms, line 9: another bikeshed, thoughts on switching this to `getHead`? 

I've done this in the storage interfaces to avoid confusion with the similarly named `root` (which refers to the ipld node that the commit points to).

dholms: actually i just decided to do it. lmk if you have any beef with it 🐮

packages/repo/src/util.ts
dholms: handled

packages/pds/src/api/com/atproto/sync.ts
pfrazee: Yeah I'm cool with that

packages/common/src/check.ts
```diff
@@ -1,13 +1,22 @@
...
+
+export interface Checkable<T> {
   parse: (obj: unknown) => T
-  safeParse: (obj: unknown) => { success: boolean }
+  safeParse: (
+    obj: unknown,
+  ) => { success: true; data: T } | { success: false; error: ZodError }
+}
+
+export interface Def<T> {
```
dholms, line 10: wrapped the defs in this object so `RepoStorage` can give better errors

packages/repo/src/storage/memory-blockstore.ts
devinivy: I think you got this right as-is, and the mismatch with `slice()` makes sense to me.  This is similar but a little different than slice since we're slicing in "the opposite direction" if you can say that.  I think the case that highlights it best is `getCommitPath(x, null)` vs. `slice(x)`.  In the former it's clear we mean to ask for all commits ending at (i.e. before) `x`; in the latter we want items starting from (i.e. after) `x`. To hammer this home, we could make `earliest` optional and default it to `null`: `getCommitPath(x)` vs. `slice(x)`.

packages/repo/src/storage/memory-blockstore.ts
```diff
@@ -1,43 +1,57 @@
...
 
 export class MemoryBlockstore extends RepoStorage {
   blocks: BlockMap
   head: CID | null = null
 
-  constructor() {
+  constructor(blocks?: BlockMap) {
     super()
-    this.blocks = new BlockMap()
+    this.blocks = blocks || new BlockMap()
```
devinivy, line 15: Would there be any value in cloning `blocks` to avoid leakage?  I could see it arguably the other way too, that it's useful for testing purposes to mutate the passed block map.

packages/repo/src/sync.ts
```diff
@@ -1,33 +1,118 @@
...
+import { MemoryBlockstore, RepoStorage } from './storage'
+import DataDiff from './data-diff'
 import Repo from './repo'
 import * as verify from './verify'
+import { DidResolver } from '@atproto/did-resolver'
+import * as util from './util'
+import { CommitData, RecordWriteOp, RepoContents } from './types'
+import { CID } from 'multiformats/cid'
+import CidSet from './cid-set'
+import { MissingBlocksError } from './error'
```
devinivy, line 10:
```suggestion
import { CID } from 'multiformats/cid'
import { DidResolver } from '@atproto/did-resolver'
import { MemoryBlockstore, RepoStorage } from './storage'
import DataDiff from './data-diff'
import Repo from './repo'
import * as verify from './verify'
import * as util from './util'
import { CommitData, RecordWriteOp, RepoContents } from './types'
import CidSet from './cid-set'
import { MissingBlocksError } from './error'
```

devinivy: A little hard to see in the suggestion, but I just pulled `multiformats/cid` and `@atproto/did-resolver` up to the top.

packages/repo/src/sync.ts
```diff
@@ -1,33 +1,118 @@
...
+    root,
+    contents: checkout.contents,
+  }
+}
+
+export const loadFullRepo = async (
+  storage: RepoStorage,
+  repoCar: Uint8Array,
+  didResolver: DidResolver,
+): Promise<{ root: CID; ops: RecordWriteOp[] }> => {
```
devinivy, line 42: Nothing to do about this here/now I don't think, but just a thought rattling around my brain.  If someone is syncing these `ops` into e.g. their own index, and they want to keep their index consistent, they'd need to process all `ops` at once transactionally.  If the `ops` were broken-up by commit, then they could gradually process commit-by-commit (or condensed into batches of commits), landing at a consistent state each step along the way.  It would be useful to be able to ask something like "give me this history in ops batches of size roughly 100, always containing full commits."

packages/repo/src/util.ts
```diff
@@ -37,7 +52,7 @@ export const readCar = async (
   }
   const root = roots[0]
   const blocks = new BlockMap()
-  for await (const block of car.blocks()) {
+  for await (const block of verifyIncomingCarBlocks(car.blocks())) {
```
devinivy, line 55: 💪

packages/repo/src/verify.ts
devinivy: Yeah, I'm not seeing any clearly better options unless there's a natural default `DidResolver`, sort of similar to how in node there's a [global/default `http.Agent`](https://nodejs.org/api/http.html#httpglobalagent).  Can also probably be smoothed over for consumers by assigning a `DidResolver` early on and threading it down here for them in some wrapper/convenience classes, etc.

devinivy (approved):
I dig this very much 💎

packages/pds/src/api/com/atproto/sync.ts
```diff
@@ -27,14 +50,26 @@ export default function (server: Server, ctx: AppContext) {
       throw new InvalidRequestError(`Could not find repo for DID: ${did}`)
     }
     const repo = await Repo.load(storage, root)
-    const fromCid = from ? common.strToCid.parse(from) : null
-    const diff = await repo.getDiffCar(fromCid)
+    const fromCid = from ? CID.parse(from) : null
```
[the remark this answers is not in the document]
devinivy: https://github.com/bluesky-social/atproto/issues/456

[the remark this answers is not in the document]
dholms: yup good call 👍 thx for logging that

packages/repo/src/readable-repo.ts
```diff
@@ -0,0 +1,90 @@
...
+    const { blocks, missing } = await this.storage.getBlocks(cids)
+    if (missing.length > 0) {
+      throw new MissingBlocksError('getContents record', missing)
+    }
+    const contents: RepoContents = {}
+    for (const entry of entries) {
+      const { collection, rkey } = util.parseRecordKey(entry.key)
+      if (!contents[collection]) {
+        contents[collection] = {}
+      }
```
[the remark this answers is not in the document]
dholms: ah nice. i have a hard time remembering these operators, but i hit this scenario enough that that is very useful

packages/pds/src/sql-repo-storage.ts
```diff
@@ -74,11 +77,33 @@ export class SqlRepoStorage extends RepoStorage {
...
+    const missingStr = cached.missing.map((c) => c.toString())
+    const res = await this.db.db
+      .selectFrom('ipld_block')
+      .innerJoin(
+        'ipld_block_creator as creator',
+        'creator.cid',
+        'ipld_block.cid',
+      )
+      .where('creator.did', '=', this.did)
+      .where('ipld_block.cid', 'in', missingStr)
```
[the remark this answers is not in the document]
dholms: ah yup. i'd done it in other methods, but missed it here. good looks 👍

packages/pds/src/sql-repo-storage.ts
```diff
@@ -139,61 +164,61 @@ export class SqlRepoStorage extends RepoStorage {
...
-      commit.prev === null
-        ? this.insertRoot(commit.root)
-        : this.updateRoot(commit.root, commit.prev)
     const insertCommitHistory = this.db.db
       .insertInto('repo_commit_history')
-      .values({
-        commit: commit.root.toString(),
-        prev: commit.prev ? commit.prev.toString() : null,
-      })
+      .values(commitHistory)
```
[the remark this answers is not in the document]
dholms: all good i tackled them 👍

packages/repo/src/mst/diff.ts
```diff
@@ -1,106 +1,129 @@
...
 
-  recordNewCid(cid: CID): void {
-    this.newCids.add(cid)
+export const mstDiff = async (
+  curr: MST,
+  prev: MST | null,
+): Promise<DataDiff> => {
+  await curr.getPointer()
+  if (prev === null) {
+    return nullDiff(curr)
```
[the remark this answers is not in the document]
dholms: heh actually there's a problem there which is that an empty tree still contains a node (with no entries) & if you diff two empty trees, you don't receive that node back in the diff.

in other words: `diff(emptyTree, emptyTree) !== diff(emptyTree, null`

packages/repo/src/mst/mst.ts
```diff
@@ -766,14 +672,46 @@ export class MST implements DataStore {
...
+      toFetch = nextLayer
+    }
+    const leafData = await this.storage.getBlocks(leaves.toList())
+    if (leafData.missing.length > 0) {
+      throw new MissingBlocksError('mst leaf', leafData.missing)
+    }
+
+    for (const leaf of leafData.blocks.entries()) {
+      await car.put(leaf)
+    }
```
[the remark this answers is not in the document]
dholms: Yup I think that's right^^ Getting them all at once is certainly better than what we were doing (involving hundreds/thousands of round trips), but there is definitely a better middle ground here

packages/repo/src/repo.ts
```diff
@@ -167,93 +140,81 @@ export class Repo {
       }
     }
 
-    const dataDiff = await data.blockDiff()
-    newBlocks.addMap(dataDiff.blocks)
+    const unstoredData = await data.getUnstoredBlocks()
+    newBlocks.addMap(unstoredData.blocks)
+
+    // ensure we're not missing any blocks that were removed and then readded in this commit
+    const diff = await DataDiff.of(data, this.data)
```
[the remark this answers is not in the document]
dholms: 🙏

packages/repo/src/repo.ts
```diff
@@ -167,93 +140,81 @@ export class Repo {
...
-    return Repo.load(this.storage, revertTo)
+    return Repo.load(this.storage, commit.commit)
   }
 
   // CAR FILES
   // -----------
 
-  async getCarNoHistory(): Promise<Uint8Array> {
-    return this.openCar((car: BlockWriter) => {
+  async getCheckout(): Promise<Uint8Array> {
```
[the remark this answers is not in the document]
dholms: yup i like it too. continuing to bit a lot of terms from git 😛 & checkout here has a slightly different meaning than a git checkout, but it's much simpler to talk about than "a-historical repo copy" lol

packages/repo/src/storage/memory-blockstore.ts
dholms: hmm that's a good point. Yeah I like cloning 👍

packages/repo/src/sync.ts
dholms: That's a good point. My motivation here was actually to simplify processing ops: for instance, if someone creates & deletes a record several times over a commit range, this will collapse it into one operation.

On second look, this should probably be an optional processing step done by the consumer: we return ops per commit & they can process them atomically by commit, or they can collapse them down into one big set of ops

packages/repo/src/mst/mst.ts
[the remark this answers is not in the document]
devinivy: Right on 👍

Followup question for my understanding— is the order of the blocks in the CAR significant?  I.e. is it important that the leaves are placed together last here?

packages/repo/src/sync.ts
dholms: this is actually trickier than i thought it'd be. you lose some info when going from diff -> op. specifically the CID associated with the del. It may make sense to start tracking cids in the operations as well as the records & include the cid in the delete operation as well 🤔

devinivy: Yeah hmmm, I can see that.  Could be nice!  If it's too much of a rabbit hole for this PR though, I'm certainly not too hung-up on it right now.

dholms: Alright I ended up doing it, because it actually does make these functions much nicer 

Split it out into it's own PR here: https://github.com/bluesky-social/atproto/pull/460

packages/repo/src/mst/mst.ts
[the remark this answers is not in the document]
dholms: Hmmm I would say no 🤔 
I just did it here to reduce blockstore roundtrips. ie "get all leaves at once instead of making a roundtrip every time you encounter one"
I don't see a huge benefit to forcing order on a carfile. Maybe baking too many assumptions into it & I expect consumers to either hold the whole thing in memory or write it to disk. I suppose can see a benefit to having a deterministic layout in a carfile if only because then a carfile takes on an identity of it's own & you can know if you have the right thing by hash or something 🤔 
Filecoin uses depth-first search for a deterministic ordering: https://ipld.io/specs/transport/car/carv1/#determinism
I suspect ours might be breadth-first search if we do so 🤔

PR #447, opened 2022-12-30
Server moderation reporting and resolution

devinivy:
Users may report a repo or record, and a PDS administrator may then resolve reports via moderation actions (e.g. a takedown #440).

Any action resolving a report must match the record or repo being reported.  A report may be resolved by multiple actions, e.g. a report of a repo may be resolved by moderation actions on multiple of that repo's post records.  And an action may resolve multiple reports, e.g. if several users reported the same post record, then a single moderation action on that post may resolve all the reports.


packages/pds/src/api/com/atproto/repo.ts
```diff
@@ -241,4 +244,70 @@ export default function (server: Server, ctx: AppContext) {
       })
     },
   })
+
+  server.com.atproto.repo.report({
```
dholms, line 248: bikeshed, but it feels wrong for this to be in the same namespace as repo crud/describe

thoughts on `com.atproto.report.create` or similar?

packages/pds/src/services/repo/index.ts
```diff
@@ -165,4 +169,100 @@ export class RepoService {
       }),
     )
   }
+
+  async report(info: {
```
dholms, line 173: similarly, these feels like a separate service from the RepoService which is mainly around repo mutations

packages/pds/src/db/tables/moderation.ts
```diff
@@ -1,13 +1,15 @@
...
 
-export const moderationActionTableName = 'moderation_action'
+export const actionTableName = 'moderation_action'
+export const reportTableName = 'moderation_report'
+export const reportResolutionTableName = 'moderation_report_resolution'
 
 export interface ModerationAction {
   id: Generated<number>
   action: typeof TAKEDOWN
   subjectType: 'com.atproto.admin.moderationAction#subjectRepo'
```
dholms, line 11: should this also allow for records?

dholms: ah that's in the next PR nvm

dholms (approved):
looks good. just the thoughts about the namespace for this 👍

packages/pds/src/api/com/atproto/repo.ts
devinivy: My thought process here was that `com.atproto.repo` was where actions surrounding repos and records would go (i.e. "get a record / create a record / delete a record / report a record")— so not really considering repo mutations to be a special carve-out, but instead just some other actions one can take on a record.

At the same time, I totally understand the tendency to want to separate out the methods that treat the repo concretely as a data structure, rather than abstractly as just another thing the PDS can talk about.

Long way of saying, `com.atproto.report.create` sounds good to me 😎

packages/pds/src/services/repo/index.ts
devinivy: I was following the same line of thinking as in my last comment, and totally down to move some things around.  What would you say to a `ModerationService`, where reports and the views of actions/reports could live?  The server admin actions of taking a moderation action and resolving a report could stay on the `AdminService`, or we could absorb them into the `ModerationService`.

It's a bit of a balance between colocating moderation functionality in a single service versus keeping admin and non-admin functionality separate from each other.  I usually tend towards the former, and treat the service/application layer as being privileged/non-authorization-aware, but there are tons of different reasonable ways to do it!

packages/pds/src/api/com/atproto/repo.ts
dholms: > I totally understand the tendency to want to separate out the methods that treat the repo concretely as a data structure, rather than abstractly as just another thing the PDS can talk about

yup great phrasing of the friction i felt^^ 

I totally see your train of thought here, but I imagine we'll have many more methods that deal with repos/records in some capacity which will bloat that namespace. Seems cleaner to have `com.atproto.repo` methods be for only datastructure operations

packages/pds/src/services/repo/index.ts
dholms: Yup `ModerationService` sounds great 👍 
I don't have a strong opinion on it co-locating the admin actions in there as well, altho my instinct is probably to do so 🤔

devinivy:
@dholms should be all set for another peek 👍 moving those things around felt pretty good.

dholms (approved):
looks great 🙌

PR #448, opened 2022-12-30
Server moderation of records

devinivy:
Implements the server moderation action for record-level takedowns (followup to #440).  Also made some minor tweaks to allow services to reference each other, and simplify deps around the event stream consumers.


packages/pds/src/services/record/index.ts
```diff
@@ -110,6 +116,7 @@ export class RecordService {
       .selectFrom('record')
       .select('collection')
       .where('did', '=', did)
+      .groupBy('collection')
```
devinivy, line 119: This is a little fix to `com.atproto.repo.describe`.

packages/pds/tests/__snapshots__/moderation.test.ts.snap
```diff
@@ -30,7 +30,7 @@ Array [
     "reasonType": "com.atproto.repo.report#spam",
     "reportedByDid": "user(0)",
     "subject": Object {
-      "$type": "com.atproto.repo.report#subjectRecord",
+      "$type": "com.atproto.repo.report#subjectRecordRef",
```
devinivy, line 33: This is a minor fix to some work in #447.

lexicons/com/atproto/admin/moderationAction.json
```diff
@@ -32,6 +32,24 @@
         "did": {"type": "string"}
       }
     },
+    "subjectRecord": {
```
dholms, line 35: does it make sense to unify these with the similar types in `com.atproto.repo.report`?

packages/pds/src/event-stream/consumers/index.ts
```diff
@@ -1,32 +1,27 @@
...
-  keypair: DidableKey,
-) => {
-  const getAuthStore = (did: string) => {
-    return auth.verifier.loadAuthStore(keypair, [], did)
-  }
-  messageQueue.listen('add_member', new AddMemberConsumer())
-  messageQueue.listen('remove_member', new RemoveMemberConsumer())
-  messageQueue.listen('add_upvote', new AddUpvoteConsumer())
-  messageQueue.listen(
+export const listen = (ctx: AppContext) => {
```
dholms, line 10: nice 👌

packages/pds/src/services/actor.ts
```diff
@@ -7,12 +7,13 @@ import { DidHandle } from '../db/tables/did-handle'
 import { Record as DeclarationRecord } from '../lexicon/types/app/bsky/system/declaration'
 import { APP_BSKY_GRAPH } from '../lexicon'
 import { actorNotSoftDeletedClause } from '../db/util'
+import { Services } from '.'
 
 export class ActorService {
-  constructor(public db: Database) {}
+  constructor(public services: Services, public db: Database) {}
```
dholms, line 13: I see what you're going for here, giving every service a reference to other services. but we already have an asymmetry in their dependencies (need blobstore/message queue for instance). any reason to not do the same here?

dholms: this pattern also causes an unfortunate circular dependency. which i generally try to avoid but also have done myself in some places so im not overly worried. can always split out with interfaces if we have the reason to

packages/pds/src/services/admin.ts
```diff
@@ -77,7 +103,7 @@ export class AdminService {
     return result
   }
 
-  async takedownActorByDid(info: { takedownId: number; did: string }) {
+  async takedownRepo(info: { takedownId: number; did: string }) {
```
dholms, line 106: nice, i like talking about this in terms of records/repos

packages/pds/src/services/index.ts
```diff
@@ -15,14 +15,27 @@ export function createServices(resources: {
...
-  return {
-    actor: ActorService.creator(),
-    auth: AuthService.creator(),
-    feed: FeedService.creator(imgUriBuilder),
-    record: RecordService.creator(messageQueue),
-    repo: RepoService.creator(messageQueue, blobstore),
-    admin: AdminService.creator(),
+  const services = {
+    get actor() {
+      return ActorService.creator(services)
```
dholms, line 20: whoa actually didn't know that this scoping worked lol

lexicons/com/atproto/admin/moderationAction.json
devinivy: Yeah, I think so, but there are a few things happening here that I wondered if we'd want to untangle at the same time:
 - The `{ did, collection, rkey, cid? }` shape is shared by `deleteRecord` and `putRecord` (though they don't carve it out into `subject`).
 - The `{ uri, cid }` pair is essentially a `com.atproto.repo.strongRef`, but there are still other methods that aren't using it such as the `*Record` outputs.  I almost used it, but it felt a bit funny to mix `com.atproto.repo.strongRef` in a union alongside something more single-use like `com.atproto.admin.moderationAction#subjectRepo`.
 - `{ did }` is arguably used by `batchWrite` and some other places to reference a repo.

I could see these as three shareable types: `recordRef`, `strongRef`, and `repoRef` respectively.  If we want to share some types across `com.atproto.*`, do you think we'd want to create a `com.atproto.system` namespace for them?  Or place them under `com.atproto.repo`?

packages/pds/src/services/actor.ts
devinivy: The dependency felt alright to me, but it's coming from the perspective that the service/application layer genuinely is "one thing" that is allowed to call into itself however it wishes.  I see this sort of analogously to how two methods on the same class both share `this`.

The situation that made me reach for this was passing `blobstore` and `messageQueue` into `AdminService` in order to get it to `RecordService`, even though neither of those would be exercised from `AdminService` nor the `RecordService` methods that were being used there.

The dependencies were starting to get leaky, and I saw sharing service creators with each service as a way to sort of acknowledge that and isolate all the dependency plumbing to one place (src/services/index.ts).  At the end of the day it's really no biggie to me— happy to add the dependency plumbing between the services.  Worst case scenario they end-up relying on each other in such a way that they all require the same constructor arguments, which would be fine!

packages/pds/src/services/index.ts
devinivy: Haha yeah!  The getter makes access to `services` lazy, so it's being referenced only when someone indexes into `services`.  It's a pretty useful tool for creating app contexts!

packages/pds/src/services/actor.ts
dholms: Maybe a good in-between here is being explicit about which services each other service relies on?

So for instance: `AdminService` relies on `RecordService` so pass the creator for that into `AdminService` but `RecordService` doesn't rely on `AdminService` so it doesn't need any dependencies. This would just force us to be honest about the dependencies between services. 

I'm not at all worried about it right now, but I think for future development it could help force us to be more honest about our logical dependencies & prevent mixing lower order & higher order services. For instance, it feels wrong to me for `RecordService` (lower order) to be able to call into `FeedService` (higher order) even tho they're both in the service layer. Thoughts? 🧐

lexicons/com/atproto/admin/moderationAction.json
dholms: ah strongRef is already in `.repo` so lets just stick them there

packages/pds/src/services/actor.ts
devinivy: I'm with ya!  I think I have a decent stand-in for now.  The moderation service, for example, will wire-up its services like this:

```ts
export class ModerationService {
  constructor(
    public db: Database,
    public messageQueue: MessageQueue,
    public blobstore: BlobStore,
  ) {}

  static creator(messageQueue: MessageQueue, blobstore: BlobStore) {
    return (db: Database) => new ModerationService(db, messageQueue, blobstore)
  }

  services = {
    repo: RepoService.creator(this.messageQueue, this.blobstore),
    record: RecordService.creator(this.messageQueue),
  }
```

I think/hope this strikes a balance between making the service dependencies explicit, granular, and isolating the plumbing.

dholms: sweet that sounds great 👍

devinivy:
@dholms should be good for a peek 👍

dholms (approved):
looks great 🙌 

thx for the back & forth on both those things 🙏

PR #451, opened 2023-01-02
API & client improvements

pfrazee:
Updates:

- Adds typeguard helpers to generated code (eg `isRecord(v: unknown): v is Record`)
- Adds validation helpers to generated code (eg `validateRecord(v: unknown): ValidationResult`)
- Adds a general `validate()` method to the `Lexicons` class.
- Adds a `validateResponse` option to xrpc-server which can be set to `false` to disable validation of responses
- Adds client-side validation of XRPC responses along with a new `XRPCInvalidResponseError` for failed validation.

Basically this PR is tackling a set of tasks we've had queued up in this space. The type-guard and validation helpers are something I've been needing in the app. The client-side validation is to begin to address issues with "faulty" servers.

### DIFF
--- a/packages/lexicon/src/types.ts
+++ b/packages/lexicon/src/types.ts
@@ -278,7 +278,7 @@ export function isValidLexiconDoc(v: unknown): v is LexiconDoc {
 }
 
 export function isObj(obj: unknown): obj is Record<string, unknown> {
-  return !!obj && typeof obj === 'object'
+  return obj !== null && typeof obj === 'object'
 }
 
 export function hasProp<K extends PropertyKey>(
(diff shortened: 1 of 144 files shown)


packages/lexicon/src/types.ts
```diff
@@ -278,7 +278,7 @@ export function isValidLexiconDoc(v: unknown): v is LexiconDoc {
 }
 
 export function isObj(obj: unknown): obj is Record<string, unknown> {
-  return !!obj && typeof obj === 'object'
+  return obj !== null && typeof obj === 'object'
```
dholms, line 281: worth throwing in any extra checks like `!Array.isArray()` here?

pfrazee: The way I use it currently, there's pretty much always a `hasProp()` check after, or other validation. Probably not needed?

dholms (approved):
awesome much needed 🙏

devinivy (approved):
😎 nice!

PR #457, opened 2023-01-04
Fix lexicon validation, do not allow undefined when required

devinivy:
I saw `app.bsky.feed.getAuthorFeed` 500ing when passed the parameter `?author=`.  This parameter is considered missing, so it's mapped to `undefined`, and it turns out there's a little bug in the lexicon validation for catching these `undefined`s, so I pulled together this little fix.

### DIFF
--- a/packages/lexicon/src/validators/complex.ts
+++ b/packages/lexicon/src/validators/complex.ts
@@ -123,7 +123,7 @@ export function object(
   // required
   if (Array.isArray(def.required)) {
     for (const key of def.required) {
-      if (!(key in value)) {
+      if (typeof value[key] === 'undefined') {
         return {
           success: false,
           error: new ValidationError(`${path} must have the property "${key}"`),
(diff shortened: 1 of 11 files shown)


packages/lexicon/src/validators/complex.ts
```diff
@@ -123,7 +123,7 @@ export function object(
   // required
   if (Array.isArray(def.required)) {
     for (const key of def.required) {
-      if (!(key in value)) {
+      if (typeof value[key] === 'undefined') {
```
dholms, line 126: one of my least fav JS quirks 😒 
"no it's not undefined, it's defined as undefined"

dholms (approved):
validation fix looks good 👍 

just wanna make sure the removal of required "term" is intentional before merge

pfrazee:
Oh silly mistake on my part. LGTM

PR #459, opened 2023-01-04
Fix resolution of server did in com.atproto.handle.resolve

devinivy:
This fixes two bugs with resolving the server's did from `com.atproto.handle.resolve`:
 - The recovery did key was being served rather than the server's PLC did.
 - Passing the public hostname of the PDS as `handle` would not resolve the server did, since it was expecting the private hostname.


dholms (approved):
good catch 👍

PR #460, opened 2023-01-04
Record write descriptions

dholms:
Adds a new type similar to `RecordWriteOp`: `RecordWriteDescript` which includes relevant CIDs to the operation. 

Generally "ops" are used when _making_ a change. "descripts" are used when _syncing_ a change.

Alongside this, commit-aware sync methods now return a `WriteLog` which is `RecordWriteDescript[][]` which can then be collapsed down into a single array of all writes that occurred over that range

Motivated by https://github.com/bluesky-social/atproto/pull/444#discussion_r1061677134


devinivy (approved):
Yesss, excellent.

PR #462, opened 2023-01-06
Record proofs

dholms:
This allows for syncing proofs of record existence or non-existence.

The provider method takes a list of paths & returns all necessary blocks to prove the contents of that path.

The consumer has two options:
- `verifyRecords` which walks the tree & returns all reachable leaves, thereby "discovering" the including records
- `verifyProofs` which takes a list of claims of existence/nonexistence and returns lists of both verified & unverified claims

This also introduces the `getRecord` XRPC method for getting a proof of an individual record.
The originally intended `getRecords` method requires an array-like in the params which is not currently supported by lexicon. We either need to establish a mechanism for arrays in parameters or make this method a procedure.

One more thing that accidentally slipped into this PR: renaming some of our functions that work with IPLD blocks to be more explicit that they're dealing with the dag-cbor codec.


packages/repo/src/verify.ts
```diff
@@ -153,4 +155,87 @@ export const verifyCommitPath = async (
   return updates
 }
 
+export const verifyProofs = async (
```
dholms, line 158: I didn't add the equivalent sync methods for these because proofs like these some more one off & im not sure consumers will want to save them. Just seemed a bit superfluous.

On the otherhand if they don't, they could always just pass in a memory blockstore 🤔

packages/repo/src/verify.ts
```diff
@@ -153,4 +155,87 @@ export const verifyCommitPath = async (
...
+    } else {
+      const expected = await cidForCbor(claim.record)
+      if (expected.equals(found)) {
+        verified.push(claim)
+      } else {
+        unverified.push(claim)
+      }
+    }
+  }
+  return { verified, unverified }
```
devinivy, line 201: Made it look easy in here.

packages/repo/tests/sync/checkout.test.ts
```diff
@@ -0,0 +1,56 @@
...
+    const hasGenesisCommit = await syncStorage.has(commitPath[0])
+    expect(hasGenesisCommit).toBeFalsy()
+  })
+
+  it('throws on a bad signature', async () => {
+    const badRepo = await util.addBadCommit(repo, keypair)
+    const checkoutCar = await sync.getCheckout(storage, badRepo.cid)
+    await expect(
+      sync.loadCheckout(syncStorage, checkoutCar, didResolver),
+    ).rejects.toThrow()
```
devinivy, line 54: Might be worth verifying what is thrown here.  I think this would do the trick:
```suggestion
    ).rejects.toThrow('Invalid signature on commit')
```

packages/repo/tests/sync/diff.test.ts
```diff
@@ -0,0 +1,72 @@
...
+    const loaded = await sync.loadFullRepo(syncStorage, car, didResolver)
+    syncRepo = await Repo.load(syncStorage, loaded.root)
+    const contents = await syncRepo.getContents()
+    expect(contents).toEqual(repoData)
+    await util.verifyRepoDiff(loaded.writeLog, {}, repoData)
+  })
+
+  it('syncs a repo that is behind', async () => {
+    // add more to providers's repo & have consumer catch up
+    const beforeData = JSON.parse(JSON.stringify(repoData))
```
devinivy, line 49: There's a nice way to do this now:
```suggestion
    const beforeData = structuredClone(repoData)
```

packages/repo/tests/sync/diff.test.ts
```diff
@@ -0,0 +1,72 @@
...
+    expect(contents).toEqual(repoData)
+    await util.verifyRepoDiff(loaded.writeLog, beforeData, repoData)
+  })
+
+  it('throws on a bad signature', async () => {
+    const badRepo = await util.addBadCommit(repo, keypair)
+    const diffCar = await sync.getDiff(storage, badRepo.cid, syncRepo.cid)
+    await expect(
+      sync.loadDiff(syncRepo, diffCar, didResolver),
+    ).rejects.toThrow()
```
devinivy, line 70: Could be nice to get a little gut check in here:
```suggestion
    ).rejects.toThrow('Invalid signature on commit')
```

packages/repo/tests/sync/narrow.test.ts
```diff
@@ -0,0 +1,140 @@
...
+
+  const doVerify = (proofs: Uint8Array, claims: RecordClaim[]) => {
+    return verify.verifyProofs(repo.did, proofs, claims, didResolver)
+  }
+
+  it('verifies valid records', async () => {
+    const claims = util.contentsToClaims(repoData)
+    const proofs = await getProofs(claims)
+    const results = await doVerify(proofs, claims)
+    expect(results.verified).toEqual(claims)
```
devinivy, line 39: Probably a bit of a paranoid move, but an extra check here would guarantee that some verification is definitely happening in the test:
```suggestion
    expect(claims.length).toBeGreaterThan(0)
    expect(results.verified).toEqual(claims)
```

devinivy (approved):
Ace—exciting to see this working.  All the recent repo abstractions made this quite clear to read through.

packages/repo/tests/sync/checkout.test.ts
devinivy: Ah, just noticed `RepoVerificationError` which is probably what we'd actually want here.

packages/pds/tests/sync.test.ts
```diff
@@ -148,6 +149,59 @@ describe('repo sync', () => {
...
+      did,
+      new Uint8Array(car.data),
+      [claim],
+      didResolver,
+    )
+    expect(result.verified.length).toBe(1)
+    expect(result.unverified.length).toBe(0)
+  })
+
+  it('sync a proof of non-existene', async () => {
```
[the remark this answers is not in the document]
dholms: 🙏

packages/repo/src/mst/mst.ts
```diff
@@ -669,6 +669,40 @@ export class MST implements DataStore {
     return leaves.length
   }
 
+  // Reachable tree traversal
+  // -------------------
+
+  // Walk reachable branches of tree & emit nodes, consumer can bail at any point by returning false
+  async *walkReachable(): AsyncIterable<NodeEntry> {
```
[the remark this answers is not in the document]
dholms: This would be for an MST that is being read out of a blockstore that may not hall all relevant blocks.

The simplest case is a proof for a single record. Then, the only reachable nodes are the blocks in that proof & the only reachable leaf is the record itself

packages/repo/src/verify.ts
dholms: 😜

packages/repo/tests/sync/checkout.test.ts
dholms: ah good looks 👍

packages/repo/tests/sync/diff.test.ts
dholms: oh beautiful how did i not know this

packages/repo/tests/sync/narrow.test.ts
dholms: i dig 👌

packages/repo/src/mst/mst.ts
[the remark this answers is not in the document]
devinivy: Ahh, got it.  For some reason my brain imagined an MST with multiple components, i.e. having multiple roots and traversing nodes connected to just one of the roots.  Aaaaaanyways!

devinivy (approved):
Even more ✅

PR #473, opened 2023-01-19
Migration test

dholms:
This is actually a relatively easy migration to check that it went well.

Instead of the somewhat contrived test I was doing earlier, I just fill the db with sample data, migrate down (delete both tables), migrate up & ensure that the data matches


packages/pds/tests/migrations/repo-sync-data.test.ts
```diff
@@ -0,0 +1,67 @@
...
+  let blocks: RepoCommitBlock[]
+
+  it('fetches the current state of the tables', async () => {
+    history = await getHistory()
+    blocks = await getBlocks()
+  })
+
+  it('migrates down', async () => {
+    await db.migrator.migrateDown()
+    await expect(getHistory()).rejects.toThrow()
```
devinivy, line 56: This line is responsible for the CI failure:
```
@atproto/pds: FAIL tests/migrations/repo-sync-data.test.ts (9.414 s)
@atproto/pds:   ● repo sync data migration › migrates down
@atproto/pds:     expect(received).rejects.toThrow()
@atproto/pds:     Received function did not throw
@atproto/pds:       54 |   it('migrates down', async () => {
@atproto/pds:       55 |     await db.migrator.migrateDown()
@atproto/pds:     > 56 |     await expect(getHistory()).rejects.toThrow()
```

devinivy (approved):
Barring that one test failure which I can't explain, this looks good 👍 !

Issue #476, opened 2023-01-20
Sort "suggested follows" by number of posts (for now)


pfrazee:
The suggested follows are going to need a sophisticated solution, but we need something better for right now. We're gonna roll with the (easily-gameable) number of posts so we can highlight active users during the beta.

Could yall update `app.bsky.actor.getSuggestions` to sort by post count? <3

PR #477, opened 2023-01-20
Sort "suggested follows" by number of posts

dholms:
Closes https://github.com/bluesky-social/atproto/issues/476

### DIFF
--- a/packages/pds/src/api/app/bsky/actor/getSuggestions.ts
+++ b/packages/pds/src/api/app/bsky/actor/getSuggestions.ts
@@ -1,5 +1,6 @@
 import AppContext from '../../../../context'
-import { notSoftDeletedClause } from '../../../../db/util'
+import { Cursor, GenericKeyset, paginate } from '../../../../db/pagination'
+import { countAll, notSoftDeletedClause } from '../../../../db/util'
 import { Server } from '../../../../lexicon'
 import { getDeclarationSimple } from '../util'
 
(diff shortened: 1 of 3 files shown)


packages/pds/src/api/app/bsky/actor/getSuggestions.ts
```diff
@@ -62,14 +74,37 @@ export default function (server: Server, ctx: AppContext) {
...
+  }
+  labeledResultToCursor(labeled: CountDidLabeledResult) {
+    return {
+      primary: labeled.primary.toString(),
+      secondary: labeled.secondary,
+    }
+  }
+  cursorToLabeledResult(cursor: Cursor) {
+    return {
+      primary: parseInt(cursor.primary),
```
devinivy, line 106: It's annoying, but we might consider a check on whether `cursor.primary` parses to a number.  Otherwise most random cursor strings sent-up here will probably get turned into 500s rather than 400s.  In other places I have something like this:
```ts
    if (isNaN(parsed)) {
      throw new InvalidRequestError('Malformed cursor')
    }
```

devinivy (approved):
Looks right!

packages/pds/src/api/app/bsky/actor/getSuggestions.ts
dholms: ah nice good call 👌

devinivy (approved):
Nice catch 👍

PR #478, opened 2023-01-24
Update module publish scripts

pfrazee:
This PR updates the publishing scripts of the following modules:

- pds
- auth
- crypto
- did-resolver
- handle
- xrpc-server
- common
- plc
- uri
- repo

I have also published v0.0.1 of each of these.

### DIFF
--- a/packages/auth/package.json
+++ b/packages/auth/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@atproto/auth",
-  "version": "0.0.3",
+  "version": "0.0.1",
   "main": "src/index.ts",
   "license": "MIT",
   "scripts": {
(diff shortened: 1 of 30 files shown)


packages/auth/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@atproto/auth",
-  "version": "0.0.3",
+  "version": "0.0.1",
```
devinivy, line 3: Huh!  Any idea why this was 0.0.3?

devinivy (approved):
Right on!

packages/auth/package.json
pfrazee: I think from the adx era

PR #483, opened 2023-01-24
Remove scenes

dholms:
Removing scenes & trends from the PDS.

This also removes the views for getAssertions & getMembers(hips) along with notifications for assertions, however it leaves behind the lexicons for assertion & confirmation records. I'm not opposed to removing those also, but they seem like the most likely candidate to be worked back in soon.

For those listening in: Scenes _will_ be making a comeback! However at the current moment they're a little half-baked & we want to be sure to flesh out the concept a bit before committing to it feature-wise

---

Not included in this PR: 
- DB migrations
- removing old records from user repos

Have plans to do those in a subsequent PR shortly


packages/pds/src/api/app/bsky/actor/updateProfile.ts
```diff
@@ -14,15 +14,7 @@ export default function (server: Server, ctx: AppContext) {
   server.app.bsky.actor.updateProfile({
     auth: ctx.accessVerifierCheckTakedown,
     handler: async ({ auth, input }) => {
-      const requester = auth.credentials.did
-
-      const did = input.body.did || requester
```
devinivy, line 19: Would not mind getting rid of these `did` inputs!  I don't think we have a use for them now, and I think we sort of want to redesign that bit of auth anyway next time we come across the same use-case.

devinivy (approved):
Looks good, and I support keeping assertions and confirmations around.  Slightly tempting to keep the `getAssertions` endpoint around, even, but I think you did the right thing 👍

packages/pds/src/api/app/bsky/actor/updateProfile.ts
dholms: ah yeah that's not a bad idea.

would clean things up & they're the type of param we can add in later (or solve w the auth system)

PR #446, opened 2022-12-29
Feature branch: storage & sync revamp

dholms:
Feature branch for revamping repo storage & sync

- https://github.com/bluesky-social/atproto/pull/436
- https://github.com/bluesky-social/atproto/pull/441
- https://github.com/bluesky-social/atproto/pull/443
- https://github.com/bluesky-social/atproto/pull/444
- https://github.com/bluesky-social/atproto/pull/460
- https://github.com/bluesky-social/atproto/pull/462
- https://github.com/bluesky-social/atproto/pull/473
- https://github.com/bluesky-social/atproto/pull/485
- https://github.com/bluesky-social/atproto/pull/487

PR #485, opened 2023-01-24
Sync mime types

dholms:
Per @devinivy's comment here: https://github.com/bluesky-social/atproto/pull/446#discussion_r1082130777

We should be using the mime type `application/vnd.ipld.car` instead of `application/cbor`.

I also removed the unused `updateRepo` method (which was unimplemented & doesn't make sense without client-held keys)


devinivy (approved):
Sweet!

PR #492, opened 2023-01-26
Allow login using email

devinivy:
This allow users to create a session using their email address (in addition to their handle).  I marked session.create's `handle` input as deprecated and added an optional `identifier` input— the PDS should be free to establish any identifier it wants with its users in addition to their handle.

### DIFF
--- a/lexicons/com/atproto/session/create.json
+++ b/lexicons/com/atproto/session/create.json
@@ -11,7 +11,8 @@
           "type": "object",
           "required": ["handle", "password"],
           "properties": {
-            "handle": {"type": "string"},
+            "handle": {"type": "string", "description": "Handle of the authenticating user (deprecated)."},
+            "identifier": {"type": "string", "description": "Handle or other identifier supported by the server for the authenticating user."},
             "password": {"type": "string"}
           }
         }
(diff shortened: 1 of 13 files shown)


pfrazee:
Okay I'm +1 to this approach. Your approach to the lexicon approach is right for published schemas but I think because this isn't published yet, what we can do instead is:

- Just switch the lexicon to `identifier` and drop `handle` entirely
- In the PDS code, still check for `handle`. Open an issue to drop that in a month(?) and comment with the issue number

devinivy:
@pfrazee one minor detail to confirm— I believe we will need to leave `identifier` optional for the time being.  Then when we stop checking for `handle` in a month, it should become required.  So technically you will be able to send a valid payload that looks like `{ password }`, and it will just behave like you gave a bad identifier/password combo.

pfrazee:
@devinivy Ah yes, you're quite right.

devinivy:
Alright, that should be patched-up!  Created #493 to track the deprecation of `handle`.

pfrazee:
LGTM!

dholms (approved):
cool cool 👌

Issue #490, opened 2023-01-26
Improve thumbnail resolution


pfrazee:
The base images are now high enough rez that they look decent, but the thumbnails are still looking bad. Devin has some ideas on how to fix this.

pfrazee:
Done!

PR #495, opened 2023-01-26
Feed thumbnail sizing, quality param bugfix

devinivy:
Two things going on here:
 - changed feed thumbnail sizing to "cover" so the shortest (rather than longest) dimension of thumbnails will now be 500px.
 - the image `quality` parameter was not being factored into URLs.  I fixed that issue and removed its usage, so we have a fix but also maintained current behavior.

Hoping this is a fix for #490, but will have to do some real world testing to confirm.  A potential path in the future is to switch to webp, which seems to give improved results particularly for low-res images.


pfrazee:
LGTM, let me know when this puppy is on staging <3

dholms (approved):
noice ✨

Issue #497, opened 2023-01-27
Filter out follows from suggestions


dholms:
This is being done client side right now & leading to multiple requests

PR #498, opened 2023-01-27
Filter out self & follows from suggestions

dholms:
Update `app.bsky.actor.getSuggestions` to not return the requesting user or any users that they already follow.

Also remove `myState` from the lexicon because it is no longer used (since the follow will always be `undefined`).

Closes https://github.com/bluesky-social/atproto/issues/497


pfrazee:
Thxu good sir!

devinivy (approved):
Looks good!  If we want this to whiz a bit faster, could add an index on post creator to help with the count-per-creator.

PR #499, opened 2023-01-27
Change esbuild version to explicitly target 0.14.50

jackmcasey:
Not an npm guy, so this might be user error, but just had a fresh build fail on me due to an unavailable version of esbuild:

`error https://registry.yarnpkg.com/esbuild-linux-32/-/esbuild-linux-32-0.14.54.tgz: getaddrinfo EAI_AGAIN registry.yarnpkg.com`

This is on node:18-alpine so not sure why it's looking for a 32 bit package, but that version of esbuild appears to be unavailable for download for any architechture.

This changes `package.json` to explicitly target a version which is still available.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "babel-eslint": "^10.1.0",
     "babel-jest": "^28.1.2",
     "dotenv": "^16.0.3",
-    "esbuild": "^0.14.48",
+    "esbuild": "0.14.50",
     "esbuild-node-externals": "^1.5.0",
     "esbuild-plugin-copy": "^1.6.0",
     "eslint": "^8.24.0",


jackmcasey:
This was opened in error, apologies for the spam!

jackmcasey:
Having checked that this allows me to build, I'm going to re-open this. It hasn't caused any issues in my testing, but that has not been exhaustive.

I haven't been able to find out why 0.14.54 was pulled either.

dholms:
Looks like it's available again?

https://registry.yarnpkg.com/esbuild-linux-32/-/esbuild-linux-32-0.14.54.tgz

jackmcasey:
Could be a problem on my end. This dockerfile is still failing to build for me with the same error. When I go to that link in my browser, it works.

```
FROM node:18-alpine
RUN apk update
RUN apk add git

WORKDIR /app

RUN git clone https://github.com/bluesky-social/atproto/
WORKDIR /app/atproto

RUN yarn install
RUN yarn build

WORKDIR packages/dev-env
ENTRYPOINT ["yarn", "run", "start"]

EXPOSE 2582
EXPOSE 2583
```

jackmcasey:
Did some more digging, failures seem to be (almost) entirely when making the request using a digital ocean IP. Guess I'm tripping a firewall rule or something.

Since it's isolated to my environment, this is probably good to close out.

dholms:
Got it. 

Thanks for poking around at it. Strange to be hitting it on that version but not others 🤔

PR #500, opened 2023-01-27
Sync revamp migration fix

dholms:
This fixes an issue from the migration in https://github.com/bluesky-social/atproto/pull/446

The issue we ran into was that block ownership was not being correctly tracked. There were some blocks that were only tied to the first user to create them & then not tracked for other users.

This fix tries to grab all the blocks tied to a user but then can dynamically ask for blocks _not_ tied to the user if it needs to. It then does the migration as before, but also creates a link in `ipld_block_creator` for each of those blocks.

Relies on https://github.com/bluesky-social/atproto/pull/501


packages/pds/tests/migrations/repo-sync-data.test.ts
```diff
@@ -1,82 +1,153 @@
...
-    })
-    close = server.close
-    db = server.ctx.db
-    const client = AtpApi.service(server.url)
-    const sc = new SeedClient(client)
-    await basicSeed(sc, server.ctx.messageQueue)
+    if (process.env.DB_POSTGRES_URL) {
+      db = Database.postgres({
+        url: process.env.DB_POSTGRES_URL,
+        schema: 'migration_duplicate_records',
```
devinivy, line 22:
```suggestion
        schema: 'migration_repo_sync_data',
```

devinivy (approved):
👍 looking good

packages/pds/tests/migrations/repo-sync-data.test.ts
dholms: my god dude. one of these days ill get it right 🤦

Issue #502, opened 2023-01-30
Increase image-size constraints (post embeds, user profiles) and served resolutions


pfrazee:
We need to improve the quality of images a smidge more. Here's my proposal:

- `app.bsky.embed.external#external` thumb:
  - maxSize 300,000 ➡️ 600,000
  - maxWidth 1000 ➡️ 2000
  - maxHeight 1000 ➡️ 2000
- `app.bsky.embed.images` image:
  - maxSize 300,000 ➡️ 600,000
  - maxWidth 1000 ➡️ 2000
  - maxHeight 1000 ➡️ 2000
- `app.bsky.actor.profile` avatar:
  - maxSize 300,000 ➡️ 600,000
  - maxWidth 1000 ➡️ 2000
  - maxHeight 1000 ➡️ 2000
- `app.bsky.actor.profile` banner:
  - maxSize 500,000 ➡️ 750,000
  - maxWidth 3000 ➡️ 6000
  - maxHeight 1000 ➡️ 2000

This is roughly a 2x of all limits. For the image-proxy serving, we also 2x all the limits and we revert the thumb sizing away from "fit" (as that wrecks the ability to serve wide images).

I suggest if we want to consider any higher limits that we look closely at the costs -- not of image storage or bandwidth, but specifically of transfer latency in the federated network. The reason we're being stingy is to ensure that peoples' feeds load in a timely fashion when the extra hops of federation are introduced.

dholms:
I'm wondering if we should go even bigger for embeds? Like up to at least a MB. Just from a quick survey of twitter, they have a lot of images that are greater than 600 kb. As Jake pointed out, these are gonna be insubstantial in comparison to videos.

I expect most images to be served off of a gateway that does some processing, which means that the images only really have to make a couple hops: client -> pds -> BGS (which then serves it off the gateway).

From the gateway, we don't have to serve the full resolution image, but I think we should give the clients some more wiggle room in terms of what they upload.

pfrazee:
I'll let yall make the call here, just want to make sure we get it right. If you're not worried about the latency of 1mb then go for it.

dholms:
kk
@pfrazee wanted to check in on
> we revert the thumb sizing away from "fit"

what's the desired behavior here for thumbs? They're already set to `cover` which `=== fill`. Is that right? Or do we want `inside` which `=== fit`?

dholms:
[...]

PR #503, opened 2023-01-30
Loosening image constraints

dholms:
This loosens image constraints:
- doubles maxWidth & maxHeight on imgs 
- raises max size to 1MB

Closes https://github.com/bluesky-social/atproto/issues/502


devinivy (approved):
I think the final step is just to update some snapshots 👍

devinivy (approved):

PR #504, opened 2023-01-30
Xrpc streams

dholms:
adding subscriptions/streams to PDS


packages/xrpc-server/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@atproto/xrpc-server",
   "version": "0.0.1",
-  "main": "src/index.ts",
+  "main": "dist/index.js",
```
devinivy, line 4: Not sure what that was about!
```suggestion
  "main": "src/index.ts",
```

packages/xrpc-stream/package.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "@atproto/xrpc-stream",
+  "version": "0.0.1",
+  "main": "dist/index.js",
```
devinivy, line 4:
```suggestion
  "main": "src/index.ts",
```

packages/xrpc-stream/src/server.ts
```diff
@@ -0,0 +1,56 @@
...
+              })
+            })
+            throw new DisconnectError()
+          } else {
+            socket.send(frame.toBytes(), { binary: true })
+          }
+        }
+      } catch (err) {
+        if (err instanceof DisconnectError) {
+          return socket.close(err.code ?? CloseCode.Policy)
```
devinivy, line 29: We might consider passing along the error code from the error frame to the second arg of `socket.close()`.  Would likely involve `DisconnectError` taking a second constructor argument for it.

packages/xrpc-stream/src/stream.ts
```diff
@@ -0,0 +1,16 @@
+import { DuplexOptions } from 'stream'
+import { createWebSocketStream, WebSocket } from 'ws'
+import { Frame } from './frames'
+
+export async function* createFrameStream(
```
devinivy, line 5: Streams are async iterables, but async iterables aren't necessarily streams, so I may have misnamed this a bit.  It's hard to name, maybe something like `byFrame(ws)`?  🤔

packages/xrpc-server/src/server.ts
```diff
@@ -47,6 +67,18 @@ export class Server {
     }
   }
 
+  // @TODO redesign this api, streaming should not have to be enabled
+  enableStreaming(server: http.Server) {
+    server.on('upgrade', async (req, socket, head) => {
+      const url = new URL(req.url || '', 'http://x')
+      const sub = this.subscriptions.get(url.pathname.replace('/xrpc/', ''))
```
devinivy, line 74: This isn't exactly a bug, but it's also not working for the right reason.  We want the URL to look like `/xrpc/{nsid}`, but there's nothing here really checking that the pathname starts with `/xrpc`.  It happens to work because `{nsid}` on its own isn't a valid pathname since it doesn't start with a `/` 🙃

packages/xrpc-server/src/server.ts
```diff
@@ -200,6 +254,56 @@ export class Server {
...
+      new XrpcStreamServer({
+        noServer: true,
+        handler: async function* (req) {
+          try {
+            const url = new URL(req.url || '', 'http://x')
+            // authenticate request
+            const auth = await config.auth?.({ req })
+            if (isHandlerError(auth)) {
+              throw XRPCError.fromError(auth)
+            }
```
devinivy, line 276: There's another world in which auth (and even param validation?) happens during the websocket upgrade (i.e. in `enableStreaming()`) rather than after it.  The upside would be that you never establish the websocket, so it's less resource intensive overall.  I didn't do this because you have to write some raw HTTP headers directly to the socket, and it felt a little funky— this ended-up being easier and more straightforward to propagate the error info.  Also not sure exactly how a 400 or 401 during the websocket upgrade looks to clients, if we switch I would want to check that out.  Another consideration, if we were ever to run this protocol over TCP then it would probably all happen over the connection as we're doing here.

devinivy: Leaving this as-is for now, as I'm not getting fantastic signals around either the correctness or client support for erroring during the websocket handshake.  It's also awkward to code-up using the ws module, as it requires writing some raw http.  I do also kind of like that these errors happen in-band, since it adapts naturally to TCP, etc.

Refs:
https://stackoverflow.com/questions/21762596/how-to-read-status-code-from-rejected-websocket-opening-handshake-with-javascrip/50685387#50685387
https://bugs.webkit.org/show_bug.cgi?id=80362

packages/xrpc-server/src/stream/frames.ts
```diff
@@ -0,0 +1,122 @@
...
+        throw new Error('Too many CBOR data items in frame')
+      }
+      i++
+    })
+    const parsedHeader = frameHeader.safeParse(header)
+    if (!parsedHeader.success) {
+      throw new Error(`Invalid frame header: ${parsedHeader.error.message}`)
+    }
+    if (body === kUnset) {
+      throw new Error('Missing frame body')
```
dholms, line 47: do we ever have empty frame bodies? or would we send `null` in that case?

packages/xrpc-server/src/server.ts
```diff
@@ -200,6 +240,91 @@ export class Server {
...
+    const assertValidXrpcParams = (params: unknown) =>
+      this.lex.assertValidXrpcParams(nsid, params)
+    const resolveLexUri = (ref: string) => this.lex.resolveLexUri(nsid, ref)
+    this.subscriptions.set(
+      nsid,
+      new XrpcStreamServer({
+        noServer: true,
+        handler: async function* (req) {
+          try {
+            const url = new URL(req.url || '', 'http://x')
```
dholms, line 258: total nit, but can we move this down next to `decodeQueryParams` so its more obvious what its used for

packages/xrpc-server/src/server.ts
```diff
@@ -200,6 +240,91 @@ export class Server {
...
+                  yield new DataFrame({ type: code, body: clone })
+                } else {
+                  yield new DataFrame({ body: item })
+                }
+              } else {
+                yield new DataFrame({ body: item })
+              }
+            }
+          } catch (err) {
+            const xrpcErrPayload = XRPCError.fromError(err).payload
```
dholms, line 297: Looks like we might be leaking internal server error messages here?

packages/xrpc-server/src/stream/server.ts
```diff
@@ -0,0 +1,59 @@
...
+import { ErrorFrame, Frame } from './frames'
+import logger from './logger'
+
+export class XrpcStreamServer {
+  wss: WebSocketServer
+  constructor(opts: ServerOptions & { handler: Handler }) {
+    const { handler, ...serverOpts } = opts
+    this.wss = new WebSocketServer(serverOpts)
+    this.wss.on('connection', async (socket, req) => {
+      socket.on('error', (err) => logger.error(err, 'websocket error'))
```
dholms, line 12: just out of curiosity, what happens on an error here? does the socket keep going? does it terminate? is there any instance in which _we_ should be terminating?

packages/xrpc-server/tests/frames.test.ts
```diff
@@ -0,0 +1,171 @@
...
+    expect(messageFrame.header).toEqual({
+      op: FrameType.Data,
+      t: 2,
+    })
+    expect(messageFrame.op).toEqual(FrameType.Data)
+    expect(messageFrame.type).toEqual(2)
+    expect(messageFrame.body).toEqual({ a: 'b', c: [1, 2, 3] })
+
+    const bytes = messageFrame.toBytes()
+    expect([...bytes]).toEqual([
```
dholms, line 21: could do `uint8arrays.equal()` here to avoid the destructuring 

no biggie tho

packages/xrpc-server/src/stream/frames.ts
```diff
@@ -0,0 +1,122 @@
...
+    } else {
+      const exhaustiveCheck: never = frameOp
+      throw new Error(`Unknown frame op: ${exhaustiveCheck}`)
+    }
+  }
+}
+
+export class DataFrame extends Frame {
+  header: DataFrameHeader
+  constructor(opts: { type?: number; body: unknown }) {
```
dholms, line 80: is `type` optional here? 

I guess in the case of a stream with only one message type? 

part of me wants to stick it on their anyway just for consistency 😅

dholms, line 80: Also can we make the `body` type a _little_ bit more stringent with `Record<string, unknown>`?
Or are there cases where the body is not a record?

also total nit, but kinda want the constructor of this to have the `body` as the first arg

constructor(body: Record<string, unknown>, opts: { type: number })

dholms (commented):
✅ ✅ ✅ 

It won't let me approve because im the PR author lol 😅 

Mostly nits & questions.

This is supremely badass thanks for getting this together 😎 💪

packages/xrpc-server/src/stream/frames.ts
devinivy: You would send an empty CBOR data item as the body, which encodes as `f7` in hex.

packages/xrpc-server/src/server.ts
devinivy: I think we should be all set on that, it should be handled inside the `payload` getter.

packages/xrpc-server/src/stream/frames.ts
devinivy: The thought I was having is that it's only really useful for unions, and even in that case you don't absolutely need it because the fallback for any unmapped `$type` is to put it in the message body rather than in the header.  So it's more of an optimization.  But if we do want to include it, we'd need to come-up with a convention for how to map the single type since it doesn't have an nsid.  Would it just be `0`?  Does that maybe mess things up if you change it to a union later?

devinivy: The original thought was that the wire protocol doesn't need to put any limits on what you can transfer with it, as long as it encodes as CBOR (strings, numbers, arrays, raw bytes).  Then at the lexicon layer maybe it gets whittled down to whatever you can actually spec for a message, possibly just a record.  One way to split the difference here might be to make this generic with a default parameter of `Record<string, unknown>`.  I could see the generic here being pretty useful anyway:
```ts
export class DataFrame<T = Record<string, unknown>> extends Frame {
  header: DataFrameHeader
  constructor(opts: { type?: number; body: T }) {
```

packages/xrpc-server/src/stream/server.ts
devinivy: Ooooh good call.  Since this is a `net.Socket`, it should close itself immediately after the error, so I think we're alright.  https://nodejs.org/api/net.html#event-error_1

packages/xrpc-server/src/stream/frames.ts
devinivy: One thing we could do is default it to `-1`, and say that mapped types are non-negative.

packages/xrpc-server/src/stream/frames.ts
devinivy: (I needed `kUnset` here since `f7` decodes to `undefined` 😆 )

packages/xrpc-server/src/stream/server.ts
devinivy: Oops, this was correct but for the wrong reason. When double-checking I see that it's _not_ an extension of net.Socket as I thought, but luckily it does have the same behavior here 👍 https://github.com/websockets/ws/issues/1823#issuecomment-740056036

packages/xrpc-server/src/stream/frames.ts
dholms: kk that's what i figured 👍

packages/xrpc-server/src/stream/server.ts
dholms: awesome thanks for the check 🙏

packages/xrpc-server/src/stream/frames.ts
dholms: Hmmm yeah I take that point. I don't have a strong opinion. Down to leave it as is for simplicities sake 👌

PR #506, opened 2023-01-30
Maintain a cache of session-clients for a service and reuse them

pfrazee:
This PR updates the API so that

```typescript
import {sessionClient} from '@atproto/api`

sessionClient.service("https://foo.com") === sessionClient.service("https://foo.com")
```

The reason for this change is to ensure that the access tokens used to control the sessions are unified across all clients. This is useful because a refresh token will invalidate if it is ever used, meaning a client needs to be careful to only attempt a refresh just once. The social app has to create multiple instances for various reasons and coordinating the sessions between them was proving too difficult. Caching and reusing instances helps to ensure no mistakes will occur.

### DIFF
--- a/packages/api/src/session.ts
+++ b/packages/api/src/session.ts
@@ -20,10 +20,26 @@ const DELETE_SESSION = 'com.atproto.session.delete'
 const CREATE_ACCOUNT = 'com.atproto.account.create'
 
 export class SessionClient extends Client {
+  private cache: Map<string, SessionServiceClient> = new Map()
+
   service(serviceUri: string | URL): SessionServiceClient {
+    const cacheKey = toCacheKey(serviceUri)
+    const inst = this.cache.get(cacheKey)
+    if (inst) {
+      return inst
+    }
     const xrpcService = new SessionXrpcServiceClient(this.xrpc, serviceUri)
-    return new SessionServiceClient(this, xrpcService)
+    const service = new SessionServiceClient(this, xrpcService)
+    this.cache.set(cacheKey, service)
+    return service
   }
+}
+
+function toCacheKey(uri: string | URL) {
+  if (typeof uri === 'string') {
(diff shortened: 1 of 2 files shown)


packages/api/tests/session.test.ts
```diff
@@ -236,4 +236,8 @@ describe('session', () => {
     expect(sessions[1]).toEqual(undefined)
     expect(client.sessionManager.active()).toEqual(false)
   })
+
+  it('reuses the client object across calls', () => {
+    expect(client).toEqual(sessionClient.service(server.url))
```
devinivy, line 241: I think we want `toBe` here, which checks for shallow equality (`===`).
```suggestion
    expect(client).toBe(sessionClient.service(server.url))
```

packages/api/src/session.ts
```diff
@@ -20,10 +20,26 @@ const DELETE_SESSION = 'com.atproto.session.delete'
...
+    this.cache.set(cacheKey, service)
+    return service
+  }
+}
+
+function toCacheKey(uri: string | URL) {
+  if (typeof uri === 'string') {
+    uri = new URL(uri)
   }
+  return uri.hostname
```
devinivy, line 42: It's unlikely to come-up, but `uri.host` might be marginally better since it takes into account the port.  An example might be different services running on `localhost`, e.g. in tests.

devinivy: If you want:
```suggestion
  return uri.host
```

devinivy (approved):
Cool cool!

pfrazee:
Actually need a sec -- hitting some complexities that suggest it might be better to do this outside of the actual api impl

pfrazee:
Okay turns out this wasn't the ideal solution, even though the root cause was correctly identified. I've updated the app to be more careful in how it handles session-oriented service instances and that solved it.

PR #511, opened 2023-01-31
Incorporate blobs into moderation actions

devinivy:
- When issuing a moderation action against a record, you may specify relevant blobs associated with the given record.
 - Admin views of records and moderation actions now include blob details.
 - Apply takedown to blobs:
   - stop serving it.
   - invalidate it in image caches.
   - disallow records from referencing the same cid going forward.


bnewbold:
Trying to think through a couple corner cases while poking through the diff:

* does a non-"takedown" action (eg, "flag" or "acknowledge") result in takedown behavior if blob CIDs are included?
* if not, consider if there is a non-"takedown" report+action ("flag" or "acknowledge") on a record that indicates CIDs, and then later a separate report+action on the same records and CIDs, do the blobs get taken down? I think they get quarantined and taken down, but the takedown might not persist as metadata on the blob for future re-uploads?
* reverting an action on a record w/ subject CIDs results in the blobs getting globally un-quarantined. Is this fine, or should we check if there any other still-enforced takedowns on the same CID before un-quarantining?
* I think there is also a possible tangled situation for posts with multiple images where one action reference one image, then another action references all images, then the first action is reverted, but the other action isn't, and the takedown state is lost on the "one" image.

These are all pretty corner-case-y, and I haven't confirmed with testing or re-reading the code if they are real issues. The core and most important and "likely" workflow functionality seems to be correct.

packages/pds/src/services/moderation/index.ts
```diff
@@ -215,13 +272,40 @@ export class ModerationService {
...
+        .where('takedownId', 'is', null)
+        .executeTakeFirst()
+      await Promise.all(
+        info.blobCids.map(async (cid) => {
+          await this.blobstore.quarantine(cid)
+          const paths = ImageUriBuilder.commonSignedUris.map((id) => {
+            const uri = this.imgUriBuilder.getCommonSignedUri(id, cid)
+            return uri.replace(this.imgUriBuilder.endpoint, '')
+          })
+          await this.imgInvalidator.invalidate(cid.toString(), paths)
```
dholms, line 305: one thing i could see here is that an error thrown by either the blobstore or the invalidator would revert the txn & prevent the takedown of some record.

this is good for maintaining a consistent state but arguably bad if the goal is "stop showing this post with as little chance for hiccups as possible"

packages/pds/src/api/com/atproto/admin/takeModerationAction.ts
```diff
@@ -17,7 +17,7 @@ export default function (server: Server, ctx: AppContext) {
     handler: async ({ input }) => {
       const { db, services } = ctx
       const moderationService = services.moderation(db)
-      const { action, subject, reason, createdBy } = input.body
+      const { action, subject, reason, createdBy, subjectBlobCids } = input.body
```
dholms, line 20: Just to make sure I have it right:

Only 1 takedown can exist against a record at a time, right?

So say you have a record with `uri1` with blobs `cid1` & `cid2`.
A takedown is issued on `uri1` & `cid1`, then a second takedown is issued on `uri1`, `cid1` & `cid2`. The latter overrides the former right?

dholms: Part of me wonders if subject cids should not be provided by the client & are just looked up in the DB when the request is made.

Like does it ever make sense to "take down post & 1 image, but leave the other"?

dholms (approved):
this looks good 👍 

our convo in slack earlier answered all my questions

devinivy:
Ended-up making a few additional changes to solidify things:
 - Enforcing one active action per repo/record/blob.
 - Change `moderation.takedownId` to `moderation.currentAction` on repo and record views, including both the id and the action taken.
 - Add `moderation.currentAction` to blob view.

dholms (approved):
👌 🎉

PR #514, opened 2023-02-02
Database notify/listen

dholms:
Adding NOTIFY / LISTEN support to our database wrapper.

There are two cases:
- sqlite: this is a database that _only one_ application will be interacting with at any point in time. Therefore we handle notify/listen with in-process callbacks
- postgres: this is a database that _one or more_ applications may be interacting with. Therefore we handle postgres' [NOTIFY](https://www.postgresql.org/docs/current/sql-notify.html) feature

### DIFF
--- a/packages/common/src/async.ts
+++ b/packages/common/src/async.ts
@@ -1,9 +1,9 @@
-export type Defferable = {
+export type Deferrable = {
   resolve: () => void
   complete: Promise<void>
 }
 
-export const createDeferrable = (): Defferable => {
+export const createDeferrable = (): Deferrable => {
   let resolve
   const promise: Promise<void> = new Promise((res) => {
     resolve = () => res()
@@ -11,14 +11,14 @@ export const createDeferrable = (): Defferable => {
   return { resolve, complete: promise }
 }
 
-export const createDeferrables = (count: number): Defferable[] => {
-  const list: Defferable[] = []
+export const createDeferrables = (count: number): Deferrable[] => {
+  const list: Deferrable[] = []
   for (let i = 0; i < count; i++) {
     list.push(createDeferrable())
   }
   return list
 }
 
-export const allComplete = async (defferables: Defferable[]): Promise<void> => {
-  await Promise.all(defferables.map((d) => d.complete))
+export const allComplete = async (deferrables: Deferrable[]): Promise<void> => {
+  await Promise.all(deferrables.map((d) => d.complete))
 }
--- a/packages/common/src/async.ts
+++ b/packages/common/src/async.ts
@@ -1,0 +1,24 @@
+export type Defferable = {
+  resolve: () => void
+  complete: Promise<void>
+}
+
+export const createDeferrable = (): Defferable => {
+  let resolve
+  const promise: Promise<void> = new Promise((res) => {
+    resolve = () => res()
+  })
+  return { resolve, complete: promise }
+}
+
+export const createDeferrables = (count: number): Defferable[] => {
+  const list: Defferable[] = []
+  for (let i = 0; i < count; i++) {
+    list.push(createDeferrable())
+  }
+  return list
+}
+
+export const allComplete = async (defferables: Defferable[]): Promise<void> => {
+  await Promise.all(defferables.map((d) => d.complete))
+}
--- a/packages/pds/package.json
+++ b/packages/pds/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@atproto/pds",
   "version": "0.0.2",
-  "main": "dist/index.js",
+  "main": "src/index.ts",
   "bin": "dist/bin.ts",
   "scripts": {
     "codegen": "lex gen-server ./src/lexicon ../../lexicons/com/atproto/*/* ../../lexicons/app/bsky/*/*",
--- a/packages/pds/package.json
+++ b/packages/pds/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@atproto/pds",
   "version": "0.0.2",
-  "main": "src/index.ts",
+  "main": "dist/index.js",
   "bin": "dist/bin.ts",
   "scripts": {
     "codegen": "lex gen-server ./src/lexicon ../../lexicons/com/atproto/*/* ../../lexicons/app/bsky/*/*",
@@ -54,6 +54,7 @@
     "pino": "^8.6.1",
     "pino-http": "^8.2.1",
     "sharp": "^0.31.2",
+    "typed-emitter": "^2.1.0",
     "uint8arrays": "3.0.0"
   },
   "devDependencies": {
--- a/packages/pds/src/db/index.ts
+++ b/packages/pds/src/db/index.ts
@@ -6,12 +6,13 @@ import {
   PoolClient as PgPoolClient,
   types as pgTypes,
 } from 'pg'
+import EventEmitter from 'events'
+import TypedEmitter from 'typed-emitter'
 import DatabaseSchema, { DatabaseSchemaType } from './database-schema'
 import { dummyDialect } from './util'
 import * as migrations from './migrations'
 import { CtxMigrationProvider } from './migrations/provider'
-import EventEmitter from 'events'
-import TypedEmitter from 'typed-emitter'
+import { dbLogger as log } from '../logger'
 
 export class Database {
   channels: Channels = {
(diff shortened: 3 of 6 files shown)


packages/pds/src/db/index.ts
```diff
@@ -114,9 +163,23 @@ export default Database
 
 export type Dialect = 'pg' | 'sqlite'
 
+export type DialectFacets = PgFacets | SqliteFacets
```
dholms, line 166: lmk if you have a better name than 'facets' for this. couldn't think of anything 😅

packages/pds/tests/db-notify.test.ts
```diff
@@ -0,0 +1,67 @@
...
+      received.push(msg || '')
+    })
+
+    dbTwo.notify('otherchannel', 'blah')
+    for (const msg of toSend) {
+      dbTwo.notify('test', msg)
+    }
+    dbTwo.notify('otherchannel', 'blah')
+
+    await wait(200)
```
dholms, line 43: don't love just throwing in an arbitrary wait here, but not sure what else to do 🤷

packages/pds/src/db/index.ts
devinivy: Only idea I have is to call it something boring like `DbConfig`, and there's essentially a different config discriminated by `config.dialect`.

packages/pds/src/db/index.ts
```diff
@@ -55,20 +58,61 @@ export class Database {
...
+    return new Database(db, { dialect: 'pg', pool, schema })
   }
 
   static memory(): Database {
     return Database.sqlite(':memory:')
   }
 
+  notify(channel: string, msg?: string) {
+    if (this.facets.dialect === 'pg') {
+      this.facets.pool.query(`NOTIFY ${channel}, '${msg}'`)
```
devinivy, line 70: Would be nice to escape `channel` and `msg` to keep SQLi far-off.  Could we use something like this?
```ts
await sql`NOTIFY ${ref(channel)}, ${msg}`.execute(this.db)
```

packages/pds/src/db/index.ts
```diff
@@ -55,20 +58,61 @@ export class Database {
...
+    } else {
+      const cbs = this.listeners[channel]
+      if (cbs && cbs.length > 0) {
+        for (const cb of cbs) {
+          cb(msg)
```
devinivy, line 75: It's a super minor thing, but to avoid timing inconsistencies it might be worth putting this behind a `process.nextTick(cb, msg)`.  Here's an example:
```ts
let item = 'a'
await db.listenFor('x', (msg) => {
  item = msg
})
db.notify('x', 'b')
// It's nice if it's well-defined whether item is 'a' or 'b' here.
```

packages/pds/src/db/index.ts
```diff
@@ -1,23 +1,26 @@
...
+  PoolClient as PgPoolClient,
+  types as pgTypes,
+} from 'pg'
 import DatabaseSchema, { DatabaseSchemaType } from './database-schema'
 import { dummyDialect } from './util'
 import * as migrations from './migrations'
 import { CtxMigrationProvider } from './migrations/provider'
 
 export class Database {
+  listeners: Record<string, ListenerCB[]> = {}
```
devinivy, line 15: I wonder if some code could disappear if this were an event-emitter.  Might find that it's easier to type that way too, with [typed-emitter](https://github.com/andywer/typed-emitter).

packages/pds/src/db/index.ts
```diff
@@ -55,20 +58,61 @@ export class Database {
...
+        }
+      }
+    }
+  }
+
+  async listenFor(channel: string, cb: ListenerCB) {
+    if (this.facets.dialect === 'pg') {
+      if (!this.notifyClient) {
+        this.notifyClient = await this.facets.pool.connect()
+        this.notifyClient.query(`LISTEN ${channel}`)
```
devinivy, line 85: May be missing something, but I _think_ (?) you can nix this thanks to the other one down on line 95.
```suggestion
```

packages/pds/src/db/index.ts
```diff
@@ -55,20 +58,61 @@ export class Database {
...
+          cb(msg)
+        }
+      }
+    }
+  }
+
+  async listenFor(channel: string, cb: ListenerCB) {
+    if (this.facets.dialect === 'pg') {
+      if (!this.notifyClient) {
+        this.notifyClient = await this.facets.pool.connect()
```
devinivy, line 84: We might want some reconnection logic for the notify client to really ensure reliability.

packages/pds/tests/db-notify.test.ts
devinivy: Has a different drawback, where if the test is failing it could hang, but one way to slice this is with a little home-spun deferrable.

```ts
    const deferrables = [createDeferrable(), createDeferrable()]
    await dbOne.listenFor('test', (msg) => {
      deferrables[received.length].resolve()
      received.push(msg || '')
    })
    // ...
    await Promise.all(deferrables.map((d) => d.complete))

// ...
const createDeferrable = () => {
  let resolve
  const promise = new Promise((res) => {
    resolve = res
  })
  return { resolve, complete: promise }
}
```

packages/pds/src/db/index.ts
```diff
@@ -78,6 +122,11 @@ export class Database {
   }
 
   async close(): Promise<void> {
+    this.notifyClient?.removeAllListeners()
+    this.notifyClient?.release()
+    if (this.facets.dialect === 'pg') {
+      await this.facets.pool.end()
+    }
```
devinivy, line 129: I have feeling `db.destroy()` might do this for you (though assuming that `db` shares the same pool).

devinivy (approved):
Feels nice to have this working! 👍 Dropped a couple thoughts, but it's only really the a. escaping and b. reconnection questions that are top of mind for me.

packages/pds/src/db/index.ts
dholms: hah i had that same feeling but got hanging tests 🙃

packages/pds/src/db/index.ts
dholms: yeah i was feeling this same tension as i started making use of it. gonna take a stab at rewriting w emitters 👌

packages/pds/src/db/index.ts
dholms: 👍

packages/pds/src/db/index.ts
devinivy: Oh, huh!  Well either way, I don't think it can do any harm :)

packages/pds/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@atproto/pds",
   "version": "0.0.2",
-  "main": "src/index.ts",
+  "main": "dist/index.js",
```
devinivy, line 4: I expect some script that runs on install must be touching these.
```suggestion
  "main": "src/index.ts",
```

packages/pds/src/db/index.ts
dholms: I don't think kysely supports NOTIFY so I can't use those unfortunately :/

However I actually don't need the payload anywhere, so I took that out & limited the channels you can send on which keeps us clear of any danger

packages/pds/src/db/index.ts
dholms: stuck some reconnection logic in there that is fairly basic, but i had a rough time trying to test it 🙃

packages/pds/tests/db-notify.test.ts
dholms: yeah nice this feels better 👍

packages/pds/src/db/index.ts
devinivy: Ah, nice!  I didn't think kysely had to explicitly support it in order to be able use it with the raw sql builder, but I also haven't tried this out!

packages/common/src/async.ts
```diff
@@ -0,0 +1,24 @@
+export type Defferable = {
```
devinivy, line 1: Not too hung-up on it, but tiny typo in here.
```suggestion
export type Deferrable = {
```

devinivy (approved):
Sweet!

packages/common/src/async.ts
dholms: oh woops good catch 👍

packages/pds/src/db/index.ts
dholms: yeah not sure. i tried it in a few different shapes and couldn't get it to like it. so i just did it this way. 

If we find we need payloads in the future, we can reassess it

PR #520, opened 2023-02-02
Account deletion hotfix

dholms:
We were doing a `SELECT WHERE IN` on an empty list which was causing a postgres syntax error.

This applies a fix that checks for empty arrays & a simple test that deletes an empty account

### DIFF
--- a/packages/pds/src/db/index.ts
+++ b/packages/pds/src/db/index.ts
@@ -127,9 +127,10 @@ export class Database {
   async close(): Promise<void> {
     this.channelClient?.removeAllListeners()
     this.channelClient?.release()
-    if (this.cfg.dialect === 'pg') {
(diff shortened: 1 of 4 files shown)


packages/pds/src/db/index.ts
```diff
@@ -127,9 +127,10 @@ export class Database {
   async close(): Promise<void> {
     this.channelClient?.removeAllListeners()
     this.channelClient?.release()
-    if (this.cfg.dialect === 'pg') {
-      await this.cfg.pool.end()
-    }
+    // @TODO investigate
```
dholms, line 130: I snuck this in here, because it was causing tests to fail for some reason. Don't remember it causing issues on my branch 🤔 

Needs further investigation

devinivy (approved):

PR #521, opened 2023-02-03
Refactor repo indexing

dholms:
This mostly just moves things around in the PDS's repo indexing in the hopes of avoiding a few footguns.

Using it in https://github.com/bluesky-social/atproto/pull/518 where I was adding another step to processing & realizing it could be very easily glossed over & cause some footguns

Also refigures the `@atproto/repo` package just a bit to knock out a `@TODO` that's been there for some time


devinivy (approved):
Good stuff 👍

PR #524, opened 2023-02-03
Handle client abort on image upload

devinivy:
This handles an issue with propagating stream errors during image upload, i.e. if a client aborts their request during the upload.


dholms (approved):
awesome! thanks for knocking this out so quick 🙇

PR #525, opened 2023-02-04
Proposed com.atproto.label.label lexicon

bnewbold:
For persisting un-signed label objects in repositories. There will likely be a handful of XRPC query endpoints in the com.atproto.label namespace in the near future.

Haven't run JS/TS codegen tools locally for this.

The schema with short key names matches the earlier label proposal, and makes sense in the context of free-floating label objects (signed or unsigned), eg as provided in an event stream. But it doesn't fit in as elegantly with current Lexicon style. Not sure what the right thing to do is there.

In theory the `labelUri` field, which should point back to the location of this record, is duplicate information when stored in an actual repo. I kept it in the record lexicon so we can still use this type for event streams. Agnostic as to whether that field should be populated when actually stored in a repo.

I added `maxLength` on a bunch of these out of habit of being conservative, but probably should not. My thinking was that these limits could always be increased (but never constrained further), but I think with lexicons we don't actually have that kind of flexibility in the ecosystem. What I would sort of want is "don't produce records larger than X, but do accept records from the network up to Y", which would be a bit more future-proof. CBOR itself does give us some reasonable constraints, so a griefer couldn't just dump 1 GByte strings in these fields, I think.

### DIFF
--- a/lexicons/com/atproto/label/label.json
+++ b/lexicons/com/atproto/label/label.json
@@ -1,0 +1,23 @@
+{
+  "lexicon": 1,
+  "id": "app.bsky.label.label",
+  "defs": {
+    "main": {
+      "type": "record",
+      "description": "Metadata tag on an atproto resource (eg, repo or record)",
+      "key": "tid",
+      "record": {
+        "type": "object",
+        "required": ["src", "uri", "val", "ts"],
+        "properties": {
+          "src": {"type": "string", "maxLength": 1024},
+          "uri": {"type": "string", "maxLength": 1024},
+          "cid": {"type": "string", "maxLength": 256},
+          "val": {"type": "string", "maxLength": 128},
+          "ts": {"type": "datetime"},
+          "labelUri": {"type": "string", "maxLength": 1024}
+        }
+      }
+    }
+  }
+}
--- a/lexicons/com/atproto/label/label.json
+++ b/lexicons/com/atproto/label/label.json
@@ -8,15 +8,39 @@
       "key": "tid",
       "record": {
         "type": "object",
-        "required": ["src", "uri", "val", "ts"],
+        "required": ["src", "uri", "val", "cts"],
         "properties": {
-          "src": {"type": "string", "maxLength": 1024},
-          "uri": {"type": "string", "maxLength": 1024},
-          "cid": {"type": "string", "maxLength": 256},
-          "val": {"type": "string", "maxLength": 128},
-          "ts": {"type": "datetime"},
-          "labelUri": {"type": "string", "maxLength": 1024}
+          "src": {
+            "type": "string",
+            "description": "DID of the actor who created this label"
+          },
+          "uri": {
+            "type": "string",
+            "description": "AT URI of the record, repository (account), or other resource which this label applies to"
+          },
+          "cid": {
+            "type": "string",
+            "description": "optionally, CID specifying the specific version of 'uri' resource this label applies to"
+          },
+          "val": {
+            "type": "string",
+            "maxLength": 128,
+            "description": "the short string name of the value or type of this label"
+          },
+          "cts": {
+            "type": "datetime",
+            "description": "timestamp when this label was created"
+          }
         }
+      }
+    },
+    "labelEvent": {
+      "type": "object",
+      "required": ["uri", "cid", "label"],
+      "properties": {
+          "uri": {"type": "string"},
+          "cid": {"type": "string"},
+          "label": {"type": "ref", "ref": "app.bsky.label.label#main"}
       }
     }
   }
--- a/lexicons/com/atproto/label/label.json
+++ b/lexicons/com/atproto/label/label.json
@@ -34,7 +34,7 @@
         }
       }
     },
-    "labelEvent": {
+    "view": {
       "type": "object",
       "required": ["uri", "cid", "label"],
       "properties": {


lexicons/com/atproto/label/label.json
```diff
@@ -0,0 +1,23 @@
...
+      "record": {
+        "type": "object",
+        "required": ["src", "uri", "val", "ts"],
+        "properties": {
+          "src": {"type": "string", "maxLength": 128},
+          "uri": {"type": "string", "maxLength": 1024},
+          "cid": {"type": "string", "maxLength": 128},
+          "val": {"type": "string", "maxLength": 128},
+          "ts": {"type": "datetime"},
+          "labelUri": {"type": "string", "maxLength": 1024}
```
devinivy, line 18: I saw your note on this, hmm!!  My gut feeling is to remove `labelUri` here, and if we need another lexicon for the streaming event then that's no biggie.  In my view, the streaming event lexicons can be treated pretty much just the same as any other XRPC lexicon— the difference of responses being an asynchronous stream of values rather than a single value doesn't seem like it has a major impact on how they should be spec'ed in lexicon, at least the way I am currently thinking about it.

Also coming to mind, if we want these records to be verifiable then we need to present them as-is over the stream so that anyone can get their canonical bytes and cid.  So if it's unspec'ed whether the record is stored with `labelUri`, but over the stream it comes with `labelUri`, then the consumer wouldn't be able to figure out the exact bytes for the record.

lexicons/com/atproto/label/label.json
```diff
@@ -0,0 +1,23 @@
...
+  "defs": {
+    "main": {
+      "type": "record",
+      "description": "Metadata tag on an atproto resource (eg, repo or record)",
+      "key": "tid",
+      "record": {
+        "type": "object",
+        "required": ["src", "uri", "val", "ts"],
+        "properties": {
+          "src": {"type": "string", "maxLength": 128},
```
devinivy, line 13: Intuitively this seems like a reasonable number, but if there's a `maxLength` to bump up higher it might be this one.  For did:plc this length is good, but for other did methods (in particular did:web), it's possible we'd want to give some more leeway.

lexicons/com/atproto/label/label.json
dholms: agreed on leaving it off here but including it over the wire 👍

lexicons/com/atproto/label/label.json
```diff
@@ -0,0 +1,23 @@
...
+          "uri": {"type": "string", "maxLength": 1024},
+          "cid": {"type": "string", "maxLength": 128},
+          "val": {"type": "string", "maxLength": 128},
+          "ts": {"type": "datetime"},
```
dholms, line 17: we've been using `createdAt` for this.

i think it helps differentiate because there are several different "timestamps" in the system
- createdAt
- indexedAt
- (not to be used as a timestamp) but possibly the cursor on the event stream
- (again, not to be used as a timestamp), but the TID

lexicons/com/atproto/label/label.json
bnewbold: Yeah, makes sense.

For over-the-wire, do you think it would be best to create a "labelEvent" additional def here in this lexicon which has two fields: `label (com.atproto.label.label#main)` and `ref (com.atproto.repo.recordRef)` (or maybe `labelRef`)? Or, to be more similar to `#view` and `#viewDetail` in `com.atptroto.admin.moderationAction`, the "labelEvent" type could be basically the same as "label" but with the `labelUri` included. Still getting working through Lexicon schema def features and style.

lexicons/com/atproto/label/label.json
bnewbold: I agree that "timestamp" can mean lots of things, it is important to clarify which meaning, and the name of the field is a good way to do that. And to be consistent with other lexicons.

(as a side-reference, scholarly publishing APIs can turn in to rats nest of different "publication times" and "update times": https://github.com/CrossRef/rest-api-doc#sorting)

I went with `ts` here to be consistent with the proposed free-floating label schema, where I feel like the byte savings might be worth it in the context of many labels getting stapled to records all over the place. Also why the other fields are short. Maybe `cts` ("created timestamp") or `cat` ("created at")? The JWT 3-char field name is `iat` ("issued at").

lexicons/com/atproto/label/label.json
devinivy: Yeah, those all feel in the ballpark for me 👍 On notifications, post, and profile views you often see something like this, which might adapt nicely here: https://github.com/bluesky-social/atproto/blob/e8c218b5c90391cd6402c43a532fc398573aa664/lexicons/app/bsky/actor/updateProfile.json#L39-L50

lexicons/com/atproto/label/label.json
bnewbold: I bumped `src` length from 128 to 1024, and CID from 128 to 256 for wiggle room.
I'm sort of disposed to just not add these length restrictions here at all, unless we decide to also add in other places (like `com.atproto.repo.strongRef`). Maybe as part of a "hardening" last pass over these lexicons.

lexicons/com/atproto/label/label.json
pfrazee: I definitely get the logic of the smaller labels but it's different than the naming conventions we've used with all the others, so if we change that up we should make sure it's with a specific intent

pfrazee:
Two thoughts:

- I'm not totally clear on the purposes of the fields; some description fields on them would definitely help.
- Not gonna make a huge fuss about it but I wonder if we could avoid the `label.label` action in the NSID

bnewbold:
I rebased and made some changes. In the big picture, I don't think this needs to be re-reviewed or merged before the general pass over Lexicons, though perhaps it should be kept in mind.

Changed:
- added descriptions
- removed maxLength on DID, CID, and ATURI fields (hoping those field types will end up with native lexicon validation; regardless should be consistent with other record types on that question)
- renamed 'ts' to 'cts' ("created timestamp"). Don't think this really resolves the "short field names" issue, but does clarify the purpose of the field
- broke out an "event" object variant with uri+cid info, and removed `labelUri` field from the record itself

`label.label` does look wrong but does also seem most correct to me right now. Some alternatives:

- `com.atproto.repo.label`
- `com.atproto.record.label`
- `com.atproto.label.record`
- `com.atproto.annotation.label`
- `com.atproto.metadata.label` or `com.atproto.meta.label`

None of these feel as "correct" as `label.label`. Maybe something will come out of the Lexicon review discussion, or we could look back to the Role & Services naming structure.

lexicons/com/atproto/label/label.json
bnewbold: I'm going to mark this resolved. I used the structure of this last `updateProfile` example, except I named the record field `label` and set the type explicitly as a ref (not `unknown`).

bnewbold:
To signpost next steps on labeling and lexicons:

- I will send proposal to include labels on labels on post and actor views, in endpoints like getTimeline, threads, author feeds, notification list
- I will send proposal for label-specific XRPC endpoints including `com.atproto.label.subscribeAllLabels` (event stream schema) and `com.atproto.label.*` endpoints to query labels by subject ATURI (including records and repos)

Both of these changes would refer back to this record type. One small change I could make to this record type is to rename `labelEvent` to `labelView`, which is a better name in the context of non-event-stream use (eg, embedding this record in other response).

Coming down the pipeline this week is lexicon refactor, which will enable string type annotations on some of the fields in this schema.

Curious for review/feedback on any blocking issues on this schema. Eg, resolution on the short field names, and whether `com.atproto.label.*` is the appropriate namespace (see alternatives in my last comment on this issue).

dholms:
Thanks for putting this back on my radar.

That all sounds good as far as next steps go 👍

No issues with this besides the bikesheds. And I'm not terribly hung up on either of them. So I'll give my thoughts, but I'll leave it up to you.

On the short field names: I'm a bit worried it's over-optimized & loses some clarity. But I also understand the perspective that these things are gonna be getting sprayed all over the place & the byte savings could actually matter. If we stick with short 3 letter names, it could be nice to borrow from preexisting semantics (like jwts) & use something like `iat` & maybe `iss` instead of `src` 

There's a part of me that wishes we'd done `.record` for all of our record NSIDs, but we didn't 😕 I suppose we still could without breaking from convention too much since this is a different namespace (all the other records are in `app.bsky`.
For vanities sake & because devs will be exposed to this so much, it would feel good to avoid the ugly `label.label`. My favorite of the ones you listed is `com.atproto.meta.label`. Another one could be `com.atproto.context.label` as in "this whole system is the 'context' system that gives additional context to objects in the network. Context can be given with labels"

dholms (approved):
comments above & happy to continue the convo, but overall looks good 👍

bnewbold:
I'm going to close this in lieu of https://github.com/bluesky-social/atproto/pull/672, which is on top of this one. Sorry to have split the conversation, there are a bunch of comments on both issues.

PR #529, opened 2023-02-07
Refactor @atproto/api to the `AtpAgent` interface

pfrazee:
This PR is meant to fix a couple issues with the API that had been surfacing in the app:

- The control flow around sessions was a little difficult to handle. This refactor tries to take sharp edges off the API so I (and others) don't make the same mistakes.
- The API package DX was a little confusing with the "client" and "sessionClient" exports. This basically unifies it around one main export, `AtpAgent`, and puts more opinion into the API around session management (notably introducing login, createAccount, and resumeSession methods).
- Monkey-patching the fetch handler is needed in react-native, and this refactor makes that pattern more obvious.

@devinivy you'll notice that I changed where in the abstraction-stack the API introduces session-token management. You previously had wrapped the xrpc `call()` in order to take advantage of its higher-level semantics. That made sense but I felt like the patch could be more concise if I did it at the `fetch()` level. Lmk if you think I missed anything; I also dropped a couple of niceties in an effort to simplify its implementation.


packages/api/src/agent.ts
```diff
@@ -0,0 +1,298 @@
...
+   */
+  async login(
+    opts: AtpAgentLoginOpts,
+  ): Promise<ComAtprotoSessionCreate.Response> {
+    try {
+      this.session = undefined
+      const res = await this.api.com.atproto.session.create({
+        identifier: opts.identifier,
+        password: opts.password,
+      })
```
devinivy, line 116: Is it all good that `this.session` is temporarily out of sync with `persistSession()` here?  I see that in the `finally {}` block it's guaranteed to eventually land, really just trying to map out how consumers using `persistSession()` should think about `this.hasSession` being `false` but not being notified of it yet.

pfrazee:
Yeah good catch, I’ll tweak that to avoid effects being visible prematurely

packages/api/src/agent.ts
```diff
@@ -0,0 +1,298 @@
...
+    }
+  }
+
+  /**
+   * Resume a pre-existing session with this agent.
+   */
+  async resumeSession(
+    session: AtpSessionData,
+  ): Promise<ComAtprotoSessionGet.Response> {
+    this.session = session
```
devinivy, line 139: This seems a lot more natural than how the session client would have you do it 👍

packages/api/src/agent.ts
```diff
@@ -0,0 +1,298 @@
...
+   * - Wraps the actual implementation in a promise-guard to ensure only
+   *   one refresh is attempted at a time.
+   */
+  private async _refreshSession() {
+    if (this._refreshSessionPromise) {
+      return this._refreshSessionPromise
+    }
+    this._refreshSessionPromise = this._refreshSessionInner()
+    await this._refreshSessionPromise
+    this._refreshSessionPromise = undefined
```
devinivy, line 223: Do we want to clear this even if `this._refreshSessionPromise` rejects?

pfrazee: Bleh yes of course. Aaaaand now I'm realizing I've got that mistake in multiple places in the app repo! Hooray!

devinivy (approved):

packages/api/src/agent.ts
pfrazee: Fixed

dholms (approved):
nice this looks slick 👌

PR #532, opened 2023-02-08
Get profiles xrpc method

devinivy:
Implements `app.bsky.actor.getProfiles`:
 - Receives up to 25 `actors` via params, which may be handles or dids.
 - Responds with profiles for any actors that exist.
 - Maintains the order of the results to match the order of the `actors` param.
 - Response shape is `{ profiles: ProfileView[] }`.
 - Also rearranged the profile view schema into `app.bsky.actor.profile#view`, similar to how `app.bsky.feed.post#view` is setup.


pfrazee:
API lgtm, will let danny review impl

packages/pds/src/api/app/bsky/actor/getProfile.ts
```diff
@@ -10,93 +9,24 @@ export default function (server: Server, ctx: AppContext) {
...
-        ])
-        .executeTakeFirst()
-
-      if (!queryRes) {
-        throw new InvalidRequestError(`Profile not found`)
+      if (!user) {
+        throw new InvalidRequestError('Profile not found')
       }
-      if (softDeleted(queryRes)) {
+      if (softDeleted(user)) {
```
dholms, line 20: actually looks like the service is handling this so you can probably ditch the extra check

packages/pds/src/services/actor/views.ts
```diff
@@ -0,0 +1,106 @@
...
+        },
+      }
+    })
+
+    return Array.isArray(result) ? views : views[0]
+  }
+}
+
+type UserResult = User & DidHandle & RepoRoot
+type ArrayEl<A> = A extends readonly (infer T)[] ? T : never
```
dholms, line 106: haha cheeky

we may have to put this in `common` or something, seems nifty

dholms (approved):
looks great 👌

packages/pds/src/api/app/bsky/actor/getProfile.ts
devinivy: I think it does this so that it can provide the `AccountTakedown` error rather than a not-found— the `true` flag passed to service is letting it return taken-down accounts.  Good call, though, it may be smart to switch to just a not-found 🤔 I am going to merge this for now preserving the behavior and we can always tweak it.

packages/pds/src/services/actor/views.ts
devinivy: Good call, done!  It was being used in a couple places, so seems like the right move.

PR #534, opened 2023-02-08
Handle SIGINT in postgres script

dholms:
This one kept biting me so I decided to knock it out.

Basically intercept SIGINT, & stop docker before existing.

Output is a bit ugly, bash isn't my forte lol. But it at least stops old volumes from sticking around


devinivy (approved):
Nice!

PR #536, opened 2023-02-09
User table fixes

dholms:
This fixes up a couple of boneheaded mistakes I made back in the day on the user table.

Namely:
- using the `handle` as the primaryKey instead of the `did`
- calling it `user`

I migrate the data over to a new table `user_account` that uses `did` as the primary key

Also took the chance to fix up the flow in account creation a little bit so that we don't have to do the whole "preregister user -> register did -> finalize user"


bnewbold:
- renaming away from `user`, and not having the handle be the pk were on my ask list, so thanks!
- if we could rename `password` to `passwordScrypt` to make it clear that it is hashed it would lower my blood pressure. low-priority.
- https://brandur.org/fragments/postgres-table-rename describes a way to do zero-downtime table renames in postgresql, using a temporary VIEW instead of the copy. it is a two-step process, and i'm not sure if it works with sqlite (probably not?). may not be worth it depending on your roll-out plan; we don't have many users yet
- I think you need both of the prior UNIQ indices on the `email` column: both lower-cased and plan. we could get away with only one if we were rigorous in application code, but let's just belt-and-suspenders for now. there is a similar situation with `did_handle` on `handle`
- at some point I might recommend moving `lastSeenNotifs` into a separate `user_state` table, because that is relatively high-churn, which could in theory cause issues down the road depending on VACUUM and transaction serialization nitty gritty. basically, `user_account` is a frequently JOIN'd table from many requests, and that doesn't pair well with high-churn, I believe. but let's punt on that for now.
- as other style things, i'd generally recommend: using `TEXT` with `CHECK` constraints over `VARCHAR` (in both sqlite and postgres, basically same behavior but altering constraints is better); make `createdAt` `TIMESTAMP WITH TIME ZONE DEFAULT now() NOT NULL` instead of a string (`createdAt` can be left to database on this specific table, normally might only want `indexedAt` to be auto-gen); have minimum length `CHECK` constraints on all string columns to prevent `""` (empty string, as opposed to `NULL`) situations, or things like `"None"` or `"undefined"`. these are all style, and apply to all tables, and are low-priority, so I would lean towards skipping for now

PostgreSQL description of this table previously, for reference:

```
                                Table "public.user"
            Column         |       Type        | Collation | Nullable | Default 
    ------------------------+-------------------+-----------+----------+---------
    handle                 | character varying |           | not null | 
    email                  | character varying |           | not null | 
    password               | character varying |           | not null | 
    lastSeenNotifs         | character varying |           | not null | 
    createdAt              | character varying |           | not null | 
    passwordResetToken     | character varying |           |          | 
    passwordResetGrantedAt | character varying |           |          | 
    Indexes:
        "user_pkey" PRIMARY KEY, btree (handle)                                                          
        "user_email_key" UNIQUE CONSTRAINT, btree (email)
        "user_email_lower_idx" UNIQUE, btree (lower(email::text))
        "user_password_reset_token_idx" UNIQUE, btree ("passwordResetToken")
```

bnewbold:
If you were feeling expansive, could also make some updates to the adjacent `user_notification` table. Current postgres schema:

```
                      Table "public.user_notification"                                                   
        Column     |       Type        | Collation | Nullable | Default 
    ---------------+-------------------+-----------+----------+---------
     userDid       | character varying |           | not null | 
     recordUri     | character varying |           | not null | 
     recordCid     | character varying |           | not null | 
     author        | character varying |           | not null | 
     reason        | character varying |           | not null | 
     reasonSubject | character varying |           |          | 
     indexedAt     | character varying |           | not null | 
    Indexes:
        "user_notification_did_indexed_at_idx" btree ("userDid", "indexedAt")
```

The urgent thing is lack of a PRIMARY KEY (the existing index probably not a good PK on it's own). Maybe `(userDid, recordUri, reasonSubject)`? I'm not confident that `(uderDid, recordUri)` is sufficient on it's own, but maybe it is.

And less urgent, on the same table, would be making `userDid` a FOREIGN KEY to the `user_account` table. Apart from keeping things tidy/legible, that would help enforce cleanup on account deletion. I don't think we ever want/need notifications on a PDS for non-local accounts.

These are scope creep on this current PR.

bnewbold (commented):
Did a quick skim of code, no issues jumped out. See comments on SQL schema.

packages/pds/src/api/com/atproto/account/create.ts
```diff
@@ -68,33 +63,44 @@ export default function (server: Server, ctx: AppContext) {
...
-          throw new InvalidRequestError(`Handle already taken: ${handle}`)
-        }
-        throw err
+      // format create op, but don't send until we ensure the username & email are available
+      const plcCreate = await ctx.plcClient.formatCreateOp(
+        ctx.keypair,
+        recoveryKey || ctx.cfg.recoveryKey,
+        handle,
+        ctx.cfg.publicUrl,
+      )
```
devinivy, line 72: Kicking myself for not seeing this before 🤦 this was the answer all along!

dholms: dude you say that, but I _wrote_ the damn thing & didn't think of it 🙃

packages/pds/src/db/migrations/20230208T222001557Z-user-table-did-pk.ts
```diff
@@ -0,0 +1,101 @@
...
+    .execute()
+
+  await db.schema
+    .createIndex('user_account_password_reset_token_idx')
+    .unique()
+    .on('user_account')
+    .column('passwordResetToken')
+    .execute()
+
+  await db.schema.dropIndex('user_password_reset_token_idx').execute()
```
devinivy, line 49: I think it's possible postgres will ask us to remove the `user_email_lower_idx` as well.  We might also want to recreate it on `user_account` if the idea is a one-to-one port of the `user` table.

packages/pds/src/db/migrations/20230208T222001557Z-user-table-did-pk.ts
```diff
@@ -0,0 +1,101 @@
...
+          'user_account.lastSeenNotifs',
+          'user_account.createdAt',
+          'user_account.passwordResetToken',
+          'user_account.passwordResetGrantedAt',
+        ]),
+    )
+    .execute()
+
+  await db.schema
+    .createIndex('user_password_reset_token_idx')
```
devinivy, line 93: There's at least the `user_email_lower_idx` to recreate too.

devinivy (approved):
Definitely a nice improvement, a lot more natural to work with 👍

packages/pds/src/db/migrations/20230208T222001557Z-user-table-did-pk.ts
bnewbold: I think as things stand we want both the `lower` and regular indices. The lower acts as a UNIQUE constraint (preventing reuse of emails), while the regular is what gets used in practice for fast lookups.

bnewbold (approved):

PR #537, opened 2023-02-09
some small low-priority patches

bnewbold:
A small grab-bag of edits from my working tree.

It took a while to get PLC and PDS running with log lines printing out as expected in my terminal. There were issues with dotenv and logging (due to library import priority), and needing to pipe the log JSON through `pino-pretty`.

Was getting a yellow warning line repeatedly about the missing `"license"` line in PDS.

I personally find it really helpful to have a language-agnostic `Makefile`, because I always forget what tool set is being used for what projects. It is kind of opinionated and old-fashioned.

I'm happy to cherry pick / rebase / whatever if only some of these should get merged to `main`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ test.sqlite
 .DS_Store
 *.log
 tsconfig.build.tsbuildinfo
+.dev.env
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,6 @@ test.sqlite
 *.log
 tsconfig.build.tsbuildinfo
 .dev.env
+\#*\#
+*~
+*.swp
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,51 @@
+
+SHELL = /bin/bash
+.SHELLFLAGS = -o pipefail -c
+
+.PHONY: help
+help: ## Print info about all commands
+	@echo "Helper Commands:"
+	@echo
+	@grep -E '^[a-zA-Z0-9_-]+:.*?## .*$$' $(MAKEFILE_LIST) | awk 'BEGIN {FS = ":.*?## "}; {printf "    \033[01;32m%-20s\033[0m %s\n", $$1, $$2}'
+	@echo
+	@echo "NOTE: dependencies between commands are not automatic. Eg, you must run 'deps' and 'build' first, and after any changes"
+
+.PHONY: build
+build: ## Compile all modules
+	yarn build
+
+.PHONY: test
+test: ## Run all tests
+	yarn test
+
+.PHONY: run-dev-env
+run-dev-env: ## Run a "development environment" shell
+	cd packages/dev-env; yarn run start
+
+.PHONY: run-pds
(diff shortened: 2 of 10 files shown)


packages/plc/src/bin.ts
```diff
@@ -1,15 +1,15 @@
 import dotenv from 'dotenv'
+const env = process.env.ENV
```
dholms, line 2: I honestly thought you weren't able to have any code in between static imports with ESM. Maybe our transpiler is able to handle it?

Regardless, its usually not considered best practice. if we want this code to run first, I'd drop it in its own `env.ts` file or similar

packages/pds/example.dev.env
```diff
@@ -0,0 +1,6 @@
+DB_POSTGRES_URL="postgres://bsky:yksb@localhost/pds_dev"
+DEBUG_MODE=1
+LOG_ENABLED="true"
+LOG_LEVEL=debug
+LOG_DESTINATION=1
```
dholms, line 5: yup sounds good 👍 

I've been outputting to a log file for development so my console is less noisey. but this seems like a reasonable default

Makefile
```diff
@@ -0,0 +1,49 @@
+
```
dholms, line 1: I'm down with the Makefile

Esp when it comes to more complicated scripts, writing them in package.json starts to become a pain

.gitignore
```diff
@@ -9,4 +9,8 @@ packages/*/coverage
 test.sqlite
 .DS_Store
 *.log
-tsconfig.build.tsbuildinfo
\ No newline at end of file
+tsconfig.build.tsbuildinfo
+.dev.env
+#*#
```
dholms, line 14: wasn't sure what this line was for so i did a quick google search & it turned up that it might need to be escaped?
https://stackoverflow.com/questions/2209810/add-glob-to-gitignore

dholms (approved):
yup im down with all of this 👍 

The one thing to be sure to hit is my comment on `plc/bin.ts`

always feel free to propose quality of life stuff such as this. not having done it is often more "million other things to do" & less "no we don't want it"

packages/plc/src/bin.ts
bnewbold: :+1: added as a separate `src/env.ts` file for both PLC and PDS

packages/pds/example.dev.env
bnewbold: Yeah, does need to be explicitly enabled via the `.env` file (or environment variables separately), but this will be the default for folks just getting started (running the Makefile)

.gitignore
bnewbold: yes! thanks. my template file is ancient, thought maybe this never worked correctly.

bnewbold:
Made some fixes and rebased.

If the `src/env.ts` changes look good, i'll merge this to `main`.

dholms (approved):
yup yup looks good 👍

PR #538, opened 2023-02-09
pds sql: add indices for feed and post aggregations

bnewbold:
NOTE: the name of the migration might need to change, as there are other PRs in flight.

I wrote this pretty quickly and have no prior experience with `kysely`, should be reviewed carefully before pushing this out.

These are some high-priority indices that yielded around a 5x improvement on single-threaded query latency in my local (laptop) testing on a small database for important XRPC endpoints like "getTimeline", "getPostThread", "getProfile", and fetching notifications (list or count).

These endpoints still seem pretty slow to me (tens of millis SQL execution time), but the speedup is quite cheap (there aren't many indices on these tables yet, and index creation should be fairly fast). We may refactor the aggregation stuff in the near future, but until then I think these are helpful.

### DIFF
--- a/packages/pds/src/db/migrations/20230208T081544325Z-post-hydrate-indices.ts
+++ b/packages/pds/src/db/migrations/20230208T081544325Z-post-hydrate-indices.ts
@@ -1,0 +1,54 @@
+import { Kysely } from 'kysely'
+
+export async function up(db: Kysely<unknown>): Promise<void> {
+  // for, eg, "upvoteCount" on posts in feed views
+  await db.schema
+    .createIndex('vote_subject_direction_idx')
+    .on('vote')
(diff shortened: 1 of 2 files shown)


devinivy (approved):
Lovely!  Thanks for knocking that out 🙏

dholms (approved):
yup yup looks great 👌

bnewbold:
I reviewed and this branch is ready to merge from my perspective. @devinivy @dholms 

Note that the migration timestamp in this branch is earlier (but same date) compared the migration in https://github.com/bluesky-social/atproto/pull/536 (user table fixes).

bnewbold:
Spoke too soon! Resolving latest `main` branch conflict real quick...

Issue #509, opened 2023-01-31
Add ways to show if someone is following the user


pfrazee:
Currently there's no efficient way to find out if a user follows you.

Here's what I'd suggest adding:

- Update `actor.getProfile`, `graph.getFollows`, and `graph.getFollowers` to include a boolean indicating if the actor follows you.
- Probably do the same with `actor.search` and `actor.getSuggestions`
- Not necessary but might be worth updating `graph.getFollows` to optionally take a target actor to quickly determine if someone follows a specific actor. Alternatively can add a separate method.

dholms:
fixed https://github.com/bluesky-social/atproto/pull/539

PR #539, opened 2023-02-09
Add following state to profile/actor views, reorg views

devinivy:
The primary purpose of this PR is to address #509 and provide a way to check if a given user follows you on a handful of XRPC methods.  While combing through the XRPC lexicons for profiles and actors I found six variants, so I've also taken a swing at culling those down to three:
1. Simple lists of actors: `did`, `handle`, `declaration`, `displayName`, `avatar`, `viewer`.
2. More interesting lists include on top of no. 1: `indexedAt`, `description`.
3. Full profile details include on top of no. 2: `banner`, `postsCount`, `followersCount`, `followingCount`, `creator`.

The `viewer` state now includes: `{ muted: boolean, followedBy: follow-uri, following: follow-uri }`, and the `myState` property which looks like `{ muted: boolean, follow: follow-uri }` remains on the profile views but is deprecated.

The only breaking changes here are that `indexedAt` and/or `createdAt` have been removed from `getMutes`, `getFollows`, `getFollowers`, `getVotes`, and `getRepostedBy`.


pfrazee:
👍 From the frontend perspective (API changes) didnt review impl

lexicons/app/bsky/actor/getProfile.json
```diff
@@ -13,37 +13,7 @@
...
-          }
-        }
-      }
-    },
-    "myState": {
-      "type": "object",
-      "properties": {
-        "follow": {"type": "string"},
-        "muted": {"type": "boolean"}
+        "schema": {"type": "ref", "ref": "app.bsky.actor.profile#view"}
```
dholms, line 16: ah good catch 👍

packages/pds/src/api/app/bsky/actor/getSuggestions.ts
```diff
@@ -64,23 +54,13 @@ export default function (server: Server, ctx: AppContext) {
...
-          : undefined,
-        indexedAt: result.indexedAt ?? undefined,
-      }))
-
       return {
         encoding: 'application/json',
         body: {
-          actors,
           cursor: keyset.packFromResult(suggestionsRes),
+          actors: await services
```
dholms, line 61: aw yiss this cleanup feels really good

packages/pds/src/services/actor/views.ts
```diff
@@ -0,0 +1,222 @@
...
+          following: profileInfo?.requesterFollowing || undefined,
+          followedBy: profileInfo?.requesterFollowedBy || undefined,
+        },
+      }
+    })
+
+    return Array.isArray(result) ? views : views[0]
+  }
+
+  // @NOTE keep in sync with feedService.getActorViews()
```
dholms, line 197: probably worth just having one view for these (in actor service)

but don't need to fix in this PR

devinivy: 💯  really wanted to do that but it got slightly dicey and I ended-up bailing on it.

packages/pds/tests/views/follows.test.ts
```diff
@@ -34,64 +33,41 @@ describe('pds follow views', () => {
...
-  const tstamp = (x: string) => new Date(x).getTime()
-
   it('fetches followers', async () => {
     const aliceFollowers = await agent.api.app.bsky.graph.getFollowers(
       { user: sc.dids.alice },
       { headers: sc.getHeaders(alice) },
     )
 
     expect(forSnapshot(aliceFollowers.data)).toMatchSnapshot()
-    expect(getCursors(aliceFollowers.data.followers)).toEqual(
```
dholms, line 52: why'd we ditch this? is it redundant? def appreciate the simplicity tbf

dholms (approved):
this cleaned up real nice ✨ 

thanks for taking it the extra step w the lex + services refactor

packages/pds/tests/views/follows.test.ts
devinivy: Technically it's redundant with the snapshot, but it called out a property of the results a bit more explicitly.  I nixed it because the data to build your own cursor for any of the elements doesn't come down from the API anymore (I think in this case it's based on the `indexedAt` time of the follow record).

packages/pds/src/services/actor/views.ts
dholms: yup yup fair enough 👍

packages/pds/tests/views/follows.test.ts
dholms: cool thats what i figured 👍

pfrazee:
> The only breaking changes here are that indexedAt and/or createdAt have been removed from getMutes, getFollows, getFollowers, getVotes, and getRepostedBy.

It looks like `getVotes` didnt get changed akshually

PR #544, opened 2023-02-10
Mod action and report params as integers rather than floats

devinivy:
The `id` parameter for getting a mod action and reports was spec'ed as a `number` when it should have been an `integer`.


dholms (approved):
yup yup 👍

PR #545, opened 2023-02-13
Notify all ancestors of reply, optimize getPostThread method

devinivy:
Here we start tracking all ancestors of each post.  This allows us to efficiently build post threads and allows us to notify all ancestors in a post thread when there is a reply.

This is done with a new index called post_hierarchy, which is a closure table `(uri, ancestorUri, depth)`.  A closure table is comparable to storing an ordered list of ancestors on each post.  But by breaking this out into a table it allows joining over that data and efficiently looking-up children in addition to ancestors.


pfrazee:
Discussed this approach to the problem with Divy, works for me

packages/pds/src/db/migrations/20230210T210132396Z-post-hierarchy.ts
```diff
@@ -0,0 +1,65 @@
...
+            .innerJoin('hierarchy', 'hierarchy.ancestorUri', 'post.uri')
+            .where('post.replyParent', 'is not', null)
+            .select([
+              'hierarchy.uri as uri',
+              sql<string>`post."replyParent"`.as('ancestorUri'),
+              sql<number>`hierarchy.depth + 1`.as('depth'),
+            ]),
+        )
+    })
+    .selectFrom('hierarchy')
```
dholms, line 42: helluva qb 👏

dholms (approved):
this looks really solid. great work on this 👏

PR #546, opened 2023-02-13
Use DAG-CBOR for XRPC streaming

devinivy:
Using DAG-CBOR for XRPC streaming, rather than a non-canonical CBOR representation.


dholms (approved):
looks good 👍 

gotta figure out these flaky tests 😅 
think it's just the timeouts for stream listening but not sure about alternatives there

Issue #480, opened 2023-01-24
Add APIs for handle changes


pfrazee:
Users need the ability to change their handle.

- Add an API for changing the user handle
  - Should be able to allocate to any open handle that the PDS offers. (In the future, this will include custom domains)
  - Must update the DID document in addition to internal records
- Add tests ensuring handle changes work correctly
  - All account behaviors work correctly (signin, etc)
  - Views are updated to provide the new handle
  - Data integrity is not lost (unlikely but gotta check we've been rigorous)

Issue #481, opened 2023-01-24
Add DNS-record solution to "handle resolution"


pfrazee:
## Background

Currently handles are resolved via the `com.atproto.handle.resolve()` XRPC method.

The behavior is as follows:

- If the `handle` param is undefined or is set to the service hostname, the resolution is canonical.
- If the `handle` param is set to a hostname other than the service, the resolution is not canonical.

Whether the XPRC call is canonical has to do with the service's authority. A given handle (e.g. `alice.com`) can only canonically resolve itself. Therefore

```
GET https://alice.com/xrpc/com.atproto.handle.resolve?handle=alice.com
```

is canonical but 

```
GET https://alice.com/xrpc/com.atproto.handle.resolve?handle=bob.com
```

is not.

## Problem

Relying on an XRPC method mandates that handles can only be set by HTTPS services with XRPC support. That overhead discourages users from using their domains as handles, especially when they're not self-hosting. Since using custom domain names as handles is one of the more fun features of ATP, we should make this easier!

## Solution

Add a second DNS-record method for setting a domain name handle. This record will be set as follows:

|Domain|Record Type|Record Value|
|-|-|-|
|`_atproto.{DOMAIN}`|`TXT`|`{DID}`|

For example, a valid DNS-record for `alice.com` might be:

```
_atproto.alice.com TXT did:plc:ragtjsm2j2vknwkz3zp4oxrd
```

We use the `_atproto` prefix to avoid potential issues around CNAME record behaviors. We set the value to the DID with no additional syntax because they are relatively unambiguous (they always start with `did:`) and because the `_atproto` prefix is unlikely to conflict with any existing record types.

Resolution of a handle then involves a race between requests to the DNS record and the XRPC method. If both respond with conflicting values then the DID document's asserted handle is used to establish the correct value. Implementers are encouraged to use DoH to avoid MITM attacks.

PR #547, opened 2023-02-13
Update handles

dholms:
Adds a method for updating handles. Pretty straightforward.

Only data integrity issue I can see is:
- the way we do mentions still encodes their old handle _as text_ in the post. So the link itself will still work (because it links to the DID), but the handle text will not change. I think this is probably the expected behavior

One thing to think through here that is not included in this PR is: how do username changes propagate to downstream systems?

Closes https://github.com/bluesky-social/atproto/issues/480
Prepares us for https://github.com/bluesky-social/atproto/issues/481


packages/pds/src/api/com/atproto/handle.ts
```diff
@@ -0,0 +1,64 @@
+import { InvalidRequestError } from '@atproto/xrpc-server'
+import * as handleLib from '@atproto/handle'
+import { Server } from '../../../lexicon'
+import AppContext from '../../../context'
+import { UserAlreadyExistsError } from '../../../services/actor'
+
+export default function (server: Server, ctx: AppContext) {
+  server.com.atproto.handle.resolve(async ({ params }) => {
```
dholms, line 8: nothing new here, just changed the file name to match the schema

packages/pds/tests/handles.test.ts
```diff
@@ -0,0 +1,156 @@
...
+    )
+
+    const aliceFollows = followers.data.followers.filter((f) => f.did === alice)
+    expect(aliceFollows.length).toBe(1)
+    expect(aliceFollows[0].handle).toBe(newHandle)
+  })
+
+  it('does not allow taking a handle that already exists', async () => {
+    const attempt = agent.api.com.atproto.handle.update(
+      { handle: 'bob.test' },
```
devinivy, line 101: Might suggest exercising the handle normalization here by futzing with casing:
```suggestion
      { handle: 'Bob.Test' },
```

packages/pds/src/services/actor/index.ts
```diff
@@ -166,6 +166,23 @@ export class ActorService {
...
+    const res = await this.db.db
+      .updateTable('did_handle')
+      .set({ handle })
+      .where('did', '=', did)
+      .whereNotExists(
+        this.db.db
+          .selectFrom('did_handle')
+          .where('handle', '=', handle)
+          .selectAll(),
+      )
```
devinivy, line 179: We should be able to lean directly on the `handle_lower` index to ensure we don't get a duplicate, which I think would actually feel pretty nice, and prove that we don't strictly rely on trusted `handle` inputs to this method.  That would involve switching this to an `onConflict((oc) => oc.doNothing())`.

devinivy (approved):
Good looks!

packages/pds/tests/handles.test.ts
dholms: yup im game 👌

packages/pds/src/services/actor/index.ts
dholms: I would prefer that too & it's what I originally tried, but there's no `onConflict` for `update` :/

Any other ideas?

devinivy: Ah dang, good call!  My only other idea is to handle the constraint error that comes out of the database, but that's a little funky to do across sqlite and pg.  I think as-is works well particularly with a test to back it up 👍

dholms: cool cool 👍 

got that test in & it works fine

devinivy (approved):

PR #550, opened 2023-02-14
Small PDS and XRPC server fixes

bnewbold:
The PDS logging thing commit is simple. Was always displaying in logs as `Bearer Invalid` even for valid auth requests.

The XRPC server fix is mostly to improve error messages in responses. I ran across this while debugging blob uploads from golang code and was getting error messages like `Invalid request encoding: false` which was confusing.

In the second commit, in test coverage, I expect the error message indicated in the `TODO` line. I spent a while trying to get the test to pass that way but don't feel like I know `jest` or javascript internals well enough to figure out why it is `TypeError: fetch failed` instead and gave up.

### DIFF
--- a/packages/xrpc-server/tests/bodies.test.ts
+++ b/packages/xrpc-server/tests/bodies.test.ts
@@ -151,6 +151,14 @@ describe('Bodies', () => {
     await expect(
       client.call('io.example.validationTest', {}, { foo: 123 }),
     ).rejects.toThrow(`Input/foo must be a string`)
+    await expect(
+      client.call(
+        'io.example.validationTest',
+        {},
+        { foo: 'hello', bar: 123 },
+        { encoding: 'image/jpeg' },
+      ),
+    ).rejects.toThrow(`Wrong request encoding (Content-Type): image/jpeg`)
 
     // 500 responses don't include details, so we nab details from the logger.
     let error: string | undefined
@@ -236,5 +244,19 @@ describe('Bodies', () => {
     )
 
     await expect(promise).rejects.toThrow('request entity too large')
+  })
+
+  it('requires any parsable Content-Type for blob uploads', async () => {
+    // not a real mimetype, but correct syntax
+    await client.call('io.example.blobTest', {}, randomBytes(BLOB_LIMIT), {
+      encoding: 'some/thing',
+    })
+
+    // empty mimetype, but correct syntax
+    const p3 = client.call('io.example.blobTest', {}, randomBytes(BLOB_LIMIT), {
+      encoding: '',
+    })
+    // TODO: await expect(p3).rejects.toThrow('Request encoding (Content-Type) required but not provided')
+    await expect(p3).rejects.toThrow('TypeError: fetch failed')
   })
 })
--- a/packages/pds/src/logger.ts
+++ b/packages/pds/src/logger.ts
@@ -17,7 +17,7 @@ export const loggerMiddleware = pinoHttp({
       const authHeader = serialized.headers.authorization || ''
       let auth: string | undefined = undefined
       if (authHeader.startsWith('Bearer ')) {
-        const token = authHeader.slice(0, 'Bearer '.length)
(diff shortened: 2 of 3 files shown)


packages/xrpc-server/tests/bodies.test.ts
```diff
@@ -237,4 +245,18 @@ describe('Bodies', () => {
...
+    // not a real mimetype, but correct syntax
+    await client.call('io.example.blobTest', {}, randomBytes(BLOB_LIMIT), {
+      encoding: 'some/thing',
+    })
+
+    // empty mimetype, but correct syntax
+    const p3 = client.call('io.example.blobTest', {}, randomBytes(BLOB_LIMIT), {
+      encoding: '',
+    })
+    // TODO: await expect(p3).rejects.toThrow('Request encoding (Content-Type) required but not provided')
```
dholms, line 259: i know what the issue is here btw, i'll push a fix

dholms: alright well the issue is that xrpc/client is defaulting that value to 'application/json' which I think is good in the general case (I can't think of a situation where we should like clients send an empty content type & nearly all of our methods are json 🤔)

So i tried deconstructing it to basic fetch semantics, but it's only working for me intermittently. Oddly it works fine when the test before it is commented out.

```ts
    // empty mimetype, but correct syntax
    const res = await fetch(`${url}/xrpc/io.example.blobTest`, {
      method: 'post',
      headers: { 'Content-Type': '' },
      body: randomBytes(BLOB_LIMIT),
      // @ts-ignore see note in @atproto/xrpc/client.ts
      duplex: 'half',
    })
    const resBody = await res.json()
    const status = res.status
    expect(status).toBe(400)
    expect(resBody.error).toBe('InvalidRequest')
    expect(resBody.message).toBe(
      'Request encoding (Content-Type) required but not provided',
    )
```

dholms: @devinivy any ideas before I spend more time spinning on this?

devinivy: Nothing is coming to mind, sorry!  Happy to pair on it if you want another set of eyes, though.

bnewbold: JSON default makes sense!

I was also running in to some weird "what the test before is doing" behavior and that was when I gave up. Don't think we should collectively burn too many cycles on this right now, we can just mark the test with a skip or something.

dholms: yeah I'd say copy this into a new test & skip it for now with a comment about why it's not working

dholms:
pushed up my pass on fixing the test, but `.skip`ed it & left a TODO + note

dholms (approved):
👍

bnewbold:
thanks for pushing this through!!

packages/xrpc-server/tests/bodies.test.ts
```diff
@@ -237,4 +246,31 @@ describe('Bodies', () => {
...
+  it('requires any parsable Content-Type for blob uploads', async () => {
+    // not a real mimetype, but correct syntax
+    await client.call('io.example.blobTest', {}, randomBytes(BLOB_LIMIT), {
+      encoding: 'some/thing',
+    })
+  })
+
+  // @TODO: figure out why this is failing dependent on the prev test being run
+  // https://github.com/bluesky-social/atproto/pull/550/files#r1106400413
+  it.skip('errors on an empty Content-type on blob upload', async () => {
```
matthieusieben, line 259: Tt fails because of the following line that break express' control flow.

https://github.com/bluesky-social/atproto/blob/169486a3fef64da086d1499ba9ba743fdfef0af2/packages/xrpc-server/src/server.ts#L194-L198

Indeed, when an error is triggered in the body (e.g. because of the max-size), the error is now propagated in two places: `next` (for the `error` handler) as well as the place where the readable `body` was being consumed (which may, or may not, write to the response, call next, etc.)

Issue #526, opened 2023-02-05
Defaults & limits for paginated queries


dholms:
Known issue with 'getTimeline' but we should check others as well

dholms:
fixed by https://github.com/bluesky-social/atproto/pull/552

PR #552, opened 2023-02-14
Apply lexicon defaults

### LINKED ISSUE #299: Ensure lexicon defaults in xrpc server
When a default value is set in an XRPC schema, we should ensure that value is actually being defaulted to in the server. Would be good to handle this at the XRPC layer.

Inspiration: https://github.com/bluesky-social/atproto/pull/297#discussion_r1012314486

devinivy:
The `default` values for number, integer, string, and boolean types are now applied during validation in xrpc-server.  Codegen has been updated to reflect that properties with a default are present.

Fixes #299 #526


dholms (approved):
awesome 🙌 🙇

PR #553, opened 2023-02-14
Post embeds

devinivy:
Enables a post to be embedded in another post, similarly to how images or external links may be embedded.  In the presentation lexicons there's a carve-out for a post that is not found, e.g. if it were taken-down or deleted.


pfrazee:
👍 Looks great

lexicons/app/bsky/embed/post.json
```diff
@@ -0,0 +1,38 @@
+{
+  "lexicon": 1,
+  "id": "app.bsky.embed.post",
```
dholms, line 3: I almost wonder if we should make this `embed.record` since that's what it's actually "describing" & may be more mappable to other data types 🤔 

For instance embedding something like a "profile".

Or (contrived example I use often), but imagine a Wordle app deployed on atproto & wanted to embed your score for the day in a native way?

The possible downside would be overloading this embed type & maybe each of those embeds should be treated differently. But at the same time, I don't see anything in the schema that makes this specific to posts

dholms: Also should we have a list of "relevant blobs" in `presentedPost`?

lexicons/app/bsky/embed/post.json
```diff
@@ -0,0 +1,38 @@
...
+    "presented": {
+      "type": "object",
+      "required":  ["post"],
+      "properties": {
+        "post": {"type": "union", "refs": ["#presentedPost", "#presentedNotFound"]}
+      }
+    },
+    "presentedPost": {
+      "type": "object",
+      "required":  ["uri", "cid", "author", "record"],
```
dholms, line 22: Should we have a list (or map) of "relevant blobs" here?

CID -> url map basically (and you can find the CID for the blob from the record)

lexicons/app/bsky/embed/post.json
devinivy: Very much hear that, good thought. @pfrazee and I worked through some of the tradeoffs here in person.  It's really tempting to do what you're saying, particularly because what we're sending down for a post (uri, cid, author, record) _would_ work equally well for other record types.  The rub is that to do this by the book it opens-up questions about how to model fallback content for clients that don't understand the embed (since simply omitting the unfamiliar embed could violate user intent), and we weren't quite ready to bite that off.  I'm still open to it but that was our thinking a couple days ago.

packages/pds/src/db/tables/post-embed.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export interface PostEmbedExternal {
+  postUri: string
+  uri: string
+  title: string
+  description: string
+  thumbCid: string | null
+}
+
+export interface PostEmbedPost {
```
dholms, line 20: hey dawg i heard you like posts?

dholms (approved):
code looks great 🙌 

just a question of the generality of the embed type 🤔

lexicons/app/bsky/embed/post.json
dholms: Hmmm I totally see that...

So basically the implication here is that "only embeds of type `app.bsky.feed.post` are allowed", right?

We need to tangle with this fallback question in general soon imo...

Totally understand the concern, but I think I'm in favor of `embed.record` with fallback to "could not display embedded record" element or something

dholms: But also if you guys already grappled with it, I'm down to trust your judgement & just get this rolling. We can always add in the generic form later

lexicons/app/bsky/embed/post.json
devinivy: 🤔 hmmm not a bad idea!  Paul said he didn't need embeds here yet, but I wouldn't mind adding them if they'll be put to good use.

Approach-wise I think it depends a little bit on whether we go the generic record route or keep it post-specific.  If we keep it post-specific, I would probably tend to include the embeds the same way we do in post views (this object already makes-up a partial post view, so it would involve adding more fields to it).  If we make it generic, I like that proposal a lot.  Actually that is already giving me ideas for the `com.atproto.admin.*` methods in which records appear...

lexicons/app/bsky/embed/post.json
dholms: There's also a decent argument that there's a difference between:
"embedding this post" (like a quote tweet)
"embedding this record" (like a link w preview)
🤷‍♂️

lexicons/app/bsky/embed/post.json
dholms: yup that makes sense 👍

lexicons/app/bsky/embed/post.json
pfrazee: We were trying to avoid having to do schema negotiation tooling now, but it's occurring to me that targeting a specific schema makes it hard for us to v2 the post. Maybe we go generic but don't worry about being super flexible or handling fallbacks well yet?

lexicons/app/bsky/embed/post.json
pfrazee: Yeah I'm 99% sure we don't want embeds within embeds. If we change our mind, I think it'll be easy to add as an optional field.

lexicons/app/bsky/embed/post.json
dholms: yup I think that's reasonable 👍 

it's in line with the rest of lex: likes can be on _anything_. but if the like doesn't make sense, it's ignored

altho nothing in this prevents a generic record, it's just the implication of "embedding a post" v "embedding a record" & if that's a useful distinction.

In my mind that distinction - if it exists - is: quote tweet v link to tweet w preview

devinivy:
@dholms should be good for another peek 👍

dholms (approved):
hot dog 🌭 ✨

PR #554, opened 2023-02-15
Allow unsetting profile images and description in updateProfile

devinivy:
A few steps here, in order to allow unsetting profile images and descriptions in the `app.bsky.actor.updateProfile` method.
 - Add support for `nullable` on lexicon object types, spec'ed similarly to `required`.
 - Add `nullable` to codegen.
 - Alter lexicons and implementation of `updateProfile` to allow unsetting avatar, banner, and description.

### DIFF
--- a/packages/lex-cli/src/codegen/lex-gen.ts
+++ b/packages/lex-cli/src/codegen/lex-gen.ts
@@ -97,10 +97,12 @@ export function genObject(
     isExported: true,
   })
   genComment(iface, def)
+  const nullable = new Set(def.nullable)
   if (def.properties) {
     for (const propKey in def.properties) {
       const req = def.required?.includes(propKey)
       const propDef = def.properties[propKey]
+      const propNullable = nullable.has(propKey)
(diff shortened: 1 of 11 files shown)


pfrazee:
Nice. I think this is the right call and the PR lgtm

packages/lex-cli/src/codegen/lex-gen.ts
```diff
@@ -127,11 +129,17 @@ export function genObject(
...
-              )}[]`,
+              type: makeType(
+                refToType(
+                  propDef.items.ref,
+                  stripScheme(stripHash(lexUri)),
+                  imports,
+                ),
+                {
+                  nullable: propNullable,
+                  array: true,
```
dholms, line 140: just out of curiosity, what does this `array: true` do?

devinivy: An example of that would be the difference between `A | B` and `(A | B)[]`.

dholms (approved):
awesome 👍 

feels very natural in lex

Issue #482, opened 2023-01-24
Add support for custom domain handles


pfrazee:
Once https://github.com/bluesky-social/atproto/issues/480 and https://github.com/bluesky-social/atproto/issues/481 are complete, we should update the PDS to support custom domain handles.

The application flow will proceed as follows:

- The user starts the custom domain flow
- The user enters their intended domain
- The app gives instructions for configuring their DNS record
- The user presses a "confirm" button which triggers a `com.atproto.handle.resolve` request to the PDS with `handle={domain}`
  - The PDS runs the handle-resolution process and responds with the DID or with a failure
- When the correct DID is returned by `resolve()`, the app sends an API request to the PDS to update the user's handle to the new domain

The PDS needs to be updated in the following ways:

- A. It must correctly handle custom domains in existing behaviors. (Basically run tests to ensure no assumptions are made.)
- B. It must support updating an account to a custom domain.
- C. It must support creating a new account with a custom domain.

A & B are fairly trivial, and B should use the same API as is used in https://github.com/bluesky-social/atproto/issues/480.

C is more complex because the user must have a DID before they can set their DNS record. Therefore we'll need to update the signup APIs to support this flow, perhaps by optimistically creating the DID document with the user's intended domain name.

As C will require some thought, I would suggest we break this task into two phases:

- [ ] Add the ability to change an existing account's handle to a custom domain (A & B)
- [ ] Add the ability to set a custom domain during signup (C)

Until C is complete, users will just create their account with an assigned subdomain then switch to their custom domain in the settings after.

dholms:
done
https://github.com/bluesky-social/atproto/pull/555

PR #555, opened 2023-02-15
DNS handles

dholms:
Adds support for DNS handles.

_Note: does not yet allow for creating an account with a DNS handle, but shouldn't be too heavy of a lift_

- DNS resolution of handles
- DNS + XRPC checks for an external handle in `handle.update`

For DNS records this does the original proposal from https://github.com/bluesky-social/atproto/issues/481:
`_atproto.dholms.dev TXT "did:plc:abc123..."`

Reason being that it's cleaner, dead simple, has no chance of collision with other namespaces, the closest thing to a "standard" way of doing things like this is DNSLink, but I see very little adoption of that outside of ipfs links in gateways.

Closes https://github.com/bluesky-social/atproto/issues/481
In support of https://github.com/bluesky-social/atproto/issues/482

---

###  🚲 🏠 
Because had to parse out the logic of "valid handles" vs "valid handles on our service".
Some decisions I made:
- limit total handle length to 40 (this is down from 255, but arguably still do big 🤔)
- limit each domain label to 20 (`none.of.these.can.be.longer.than.twenty`) 
- do not allow 2 char subdomains on our service, but _do_ allow them from external handles
- only enforce reserved handles for domains on our service


packages/pds/src/api/com/atproto/handle.ts
```diff
@@ -60,9 +78,34 @@ export default function (server: Server, ctx: AppContext) {
...
           throw err
         }
-
         await ctx.plcClient.updateHandle(requester, handle, ctx.keypair)
       })
     },
   })
 }
+
+const resolveExternalHandle = async (
```
dholms, line 87: I'd rather race these, but I was having a hard time doing it that didn't leave the second promise dangling & we don't have access to piping an abortcontroller through the ApiAgent

packages/pds/src/api/com/atproto/handle.ts
```diff
@@ -13,17 +16,25 @@ export default function (server: Server, ctx: AppContext) {
...
+        const supportedHandle = ctx.cfg.availableUserDomains.some((host) =>
+          handle.endsWith(host),
+        )
+        // this should be in our DB & we couldn't find it, so fail
+        if (supportedHandle) {
+          throw new InvalidRequestError('Unable to resolve handle')
+        }
+
+        // this is not someone on our server, but we help with resolving anyway
+        const resolved = await resolveExternalHandle(ctx.cfg.scheme, handle)
```
dholms, line 32: We talked about doing this: but just want to double check that we are planning to do resolves for users not registered on our server 🤔 

It's relatively inexpensive so I don't see a big downside & shouldnt ever be more than 1 extra hop

dholms: @pfrazee

packages/pds/src/api/com/atproto/handle.ts
devinivy: Curious about this— what do you mean by "dangling" here?  Still, DNS should be quick, so I agree this is probably okay :)

devinivy: P.S. really would be nice to support aborts in the client lib!

dholms: basically: we don't wait for the promise to resolve, so the test runner gets mad at me because we have some async code still running after test ends

only way I can think to get out of that is with aborts

agreed on supporting them 👍

packages/pds/src/api/com/atproto/handle.ts
pfrazee: Yeah I figure do this for now

pfrazee:
> - limit total handle length to 40 (this is down from 255, but arguably still do big 🤔)
> - limit each domain label to 20 (none.of.these.can.be.longer.than.twenty)

So on these -- what's the reason for that? The original logic was to follow the rules of DNS so that any valid domain name (that doesnt cause problems generally) will work with ATP.

packages/handle/src/index.ts
```diff
@@ -54,21 +35,54 @@ export const normalize = (handle: string): string => {
...
   handle: string,
   availableUserDomains: string[],
+  reserved = reservedSubdomains,
 ): boolean => {
   try {
-    ensureValid(handle, availableUserDomains)
+    ensureServiceConstraints(handle, availableUserDomains, reserved)
   } catch (err) {
     if (
       err instanceof InvalidHandleError ||
```
devinivy, line 88: I wonder if we need to add a case here for `UnsupportedDomainError`.
```suggestion
      err instanceof InvalidHandleError ||
      err instanceof UnsupportedDomainError ||
```

pfrazee:
Okay I don't feel strongly enough on this to hang it up so go for it

packages/handle/tests/handle.test.ts
```diff
@@ -1,78 +1,98 @@
...
+    ensureServiceConstraints('john-john.test', domains)
+    ensureServiceConstraints('john.bsky.app', domains)
+    ensureServiceConstraints('0john.test', domains)
+    ensureServiceConstraints('12345.test', domains)
   })
 
   it('allows punycode handles', () => {
-    check('xn--ls8h.test') // 💩.test
-    check('xn--bcher-kva.tld') // bücher.tld
+    ensureValid('xn--ls8h.test') // 💩.test
```
devinivy, line 34: ✨

devinivy (approved):
Ace!!  Realize there may be a few minor changes here, but this all looks solid.

packages/handle/src/index.ts
dholms: yup good find 👍

devinivy (approved):

PR #556, opened 2023-02-16
MST review and tests

bnewbold:
In particular, ports the older fanout=32 corner-case tests to fanout=16. I generated new paths with correct level and ordering, but haven't actually verified that the resulting structures have the shape of the ASCII diagrams in text.

Added a couple tests demonstrating the weirdness of "wide" (non-ASCII) characters in MST key paths. The "right thing to do" is probably to change the MST keys from "string" to "bytes", but that will change an IPLD field type, which changes encoding/serialization/deserialization, so would be tricky to roll out.

Many of these tests are being used as early interoperability tests cross-language. AKA, have implemented the exact same tests in golang (indigo) and rust (adenosine), and are working through inconsistencies.

(still work in progress as I confirm that inconsistencies are bugs in other implementations not TS; also want to push through repo-level semantics in this PR)

### DIFF
--- a/packages/repo/src/mst/mst.ts
+++ b/packages/repo/src/mst/mst.ts
@@ -414,7 +414,6 @@ export class MST implements DataStore {
   // returns entry at index
   async atIndex(index: number): Promise<NodeEntry | null> {
     const entries = await this.getEntries()
-    // TODO: shouldn't this error instead of returning null?
     return entries[index] ?? null
   }
 
--- a/packages/repo/src/mst/mst.ts
+++ b/packages/repo/src/mst/mst.ts
@@ -21,6 +21,15 @@ import * as parse from '../parse'
  * This is a merkle tree, so each subtree is referred to by it's hash (CID).
  * When a leaf is changed, ever tree on the path to that leaf is changed as well,
  * thereby updating the root hash.
+ *
+ * For atproto, we use SHA-256 as the key hashing algorithm, and ~16 fanout
+ * (4-bits of zero per layer).
+ *
+ * NOTE: currently keys are strings, not bytes. Because UTF-8 strings can't be
+ * safely split at arbitrary byte boundaries (the results are not necessarily
+ * valid UTF-8 strings), this means that "wide" characters not really supported
+ * in keys, particularly across programming language implementations. We
+ * recommend sticking with simple alphanumeric (ASCII) strings.
  */
 
 /**
@@ -96,6 +105,7 @@ export class MST implements DataStore {
     opts?: Partial<MstOpts>,
   ): Promise<MST> {
     const pointer = await util.cidForEntries(entries)
+    // TODO: why isn't this layer=null
     const { layer = 0, fanout = DEFAULT_MST_FANOUT } = opts || {}
     return new MST(storage, fanout, pointer, entries, layer)
   }
@@ -111,6 +121,7 @@ export class MST implements DataStore {
     return new MST(storage, fanout, pointer, entries, layer)
   }
 
+  // this is really a *lazy* load, doesn't actually touch storage
   static load(
     storage: ReadableBlockstore,
     cid: CID,
@@ -404,6 +415,7 @@ export class MST implements DataStore {
   // returns entry at index
   async atIndex(index: number): Promise<NodeEntry | null> {
     const entries = await this.getEntries()
+    // TODO: shouldn't this error instead of returning null?
     return entries[index] ?? null
   }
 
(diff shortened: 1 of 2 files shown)


bnewbold:
This is ready to review and merge. We are currently planning to rework how MST keys are stored, but I think these are still valuable tests for the current implementation across languages. And they result in test coverage and some behavioral coverage which will increase confidence in MST key changes, even if the CID values need to be updated.

In particular during review, there are two new TODO comments on existing code in this commit: https://github.com/bluesky-social/atproto/pull/556/commits/f3c634155922bfaf07eba88d155c8613d9b9ba31

bnewbold:
Just pushed a rebase on top of `main`

devinivy (approved):
Aside from the freshly tagged TODOs which I expect we may be able to resolve with @dholms's help, this looks great to me!

packages/repo/src/mst/mst.ts
```diff
@@ -96,6 +105,7 @@ export class MST implements DataStore {
     opts?: Partial<MstOpts>,
   ): Promise<MST> {
     const pointer = await util.cidForEntries(entries)
+    // TODO: why isn't this layer=null
```
dholms, line 108: Hmmm not sure that it really matters & maybe it's just preference. 

`null` is mainly about "not lazy loaded yet" whereas I think of a new MST as being default `0` 

I'm down to change to `null` & perhaps it is actually more honest. Doesn't affect any of the logic 👍

dholms: also a total nit but actually kinda important cause I search on it all the time, we normally do `@TODO` instead of `TODO`

packages/repo/src/mst/mst.ts
```diff
@@ -404,6 +415,7 @@ export class MST implements DataStore {
   // returns entry at index
   async atIndex(index: number): Promise<NodeEntry | null> {
     const entries = await this.getEntries()
+    // TODO: shouldn't this error instead of returning null?
```
dholms, line 418: hmmm why's that? 🤔

packages/repo/tests/mst.test.ts
```diff
@@ -277,5 +277,256 @@ describe('utils', () => {
...
+    expect(countPrefixLen('abcde', 'abc1')).toBe(3)
+    expect(countPrefixLen('abcde', 'abb')).toBe(2)
+    expect(countPrefixLen('abcde', 'qbb')).toBe(0)
+    expect(countPrefixLen('', 'asdf')).toBe(0)
+    expect(countPrefixLen('abc', 'abc\x00')).toBe(3)
+    expect(countPrefixLen('abc\x00', 'abc')).toBe(3)
+  })
+
+  it('counts string (not byte) prefix length', () => {
+    // TODO: these are not cross-language consistent
```
dholms, line 291: hopefully we can wipe all this out soon 😛

bnewbold: indeed! was a good dry run for discussion about computing post lengths as well

packages/repo/src/mst/mst.ts
bnewbold: Hrm, maybe just a different norm/expectation in Javascript land (vs python/go/rust).

Checking behavior of array indexing real quick, what about `undefined`?

packages/repo/src/mst/mst.ts
bnewbold: It looks like all call sites of `create()` already setting the layer explicitly, also. I felt like it should be "not lazy loaded yet" until a layer was actually computed. `layer=0` is an actual well-defined layer.

Changed to `null` and tests pass.

bnewbold:
Pushed small changes, which I think resolved everything. One small question back to Daniel above, but that is just my curiosity.

I think this is ready to merge, and there are no particular concerns for deployment that I am aware of. There is just that tiny `0 -> null` change to the actual code, everything else is just additional test coverage.

I plan to merge to main tomorrow, and will let it roll-up with other changes for deployment.

packages/repo/src/mst/mst.ts
dholms: Yeah we call this in quite a few places where we expect the index to possibly be out of bounds

I think you're write that `undefined` is probably the better match in this case. Doesn't affect behavior at all:
```ts
  async atIndex(index: number): Promise<NodeEntry | undefined> {
    const entries = await this.getEntries()
    return entries[index]
  }
```

dholms (approved):
all looks good!

PR #563, opened 2023-02-16
Repo sync semantic changes

dholms:
A couple changes to our sync stream semantics.

Two main changes here:
- we use a cursor instead of time for backfill. we were already doing this internally in the server, so we just expose that value
- we flatten the repo_append & repo_rebase event types into one event type. the implication of this is that we'll likely be sending carSlices on rebase as well

A couple choices I made for the stream to deal with these new semantics:
- if the provided cursor would start the stream _before_ the available backfill time we:
  - send an InfoFrame with code `OutdatedCursor`
  - start the stream from the available backfill time
- if the provided cursor is greater than the highest cursor, we throw a `FutureCursor` error
  - The reason for throwing & not just serving from present or waiting for that future, is that a consumer should never have a reason to provide a future cursor. So it's likely some botched book keeping on their end that led to it
  - If a consumer just wishes to receive all _new_ events, then they can provide no cursor at all


devinivy (approved):
Lookin good 👍 Nice to see infos and errors doing their thing.

Issue #549, opened 2023-02-14
Add `com.atproto.sync.getBlocks` XRPC method


dholms:
A simple sync route for requesting one or more blocks by CID. 

Used in the case where you find a hole in your repo & don't know which commit that block is from

PR #564, opened 2023-02-17
Add `com.atproto.sync.getBlocks` XRPC method

dholms:
A simple sync route for requesting one or more blocks by CID.

Used in the case where you find a hole in your repo & don't know which commit that block is from

Closes https://github.com/bluesky-social/atproto/issues/549

### DIFF
--- a/packages/pds/src/api/com/atproto/sync/getBlocks.ts
+++ b/packages/pds/src/api/com/atproto/sync/getBlocks.ts
@@ -1,0 +1,25 @@
+import { CID } from 'multiformats/cid'
+import { InvalidRequestError } from '@atproto/xrpc-server'
+import { Server } from '../../../../lexicon'
+import SqlRepoStorage from '../../../../sql-repo-storage'
+import AppContext from '../../../../context'
+import { blocksToCar } from '@atproto/repo'
+
+export default function (server: Server, ctx: AppContext) {
+  server.com.atproto.sync.getBlocks(async ({ params }) => {
+    const { did } = params
+    const cids = params.cids.map((c) => CID.parse(c))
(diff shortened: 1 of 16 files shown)


packages/pds/src/api/com/atproto/sync/getBlocks.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export default function (server: Server, ctx: AppContext) {
+  server.com.atproto.sync.getBlocks(async ({ params }) => {
+    const { did } = params
+    const cids = params.cids.map((c) => CID.parse(c))
+    const storage = new SqlRepoStorage(ctx.db, did)
+    const got = await storage.getBlocks(cids)
+    if (got.missing.length > 0) {
+      const missingStr = got.missing.map((c) => c.toString())
+      throw new InvalidRequestError(`Could not find cids: ${missingStr}`)
```
devinivy, line 16: Just asking for good measure—would there be any value in exposing `missing` in the output rather than making it an error?

devinivy: Ah, I guess there's not really a good place for it, is there?

devinivy (approved):
Big ✅

packages/pds/src/api/com/atproto/sync/getBlocks.ts
dholms: yeah i thought about that 🤔 

I'm not opposed to just _not_ returning them & making the client figure it out. 

My main thought is that this route should be used for relatively few cids at a time & those cids are like such that if you don't have them _all_ you can't do whatever you were trying to do

Issue #496, opened 2023-01-26
Content sniffing on PNG uploads erroring


dholms:
`file-type` library throws an `EndOfStreamError`

bnewbold:
Oof, yeah, this is a real footgun. Ran in to it today and ground gears for a while.

dholms:
Yup I'd like to fix it this week. Shouldn't be too bad & it's a pretty glaring issue

PR #570, opened 2023-02-20
Fix content sniffing on streams

dholms:
This wasn't an issue with the library, I was just using it wrong 🙃 

The `file-type` lib only reads as far as it needs to then stops reading. 

So we must `clone` the stream, read till we find the file type, then destroy the stream so it doesn't hold up the other processors

Closes https://github.com/bluesky-social/atproto/issues/496


devinivy (approved):
Nice and weird!  Good stuff.

PR #572, opened 2023-02-21
Logical separation of app view

devinivy:
The goal here is to isolate app view functionality for the bsky application logically from the rest of the PDS, which is more focused on hosting.  The app-view functionality lives inside `packages/pds/src/app-view/`, where lots of files have been moved or adapted out of `packages/pds/src/`.

#### Database layer
In the db layer the following tables have been separated into the app view, all related to indexing bsky records: `duplicate_record`, `assertion`, `profile`, `post`, `post_hierarchy`, `post_entity`, `post_embed_image`, `post_embed_external`, `vote`, `repost`, `follow`.

A few exceptions to the factoring: `vote` is shared by the `feed.setVote` method in the PDS for now.  The `repo_root` and `record` tables are shared so that the app view can tell if a repo/record is moderated.  And `did_handle` is used by both.  There is also a little tangle with `mutes`, which logically lives under the PDS but is used in the app-view.

#### Service layer
The actor service has been split into the PDS's account service and the app-view's actor service.  The PDS inherits the `com.atproto` terminology ("account"), where app-view has the `app.bsky` terminology ("actor").  All actor view logic is isolated within the the app-view, though it is currently used in two places in the PDS: `graph.getMutes` and `notification.list` methods.

The record service has been split into the PDS's record service and the app-view's indexing service.  In short, all bsky-specific indexing logic from the record service (i.e. the plugin/processing system) has been split out into the app-view's indexing service.

In order to maintain a loose coupling between these two services, a `MessageDispatcher` implementing `MessageQueue` that handles messages immediately (i.e. in the same transaction as the `send()`) has been introduced.  For example, when the user writes a record to their repo the PDS dispatches an "index_record" message.  The app-view receives this message, performs any bsky-specific indexing, then may push some "create_notification" messages onto the message queue, which are picked-up by the PDS.

The services for the app-view live under `services.appView`.  I see the service layer as sort of the midpoint of the architecture (between method handlers and the db model), so it's a nice place to have any coupling between app-view and PDS explicitly documented.

Still a few things to untangle, but now there's some proper logical separation between the PDS and app-view, and the overlap between the two is documented here as well as with `@NOTE` or `@TODO`s in the code.


packages/pds/src/app-view/services/indexing/index.ts
```diff
@@ -0,0 +1,125 @@
...
+      this.db.db
+        .deleteFrom('post_embed_image')
+        .where('post_embed_image.postUri', 'in', postByUser)
+        .execute(),
+      this.db.db
+        .deleteFrom('post_embed_external')
+        .where('post_embed_external.postUri', 'in', postByUser)
+        .execute(),
+      this.db.db
+        .deleteFrom('duplicate_record')
```
dholms, line 95: as we focus on backlinks more, we may want to make `duplicate_record` a bit smarter

dholms (approved):
sorry for the wait for review 

this is all coming together quite nicely!

PR #576, opened 2023-02-22
lexicon-level identifier validation helpers and test cases

bnewbold:
These are partially redundant with existing Typescript packages to validate handles, NSIDs, ATURIs, and DIDs. There are also exist third party libraries to validate some of these.

The motivation to re-implement these here is that we might want to validate differently at the lexicon level. Eg, we should be "permissive" in our validation if we are going to reject records entirely (at least, with record validation enabled) because they don't pass these validations. As a specific example, we have a bunch of extra restrictions on registering handles currently, like a 20 char limit, which we probably don't want to enforce on records from third parties in the network.

A second motivation is to have tested examples of how to validate these fields, especially using "just" regexes, which other implementations can copy/paste from with confidence.

Another motivation is to just independently look all these over as part of hardening review work.

Work in progress. Will be pulling most of the TODOs out into the related notion doc for discussion there. Do want to share this progress though.


bnewbold:
Missing currently are TID tests.

It might also be good to have "datetime" validation in here (in the context of that as a Lexicon string format). Eg, corner cases of timezone formatting, maybe we want them to always be UTC and `Z` suffix. Inserted leap seconds. Millisecond or microsecond timestamp precision.

packages/lexicon/src/identifiers.ts
```diff
@@ -0,0 +1,317 @@
...
+// This longer handle verifier gives understandable errors
+export const lexVerifyHandle = (handle: string): void => {
+  // check that all chars are boring ASCII
+  if (!/^[a-zA-Z0-9.-]*$/.test(handle)) {
+    throw new Error(
+      'Disallowed characters in handle (ASCII letters, digits, dashes, periods only)',
+    )
+  }
+
+  if (handle.length > 253) {
```
dholms, line 40: The existing handle constraints are split into two sections: handle constraints & service constraints. the 40 max char total & 20 max char on each part is meant to be a handle constraint, enforced across the whole network.

Im personally still pro: "no handles over 40 chars". I just think it's bad ux & i kinda wanna enforce that everywhere we can to set a hard constraint. We already have extra constraints over domains. ie enforcing boring ascii

Do you much prefer just enforcing domain length constraints here instead?

packages/lexicon/src/identifiers.ts
```diff
@@ -0,0 +1,317 @@
...
+
+export const lexVerifyDidRegex = (did: string): void => {
+  // simple regex to enforce most constraints via just regex and length.
+  // hand wrote this regex based on above constraints
+  if (!/^did:[a-z]+:[a-zA-Z0-9._:%-]*[a-zA-Z0-9._-]$/.test(did)) {
+    throw new Error("DID didn't validate via regex")
+  }
+
+  if (did.length > 8 * 1024) {
+    throw new Error('DID is far too long')
```
dholms, line 186: haha 😆

packages/lexicon/src/identifiers.ts
```diff
@@ -0,0 +1,317 @@
...
+  }
+  if (handle.length > 253) {
+    throw new Error('Handle is too long (253 chars max)')
+  }
+}
+
+// Human readable constraints on NSID:
+// - a valid domain in reversed notation. which means the same as a loose domain!
+// - "nsid-ns" is a special situation, not allowed here
+// - not clear if final "name" should be allowed in authority or not
```
dholms, line 84: dont quite grok what this means

dholms:
Overall looking really good 👍 

I'd like to not have multiple code paths for validation of stuff. Thoughts on moving each of these validators into their respective package?

If there are cases where we need additional validation on top of this, I would steal like to use these validators & just layer on additional validation alongside it

packages/lexicon/src/identifiers.ts
bnewbold: I don't think we are enforcing anything extra on top of domains? Punycode is still allowed in the handles themselves (which are how international domains work).

If we advertise "bring your own domain" as a feature, it does feel weird to then disallow some domains. I feel like the vast, vast majority of folks will use short handles, and we can truncate the rest at display time.

But, this is ultimately just a protocol decision and I would defer. 40 chars is fairly long.

packages/lexicon/src/identifiers.ts
bnewbold: eg, in the case of `com.example.fun.in.sun.someThing`, the `someThing` is not part of the authority, but is `sun`?

IIRC (didn't verify now), the NSID spec says it is, so maybe not ambiguous.

bnewbold:
I rebased on current HEAD and pushed a couple small typos in the comments, and clarified that single-char domains are allowed (I think this is different from current handle validation?).

I'm open to having these refactored in to whatever location. The main goal is to have consistent definitions and test cases that we can use across implementations.

I'm going to need to stay focused on other work this week. I can just hand this PR off to you for you to do as you like? Eg, refactor, modify, split out in to other packages as you see fit.

packages/lexicon/src/identifiers.ts
bnewbold: _/me dashes out to register a very long domain name just to troll @dholms_

dholms:
Alrighty @bnewbold got these all integrated into their respective libraries. Also took the oppurtunity to rename the `handle` library to `identifier` & add did validation in there as well

bnewbold:
Looks groovy, loving it!

PR #578, opened 2023-02-22
Bump package versions

dholms:
Just bumping package versions for common & crypto as I'm using them in the new plc repo


devinivy (approved):
I think it's nice for the commit w/ version bump to be a side-effect of the release process itself, but either way lookin good 👍

PR #601, opened 2023-02-23
Add standard container image labels

Jacob2161:
Not required but nice to have them in there so they can be used by various UIs.


dholms (approved):
👍 

i'll copy the PLC ones over to my new PLC repo as well 👀

PR #603, opened 2023-02-24
XRPC subscription consumer

devinivy:
Taking care of the ceremony around consuming an XRPC subscription:
 - Implementation is not integrated into core XRPC client yet, mostly to avoid compatibility issues with the web, etc. where subscriptions aren't used today.
 - Handles reconnects with backoff, jitter, and param updates.
 - Handles message validation.
 - Handles streaming error semantics.


packages/xrpc-server/src/stream/stream.ts
```diff
@@ -11,3 +12,15 @@ export async function* byFrame(ws: WebSocket, options?: DuplexOptions) {
...
 }
+
+export async function* byMessage(ws: WebSocket, options?: DuplexOptions) {
+  for await (const frame of byFrame(ws, options)) {
+    if (frame.isMessage()) {
+      yield frame
+    } else if (frame.isError()) {
+      throw new XRPCError(-1, frame.code, frame.message)
+    } else {
+      throw new XRPCError(ResponseType.Unknown, undefined, 'Unknown frame type')
```
dholms, line 23: I see that Infos are still around but not handled here, will they get thrown?

packages/xrpc-server/src/stream/subscription.ts
```diff
@@ -0,0 +1,137 @@
...
+}
+
+export default Subscription
+
+class AbnormalCloseError extends Error {
+  code = 'EWSABNORMALCLOSE'
+}
+
+function isReconnectable(err: unknown, reconnects: number | null): boolean {
+  // The socket never opened on initial connection.
```
dholms, line 73: shouldn't we try to reconnect if initial connection failed due to network error?

packages/xrpc-server/src/stream/stream.ts
devinivy: Good eye.  That's right, they will get thrown.  I figured it didn't make sense to figure out how to fit them into this interface since they are about to disappear, should have left a note about that!  Think it's alright?

packages/xrpc-server/src/stream/subscription.ts
devinivy: I struggled with this one.  The thing you do ideally want to be able to catch on init is any form of misconfiguration.  So then the question becomes, how do you distinguish misconfiguration from downtime on the producer side?  And that becomes hard to answer.  Any thoughts on how to slice this?

devinivy: I talked to @Jacob2161 about this a little bit.  He recommended having reconnection logic for the initial connection but maybe making it a bit more aggressive than the backoff strategy (e.g. repeatedly trying once a second) that occurs when you lose an existing connection.

If the app view weren't also serving XRPC requests I think it would make more sense to hard error and let Docker recycle the container as I'm doing right now, but since that's not the case I agree that we should have reconnection logic on the initial connection.  To mitigate the misconfiguration question I can make sure there's good logging going on.

devinivy:
@dholms should be ready for another peek!
 - Added reconnection for initial setup, attempting a connection every second.
 - Added a hook so that reconnects can be logged.
 - Added cancellation to subscriptions, and used this mechanism to initiate an error/reconnect on abnormal aborts.

packages/xrpc-server/src/stream/stream.ts
dholms: yup that's fine 👍 

I'd say just leave a note for now. or yield them as that will more closely match the actual behavior

packages/xrpc-server/src/stream/subscription.ts
dholms: nice that all sounds great to me 👍

dholms (approved):
looks excellent  🙌

packages/xrpc-server/src/stream/stream.ts
devinivy: Good call, I've started yielding them here 👍

PR #604, opened 2023-02-27
Running a task as a leader in the app view

devinivy:
Adds some tooling to help coordinate multiple app views so that only one of them— i.e. the leader—runs a given task at a time.  This will be combined with #603 in order to ensure that there's only one consumer of events from the BGS at a time.


dholms (approved):
sweet looks great 💪

Issue #523, opened 2023-02-03
Basic access to original blob file via Lexicon endpoint


bnewbold:
We should have a simple XRPC endpoint to fetch blobs by CID.

A simple version might:
- take just a CID as a parameter
- always return bytes, with HTTP content type `application/octet-stream`
- check for hard-takedowns and return HTTP `410 Gone` in that case (distinct from 404)

In the future we might also implement a "get all blobs for repo as CAR" endpoint, but we shouldn't block on that. Current motivations are to allow third-party content labeling (on original images), and crude user content export (eg, pull out user's own images).

dholms:
hmm should the content-type always be `octet-stream` or should the pds offer up whatever content type it figured it to be (whether proposed by user or sniffed)

devinivy:
In my view the PDS should provide the content-type to the best of its ability, with octet-stream as an acceptable fallback.

bnewbold:
Good question, I think it is important to get this right.

If we serve correct `Content-Type`, that opens the door to folks deeplinking images without a proxy or cache. Maybe that is fine? But maybe expensive.

I generally think of filetype detection as being pretty non-deterministic and subjective (and changing over time). Maybe that is trauma from working at archive.org with many millions of horrible `application/pdf` blobs. One thing that comes to mind are those wonderful/horrible PoC||GTFO "file format polyglots" which are valid PDF files, ZIP files, and PDF files. All at the same time, exact byte strings.

We could add more context to each GET request. In the full-context version, the ATURI (or repo+collection+rkey) would be specified along with the CID requested. I think in that case the content type is pretty unambiguous? Unless a user is clever/evil and uploads the same blob under different types on the same post. Probably not worth worrying about.

A partial-context version would be to supply the repo (DID) along with the blob CID. That would let PDS do bandwidth accounting and/or tiered storage or something if it wanted? And scope down the content-type question a bit more.

I'm probably over-thinking it and we should just set the `Content-Type` to whatever metadata the PDS can find first for the blob. This might be inconsistent or weird in corner-cases, but we can let downstream users sort that out.

dholms:
Hmm all good points 🤔 

You kind've have me convinced here. This is "raw data route" & a service may serve it with the proper content-types on a separate route if they like (ie some sort of "gateway" route)

[...]

PR #606, opened 2023-02-28
Get blobs

dholms:
Added blobs getter method

Had a lil bit of feature creep on this
- Refactored the repo routes when I thought this was in the repo namespace
- added `creator` to the blobs table so that we track blob metadata within the context of a given repo
  - I've been meaning to do this for a while but the convo we had in the issue was good motivation to get it done

Closes https://github.com/bluesky-social/atproto/issues/523


packages/xrpc-server/src/server.ts
```diff
@@ -222,8 +223,12 @@ export class Server {
             output?.encoding === 'json'
           ) {
             res.status(200).json(output.body)
-          } else if (output) {
+          } else if (output?.body instanceof Readable) {
             res.header('Content-Type', output.encoding)
+            res.status(200)
+            forwardStreamErrors(output.body, res)
+            output.body.pipe(res)
```
dholms, line 230: still confused when to use `pipeline` vs this tbh 😅

lexicons/com/atproto/repo/getBlob.json
```diff
@@ -0,0 +1,20 @@
...
+      "description": "Fetch a blob.",
+      "parameters": {
+        "type": "params",
+        "required": ["cid"],
+        "properties": {
+          "cid": {"type": "string", "description": "The CID of the blob to fetch"}
+        }
+      },
+      "output": {
+        "encoding": "application/octet-stream"
```
devinivy, line 16: Even if we choose to serve the files as application/octet-stream, I'd still be +1 to leave it open `*/*` in the lex spec.  My basic argument is knowledge is power haha— I don't think you really lose anything by allowing hosts to say what they consider the bytes that they are serving to represent.

packages/xrpc-server/src/server.ts
devinivy: Seems good as-is to me here! 👍

devinivy (approved):
I know it's just a draft, but this is lookin good!

packages/pds/src/api/com/atproto/sync/getBlob.ts
```diff
@@ -0,0 +1,26 @@
...
+      .executeTakeFirst()
+    if (!found) {
+      throw new InvalidRequestError(`blob not found: ${params.cid}`)
+    }
+    const cid = CID.parse(params.cid)
+    const blobStream = await ctx.blobstore.getStream(cid)
+    res.setHeader('Content-Length', found.size)
+    // @TODO better codegen for */* mimetype
+    return {
+      encoding: 'application/octect-stream' as any,
```
devinivy, line 22:
```suggestion
      encoding: 'application/octet-stream' as any,
```

packages/pds/src/db/migrations/20230210T210132396Z-post-hierarchy.ts
```diff
@@ -41,11 +41,15 @@ export async function up(
...
     .selectFrom('hierarchy')
 
-  await db
-    .insertInto('post_hierarchy')
-    .columns(['uri', 'ancestorUri', 'depth'])
-    .expression(postHierarchyQb.select(['uri', 'ancestorUri', 'depth']))
-    .execute()
+  const res = await postHierarchyQb.selectAll().limit(1).execute()
+
+  if (res.length > 0) {
```
devinivy, line 46: Did some kind of error pop out of this?

packages/pds/tests/moderation.test.ts
```diff
@@ -568,8 +568,8 @@ describe('moderation', () => {
 
     it('only allows blob to have one current action.', async () => {
       const img = sc.posts[sc.dids.carol][0].images[0]
-      const postA = await sc.post(sc.dids.alice, 'image A', undefined, [img])
-      const postB = await sc.post(sc.dids.alice, 'image B', undefined, [img])
+      const postA = await sc.post(sc.dids.carol, 'image A', undefined, [img])
+      const postB = await sc.post(sc.dids.carol, 'image B', undefined, [img])
```
devinivy, line 572: Just curious what called for this change.

packages/pds/src/api/com/atproto/sync/getBlob.ts
dholms: oh god. bless you

packages/pds/tests/moderation.test.ts
dholms: these blobs were uploaded by carol so we needed to use them in her posts instead.

since we're associating blobs with users now!

packages/pds/src/db/migrations/20230210T210132396Z-post-hierarchy.ts
dholms: oh yup. but i think it was a previous boneheaded mistake that i think was causing it. think i can remove it now

packages/pds/src/db/migrations/20230301T003452036Z-blob-creator.ts
```diff
@@ -0,0 +1,76 @@
...
+      'size',
+      'tempKey',
+      'width',
+      'height',
+      'createdAt',
+    ])
+    .expression((exp) =>
+      exp
+        .selectFrom('blob')
+        .innerJoin('repo_blob', 'repo_blob.cid', 'blob.cid')
```
devinivy, line 33: One potential issue that could happen here is that the same blob is in multiple commits for the same user.  I think that would lead to a constraint violate on the primary key.  It's possible just adding an on conflict clause would patch that up though.

packages/pds/tests/migrations/blob-creator.test.ts
```diff
@@ -0,0 +1,142 @@
+import { Database } from '../../src'
+import { randomStr } from '@atproto/crypto'
+import { cidForCbor, TID } from '@atproto/common'
+import { Kysely } from 'kysely'
+import { APP_BSKY_SYSTEM } from '../../src/lexicon'
+import { AtUri } from '@atproto/uri'
+
+describe('user table did pkey migration', () => {
```
devinivy, line 8:
```suggestion
describe('blob creator migration', () => {
```

devinivy (approved):
Excellent stuff!  Here are a few cases I was unsure about after review, though I'm not certain all/any of them actually require changes:

 - Two different users upload the same blob in succession.  I think that the blobstore would end-up trying to make the same blob permanent twice now (whereas before the singular `tempKey` state ensured it would happen once), and I'm not sure that all the blobstore implementations were originally designed with that in mind.  I'm definitely fine with making `blobstore.makePermanent()` formally a no-op when the blob is already permanent, seems like it should work that way regardless.
 - The one I mentioned in the migration comment: prior to migration, one user used the same blob in multiple commits/records.
 - The one I mentioned in the repo blob service comment: the user uploads a blob, it disappears from temp storage, then the user uploads and tethers a blob.

packages/pds/src/db/migrations/20230301T003452036Z-blob-creator.ts
dholms: that's a good point 👍 
I think just the on conflict will do. I think it's reasonable to say "a given blob can only be used one way per repo"

packages/pds/src/services/repo/blobs.ts
```diff
@@ -40,12 +42,7 @@ export class RepoBlobs {
         height: imgInfo?.height || null,
         createdAt: new Date().toISOString(),
       })
-      .onConflict((oc) =>
-        oc
-          .column('cid')
-          .doUpdateSet({ tempKey })
-          .where('blob.tempKey', 'is not', null),
-      )
+      .onConflict((oc) => oc.doNothing())
```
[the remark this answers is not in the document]
dholms: ah yup i see. got a bit carried away. nice catch!

dholms:
Sweet thanks for the feedback! 

I'm down making `makePermanent` of the same CID a no-op. That's already how it worked on the disk blobstore.

S3 would've worked, but it was overriding the last set value with the new one. I added a check to see if it was stored already & if so, just delete from temp without moving 👍

devinivy (approved):
Looks good!

bnewbold:
Shipping image labeling work is blocked on this branch, would be great to get it merged soon.

I have a locally rebased branch for testing, on top of the repo v2 merge commit on `main` (not on top of current `main` HEAD). Had to rename/reorder database migration. There is a conflict with `main` HEAD where some repo files got entirely moved/refactored.

devinivy:
@bnewbold I think we should be able to get this merged before too long today 👍

PR #610, opened 2023-03-01
Enable app-view to build indexes from subscription to a repo host

devinivy:
This builds on top of #603 and #604, to integrate the app-view with federation/sync.  When configured with `APP_VIEW_REPO_HOST`, the app view will not use in-process sync, and instead build indexes from events coming over `com.atproto.sync.subscribeAllRepos` from the repo host.

Still need to do a bit of massaging of how the indexes are built, in particular to ensure they do not rely on record ordering across repositories.  We'll also need to handle repo moves and handle updates— this essentially just moves in-process updates from the PDS to out-of-process updates via federation.


packages/pds/src/app-view/subscription/repo.ts
```diff
@@ -0,0 +1,187 @@
...
+} from '@atproto/repo'
+import { PreparedWrite } from '../../repo'
+import { OutputSchema as Message } from '../../lexicon/types/com/atproto/sync/subscribeAllRepos'
+import { lexicons } from '../../lexicon/lexicons'
+import Database from '../../db'
+import AppContext from '../../context'
+import { Leader } from '../../db/leader'
+import { appViewLogger } from '../logger'
+
+const METHOD = 'com.atproto.sync.subscribeAllRepos'
```
dholms, line 21: you can make this typesafe by importing `ids` from `lexicon/lexicons` & using `ids.ComAtprotoSyncSubscribeAllRepos`

packages/pds/src/app-view/subscription/repo.ts
```diff
@@ -0,0 +1,187 @@
...
+export const REPO_SUB_ID = 1000
+
+export class RepoSubscription {
+  leader = new Leader(REPO_SUB_ID, this.ctx.db)
+  destroyed = false
+  constructor(public ctx: AppContext, public service: string) {}
```
dholms, line 27: we'll want this to support multiple providers eventually, but not needed rn

packages/pds/src/app-view/subscription/repo.ts
```diff
@@ -0,0 +1,187 @@
...
+          { err, reconnects, initial },
+          'repo subscription reconnect',
+        )
+      },
+      validate: (value) => {
+        try {
+          return lexicons.assertValidXrpcMessage<Message>(METHOD, value)
+        } catch (err) {
+          appViewLogger.warn(
+            { err, service: this.service },
```
dholms, line 132: maybe worth logging the message or some parts of it if possible?

this may help with debugging BGS if it sends bad messages 🤔

dholms (approved):
couple small things & questions, but overall this looks really great 💪 I'm stoked

packages/pds/src/app-view/subscription/repo.ts
```diff
@@ -0,0 +1,187 @@
...
+          throw new Error('Repo sub completed, but should be persistent')
+        }
+      } catch (err) {
+        appViewLogger.error(
+          { err, service: this.service },
+          'repo subscription errored',
+        )
+      }
+      if (!this.destroyed) {
+        await wait(5000 + jitter(1000))
```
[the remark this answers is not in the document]
devinivy: Yeah, that's right.  If the leader goes down, this is how it would get picked back up.

packages/pds/src/app-view/subscription/repo.ts
devinivy: Ah nice, forgot about that!

packages/pds/src/app-view/subscription/repo.ts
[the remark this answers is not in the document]
dholms: yeah that's fine for now 🤷 just wanted to make sure i understood

devinivy:
@dholms should be good for a final peek if you're interested!

PR #599, opened 2023-02-23
Sync subscription tweaks

dholms:
Couple tweaks
- `prev` is nullable instead of optional
- `cursor` can be set to `0`

Closes https://github.com/bluesky-social/atproto/issues/577

PR #611, opened 2023-03-01
Ensure subscribeAllRepos syncs from cursor 0

devinivy:
Just some follow-on to #599 to ensure `subscribeAllRepos` behaves correctly when the `cursor` param is 0.


dholms (approved):
oh beautiful. thanks for noticing that 🙇

PR #614, opened 2023-03-02
Repo ops on sync stream

dholms:
Adds repo ops to sync stream.

We track ops alongside writes.

We don't bother backfilling old ops, instead we just reset the sequenced history (by deleting old rows)

### DIFF
--- a/lexicons/com/atproto/sync/subscribeAllRepos.json
+++ b/lexicons/com/atproto/sync/subscribeAllRepos.json
@@ -17,7 +17,7 @@
       "message": {
         "schema": {
           "type": "object",
-          "required": ["seq", "event", "repo", "commit", "prev", "blocks", "blobs", "time"],
+          "required": ["seq", "event", "repo", "commit", "prev", "blocks", "ops", "blobs", "time"],
           "nullable": ["prev"],
           "properties": {
             "seq": {"type": "integer"},
@@ -32,7 +32,10 @@
             "commit": {"type": "string"},
             "prev": {"type": "string"},
             "blocks": {"type": "unknown"},
-            "ops": {"type": "ref", "ref": "#repoOp"},
+            "ops": {
+              "type": "array", 
+              "items": { "type": "ref", "ref": "#repoOp"}
+            },
             "blobs": {
               "type": "array",
               "items": {"type": "string"}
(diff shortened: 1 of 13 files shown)


packages/pds/src/sequencer/index.ts
```diff
@@ -112,16 +112,32 @@ export class Sequencer extends (EventEmitter as new () => SequencerEmitter) {
       .selectFrom(seqQb.as('repo_seq'))
       .innerJoin('repo_blob', (join) =>
         join
-          .onRef('repo_blob.did', '=', 'repo_blob.did')
+          .onRef('repo_blob.did', '=', 'repo_seq.did')
```
devinivy, line 115: 👀 💎

lexicons/com/atproto/sync/subscribeAllRepos.json
```diff
@@ -46,6 +50,16 @@
       "errors": [
         {"name": "FutureCursor"}
       ]
+    },
+    "repoOp": {
+      "type": "object",
+      "required": ["action", "path", "cid"],
+      "nullable": ["cid"],
+      "properties": {
+        "action": {"type": "string"},
```
devinivy, line 59: I wonder if there could be some value in tacking `knownValues` on here.  Since the actions are closed (create, update, delete) I don't think that tokens would be needed.

packages/pds/src/db/tables/repo-op.ts
```diff
@@ -0,0 +1,11 @@
+export interface RepoOp {
+  did: string
+  commit: string
+  action: string
```
devinivy, line 4: Whether or not it ends-up in the lex schema, we could type this as `WriteOpAction`.

devinivy (approved):
That was quick!  Looks great.

packages/pds/src/sequencer/index.ts
dholms: yeah can't believe that one slipped through 😅

lexicons/com/atproto/sync/subscribeAllRepos.json
dholms: yup that's a good point 👍

PR #618, opened 2023-03-02
Ensure app-view is permissive of reindexing

devinivy:
This ensures that the app-view indexing logic allows reprocessing repo operations.  This will come in handy to fill-in any skipped events post-facto, replay history to fix buggy indexes, and just generally be prepared for the eventuality of consuming the same repo ops multiple times.

I also started starting requiring that a timestamp be provided for indexing, since that seems to be the core use-case and it makes the service more deterministic.


dholms (approved):
looks great 🙌

PR #624, opened 2023-03-03
Lengthen token lifetimes

devinivy:
The refresh token lifetime can be much longer, particularly because they are rotated on each use.  This bumps their lifetime up from 7 to 90 days.

Our access token lifetimes were also quite conservative, and have been lengthened for more practical/everyday usage.  This bumps their lifetime up from 15 to 120 mins.


pfrazee:
👍

dholms (approved):
nice 👍

PR #626, opened 2023-03-04
Fixing typo in com.atproto.handle.resolve

ngerakines:
This looks like a copy/paste typo where parameters had some nested values that didn't align with the spec.


ngerakines:
I think the documentation was just outdated and didn't reference the XRPC Parameters.

PR #628, opened 2023-03-04
Feed and search query perf tuning

devinivy:
In the author feed, timeline feed, and user search there are cases where we're using UNION where we can be using UNION ALLs, which tells the database it doesn't need to do the work of deduping the result set.  Here's why we can switch them in each case:
 - The timeline feed unions posts and reposts, and they each have a fixed `type` column which is either `'post'` or `'repost'`.  So we know ahead of time that there will not be duplicates across the result sets, as the `type` can't match.
 - The author feed works the same as the timeline in this respect 👆 
 - The user search queries union accounts and profiles, then chooses the best result from their combination.  In this case we are already deduping after the union using a `distinct on`, so we don't need to ask postgres to do the work of deduping the union.


dholms (approved):
nice 💪

PR #630, opened 2023-03-06
Fix README of handle package

jiftechnify:
It seems that 4th example is meant to demonstrate  unhappy path of `ensureValid`, but not `isValid`.


devinivy (approved):
Nailed it, thanks!

PR #631, opened 2023-03-06
Enable indexing of record updates, consume repo ops on sync subscription

devinivy:
This accomplishes a few things:
 - Prepare the indexer for handling record updates (in addition to creates and deletes).
 - Update `app.bsky.actor.updateProfile` to take advantage of that new machinery.
 - Update repo subscription to handle repo ops from #614.
 - Make some updates to notification generation so that updating a post, for example, doesn't create duplicate notifications.

This does not enable updates in `batchWrite` or `putRecord`, but we should be good to enable that whenever we see fit.


packages/pds/src/app-view/services/indexing/index.ts
```diff
@@ -107,7 +117,6 @@ export class IndexingService {
         .execute(),
     ])
     await Promise.all([
-      this.db.db.deleteFrom('record').where('did', '=', did).execute(),
```
devinivy, line 110: This moved to the PDS, since the record table really isn't governed by the app view.

packages/pds/src/app-view/services/indexing/processor.ts
```diff
@@ -83,7 +85,64 @@ export class RecordProcessor<T, S> {
...
+      )
+    }
+    return this.params.eventsForDelete(deleted, inserted)
+  }
+
   async deleteRecord(uri: AtUri, cascading = false): Promise<Message[]> {
+    await this.db
+      .deleteFrom('duplicate_record')
+      .where('uri', '=', uri.toString())
+      .execute()
```
devinivy, line 145: It occurred to me that deleted records could have been hanging around as dupes, which could allow them to later get indexed after their deletion.

packages/pds/tests/subscription/repo.test.ts
```diff
@@ -138,7 +143,10 @@ describe('sync', () => {
       forSnapshot(originalTl.data.feed),
     )
 
-    await expect(getTableDump()).resolves.toEqual(originalTableDump)
+    // Permissive of indexedAt times changing
```
devinivy, line 146: Reprocessing an update does update the `indexedAt` time.  Ideally I think it wouldn't, but I'm not sure if it's really worth adding the check to work around.

packages/pds/src/app-view/services/indexing/processor.ts
dholms: oh damn nice find 👍

packages/pds/src/app-view/services/indexing/index.ts
```diff
@@ -45,26 +46,35 @@ export class IndexingService {
...
+        ? await indexer.insertRecord(uri, cid, obj, timestamp)
+        : await indexer.updateRecord(uri, cid, obj, timestamp)
     await this.messageQueue.send(this.db, events)
   }
 
   async deleteRecord(uri: AtUri, cascading = false) {
     this.db.assertTransaction()
-    const table = this.findTableForCollection(uri.collection)
-    const events = await table.deleteRecord(uri, cascading)
+    const indexer = this.findIndexerForCollection(uri.collection)
```
dholms, line 67: like this name change 👌

dholms (approved):
looks great 🥇

PR #632, opened 2023-03-07
Handle fixes

dholms:
We were being a bit aggressive with errors & they were getting returned as 500s.

This also changes it so that we do not resolve the domain of the PDS to a DID. We have not yet established whether services in the network should have handles.

### DIFF
--- a/packages/pds/tests/account.test.ts
+++ b/packages/pds/tests/account.test.ts
@@ -449,18 +449,4 @@ describe('account', () => {
       agent.api.com.atproto.session.create({ identifier: handle, password }),
     ).resolves.toBeDefined()
   })
-
-  it('resolves did for account', async () => {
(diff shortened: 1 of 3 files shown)


devinivy (approved):

packages/pds/tests/account.test.ts
```diff
@@ -449,18 +449,4 @@ describe('account', () => {
       agent.api.com.atproto.session.create({ identifier: handle, password }),
     ).resolves.toBeDefined()
   })
-
-  it('resolves did for account', async () => {
-    const resolved = await agent.api.com.atproto.handle.resolve({ handle })
-    expect(resolved.data).toEqual({ did })
-  })
```
devinivy, line 456: This test seems like it may still be good.

devinivy (approved):
Solid!

PR #633, opened 2023-03-07
Format invite codes with dashes

dholms:
Format invite codes with dashes instead of dots so that they are not mistaken for links:

`bsky.social-abc12 -> bsky-social-abc12`

### DIFF
--- a/packages/pds/src/api/com/atproto/account/createInviteCode.ts
+++ b/packages/pds/src/api/com/atproto/account/createInviteCode.ts
@@ -10,9 +10,10 @@ export default function (server: Server, ctx: AppContext) {
       const { useCount } = input.body
 
       // generate a 5 char b32 invite code - preceded by the hostname
-      // ex: bsky.app-abc12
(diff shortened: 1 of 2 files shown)


packages/pds/src/api/com/atproto/account/createInviteCode.ts
```diff
@@ -10,9 +10,10 @@ export default function (server: Server, ctx: AppContext) {
       const { useCount } = input.body
 
       // generate a 5 char b32 invite code - preceded by the hostname
-      // ex: bsky.app-abc12
+      // with '.'s replaced by '-'s so it is not mistakable for a link
+      // ex: bsky-app-abc12
       const code =
-        ctx.cfg.publicHostname +
+        ctx.cfg.publicHostname.replaceAll('.', '-') +
```
devinivy, line 16: `replaceAll()`!  I'm still in the past with global regexes.  Right on.

devinivy (approved):
Perf!

PR #634, opened 2023-03-07
Bugfix: resolving the root of an available domain

dholms:
The server was not properly bailing out here & was recursively calling itself to try to resolve it's own handle


devinivy (approved):
👀 ✅ Looks great

Issue #543, opened 2023-02-09
Flaky test for PDS sync


devinivy:
On occasion this sync test in the PDS flakes out:
```
@atproto/pds: FAIL tests/sync.test.ts (5.811 s)
@atproto/pds:   ● repo sync › sync a record proof
@atproto/pds:     expect(received).toBe(expected) // Object.is equality
@atproto/pds:     Expected: 1
@atproto/pds:     Received: 0
@atproto/pds:       210 |       record: repoData[collection][rkey],
@atproto/pds:       211 |     }
@atproto/pds:     > 212 |     expect(records.length).toBe(1)
@atproto/pds:           |                            ^
@atproto/pds:       213 |     expect(records[0].record).toEqual(claim.record)
@atproto/pds:       214 |     const result = await repo.verifyProofs(
@atproto/pds:       215 |       did,
@atproto/pds:       at Object.<anonymous> (tests/sync.test.ts:212:28)
```

dholms:
oh no i wonder why 🧐

dholms:
oof just got it myself: https://github.com/bluesky-social/atproto/actions/runs/4138359513/jobs/7154672527

PR #636, opened 2023-03-08
MST util bug fixes for sync

dholms:
Two fixes in here: 
- fixes the problem with finding records in proofs
  - we were recursing into `walk` instead of `walkReachable` which was throwing
- fixes an issue with overeager diffs
  - we were tracking "new cids" that weren't actually "new", they were just in a different location in the MST. so now we track "removed" cids as well & dedupe this in a similar way to how we handle adds + deletes

Closes https://github.com/bluesky-social/atproto/issues/543


devinivy (approved):
This is good news 👍 👍

PR #638, opened 2023-03-08
Add hostname param to notifyOfUpdate

dholms:
Small tweak that adds `hostname` param to `com.atproto.sync.notifyOfUpdate`

Also the `required` value on `requestCrawl` was different from the property value so I standardized that on `hostname` as well

### DIFF
--- a/lexicons/com/atproto/sync/notifyOfUpdate.json
+++ b/lexicons/com/atproto/sync/notifyOfUpdate.json
@@ -4,7 +4,14 @@
   "defs": {
     "main": {
       "type": "query",
-      "description": "Notify a crawling service of a recent update. Often when a long break between updates causes the connection with the crawling service to break."
+      "description": "Notify a crawling service of a recent update. Often when a long break between updates causes the connection with the crawling service to break.",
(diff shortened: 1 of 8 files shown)


devinivy (approved):
👍

bnewbold:
Perhaps scope creep for this PR, but can we be more explicit about whether to use `wss://` (SSL)? Eg, is the protocol spec that we must always use `wss://` except for localhost testing and development?

If there are cases where we might use `ws://` it would be nice to have a flag or indication per host. Once speculative example would be complex dev setups with multiple subscriptions, some local (not SSL), some remote (with SSL).

Also, can `hostname` include a port number, if not 80 or 443?

dholms:
I'm gonna get this one in for right now

We need to have a bit more of a discussion about what you bring up, cause there will be many use cases when you want to request a service to make a request to another service.

My preferred solution here is to give services handles & then they can declare their endpoint (including protocol handler + port if they like)

The resolution step is the tricky one there because `com.atproto.handle.resolve` presupposes https 🤔

PR #639, opened 2023-03-09
Add some indices for the record table

dholms:
A couple of indices for record table:
- `(did, cid)` for finding record cids related to a user
  - this will also be useful for rebases
- `(did, collection)` for listing records in a colleciton
  - not actually needed rn, but this would be reasonable to have & speed up some queries elsewhere


devinivy (approved):
Looks great 👍

PR #641, opened 2023-03-09
App migration setup

dholms:
A table & utilities for running application migrations.

Only available in Postgres


packages/pds/tests/db.test.ts
```diff
@@ -174,4 +175,121 @@ describe('db', () => {
...
+        'fulfilled',
+        'fulfilled',
+        'fulfilled',
+        'fulfilled',
+        'fulfilled',
+        'fulfilled',
+      ])
+
+      const after = await db.db.selectFrom('did_handle').select('did').execute()
+      expect(after).toEqual([{ did: 'did:plc:0' }])
```
dholms, line 292: You had this as `0` but I was getting `0` & `0` seems right 🤔 

Thoughts?

devinivy: Yup, that seems right!  Not sure how that happened.

PR #652, opened 2023-03-13
Initial setup of independent bsky app view

devinivy:
This is just the first pass on creating an independent bsky app view.  The basic approach was to peel out all of `pds/src/app-view` plus the other parts of the pds that the in-process app view depended upon, such as the `ipld_block`, `repo_root`, `did_handle`, `record` tables.  Basically, to just get a functioning app view that runs completely independently of the pds.  Here are some details:
 - Culled pds-specific dependencies.
 - Removed pds-specific test files.
 - Setup server and test entrypoints, passing tests on core indexing (but not presentation) logic.
 - Culled database tables, and condensed into a single migration.
 - Removed sqlite dialect, centering on postgres.
 - Removed mutes and notification functionality.
 - Removed dependence on message queue.
 - Updated xrpc method implements to match updated interfaces (e.g. for removed dialect, removed `appView` namespacing under `services`).
 - Temporary auth setup, simply using dids as a bearer token.

The status here is that essentially that the main subscription and indexing logic is working, but there are some updates required to get the xrpc methods serving content properly.  So the `views/` tests aren't passing yet.  There is an incoming PR that gets all the tests passing.

A note, I opted not to `git mv` all these files out of the pds because I think it will be worthwhile to preserve a working pds/app-view combo while this is in flight.  So there's temporarily some duplicate implementation going on between the pds app view and new independent bsky app view.  I propose to essentially freeze development the in-pds app view.  Soon these implementations will diverge, particularly as we update the app view indexers to generate more efficient indexes— that effort will go towards the new bsky app view, but not the in-pds app view.


packages/bsky/Dockerfile
```diff
@@ -0,0 +1,51 @@
+FROM node:18-alpine as build
+
+# Move files into the image and install
+WORKDIR /app
+COPY ./*.* ./
+# NOTE pds's transitive dependencies go here: if that changes, this needs to be updated.
```
devinivy, line 6:
```suggestion
# NOTE bsky's transitive dependencies go here: if that changes, this needs to be updated.
```

packages/bsky/service/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "plc-service",
```
devinivy, line 2:
```suggestion
  "name": "bsky-service",
```

packages/bsky/src/config.ts
```diff
@@ -1,204 +1,72 @@
...
     const publicUrl = process.env.PUBLIC_URL || undefined
-    const hostname = process.env.HOSTNAME || 'localhost'
-    let scheme
-    if ('TLS' in process.env) {
-      scheme = process.env.TLS === '1' ? 'https' : 'http'
-    } else {
-      scheme = hostname === 'localhost' ? 'http' : 'https'
-    }
     const envPort = parseInt(process.env.PORT || '', 10)
     const port = isNaN(envPort) ? 2583 : envPort
```
dholms, line 26: nit, but maybe this is a diff default port from PDS?

packages/bsky/src/index.ts
```diff
@@ -2,90 +2,54 @@
 // this is a relatively non-invasive change to express
 // they get handled in the error.handler middleware
 // leave at top of file before importing Routes
-import 'express-async-errors'
+import 'express-async-errors' // @TODO(bsky) remove
```
dholms, line 5: oh?

I guess XRPC is pretty good at catching errors 🤔

packages/bsky/tests/_util.ts
```diff
@@ -86,21 +68,47 @@ export const runTestServer = async (
...
+  // run app view
+
+  const cfg = new ServerConfig({
+    version: '0.0.0',
+    didPlcUrl: plcUrl,
+    publicUrl: 'https://bsky.public.url',
+    imgUriSalt: '9dd04221f5755bce5f55f47464c27e1e',
+    imgUriKey:
+      'f23ecd142835025f42c3db2cf25dd813956c178392760256211f9d315f8ab4d8',
+    repoProvider: `ws://localhost:${pdsPort}`,
```
dholms, line 102: sick 😎

packages/bsky/package.json
```diff
@@ -8,9 +8,8 @@
     "codegen": "lex gen-server ./src/lexicon ../../lexicons/com/atproto/*/* ../../lexicons/app/bsky/*/*",
```
dholms, line 8: yup might not even need to codegen atproto schemas 👀

dholms (approved):
Nice this looks great 💪 

I'm a _little_ bit nervous about having to keep the appview & in-app pds versions in sync. but hopefully it isn't _that_ bad 😅

packages/bsky/package.json
devinivy: Oooh that would be nice.  Right now I think we only use one or two of the sync lexicons.  Maybe we can special-case those—I will take a look at it once non-repo updates settles!

packages/bsky/src/index.ts
devinivy: Yeah, I think we're pretty well taken care-of tbh!

Issue #580, opened 2023-02-22
Lexicon updates: new types, string formats


pfrazee:
- Add the following types to Lexicon
  - `bytes`
    - In CBOR, uses the binary tag
    - In JSON, looks like `{$type: 'bytes', value: string}` with value using a base64 encoding
  - `cid-internal-ref`
    - In CBOR, uses the link tag
    - In JSON, looks like `{$type: 'cid-internal-ref', value: string}` with value using a base64 encoding
- Remove the `datetime` type
- Add the `format` modifier to strings
  - This should apply additional validation to strings
- Add the following string formats:
  - `datetime`
  - `at-uri`
  - `did`
  - `cid`
- Update all lexicons as follows:
  - Change any `datetime` type to a `string` with the datetime format
  - Add formats to any strings that might benefit from them

bnewbold:
In the example for `cid-internal-ref` for JSON, do you mean {$type: 'cid-internal-ref', value: string}`? If not, it seems like it would not safely round-trip without schema between CBOR and JSON.

pfrazee:
Yes, apologies

dholms:
fixed https://github.com/bluesky-social/atproto/pull/658

PR #653, opened 2023-03-13
Lexicon string formats

dholms:
This removes the `datetime` primitive & introduces a `format` modifier on strings with 4 possibilities:
- `datetime`
- `at-uri`
- `cid`
- `did`

---

Some points of friction that I wouldn't mind a second opinion on:

The fact that I added formats in so many places really called out the fact that we weren't adding them in other places: handle, nsid/collection, email, etc.
I'm not sure if these are necessary to add, but they may be nice. I think handle is probably the one that calls for it the most

Nearly all the time, you just want to give a string one format. However there are a couple cases where this is not the case:
- actor refs: these are `did | handle`, but handle does not have a `format` so they just become `string`
- `subject` in moderation methods: I threw out some options for this in slack. @bnewbold proposed just making these `at-uri` & prefixing dids with `at://` (which is valid). Alternately, since these are admin-facing routes, we just leave the param unformatted.

My opinion on both is that we just leave them as they are:
- no other format modifiers for the time being, add them later if we hit some need for them
- `format` is always just a string. If it needs to be more open than that, then you have to handle validation yourself outside of lex. I don't think either of the examples I had above are going to bite us & this shouldn't ever be an issue at the record layer (where this really matters)

In support of https://github.com/bluesky-social/atproto/issues/580


bnewbold:
I'd be :heavy_plus_sign: to formats for `handle` and `nsid`. Wouldn't be against `email` as well, though the utility is pretty limited.

An option for the `did | handle` case would be a specific string format `did-or-handle`. Not the most aesthetic, but it is a common-enough pattern in our APIs, and likely to be the sort of field/param that winds up in other Lexicons. One way to avoid the multi-type field entirely would be separate `did` and `handle` fields, and some kind of `XOR` requirement between them. That feels even hackier though.

lexicons/com/atproto/report/subject.json
```diff
@@ -6,25 +6,37 @@
...
         }
       },
       "record": {
         "type": "object",
         "required": ["did", "collection", "rkey"],
         "properties": {
-          "did": {"type": "string", "description": "The DID of the repo."},
+          "did": {
+            "type": "string",
+            "format": "cid",
```
devinivy, line 22:
```suggestion
            "format": "did",
```

lexicons/com/atproto/sync/subscribeAllRepos.json
```diff
@@ -28,19 +28,19 @@
                 "rebase"
               ]
             },
-            "repo": {"type": "string"},
-            "commit": {"type": "string"},
-            "prev": {"type": "string"},
+            "repo": {"type": "string", "format": "did"},
+            "commit": {"type": "string", "format": "cid"},
+            "prev": {"type": "string", "format": "cid"},
```
devinivy, line 33: Accurate that some of these will soon become cid-internal-refs?

packages/lexicon/src/validators/primitives.ts
```diff
@@ -248,39 +246,76 @@ export function string(
...
 
-  // type
-  const type = typeof value
-  if (type !== 'string') {
+export function atUriFormat(path: string, value: string): ValidationResult {
+  try {
+    const uri = new AtUri(value)
+    if (!value.startsWith('at://') || uri.toString() !== value) {
+      throw new Error()
+    }
```
devinivy, line 286: I am not sure if we need the `uri.toString() !== value` check.  Two reasons there:
 - The `AtUri` constructor should throw on an invalid uri.
 - It might create false positives since there isn't necessarily a canonical representation of a URI (e.g. trailing slashes, empty search params or hash fragments, unnecessary uri encoding).

Just to toss a possibility out there:
```suggestion
    if (!value.startsWith('at://')) {
      throw new Error() // Fast path
    }
    const uri = new AtUri(value)
    if (!uri.host) {
      throw new Error()
    }
```

packages/lexicon/src/validators/primitives.ts
```diff
@@ -248,39 +246,76 @@ export function string(
...
-      if (typeof value !== 'string' || !isValidISODateString(value)) {
-        throw new ValidationError(
-          `${path} must be an iso8601 formatted datetime`,
-        )
-      }
-    } catch {
-      throw new ValidationError(`${path} must be an iso8601 formatted datetime`)
+export function didFormat(path: string, value: string): ValidationResult {
+  const parts = value.split(':')
+  if (parts.length < 3 || parts[0] !== 'did') {
```
devinivy, line 298: I think these bad boys are always supposed to have three parts.
```suggestion
  if (parts.length !== 3 || parts[0] !== 'did') {
```

I also think they are all meant to be present, e.g. this isn't a did:  `did::`.  So could also have a check like:
```js
const [tag, method, id, ...others] = value.split(':')
if (tag !== 'did' || !method || !id || others.length) {
```

devinivy, line 298: We could add a check here for a missing method to avoid non-dids like `did::`.
```suggestion
  if (parts.length < 3 || parts[0] !== 'did' || !parts[1]) {
```

packages/pds/tests/views/author-feed.test.ts
```diff
@@ -239,7 +239,7 @@ describe('pds author feed views', () => {
     await agent.api.com.atproto.admin.reverseModerationAction(
       {
         id: action.id,
-        createdBy: 'X',
+        createdBy: 'did:exampl:admin',
```
devinivy, line 242:
```suggestion
        createdBy: 'did:example:admin',
```

packages/lexicon/src/validators/primitives.ts
```diff
@@ -248,39 +246,76 @@ export function string(
...
+      error: new ValidationError(`${path} must be a did`),
     }
   }
+  return { success: true, value }
+}
 
+export function cidFormat(path: string, value: string): ValidationResult {
+  try {
+    const cid = CID.parse(value)
+    if (cid.toString() !== value) {
```
devinivy, line 310: I don't think this check will cause any false positives, but I think it may not be possible for it to ever fail here.

devinivy (approved):
Super thorough— good stuff!

lexicons/com/atproto/sync/subscribeAllRepos.json
dholms: yup yup that's right 👍

packages/lexicon/src/validators/primitives.ts
dholms: yup good point 👍 

i'll be merging in bryans identifier validators soon: https://github.com/bluesky-social/atproto/pull/576

packages/lexicon/src/validators/primitives.ts
dholms: yup yup 👍

packages/lexicon/src/validators/primitives.ts
dholms: yeah fair point. i'll take it out 🤷

dholms:
I'm +1 on @bnewbold's feedback

Formats are cheap & I think we should feel comfortable adding them. 

How about the following list?
- datetime
- at-uri
- cid
- did
- handle
- nsid
- did-or-handle
- subject (did or at-uri)

dholms:
Gonna get this in & we can keep the discussion going 👍

lexicons/com/atproto/report/subject.json
devinivy: Hahaha love that.  That's my b!  I must have missed this in a cleanup.

Issue #579, opened 2023-02-22
Remove declarations and declarationCids


pfrazee:
- Non-breaking changes
  - Remove `app.bsky.system.*`
  - Stop producing the declaration record during account creation
- Breaking changes
  - Update the `subject of `app.bsky.graph.follow` to be a string with the `did` format

dholms:
fixed https://github.com/bluesky-social/atproto/pull/658

PR #654, opened 2023-03-13
Bsky app view view methods, actor indexing

devinivy:
Piggy-backing on #652, this completes the initial round of core functionality of the bsky app view.  All tests are passing, ~~though still need to get postgres running in CI in order to demonstrate that.~~

 - Nix `ipld_block` table, repurpose `record` table for storing record contents.
   - Storing in json so that it's essentially ready to be served, no more conversion to/from cbor in the views.
 - Nix `did_handle`, `repo_root` tables, replace with single `actor` table.
 - Index actors.  Handles are resolved once (confirming bidirectional link) when the repo is first encountered.
   - Will need to take into account handle updates.
 - Allow skipping over ops for records that the app view doesn't need to index.
 - Stop indexing declarations, assertions, confirmations and remove related tables.
 - Looking ahead to #579, I've started just stubbing-out the declarations served in the views.
 - Sweep all view tests, updating so that indexing fully processes as needed.
   - Remove some pds-specific tests that appeared mixed-in to the view tests.
 - Remove unused notification types `'assertion'` and `'invite'`.
 - Misc tidy and fixes.


packages/bsky/src/services/indexing/index.ts
```diff
@@ -49,28 +45,59 @@ export class IndexingService {
...
+      .selectAll()
+      .executeTakeFirst()
+    if (actor) {
+      return // @TODO deal with handle updates
+    }
+    const { pds, handle } = await this.didResolver.resolveAtpData(did)
+    const handleToDid = await resolveExternalHandle(pds, handle)
+    if (did !== handleToDid) {
+      return // No bidirectional link between did and handle
+    }
```
devinivy, line 79: @dholms wanted to double-check, does this seem like the right approach for indexing the actor's handle?

packages/bsky/src/services/indexing/index.ts
```diff
@@ -111,22 +137,34 @@ export class IndexingService {
...
+    if (err instanceof handleLib.NoHandleRecordError) {
+      // no worries it's just not found
+    } else {
+      subLogger.error({ err, handle }, 'could not resolve dns handle')
+    }
+  }
+  try {
+    const agent = new ApiAgent({ service: pds }) // @TODO all good to use pds rather than the handle itself?
+    const res = await agent.api.com.atproto.handle.resolve({ handle })
+    return res.data.did
```
devinivy, line 166: @dholms also was hoping for an extra eye on this.  Since I happen to have the user's pds from their did doc, I wondered if this may be more reliable than using their handle as the `service` without sacrificing any correctness.

dholms: ah no unfortunately not, you should use their handle.

Reason being: I can claim the handle `dholms.xyz` _either_ using a dns record _or_ hosting a server with the handle.resolve xrpc method there

dholms: Altho their server _could_ probably still resolve that request for you, I think it's better to go to the source

packages/bsky/src/services/indexing/index.ts
dholms: yup yup that seems good 👍 

I guess just some worries about retry logic there with DNS propagation? 

We oughta get in those authoritative name server lookups soon 😅

packages/bsky/src/db/migrations/20230309T045948368Z-init.ts
```diff
@@ -232,68 +210,38 @@ export async function up(db: Kysely<unknown>): Promise<void> {
...
-
-  // @TODO(bsky) ipldBlock
-  await db.schema
-    .createTable('ipld_block')
-    .addColumn('cid', 'varchar', (col) => col.notNull())
-    .addColumn('creator', 'varchar', (col) => col.notNull())
-    .addColumn('size', 'integer', (col) => col.notNull())
-    .addColumn('content', sql`bytea`, (col) => col.notNull())
-    .addPrimaryKeyConstraint('ipld_block_with_creator_pkey', ['creator', 'cid'])
+    .addColumn('takedownId', 'integer') // @TODO intentionally missing index
```
dholms, line 236: why intentionally missing index?

packages/bsky/src/db/migrations/20230309T045948368Z-init.ts
```diff
@@ -232,68 +210,38 @@ export async function up(db: Kysely<unknown>): Promise<void> {
...
+  //record
   await db.schema
     .createTable('record')
     .addColumn('uri', 'varchar', (col) => col.primaryKey())
     .addColumn('cid', 'varchar', (col) => col.notNull())
     .addColumn('did', 'varchar', (col) => col.notNull())
-    .addColumn('collection', 'varchar', (col) => col.notNull())
-    .addColumn('rkey', 'varchar', (col) => col.notNull())
-    .addColumn('takedownId', 'integer') // @TODO intentionally missing index
+    .addColumn('json', 'text', (col) => col.notNull())
```
dholms, line 234: aw yiss

packages/bsky/src/services/indexing/processor.ts
```diff
@@ -57,6 +56,17 @@ export class RecordProcessor<T, S> {
     if (!this.matchesSchema(obj)) {
       throw new Error(`Record does not match schema: ${this.params.lexId}`)
     }
+    await this.db
+      .insertInto('record')
+      .values({
+        uri: uri.toString(),
+        cid: cid.toString(),
+        did: uri.host,
+        json: JSON.stringify(obj),
```
dholms, line 65: need to make sure we use the lex serializer on these 

for cids/bytes/blob refs

dholms: maybe throw a TODO on or something cause this could end up causing unexpected bugs

dholms: goes every anywhere that records get serialized to/from json

packages/bsky/tests/_util.ts
```diff
@@ -129,10 +134,26 @@ export const runTestServer = async (
   const bskyServer = await bsky.start()
   const bskyPort = (bskyServer.address() as AddressInfo).port
 
+  // Map pds public url to its local url
+  ApiAgent.configure({
```
dholms, line 138: oh nice gotta keep this in my back pocket

dholms: some dev tools around this could be quite nice tbh

dholms (approved):
looks good ✨ 

left a couple comments but nothing too big

packages/bsky/src/db/migrations/20230309T045948368Z-init.ts
devinivy: Mostly because I wasn't sure how we were going to land on this, but wanted to keep the column around since it's already all wired-up in the views.  I have a better sense of it as of this week— I will add the indexes back.

devinivy: Just looked at this— I forgot!  I wrote index but I probably meant foreign key, as the `moderation_report` table doesn't exist yet (I didn't know when writing this code whether it would ever exist).  I will update the comment.

Issue #591, opened 2023-02-22
Replace all user parameters with actor or repo


pfrazee:
- If in the com.atproto namespace, we should use `repo`
- If in the app.bsky, we should use `actor`
- Reasoning:
    - We’ve chosen actor & repo very specifically for each of these namespaces, and user gets thrown in randomly right now. It’s important to be consistent on this.

dholms:
fixed https://github.com/bluesky-social/atproto/pull/658

Issue #548, opened 2023-02-14
`actor` param on some methods, `user` on others


dholms:
We use the `actor` param for some methods such as `getProfile`, but with others we call the param `user` - such as `getFollows` & `getFollowers`. We should standardize this on all methods

myConsciousness:
@dholms ,

I found this in JSON returned by API.

For example, `app.bsky.actor.search` returns JSON as follows:

```json
{
  "cursor": ".7777778::tezos.bsky.social",
  "users": [
    {
      "did": "did:plc:ji4f2ypncqko5cm62pmyxh67",
      "declaration": {
        "actorType": "app.bsky.system.actorUser",
        "cid": "bafyreid27zk7lbis4zw5fz4podbvbs4fc5ivwji3dmrwa6zggnj4bnd57u"
      },
      "handle": "tester.bsky.social",
      "displayName": "tester",
      "description": "testing account for apps",
      "indexedAt": "2023-03-07T01:14:50.384Z",
      "viewer": {
        "muted": false,
        "following": "at://did:plc:iijrtk7ocored6zuziwmqq3c/app.bsky.graph.follow/3jqdex7dcrt2q"
      }
    },
}
```

But other endpoint returns as follows:

```json
{
  "cursor": "235::did:plc:5trbo6ijzaizkszgcpvx7gnf",
  "actors": [
    {
      "did": "did:plc:fjsmdevv3mmzc3dpd36u5yxc",
      "declaration": {
        "actorType": "app.bsky.system.actorUser",
        "cid": "bafyreid27zk7lbis4zw5fz4podbvbs4fc5ivwji3dmrwa6zggnj4bnd57u"
      },
      "handle": "hnbot.bsky.social",
      "viewer": { "muted": false }
    },
}
```

dholms:
Nope `User` & `Actor` are just unstandardrized values. You can see the standardization I'm doing here: https://github.com/bluesky-social/atproto/pull/655

The former returns a more detailed view than the latter

dholms:
fixed https://github.com/bluesky-social/atproto/pull/658

PR #655, opened 2023-03-13
Lexicon: Replace all user parameters with actor or repo

dholms:
Remove vague "user" references.

If in the `com.atproto` namespace, we use "repo"
If in the `app.bsky`, we use "actor"

Closes https://github.com/bluesky-social/atproto/issues/591
Closes https://github.com/bluesky-social/atproto/issues/548


lexicons/app/bsky/feed/getAuthorFeed.json
```diff
@@ -4,12 +4,12 @@
   "defs": {
     "main": {
       "type": "query",
-      "description": "A view of a user's feed.",
+      "description": "A view of an actor's feed.",
       "parameters": {
         "type": "params",
-        "required": ["author"],
+        "required": ["actor"],
```
dholms, line 10: I took a bit of liberty here & changed it to `actor` even tho it wasn't ambiguous like `user`. Let me know if you disagree

devinivy:
I just pushed up that change I mentioned above— we can always revert if we'd rather not include it 👍

dholms:
Nope I like it! & now that we have formats, it's totally clear what the value is supposed to be by looking at the schema 👍

devinivy (approved):

PR #657, opened 2023-03-14
PDS refresh token grace period

devinivy:
This implements a grace period during which the same refresh token may be used multiple times before being fully rotated.  When the refresh token is used its expiration time is shortened to the grace period.

Using the same refresh token multiple times during its grace period also produces new tokens with the same `jti`, meaning that during the grace period you essentially can only get one new refresh token, though you can ask for it multiple times.  This is intended to mitigate an attacker from minting their own "chain" of refresh tokens: attackers would still have to share credentials with the legitimate user and therefore possibly lose access if that user deletes their session.

I've also added a cleanup routine that runs when refresh tokens are used, so that expired tokens for that user are no longer stored.  Previously we only deleted refresh tokens that were rotated, but now we may also remove expired refresh tokens that may never have been rotated.


dholms (approved):
yup yup super clean solution & should save us some real headaches🙌

PR #659, opened 2023-03-14
Lexicon: remove actor declaration and assertion/confirmations

### LINKED ISSUE #579: Remove declarations and declarationCids

### LINKED ISSUE #586: Remove assertion and confirmation
- Remove:
  - `app.bsky.graph.assertion`
  - `app.bsky.graph.confirmation`
  - `app.bsky.graph.assertMember`
  - `app.bsky.graph.assertCreator`

We think these are the right idea but they're not being used yet, so they can come back in the future

devinivy:
This removes declarations (incl. actor type and declaration cids), assertions, and confirmations.  Follow subject is updated to be a plain did rather than a did/declaration pair.

The app-view indexes are cleared of all these records, but there will be an additional PR (incoming) to migrate them out of repos as well.

One unrelated change in here is that I removed [pds/tests/migrations/duplicate-records.ts](https://github.com/bluesky-social/atproto/compare/lex/xprc-user-params...lex/remove-declaration?expand=1#diff-3f95f2d81e1adaad04c4a761967d04ea905d5fb13b447311d8e41a208deb55b8) since it was out of date, doesn't run as part of the test suite, and would have required some non-trivial changes to bring up to date.

Note that this is waterfalling into #655, we should let that land first.  Resolves #579, resolves #586.


lexicons/app/bsky/graph/follow.json
```diff
@@ -10,7 +10,7 @@
         "type": "object",
         "required": ["subject", "createdAt"],
         "properties": {
-          "subject": {"type": "ref", "ref": "app.bsky.actor.ref"},
+          "subject": {"type": "string", "format": "did"},
```
dholms, line 13: oh hell yeah 😎

dholms (approved):
big yup 👍 

just make sure to undo the migration change before you merge it in

PR #660, opened 2023-03-14
Close open file handles

dholms:
Couple spots where we weren't closing open file handles. Started popping up in tests

### DIFF
--- a/packages/common/src/fs.ts
+++ b/packages/common/src/fs.ts
@@ -1,0 +1,14 @@
+import fs from 'fs/promises'
+import { isErrnoException } from './util'
+
+export const fileExists = async (location: string): Promise<boolean> => {
+  try {
+    await fs.access(location, fs.constants.F_OK)
+    return true
+  } catch (err) {
+    if (isErrnoException(err) && err.code === 'ENOENT') {
+      return false
+    }
+    throw err
+  }
+}
--- a/packages/pds/src/image/server.ts
+++ b/packages/pds/src/image/server.ts
@@ -133,21 +133,18 @@ export class BlobDiskCache implements BlobCache {
   }
 
   async get(fileId: string) {
-    let handle: fs.FileHandle | undefined = undefined
+    const filePath = path.join(this.tempDir, fileId)
     try {
-      handle = await fs.open(path.join(this.tempDir, fileId), 'r')
-      const { size } = await handle.stat()
+      const { size } = await fs.stat(filePath)
       if (size === 0) {
         throw new BlobNotFoundError()
       }
-      return Object.assign(handle.createReadStream(), { size })
+      return Object.assign(fsSync.createReadStream(filePath), { size })
     } catch (err) {
       if (isErrnoException(err) && err.code === 'ENOENT') {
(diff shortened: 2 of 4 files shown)


packages/pds/src/image/server.ts
```diff
@@ -145,6 +146,8 @@ export class BlobDiskCache implements BlobCache {
         throw new BlobNotFoundError()
       }
       throw err
+    } finally {
+      await handle?.close()
```
devinivy, line 150: Naively I would expect this to close the file descriptor in such a way that it prevents consumers from reading the returned stream—do you know why that isn't the case?  Or if that does turn out to be case, maybe this just needs to go in the `catch {}` path?

packages/pds/src/storage/disk-blobstore.ts
```diff
@@ -112,9 +112,13 @@ export class DiskBlobStore implements BlobStore {
...
+    if (!exists) {
+      throw new BlobNotFoundError()
+    }
     try {
-      const handle = await fs.open(this.getStoredPath(cid), 'r')
-      return handle.createReadStream()
+      return fsSync.createReadStream(path)
     } catch (err) {
       if (isErrnoException(err) && err.code === 'ENOENT') {
         throw new BlobNotFoundError()
```
devinivy, line 124: This error can't really occur here anymore, I don't think—I would expect it to instead be emitted on the returned stream.  That could technically still happen if the file is removed between the `fileExists()` check above and when the file stream is created, though I don't think that is especially likely or a major deal.

packages/pds/src/image/server.ts
dholms: yeah i think you're right i totally missed that 🙃 

Okay i just used `fsSync` to return the readstream

packages/pds/src/storage/disk-blobstore.ts
dholms: yup yup good call 👍 

fixed both of these up & changed my approach just a bit

avoiding open handles in general, just wanna create a read stream & let node handle closing it

devinivy (approved):
Yup!

packages/common/src/fs.ts
```diff
@@ -0,0 +1,14 @@
+import fs from 'fs/promises'
+import { isErrnoException } from './util'
+
+export const fileExists = async (location: string): Promise<boolean> => {
+  try {
+    await fs.access(location, fs.constants.F_OK)
```
dholms, line 6: `exists` is deprecated & `existsSync` is well... synchronous. 

`access` is lighterweight than `stat` & gives us what we want

devinivy: Ace!

packages/pds/src/storage/disk-blobstore.ts
dholms: ohhh got it you're probably right

ah well, this feels a bit more explicit to me so i'll stick with it

devinivy: I think that sounds good 💎

Issue #592, opened 2023-02-22
Replace all `before` or `after` RPC method params with `cursor`


pfrazee:
- Reasoning:
    - Cursors are actually not datetimes and it’s difficult as a client author to know which you should use
    - If ranges are needed, they should be re-added in the future

dholms:
fixed https://github.com/bluesky-social/atproto/pull/658

PR #661, opened 2023-03-14
Lex: standardize on `cursor`

dholms:
Standardizing on `cursor` instead of `before`

In the case of rkeys for list records, we use `rkeyStart` and `rkeyEnd`

Closes https://github.com/bluesky-social/atproto/issues/592


lexicons/com/atproto/repo/listRecords.json
```diff
@@ -12,8 +12,8 @@
           "repo": {"type": "string", "description": "The handle or DID of the repo."},
           "collection": {"type": "string", "description": "The NSID of the record type."},
           "limit": {"type": "integer", "minimum": 1, "maximum": 100, "default": 50, "description": "The number of records to return."},
-          "before": {"type": "string", "description": "A TID to filter the range of records returned."},
-          "after": {"type": "string", "description": "A TID to filter the range of records returned."},
+          "rkeyStart": {"type": "string", "description": "The lowest sort-ordered rkey to start from (exclusive)"},
+          "rkeyEnd": {"type": "string", "description": "The highest sort-ordered rkey to stop at (exclusive)"},
```
devinivy, line 16: I was a little surprised that these are both exclusive, just confirming that's correct! 👍

devinivy (approved):
💪 perfect

lexicons/com/atproto/repo/listRecords.json
dholms: They were both exclusive in our prev code fwiw

I went back and forth on this & toyed with making rkeyEnd inclusive. But pagination works both directions on this & my thinking is that if you know the rkey already, then you probably know the record as well.

Maybe what it should be is: make the start exclusive & the end inclusive in normal order & the opposite in reverse order? but the semantics are getting a bit complicated then 🙃

dholms: I'm down to simplify this route in general tbh. I'm not sure it's actually used anywhere & there are some tricky semantics around cursors/ordering in it

devinivy: Ahh yeah, that makes sense 👍 I don't think we need to change anything (particularly here/now), just wanted to ensure the description was accurate.

PR #662, opened 2023-03-14
List blobs method

dholms:
Added in an XRPC method for listing blob cids for a given commit range


devinivy (approved):
Solid!  We just gotta remember to hit the new method with a `format: 'cid'` when the lex refactor lands 👍

PR #664, opened 2023-03-15
Lexicon: declaration, assertion/confirmation, follow app migration

devinivy:
Two app migrations contained here:
1. Remove declaration, assertion, and confirmation records.
2. Update follow records so that their `subject` is a did rather than an actor ref.  Invalid follow records are deleted.

In both we use the repo service, which takes care of updating repositories plus the generic `record` index.  If any app-view related changes are needed, those are handled at the end of the migrations— that would include updates to `duplicate_record` and `follow` tables.  A no-op `messageDispatcher` is used rather than a real one in order to avoid app-view indexing logic from kicking in.

### DIFF
--- a/packages/pds/src/app-migrations/remove-declarations.ts
+++ b/packages/pds/src/app-migrations/remove-declarations.ts
@@ -41,7 +41,7 @@ async function main(tx: Database, ctx: AppContext) {
       'app.bsky.graph.assertion',
       'app.bsky.graph.confirmation',
     ])
-    .select(['did', 'collection', 'rkey'])
+    .select(['record.did as did', 'collection', 'rkey'])
     .execute()
 
   const deletionsByDid = recordsToDelete.reduce((collect, record) => {
@@ -49,16 +49,16 @@ async function main(tx: Database, ctx: AppContext) {
     collect[record.did].push(prepareDelete(record))
     return collect
   }, {} as Record<string, PreparedDelete[]>)
-
   const entries = Object.entries(deletionsByDid)
-  const chunks = chunkArray(entries, Math.ceil(entries.length / 50))
 
   console.log(
     SHORT_NAME,
     `${recordsToDelete.length} deletions across ${entries.length} dids`,
   )
+
   let didsComplete = 0
   let deletionsComplete = 0
+  const chunks = chunkArray(entries, Math.ceil(entries.length / 50))
 
   await Promise.all(
     chunks.map(async (chunk) => {
--- a/packages/pds/src/app-migrations/remove-declarations.ts
+++ b/packages/pds/src/app-migrations/remove-declarations.ts
@@ -1,7 +1,9 @@
 import { chunkArray } from '@atproto/common'
+import assert from 'assert'
 import AppContext from '../context'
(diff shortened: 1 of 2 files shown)


packages/pds/src/app-migrations/update-follow-subjects.ts
```diff
@@ -0,0 +1,169 @@
...
+              assertValidRecord(record)
+              return prepareUpdate({
+                did,
+                collection: ids.AppBskyGraphFollow,
+                rkey: follow.rkey,
+                record,
+                validate: false,
+              })
+            } catch {
+              const del = prepareDelete({
```
dholms, line 76: just curious, how many hit this in trial runs?

dholms (approved):
These both looks really great 🙌 

Would you be up for adding some quick sanity checks at the end, that throw if they fail? like "count of follows = count before migration - count deleted" & "no declaration/assertion/confirmation left in records?

packages/pds/src/app-migrations/remove-declarations.ts
```diff
@@ -0,0 +1,78 @@
...
+    noopDispatcher,
+    ctx.blobstore,
+  )
+
+  // For each user remove declaration, assertion, confirmation records.
+
+  // Should be 4-5k records
+  const recordsToDelete = await tx.db
+    .selectFrom('record')
+    .innerJoin('repo_root', 'repo_root.did', 'record.did') // Ignore any records not in a repo
```
dholms, line 38: did you encounter any of these in your tests?

devinivy: There were 7, I think.  I looked at them all and they were indeed malformed in some way (either a handle rather than a did, or empty did).

packages/pds/src/app-migrations/update-follow-subjects.ts
devinivy: There were 7, I think. I looked at them all and they were indeed malformed in some way (either a handle rather than a did, or empty did).

packages/pds/src/app-migrations/remove-declarations.ts
devinivy: I didn't check actually, but I could do a run without this and find out.

devinivy: Okay this was overly paranoid, it just doesn't happen on any environment 😆 Will remove.

packages/pds/src/app-migrations/update-follow-subjects.ts
```diff
@@ -0,0 +1,169 @@
...
+                collection: ids.AppBskyGraphFollow,
+                rkey: follow.rkey,
+              })
+              updatesTurnedDeletes.push(del.uri.toString())
+              return del
+            }
+          }),
+        )
+
+        await repoTx.processWrites(did, updates, now)
```
dholms, line 87: ah probably wanna ensure updates.length > 0

devinivy:
@dholms should be good for another look!

dholms (approved):
looks great 👍

PR #668, opened 2023-03-15
Lexicon: votes to likes app migration

devinivy:
Contains the app migration to turn votes into likes.

Using the repo service here similar to #664, which takes care of updating repositories plus the generic record index. App-view related changes are handled at the end of the migrations— that would include updates to `duplicate_record` and `like` tables, in particular updating uris and cids.  A no-op `messageDispatcher` is used rather than a real one in order to avoid app-view indexing logic from kicking in.  The rkeys of the vote records are preserved into the like records.

### DIFF
--- a/packages/pds/src/app-migrations/votes-to-likes.ts
+++ b/packages/pds/src/app-migrations/votes-to-likes.ts
@@ -23,7 +23,7 @@ async function main(tx: Database, ctx: AppContext) {
   const now = new Date().toISOString()
 
   // The message dispatcher usually ensures that the app view indexes these updates,
-  // but that has been taken care of via a db migration to remove the indexes entirely.
+  // but that has been taken care of with a db migration, plus the end of this migration.
   const noopDispatcher = new MessageDispatcher()
   noopDispatcher.destroy()
 
--- a/packages/pds/src/app-migrations/votes-to-likes.ts
+++ b/packages/pds/src/app-migrations/votes-to-likes.ts
@@ -1,0 +1,209 @@
+import { cborBytesToRecord, chunkArray } from '@atproto/common'
+import { sql } from 'kysely'
+import AppContext from '../context'
+import Database from '../db'
+import { appMigration } from '../db/leader'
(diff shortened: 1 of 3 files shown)


packages/pds/src/app-migrations/votes-to-likes.ts
```diff
@@ -0,0 +1,209 @@
...
+    }),
+  )
+
+  console.log(
+    SHORT_NAME,
+    `${createsTurnedDeletes.length} creates-turned-deletes`,
+    createsTurnedDeletes,
+  )
+
+  if (createsTurnedDeletes.length > voteCount / 1000) {
```
dholms, line 118: is this just an arbitrary limit?

packages/pds/src/app-migrations/votes-to-likes.ts
```diff
@@ -0,0 +1,209 @@
...
+                record,
+                validate: false, // Validated above
+              }),
+            ]
+          }),
+        )
+
+        // Chunk to avoid hitting e.g. postgres param limits
+        for (const writeChunk of chunkArray(writeResults, 1000)) {
+          await repoTx.processWrites(did, writeChunk.flat(), now)
```
dholms, line 99: can't imagine it hitting this case, but maybe still worth ensuring that the chunk.length > 0

dholms (approved):
looks good! again, it might feel good to have a couple quick sanity checks at the end

packages/pds/src/app-migrations/votes-to-likes.ts
devinivy: Yeah, the idea is that if you see a non-trivial number of these that the most likely cause is that the logic to reformat the record from a vote to a like is wrong.  And if there are extremely few of them it's probably just bad data that already existed, which is fine.

devinivy:
@dholms sweet, should be good for another peek 👍

dholms (approved):
yup yup yup 👍

PR #672, opened 2023-03-16
More label lexicons

bnewbold:
New endpoints:

- `com.atproto.label.query`: lookup labels on URI patterns. intended uses are to look up all labels on a specific repo or specific record; look up all labels on a specific repo and all labels on all posts by that repo (collection-level wildcard)
- `com.atproto.label.subscribeAllLabels`: an event stream similar to `subscribeAllRepos`

And adding arrays of labels to "views" of posts and profiles.

This branch is on top of the `com.atproto.label.label` branch (PR: https://github.com/bluesky-social/atproto/pull/525).

The description on the "uriPatterns" param is sort of terse, when I mean by "inclusive" is that `at://did:plc:abc123*` would match labels on `at://did:plc:abc123` (exact match) in addition to `at://did:plc:abc123/app.bsky.feed.post/abc123` (etc). That wouldn't be the case if you did `at://did:plc:abc123/*` (with the extra slash before the `*`).

Some things i'm looking for feedback on:

- should I be using the `#view` variant of label record to pull the CID and URI (of the label itself) in? that feels like it really commits us to labels-in-repos.
- intend to eventually add an additional optional `providers` array to the query endpoint, which would only return labels from the listed DIDs (`src` field in the label). Maybe I should just add that now?
- do I need to worry about what the "cursor" value will be for pagination of results? I don't think so. Would probably just be a table primary key encoded as base64 or something like that.
- I originally had a `get` variant of `query`, but I think that `query` is flexible to both.

### DIFF
--- a/lexicons/app/bsky/actor/defs.json
+++ b/lexicons/app/bsky/actor/defs.json
@@ -18,7 +18,7 @@
         "viewer": {"type": "ref", "ref": "#viewerState"},
         "labels": {
           "type": "array",
-          "items": {"type": "ref", "ref": "com.atproto.label.label"}
+          "items": {"type": "ref", "ref": "com.atproto.label.defs#label"}
         }
       }
     },
@@ -43,7 +43,7 @@
         "viewer": {"type": "ref", "ref": "#viewerState"},
         "labels": {
           "type": "array",
-          "items": {"type": "ref", "ref": "com.atproto.label.label"}
+          "items": {"type": "ref", "ref": "com.atproto.label.defs#label"}
         }
       }
     },
@@ -72,7 +72,7 @@
         "viewer": {"type": "ref", "ref": "#viewerState"},
         "labels": {
           "type": "array",
-          "items": {"type": "ref", "ref": "com.atproto.label.label"}
+          "items": {"type": "ref", "ref": "com.atproto.label.defs#label"}
         }
       }
     },
--- a/lexicons/app/bsky/actor/defs.json
+++ b/lexicons/app/bsky/actor/defs.json
@@ -15,7 +15,11 @@
           "maxLength": 640
         },
         "avatar": { "type": "string" },
-        "viewer": {"type": "ref", "ref": "#viewerState"}
+        "viewer": {"type": "ref", "ref": "#viewerState"},
+        "labels": {
+          "type": "array",
+          "items": {"type": "ref", "ref": "com.atproto.label.label"}
+        }
       }
     },
     "profileView": {
@@ -36,7 +40,11 @@
         },
         "avatar": { "type": "string" },
         "indexedAt": {"type": "string", "format": "datetime"},
-        "viewer": {"type": "ref", "ref": "#viewerState"}
+        "viewer": {"type": "ref", "ref": "#viewerState"},
+        "labels": {
+          "type": "array",
+          "items": {"type": "ref", "ref": "com.atproto.label.label"}
+        }
       }
     },
     "profileViewDetailed": {
@@ -61,7 +69,11 @@
         "followsCount": {"type": "integer"},
         "postsCount": {"type": "integer"},
         "indexedAt": {"type": "string", "format": "datetime"},
-        "viewer": {"type": "ref", "ref": "#viewerState"}
+        "viewer": {"type": "ref", "ref": "#viewerState"},
+        "labels": {
+          "type": "array",
+          "items": {"type": "ref", "ref": "com.atproto.label.label"}
+        }
       }
     },
     "viewerState": {
--- a/lexicons/com/atproto/label/query.json
+++ b/lexicons/com/atproto/label/query.json
@@ -1,0 +1,42 @@
+{
+  "lexicon": 1,
+  "id": "com.atproto.label.query",
+  "defs": {
+    "main": {
+      "type": "query",
+      "description": "Find labels relevant to the provided URI patterns.",
+      "parameters": {
+        "type": "params",
+        "required": ["uriPatterns"],
+        "properties": {
+          "uriPatterns": {
+              "type": "array",
+              "items": {"type": "string"},
+              "description": "List of AT URI patterns to match. Each may be a prefix (ending with '*'; will match inclusive of the string leading to '*'), or a full URI"
+          },
+          "sources": {
+              "type": "array",
+              "items": {"type": "string"},
+              "description": "Optional list of label sources (DIDs) to filter on"
+          },
+          "limit": {"type": "integer", "minimum": 1, "maximum": 100, "default": 20},
+          "cursor": {"type": "string"}
+        }
+      },
+      "output": {
+        "encoding": "application/json",
+        "schema": {
+          "type": "object",
+          "required": ["labels"],
+          "properties": {
+            "cursor": {"type": "string"},
+            "labels": {
+              "type": "array",
+              "items": {"type": "ref", "ref": "com.atproto.label.label"}
+            }
+          }
+        }
+      }
+    }
+  }
+}
--- a/lexicons/com/atproto/label/query.json
+++ b/lexicons/com/atproto/label/query.json
@@ -16,7 +16,7 @@
           },
           "sources": {
               "type": "array",
-              "items": {"type": "string"},
+              "items": {"type": "string", "format": "did"},
               "description": "Optional list of label sources (DIDs) to filter on"
           },
           "limit": {"type": "integer", "minimum": 1, "maximum": 100, "default": 20},
--- a/lexicons/com/atproto/label/subscribeLabels.json
+++ b/lexicons/com/atproto/label/subscribeLabels.json
@@ -16,24 +16,40 @@
       },
       "message": {
         "schema": {
-          "type": "object",
-          "required": ["seq", "labels"],
-          "nullable": ["prev"],
-          "properties": {
-            "seq": {"type": "integer"},
-            "labels": {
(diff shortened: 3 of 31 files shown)


lexicons/com/atproto/label/query.json
```diff
@@ -0,0 +1,36 @@
...
+    "main": {
+      "type": "query",
+      "description": "Find labels relevant to the provided URI patterns.",
+      "parameters": {
+        "type": "params",
+        "properties": {
+          "uriPatterns": {
+              "type": "array",
+              "items": {"type": "string"},
+              "description": "List of AT URI patterns to match. Each may be a prefix (ending with '*'; will match inclusive of the string leaing to '*'), or a full URI"
```
dholms, line 14: typo on "leaing"

lexicons/com/atproto/label/subscribeAllLabels.json
```diff
@@ -0,0 +1,39 @@
+{
+  "lexicon": 1,
+  "id": "com.atproto.label.subscribeAllLabels",
```
dholms, line 3: we changed the repos one to `subscribeRepos` & with the idea that we'll eventually just add ways to filter your subscription with a default of firehose

probably makes sense here as well

lexicons/com/atproto/label/subscribeAllLabels.json
```diff
@@ -0,0 +1,39 @@
...
+          "cursor": {
+            "type": "integer",
+            "description": "The last known event to backfill from."
+          }
+        }
+      },
+      "message": {
+        "schema": {
+          "type": "object",
+          "required": ["seq", "event", "repo", "commit", "prev", "blocks", "ops", "blobs", "time"],
```
dholms, line 20: requireds lurking around

lexicons/com/atproto/label/subscribeAllLabels.json
```diff
@@ -0,0 +1,39 @@
...
+          "nullable": ["prev"],
+          "properties": {
+            "seq": {"type": "integer"},
```
dholms, line 23: if a service offers both repo & label subscriptions, the idea is they'd be separate `seq`s right?

dholms (approved):
This all seems good to me 👍 

I wish we had a better syntax than `at://did:plc:abc123*`. feels gross to me but I guess it's right 🙃 

> should I be using the #view variant of label record to pull the CID and URI (of the label itself) in? that feels like it really commits us to labels-in-repos.

Yknow im not sure how much we'll end up using `label#view`. I might just ditch it until we need it. I would also like to not lock us into labels-in-repos right now.

> intend to eventually add an additional optional providers array to the query endpoint, which would only return labels from the listed DIDs (src field in the label). Maybe I should just add that now?

As long as it's optional, we can add it in the future 👍  If you feel pretty confident in the semantics tho, you can go ahead and throw it on.

> do I need to worry about what the "cursor" value will be for pagination of results?

Nope cursors are meant to be opaque & at the discretion of the service 👍 

> I originally had a get variant of query, but I think that query is flexible to both.

Seems find 👍 We can add if we find a need for it

lexicons/com/atproto/label/subscribeAllLabels.json
bnewbold: That is my understanding. the scope of sequences (distinct set of `seq` numbers) would be (host, streamNsid).

I think we want to keep the door open to merging subscription feeds into a unified feed, and I don't think this blocks that, assuming that the merged feed would have a different endpoint. Also seq numbers can in theory have gaps between them, so a unified seq number *could* be used, and then endpoints just filter to the specific type at that endpoint. Haven't thought this through much, just thinking outloud.

lexicons/com/atproto/label/subscribeAllLabels.json
bnewbold: I made it `subscribeLabels`, though now that I think of it could even just be `subscribe`.

bnewbold:
Pushed an update with review and conversation changes:

- removed `#view` from `com.atproto.label.label`
- renamed subscribe endpoint to `com.atproto.label.subscribeLabels`
- made `uriPatterns` required for query
- added `sources` filter on query (optional array of DIDs)
- fixed typo and removed bogus required from subscribe

My current disposition is to: wait until lexicon refactors land and get deployed to prod; rebase on that new `main` and add string-type annotations here (eg, DID and ATURI fields); merge this. Can't think of a specific way that this would hold you up with the lex refactor, but that is a lot of moving pieces and don't want to trip things up. I will do codegen in golang before then to start implementing these.

I'm not sure if the `cid` field on the label should be an internal type or string-y type. It would be referencing a record version, like a strong ref.

lexicons/com/atproto/label/query.json
```diff
@@ -0,0 +1,42 @@
+{
+  "lexicon": 1,
+  "id": "com.atproto.label.query",
```
devinivy, line 3: It may be more consistent with the style post-lex refactor to use something verb-noun like `queryLabels` here.

devinivy (commented):
Left a note but this looks good!  Do kinda wish the `label.label` repetition weren't a thing, but the name itself seems right.  Only other ideas that come to mind for an alternative namespace are `com.atproto.content` or `com.atproto.labeling`.  Probably not a huge deal, though.  Nice!

bnewbold:
Squashed my earlier commits, then in separate commits (so you can view commit-by-commit if you want) did some post-lex-refactor updates:

- moved profile/feed/notification inclusion to different files/schemas. I kept these the same as before, and verified that they show up the same places as "viewer" metadata (as a heuristic), plus in notifications (which doesn't have "viewer" metadata)
- renamed to `label.queryLabels` (to match verb-noun norm)
- added string formats. kept `uri` as `format="uri"` instead of `"at-uri"` because we might stuff other target types in there (like DNS URIs for infra). open to going with `"at-uri"` if we think that referencing infra by handles will be robust enough for doing, eg, labeling of PDS instances
- bumped query batch sizes. labels are pretty small/cheap and I think returning more that for other object types is fine

Open thoughts still on my mind (after all this time!):

- does `label.label` even need to be a 'record' schema? I guess it might as well be. having it be a record will potentially result in `$type: "com.atproto.label.label"` getting included always, I think, which is unnecessary overhead
- have a plan for label negation which is to put a `~` in front of the value to indicate negation. this could instead be an optional boolean field. not sure which is better. negations would likely be intended for infra-to-infra only and not displayed by clients. it would indicate "wipe any previous label" (with the same src, uri, val)

I think i'm also going to have a small round of changes for integrating labels in to the reporting lexicons (for redsky), but that can be a separate PR after this lands.

lexicons/com/atproto/label/subscribeLabels.json
```diff
@@ -0,0 +1,39 @@
...
+          "properties": {
+            "seq": {"type": "integer"},
+            "labels": {
+              "type": "array",
+              "items": { "type": "ref", "ref": "com.atproto.label.label" }
+            }
+          }
+        }
+      },
+      "infos": [
```
dholms: ah we've ended up deprecated `infos` as a lexicon/wire-protocol carveout & now they're just in the message schema itself

See subscribeRepos for an example: https://github.com/bluesky-social/atproto/blob/main/lexicons/com/atproto/sync/subscribeRepos.json

lexicons/app/bsky/actor/defs.json
```diff
@@ -15,7 +15,11 @@
           "maxLength": 640
         },
         "avatar": { "type": "string" },
-        "viewer": {"type": "ref", "ref": "#viewerState"}
+        "viewer": {"type": "ref", "ref": "#viewerState"},
+        "labels": {
+          "type": "array",
+          "items": {"type": "ref", "ref": "com.atproto.label.label"}
```
dholms, line 21: I think... this might actually be the only case where a record schema is used as a `ref` 🤔 

I'm not sure if we've exactly specified the semantics of that? ie if `$type` is required in that case.

I would think not. Similar to schema fragments, it's included in a union but not a ref...

dholms (approved):
I think this all looks great 👍 

Just the question I posted about a record schema being used as a ref.

lexicons/app/bsky/actor/defs.json
dholms: might want @pfrazee to take a look at this 🤔 

We include `$type` when we need to differentiate between two options, which we don't need to do here. But without `$type`, the record will hash to a different CID than expected

dholms: (and i continue leaning towards making labels there own free-floating thing instead of a record in a repo 😉 )

bnewbold:
Updated to add `createLabelVals` (and `negateLabelVals`) on moderation actions.

The idea is that this is the direct and simple way to create labels as a result of a moderation action, either in isolation or in response to a report (via separate "resolves report" endpoint). By using moderation actions, instead of directly inserting a label, we leave an internal paper trail for manual/admin decisions.

The intended behavior is that the action creation would fail if a conflicting or pre-existing label already exists. For negation, there would need to be an existing label. Reverting an action with an associated negation would result in a label being created. The list of label vals (created and negated) would be stored along with the moderation action immutably. One potential change would be to fully hydrate labels in the action view detail; the issue with that would be the label getting updated/mutated by other actions over time (eg, after an action revert).

As noted in the commit message, by including "Vals" in the endpoint name, we leave the door open to submitting full label objects in the future, which might be preferable depending on how distributed auth shakes out.

PR #673, opened 2023-03-16
Lex: support cids & bytes

dholms:
This adds cids & bytes to lexicon.

Cids can take two shapes: 
- `type: string, format: cid` which encodes the value as a string & enforces its parseability as a CID 
- `type: cid-internal-ref` which encodes the value as CID (in ipld/dag-cbor or ipld/dag-json format). This is not expected to be used often, only for internal references within a repository, not for integrity constraints on records.

This augments xrpc-server & xrpc-client to encode/decode CIDs & bytes as their dag-json equivalent on send/receive

This also augments codegen to give them their actual types: `Uint8Array` & `CID`
Note: we're just importing CID in every file. This is a bit unfortunate, but we already have unused imports in gened files & it was gonna be a larger change to the codegen library to do it otherwise.

In support of https://github.com/bluesky-social/atproto/issues/580


packages/xrpc/src/util.ts
```diff
@@ -163,5 +164,8 @@ export function httpResponseBodyParse(
       }
     }
   }
+  if (data instanceof ArrayBuffer) {
+    return new Uint8Array(data)
```
dholms, line 168: snuck this in. we have `Uint8Arrays` in the types but they were actually ArrayBuffers

packages/common/src/ipld.ts
```diff
@@ -24,25 +43,101 @@ export const verifyCidForBytes = async (cid: CID, bytes: Uint8Array) => {
...
+  | string
+  | null
+  | bigint
+  | CID
+  | Uint8Array
+  | Array<IpldValue>
+  | { [key: string]: IpldValue }
+  | { [key: number]: IpldValue }
+
+export const jsonToIpldValue = (val: JsonValue): IpldValue => {
```
dholms, line 77: I was having a bunch of issues with the dag-json library so I just reimplemented 🙃

packages/common/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@atproto/common",
   "version": "0.1.1",
-  "main": "src/index.ts",
+  "main": "dist/index.js",
```
devinivy, line 4:
```suggestion
  "main": "src/index.ts",
```

packages/common/src/ipld.ts
```diff
@@ -24,25 +43,101 @@ export const verifyCidForBytes = async (cid: CID, bytes: Uint8Array) => {
...
+      if (Object.keys(val).length > 1) {
+        throw new Error()
+      }
+      const maybeBytes = maybeCid['bytes']
+      if (maybeBytes) {
+        if (
+          Object.keys(maybeCid).length > 1 ||
+          typeof maybeBytes !== 'string'
+        ) {
+          console.log('this err')
```
devinivy, line 92:
```suggestion
```

packages/lex-cli/src/codegen/lex-gen.ts
```diff
@@ -70,6 +73,8 @@ export function genUserType(
     case 'image':
     case 'video':
     case 'audio':
+    case 'bytes':
+    case 'cid-internal-ref':
```
devinivy, line 77: 👻

packages/lexicon/src/validators/primitives.ts
```diff
@@ -338,12 +343,70 @@ export function cidFormat(path: string, value: string): ValidationResult {
...
+  return { success: true, value }
+}
+
+export function cidInternalRef(
+  lexicons: Lexicons,
+  path: string,
+  def: LexUserType,
+  value: unknown,
+): ValidationResult {
+  if (CID.asCID(value) === null) {
```
devinivy, line 400: 👍 good call vs. the `instanceof` check.

packages/xrpc/src/util.ts
```diff
@@ -144,7 +145,7 @@ export function httpResponseBodyParse(
     if (mimeType.includes('application/json') && data?.byteLength) {
       try {
         const str = new TextDecoder().decode(data)
-        return JSON.parse(str)
+        return jsonStringToIpld(str)
```
devinivy, line 148: Might be good to get an early read on whether `@atproto/api` will still work on clients where we need it to.  Not sure how some deps in `@atproto/common` including `multiformats/cid` will play on the web, for example.

packages/xrpc/package.json
```diff
@@ -18,6 +18,7 @@
   },
   "license": "MIT",
   "dependencies": {
+    "@atproto/common": "*",
```
devinivy, line 21: Might be good to get an early read on whether `@atproto/api` will still work on clients where we need it to.  Not sure how some deps in `@atproto/common` including `multiformats/cid` will play on the web, for example.

devinivy (approved):
Yep!

packages/common/package.json
dholms: you'll be happy to know i battled with this for 30 min yesterday wondering why my changes to common were not affecting anything 🙃

packages/xrpc/package.json
dholms: `multiformats/cid` is totally fine on web, we used it a bunch at Fission 👍 

I think everything in here should still be playing nicely 🤔

PR #674, opened 2023-03-16
Lexicon: richtext facets, replacing post entities

### LINKED ISSUE #587: Add `app.bsky.richtext.facet`
- Add `app.bsky.richtext.facet`
    - Has an `index` which is `#textSlice`
    - and `value` which is an open union of `#link` and `#mention`
    - `#link` has a `uri` field
    - `#mention` has a `did` field
- Update `app.bsky.feed.post`...
    - Mark `entities` as deprecated but leave it in
    - Add `facets` as an optional field
    - This will make migration smooth — we keep supporting `entities` but start writing and using `facets`
- Update notification indexing to read the `facets` field but keep around the old `entities` field with a plan to deprecate
- Reasoning
    - Entities was created before Lexicon was fleshed out and it needs to use specific types, especially because it has indexing implications
    - Different kinds of facets have different field needs
    - As we expand on content-types, we’ll need a more robust rich text system, which entities basically was
    - The term “entities” confused people

devinivy:
This deprecates the post entities system and introduces richtext facets through `app.bsky.richtext.facet`.

Resolves #587


lexicons/app/bsky/feed/post.json
```diff
@@ -12,8 +12,13 @@
           "text": {"type": "string", "maxLength": 256},
           "entities": {
             "type": "array",
+            "description": "Deprecated: replaced by app.bsky.richtext.facet.",
```
devinivy, line 15:
```suggestion
            "description": "Deprecated: replaced by facets.",
```

lexicons/app/bsky/richtext/facet.json
```diff
@@ -0,0 +1,39 @@
...
+      "properties": {
+        "did": {"type": "string", "format": "did"}
+      }
+    },
+    "link": {
+      "type": "object",
+      "description": "A facet value for links.",
+      "required": ["uri"],
+      "properties": {
+        "uri": {"type": "string"}
```
devinivy, line 26: Minor note: in #587 I think the intention was to use a uri format here, but we decided to just have the at-uri format for now, which wouldn't apply here.

packages/pds/src/db/migrations/20230316T142245535Z-post-entities-to-facets.ts
```diff
@@ -0,0 +1,9 @@
+import { Kysely } from 'kysely'
+
+export async function up(db: Kysely<unknown>): Promise<void> {
+  await db.schema.alterTable('post_entity').renameTo('post_facet').execute()
```
dholms, line 4: I remember this table is missing a primary key. Would be a good time to get one on it

But also: I don't think we actually use this table for anything. Maybe we could drop??

dholms (approved):
looks good 👍 

Just the question of primary key or drop table. I lean towards drop unless you see a reason to keep it around 👀

packages/pds/src/db/migrations/20230316T142245535Z-post-entities-to-facets.ts
devinivy: Yeah good call, let's just drop this bad boy.

PR #675, opened 2023-03-16
Bugfix: don't emit events lost commits

dholms:
We do an _inner join_ (instead of a left join) on repo history to ensure that we're only emitting events for commits that we still have around

### DIFF
--- a/packages/pds/src/sequencer/index.ts
+++ b/packages/pds/src/sequencer/index.ts
@@ -58,23 +58,10 @@ export class Sequencer extends (EventEmitter as new () => SequencerEmitter) {
     limit?: number
   }): Promise<RepoAppendEvent[]> {
     const { earliestSeq, earliestTime, limit } = opts
+
     let seqQb = this.db.db
       .selectFrom('repo_seq')
-      .selectAll()
-      .orderBy('seq', 'asc')
-    if (earliestSeq !== undefined) {
-      seqQb = seqQb.where('seq', '>', earliestSeq)
-    }
-    if (earliestTime !== undefined) {
-      seqQb = seqQb.where('sequencedAt', '>=', earliestTime)
-    }
-    if (limit !== undefined) {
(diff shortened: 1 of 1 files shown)


packages/pds/src/sequencer/index.ts
```diff
@@ -87,10 +74,26 @@ export class Sequencer extends (EventEmitter as new () => SequencerEmitter) {
...
+    if (events.length < 1) {
+      return []
+    }
+
+    const seqs = events.map((evt) => evt.seq)
 
     const getBlocks = this.db.db
-      .selectFrom(seqQb.as('repo_seq'))
+      .selectFrom('repo_seq')
+      .where('repo_seq.seq', 'in', seqs)
```
devinivy, line 96: Only concern here is what happens if `seqs` gets large, e.g. hitting postgres param limits.  Not sure how likely that is, though.  Either way, I'm interested to hear about the impetus for switching away from using a subquery here.

devinivy (approved):
One small note/question, but this is looking good 👍

packages/pds/src/sequencer/index.ts
dholms: We never call for more than 50 so it isn't a problem pragmatically. but of course, that's not encapsulated in this function at all
I'm down to chunk the inputs on these even if we don't expect that code to be hit yet

The impetus for switching away is basically that the query is more expensive now because it has to join on `repo_commit_history` in order to get an accurate count. And we're doing that join 4 times instead of once 🤔 

I'm not opposed to sticking it in the QB, just seemed like more complex queries to me. But maybe the work of sending seqs in & out of PG is worse?

devinivy: I think if that count is always  <= 50 then this looks great. Maybe we can just add a note to indicate why it's safe.

dholms: added note

Issue #581, opened 2023-02-22
Updates to blob behaviors: `$type`, cid encoding, etc


pfrazee:
- Update all blob usages (values of type blob, image, video, audio) to include a `$type` field with the relevant type included
  - Enforce this during write validation but not during reads
- When translating blobs to CBOR, encode the `cid` field to the link tag when possible

dholms:
fixed https://github.com/bluesky-social/atproto/pull/658

PR #679, opened 2023-03-16
Lex: blob ref types

dholms:
Adds a new structure to encoding blobs into records:
```ts
{
  $type: 'blob',
  ref: CID,
  mimeType: string
}
```

Note: I chose to _only_ have $type: 'blob'. I don't see a benefit to having $type: 'image', etc because that information is available via the mimeType. The purpose of different lexicon types was to be able to impose media-type specific constraints (max height for instance).

Adds a `BlobRef` type to lexicon library. When records are parsed both the old & the new representation of blobs are parsed into this type.

Also a `LexValue` type which is all valid JSON values + IPLD values (CIDs/bytes) + BlobRefs. In the future we can extend this type to parse out formatted strings as well: At-uris, handles, nsids, etc if we like.

This type & the lex encoders/decoders are consumed by the repo as well. Records within the Repo are now typed `Record<string, LexValue>` as opposed to `Record<string, unknown>`.

Closes https://github.com/bluesky-social/atproto/issues/581


packages/lexicon/src/serialize.ts
```diff
@@ -0,0 +1,111 @@
...
+  | { [key: string]: LexValue }
+  | { [key: number]: LexValue }
+
+export type RepoRecord = Record<string, LexValue>
+
+export const lexValueToIpld = (val: LexValue): IpldValue => {
+  // convert blobs
+  if (val instanceof BlobRef) {
+    return val.original
+  }
```
devinivy, line 28: If we don't want to reference anything lexicon-related during serialization, we can respect `toJSON()` the same way `JSON.stringify()` itself does.  Whenever you come across an object with a `toJSON()` method during serialization, you'd replace the object with the result of `toJSON()`.  Even if it's only in the serialization path and not the deserialization path, it at least gives you an out so that repo code doesn't need to know about `BlobRef`.  `BlobRef` would just implement `toJSON()` 👍

dholms: ah nice we probably _do_ want a `toJSON` method on this. But I think in this case, we don't want to actually return JSON, we want to return IPLD (which includes bytes & cids)

packages/lexicon/src/serialize.ts
```diff
@@ -0,0 +1,111 @@
...
+  schema,
+} from '@atproto/common'
+import { BlobRef, jsonBlobRef } from './blob-refs'
+
+export type LexValue =
+  | IpldValue
+  | BlobRef
+  | Array<LexValue>
+  | { [key: string]: LexValue }
+  | { [key: number]: LexValue }
```
devinivy, line 20: I'm trying to think of how this type is possible since keys are generally always strings.  But realize I may be / probably am missing something!

packages/pds/src/app-view/services/indexing/plugins/post.ts
```diff
@@ -93,7 +93,7 @@ const insertFn = async (
       uri: external.uri,
       title: external.title,
       description: external.description,
-      thumbCid: external.thumb?.cid || null,
+      thumbCid: external.thumb?.ref?.toString() || null,
```
devinivy, line 96: I thiiink that if  `thumb` exists then it's a `BlobRef` and you're guaranteed to have a `ref`.

packages/pds/src/app-view/services/indexing/plugins/profile.ts
```diff
@@ -27,8 +27,8 @@ const insertFn = async (
       creator: uri.host,
       displayName: obj.displayName,
       description: obj.description,
-      avatarCid: obj.avatar?.cid,
-      bannerCid: obj.banner?.cid,
+      avatarCid: obj.avatar?.ref?.toString(),
+      bannerCid: obj.banner?.ref?.toString(),
```
devinivy, line 31:
```suggestion
      avatarCid: obj.avatar?.ref.toString(),
      bannerCid: obj.banner?.ref.toString(),
```

packages/pds/tests/views/__snapshots__/author-feed.test.ts.snap
```diff
@@ -71,8 +71,11 @@ Array [
               Object {
                 "alt": "tests/image/fixtures/key-landscape-small.jpg",
                 "image": Object {
-                  "cid": "cids(3)",
+                  "$type": "blob",
                   "mimeType": "image/jpeg",
+                  "ref": Object {
+                    "/": "cids(3)",
+                  },
```
devinivy, line 78: Since in the record this is an `'image'` lex type, should we have `"$type": "image"` here?  Or is the story there that those lex types (also e.g. `'video'`) are just for validation on write?

packages/repo/src/parse.ts
```diff
@@ -1,21 +1,34 @@
...
+  blocks: BlockMap,
+  cid: CID,
+): Promise<{ record: RepoRecord; bytes: Uint8Array }> => {
+  const bytes = blocks.get(cid)
+  if (!bytes) {
+    throw new MissingBlockError(cid, 'record')
+  }
+  const record = await cborToLexRecord(bytes)
+  return { record, bytes }
+}
```
devinivy, line 17: What are your thoughts on keeping lex records pretty much entirely out of the repo/storage code, and working exclusively with ipld values there.  In a more lexicon-aware such as an xrpc server (i.e. the pds) you would end-up calling `ipldToLex()` and `lexToIpld()` when talking to the repo or storage.  I know you spent a lot of time considering this kind of thing, so mostly interested in your current thinking on it.

packages/repo/src/types.ts
```diff
@@ -50,14 +51,14 @@ export type RecordCreateOp = {
...
+  record: RepoRecord
 }
 
 export type RecordUpdateOp = {
   action: WriteOpAction.Update
   collection: string
   rkey: string
-  record: Record<string, unknown>
+  record: RepoRecord
 }
```
devinivy, line 62: Re: my last comment on factoring lex away from the repo, this is a useful example of what was on my mind.  `RecordUpdateOp['record']` would be an ipld value, and `updateWriteToOp()` in `packages/pds/src/repo/prepare.ts` would convert the lex record to ipld:
```ts
export const updateWriteToOp = (write: PreparedUpdate): RecordUpdateOp => ({
  action: WriteOpAction.Update,
  collection: write.uri.collection,
  rkey: write.uri.rkey,
  record: lexToIpld(write.record),
})
```

packages/xrpc-server/src/server.ts
```diff
@@ -222,7 +223,7 @@ export class Server {
             output?.encoding === 'application/json' ||
             output?.encoding === 'json'
           ) {
-            const json = ipldValueToJson(output.body)
+            const json = lexToJson(output.body)
```
devinivy, line 226: Neat!

packages/xrpc-server/src/util.ts
```diff
@@ -128,8 +125,8 @@ export function validateInput(
   // if input schema, validate
   if (def.input?.schema) {
     try {
-      const ipld = req.body ? jsonToIpldValue(req.body) : req.body
-      req.body = lexicons.assertValidXrpcInput(nsid, ipld)
+      const lexBody = req.body ? jsonToLex(req.body) : req.body
+      req.body = lexicons.assertValidXrpcInput(nsid, lexBody)
```
devinivy, line 129: ✅ !

devinivy (approved):
Super impressive chunk of work!  There was just one behavioral question I had that I am curious to understand re: image/video types.

The other question is less important to work through to me since it was about factoring code (specifically around lex and the repo), which is something we can always tweak in the future if we want to.

This was a fun ride 👍 I dig it.

packages/pds/src/app-view/services/indexing/plugins/post.ts
dholms: Oh do you not have to keep using `?`?

I can just do `external.thumb?.ref.toString()`?

dholms: didn't know that's how the language feature worked 👀

packages/pds/tests/views/__snapshots__/author-feed.test.ts.snap
dholms: Yeah I put a note about that in the PR description. My original intention was to do `$type: image`, but I wasn't seeing any benefit out of it. You get everything you need from the mimetype. And it was making certain things - like using the output of upload.blob as an input for an image type - kinda annoying & complex. I see the lex type mainly being useful for validation.

packages/repo/src/parse.ts
dholms: Yeah I'm not opposed to going back to that. I just found that it was just prone to a lot of footguns. A lot of places where you have to make sure you correctly serialize your data before sending it to the repo. & the type system has a hard time calling all of those out.

I'm also not sure if it's bad for the repo to be _lightly_ aware of lex semantics. Like maybe you aren't allowed to put _anything_ at repo leaves. Maybe they _have_ to be a lexicon-record-deserealizable value 🤔

packages/repo/src/types.ts
dholms: Yup that is how I originally had it 👍 

The basic problem is that there is a decent bit of repo calling code where you have to remember to do that serialization. & just sticking it on the entry point to the repo pretty dramatically reduced footguns

dholms: I'm not opposed to revisiting if you think this is a leaky abstraction. But personally I'm starting to think that we move less in the direction of "repo as grabbag of _any data_ at _any key_" & more so as "repo is lightly lexicon aware, enforces data key path semantics & parseability as lexicon record at the leaves"

packages/lexicon/src/serialize.ts
dholms: I honestly thought js objects could have numbers as keys but after a quick look, turns out those are converted to strings under the hood 😅 

I'll do away with this 👍

packages/pds/src/app-view/services/indexing/plugins/post.ts
devinivy: Ah right on!  Yeah, I find it's kind nice since it allows you to be precise about where there may be a missing value :) and even more so with typescript!

packages/pds/tests/views/__snapshots__/author-feed.test.ts.snap
devinivy: I'm down with that 👍

packages/repo/src/parse.ts
devinivy: That makes sense, sounds good.  Would definitely want good support for that in the type system, so if that's not coming very naturally then that's good enough reason to punt on it for me.  Plus we can always come back to it, I don't see it as a super high stakes thing 👍

devinivy (approved):

PR #683, opened 2023-03-16
Lexicon: optional profile display name

### LINKED ISSUE #590: Make profile `displayName` optional
- Reasoning:
    - It’s technically optional already because the profile record may not exist. This change just more accurately reflects the system in total.

devinivy:
This makes the profile record `displayName` field optional.  Even though the `updateProfile` method will disappear soon, I have also updated it to handle this case for completeness.

Closes #590


packages/pds/src/db/migrations/20230316T225303411Z-profile-display-name-empty.ts
```diff
@@ -0,0 +1,120 @@
+import { Kysely, sql } from 'kysely'
+
+// Goal here is just to make displayName nullable. Need to recreate table due to dqlite limitations.
```
devinivy, line 3:
```suggestion
// Goal here is just to make displayName nullable. Need to recreate table due to sqlite limitations.
```

devinivy:
@dholms no special rush on my part, but just wanted to flag this for you— once this one is reviewed/merged it will allow a few others based on it to go through too. 🙏

dholms:
@devinivy 

was waiting to hear your thoughts on just dropping the constraint on the column in postgres 🤔 

seems like it'd be a much easier/faster migration to actually run

devinivy:
Ah sorry, didn't catch that!  Just tossed that in there 👍

dholms (approved):
awesome! thanks for doing that!

there's some codepaths for PG that'll never be hit lower in the migration if ya wanna clean that up too before merge 🙏

devinivy:
Yupyup, good calls!  Tidied those up.

PR #684, opened 2023-03-17
Lexicon: optional aggregated fields on views

### LINKED ISSUE #589: Make all aggregated counts in views optional
- Affects profile view and post view
- Reasoning:
    - There may be cases where view providers are unable or not designed to surface this information. Rather than give an incorrect count (ie “0”) they should give undefined. Client UIs should handle this by not showing the information.

devinivy:
This makes the aggregated count fields on post and profile views (e.g. `post.likeCount`, `profile.followCount`) optional.   Also made `viewer` state optional on the post view, since it falls into a similar category and is already optional on other views.  Finally, I removed the deprecated `myState` field on the profile view.

Resolves #589

### DIFF
--- a/lexicons/app/bsky/actor/defs.json
+++ b/lexicons/app/bsky/actor/defs.json
@@ -38,8 +38,7 @@
         "postsCount": {"type": "integer"},
         "creator": {"type": "string"},
         "indexedAt": {"type": "string", "format": "datetime"},
-        "viewer": {"type": "ref", "ref": "#viewerState"},
-        "myState": {"type": "ref", "ref": "#myState", "description": "Deprecated"}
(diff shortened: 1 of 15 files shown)


lexicons/app/bsky/actor/defs.json
```diff
@@ -38,8 +38,7 @@
         "postsCount": {"type": "integer"},
         "creator": {"type": "string"},
         "indexedAt": {"type": "string", "format": "datetime"},
-        "viewer": {"type": "ref", "ref": "#viewerState"},
-        "myState": {"type": "ref", "ref": "#myState", "description": "Deprecated"}
```
dholms, line 42: nice

PR #688, opened 2023-03-17
Lex: more string formats

dholms:
This introduces 4 new string formats:
- `handle`
- `nsid`
- `uri` (generally, as opposed to the specific case of `at-uri`)
- `at-identifier` (this is just the union of `did` & `handle`)

I did not add the `at-uri` format to moderation method subjects. Wanted to coordinate with @devinivy on that to make sure it makes sense.

I also briefly tried out switching cid references in the cbor-only `subscribeRepos` method to `cid-internal-ref` but hit a hiccup because dag-cbor does not support concatenated cbor & cborx was not playing nicely with it.

Closes https://github.com/bluesky-social/atproto/issues/580


devinivy:
Might be a little side-quest, but I think we should be able to add support for CIDs to cbor-x with a [custom extension](https://github.com/kriszyp/cbor-x#custom-extensions).  We have an open thread https://github.com/rvagg/cborg/issues/75 on cborg (used by @ipld/dag-cbor) to support streaming cbor, but it's probably not the shortest path to getting this working yet.

dholms:
Yeah I thought about that 🤔 

I'll take a look & see what i can cook up

dholms:
hot dog: https://github.com/bluesky-social/atproto/pull/693

packages/lexicon/src/validators/formats.ts
```diff
@@ -0,0 +1,107 @@
...
+      error: new ValidationError(
+        `${path} must be an iso8601 formatted datetime`,
+      ),
+    }
+  }
+  return { success: true, value }
+}
+
+export function uri(path: string, value: string): ValidationResult {
+  const isUri = value.match(/\w+:(\/?\/?)[^\s]+/) !== null
```
devinivy, line 25:
```suggestion
  const isUri = value.match(/^\w+:(\/?\/?)[^\s]+$/) !== null
```

devinivy, line 25: Since `[^\s]+` covers `\/?\/?`, I am not sure the `(\/?\/?)` is affecting the result.  But I also see why it's there / what it's getting at.  Just to toss a close alternative out there:
```suggestion
  const isUri = value.match(/\w+:(?:\/\/)?[^\s\/][^\s]*/) !== null
```

If we're not 100% sure how we want it, I think it's also totally cool to leave it as-is and come back to it later.

packages/pds/src/services/record/index.ts
```diff
@@ -267,9 +264,14 @@ export class RecordService {
...
   if (record?.['$type'] === ids.AppBskyGraphFollow) {
     const subject = record['subject']
-    const validLink =
-      typeof subject === 'string' && didFormat('subject', subject).success
-    if (!validLink) return []
+    if (typeof subject !== 'string') {
+      return []
+    }
+    try {
+      ident.ensureValidDid(subject)
```
devinivy, line 271: Very glad about this, I felt a little dirty dipping into lexicon internals.

devinivy (approved):
👏 lovely

packages/lexicon/src/validators/formats.ts
dholms: nice im game 👌

PR #690, opened 2023-03-18
Lexicon: repo write method semantics clarifications

### LINKED ISSUE #588: Add optional `rkey` parameter to record write methods
This will enable choosing the keyname of records

### LINKED ISSUE #617: Clarify teminology/semantics around repo updates
Currently there are three kinds of repo writes: creates, updates or puts, and deletions.  That is reflected in three different places right now:
 - `com.atproto.repo.[create|put|delete]Record`
 - `com.atproto.repo.batchWrite`
 - The interface to the repo itself, i.e. through `RecordWriteOp` and `WriteOpAction`.

Everything here is clear, except there are some details around updates/puts that could be made a bit more consistent.  For example, in xrpc you see `com.atproto.repo.putRecord` and the description of `com.atproto.repo.batchWrite` reads "Apply a batch transaction of creates, puts, and deletes.".  But at the same time, at the repo level the write action is called `WriteOpAction.Update` and the action type in `com.atproto.repo.batchWrite` is called "update".

So there's a terminology question around "put" versus "update".  But that also brings-up a semantic question, since often the term "put" communicates an idempotent version of an update (e.g. in HTTP), where the result does not depend on whether the record already exists.  An example of how this could be used in bsky, for example, would be for profile records, which have the fixed key "self"— you may want to "put" a new profile record to update your profile without needing to know whether you've already created a profile.

In my view it's worthwhile to lean into idempotent operations (e.g. for puts and deletions) for atproto's use-cases, since it's more resilient to writers not necessarily knowing the current state of the repo living remotely on the user's PDS.  We would still keep create operations around so that the repo/PDS can generate record keys as needed.  When using a "put", if the user wanted to ensure they were creating a record for the first time, we could support that through the compare-and-swap interface outlined in #582.

### LINKED ISSUE #651: createRecord should always use a TID for the rkey
The `createRecord` API should always use a TID if no rkey is supplied to ensure that the PDS behavior is consistent.

devinivy:
This clarifies the semantics of repo write xrpc methods through a. improved descriptions, b. tests, c. some implementation fixes.

TLDR there is that `deleteRecord` and `putRecord` ensure the post-condition of the record not existing (in the case of a delete) or existing (in the case of a put), and will succeed regardless of whether the record already exists or not.  The `applyWrites` method, on the other hand, is lower-level and deals with non-idempotent repo ops of create, update, and delete.

Also removes setting of `rkey` based upon lexicons on the PDS, and pushes that down into the client library.  This is relevant, for example, with profile records whose key should be the literal `self`: now the client `profile.create()` method handles that case rather than the PDS.

Resolves #651, resolves #617, resolves #588


packages/pds/src/api/com/atproto/repo/createRecord.ts
```diff
@@ -20,17 +20,15 @@ export default function (server: Server, ctx: AppContext) {
...
       const now = new Date().toISOString()
+
       let write: PreparedCreate
       try {
         write = await repo.prepareCreate({
           did,
           collection,
           record,
-          rkey,
+          rkey: rkey || repo.determineRkey(),
```
dholms, line 31: I might just take out `determineRkey` & do `TID.nextStr()` directly 🤔

dholms: This makes it look like something is still going on

dholms (approved):
nice 👌

PR #691, opened 2023-03-18
Lexicon: embeds in record embeds

### LINKED ISSUE #619: A few changes to the record embeds
Hey @devinivy now that I'm finally futzing with the record embeds, I'm hitting a couple things I could use changes on. We could group this into the lexicon refactor.

- We need the `indexedAt` field somewhere in there I think. Right now we're using `createdAt` which can be effed with.
- I was wrong. We do need the presented form to have the presented embeds of the embedded record. Specifically for images. My bad.
- Right now the presented form ends up with `record.record`. Not a huge deal but if inspiration hits, it'd be great to have another name for the outer bit.

devinivy:
A few changes to record embeds in here:
 - Embed view lexicons are named as `#view` rather than `#presented`, for consistency.
 - Embed views now have a standard wrapper shape `{ $type, value }`, whereas before we had different keys `{ $type, record }`, `{ $type, external }`, etc.  That wasn't necessary since `$type` acts as a discriminator for the wrapped value.  One small upside there is that the record under a record embed is now `(embed.value).record` rather than `(embed.record).record`, which feels a little nicer to write.
 - Record embed views now contain `embeds` which can include one or more views of embeds found within the record.  If there are multiple, it would be up to the consumer to distinguish them by inspecting `embed.value.record`.  Embeds are only hydrated one level deep.
 - Fixes to external embeds: the `uri` field was marked as an at-uri, whereas it's actually not limited to `at://` (most likely will be a web uri `https://`).

Resolves #619

### DIFF
--- a/lexicons/app/bsky/embed/external.json
+++ b/lexicons/app/bsky/embed/external.json
@@ -17,7 +17,7 @@
       "type": "object",
       "required":  ["uri", "title", "description"],
       "properties": {
-        "uri": {"type": "string", "format": "at-uri"},
+        "uri": {"type": "string"},
         "title": {"type": "string"},
         "description": {"type": "string"},
         "thumb": {
@@ -29,21 +29,21 @@
         }
       }
     },
-    "presented": {
+    "view": {
       "type": "object",
       "required": ["external"],
       "properties": {
-        "external": {
+        "value": {
           "type": "ref",
-          "ref": "#presentedExternal"
+          "ref": "#viewExternal"
         }
       }
     },
-    "presentedExternal": {
+    "viewExternal": {
       "type": "object",
       "required":  ["uri", "title", "description"],
       "properties": {
-        "uri": {"type": "string", "format": "at-uri"},
+        "uri": {"type": "string"},
         "title": {"type": "string"},
         "description": {"type": "string"},
         "thumb": {"type": "string"}
--- a/lexicons/app/bsky/embed/record.json
+++ b/lexicons/app/bsky/embed/record.json
@@ -19,12 +19,24 @@
     },
     "viewRecord": {
       "type": "object",
-      "required":  ["uri", "cid", "author", "record"],
+      "required":  ["uri", "cid", "author", "record", "indexedAt"],
       "properties": {
         "uri": {"type": "string", "format": "at-uri"},
         "cid": {"type": "string", "format": "cid"},
         "author": {"type": "ref", "ref": "app.bsky.actor.defs#withInfo"},
-        "record": {"type": "unknown"}
+        "record": {"type": "unknown"},
+        "embeds": {
+          "type": "array",
+          "items": {
+            "type": "union",
+            "refs": [
+              "app.bsky.embed.images#view",
+              "app.bsky.embed.external#view",
+              "app.bsky.embed.record#view"
+            ]
+          }
+        },
+        "indexedAt": {"type": "string", "format": "datetime"}
       }
     },
     "viewNotFound": {
--- a/lexicons/app/bsky/embed/record.json
+++ b/lexicons/app/bsky/embed/record.json
@@ -10,14 +10,14 @@
         "record": {"type": "ref", "ref": "com.atproto.repo.strongRef"}
       }
     },
-    "presented": {
+    "view": {
       "type": "object",
       "required":  ["record"],
       "properties": {
-        "record": {"type": "union", "refs": ["#presentedRecord", "#presentedNotFound"]}
+        "value": {"type": "union", "refs": ["#viewRecord", "#viewNotFound"]}
       }
     },
-    "presentedRecord": {
+    "viewRecord": {
       "type": "object",
       "required":  ["uri", "cid", "author", "record"],
       "properties": {
@@ -27,7 +27,7 @@
         "record": {"type": "unknown"}
       }
     },
-    "presentedNotFound": {
+    "viewNotFound": {
       "type": "object",
       "required":  ["uri"],
       "properties": {
--- a/packages/pds/src/app-view/services/feed/index.ts
+++ b/packages/pds/src/app-view/services/feed/index.ts
@@ -3,7 +3,7 @@ import * as common from '@atproto/common'
 import Database from '../../../db'
 import { countAll, notSoftDeletedClause } from '../../../db/util'
 import { ImageUriBuilder } from '../../../image/uri'
-import { isPresented as isPresentedImage } from '../../../lexicon/types/app/bsky/embed/images'
+import { isView as isViewImage } from '../../../lexicon/types/app/bsky/embed/images'
 import { PostView } from '../../../lexicon/types/app/bsky/feed/defs'
 import { ActorViewMap, FeedEmbeds, PostInfoMap, FeedItemType } from './types'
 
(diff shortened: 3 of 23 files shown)


lexicons/app/bsky/embed/external.json
```diff
@@ -17,7 +17,7 @@
       "type": "object",
       "required":  ["uri", "title", "description"],
       "properties": {
-        "uri": {"type": "string", "format": "at-uri"},
+        "uri": {"type": "string"},
```
dholms, line 20: i snuck in the `uri` format btw 👀

lexicons/app/bsky/embed/record.json
```diff
@@ -10,24 +10,36 @@
...
       "type": "object",
-      "required":  ["uri", "cid", "author", "record"],
+      "required":  ["uri", "cid", "author", "record", "indexedAt"],
       "properties": {
         "uri": {"type": "string", "format": "at-uri"},
         "cid": {"type": "string", "format": "cid"},
         "author": {"type": "ref", "ref": "app.bsky.actor.defs#withInfo"},
-        "record": {"type": "unknown"}
+        "record": {"type": "unknown"},
+        "embeds": {
```
dholms, line 28: is this an array or just a union?

dholms: A single record only has one embed right?

packages/pds/src/app-view/services/feed/index.ts
```diff
@@ -220,14 +224,19 @@ export class FeedService {
         records.map((p) => p.did),
         requester,
       ),
+      this.embedsForPosts(
+        records.map((p) => p.uri),
+        requester,
+        _depth + 1,
```
dholms, line 230: does this go arbitrarily deep?

so like in the case of a quote post of a quote post of a quote post of an image, this'll return that image?

dholms: oh wait just saw the _depth > 1 check. so only checks one deep?

lexicons/app/bsky/embed/record.json
devinivy: A post record has at most one embed, but if we want this to remain generic for any record type, I figured there could be multiple embeds.
> If there are multiple, it would be up to the consumer to distinguish them by inspecting `embed.value.record`.

packages/pds/src/app-view/services/feed/index.ts
devinivy: Exactly, there's a depth check in two places:
 - At the top of the function to ensure it doesn't recurse.
 - When mapping to `embeds` in the view.  The purpose there is to distinguish `embeds` being missing ("embeds are not included" e.g. due to depth) versus empty `[]` ("embeds are included but there are none").

packages/pds/src/app-view/services/feed/index.ts
```diff
@@ -265,20 +274,26 @@ export class FeedService {
...
+          $type: 'app.bsky.embed.record#view',
+          value: formatted
             ? {
-                $type: 'app.bsky.embed.record#presentedRecord',
+                $type: 'app.bsky.embed.record#viewRecord',
                 uri: formatted.uri,
                 cid: formatted.cid,
                 author: formatted.author,
                 record: formatted.record,
+                embeds: formatted.embed
```
dholms, line 288: this is one too many ternaries for my tired saturday brain to handle haha 🙃 

I don't quite understand the difference between undefined & the empty array here?

& similar to my prev comment, is there ever a case when the array contains more than one value?

devinivy: Just dropped a couple comments above that will hopefully make it clearer what I was going for.  However we land on it, I'll clean this up and drop some comments in the code 👍

lexicons/app/bsky/embed/record.json
dholms: Hmmm i think im inclined to make it only one embed 🤔 

If you want multiple embeds then you define a new type for that in the union. Similar to how we handle images which is actually up to 4 embeds but represented as only a single embed

This let's us set constraints on how many objects of what type are embedded instead of an unconstrained array

Thoughts?

dholms: I guess that also only works if this union is only for schemas that _we_ control. because another developer could take an alternate approach and have some record contain an array of embeds. But I suppose then we could just define a new single embed type that points to that array of embeds.  🤔

dholms: Idk im not fully convinced either way. But I think when we originally had this convo about embeds we chose not to make it an array because then we can set constraints on each embed type. Everything is still expressible, it might just be one semantic layer deeper

devinivy: I see what you're getting at 👍 I do think I still lean towards the array but I'm also definitely not married to it.  Part of my view on it is treating embeds the same way we might treat any other type: if record embed views needed to present all blobs that appear in the record, for example, I imagine we might end-up with an array of blobs here.  That would be necessary to cover cases like profile records which have multiple blobs, i.e. both an avatar and a background banner.  I was imagining a similar case where a record could conceivably contain two or more separate embed fields, since there's nothing restricting that.

Here's a case to illustrate where my head was at!  Consider a "question and answer" record, where an actor can present a post that poses a question next to a post that answers that question.  It might be modeled like this, with an embed for the question post and another for the answer post:
```json
{
  "commentary": {"type": "string"},
  "questionEmbed": {"type": "ref", "ref": "app.bsky.embed.record"},
  "answerEmbed": {"type": "ref", "ref": "app.bsky.embed.record"}
}
```

At the same time I'd be totally cool saying that we're just not going to cater to that for record embeds, that there's a convention that you should avoid creating records with multiple embed fields 👍 We also do always have the option to go back on the idea of this being a generic record embed, and instead making it specifically a post embed.

dholms: The argument would be that the presentation of that possibility would be
```json
"questionAndAnswerEmbedView": {
  "questionEmbed": {"type": "ref", "ref": "app.bsky.embed.record"},
  "answerEmbed": {"type": "ref", "ref": "app.bsky.embed.record"}
}
```

And similarly any array of heterogenous embeds can be encapsulated in a lex type that boils it down to a single value that's expressive about what exactly is in the array.

With that being said, it starts to sound a bit overly complex/galaxy brained as a way of encoding what's essentially a list of supplemental "view" data

dholms: I think I'd feel more strongly if this was primary data. But since it's supplemental data & the actual structure is imposed by the record, I'm game to leave this field relatively unstructured & just return an array of heterogenous embeds.

devinivy: Cool cool 👍 it's just a little tricky to deal with view information for any records in a unified way.  In the case of a complex record like the question-and-answer one, I would expect a consumer to discover embeds in the original record data.  And then fish the view information for those embeds out of this array.

I imagine same might go for blobs if we introduce those here: you discover a blob in the original record, and then fish the view of that blob (e.g. a url to a resized image) out of some collection/array of blob views.

dholms: yup yup that checks out. im game with just the array here. thx for the back and forth 🙏

dholms (approved):
💎  :shipit:

PR #692, opened 2023-03-18
Lexicon: notify users when they are quoted

### LINKED ISSUE #619: A few changes to the record embeds

devinivy:
When users are quoted via a record embed they receive a `quote` notification.

Resolves #619 in concert w/ #691

### DIFF
--- a/packages/pds/src/app-view/services/indexing/plugins/post.ts
+++ b/packages/pds/src/app-view/services/indexing/plugins/post.ts
@@ -143,38 +143,48 @@ const findDuplicate = async (): Promise<AtUri | null> => {
 
 const eventsForInsert = (obj: IndexedPost) => {
   const notifs: Message[] = []
-  for (const facet of obj.facets) {
-    if (facet.type === 'mention') {
-      if (facet.value !== obj.post.creator) {
-        notifs.push(
-          messages.createNotification({
-            userDid: facet.value,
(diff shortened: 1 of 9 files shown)


packages/pds/src/app-view/services/indexing/plugins/post.ts
```diff
@@ -143,38 +143,48 @@ const findDuplicate = async (): Promise<AtUri | null> => {
 
 const eventsForInsert = (obj: IndexedPost) => {
   const notifs: Message[] = []
+  const notified = new Set([obj.post.creator])
+  const maybeNotify = (notif: messages.NotificationInfo) => {
```
dholms, line 147: nice 👍 

so one notif per post & order of precendence goes:
- mention
- quote
- reply

dholms (approved):
excellent 🙌

PR #694, opened 2023-03-20
docs: fixed outdated usage

myConsciousness:
Hi amazing developers,

I found a commit that was recently merged into main that has not had the README fixed.

In `identifier`, `isValid` and `ensureValid` are no longer used, so I replaced them with the handle and did functions instead.


devinivy (approved):
Looks good, thanks!

PR #696, opened 2023-03-20
feat: added `InvalidDidError`

myConsciousness:
Hi,

I gave a specific name to the error about the DID validation. And for the name, I referred to `InvalidHandleError` in `handle.ts`.


dholms (approved):
Looks great! Thanks for doing that 🙏

PR #698, opened 2023-03-20
Update subscribeRepos lexicon

dholms:
- Adds non-repo updates (handle change, account migration, account tombstone)
- moves info frame to be a message frame
- changes "repo_append" -> "append"
- adds new `#signpost` event for transmitting some info about a commit that was too large to be pushed on the stream
  - i kinda wiled out & did something unexpected here. lmk if you don't like and I'll just change it to be a normal boolean flag


lexicons/com/atproto/sync/subscribeRepos.json
```diff
@@ -0,0 +1,134 @@
...
+      }
+    },
+    "info": {
+      "type": "object",
+      "required": ["name"],
+      "properties": {
+        "name": {
+          "type": "string",
+          "knownValues": ["OutdatedCursor"]
+        }
```
devinivy, line 113: The info frames also have/had a `message` field if we want to include that. 
```suggestion
        "name": {
          "type": "string",
          "knownValues": ["OutdatedCursor"]
        },
        "message": {
          "type": "string"
        }
```

dholms:
Yup good thoughts. I changed it to the much more obvious `#tooLarge` & added in a `block` field for just the dag-cbor encoded commit.

PR #699, opened 2023-03-21
Lex: link/bytes/blob revamp

dholms:
Removes lex types for
- `image`
- `video`
- `audio`
These all just use type `blob` now

Renames `cid-internal-ref` to `cid-link`

Removes dagJson encoding in favor of a lexicon encoding:
- cid-link: `{ $link: 'bafy...' }`
- bytes: `{ $bytes: 'base64encoded...' }`

Adds `size` (in bytes) to the encoded blob refs.


packages/common/src/ipld.ts
```diff
@@ -70,30 +70,27 @@ export type IpldValue =
...
   if (check.is(val, dagJsonVal)) {
-    if (!check.is(val, dagJsonValStrict)) {
-      // enforce strictly. don't allow other values in dag-json formatted fields
-      throw new Error(`improperly formatted dag-json: ${val}`)
-    }
     if (check.is(val, dagJsonCid)) {
-      return CID.parse(val['/'])
+      return CID.parse(val.$link)
     }
-    return base64.decode(`m${val['/'].bytes}`) // add mbase prefix according to dag-json code
```
devinivy, line 96: 👋 byebye multibase

devinivy (approved):
Ace 💎 

I think there's more code in the PDS we can remove related to blob validation (e.g. for `maxWidth`/`maxHeight`), but it should be all good to do that in another step if we like.

dholms:
oh duh i knew i was forgetting something. I'll do that real quick 👀

PR #700, opened 2023-03-21
Lex: change number to float

dholms:
Updates lexicon `number` primitive to `float`.

We may need some extra thought around this involving roundtrips to json


devinivy (approved):
Lookin good!

PR #701, opened 2023-03-21
fix example invite code

SafaSafari:
kindly please first learn base32 rfc then use it :))


dholms (approved):
hey it's the guy that inspired us to bump this up to 7 😉 

thanks for the edit!

PR #703, opened 2023-03-22
Integrate lex changes into subscribeRepos route

dholms:
This integrates the lex changes from https://github.com/bluesky-social/atproto/pull/698 into the subscribeRepos route

Namely work around removing InfoFrames & using the union ref on `message` to discriminate MessageFrames


packages/pds/src/app-view/subscription/repo.ts
```diff
@@ -27,6 +31,13 @@ export class RepoSubscription {
         const { ran } = await this.leader.run(async ({ signal }) => {
           const sub = this.getSubscription({ signal })
           for await (const msg of sub) {
+            if (!isCommit(msg)) {
+              appViewLogger.warn(
+                { msg },
+                'unexpected message on repo subscription stream',
+              )
+              continue
```
devinivy, line 39: Arguably we should move the cursor forward in this situation.  But not terribly worried about it, since this area will get more attention soon.

packages/xrpc-server/src/stream/stream.ts
```diff
@@ -15,7 +15,7 @@ export async function* byFrame(ws: WebSocket, options?: DuplexOptions) {
 
 export async function* byMessage(ws: WebSocket, options?: DuplexOptions) {
   for await (const frame of byFrame(ws, options)) {
-    if (frame.isMessage() || frame.isInfo()) {
+    if (frame.isMessage()) {
       // @NOTE info frames will soon be folded into messages
```
devinivy, line 19:
```suggestion
```

packages/xrpc-server/tests/subscriptions.test.ts
```diff
@@ -99,12 +95,10 @@ describe('Subscriptions', () => {
     const countdown = Number(params.countdown ?? 0)
     for (let i = countdown; i >= 0; i--) {
       yield {
-        $type:
-          i % 2 === 0 ? 'io.example.stream2#even' : 'io.example.stream2#odd',
+        $type: i % 2 === 0 ? '#even' : '#odd',
         count: i,
       }
     }
-    yield { $type: 'io.example.stream2#done' }
```
devinivy, line 107: What should the behavior be for this?  I imagine it might end-up in the header discriminated as `#done`.

packages/xrpc-server/tests/subscriptions.test.ts
```diff
@@ -99,12 +95,10 @@ describe('Subscriptions', () => {
     const countdown = Number(params.countdown ?? 0)
     for (let i = countdown; i >= 0; i--) {
       yield {
-        $type:
-          i % 2 === 0 ? 'io.example.stream2#even' : 'io.example.stream2#odd',
+        $type: i % 2 === 0 ? '#even' : '#odd',
```
devinivy, line 98: I think in a perfect world you can yield `{ $type: 'io.example.stream2#even' }`, and `#even` get's placed as the discriminator in the header.  I saw the compaction of the `$type` as being sort of an implementation detail of the wire protocol, and the application having the ability to deal with lex unions just as it normally would.  Interested to hear your thinking on it!

devinivy (approved):
Had a few tiny thoughts, but this is lookin real nice!  I think the wire protocol will be much more legible this way 💎

packages/xrpc-server/tests/subscriptions.test.ts
dholms: yup that's right^^

altho this seemed strange since it was not in the schema 🤔 

I'm down to add something back in tho

packages/pds/src/app-view/subscription/repo.ts
dholms: yeah you're right that we should ideally. but I think I'll hold off on it until we start actually sending/processing those messages. there's not any way for this to currently get into a bad state

packages/xrpc-server/tests/subscriptions.test.ts
devinivy: Agree, the same case doesn't really happen now that `codes` is gone.  I think it was a test of what happens when you yield a message with a type that doesn't have a matching code.

packages/xrpc-server/tests/subscriptions.test.ts
dholms: yup yup you're right. I fixed this up 👍

packages/xrpc-server/src/stream/subscription.ts
```diff
@@ -54,7 +54,12 @@ export class Subscription<T = unknown> {
       try {
         const cancelable = { signal: ac.signal }
         for await (const message of byMessage(ws, cancelable)) {
-          const result = this.opts.validate(message.body) // @TODO map $type
+          const t = message.header.t
+          const body = message.body
+          if (body && t !== undefined) {
+            body['$type'] = t.startsWith('#') ? this.opts.method + t : t
+          }
```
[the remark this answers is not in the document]
dholms: yeah probably not a huge worry but pushed up a change here so that we don't mutate body 👍

PR #704, opened 2023-03-22
Speed up CI tests

dholms:
A few changes to our CI inspired by [this article](https://javascript.plainenglish.io/optimizing-jest-for-faster-ci-performance-with-github-actions-f4d7100c86c5)

- allow jest to cache things in between runs (this is a reported 50% speed up)
- set `maxWorkers` to `2`. GitHub Action VMs have 2 cores and jest by default only uses 1 of them
- shard jest across 3 different VMs to further parallelize

### DIFF
--- a/.github/workflows/repo.yaml
+++ b/.github/workflows/repo.yaml
@@ -12,6 +12,9 @@ jobs:
       - run: yarn install --frozen-lockfile
       - run: yarn build
   test:
+    strategy: 
+      matrix:
+        shard: [1/3, 2/3, 3/3]
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


dholms:
2x speed up. better than nothing 🤷‍♂️

.github/workflows/repo.yaml
```diff
@@ -19,15 +19,22 @@ jobs:
...
         with:
           node-version: 18
           cache: "yarn"
+      - uses: actions/cache@v3
+        with:
+          path: /tmp/jest_rt
+          key: ${{ hashFiles('**/yarn.lock') }}-${{ matrix.shard }}
       - run: yarn install --frozen-lockfile
-      - run: yarn test
+      - run: yarn test:withFlags --maxWorkers=2 --shard=${{ matrix.shard }} --passWithNoTests
```
devinivy, line 37: Would there be an issue using the double-dash here rather than in the new command?  E.g.:
```
yarn test -- --maxWorkers=2 --shard=${{ matrix.shard }} --passWithNoTests
```

Truly no issue with it as-is, just more used to doing something like that in these kinds of situations.

devinivy (approved):
Cool!

.github/workflows/repo.yaml
dholms: It didn't want to pass through the arguments correctly when I tried that^^

Lerna tries to interpret the `maxWorkers` argument

devinivy: Ah, fun!

PR #706, opened 2023-03-23
Blob resolution on bsky app view

devinivy:
This sets-up an endpoint for resolving and verifying blob contents based on a did and cid, by reaching out to `com.atproto.sync.getBlob`.  The image processing server and image URLs have been updated to reflect this.


packages/bsky/tests/_util.ts
```diff
@@ -134,14 +135,32 @@ export const runTestServer = async (
   const bskyServer = await bsky.start()
   const bskyPort = (bskyServer.address() as AddressInfo).port
 
+  // Map pds public url to its local url when resolving from plc
+  const origResolveDid = DidResolver.prototype.resolveDid
```
dholms, line 139: 🆒 

in general, i think we should work on local testing utilities like this. Wouldn't mind for this to eventually just live on did-resolver

packages/bsky/tests/blob-resolver.test.ts
```diff
@@ -0,0 +1,83 @@
...
+    )
+    expect(status).toEqual(400)
+    expect(data).toEqual({
+      error: 'BadRequestError',
+      message: 'Invalid did',
+    })
+  })
+
+  it('400s on invalid cid.', async () => {
+    const { data, status } = await client.get(`/blob/${fileDid}/barfy`)
```
dholms, line 67: 🤮

packages/bsky/src/api/blob-resolver.ts
```diff
@@ -0,0 +1,79 @@
...
+      const verifyCid = new VerifyCidTransform(cid)
+
+      // Send chunked response, destroying stream early (before
+      // closing chunk) if the bytes don't match the expected cid.
+      res.statusCode = 200
+      res.setHeader(
+        'content-type',
+        getBlob.headers['content-type'] || 'application/octet-stream',
+      )
+      pipeline([imageStream, verifyCid, res], ignore)
```
dholms, line 50: very slick

dholms: altho do we want to handle errors at all? even just logging them or something? or will they get caught elsewhere?

packages/bsky/src/api/blob-resolver.ts
```diff
@@ -0,0 +1,79 @@
...
+
+      const imageStream: Readable = getBlob.data
+      const verifyCid = new VerifyCidTransform(cid)
+
+      // Send chunked response, destroying stream early (before
+      // closing chunk) if the bytes don't match the expected cid.
+      res.statusCode = 200
+      res.setHeader(
+        'content-type',
+        getBlob.headers['content-type'] || 'application/octet-stream',
```
dholms, line 48: I checked the `getBlob` route on the PDS & I think it's always giving `application/octet-stream` 🤔 

May wanna sneak a change in for that here if we're feeling up for it

dholms: I forget, does imgproxy need a hint as to the content type? Or it can just take raw bytes & figure out the rest just fine?

dholms (approved):
Stoked. this is pretty awesome 🙌

packages/bsky/tests/_util.ts
devinivy: That sounds interesting, I'll be the lookout for opportunities to do that.  Still sussing-out the best way to slice this.  For example, I would have loved to just stick with configuring `AtpAgent` with a custom fetch handler, which is pretty clean.  But I don't use the `AtpAgent` in the image service, since I want to stream the result and we don't have that facility yet in @atproto/api—so modifying the PLC result ended-up being the shortest path in this case.

packages/bsky/src/api/blob-resolver.ts
devinivy: Hmmm good call!  I think they go into express and http-land and get handled there, since it's an error during response transmission which is a little atypical.  I will add a special log for it in here since it's pretty straightforward to do that.

packages/bsky/src/api/blob-resolver.ts
devinivy: It should figure it out just fine without the content-type hint 👍 I'll make that change though, looks pretty straightforward.

PR #658, opened 2023-03-14
Lexicon refactor

dholms:
Feature branch for lex refactor
- https://github.com/bluesky-social/atproto/pull/653
- https://github.com/bluesky-social/atproto/pull/655
- https://github.com/bluesky-social/atproto/pull/659
- https://github.com/bluesky-social/atproto/pull/661
- https://github.com/bluesky-social/atproto/pull/664
- https://github.com/bluesky-social/atproto/pull/665
- https://github.com/bluesky-social/atproto/pull/668
- https://github.com/bluesky-social/atproto/pull/673
- https://github.com/bluesky-social/atproto/pull/674
- https://github.com/bluesky-social/atproto/pull/679
- https://github.com/bluesky-social/atproto/pull/680
- https://github.com/bluesky-social/atproto/pull/683
- https://github.com/bluesky-social/atproto/pull/684
- https://github.com/bluesky-social/atproto/pull/685
- https://github.com/bluesky-social/atproto/pull/686
- https://github.com/bluesky-social/atproto/pull/688
- https://github.com/bluesky-social/atproto/pull/690
- https://github.com/bluesky-social/atproto/pull/691
- https://github.com/bluesky-social/atproto/pull/692
- https://github.com/bluesky-social/atproto/pull/693
- https://github.com/bluesky-social/atproto/pull/699
- https://github.com/bluesky-social/atproto/pull/700
- https://github.com/bluesky-social/atproto/pull/707
- https://github.com/bluesky-social/atproto/pull/708
- https://github.com/bluesky-social/atproto/pull/717
- https://github.com/bluesky-social/atproto/pull/718
- https://github.com/bluesky-social/atproto/pull/712
- https://github.com/bluesky-social/atproto/pull/720
- https://github.com/bluesky-social/atproto/pull/723
- https://github.com/bluesky-social/atproto/pull/724
- https://github.com/bluesky-social/atproto/pull/705
- https://github.com/bluesky-social/atproto/pull/727
- https://github.com/bluesky-social/atproto/pull/728
- https://github.com/bluesky-social/atproto/pull/729
- https://github.com/bluesky-social/atproto/pull/731
- https://github.com/bluesky-social/atproto/pull/732
- https://github.com/bluesky-social/atproto/pull/733
- https://github.com/bluesky-social/atproto/pull/735

PR #707, opened 2023-03-23
Lexicon: fix removed lexicons

devinivy:
Addressing review from Paul on #658: he found a few lexicons that should have been removed in the lex refactor.  Also removed `creator` from the profile view, which was likely a byproduct of scenes and declarations.


pfrazee:
LGTM

PR #708, opened 2023-03-23
Lexicon: support separate images embed

devinivy:
This enables quote posts to contain their own image embeds.  It's achieved by allowing any post to have an images embed separate from other embeds.  In the case that a post contains two images embeds, the one in the `images` field is the one that should be presented to the user.
 - Add `images` field to post record, containing an images embed.
 - Add `images` field to post view, containing an images embed view.
 - Remove images embed view from post view `embed` field, to remove ambiguity from client.
 - Handle uploads and indexing of post `images`.
 - Ensure multiple embeds make it into a record embed view's `embeds` field, e.g. when a post contains both a quote and images.

Two things to look at here:
1. In the post view lexicon I sort of force image embeds to be presented in the `images` field, even though in legacy cases they could be stored in the post record's `embed` field.  I thought this was the right call to avoid ambiguity, but it was also a hard decision.
2. Slight awkwardness around the post record's `images` field since it contains an images embed, which also has a single `images` field.  So one might write `postRecord.images.images`.  On the view it looks like `postView.images.value` as of #691.

### DIFF
--- a/lexicons/app/bsky/feed/defs.json
+++ b/lexicons/app/bsky/feed/defs.json
@@ -10,12 +10,16 @@
         "cid": {"type": "string", "format": "cid"},
         "author": {"type": "ref", "ref": "app.bsky.actor.defs#withInfo"},
         "record": {"type": "unknown"},
+        "images": {
+          "type": "ref",
+          "ref": "app.bsky.embed.images#view"
+        },
         "embed": {
           "type": "union",
           "refs": [
-            "app.bsky.embed.images#view",
             "app.bsky.embed.external#view",
-            "app.bsky.embed.record#view"
+            "app.bsky.embed.record#view",
+            "app.bsky.embed.images#view"
           ]
         },
         "replyCount": {"type": "integer"},
--- a/lexicons/app/bsky/feed/defs.json
+++ b/lexicons/app/bsky/feed/defs.json
@@ -18,8 +18,7 @@
           "type": "union",
           "refs": [
             "app.bsky.embed.external#view",
-            "app.bsky.embed.record#view",
-            "app.bsky.embed.images#view"
+            "app.bsky.embed.record#view"
           ]
         },
         "replyCount": {"type": "integer"},
--- a/packages/pds/src/app-view/services/indexing/plugins/post.ts
+++ b/packages/pds/src/app-view/services/indexing/plugins/post.ts
@@ -2,7 +2,10 @@ import { sql } from 'kysely'
 import { CID } from 'multiformats/cid'
 import { AtUri } from '@atproto/uri'
 import { Record as PostRecord } from '../../../../lexicon/types/app/bsky/feed/post'
-import { isMain as isEmbedImage } from '../../../../lexicon/types/app/bsky/embed/images'
+import {
+  isMain as isEmbedImages,
+  Main as EmbedImages,
+} from '../../../../lexicon/types/app/bsky/embed/images'
 import { isMain as isEmbedExternal } from '../../../../lexicon/types/app/bsky/embed/external'
 import { isMain as isEmbedRecord } from '../../../../lexicon/types/app/bsky/embed/record'
 import {
@@ -26,7 +29,8 @@ type PostEmbedRecord = DatabaseSchemaType['post_embed_record']
 type IndexedPost = {
   post: Post
   facets: { type: 'mention' | 'link'; value: string }[]
-  embed?: PostEmbedImage[] | PostEmbedExternal | PostEmbedRecord
+  images?: PostEmbedImage[]
+  embed?: PostEmbedExternal | PostEmbedRecord
   ancestors: PostHierarchy[]
 }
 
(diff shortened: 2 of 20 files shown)


devinivy:
Going to make a couple small changes to this, coming out of a chat with Paul.

devinivy:
@pfrazee I took a swing at those changes.
 - Flatten images embed onto post record as `images: ImageRef[]` rather than `images: ImagesEmbed`.
 - Flatten images embed view onto post view as `images: ImageView[]` rather than `images: ImagesEmbedView`.
 - The images from `post.images` still appear a proper embed when nested in a record embed view's `embeds` field.
   - For example, if you quote a post that has images and an external embed then the `embeds` field of the view for the quote post will have two items in it: an images embed view and an external embed view.
 - Sync-up the embed views to match the structure of the embeds in the record.  So an external embed's view has a wrapper with an `external` key, for example.  So now all embed views do not have identical `{ value }` wrappers.
 - The record embed view is shaped `{ uri, cid, value, ... }` rather than `{ uri, cid, record, ... }`, which means you wont ever write `postView.embed.record.record`.  Instead it would be `postView.embed.record.value`.  It turns out this is actually more consistent with the other APIs anyway!

packages/pds/src/app-view/services/indexing/plugins/post.ts
```diff
@@ -76,17 +80,19 @@ const insertFn = async (
...
   })
   // Embed indices
-  let embed: PostEmbedImage[] | PostEmbedExternal | PostEmbedRecord | undefined
-  if (isEmbedImage(obj.embed)) {
-    const { images } = obj.embed
-    embed = images.map((img, i) => ({
+  let images: PostEmbedImage[] | undefined
+  let embed: PostEmbedExternal | PostEmbedRecord | undefined
+  if (obj.images || isEmbedImages(obj.embed)) {
+    const embeddedImages = obj.images || (obj.embed as EmbedImages).images
```
dholms, line 86: so you can't have images in both, right? it's one or the other

packages/pds/src/repo/prepare.ts
```diff
@@ -26,7 +30,7 @@ import {
 
 // @TODO do this dynamically off of schemas
```
dholms, line 31: ah man we should do this off of blob refs now 😛 

i may take a stab at that soon

lexicons/app/bsky/feed/defs.json
```diff
@@ -10,10 +10,14 @@
...
         "record": {"type": "unknown"},
+        "images": {
+          "type": "array",
+          "items": {"type": "ref", "ref": "app.bsky.embed.images#viewImage"},
+          "maxLength": 4
+        },
         "embed": {
           "type": "union",
           "refs": [
-            "app.bsky.embed.images#view",
```
dholms, line 16: I think that's the right call 👍

dholms (approved):
Nice this is honestly much cleaner with images as it's own field. Especially given how common it is 👌

pfrazee:
Nice, looking good.

I just, tragically, had one last thought hit me. We're looking at images here. Is it possible we'll want other kinds of media to get promoted like this? Like videos?

Even more annoyingly, are we 100% sure we'll never want both a external (link card) and quote post? I think I missed that thought when we were talking about making embeds an array 😐

Let's chat about this tomorrow morning. I also have a thought about arrays of open unions & schema negotiation rattling around.

packages/pds/src/repo/prepare.ts
devinivy: I was tempted, haha!

pfrazee (approved):
Whew, okay yes. LGTM. 🙏

dholms (commented):
Code looks good 👍 

I guess I have some questions about this approach still...

To take the two extremes:
Lenient - an array of embeds/media. If you jammed to much stuff in there, it's up to the client to handle it
Rigid - only allowable combinations are representable

I'm worried we fall somewhere in the middle where we don't get the benefits of either.

In this example, you can do an external + 4 images. But not an external + embed. Is that right? 

The embed type that is opening this up is `external` & I'm wondering if we try to encapulate that some how by doing one of a couple things:
- moving `external` off into it's own field (instead of media) so that it can be composed with other embed types
- making a higher level embed type which is something like `externalWithEmbeds` that leaves everything in embeds (completely backwards compatible!) & wraps the allowed combinations in a type

In general, I'm worried that if we start introducing more embed types, that this system will continue to contort in strange ways. And I thought the plan was to handle that complexity by forcing _every allowable combo of embeds_ to have a schema.

pfrazee:
Yep, we had this exact discussion. It's a bit of a "flexible but complex" versus "consistent but rigid" debate. I could go either way on this, but here are the reasons we landed on the latter:

1. Implementing the array model requires more client side work
2. The array model may be handled differently between clients and may make publishing less predictable for users
3. We can always v2 the post as-needed

I really don't feel strongly about this, though, unless there's another consideration that I've memory-holed which is very possible

devinivy:
Closed in favor of #723

schwentker:
hi perhaps the image size, dpi etc has been indicated above 

I'm trying to post images with text in them and cant find any work around for the images getting truncated. 

bsky-devrel1

bsky-truncated-image

pfrazee:
This is just an annoying failure of the web app. We’ll fix that soon (I hope)On Mar 31, 2023, at 7:59 PM, Robert Schwentker ***@***.***> wrote:﻿
hi perhaps the image size, dpi etc has been indicated above
I'm trying to post images with text in them and cant find any work around for the images getting truncated.

PR #709, opened 2023-03-23
Drop scene tables

dholms:
This drops the tables relevant to tracking scenes.

I almost made the edit to `did_handle` to drop the actorType & declarationCid columns, but that's already being handled by https://github.com/bluesky-social/atproto/pull/658


devinivy (approved):
✅ !!

PR #710, opened 2023-03-23
Delete scenes app migration

dholms:
This is an app migration for deleting the accounts of scenes.

We grab all scenes from the `did_handle` table & then call the delete methods on each relevant service - as we would in `com.atproto.account.delete`

This should be run _before_ deploying https://github.com/bluesky-social/atproto/pull/658

### DIFF
--- a/packages/pds/src/app-migrations/remove-scenes.ts
+++ b/packages/pds/src/app-migrations/remove-scenes.ts
@@ -35,6 +35,8 @@ async function main(ctx: AppContext, tx: Database) {
     }),
   )
 
+  console.log(`deleted ${sceneDids.length} scenes`)
+
   // sanity check
 
   const check1 = tx.db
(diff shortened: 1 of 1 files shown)


packages/pds/src/app-migrations/remove-scenes.ts
```diff
@@ -0,0 +1,66 @@
...
+    .selectFrom('did_handle')
+    .where('did', 'in', sceneDids)
+    .selectAll()
+    .execute()
+
+  const check3 = tx.db
+    .selectFrom('record')
+    .where('did', 'in', sceneDids)
+    .selectAll()
+    .execute()
```
devinivy, line 58: I think this will also remove some assertions, if we want to do any dummy-checking in the app view.

devinivy (approved):
Perfect!

packages/pds/src/app-migrations/remove-scenes.ts
dholms: Nice I added that check 👌

devinivy (approved):

PR #711, opened 2023-03-23
Rework sequencer storage & add handle update

dholms:
This reworks our sequencer storage so that we do the work upfront instead of collating it when we serve.
This does cause some duplicated data however, I do not think it's a concern for several reasons:
- we are not storing large events here
- given our 24hr backfill, we can garbage collect old seqs whenever

This allows our `repo_seq` table to hold heterogenous event types. It will also speed up the subscription route.

This PR also adds in support for handle update events.

---

- Dropped `repo_op` table as we weren't using it anymore
- For `tooBig` events I chose a limit of 200 ops or 1MB of block data.
- Put a limit of 200 ops on `batchWrite`
- I cbor encoded messages becase:
  - it let me encode CIDs
  - but mostly, sqlite doesn't support json & this seemed better than json stringifying
  - it does unfortunately make our PG database more opaque
- I added an `invalidatedBy` column to seq which tracks if an event is invalidated by a later event namely:
  - commits are invalidated by a later rebase
  - handle updates are invalidated by a later handle update


packages/pds/src/api/com/atproto/repo/applyWrites.ts
```diff
@@ -29,6 +29,9 @@ export default function (server: Server, ctx: AppContext) {
           'Unvalidated writes are not yet supported.',
         )
       }
+      if (tx.writes.length > 200) {
+        throw new InvalidRequestError('Two many writes. Max: 200')
+      }
```
devinivy, line 34: I assume this check lives here because you don't want it to be a hard limit in the lexicon (i.e. for all PDS implementations), is that right?

packages/pds/src/api/com/atproto/repo/applyWrites.ts
```diff
@@ -29,6 +29,9 @@ export default function (server: Server, ctx: AppContext) {
           'Unvalidated writes are not yet supported.',
         )
       }
+      if (tx.writes.length > 200) {
+        throw new InvalidRequestError('Two many writes. Max: 200')
```
devinivy, line 33:
```suggestion
        throw new InvalidRequestError('Too many writes. Max: 200')
```

packages/pds/src/db/migrations/20230323T202553064Z-rework-seq.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  const binaryDatatype = dialect === 'sqlite' ? sql`blob` : sql`bytea`
+  await builder
+    .addColumn('did', 'varchar', (col) => col.notNull())
+    .addColumn('eventType', 'varchar', (col) => col.notNull())
+    .addColumn('event', binaryDatatype, (col) => col.notNull())
+    .addColumn('sequencedAt', 'varchar', (col) => col.notNull())
+    .execute()
+
+  await db.schema
```
devinivy, line 36: Having trouble recalling which types of queries these indices support.  I kinda like the comments we've added in some of the other migrations documenting this, have found it useful!

packages/pds/src/db/migrations/20230323T202553064Z-rework-seq.ts
```diff
@@ -0,0 +1,100 @@
...
+  await db.schema.dropIndex(repoSeqCommitIndex).execute()
+  await db.schema.dropIndex(repoSeqDidIndex).execute()
+  await db.schema.dropTable(repoSeqTable).execute()
+  await db.schema.dropTable(repoOpTable).execute()
+
+  let builder = db.schema.createTable(repoSeqTable)
+  if (dialect === 'pg') {
+    builder = builder
+      .addColumn('seq', 'bigserial', (col) => col.primaryKey())
+      .addColumn('invalidatedBy', 'bigint')
```
devinivy, line 21: I dig the `invalidatedBy` idea!  I wonder if we should also make it a foreign key.

packages/pds/src/sequencer/events.ts
```diff
@@ -0,0 +1,144 @@
...
+  commitData: CommitData,
+  writes: PreparedWrite[],
+) => {
+  let tooBig: boolean
+  const ops: CommitEvtOp[] = []
+  const blobs = new CidSet()
+  let carSlice: Uint8Array
+
+  // max 200 ops or 1MB of data
+  if (writes.length > 200 || commitData.blocks.byteSize > 1024000) {
```
devinivy, line 26: Such a goofy nit, sorry! 🙈 But I wonder if we should either go for 1000^2 or 1024^2 for the max byte size, rather than 1024*1000.  Realize you may have had a good reason for that I'm overlooking though!

packages/pds/src/sequencer/events.ts
```diff
@@ -0,0 +1,144 @@
...
+    tooBig = true
```
devinivy, line 27: Just my opinion but I think it might be good for compatibility to place these constraints right in the lexicon for the subscription endpoint.  Everyone agreeing on what is "too big" of an event will be useful for consumers.  E.g. if consumers are going to keep a buffer of a fixed size for placing the car, everyone will agree on how big that buffer must be.

packages/pds/src/sequencer/events.ts
```diff
@@ -0,0 +1,144 @@
...
+    const justRoot = new BlockMap()
+    justRoot.add(commitData.blocks.get(commitData.commit))
+    carSlice = await blocksToCar(commitData.commit, justRoot)
```
devinivy, line 30: 💎

packages/pds/src/sequencer/index.ts
```diff
@@ -56,24 +58,14 @@ export class Sequencer extends (EventEmitter as new () => SequencerEmitter) {
...
-      .select([
-        'repo_seq.seq as seq',
-        'repo_seq.did as did',
-        'repo_seq.commit as commit',
-        'repo_seq.sequencedAt as sequencedAt',
-        'repo_commit_history.prev as prev',
-      ])
+      .selectAll()
       .orderBy('seq', 'asc')
+      .where('invalidatedBy', 'is', null)
```
devinivy, line 68: 🤤 so nice, this thing is gonna whiz.

devinivy (approved):
👏 so nice

packages/pds/src/api/com/atproto/repo/applyWrites.ts
dholms: lol woops

packages/pds/src/api/com/atproto/repo/applyWrites.ts
dholms: Yea exactly. There are cases where you may want to do more than 200 writes & i feel that we shouldn't put this at the lexicon level. There's a good chance we'll remove this check at some point when we have a better handle on the repercussions of large mutations

packages/pds/src/db/migrations/20230323T202553064Z-rework-seq.ts
dholms: yup i'll add some comments 👍

packages/pds/src/db/migrations/20230323T202553064Z-rework-seq.ts
dholms: ah yeah I'd be game 👍

packages/pds/src/sequencer/events.ts
dholms: yup that's a good point 👍

PR #714, opened 2023-03-24
Lexicon: browser friendly api package

devinivy:
In the lex refactor the @atproto/api package became a bit unfriendly to non-none environments, so this introduces some updates to make it web-friendly version.

 - Split web-safe parts of `@atproto/common` out into `@atproto/common-web`.  The `@atproto/common` package is inclusive of `@atproto/common-web`.
 - Removed cbor-to-lex-related functionality from `@atproto/lexicon`, some of it finding a new home in `@atproto/repo`.
 - Updated `@atproto/api` and `@atproto/common-web` build to be catered to browser environments.  Updated `@atproto/identifier` to give a hint for browser bundling (sidestepping dns package for resolving an identifier).
 - Misc updates and tidying.


devinivy:
See also https://github.com/bluesky-social/atproto/pull/715 which is similar but a slightly different approach (no separate common-web package).

pfrazee:
Tough call! Both seem good to me. I have a vague feeling the common-web approach is going to give more direct control over the outcome whereas #715 is cleaner but runs more risk of fighting with build outputs.

dholms (approved):
Yup I like this 👍 

I think the methods you moved to `repo` are actually a bit happier there & tbh it resolves some of the friction that I had about using them in `repo` in the first place

I like this better than the approach in https://github.com/bluesky-social/atproto/pull/715

Both are fine, but I find this more explicit & I've just had so many hard to debug nightmares with package.json browser/exports/etc, that I get a bit less anxious with this approach 😅

PR #715, opened 2023-03-24
Lexicon: browser friendly api package (alternate)

devinivy:
We could also do this rather than the approach in #714.  The difference is that in #714 there's explicitly a common-web package for web-friendly utilities, and here we use the `browser` field in package.json to ask the bundler to stub-out certain exports from common on a per-file basis.  This is a little cleaner IMO but open to either option.


devinivy:
Closed in favor of #714

PR #717, opened 2023-03-25
Lexicon: standardize record repo interfaces

devinivy:
This unifies all the `com.atproto.repo.*` xrpc methods so that the repo to read/write to is specified with a `repo` field, which can be an at-identifier (i.e. did or handle). Previously the write endpoints instead took a `did` field as an input, while the read endpoints a `repo`.


pfrazee:
LGTM, thanks for hitting this

PR #718, opened 2023-03-25
Lexicon: remove follows of scenes

devinivy:
Removes follows of scene records.  I used a static list here since it's short, it's fixed, and the `actorType` information would be dropped from the db before running the app migration unless we wanted to do some special coordination around that.


dholms (approved):
aw sweet tbh i was just gonna let this one slip but stoked that you did it. hardcoding the scenes made a lot of sense as an approach. good call 🙏

PR #719, opened 2023-03-27
Bsky app view indexing per repo

devinivy:
The main goal here is to allow independent indexing in-process per repository.  That means that indexing work can be parallelized on a user-by-user basis.  Currently this all happens in-process.  A result of that is that we also need to be able to index updates out-of-order, which affects our thread indexer (e.g. if a reply comes in before the post it was replying to).

Three ideas going on here:
 - Each message is popped onto an in-memory queue that is partitioned by did.
 - When a message is completed (including failed), we check if there are any consecutive messages completed ahead of the subscription cursor.  If there are some, we move the cursor forward.
 - For thread indexing, we now add a step to index post descendents in the case that posts within a thread are indexed out-of-order.  To do this we also take a lock on the thread via the thread's root post.

I also did a little work here to make the tests more reliable.


packages/bsky/src/services/indexing/plugins/post.ts
```diff
@@ -99,19 +100,46 @@ const insertFn = async (
...
-      depth: 0,
-    })
-    .onConflict((oc) => oc.doNothing()) // Supports post updates
+    .values([
+      {
+        uri: post.uri,
+        ancestorUri: post.uri,
+        depth: 0,
+      },
+      ...(post.replyParent
```
dholms, line 125: any chance we could do this with fewer operators outside of the db query? my brain starts to glaze over when i see this many symbols 😅

packages/bsky/src/services/indexing/plugins/post.ts
```diff
@@ -124,11 +152,43 @@ const insertFn = async (
...
           ]),
       )
-      .onConflict((oc) => oc.doNothing()) // Supports post updates
+      .onConflict((oc) => oc.doNothing())
       .returningAll()
       .execute()
+    ancestors.push(...deepAncestors)
+  }
+
+  // Copy all post's relations down to its descendents. This ensures
```
dholms, line 161: very slick 👏 🙇

packages/bsky/src/services/indexing/plugins/post.ts
```diff
@@ -153,8 +213,11 @@ const eventsForInsert = (obj: IndexedPost) => {
     }
   }
   const notified = new Set([obj.post.creator])
-  const ancestors = [...obj.ancestors].sort((a, b) => a.depth - b.depth)
+  const ancestors = [...(obj.ancestors ?? [])].sort((a, b) => a.depth - b.depth)
```
dholms, line 216: out of curiosity, whats the point of the destructuring right back into an array?

packages/bsky/src/services/indexing/plugins/post.ts
```diff
@@ -170,6 +233,14 @@ const eventsForInsert = (obj: IndexedPost) => {
       )
     }
   }
+
+  if (obj.descendents) {
+    // May generate notifications for out-of-order indexing of replies
```
dholms, line 238: yup yup that makes sense. the timing on these if they're out of order should be close anyway 👌

packages/bsky/src/subscription/repo.ts
```diff
@@ -10,74 +10,110 @@ import Database from '../db'
...
-                await Promise.all([
-                  this.handleOps(tx, ops, msg.time),
-                  this.handleActor(tx, msg.repo, msg.time),
-                ])
-                await this.setState(tx, { cursor: msg.seq })
+            const item = this.consecutive.push(msg)
+            this.repoQueue
+              .add(msg.repo, () => this.handleMessage(msg))
+              .catch((err) => {
+                subLogger.error(
```
dholms, line 41: making sure i understand, we just drop messages we can't process, right?

dholms (approved):
nice this is looking great

packages/bsky/src/services/indexing/plugins/post.ts
devinivy: It's a habit I have since `sort()` mutates the array.  Luckily they are adding non-mutating versions of these array methods to the language 🙃

packages/bsky/src/subscription/repo.ts
devinivy: Yep that's right, I'll add a comment about that with a little rationale.

packages/bsky/src/services/indexing/plugins/post.ts
devinivy: I ended-up filtering out the `depth === 0` item up here rather than in the loop, which made the clone unnecessary 👍

PR #720, opened 2023-03-27
Add grapheme utilities to api RichText

pfrazee:
Adds `.length` and `.graphemeLength` to the `RichText` API


packages/api/src/rich-text/unicode.ts
```diff
@@ -23,6 +25,10 @@ export class UnicodeString {
     return this.utf8.byteLength
   }
 
+  get graphemeLength() {
+    return graphemeLen(this.utf16)
+  }
```
devinivy, line 30: Definitely not necessary now, but it might be nice to eventually cache this result.

dholms (approved):
sweet 👌 👌 

(btw working on fixing the CI test failures)

PR #721, opened 2023-03-27
Streaming test fix

dholms:
Fixed up the streaming/sequencing tests so that they finish deterministically & do not rely on timing

This was some really goofy/heady JS async stuff. These utilities are only used in tests. But lmk if you disagree with the approach or have other ideas


packages/common-web/src/async.ts
```diff
@@ -1,17 +1,32 @@
-import { bailableWait } from './util'
+import { wait } from './util'
+
+export const emptyPromise: Promise<void> = new Promise((resolve) => resolve())
```
devinivy, line 3: No biggie here either way, just a more concise way to write the exact same thing:
```suggestion
export const emptyPromise: Promise<void> = Promise.resolve()
```

devinivy (approved):
My hero!

packages/common-web/src/async.ts
dholms: oh far out didn't realize that

PR #722, opened 2023-03-27
Use swc for jest

dholms:
This uses [swc](https://swc.rs/) a rust-based compiler for jest which gives us some massive speedups.

We should consider using this for building as well, however tests were the critical path at the moment

### DIFF
--- a/packages/repo/src/repo.ts
+++ b/packages/repo/src/repo.ts
@@ -31,6 +31,7 @@ export class Repo extends ReadableRepo {
 
   constructor(params: Params) {
     super(params)
+    this.storage = params.storage
   }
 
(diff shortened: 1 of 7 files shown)


packages/repo/src/repo.ts
```diff
@@ -31,6 +31,7 @@ export class Repo extends ReadableRepo {
 
   constructor(params: Params) {
     super(params)
+    this.storage = params.storage
```
devinivy, line 34: Seemed slightly unrelated, just wanted to make sure it landed in here intentionally.

devinivy (approved):
🙏 🙏 🙏

PR #724, opened 2023-03-28
Lex: actor view semantics

dholms:
This renames `actor.def#withInfo` & standardizes on 3 different profile view options:
- `#profileViewBasic`: previous `#withInfo`. used in feeds & search when we don't want to be too verbose
- `#profileView`: also includes description & indexedAt time. used in lists of accounts like `getLikes`, `getFollows`, etc
- `#profileViewDetailed`: full profile info. used in getProfile & getProfiles


devinivy (approved):
Excellent

PR #725, opened 2023-03-28
Bsky handling of rebases and too-big commits

devinivy:
Implements indexing logic for a "too-big" repo sync event: 
 - Fetch and verify a checkout at that commit for the user.
 - Unindex the user and immediately index their latest profile.
 - Index the records in checkout, 100 record at a time.

Added a new `verifyCheckoutWithCids` which is similar to `verifyCheckout`, but it retains the cids of each record in the returned repo contents.

Since we re-index records which may already exist and their `indexedAt` time will be updated, we now keep a `sortAt` column in the database automatically computed to be the earliest of `indexedAt` and `createdAt`— this is now used for sorting in endpoints such as `getAuthorFeed`, `getFollows`, etc.

I also added retries to all http requests made by bsky out to actors' PDSes.

Rebases are also handled here— we keep track of the latest commit data cid for each actor, and when a rebase comes through we no-op if the data cid hasn't changed.  Otherwise we reindex the repository, similarly to a too-big commit.

Will add some more tests to the repo subscription once #730 lands and makes its way into the app view.


packages/repo/src/verify.ts
```diff
@@ -50,6 +41,25 @@ export const verifyCheckout = async (
...
+  if (repo.did !== did) {
+    throw new RepoVerificationError(`Invalid repo did: ${repo.did}`)
+  }
+  const validSig = await util.verifyCommitSig(repo.commit, signingKey)
+  if (!validSig) {
+    throw new RepoVerificationError(
+      `Invalid signature on commit: ${repo.cid.toString()}`,
+    )
+  }
+  return repo
```
dholms, line 60: one concern here is that without walking the datastore (with a diff), we're not actually verifying that the repo is properly structured

so if you say `verifyCheckoutToRepo` expecting that the repo you get is verified, it may be poorly formatted once you get down into the datastore.

That being said, if you try to do anything with a poorly formatted part, it'll still error so probably not a huge concern 🤔

packages/bsky/src/services/indexing/index.ts
```diff
@@ -93,6 +103,57 @@ export class IndexingService {
     }
   }
 
+  async indexRepo(did: string, commit: string) {
+    const now = new Date().toISOString()
+    const { pds, signingKey } = await this.didResolver.resolveAtpData(did)
+    const { sync } = new AtpAgent({ service: pds }).api.com.atproto
+
+    const { data: car } = await sync.getCheckout({ did, commit })
+    const { root, blocks } = await readCarWithRoot(new Uint8Array(car))
```
dholms, line 112: after lex refactor this'll actually be returning the right type & you won't have to wrap in `uint8array`!

packages/bsky/src/services/indexing/index.ts
```diff
@@ -93,6 +103,57 @@ export class IndexingService {
...
+    const storage = new MemoryBlockstore(blocks)
+    const repo = await verifyCheckoutToRepo(storage, root, did, signingKey)
+
+    // First, replace the user with just their current profile if it exists
```
dholms, line 116: is there a reason for handling profile before the rest of the records? & in a different tx as well?

packages/bsky/src/services/indexing/index.ts
```diff
@@ -93,6 +103,57 @@ export class IndexingService {
...
 
+  async indexRepo(did: string, commit: string) {
+    const now = new Date().toISOString()
+    const { pds, signingKey } = await this.didResolver.resolveAtpData(did)
+    const { sync } = new AtpAgent({ service: pds }).api.com.atproto
+
+    const { data: car } = await sync.getCheckout({ did, commit })
+    const { root, blocks } = await readCarWithRoot(new Uint8Array(car))
+    const storage = new MemoryBlockstore(blocks)
+    const repo = await verifyCheckoutToRepo(storage, root, did, signingKey)
```
dholms, line 114: It kinda seems like the previously implemented `verifyCheckout` handles everything that you're doing here. Any particular reason for creating the new function & reading this as a repo instead?

packages/bsky/src/services/indexing/index.ts
devinivy: It's really just a UX thing.  When we start reindexing the user their data will disappear then gradually flow back in.  I figure it would be friendly to ensure that their profile is immediately up to date since it's sort of the "face" of that user.

packages/bsky/src/services/indexing/index.ts
devinivy: What I was missing there was the CIDs—was just avoiding recomputing them.

packages/repo/src/verify.ts
devinivy: Ah, good point!

packages/bsky/src/services/indexing/index.ts
```diff
@@ -93,6 +103,57 @@ export class IndexingService {
...
+
+    // First, replace the user with just their current profile if it exists
+    await this.db.transaction(async (tx) => {
+      const indexingTx = new IndexingService(tx, this.didResolver)
+      await indexingTx.deleteForUser(did)
```
dholms, line 119: I wonder if it could be nice to diff the current indexed records against the new ones so that we're not doing a mass delete + mass add 🤔 

this will also update all the `indexedAt` times for every record - even if it wasn't actually mutated during this change

but i can see the headaches & data load of that back and forth as well 🤔

packages/bsky/src/services/indexing/index.ts
dholms: ah right yup that makes sense 👍

packages/bsky/src/services/indexing/index.ts
dholms: oh got it got it. yup that's fine. but also maybe we could just update `verifyCheckout` to give back cids as well 👀 

those are utility methods meant for this exact case. i wrote them in isolation so wasn't quite sure what was needed so we should make them match the common use

could do an "in between" `verifyCheckoutCids` which returns a map similar to `RepoContents` but with cids at the leaves

packages/bsky/src/services/indexing/index.ts
devinivy: I expect this to get more sophisticated over time, but this should also be rare/discouraged so I figured I'd just go with the simplest implementation to start.  In the worst case scenarios that a repo was truncated or totally replaced this will perform the best, it's the awkward cases where someone actually performed 500 ops that this will be suboptimal.

Re: the `indexedAt` times, it's a good point and I observed this when testing it out.  IMO it's acceptable, aside from the fact that we often sort on `indexedAt`.  My plan is to implement Paul's strategy in the indices, keeping around a sort field based on the combination of of `createdAt` and `indexedAt`.

dholms: kk cool that should be fine 👌 

yup this will just need to go along with a bit of extra logic around times 👍

devinivy:
@dholms this is good for another review.  This is still marked a draft, but in general I consider this work complete—just need to route the right messages to `indexRepo()` once #658 lands, which should just be a couple lines.

devinivy:
@dholms this is now fully up to date with lex changes, so it handles rebases and too-big commits.

packages/bsky/src/db/migrations/20230309T045948368Z-init.ts
```diff
@@ -63,6 +63,12 @@ export async function up(db: Kysely<unknown>): Promise<void> {
     .addColumn('replyParentCid', 'varchar')
     .addColumn('createdAt', 'varchar', (col) => col.notNull())
     .addColumn('indexedAt', 'varchar', (col) => col.notNull())
+    .addColumn('sortAt', 'varchar', (col) =>
+      col
+        .generatedAlwaysAs(sql`least("createdAt", "indexedAt")`)
```
dholms, line 68: 🆒

packages/bsky/src/db/migrations/20230309T045948368Z-init.ts
```diff
@@ -199,6 +221,7 @@ export async function up(db: Kysely<unknown>): Promise<void> {
     .addColumn('did', 'varchar', (col) => col.primaryKey())
     .addColumn('handle', 'varchar', (col) => col.unique())
     .addColumn('indexedAt', 'varchar', (col) => col.notNull())
+    .addColumn('commitDataCid', 'varchar')
```
dholms, line 224: possibility we want this on a different table since it'll be much higher churn than the rest of the table 🤔

packages/bsky/src/services/indexing/index.ts
```diff
@@ -93,6 +103,81 @@ export class IndexingService {
...
+        )
+      }
+    })
+
+    // Then iterate over all records and index them in batches
+    const contentList = [...walkContentsWithCids(checkout.contents)]
+    const chunks = chunkArray(contentList, 100)
+
+    for (const chunk of chunks) {
+      await this.db.transaction(async (tx) => {
```
dholms, line 145: sorry to come back to this, but what's the recovery path if this part fails?

the user is then left with no data except their profile, right?

dholms: a bit worried that a bad commit can blow up a user's account? would be better for it just to not update I think

packages/bsky/src/subscription/repo.ts
```diff
@@ -124,8 +131,16 @@ export class RepoSubscription {
   // @TODO handle too-big commits and rebases
   private async handleCommit(msg: message.Commit) {
     const { db, services } = this.ctx
-    const processOps = async (tx: Database, ops: PreparedWrite[]) => {
-      const indexingTx = services.indexing(tx)
+    const { root, rootCid, ops } = await getOps(msg)
+    const indexRecords = async (indexingTx: IndexingService) => {
```
dholms, line 135: nice

packages/bsky/src/services/indexing/index.ts
dholms: oh wait are these nested txs? if so, mind walking me through the different cases here? I'm not super versed in nested txs

dholms (approved):
Looks good! This shaped up nicely.

Just the concerns about: transactionality of updates & high churn column on table

packages/bsky/src/util/retry.ts
```diff
@@ -0,0 +1,69 @@
...
+
+// Adds randomness +/-15% of value
+function jitter(value: number) {
+  const delta = value * 0.15
+  return value + randomRange(-delta, delta)
+}
+
+function randomRange(from: number, to: number) {
+  const rand = Math.random() * (to - from)
+  return rand - from
```
devinivy, line 68:
```suggestion
  return rand + from
```

packages/bsky/src/services/indexing/index.ts
devinivy: Cool cool!

To get more concrete—the issue I'm picturing with processing partial writes within a commit is that you might butt into conflicts w/ application semantics.  Switching a downvote to an upvote, for example, might involve removing a downvote and creating an upvote, and doing half of that process isn't going to land you in a consistent state.

This same issue seems much less immediate with processing a full checkout, which consists of only creates, rather than updates/deletes which are far more likely to pair with other ops that are expected to be applied in tandem.

dholms: It's kinda tricky tho cause at least in that case, I think I would prefer for my downvote to go away even if my upvote is not created. Similarly if I try to batch a bunch of creates & deletes in a single commit as we did during the app migrations. 

That transactionality def seems important to me at the data layer, but I wonder if at the view layer it's best to "just get as close as possible".

I can think of cases where you may prefer transactionality at the view layer, but I can think of more where you'd prefer lenient handling 

Idk it's a tricky question fer sure

packages/bsky/src/db/migrations/20230309T045948368Z-init.ts
devinivy: I peeled this off, tracking additional sync state for the actor:
 - Commit cid so that we can track what commit they should be at, potentially reindex their current state from scratch or perform debugging.
 - Tally of total too-big commits and rebases, to get a sense for if someone is performing a ton of these operations.

packages/bsky/src/services/indexing/index.ts
devinivy: I tidied this area up and added some logic to skip over invalid records when doing a full repo sync.

PR #727, opened 2023-03-28
Add exports to api package: blobref & lex/json converters

pfrazee:
These exports are needed for writing a custom fetch handler and generally if you're manipulating blob refs. I also updated the api docs for the other ~poor saps~ engineers using react native.


dholms (approved):
Yup this all looks great 👍

PR #728, opened 2023-03-28
Lexicon: fix lingering references to recordRef

devinivy:
In the lex refactor's "big renaming" `com.atproto.repo.recordRef` was removed in favor of `strongRef`, but was still kicking around in the pds admin/moderation tooling.  This fixes that.  Also one tiny change to an unrelated utility to make some type checking happy.


dholms (approved):
nice this looks good 👍

PR #729, opened 2023-03-28
Lexicon: facet features

devinivy:
Two changes to richtext facets:
 - Each facet can have multiple "features", rather than a single "value."
 - The slice within the facet now more clearly deals with utf8 bytes rather than e.g. javascript string indexes.


dholms (approved):
nice this looks great to me 👍

PR #730, opened 2023-03-28
Repo rebases

dholms:
This implements rebases in both the repository library & on the PDS.

Note: this does not yet add a rebase XRPC method to the PDS as it can be a costly operation. However, we will likely want to support that at some point.

We also only currently support the simplest case of rebase: no operations, just removing history

For the PDS, the rebase process is:
- wipe the existing bookkeeping (`repo_commit_block` & `repo_commit_history`)
- add the new repo block
- walk the repo & get a list of all cids that are still in it (including the new root block)
- reindex all of those in `repo_commit_block` & `repo_commit_history` 
- left join `ipld_block` with `repo_commit_block` & delete all blocks from `ipld_block` that do not have a related `repo_commit_block`

After that we:
- assign all existing blobs to be associated with the new root
- emit a `seq` event for the rebase, containing _only_ the new root block
- invalidate all previous seqs for that repo


packages/pds/src/services/repo/blobs.ts
```diff
@@ -124,6 +124,14 @@ export class RepoBlobs {
       .execute()
   }
 
+  async processRebaseBlobs(did: string, newRoot: CID) {
```
devinivy, line 127: Maybe a question for another day, but I think as-is blobs on deleted records will stick around, including after a rebase.  So there could be blobs associated with the rebase commit, but no record present in that commit referencing them.  Hmmm!

packages/pds/src/sql-repo-storage.ts
```diff
@@ -139,18 +146,68 @@ export class SqlRepoStorage extends RepoStorage {
...
+    const allCids = [...rebase.preservedCids, ...rebase.blocks.cids()]
+    await this.indexCommitCids([
+      { commit: rebase.commit, prev: null, cids: allCids },
+    ])
+    await this.db.db
+      .deleteFrom('ipld_block')
+      .where(
+        'cid',
+        'in',
+        this.db.db
```
devinivy, line 172: Do we want this to also be constrained by `did`?
```suggestion
      .deleteFrom('ipld_block')
      .where('creator', '=', this.did)
      .where(
        'cid',
        'in',
        this.db.db
```

devinivy: Ohh I see, this is just sort of general cleanup that runs 👍

packages/pds/src/sql-repo-storage.ts
```diff
@@ -139,18 +146,68 @@ export class SqlRepoStorage extends RepoStorage {
...
+          .leftJoin('repo_commit_block', (join) =>
+            join
+              .onRef('block.creator', '=', 'repo_commit_block.creator')
+              .onRef('block.cid', '=', 'repo_commit_block.block'),
+          )
+          .where('repo_commit_block.creator', 'is', null)
+          .select('block.cid'),
+      )
+      .execute(),
+      await this.updateHead(rebase.commit, null)
```
devinivy, line 183:
```suggestion
      .execute()
    await this.updateHead(rebase.commit, null)
```

devinivy (approved):
💎 looks solid!

packages/repo/src/types.ts
```diff
@@ -118,23 +132,3 @@ export type RecordClaim = {
...
-  add(key: string, value: CID): Promise<DataStore>
-  update(key: string, value: CID): Promise<DataStore>
-  delete(key: string): Promise<DataStore>
-  get(key: string): Promise<CID | null>
-  list(count?: number, after?: string, before?: string): Promise<DataValue[]>
-  listWithPrefix(prefix: string, count?: number): Promise<DataValue[]>
-  getUnstoredBlocks(): Promise<{ root: CID; blocks: BlockMap }>
-  writeToCarStream(car: BlockWriter): Promise<void>
-  cidsForPath(key: string): Promise<CID[]>
-}
```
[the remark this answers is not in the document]
dholms: Yeah that was added in a certain frame of mind. But it's only been adding friction recently & it's not clear when/if we'll be supporting other datastores 👌

packages/pds/src/services/repo/blobs.ts
dholms: ah that's a good point 🤔 

alright I guess I'll probably have to list repo-blobs then

packages/pds/src/sql-repo-storage.ts
dholms: altho actually youre right because we don't have an index on just `cid`

devinivy (approved):

PR #732, opened 2023-03-29
Lexicon: optimize lex de/serialization

devinivy:
We were getting an unfriendly slowdown in the lex-refactor, and traced it down pretty exclusively to the new de/serialization processing.  By moving away from our `check.is()` system to simple checks wherever possible during de/serialization, we get a big speedup.

### DIFF
--- a/packages/common-web/src/ipld.ts
+++ b/packages/common-web/src/ipld.ts
@@ -19,7 +19,7 @@ export type IpldValue =
   | { [key: string]: IpldValue }
 
 // @NOTE avoiding use of check.is() here only because it makes
-// these implementations slow, and they often live it hot paths.
+// these implementations slow, and they often live in hot paths.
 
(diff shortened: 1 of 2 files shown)


packages/common-web/src/ipld.ts
```diff
@@ -20,65 +18,60 @@ export type IpldValue =
...
-
-const dagJsonBytes = z
-  .object({
-    $bytes: z.string(),
-  })
-  .strict()
-
-const dagJsonVal = z.union([dagJsonCid, dagJsonBytes])
+// @NOTE avoiding use of check.is() here only because it makes
+// these implementations slow, and they often live it hot paths.
```
devinivy, line 22:
```suggestion
// these implementations slow, and they often live in hot paths.
```

dholms (approved):
nice this looks excellent 🙌

PR #733, opened 2023-03-29
Lexicon: app migration to tidy notifications

devinivy:
Three things going on here:
 - Remove notifications for trends and assertions (i.e. old scene notifs).
 - Update record uris in notifications for the change from votes to likes.
 - Update record cids in notifications for likes and follows.


dholms (approved):
looks excellent 👍

PR #734, opened 2023-03-29
IPLD test vectors

dholms:
This adds test vectors for converting between JSON <-> IPLD <-> CBOR -> CID

As support, it also adds a new method for checking equality of two IPLD objects.


dholms:
Yeah honestly I'm not even sure how necessary the IPLD check is in there. 

If you can go JSON <-> CBOR just fine then your IPLD should be coming out just fine as well.

Those vectors were honestly mostly for readability

bnewbold:
Something with `$blob` would be good, given that we handle it specially.

Some emoji/unicode in strings would probably be good for use in test vectors, but not a big priority (I think we have coverage of that elsewhere).

The last type that we sort-of-not-really support is `float`. I'm tempted to say we should include a test vector with a float in there just to demonstrate that it is a bad idea or hard to repro?

bnewbold:
In particular, if you do a `$blob` example, take care to use a CID which is of "raw" multicodec, not `dag-cbor` multicodec like the existing `$link`s.

dholms:
kk added some tests for unicode & blob-refs with a "raw" multicodec.

I also added a test for floats altho we are not yet handling the "float that looks like an int" case. That is a "please steer away from doing this unless you _really_ know what you're doing" type of situation

PR #735, opened 2023-03-29
Lexicon: notification reason fix

devinivy:
The app migration from #733 was intended to update the notification `reason` in addition to the record uris, for like notifications.


dholms (approved):
hot dog 🌭

PR #737, opened 2023-03-30
Test vectors for secp and p256 signature verification

devinivy:
Just throwing up some test vectors to help us dummy-check interop with other implementations.

### DIFF
--- a/packages/crypto/src/multibase.ts
+++ b/packages/crypto/src/multibase.ts
@@ -1,4 +1,5 @@
 import * as uint8arrays from 'uint8arrays'
+import { SupportedEncodings } from 'uint8arrays/to-string'
 
 export const multibaseToBytes = (mb: string): Uint8Array => {
   const base = mb[0]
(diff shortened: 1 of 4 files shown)


packages/crypto/src/multibase.ts
```diff
@@ -24,3 +25,29 @@ export const multibaseToBytes = (mb: string): Uint8Array => {
       throw new Error(`Unsupported multibase: :${mb}`)
   }
 }
+
+export const bytesToMultibase = (
```
dholms, line 29: this package already relies on `multiformats` which has these methods implemented. so we could just export those. not sure why i didn't do so for `multibaseToBytes` 🤔

packages/crypto/src/secp256k1/operations.ts
```diff
@@ -14,3 +14,12 @@ export const verifyDidSig = async (
   const msgHash = await secp.utils.sha256(data)
   return secp.verify(sig, msgHash, keyBytes)
 }
+
+export const verify = async (
```
dholms, line 18: could have `verifyDidSig` use this fn

packages/crypto/tests/signatures.test.ts
```diff
@@ -0,0 +1,100 @@
...
+          signatureBytes,
+        )
+        expect(verified).toEqual(true)
+      } else if (vector.algorithm === SECP256K1_JWT_ALG) {
+        const verified = await secp.verify(
+          keyBytes,
+          messageBytes,
+          signatureBytes,
+        )
+        expect(verified).toEqual(true)
```
dholms, line 51: might be worth adding a bad signature or two in here as well

like just a signature for another curve for instance

PR #742, opened 2023-03-31
Use TypedArrays instead of ArrayBuffers with WebCrypto

dholms:
The fundamental problem here that we were passing in an underlying ArrayBuffer from the cbor decoding library to webcrypto instead of the typed array.

For context: A TypedArray (Uint8Array or Buffer) in node is a _view_ into an underlying ArrayBuffer. The cbor encoding library must use ArrayBuffers under the hood & then pass back a slice of that ArrayBuffer as a Buffer. We then passed in the underlying ArrayBuffer which was larger than the actual data itself, thereby signing the wrong data.

The cbor decoding library would use an ArrayBuffer of size 128 as long as the data payload was <=128 & then 8kb if larger. A commit without a `prev` is <128 & a commit with a `prev` is > 128, thus why this was only happening on subsequent commits.

We were not running into this in the past because our commit objects were smaller (before the repov2 rewrite).

WebCrypto used to required ArrayBuffers to be passed in (it would not handle views), but it now allows either to be used. I thought I had updated all uses of this, but missed the one on `sign`. I've changed that to not pass in the underlying buffer,  wrapped every webcrypto input with `new UInt8Array` to ensure that we're not passing in any underlying arraybuffers and added some tests.

### DIFF
--- a/packages/crypto/src/p256/operations.ts
+++ b/packages/crypto/src/p256/operations.ts
@@ -46,8 +46,8 @@ export const verify = async (
   return webcrypto.subtle.verify(
     { name: 'ECDSA', hash: { name: 'SHA-256' } },
     importedKey,
-    sig,
-    data,
+    new Uint8Array(sig),
+    new Uint8Array(data),
   )
 }
 
@@ -56,7 +56,7 @@ export const importEcdsaPublicKey = async (
 ): Promise<CryptoKey> => {
   return webcrypto.subtle.importKey(
     'raw',
-    keyBytes,
+    new Uint8Array(keyBytes),
     { name: 'ECDSA', namedCurve: 'P-256' },
     true,
     ['verify'],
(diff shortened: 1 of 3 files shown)


packages/crypto/src/p256/operations.ts
```diff
@@ -46,8 +46,8 @@ export const verify = async (
   return webcrypto.subtle.verify(
     { name: 'ECDSA', hash: { name: 'SHA-256' } },
     importedKey,
-    sig,
-    data,
+    new Uint8Array(sig),
+    new Uint8Array(data),
```
devinivy, line 50: Not sure I totally get why these are needed. Or is the idea to be a bit defensive here?

devinivy (approved):
🙏 glad this got rooted right out!

packages/crypto/src/p256/operations.ts
dholms: Yeah just being extra defensive. it's probably not necessary. But the different types of binary representations freak me out enough that I just decided to wrap it

PR #744, opened 2023-04-01
Bsky indexing non-repo updates

devinivy:
This handles non-repo updates from the `subscribeRepo` xrpc endpoint in the app view.
 - Handles are revalidated on a handle-updated message.
 - Actors are unindexed on a tombstone message if their did no longer resolves.
 - No-op on a migration message, since app view doesn't need to take any action here.

I made a change in the did resolver so that `notFound` docs are always positively not found, rather than possibly not being found due to some kind of network or service error (though this is a bit fuzzier for did-web).

We still need to deal with the case that a did's handle stops resolving, but it's not 100% settled yet how this should present itself in `getProfile`, etc.


packages/did-resolver/src/web-resolver.ts
```diff
@@ -48,10 +48,13 @@ export const makeResolver = (opts: WebResolverOptions): DIDResolver => {
...
     } catch (err) {
-      return errors.notFound()
+      if (err instanceof AxiosError && err.response) {
+        return errors.notFound() // Positively not found, versus due to e.g. network error
+      }
+      throw err
     }
 
-    // TODO: this excludes the use of query params
+    // @TODO: this excludes the use of query params
```
dholms, line 57: bless 🙏

packages/bsky/src/subscription/repo.ts
```diff
@@ -118,17 +118,16 @@ export class RepoSubscription {
       await this.handleCommit(msg)
     } else if (message.isHandle(msg)) {
       await this.handleUpdateHandle(msg)
-    } else if (message.isMigrate(msg)) {
-      await this.handleMigrate(msg)
     } else if (message.isTombstone(msg)) {
       await this.handleTombstone(msg)
+    } else if (message.isMigrate(msg)) {
+      // Ignore migrations
```
dholms, line 124: yeah tbh im not sure appview will ever need to handle migrations. more of a bgs thing

dholms (approved):
looks great 🙌 🙇

PR #745, opened 2023-04-02
lex refactor hot fixes

dholms:


devinivy:
Merging this just to avoid some tricky coordination.  There will be one bit of followup to remove the profile display name and description truncation behavior soon.

PR #749, opened 2023-04-03
Update reason type in `feed.ts`

christiandeange:
Missed a spot when refactoring feed definitions.
The old reference no longer exists.


devinivy (approved):
Good catch, thanks!

PR #751, opened 2023-04-03
Paginate listRecords based on cursor

dholms:
Switches pagination in `listRecords` from `rkeyStart` & `rkeyEnd` to `cursor`.

This makes the semantics match other pagination routes. rkeyStart/End were not clear & leading to uncertainty. They would only be used for requesting some internal range of rkeys which just does not seem like a common usecase.

I've left those parameters around for now but marked them deprecated. I propose we switch the client over, wait a week or two & then actually remove them.

### DIFF
--- a/lexicons/com/atproto/repo/listRecords.json
+++ b/lexicons/com/atproto/repo/listRecords.json
@@ -12,8 +12,9 @@
           "repo": {"type": "string", "format": "at-identifier", "description": "The handle or DID of the repo."},
           "collection": {"type": "string", "format": "nsid", "description": "The NSID of the record type."},
           "limit": {"type": "integer", "minimum": 1, "maximum": 100, "default": 50, "description": "The number of records to return."},
-          "rkeyStart": {"type": "string", "description": "The lowest sort-ordered rkey to start from (exclusive)"},
-          "rkeyEnd": {"type": "string", "description": "The highest sort-ordered rkey to stop at (exclusive)"},
+          "cursor": {"type": "string"},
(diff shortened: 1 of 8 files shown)


lexicons/com/atproto/repo/listRecords.json
```diff
@@ -12,8 +12,9 @@
           "repo": {"type": "string", "format": "at-identifier", "description": "The handle or DID of the repo."},
           "collection": {"type": "string", "format": "nsid", "description": "The NSID of the record type."},
           "limit": {"type": "integer", "minimum": 1, "maximum": 100, "default": 50, "description": "The number of records to return."},
-          "rkeyStart": {"type": "string", "description": "The lowest sort-ordered rkey to start from (exclusive)"},
-          "rkeyEnd": {"type": "string", "description": "The highest sort-ordered rkey to stop at (exclusive)"},
+          "cursor": {"type": "string"},
+          "rkeyStart": {"type": "string", "description": "DEPRECATED: The lowest sort-ordered rkey to start from (exclusive)"},
+          "rkeyEnd": {"type": "string", "description": "DEPRECATED: The highest sort-ordered rkey to stop at (exclusive)"},
```
pfrazee, line 17: Did you mean to put deprecated on these?

dholms: yup sorry didn't get to write a descript cause i hopped on a call. my plan was to stop supporting these entirely here in the next week or so once you switch over to `cursor`

PR #752, opened 2023-04-03
Prep bsky app view for unauthed usage

devinivy:
A few separate things going on here to prep the app view for public usage:
 - Allow all endpoints except `getTimeline` to be accessed without auth.
 - Support CORS for cases where a web client talks directly to the app view without going through a PDS.
 - Tidy/cull the entrypoint and configuration for the app view.
 - Some fixes to the dockerfile.
 - Add a workflow for building the docker image and pushing to AWS.
 - Allow running bsky app view without running repo subscription.


packages/bsky/Dockerfile
```diff
@@ -6,12 +6,11 @@ COPY ./*.* ./
 # NOTE bsky's transitive dependencies go here: if that changes, this needs to be updated.
 COPY ./packages/bsky ./packages/bsky
 COPY ./packages/api ./packages/api
-COPY ./packages/aws ./packages/aws
```
dholms, line 9: we might need cloudfront in here for doing invalidations 🤔

dholms: but i guess we can get that in when we get to blob takedowns in appview

dholms (approved):
looks great 💪

packages/bsky/Dockerfile
devinivy: Yeah, I think we'll pull it back in once there are codepaths using it.

PR #756, opened 2023-04-04
Include email on server.getSession

dholms:
Left as optional because services may support other identifiers (phone number for instance)


devinivy (approved):
Looks good!

pfrazee:
Looks like some more tests need updating (?) but lgtm, lmk when you can push to prod!

PR #760, opened 2023-04-04
Add email to createSession

pfrazee:
This PR:

- Adds an optional `email` to `com.atproto.server.createSession` response
- Updates the api package to store email in the session data it tracks


dholms (approved):
yup yup 👍

PR #764, opened 2023-04-05
Improve admin invite code views

dholms:
A few things:
- add filter by `invitedBy` on search
  - this only works on a _blank_ term, otherwise we were having to jam a bunch of invite logic into our search functionality which didn't feel great. & also doesn't seem like a common usecase
- add some tests for admin views
- makes searchRepos only return `invitedBy` & getRepo return `invitedBy` & `invites`. since `invites` is a larger value, we may not want to run it on every search


pfrazee:
👍 LGTM

devinivy (approved):
Great 👍 !

PR #770, opened 2023-04-06
Feed item index

devinivy:
Adds a new `feed_item` table from which `getTimeline` and `getAuthorFeed` can be more easily built.  It essentially combines relevant data from the `post` and `repost` tables to aid serving feeds.


packages/pds/src/db/migrations/20230406T185855842Z-feed-item-init.ts
```diff
@@ -0,0 +1,123 @@
...
+    .addColumn('uri', 'varchar', (col) => col.primaryKey())
+    .addColumn('cid', 'varchar', (col) => col.notNull())
+    .addColumn('type', 'varchar', (col) => col.notNull())
+    .addColumn('postUri', 'varchar', (col) => col.notNull())
+    .addColumn('postAuthorDid', 'varchar', (col) => col.notNull())
+    .addColumn('originatorDid', 'varchar', (col) => col.notNull())
+    .addColumn('sortAt', 'varchar', (col) => col.notNull())
+    .execute()
+  // Fill table. This query should be safe to be run any time to update the index.
+  await db
```
dholms, line 17: I'm wondering if we shouldn't just run this whole thing in a separate app migration 🤔

packages/pds/src/app-view/db/tables/feed-item.ts
```diff
@@ -0,0 +1,13 @@
+export const tableName = 'feed_item'
+
+export interface FeedItem {
+  uri: string
+  cid: string
+  type: 'post' | 'repost'
+  postUri: string
+  postAuthorDid: string
```
dholms, line 8: I think you could drop this column since - like replyRoot/replyParent - you can grab this off the join 🤔

packages/pds/src/app-view/services/indexing/plugins/post.ts
```diff
@@ -52,15 +52,32 @@ const insertFn = async (
...
-    .returningAll()
-    .executeTakeFirst()
+  const [insertedPost] = await Promise.all([
+    db
+      .insertInto('post')
+      .values(post)
+      .onConflict((oc) => oc.doNothing())
+      .returningAll()
+      .executeTakeFirst(),
+    await db
```
dholms, line 62: can delete this await

packages/pds/src/app-view/services/indexing/plugins/post.ts
```diff
@@ -52,15 +52,32 @@ const insertFn = async (
...
+      .insertInto('feed_item')
+      .values({
+        type: 'post',
+        uri: post.uri,
+        cid: post.cid,
+        postUri: post.uri,
+        postAuthorDid: post.creator,
+        originatorDid: post.creator,
+        sortAt:
+          post.indexedAt < post.createdAt ? post.indexedAt : post.createdAt,
```
dholms, line 72: sweet 👌

packages/pds/tests/event-stream/__snapshots__/sync.test.ts.snap
```diff
@@ -186,14 +186,28 @@ Array [
       "uri": "record(2)",
       "viewer": Object {},
     },
+    "reason": Object {
```
dholms, line 189: did these snaps change so much cause of the new `sortBy`?

dholms (approved):
looks good 👍 

couple small things

I think it might be good to talk through the deploy strategy here a bit before lettin er rip

packages/pds/src/app-view/db/tables/feed-item.ts
devinivy: One reason to keep it would be if we want to put any indexes on it for building feeds, but I think where things landed it didn't require any, so seems all good to remove 👍

packages/pds/tests/event-stream/__snapshots__/sync.test.ts.snap
devinivy: Yeah, that's right.  I think the current behavior is more correct because the feed usually shouldn't be reordered based on the updated `indexedAt` time (at least when folks are playing nice with their self-reported `createdAt` times).

devinivy:
Closing, as this has been split into #773 and #774.  But this currently contains the full set of changes, so may be useful to look here for recent changes that have taken place (i.e. since the last review).

PR #775, opened 2023-04-07
Fix AtUri parse on hostnames that start with numbers

pfrazee:
Users with numbers in their handles are getting cryptic "Error: uri must be a valid at-uri" messages when the app tries to load their threads. Turns out I botched the grammar for hostnames here.


devinivy (approved):

pfrazee:
Published `@atproto/uri@0.0.2`

PR #778, opened 2023-04-08
Disable output validation when not in debug mode

devinivy:
Output validation generally isn't needed outside of debug mode.  We may want to consider validating a sample output in the future when not in debug mode.


dholms (approved):
great call 👏

PR #784, opened 2023-04-10
Sync/listRepos

dholms:
Adds an enumeration route to list repos hosted on a service, including did & current head.

Paginates by account creation time

### DIFF
--- a/packages/pds/src/api/com/atproto/sync/listHostedRepos.ts
+++ b/packages/pds/src/api/com/atproto/sync/listHostedRepos.ts
@@ -1,0 +1,60 @@
+import { InvalidRequestError } from '@atproto/xrpc-server'
+import { Server } from '../../../../lexicon'
+import AppContext from '../../../../context'
+import { Cursor, GenericKeyset, paginate } from '../../../../db/pagination'
+
+export default function (server: Server, ctx: AppContext) {
+  server.com.atproto.sync.listHostedRepos(async ({ params }) => {
+    const { limit, cursor } = params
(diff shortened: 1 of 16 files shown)


packages/pds/src/api/com/atproto/sync/listHostedRepos.ts
```diff
@@ -0,0 +1,61 @@
...
+      .selectFrom('repo_root')
+      .innerJoin('user_account', 'user_account.did', 'repo_root.did')
+      .where('repo_root.takedownId', 'is', null)
+      .select([
+        'repo_root.did as did',
+        'repo_root.root as head',
+        'user_account.createdAt as createdAt',
+      ])
+    let builder = ctx.db.db.selectFrom(innerBuilder.as('repos')).selectAll()
+    const keyset = new TimeDidKeyset(ref('createdAt'), ref('did'))
```
devinivy, line 20: We probably don't need to have a wrapping `select *` here, believe it should just involve qualifying the keyset columns, e.g. `createdAt -> user_account.createdAt` and `did -> user_account.did`.

devinivy: If there's ever an index placed on `user_account` to support this pagination, we may want to check whether selecting from `user_account` and joining over to `repo_root` is smoother for the db.  I am pretty sure postgres figures this out, though, as long as both sort columns are on the same table.

devinivy (approved):
Had some misc query thoughts, but this looks great.

PR #785, opened 2023-04-10
Synchronous notifs

dholms:
Remove notifs from the message queue on the PDS & do them synchronously within the db tx

This allowed us to remove the SQL MessageQueue abstraction

Also cleaned up all the app migrations from the lex refactor, don't think we need them in here anymore


packages/pds/src/app-view/services/indexing/processor.ts
```diff
@@ -180,7 +174,29 @@ export class RecordProcessor<T, S> {
...
+  }
+
+  async handleNotifs(op: { deleted?: S; inserted?: S }) {
+    let notifs: UserNotification[] = []
+    if (op.deleted) {
+      const forDelete = this.params.notifsForDelete(
+        op.deleted,
+        op.inserted ?? null,
+      )
+      if (forDelete.toDelete) {
```
devinivy, line 188: May just protect us from ourselves, when we eventually return a `toDelete: []` somewhere.
```suggestion
      if (forDelete.toDelete?.length) {
```

packages/pds/src/context.ts
```diff
@@ -82,10 +80,6 @@ export class AppContext {
     return this.opts.services
   }
 
-  get messageQueue(): MessageQueue {
-    return this.opts.messageQueue
-  }
```
devinivy, line 87: 👋

packages/pds/src/db/migrations/20230411T175730759Z-drop-message-queue.ts
```diff
@@ -0,0 +1,32 @@
+import { Kysely } from 'kysely'
+import { Dialect } from '..'
+
+const messageQueueTable = 'message_queue'
+const messageQueueCursorTable = 'message_queue_cursor'
+
+export async function up(db: Kysely<unknown>): Promise<void> {
```
devinivy, line 7: We spoke about potentially wanting to split off this migration into a separate PR.

devinivy (approved):
A couple small notes, but this looks great.  Main consideration is just whether to split out the migration. 💎

PR #788, opened 2023-04-11
Lex: additional report types

bnewbold:
Also added basic descriptions, not sure if we actually want those or should pull them out.


dholms (approved):
LGTM 👍

PR #792, opened 2023-04-12
Adjustable db pool size

dholms:
Add env var for adjustable postgres pool size: rec trying out 20 first, default is 10

### DIFF
--- a/packages/pds/service/index.js
+++ b/packages/pds/service/index.js
@@ -32,6 +32,7 @@ const main = async () => {
   const db = Database.postgres({
     url: pgUrl(env.dbCreds),
     schema: env.dbSchema,
+    poolSize: env.dbPoolSize
   })
   const s3Blobstore = new S3BlobStore({ bucket: env.s3Bucket })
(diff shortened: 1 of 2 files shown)


packages/pds/service/index.js
```diff
@@ -32,6 +32,7 @@ const main = async () => {
   const db = Database.postgres({
     url: pgUrl(env.dbCreds),
     schema: env.dbSchema,
+    poolSize: env.dbPoolSize
```
devinivy, line 35:
```suggestion
    poolSize: env.dbPoolSize,
```

packages/pds/service/index.js
```diff
@@ -93,6 +94,7 @@ const getEnv = () => ({
   dbCreds: JSON.parse(process.env.DB_CREDS_JSON),
   dbMigrateCreds: JSON.parse(process.env.DB_MIGRATE_CREDS_JSON),
   dbSchema: process.env.DB_SCHEMA,
+  dbPoolSize: process.env.DB_POOL_SIZE,
```
devinivy, line 97: We might want to parse this into a number when it's present.

devinivy (approved):

PR #805, opened 2023-04-12
Label profiles in feeds

dholms:
Ensure that we add labels to profiles in feeds


packages/pds/src/app-view/services/feed/index.ts
```diff
@@ -124,6 +127,7 @@ export class FeedService {
             following: cur?.requesterFollowing || undefined,
             followedBy: cur?.requesterFollowedBy || undefined,
           },
+          labels: labels[cur.did],
```
devinivy, line 130:
```suggestion
          labels: labels[cur.did] ?? [],
```

devinivy (approved):

PR #811, opened 2023-04-13
Minor query optimization for getBlocksForCommits()

devinivy:
By using a values list for the `where in (...commits)` clause, postgres is able to do a hash lookup on the commit list.  This is useful for long lists, which is common in `getBlocksForCommits()` via the `com.atproto.sync.getRepo` method.


dholms (approved):
oh badass dude 😎

PR #812, opened 2023-04-13
Add admin.updateAccountEamil

dholms:
Holding off on adding in user email updates because there are some security issues to take into account there

### DIFF
--- a/packages/pds/src/services/account/index.ts
+++ b/packages/pds/src/services/account/index.ts
@@ -145,6 +145,17 @@ export class AccountService {
     await sequenceHandleUpdate(this.db, did, handle)
   }
 
+  async updateEmail(did: string, email: string) {
+    const res = await this.db.db
+      .updateTable('user_account')
+      .set({ email })
+      .where('did', '=', did)
+      .executeTakeFirst()
+    if (res.numUpdatedRows < 1) {
+      throw new UserAlreadyExistsError()
+    }
+  }
+
   async updateUserPassword(did: string, password: string) {
     const passwordScrypt = await scrypt.hash(password)
     await this.db.db
--- a/packages/pds/src/services/account/index.ts
+++ b/packages/pds/src/services/account/index.ts
@@ -146,14 +146,11 @@ export class AccountService {
   }
 
   async updateEmail(did: string, email: string) {
-    const res = await this.db.db
+    await this.db.db
       .updateTable('user_account')
-      .set({ email })
+      .set({ email: email.toLowerCase() })
       .where('did', '=', did)
       .executeTakeFirst()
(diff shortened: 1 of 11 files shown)


packages/pds/src/services/account/index.ts
```diff
@@ -145,6 +145,17 @@ export class AccountService {
...
   }
 
+  async updateEmail(did: string, email: string) {
+    const res = await this.db.db
+      .updateTable('user_account')
+      .set({ email })
+      .where('did', '=', did)
+      .executeTakeFirst()
+    if (res.numUpdatedRows < 1) {
+      throw new UserAlreadyExistsError()
```
devinivy, line 155: I think this error is probably something more like "user doesn't exist".  Maybe we can even get rid of the `getAccount()` check in the route.

packages/pds/src/services/account/index.ts
```diff
@@ -145,6 +145,17 @@ export class AccountService {
     await sequenceHandleUpdate(this.db, did, handle)
   }
 
+  async updateEmail(did: string, email: string) {
+    const res = await this.db.db
```
devinivy, line 149: Should we do some minimal form of email validation?  Just poked-around quickly at registration and didn't see any, though hmm 🤔

packages/pds/src/services/account/index.ts
```diff
@@ -145,6 +145,17 @@ export class AccountService {
...
+      .updateTable('user_account')
+      .set({ email })
```
devinivy, line 151: I think we may want to normalize the email to lowercase as we do elsewhere:
```suggestion
      .set({ email: email.toLowerCase() })
```

packages/pds/tests/account.test.ts
```diff
@@ -168,6 +168,36 @@ describe('account', () => {
     ])
   })
 
+  it('allows administrative email updates', async () => {
+    await agent.api.com.atproto.admin.updateAccountEmail(
+      {
+        account: handle,
+        email: 'alice-new@test.com',
```
devinivy, line 175: Could toss some caps in here to test the normalization if we add that to the backend.

packages/pds/src/services/account/index.ts
dholms: ah yup lol. i did a hack copy-paste job

packages/pds/src/services/account/index.ts
dholms: ah good call 👍

packages/pds/tests/account.test.ts
dholms: 👍

PR #819, opened 2023-04-15
Fix duplication of constants in the crypto package

MarshalX:
The "BASE58_DID_PREFIX" constant in `const.ts` was unused, but the same constant value was declared twice in `did.ts`


dholms (approved):
yup good find 👍  thanks!

PR #822, opened 2023-04-16
fix: add  for link facet.

susumuota:
To keep consistency with mention facet.


dholms:
Hey @pfrazee, looking at this I don't think we need the `$type` for either, since it's a ref & not a union. Correct?

dholms (changes requested):
I think we actually want the inverse of this & not send along the `$type` field on _either_ the link _or_ the mention

`$types` are necessary when differentiating between lexicons at runtime - namely, in records & in open unions. Since this is just a ref & not an open union, we do not need to include the `$type`

I'll hold off on the fix for now in case you still wanna get this in 😉 

If not, I can close & we'll get a fix in ourselves

susumuota:
@dholms Thanks! 

My point is to keep consistency between link and mention. 
I always prefer the simple way if something is not required. 😄

PR #826, opened 2023-04-17
App passwords

dholms:
Implement routes for:
- creating app passwords
- revoking app passwords
- listing app password names (but not passwords because they are scrypted)

Access tokens from app passwords have a different scope on them (which is a subset of the general "access") & the auth verifier for _most_ routes is augmented to allow requests from both scopes. Some routes - such as account deletion or creating new app passwords - do not allow requests from app passwords. 

At the DB layer, we scrypt app passwords using the user's did as a salt. This prevents us from having to scrypt a password (costly operation) many times for many different salts. This still protects against rainbow table attacks.


packages/pds/src/db/scrypt.ts
```diff
@@ -1,24 +1,50 @@
...
+}
+
+export const getDerivedHash = (
+  password: string,
+  salt: string,
+): Promise<string> => {
   return new Promise((resolve, reject) => {
-    const [salt, hash] = storedHash.split(':')
     crypto.scrypt(password, salt, 64, (err, derivedHash) => {
       if (err) reject(err)
```
devinivy, line 37:
```suggestion
      if (err) return reject(err)
```

packages/pds/src/db/tables/app-password.ts
```diff
@@ -0,0 +1,10 @@
+export interface AppPassword {
+  accountDid: string
```
devinivy, line 2: No issue here at all, but I'm interested to hear whether there's any special reason this is `accountDid` rather than just `did`, as you might find in some of the other tables.

packages/pds/src/auth.ts
```diff
@@ -105,13 +122,17 @@ export class ServerAuth {
...
+    scopes: AuthScope[],
+    options?: jwt.VerifyOptions,
+  ): jwt.JwtPayload {
     try {
       const payload = jwt.verify(token, this._secret, options)
       if (typeof payload === 'string' || 'signature' in payload) {
         throw new InvalidRequestError('Malformed token', 'InvalidToken')
       }
-      if (scope && payload.scope !== scope) {
+      if (scopes.length > 0 && !scopes.includes(payload.scope)) {
```
devinivy, line 135: I don't think it matters in any code we have kicking around, but I'm tempted to treat empty `scopes` as always invalid rather than always valid.

packages/pds/tests/app-passwords.test.ts
```diff
@@ -0,0 +1,103 @@
...
+        repo: appAgent.session?.did,
+      },
+      {
+        text: 'Testing testing',
+        createdAt: new Date().toISOString(),
+      },
+    )
+  })
+
+  it('restricts certain actions', async () => {
```
devinivy, line 67: Just double-checking that `createAppPassword` is the only method we want to restrict in this PR.

packages/pds/tests/app-passwords.test.ts
```diff
@@ -0,0 +1,103 @@
+import AtpAgent from '@atproto/api'
```
devinivy, line 1: One other case I can think of that might be nice in here is is that tokens received after a refresh continue to be restricted as app-password credentials.

pfrazee:
APIs lgtm from the app perspective

packages/pds/src/db/scrypt.ts
```diff
@@ -1,24 +1,50 @@
...
+}
+
+export const hashWithSalt = (
+  password: string,
+  salt: string,
+): Promise<string> => {
   return new Promise((resolve, reject) => {
-    const salt = crypto.randomBytes(16).toString('hex')
     crypto.scrypt(password, salt, 64, (err, hash) => {
       if (err) reject(err)
```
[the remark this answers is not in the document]
dholms: ah good looks

packages/pds/src/db/tables/app-password.ts
dholms: no good reason lol. i'll update it 👍

packages/pds/tests/app-passwords.test.ts
dholms: I think so? I can't think of any others besides maybe `revokeAppPassword`

Honestly I'm not even sure we need to restrict `createAppPassword` 🤔 

`deleteAccount` requires you to add in your account password anyways

packages/pds/tests/app-passwords.test.ts
dholms: ah good call 👍

packages/pds/tests/app-passwords.test.ts
devinivy: I think it's a good call to restrict `createAppPassword` 👍

devinivy (approved):

PR #827, opened 2023-04-17
Appview proxy

dholms:
Proxying requests from PDS to AppView

For all view requests, we forward the request to the configured appview, then hydrate mutes into the response before sending it back to the client.

---

Some implementation notes:

I've made the PDS binary either "everything or nothing proxied to appview". We can set this up per-user, but would require 
some extra user state kicking around & a DB hit at the start of each request. I'm not sure if we'll want to do this given we're going to be deploying & testing in the sandbox.

Each PDS only has _one_ configured AppView. In the future we can let a user set their preferred AppView.

I disabled the actor search tests since they kept timing out. In general, our bulk user data tests do not perform well & we probably need to do something about that (I _think_ that this is an scrypt thing & we might be able to fix by avoiding scrypt in dev envs). If you have other ideas plz let me know. 
That being said, I am concerned about the performance of `searchActorsTypeahead` being proxied to the appview. We _may_ want to handle that some other way...

For testing, I just copied over our view tests into a new `proxied` folder. This has led to a decent bit of duplicated logic/tests, but as these services come together & become more stable, we should be able to start cleaning it up. For now, since both paths are supported, I feel that it's important to test both paths. Let me know if there is anything you would do in the meantime.

For testing, I also switched the pds to default test with postgres. There is a new `yarn test:sqlite` that will use sqlite & not run the proxied tests (which rely on postgres because of appview)

---

Future work:
- simple service <> service auth instead of justing using the DID as a bearer token


packages/bsky/src/services/feed/index.ts
```diff
@@ -343,12 +343,10 @@ export class FeedService {
       repostCount: post.repostCount,
       likeCount: post.likeCount,
       indexedAt: post.indexedAt,
-      viewer: post.viewer
-        ? {
-            repost: post.requesterRepost ?? undefined,
-            like: post.requesterLike ?? undefined,
-          }
-        : undefined,
+      viewer: {
```
dholms, line 346: was there a reason for these changes? seems like it was always getting set to undefined

packages/pds/tests/proxied/__snapshots__/thread.test.ts.snap
```diff
@@ -0,0 +1,812 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
devinivy, line 1: Just curious—are the proxied snapshots pretty much identical to the non-proxied versions?

packages/pds/src/app-view/proxied/index.ts
```diff
@@ -0,0 +1,519 @@
...
+        body: profile,
+      }
+    },
+  })
+
+  server.app.bsky.actor.getProfiles({
+    auth: ctx.accessVerifier,
+    handler: async ({ auth, params }) => {
+      const requester = auth.credentials.did
+      const res = await agent.api.app.bsky.actor.getProfiles(
```
devinivy, line 49: What's the story around propagating errors— does that work relatively well?

devinivy (commented):
Approach lookin good to me!  Left a couple small thoughts/questions.  My general thought is that making proxying be close to a pass-through is a good look—the less machinery kickin around there the better.  (Of course that has its limits when we have to go rooting around the response anyway.)

Optional things we might consider in the future (i.e. not critical):
 - status codes / error messaging for when appview is not available.
 - could add some modest retrying.
 - passing-through errors from the appview pretty faithfully (e.g. rather than going through the XRPC error abstractions).

packages/pds/tests/proxied/__snapshots__/thread.test.ts.snap
dholms: pretty much. only 3 differences:
- there are fewer of them
- labels are not in them (yet)
- the img links changed

I was fairly rigorous in parsing through them to verify. but we may want to do so again once we have labels in the appview

packages/pds/src/app-view/proxied/index.ts
```diff
@@ -0,0 +1,519 @@
...
+  }
+
+  server.app.bsky.actor.getProfile({
+    auth: ctx.accessVerifier,
+    handler: async ({ auth, params }) => {
+      const requester = auth.credentials.did
+      const res = await agent.api.app.bsky.actor.getProfile(
+        params,
+        headers(requester),
+      )
```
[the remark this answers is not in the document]
dholms: yeah that's a good point... 

altho of course the types are pretty nice when working with the data 😅 

is there any additional complexity there if we do not trust the service? ie user-configured routes (don't need to address right away)

packages/pds/src/app-view/proxied/index.ts
dholms: yeah it actually works flawlessly. it's already the correct error class & everything

devinivy (approved):
Oh yeah!

PR #828, opened 2023-04-17
Stop sequencer listener in outbox when subscription aborts

devinivy:
When the xrpc subscription on `subscribeRepos` forcibly ends, we now stop the outbox's subscription to the event sequencer.


dholms (approved):
yup yup great find. that looks right 👌

PR #829, opened 2023-04-17
Notifications from the appview

devinivy:
The appview provides notifications statelessly (i.e. without tracking the last seen date).  This involves adding optional `sortAt` parameters to the notification methods.  Made some modest changes to the notification table as compared to the pds, in particular adding an auto-incrementing primary key and renaming the `userDid` column to `did`.


packages/bsky/src/api/app/bsky/notification/listNotifications.ts
```diff
@@ -0,0 +1,88 @@
...
+      )
+      notifBuilder = paginate(notifBuilder, {
+        cursor,
+        limit,
+        keyset,
+      })
+
+      const notifs = await notifBuilder.execute()
+
+      // @NOTE calling into app-view, will eventually be replaced
```
dholms, line 49: probs can delete this

devinivy:
```suggestion
```

dholms (approved):
this all LGTM 👍 

so for deduping updates, (ie getting the same commit w same like on the same record from two different sources) what would happen is that:
- first commit would create the notif for the like
- second commit would delete the existing notif & create a _new_ notif for the like

is that right?

devinivy:
In that case the one notification should actually just stick around.  Here's the logic for handling notifications on record creation: https://github.com/bluesky-social/atproto/blob/b69a6da033b03fbbf8bcb994f2557d781e921116/packages/bsky/src/services/indexing/processor.ts#L71-L81

If the record is created twice, then `inserted` there will be `null` since the `insertFn` for likes no-ops using an `on conflict do nothing`: https://github.com/bluesky-social/atproto/blob/b69a6da033b03fbbf8bcb994f2557d781e921116/packages/bsky/src/services/indexing/plugins/like.ts#L19-L33

PR #839, opened 2023-04-19
Do not give notifications for replies to ancestors that have been deleted 

dholms:
We were basing notifications for threads off of our post_hierarchy (ancestor) table.

However that structure stays around even after a post has been deleted.

Before sending a notification for a post to an ancestor, first ensure that that ancestor is still around

Changes made to both PDS & AppView

### DIFF
--- a/packages/bsky/src/services/indexing/plugins/post.ts
+++ b/packages/bsky/src/services/indexing/plugins/post.ts
@@ -21,11 +21,16 @@ type Post = Selectable<DatabaseSchemaType['post']>
 type PostEmbedImage = DatabaseSchemaType['post_embed_image']
 type PostEmbedExternal = DatabaseSchemaType['post_embed_external']
 type PostEmbedRecord = DatabaseSchemaType['post_embed_record']
+
+type Ancestor = PostHierarchy & {
+  doesNotExst?: boolean
+}
+
 type IndexedPost = {
   post: Post
   facets?: { type: 'mention' | 'link'; value: string }[]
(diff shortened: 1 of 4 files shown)


packages/bsky/src/services/indexing/plugins/post.ts
```diff
@@ -21,11 +21,16 @@ type Post = Selectable<DatabaseSchemaType['post']>
 type PostEmbedImage = DatabaseSchemaType['post_embed_image']
 type PostEmbedExternal = DatabaseSchemaType['post_embed_external']
 type PostEmbedRecord = DatabaseSchemaType['post_embed_record']
+
+type Ancestor = PostHierarchy & {
+  doesNotExst?: boolean
```
devinivy, line 26: No biggie, but in PDS version this reads `doesNotExist` with the "i" present.

devinivy (approved):
Lookin good to me!

Just to toss out an option that came to mind while reading it— another approach is to make this more generic by filtering out any notifications that have a `reasonSubject` that doesn't exist.  That could be done either at query time (i.e. in the `listNotifications` method) or right before/after the notifications are inserted.

dholms:
Now there's an idea.... I can't believe I didn't think of that 😅

dholms:
I liked that idea so whipped this up real quick: https://github.com/bluesky-social/atproto/pull/843

was much easier/faster too 😛

dholms:
closed in favor of https://github.com/bluesky-social/atproto/pull/843

PR #840, opened 2023-04-19
Bsky appview admin and moderation endpoints

devinivy:
Core admin/moderation tooling is now present in the appview:

 - Implement com.atproto.admin.* methods on appview.
 - Implement com.atproto.moderation.createReport method on appview.
 - Implement com.atproto.identity.resolveHandle method on appview.
 - Update dev-env so that it can be run with a bsky appview.
 - Update appview blob resolver so that it doesn't resolve taken-down blobs.
 - Invalidate image cache on takedown.

Does not have automated labeling or serve labels on non-admin views yet— that will be in a followup PR.


packages/bsky/src/db/tables/label.ts
```diff
@@ -0,0 +1,12 @@
+export const tableName = 'label'
+
+export interface Label {
+  src: string
+  uri: string
+  cid: string
+  val: string
+  neg: 0 | 1 // @TODO convert to boolean in app-view
```
dholms, line 8: oh this is a boolean now! got it right in migration 👌

packages/bsky/src/services/moderation/index.ts
```diff
@@ -0,0 +1,455 @@
...
+  async takedownRepo(info: { takedownId: number; did: string }) {
+    await this.db.db
+      .updateTable('actor')
+      .set({ takedownId: info.takedownId })
+      .where('did', '=', info.did)
+      .where('takedownId', 'is', null)
+      .executeTakeFirst()
+  }
+
+  async reverseTakedownRepo(info: { did: string }) {
```
dholms, line 291: should we be invalidating all related blobs here?

packages/bsky/src/api/blob-resolver.ts
```diff
@@ -30,7 +31,24 @@ export const createRouter = (ctx: AppContext): express.Router => {
         return next(createError(400, 'Invalid cid'))
       }
 
-      const { pds } = await ctx.didResolver.resolveAtpData(did) // @TODO cache did info
+      const [{ pds }, takedown] = await Promise.all([
+        ctx.didResolver.resolveAtpData(did), // @TODO cache did info
+        ctx.db.db
+          .selectFrom('moderation_action_subject_blob')
```
dholms, line 37: do you need a `.select()` or `selectAll()` here?

packages/bsky/src/api/com/atproto/identity/resolveHandle.ts
```diff
@@ -13,6 +13,16 @@ export default function (server: Server, ctx: AppContext) {
     if (user) {
       did = user.did
     } else {
+      const publicHostname = ctx.cfg.publicUrl
```
dholms, line 16: we just gotta ensure we don't have any alternate subdomains for the appview 😛

dholms (approved):
left some comments but overall lookin rad 😎

packages/bsky/src/api/com/atproto/identity/resolveHandle.ts
devinivy: Yeah 😳 I wonder if we should eventually wrap this into a more general configuration allowing multiple hostnames or even wildcards.  I'll leave a note here for now.

packages/bsky/src/api/blob-resolver.ts
devinivy: I double-checked and it does work without it, but good call!  Definitely should have one.

packages/bsky/src/services/moderation/index.ts
devinivy: That's a good question.  To date blobs haven't been a part of repo takedowns, but it's possible to start allowing it with a little effort.  Here's where that currently gets stopped: https://github.com/bluesky-social/atproto/blob/f30887bc1c21936ff2b79832d4f80e04edfdb9e2/packages/pds/src/services/moderation/index.ts#L215-L217

PR #841, opened 2023-04-19
Thread mutes

dholms:
⚠️ WIP ⚠️

### DIFF
--- a/lexicons/app/bsky/feed/mutePostThread.json
+++ b/lexicons/app/bsky/feed/mutePostThread.json
@@ -1,0 +1,20 @@
+{
+  "lexicon": 1,
+  "id": "app.bsky.feed.mutePostThread",
+  "defs": {
+    "main": {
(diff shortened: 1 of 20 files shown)


pfrazee:
Similar to other kinds of mutes, I need to have "thread muted" state attached to posts when they're served so that I can correctly display the mute/unmute toggle.

If we don't want to pay that overhead everywhere I can request that state when the dropdown opens but I dont love the latency that will introduce to the UI

dholms:
ah right... yup forgot about that

how should thread mutes work with getTimeline feed? those posts will still show up right?

pfrazee:
Yep correct. Also I'm applying this filter clientside right now -- not sure 100% what the plan is for server/client filtering of notifs but I have all the needed logic (for thread muting) in the client

dholms:
closing for now - probably need a new pass on this

i'll keep the branch around in case we come back to it

Issue #823, opened 2023-04-17
Use production URL in DidResolver as default plcUrl.


simpolism:
## Description

See [TODO in the DID resolver](https://github.com/bluesky-social/atproto/blob/main/packages/did-resolver/src/resolver.ts#L21).

Would like access to a publicly accessible plcUrl, so I can verify what DID resolution looks like without having to follow local plc server setup/configuration (if it's even available, I didn't check). Easiest to supply as default, per TODO.

Issue #824, opened 2023-04-17
Document which packages are web-friendly


simpolism:
Tried to use @atproto/did-resolver in the browser -- got an error because it depends on @atproto/common.

Would love to have a list of which packages are usable in the browser and which are nodejs only.

dholms:
Good find 👍 

That one should probably be web-friendly

PR #842, opened 2023-04-19
Clean up Did-resolver

dholms:
A few things here:
- no longer use the published did-resolver library. We weren't getting anything from it & was adding complexity to the codebase
- default plc url to the production plc service (`https://plc.directory`)
- add in flexible caching ability to the did-resolver
- better error handling
- remove node dependencies 
  - `@atproto/common` -> `@atproto/common-web`
  - remove the one node dep in `@atproto/crypto` around streaming sha256

Closes https://github.com/bluesky-social/atproto/issues/823
Partially addresses https://github.com/bluesky-social/atproto/issues/824

### DIFF
--- a/packages/bsky/src/services/indexing/index.ts
+++ b/packages/bsky/src/services/indexing/index.ts
@@ -80,7 +80,7 @@ export class IndexingService {
     if (actor && !force) {
       return
     }
-    const { handle } = await this.didResolver.resolveAtpData(did)
+    const { handle } = await this.didResolver.resolveAtprotoData(did)
     const handleToDid = await resolveExternalHandle(handle)
     if (did !== handleToDid) {
       return // No bidirectional link between did and handle
@@ -104,7 +104,7 @@ export class IndexingService {
   async indexRepo(did: string, commit: string) {
     this.db.assertTransaction()
     const now = new Date().toISOString()
-    const { pds, signingKey } = await this.didResolver.resolveAtpData(did)
+    const { pds, signingKey } = await this.didResolver.resolveAtprotoData(did)
     const { api } = new AtpAgent({ service: pds })
 
     const { data: car } = await retryHttp(() =>
@@ -194,7 +194,7 @@ export class IndexingService {
   async tombstoneActor(did: string) {
     this.db.assertTransaction()
     const doc = await this.didResolver.resolveDid(did)
-    if (doc.didResolutionMetadata.error === 'notFound') {
+    if (doc === null) {
       await Promise.all([
         this.unindexActor(did),
         this.db.db.deleteFrom('actor').where('did', '=', did).execute(),
--- a/packages/did-resolver/src/base-resolver.ts
+++ b/packages/did-resolver/src/base-resolver.ts
@@ -10,15 +10,33 @@ export abstract class BaseResolver {
 
   abstract resolveDidNoCheck(did: string): Promise<unknown | null>
 
-  async refreshCache(did: string): Promise<void> {
-    if (!this.cache) return
-    const got = await this.resolveDidNoCheck(did)
-    if (check.is(got, didDocument)) {
-      await this.cache.cacheDid(did, got)
+  validateDidDoc(did: string, val: unknown): DidDocument {
+    if (!check.is(val, didDocument)) {
+      throw new PoorlyFormattedDidDocumentError(did, val)
     }
+    if (val.id !== did) {
+      throw new PoorlyFormattedDidDocumentError(did, val)
+    }
+    return val
   }
 
-  async resolveDid(did: string): Promise<DidDocument | null> {
+  async refreshCache(did: string): Promise<DidDocument | null> {
+    if (!this.cache) return this.resolveDid(did)
+    const got = await this.resolveDidNoCheck(did)
+    if (got === null) return null
+    const doc = this.validateDidDoc(did, got)
+    await this.cache.cacheDid(did, doc)
+    return doc
+  }
+
+  async resolveDid(
+    did: string,
+    forceRefresh = false,
+  ): Promise<DidDocument | null> {
+    if (forceRefresh) {
+      return this.refreshCache(did)
+    }
+
     if (this.cache) {
       const fromCache = await this.cache.checkCache(did)
       if (fromCache?.expired) {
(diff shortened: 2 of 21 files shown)


packages/bsky/src/services/indexing/index.ts
```diff
@@ -194,7 +194,7 @@ export class IndexingService {
   async tombstoneActor(did: string) {
     this.db.assertTransaction()
     const doc = await this.didResolver.resolveDid(did)
-    if (doc.didResolutionMetadata.error === 'notFound') {
+    if (doc === null) {
```
devinivy, line 197: The key thing I'm looking for here is that the doc positively doesn't exist, versus not being able to find it due to a network error, service down, or similar.  I think you're capturing that here, just wanted to double-check!

packages/did-resolver/src/base-resolver.ts
```diff
@@ -0,0 +1,92 @@
...
+      }
+      if (fromCache) {
+        return fromCache.doc
+      }
+    }
+
+    const got = await this.resolveDidNoCache(did)
+    if (got === null) return null
+    await this.cache?.cacheDid(did, got)
+    return got
```
devinivy, line 50: I could imagine a world in which a missing doc is cacheable.  I may need to think it through a bit more though, hmm 🤔

packages/did-resolver/src/memory-cache.ts
```diff
@@ -0,0 +1,48 @@
...
+    if (doc) {
+      await this.cacheDid(did, doc)
+    }
+  }
+
+  async checkCache(did: string): Promise<CacheResult | null> {
+    const val = this.cache[did]
+    if (!val) return null
+    const now = Date.now()
+    const expired = now > val.updatedAt + this.ttl
```
devinivy, line 37: Should we remove expired items from the cache here?

packages/did-resolver/src/memory-cache.ts
```diff
@@ -0,0 +1,48 @@
...
+}
+
+export class MemoryCache extends DidCache {
+  public ttl: number
+  constructor(ttl?: number) {
+    super()
+    this.ttl = ttl ?? DAY
+  }
+
+  public cache: Record<string, CacheVal> = {}
```
devinivy, line 17: A `Map` may be better suited to backing this cache (even though the keys are strings anyway), particularly if we start removing items from the cache.

devinivy: Also slightly nicer from a typing perspective I think, since it knows `get()` may come back `undefined`, but `cache[did]` is always considered to be present.

packages/did-resolver/src/base-resolver.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  async resolveDid(
+    did: string,
+    forceRefresh = false,
+  ): Promise<DidDocument | null> {
+    if (this.cache && !forceRefresh) {
+      const fromCache = await this.cache.checkCache(did)
+      if (fromCache?.expired) {
+        this.refreshCache(did)
+      }
```
devinivy, line 41: I think that it can be useful to have separate ideas of expiration (not servable) versus staleness (servable but requires revalidation).  As things currently stand this item could be long-expired, and we'd still serve the current value.  I don't think we necessarily need to get into now and make any changes, but my inclination might be to have expiration working first for correctness, then layer-in staleness as an optimization.

packages/bsky/src/services/indexing/index.ts
devinivy: I found the logic for it 👍 I think I was the last to touch that— I be we can tune it up to be more accurate, but seems reasonable enough for now!

devinivy (approved):
Much more straightforward! 👏 Only questions/thought I had are on caching minutiae. Looks nice 👍

packages/bsky/src/services/indexing/index.ts
dholms: yup yup cool 👍 

i try to return null when it doesn't exist & throw an error for a network error 👌

packages/did-resolver/src/base-resolver.ts
dholms: I could see that... & not opposed to adding that in if we find it'd be helpful. My thought is that if you cached the missing doc & are asking to resolve the did, basically everytime you'll want to bypass the cache (cause you probably wouldn't be resolving if you thought it did not exist)

packages/did-resolver/src/memory-cache.ts
dholms: The did resolver takes care of doing that ^^ I think it's useful to have the stale value around until it finishes revalidating

packages/did-resolver/src/base-resolver.ts
dholms: where do you draw the line between staleness & expired? staleness being "any cache hit" & expired being "older than TTL"?

devinivy: I usually think of it as two timeouts.  For example, in http it might look like this:
`Cache-Control: max-age=20, stale-while-revalidate=40`
 - for 0-20 seconds serve the cached version
 - for 20-60 seconds serve the cached version, and revalidate in the background
 - for 60+ seconds do not serve the cached version, revalidate and serve the latest result

packages/did-resolver/src/memory-cache.ts
dholms: yup yup good call switched it over 👌

packages/did-resolver/src/base-resolver.ts
dholms: ahh got it got it 👍 

i'll take a pass 👍

dholms:
Sweet got in the diff between stale & expired and also wrote some tests for the cache 👌

packages/did-resolver/src/memory-cache.ts
```diff
@@ -8,16 +8,18 @@ type CacheVal = {
 }
 
 export class MemoryCache extends DidCache {
-  public ttl: number
-  constructor(ttl?: number) {
+  public staleTTL
```
devinivy, line 11: I wonder if this is getting implicitly typed as `any`.
```suggestion
  public staleTTL: number
```

packages/did-resolver/src/base-resolver.ts
```diff
@@ -36,16 +36,19 @@ export abstract class BaseResolver {
   ): Promise<DidDocument | null> {
     if (this.cache && !forceRefresh) {
       const fromCache = await this.cache.checkCache(did)
-      if (fromCache?.expired) {
-        this.refreshCache(did)
+      if (fromCache?.stale) {
+        await this.refreshCache(did)
```
devinivy, line 40: I don't think we necessarily need the `await` here since the stale value is good to be served immediately while it gets refreshed.

devinivy (approved):
Nice!

packages/did-resolver/src/memory-cache.ts
dholms: ah man it was. good catch 👍

packages/did-resolver/src/base-resolver.ts
dholms: I went back & forth on this. but i decided to keep the await there & then the cache gets to decide how async it should be

This simple memory cache does not do it async. but the sql cache that we use on the appview keeps a pqueue that it pushes it on & does the refresh async

dholms: the other option is to get the pqueue into the base-resolver but that all seemed a little bit heavy

devinivy: Seems good as-is 👍

PR #845, opened 2023-04-19
Simple service auth

dholms:
Simple service-to-service auth

This is a JWT that is used by one service to communicate with another service on a users behalf. These JWTs are asymmetrically signed by the user's repo signing key & include the user's did as the issuer.

To verify auth, we must do a did lookup, therefore it's important that we have reasonable caching of did documents.

---

### JWT details

I'm pretty sure I stayed on-spec for JWTs (stole some code from UCAN lib, where we went to decent ends to make sure we were on-spec), but it's worth a glance to be sure

```
// header
{
  typ: "JWT",
  alg: "ES256K" // jwt alg for secp256k1
}

// payload
{
  iss: did:example:usersDid, // _user's_ did
  exp: 1681870952 // expiration time in _seconds_
}
```
To produce the signature:
- base64url encode the header & payload
- join them with a `"."` between
- sha256 & sign the utf8 bytes
- base64url encode the signature


packages/bsky/src/auth.ts
```diff
@@ -0,0 +1,85 @@
...
+export const getJwtStrFromReq = (req: express.Request): string | null => {
+  const { authorization = '' } = req.headers
+  if (!authorization.startsWith('Bearer ')) {
+    return null
+  }
+  return authorization.replace('Bearer ', '').trim()
+}
+
+const parseB64UrlToJson = (b64: string) => {
+  return JSON.parse(ui8.toString(ui8.fromString(b64, 'utf8')))
```
devinivy, line 79: Could def be missing a detail here, but I expected the encoding to be something like `base64url` rather than `utf8`.

packages/bsky/src/auth.ts
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  const msgBytes = ui8.fromString(parts.slice(0, 2).join('.'), 'utf8')
+  const sigBytes = ui8.fromString(sig, 'base64url')
+
+  const atpData = await didResolver.resolveAtpData(payload.iss)
+  let validSig: boolean
+  try {
+    validSig = await crypto.verifySignature(
+      atpData.signingKey,
```
devinivy, line 50: I think ideally the header's `alg` would be verified to match the signing key.

packages/bsky/src/auth.ts
```diff
@@ -0,0 +1,85 @@
...
+  } catch (err) {
+    throw new AuthRequiredError(
+      'could not verify jwt signature',
+      'BadJwtSignature',
+    )
+  }
+  if (!validSig) {
+    throw new AuthRequiredError(
+      'jwt signature does not match jwt issuer',
+      'BadJwtSig',
```
devinivy, line 63: Just syncing-up with the similar code used above:
```suggestion
      'BadJwtSignature',
```

packages/pds/src/auth.ts
```diff
@@ -234,5 +233,28 @@ export const adminVerifier =
...
+    exp,
+  }
+  const toSignStr = `${jsonToB64Url(header)}.${jsonToB64Url(payload)}`
+  const toSign = ui8.fromString(toSignStr, 'utf8')
+  const sig = await keypair.sign(toSign)
+  return `${toSignStr}.${ui8.toString(sig, 'base64url')}`
+}
+
+const jsonToB64Url = (json: Record<string, unknown>): string => {
+  return ui8.toString(ui8.fromString(JSON.stringify(json), 'utf8'))
```
devinivy, line 259: Same here, I feel like I'm missing where `base64url` comes into play—I think `toString()` and `fromString() defaults to encoding to `utf8`, so what I expected to see is more like this:
```suggestion
  return ui8.toString(ui8.fromString(JSON.stringify(json)), 'base64url')
```

packages/bsky/src/auth.ts
```diff
@@ -0,0 +1,85 @@
...
+
+const verifyJwt = async (
+  didResolver: DidResolver,
+  jwtStr: string,
+): Promise<string> => {
+  const parts = jwtStr.split('.')
+  if (parts.length !== 3) {
+    throw new AuthRequiredError('poorly formatted jwt', 'BadJwt')
+  }
+  const payload = parseB64UrlToJson(parts[1]) as JwtPayload
```
devinivy, line 35: We might consider verifying `payload` is an object.  Someone could craft a bad token where `payload` is `null`, and we'd end-up 500'ing when checking `exp` below.

packages/bsky/src/auth.ts
```diff
@@ -0,0 +1,85 @@
...
+  const sig = parts[2]
+
+  // @TODO add aud check?
+  if (Date.now() / 1000 > payload.exp) {
```
devinivy, line 39: Might be worth confirming that `exp` is present and a number as part of the expiration check.

packages/bsky/src/auth.ts
```diff
@@ -0,0 +1,85 @@
...
+    throw new AuthRequiredError('jwt expired', 'JwtExpired')
+  }
+
+  const msgBytes = ui8.fromString(parts.slice(0, 2).join('.'), 'utf8')
+  const sigBytes = ui8.fromString(sig, 'base64url')
+
+  const atpData = await didResolver.resolveAtpData(payload.iss)
```
devinivy, line 46: We may want to confirm `payload.iss` is a string before passing it around, just may be able to trigger an unexpected error.

devinivy (commented):
Love this approach 💎 Just had a few open questions about decoding and verification that should be straightforward to work through!

packages/bsky/src/auth.ts
dholms: yup you're right, was supposed to be `  return JSON.parse(ui8.toString(ui8.fromString(b64, 'base64url'), 'utf8'))`
good catch 🙇

packages/bsky/src/auth.ts
dholms: 👍

packages/pds/src/auth.ts
dholms: yup yup 👍 

altho it's actually
`return ui8.toString(ui8.fromString(JSON.stringify(json), 'utf8'), 'base64url')`

i'm gonna split these into helper methods to be less confusing

packages/bsky/src/auth.ts
dholms: yup good call 👍

packages/bsky/src/auth.ts
dholms: yup 👍

packages/bsky/src/auth.ts
dholms: sweet i just got an extra verification of payload step in in general 👌

packages/bsky/src/auth.ts
```diff
@@ -76,7 +76,19 @@ export const getJwtStrFromReq = (req: express.Request): string | null => {
 }
 
 const parseB64UrlToJson = (b64: string) => {
-  return JSON.parse(ui8.toString(ui8.fromString(b64, 'utf8')))
+  return JSON.parse(common.b64UrlToUtf8(b64))
+}
+
+const parsePayload = (b64: string): JwtPayload => {
+  const payload = parseB64UrlToJson(b64)
+  if (typeof payload !== 'object') {
```
devinivy, line 84: If for some reason the JSON were `null` 🙃 
```suggestion
  if (!payload || typeof payload !== 'object') {
```

dholms: ugh dammit you right 🙇

devinivy (approved):

devinivy:
Note to self— once this lands incorporate #863 into the appview.  I went to do it as part of #840, but enough will change with auth that I think it makes sense to land this one first.

devinivy (approved):

PR #848, opened 2023-04-20
Improve streaming sync routes

dholms:
Previously, we were reading the car file to a buffer before writing it out to the response.

Now we just stream it directly out

### DIFF
--- a/packages/common/src/streams.ts
+++ b/packages/common/src/streams.ts
@@ -38,6 +38,24 @@ export const streamToBytes = async (stream: Readable): Promise<Uint8Array> => {
   return new Uint8Array(Buffer.concat(bufs))
 }
 
+const pushIterableToStream = async (
+  iter: AsyncIterable<Uint8Array>,
+  stream: Readable,
+) => {
+  for await (const bytes of iter) {
+    stream.push(bytes)
+  }
+  stream.push(null)
+}
+
+export const byteIterableToStream = (
+  iter: AsyncIterable<Uint8Array>,
+): Readable => {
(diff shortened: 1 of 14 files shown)


packages/common/src/streams.ts
```diff
@@ -38,6 +38,24 @@ export const streamToBytes = async (stream: Readable): Promise<Uint8Array> => {
   return new Uint8Array(Buffer.concat(bufs))
 }
 
+const pushIterableToStream = async (
+  iter: AsyncIterable<Uint8Array>,
+  stream: Readable,
+) => {
+  for await (const bytes of iter) {
+    stream.push(bytes)
+  }
```
devinivy, line 47: If this throws, I think we should ensure `stream` gets destroyed here.

devinivy: Actually, `Readable.from()` may even do all this for us.

packages/pds/src/sql-repo-storage.ts
```diff
@@ -167,6 +167,7 @@ export class SqlRepoStorage extends RepoStorage {
         size: bytes.length,
         content: bytes,
       })
+      this.cache.set(cid, bytes)
```
devinivy, line 170: Just making sure this was an intentional addition!

devinivy (approved):
This rocks 💎

packages/common/src/streams.ts
dholms: wow nice i searched around for that exact fn & didn't find it 🙃

devinivy (approved):

PR #851, opened 2023-04-20
Add labeler functionality to the appview

devinivy:
This ports the labeler functionality originally implemented in the PDS into the appview.

Depends on #840, so we'll want to merge that one first.


packages/bsky/src/api/blob-resolver.ts
```diff
@@ -31,42 +32,12 @@ export const createRouter = (ctx: AppContext): express.Router => {
...
-        getBlob({ pds, did, cid: cidStr }),
-      )
-
-      const imageStream: Readable = blobResult.data
-      const verifyCid = new VerifyCidTransform(cid)
+      const verifiedImage = await resolveBlob(did, cid, ctx)
 
-      // Send chunked response, destroying stream early (before
-      // closing chunk) if the bytes don't match the expected cid.
-      res.statusCode = 200
```
dholms, line 61: did we end up dropping these headers/status code?

packages/bsky/src/labeler/base.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  abstract labelText(text: string): Promise<string[]>
+  abstract labelImg(img: stream.Readable): Promise<string[]>
+
+  async processAll() {
+    await this.processingQueue?.onIdle()
+  }
+
+  async destroy() {
+    const pQueue = this.processingQueue
```
dholms, line 76: do we still wanna do this on a pQueue in the appview? it's less pressing if it holds up a tx for a second here, but also it'll still increase time till a record makes it into views 🤔 

honestly a durable queue would be nice here... this probably suffices for now

packages/bsky/tests/labeler/labeler.test.ts
```diff
@@ -0,0 +1,205 @@
...
+
+  it('retrieves both profile & repo labels on profile views', async () => {
+    await ctx.db.db
+      .insertInto('label')
+      .values({
+        src: labelerDid,
+        uri: alice,
+        cid: '',
+        val: 'repo-label',
+        neg: 0,
```
dholms, line 157: another place that should be a bool

packages/bsky/tests/views/__snapshots__/notifications.test.ts.snap
```diff
@@ -218,7 +226,7 @@ Array [
           "uri": "record(2)",
         },
       },
-      "text": "hear that",
+      "text": "hear that label_me label_me_2",
```
dholms, line 229: looks like labels aren't making it onto this notif?

dholms: yeah checked code & pds does labeling in listNotifications & not implemented in appview yet

dholms (approved):
overall lookin good 👌 

big thing is ensuring that labels get onto notification views!

and want to make sure we've thought through the continued use of p-queue -- even tho i think that it's the best option for the moment until we get a durable queue in

packages/bsky/src/api/blob-resolver.ts
devinivy: Oops, really not sure how I did that 🙃 thanks for catching!

packages/bsky/src/labeler/base.ts
devinivy: I agree with you.  My inclination is to leave it relatively in sync with the pds's implementation til things settle.  I also think getting records into the view quickly is important.  A proper queue would/will be 💎

packages/bsky/tests/views/__snapshots__/notifications.test.ts.snap
devinivy: Oh man, fail!  Thanks for catching that!

packages/bsky/tests/labeler/labeler.test.ts
devinivy: Thanks for surfacing!  Now that it's in the db model typescript's got my back too 😎 So these issues should be pretty well taken care of.

PR #856, opened 2023-04-21
Did caching

dholms:
Split off from [simple-service-auth](https://github.com/bluesky-social/atproto/pull/845) for ease of review

This adds a Postgres DID cache to the appview building off of https://github.com/bluesky-social/atproto/pull/842

We do stale-while-revalidate when TTL is passed & can override the cache for sensitive operations such as updating indexed handles


packages/bsky/src/did-cache.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  async checkCache(did: string): Promise<CacheResult | null> {
+    const res = await this.db.db
+      .selectFrom('did_cache')
+      .where('did', '=', did)
+      .selectAll()
+      .executeTakeFirst()
+    if (!res) return null
+    const now = Date.now()
+    const updatedAt = new Date(res.updatedAt).getTime()
```
devinivy, line 43: `Date` may not have to come into play here since `res.updatedAt` is a unix timestamp itself.

packages/bsky/src/db/tables/did-cache.ts
```diff
@@ -0,0 +1,13 @@
+import { DidDocument } from '@atproto/did-resolver'
+
+export interface DidCache {
+  did: string
+  doc: DidDocument
```
devinivy, line 5: Is it all good for tombstoned dids to be cached here?

devinivy (approved):
Excellent 👍!!

packages/bsky/src/did-cache.ts
dholms: oh yeah i had that as Date originally 🙃

packages/bsky/src/db/tables/did-cache.ts
dholms: I cleaned it up a bit so if we try to refresh & it's missing, we'll clear out the cache. should fix tombstoned dids 👍

devinivy (approved):

PR #858, opened 2023-04-21
Materialize post and profile aggregations on write

devinivy:
This work materializes the following profile and post aggregations into a table:
 - post like count
 - post reply count
 - post repost count
 - profile followers count
 - profile following count
 - profile post count

The work is done in the background on each write, so it's not in the critical path of writes.  The aggregations are computed directly from the `post`, `follow`, `like`, and `repost` tables, so they can always be refreshed in a pinch.  Another change in here is that I created a single "background queue", and also put labeler work on it.

In #859 we start using the aggregations in reads.


packages/pds/src/app-view/db/tables/post-agg.ts
```diff
@@ -0,0 +1,14 @@
+import { Generated } from 'kysely'
+
+export const tableName = 'post_agg'
+
+export interface PostAgg {
+  uri: string
```
dholms, line 6: we'll wanna track counts for different cids at some point. not critical rn tho

packages/pds/src/app-view/services/indexing/index.ts
```diff
@@ -18,18 +20,18 @@ export class IndexingService {
     profile: Profile.PluginType
   }
 
-  constructor(public db: Database, public messageDispatcher: MessageQueue) {
+  constructor(public db: Database, public backgroundQueue: BackgroundQueue) {
     this.records = {
-      post: Post.makePlugin(this.db.db),
```
dholms, line 23: nice good call on making this not pass the underlying DatabaseSchema obj 👌

packages/pds/src/db/index.ts
```diff
@@ -113,6 +114,11 @@ export class Database {
     }
   }
 
+  onCommit(fn: () => void) {
```
dholms, line 117: cool

dholms (approved):
yup yup this is excellent 🙌

PR #859, opened 2023-04-21
Serve materialized post and profile aggregations

devinivy:
This is a followup to #858, to actually start serving the materialized aggregations for posts and profiles.


dholms (approved):
looks great 👌 😎

PR #860, opened 2023-04-21
Format commits outside of tx

dholms:
This moves formatting of commits outside of the write transactions so that they do not hold up connections from the pool.

For each write, we attempt up to 10 times with a 100ms delay between each. Each attempt only takes some short reprocessing where we replay the commit on top of the current state of the repo (if the lock is up). We use a persistent `SqlRepoStorage` instance between attempts so that already computed blocks can be cached & not re-fetched from storage

### DIFF
--- a/packages/pds/src/api/com/atproto/repo/createRecord.ts
+++ b/packages/pds/src/api/com/atproto/repo/createRecord.ts
@@ -12,6 +12,7 @@ import {
 import AppContext from '../../../../context'
 import { ids } from '../../../../lexicon/lexicons'
 import Database from '../../../../db'
+import { ConcurrentWriteError } from '../../../../services/repo'
 
 export default function (server: Server, ctx: AppContext) {
   server.com.atproto.repo.createRecord({
@@ -60,10 +61,12 @@ export default function (server: Server, ctx: AppContext) {
       try {
         await ctx.services
           .repo(ctx.db)
-          .processWrites({ did, writes, swapCommitCid }, 5, 100)
+          .processWrites({ did, writes, swapCommitCid }, 10)
       } catch (err) {
         if (err instanceof BadCommitSwapError) {
           throw new InvalidRequestError(err.message, 'InvalidSwap')
+        } else if (err instanceof ConcurrentWriteError) {
+          throw new InvalidRequestError(err.message, 'ConcurrentWrites')
         }
         throw err
       }
--- a/packages/pds/src/api/com/atproto/repo/createRecord.ts
+++ b/packages/pds/src/api/com/atproto/repo/createRecord.ts
@@ -60,7 +60,7 @@ export default function (server: Server, ctx: AppContext) {
       try {
         await ctx.services
           .repo(ctx.db)
-          .attemptWrites({ did, writes, swapCommitCid }, 5, 100)
+          .processWrites({ did, writes, swapCommitCid }, 5, 100)
       } catch (err) {
         if (err instanceof BadCommitSwapError) {
           throw new InvalidRequestError(err.message, 'InvalidSwap')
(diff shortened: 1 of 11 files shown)


packages/pds/src/api/com/atproto/repo/createRecord.ts
```diff
@@ -52,26 +52,24 @@ export default function (server: Server, ctx: AppContext) {
...
-          )
-        } catch (err) {
-          if (err instanceof BadCommitSwapError) {
-            throw new InvalidRequestError(err.message, 'InvalidSwap')
-          } else {
-            throw err
-          }
+      const backlinkDeletions = validate
+        ? await getBacklinkDeletions(ctx.db, ctx, write)
+        : []
```
devinivy, line 57: I think the backlink result here could technically get out of sync, but it may be acceptable since it's not critical functionality.  Come to think of it, I suppose that was already probably an issue to some extent 🤔

packages/pds/src/db/index.ts
```diff
@@ -176,6 +176,10 @@ export class Database {
     assert(this.isTransaction, 'Transaction required')
   }
 
+  assertNotTransaction() {
+    assert(!this.isTransaction, 'Cannot be in a transaction')
+  }
```
devinivy, line 181: 👀

packages/pds/src/sql-repo-storage.ts
```diff
@@ -30,25 +29,17 @@ export class SqlRepoStorage extends RepoStorage {
...
-  private async getAndLockHeadPg(): Promise<RepoRoot | null> {
-    const selectAndSkipLock = async () => {
-      return await this.db.db
-        .selectFrom('repo_root')
-        .selectAll()
-        .forUpdate()
-        .skipLocked()
-        .where('did', '=', this.did)
-        .executeTakeFirst()
+  async lockHead(): Promise<CID | null> {
```
devinivy, line 32: I think this is a nice little improvement to the interface 👍

packages/pds/src/sql-repo-storage.ts
```diff
@@ -30,25 +29,17 @@ export class SqlRepoStorage extends RepoStorage {
   // instead of waiting for row locks, we attempt to grab the root skipping over locked rows
   // if we get no result, we wait for 200ms so the previous write has the chance to go through & then try again
   // if the lock is still held, we check (without a row lock) to see the row exists at all
```
devinivy, line 31: These may be out of date now.  I think it's worth maintaining a note that it may return `null` even if the head exists.

packages/pds/src/services/repo/index.ts
```diff
@@ -174,7 +225,7 @@ export class RepoService {
   async rebaseRepo(did: string, now: string, swapCommit?: CID) {
     this.db.assertTransaction()
     const storage = new SqlRepoStorage(this.db, did, now)
-    const currRoot = await storage.getHead(true)
+    const currRoot = await storage.lockHead()
     if (!currRoot) {
       throw new InvalidRequestError(
         `${did} is not a registered repo on this server`,
```
devinivy, line 231: Should this be a `ConcurrentWriteError` as well?

devinivy (approved):
Surprisingly easy read 👍 👍

packages/pds/src/sql-repo-storage.ts
dholms: agreed. less footgun-y

packages/pds/src/sql-repo-storage.ts
dholms: yup good point 👍

packages/pds/src/api/com/atproto/repo/createRecord.ts
dholms: yeah i thought about that... but my reaction was basically what you said:
- was already an issue
- not critical path
- didn't wanna have to mess with it 😅

packages/pds/src/services/repo/index.ts
dholms: yup 👍

devinivy (approved):

PR #866, opened 2023-04-22
Remove profile labels

dholms:
We weren't ready for them yet & they were causing ui quirks in the app


devinivy (approved):
Yup!!

packages/pds/src/labeler/util.ts
```diff
@@ -14,8 +14,9 @@ type RecordFields = {
 export const getFieldsFromRecord = (record: unknown): RecordFields => {
   if (isPost(record)) {
     return getFieldsFromPost(record)
-  } else if (isProfile(record)) {
-    return getFieldsFromProfile(record)
+    // @TODO add back in profile labeling
```
devinivy, line 17: One alternative here is to keep labeling profile information, but stop mixing it into label results on profile views.

dholms: yeah i thought about that... but this will decrease unnecessary load on hive. also honestly it was easier to comment out one line then make sure we're fixing all the db queries that do so 😛

PR #867, opened 2023-04-22
Add more config for pds db pool

devinivy:
Allow setting some more postgres connection pool settings via env vars for the PDS.


dholms (approved):
yup yup yup

PR #868, opened 2023-04-22
Better handle races in createAccount

dholms:
A few things here:
- move scrypt & plc op formatting outside of the db txn
- still verify availability of invite code before doing scrypto (outside of tx)
- remove row lock & replace with an integrity check at the end of the txn

### DIFF
--- a/packages/pds/src/api/com/atproto/server/createAccount.ts
+++ b/packages/pds/src/api/com/atproto/server/createAccount.ts
@@ -1,6 +1,7 @@
 import { InvalidRequestError } from '@atproto/xrpc-server'
 import * as ident from '@atproto/identifier'
 import * as plc from '@did-plc/lib'
+import * as scrypt from '../../../../db/scrypt'
 import { Server } from '../../../../lexicon'
 import { countAll } from '../../../../db/util'
 import { UserAlreadyExistsError } from '../../../../services/account'
@@ -10,6 +11,14 @@ export default function (server: Server, ctx: AppContext) {
   server.com.atproto.server.createAccount(async ({ input, req }) => {
     const { email, password, inviteCode, recoveryKey } = input.body
 
+    if (ctx.cfg.inviteRequired && !inviteCode) {
+      throw new InvalidRequestError(
+        'No invite code provided',
+        'InvalidInviteCode',
+      )
+    }
+
+    // validate handle
     let handle: string
     try {
       handle = ident.normalizeAndEnsureValidHandle(input.body.handle)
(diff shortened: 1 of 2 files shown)


packages/pds/src/api/com/atproto/server/createAccount.ts
```diff
@@ -137,3 +132,29 @@ export default function (server: Server, ctx: AppContext) {
...
+export const ensureCodeIsAvailable = async (
+  db: Database,
+  inviteCode: string,
+  withLock = false,
+): Promise<void> => {
+  const invite = await db.db
+    .selectFrom('invite_code')
+    .selectAll()
+    .where('code', '=', inviteCode)
+    .if(withLock && db.dialect === 'pg', (qb) => qb.forUpdate().skipLocked())
```
devinivy, line 145: We have a skip-locked here but no retrying mechanism the way you originally setup with repo_root— just wanted to make sure that's intentional.  Maybe the skip isn't as necessary given that a sanity check has already been done by the time a lock is taken.

packages/pds/src/api/com/atproto/server/createAccount.ts
```diff
@@ -101,6 +95,7 @@ export default function (server: Server, ctx: AppContext) {
         throw err
       }
 
+      // insert invite code use & do an integrity check to make sure we didn't over-use it
```
devinivy, line 98: Based on this comment and the PR description, I sort of thought I'd see some kind of `ensureCodeIsAvailable()` check after the `invite_code_use` is created.  But at the same time, the way it's currently written seems perfectly fine to me!  Did the approach change a little bit since this comment was written? Or maybe I'm just misinterpreting it!

devinivy (approved):
Had a couple Qs, but the approach seems solid to me.  Even just getting that scrypt out of there was clutch!

packages/pds/src/api/com/atproto/server/createAccount.ts
dholms: ah the comment was from when i _did_ have that exact check. but I ended up rearranging it 👍

packages/pds/src/api/com/atproto/server/createAccount.ts
dholms: Yeah I wanted to keep complexity low & this tx is so short-lived that this should really only be hit if someone is trying to abuse & use the same invite code twice in rapid succession. I'm fine with throwing them errors 🤷‍♂️

devinivy: I suppose there's a chance two different folks have the same invite w/ multiple uses and try it at the same time, but chances there are negligible and retrying still works just fine 👍

PR #871, opened 2023-04-22
More forgiving config for empty db schema

devinivy:
Followup to #867— also making config of an empty db schema more forgiving, i.e. handling unset versus empty string the same way.


dholms (approved):
yup yup

PR #875, opened 2023-04-23
minor Makefile tweak: run-dev-pds, not run-pds

bnewbold:
low-priority, but indigo and PLC both have the `run-dev-*` format, and I keep typo-ing when doing integration tests.


dholms (approved):
yup looks good 👍

PR #876, opened 2023-04-23
Get posts method

dholms:
Add a method for getting an array of post views

Added to both PDS & AppView

### DIFF
--- a/packages/bsky/tests/views/posts.test.ts
+++ b/packages/bsky/tests/views/posts.test.ts
@@ -1,0 +1,85 @@
+import AtpAgent from '@atproto/api'
+import { TestEnvInfo, runTestEnv } from '@atproto/dev-env'
+import { forSnapshot, processAll, stripViewerFromPost } from '../_util'
+import { SeedClient } from '../seeds/client'
+import basicSeed from '../seeds/basic'
+
+describe('pds posts views', () => {
+  let testEnv: TestEnvInfo
+  let agent: AtpAgent
+  let sc: SeedClient
+
+  beforeAll(async () => {
+    testEnv = await runTestEnv({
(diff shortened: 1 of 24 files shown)


packages/bsky/tests/views/posts.test.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    const authed = await agent.api.app.bsky.feed.getPosts(
+      { uris },
+      { headers: sc.getHeaders(sc.dids.alice, true) },
+    )
+    const unauthed = await agent.api.app.bsky.feed.getPosts({
+      uris,
+    })
+    const stripped = authed.data.posts.map((p) => stripViewerFromPost(p))
+    // @TODO uncomment
```
dholms, line 63: I could use an eye on this @devinivy 

I fought with it for like 30 min & feel like I'm missing something obvious

Logging the stripped data looks totally fine, but for some reason in the jest compare, it says that `viewer` is still around 🧐

devinivy: Ah, it was just an issue with `stripViewerFromPost()` not dealing with record-with-media embeds.  Just pushed a fix to this branch 👍

devinivy (approved):

packages/bsky/tests/views/posts.test.ts
dholms: well now i feel silly. why tf it was it console logging it out just fine then?

dholms: ah well 🤷‍♂️

devinivy: Ah, it's all good!  No idea, sounds like a ghost in the machine.  I def spent a little time looking through `unauthed` when the issue was in `stripped`.

PR #878, opened 2023-04-24
add BR and JP specific handles to reserved list

bnewbold:
tried to find top twitter followers by country, then top news by country. not perfect, this was just a couple minutes of lookups.


dholms (approved):
looks good 👍 

merged in main so that it'll build properly 👍

PR #879, opened 2023-04-24
Ensure indexes are used on actor search queries

devinivy:
Turns out actor search was pretty expensive because we were _juust_ missing use of the trigram index.  This adds an additional word similarity condition that is less granular than the `distance` check, but is able to use the index to exclude results.  Result sets shouldn't change, but they should be an order of magnitude faster to query.

The `pg_trgm.strict_word_similarity_threshold` is the threshold used by the `<<%` strict word similarity operator to determine whether to include a result.  It's essentially the same thing as the `<<<->` operator but with a fixed distance threshold (and the threshold is defined as `1 - distance`).

### DIFF
--- a/packages/dev-env/src/test-env.ts
+++ b/packages/dev-env/src/test-env.ts
@@ -191,7 +191,7 @@ export const runBsky = async (cfg: BskyConfig): Promise<BskyServerInfo> => {
     schema: cfg.dbPostgresSchema,
   })
 
-  // Separate migration db in case migration changes some connection state that we need in the tests, e.g. "alter table ... set ..."
+  // Separate migration db in case migration changes some connection state that we need in the tests, e.g. "alter database ... set ..."
   const migrationDb = bsky.Database.postgres({
     url: cfg.dbPostgresUrl,
     schema: cfg.dbPostgresSchema,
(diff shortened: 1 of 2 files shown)


packages/pds/tests/_util.ts
```diff
@@ -108,10 +108,22 @@ export const runTestServer = async (
         })
       : Database.memory()
 
+  // Separate migration db on postgres in case migration changes some
+  // connection state that we need in the tests, e.g. "alter table ... set ..."
```
devinivy, line 112:
```suggestion
  // connection state that we need in the tests, e.g. "alter database ... set ..."
```

packages/dev-env/src/test-env.ts
```diff
@@ -191,11 +191,17 @@ export const runBsky = async (cfg: BskyConfig): Promise<BskyServerInfo> => {
     schema: cfg.dbPostgresSchema,
   })
 
+  // Separate migration db in case migration changes some connection state that we need in the tests, e.g. "alter table ... set ..."
```
devinivy, line 194:
```suggestion
  // Separate migration db in case migration changes some connection state that we need in the tests, e.g. "alter database ... set ..."
```

dholms (approved):
awesome! 🙌 😎

PR #880, opened 2023-04-24
Fix strict_word_similarity_threshold configuration

devinivy:
Follow-up to #879.  Managed services don't necessarily allow you to set `strict_word_similarity_threshold` in the way we were doing, so moving it to a config on each session/connection.


dholms (approved):
yup yup 👍

PR #885, opened 2023-04-24
Return email in mod views according to lexicon

dholms:
We did not update the impl to match the new lexicon for these. Return email as a property instead of nested under "account"

### DIFF
--- a/packages/pds/src/services/moderation/views.ts
+++ b/packages/pds/src/services/moderation/views.ts
@@ -94,7 +94,7 @@ export class ModerationViews {
       return {
         did: r.did,
         handle: r.handle,
-        account: email ? { email } : undefined,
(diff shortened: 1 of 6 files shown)


packages/pds/src/services/moderation/views.ts
```diff
@@ -94,7 +94,7 @@ export class ModerationViews {
       return {
         did: r.did,
         handle: r.handle,
-        account: email ? { email } : undefined,
+        account: email ?? undefined,
```
devinivy, line 97: I think we want to name the property `email`:
```suggestion
        email: email ?? undefined,
```

dholms: oh yup lol good call

PR #887, opened 2023-04-24
Add getSuggestions and getPosts to BskyAgent

thinkverse:
This PR adds two new shortcuts to the `BskyAgent`, namely the `getSuggestions` and `getPosts` methods. The reasoning for this is quicker, faster, and easier access to two methods which will most likely see a lot of uses going forward instead of hiding them behind more advanced API calls.


pfrazee:
LGTM! Thanks! Releasing in 0.2.10.

PR #890, opened 2023-04-25
Fix leaky database transactions

dholms:
We were having an issue with database transactions where statements were still being written to the connection after an error had thrown & the transaction had rolled back which was leading to inconsistent state in the database.

This fix uses kysely's plugin system. We add a check to the query executor such that if an error has been throwing during the transaction, every other statement on that transaction database will throw as well.

As far as I can tell, this is the last processing step before the statement is sent off to the database. The previous approach we discussed - of setting the underlying `DatabaseSchema` instance to null & throwing when you use it may still give a chance for a query to sneak in after the tx is dead

### DIFF
--- a/packages/bsky/src/db/index.ts
+++ b/packages/bsky/src/db/index.ts
@@ -56,21 +56,15 @@ export class Database {
   }
 
   async transaction<T>(fn: (db: Database) => Promise<T>): Promise<T> {
-    let res: Awaited<T>
     const leakyTxPlugin = new LeakyTxPlugin()
-    try {
-      res = await this.db
-        .withPlugin(leakyTxPlugin)
-        .transaction()
-        .execute(async (txn) => {
-          const dbTxn = new Database(txn, this.cfg)
-          const txRes = await fn(dbTxn)
-          return txRes
-        })
-    } catch (err) {
-      leakyTxPlugin.endTx()
-      throw err
-    }
+    const res = await this.db
+      .withPlugin(leakyTxPlugin)
+      .transaction()
+      .execute(async (txn) => {
+        const dbTxn = new Database(txn, this.cfg)
+        const txRes = await fn(dbTxn).finally(() => leakyTxPlugin.endTx())
+        return txRes
+      })
     return res
   }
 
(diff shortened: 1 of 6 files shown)


packages/bsky/src/db/index.ts
```diff
@@ -46,11 +56,21 @@ export class Database {
...
+      res = await this.db
+        .withPlugin(leakyTxPlugin)
+        .transaction()
+        .execute(async (txn) => {
+          const dbTxn = new Database(txn, this.cfg)
+          const txRes = await fn(dbTxn)
+          return txRes
+        })
+    } catch (err) {
+      leakyTxPlugin.endTx()
```
devinivy, line 71: To fully avoid a race I wonder if we should end the transaction immediately before the commit/rollback (and in both cases), i.e. as soon as we know `fn` should be done with it.  What do you think?

devinivy: Could look like:
```ts
const txRes = await fn(dbTxn)
  .finally(() => leakyTxPlugin.endTx())
```

packages/bsky/src/db/index.ts
```diff
@@ -115,3 +135,24 @@ type PgOptions = {
   pool?: PgPool
   schema?: string
 }
+
+class LeakyTxPlugin implements KyselyPlugin {
```
devinivy, line 139: Great idea 👍 👍 This is definitely the most reliable way to nab these.

packages/pds/src/sql-repo-storage.ts
```diff
@@ -237,9 +237,6 @@ export class SqlRepoStorage extends RepoStorage {
           root: cid.toString(),
           indexedAt: this.getTimestamp(),
         })
-        .onConflict((oc) =>
-          oc.column('did').doUpdateSet({ root: cid.toString() }),
-        )
```
devinivy, line 242: So the `prev` here isn't necessarily the `prev` listed in the commit, it could also reference the commit being replaced in a rebase, right?  Just wanted to double-check, since I could imagine this on-conflict handling rebases.

devinivy (approved):
Much nicer solution than I expected would be available to us.  Love it 💪

packages/pds/src/sql-repo-storage.ts
devinivy: I think it's all good, just wanted to double-check 👍

PR #891, opened 2023-04-25
appview: tweak repo error log line to not clobber 'service'

bnewbold:
Somehow the output of this log line has been ending up in datadog under the wrong 'service' (with the 'service' being an "wss://"-prefixed URL, not "pds" or "bav" or whatever expected).

Tweaks the 'service' key, and some others that might have conflict.


dholms (approved):
sweet good call 👍

PR #893, opened 2023-04-25
Fix SSRF vulnerability on blob

yafesdot:
**Description**

First of all, sorry in advance if this PR is a bit hilarious as this is one of my first contributions to any open-source project of this scale. SSRF potential may have seemed to me because I did not fully understand the protocol is also a possibility. However, if there is such vulnerability potential and more needs to be done to fix it than I have committed, please point it out in the comments, even if you are not a maintainer!

**Change(s) Made**

- Added URL validation to `getBlob()` function to prevent SSRF attacks.

**Related Issues**

None. Since I didn't come across any guidelines whatsoever I didn't open any on this topic either.

### DIFF
--- a/packages/bsky/src/api/blob-resolver.ts
+++ b/packages/bsky/src/api/blob-resolver.ts
@@ -118,14 +118,6 @@ export async function resolveBlob(
 
 async function getBlob(opts: { pds: string; did: string; cid: string }) {
   const { pds, did, cid } = opts
-  // Check pds protocol and hostname to prevent potential SSRF
-  const { hostname, protocol } = new URL(pds)
-  if (!['http:', 'https:'].includes(protocol)) {
-    throw createError(400, 'Invalid pds protocol')
-  }
-  if (!hostname) {
-    throw createError(400, 'Invalid pds hostname')
-  }
   return axios.get(`${pds}/xrpc/com.atproto.sync.getBlob`, {
     params: { did, cid },
     decompress: true,
--- a/packages/bsky/src/api/blob-resolver.ts
+++ b/packages/bsky/src/api/blob-resolver.ts
@@ -118,6 +118,14 @@ export async function resolveBlob(
 
 async function getBlob(opts: { pds: string; did: string; cid: string }) {
   const { pds, did, cid } = opts
+  // Check pds protocol and hostname to prevent potential SSRF
+  const { hostname, protocol } = new URL(pds)
+  if (!['http:', 'https:'].includes(protocol)) {
+    throw createError(400, 'Invalid pds protocol')
+  }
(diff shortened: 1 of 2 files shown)


pfrazee:
Appreciate the PR. I suppose we have to look at this more closely.

Right now IIRC all blob URLs are being provided by your PDS and should be from their CDN, so it's assume they're trusted if they're handed to this API. (Correct me if I'm wrong dan or devin.) I'm unsure what kind of validation we could/would apply here that would secure this further.

devinivy:
I think some validation here is warranted, but I do need to take a closer look at it.  In this case we're hitting someone's PDS as described by their DID document, and the blob resolver shouldn't totally trust the DID document to point to a legitimate URL.  Will dig-in to this soon 👍

yafesdot:
Oopsie

roj1512:
Don't mind it @yafesdot. You just pinged them.

yafesdot:
@roj1512 I *just* forgot that we can't just rename branches in git :woozy_face:, that's why...

roj1512:
That was for good. Idk why they're not digging into it (and other PRs) anyway.

dholms:
We've just been horribly busy lately, getting through some PRs now. Thanks for your patience :)

dholms (changes requested):
Okay this looks like a reasonable fix to me, but I think it would be better applied at the entry point for the untrusted address. 

Which is the `getPds` function in https://github.com/bluesky-social/atproto/blob/main/packages/did-resolver/src/atproto-data.ts

yafesdot:
Oh sorry, I just saw your review and requests. I'll make the changes you want!

Matrix89:
@yafesdot Just a heads-up, when you discover an exploit, a security hole or anything that can be used by a malicious actor for an attack, please first look if the project has a disclosure program of some sort, before mentioning the issue in public

yafesdot:
> @yafesdot Just a heads-up, when you discover an exploit, a security hole or anything that can be used by a malicious actor for an attack, please first look if the project has a disclosure program of some sort, before mentioning the issue in public

I do when needed, but this one is not that much of a threatening one, so yeah.

dholms:
Yup yup this one was fine. Low attack vector & the vector is not even accessible in production deployments right now.

This did prompt me to draft a security readme for future vulnerability reports: https://github.com/bluesky-social/atproto/pull/954

Thanks again for the help! I'll keep an eye out for the changes

yafesdot:
@dholms it should work that way, but I'm unfortunately not that sure about that since I couldn't find a way to test it out directly.

dholms (approved):
yup looks good 👍

PR #894, opened 2023-04-25
Return `XRPCError` instead of 404 if a handler errors

roj1512:
Currently, if there is an uncaught error inside a handler, a 404 response is returned instead of the error itself. This fixes that.

### DIFF
--- a/packages/xrpc-server/src/server.ts
+++ b/packages/xrpc-server/src/server.ts
@@ -203,13 +203,18 @@ export class Server {
         const locals: RequestLocals = req[kRequestLocals]
 
         // run the handler
-        const outputUnvalidated = await handler({
-          params,
-          input,
-          auth: locals.auth,
-          req,
-          res,
-        })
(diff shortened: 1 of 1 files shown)


roj1512:
I need to re-describe this issue because it is returning 404 only for unusual errors and perhaps fix it in the `errorMiddleware` instead.

roj1512:
I highly recommend moving to [Koa](https://github.com/koajs/koa) for to make the asynchronous routes more convenient.

dholms (approved):
Just curious, was this motivated by a falsy value getting thrown somewhere in our codebase, or just preemptively heading that off?

This looks good to me 👍 
I'll run tests & we can see about merging

roj1512:
No, this code base didn’t trigger that issue. Once I saw you’re using Express, I immediately thought that there might be a problem with the async routes as it does not properly support async routes compared to something like Koa, for example.

PR #896, opened 2023-04-25
Finalize support for new moderation report reasons

devinivy:
Ideally in the future codegen will help us with these kinds of unions— this should conclude some of the "manual" work to support the new moderation report types.

### DIFF
--- a/packages/bsky/src/api/com/atproto/moderation/util.ts
+++ b/packages/bsky/src/api/com/atproto/moderation/util.ts
@@ -13,6 +13,10 @@ import {
 import {
   REASONOTHER,
   REASONSPAM,
+  REASONMISLEADING,
+  REASONRUDE,
+  REASONSEXUAL,
+  REASONVIOLATION,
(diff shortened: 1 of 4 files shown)


bnewbold:
@devinivy when I did a grep of the codebase for REASONSPAM, this database table line came up:
https://github.com/bluesky-social/atproto/blob/main/packages/pds/src/db/tables/moderation.ts#L46

I'm not sure if that comes through as an ENUM, or if that behavior is different in sqlite or postgresql (hopefully not an issue!)

there are also some existing tests the new values could slot in to fairly easily, though i'm not sure that is blocking

bnewbold:
oh, I see, you do update that table def

devinivy:
Yep yep!  It's just a typedef used at the app level— in the database these are varchars rather than postgres enums.  So I think we should be all good.

bnewbold:
LGTM. added @dholms as a reviewer if a more formal lookover needed

dholms (approved):
LGTM 👍 just wondering if we want a "reason illegal" in here

PR #854, opened 2023-04-20
Add labels onto record embeds

devinivy:
This work ensures labels are hydrated onto record embeds.

PR #897, opened 2023-04-25
Appview proxying bugfixes

dholms:
Two main ones here:
- proxy `feed.getPosts`
- when fetching mutes, handle the case of an empty array


devinivy:
@dholms if you merge `main` back into this, I think CI should be happy again.  It broke when #854 merged but has been patched-up.

devinivy (approved):

PR #898, opened 2023-04-25
Support backfill into bsky appview

devinivy:
The bsky appview can now be configured so that, when it receives an outdated cursor from `subscribeRepos`, it backfills all repository data from the repo host before continuing to consume events.


devinivy:
Superceded by #1224 #1181

PR #906, opened 2023-04-26
Sequencer race condition

dholms:
We were hitting a race condition in the sequencer.

At its core this had to do with the fact that a postgres SERIAL number is decided on _insert_, but queryable on _commit_.

A typical race would go
- repoA makes a write, they insert into repo_seq & get sequence number 100
- repoB makes a write, they insert into repo_seq & get sequence number 101
- repoB tx commits & notifies outgoing streams of new evt
- outgoing streams check DB, see grab event 101, send it out & update their `lastSeen` to 101
- repoA commits & notifies outing streams of a new evt
- outgoing streams check DB, see evt 101, & don't send out evt 100

We apply two fixes, one on write & one on read:

On write, we sequence writes as _the very last thing_ in the tx, instead of trying to parallelize with other writes in the tx. This should make the race condition occur much less often

On read, we:
- grab recent events
- if each event is 1 higher than the lastSeen, we know there is nothing waiting, so we send them out
- if there are gaps, it could be a race condition or a tx rollback
  - in this case, we wait 50ms in the hopes an tx will land, retry the range again & output

(yes this is _a lil_ hacky, should work 99.9% of time, but may want to make more sophisticated eventually)


bnewbold:
I would consider bumping the 50ms a bit? or multiple attempts with 10ms gaps up to 250ms or something? trying to imagine loaded-server kernel concurrency tempo and just throwing out a random number. just rambling, probably not super important.

I bet the "move SERIAL to end of TX" will almost 100% solve this, I should have recommended that when we chatted. last time I had a problem like this I did that as a partial mitigation and it reduce the problem to ones-per-year, even with high and spike-y database load.

dholms:
yeah tbh i had thought it _was_ at the end of the tx. but turns out i was wrong.

I think just the writes change will help in almost every case. My worry with bumping it up too much is that we'll most often hit this with ROLLBACKs & so we're just gonna be unnecessarily waiting & hitting the DB most of the time 🤔 

Wait about up to 2 DB hits with a 50ms wait between each?

devinivy (approved):
Looks great!

PR #909, opened 2023-04-27
Improve readability in `instanceof` assertions

roj1512:


dholms (approved):
Yup yup looks good 👍 

Thanks for the helpful PRs

PR #913, opened 2023-04-27
Parse URLs after sentence dot corectly

KevSlashNull:
This fixes the parsing of URLs that are prefixed by a dot (`.`), i.e. when the user forgot a space between the end of a sentence and a link.

image

The fix is done by preventing the second alternative capture group from having a domain end in `.https://` (which would be an invalid domain anyway), since that would capture `out.https` in the example message `Check this out.https://afterdot.com`.

### DIFF
--- a/packages/api/src/rich-text/detection.ts
+++ b/packages/api/src/rich-text/detection.ts
@@ -34,7 +34,7 @@ export function detectFacets(text: UnicodeString): Facet[] | undefined {
   {
     // links
     const re =
-      /(^|\s|\()((https?:\/\/[\S]+)|((?<domain>[a-z][a-z0-9]*(\.[a-z0-9]+)+)[\S]*))/gim
+      /(^|\s|\(|\b)((https?:\/\/[\S]+)|((?<domain>[a-z][a-z0-9]*(?!\.https?:\/\/)(\.[a-z0-9]+)+)[\S]*))/gim
     while ((match = re.exec(text.utf16))) {
       let uri = match[2]
       if (!uri.startsWith('http')) {
--- a/packages/api/tests/rich-text-detection.test.ts
+++ b/packages/api/tests/rich-text-detection.test.ts
@@ -31,6 +31,8 @@ describe('detectFacets', () => {
     'start middle https://end.com/foo/bar?baz=bux#hash',
     'https://newline1.com\nhttps://newline2.com',
     '👨‍👩‍👧‍👧 https://middle.com 👨‍👩‍👧‍👧',
+    'Check this out.https://afterdot.com',
+    'Check this out!https://afterdot.com',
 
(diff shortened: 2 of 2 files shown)


dholms (approved):
this looks reasonable to me, but regex is not my forte

gonna run tests & ping in @devinivy or @pfrazee for a second look

devinivy:
Tested this out and it seems to work as advertised, nice 👍 TBH would be nice to get signoff from Paul, more so on the behavior than the implementation.

pfrazee:
At a glance looks fine, let me get back to this after I get some FE review

jesopo:
while we're here, there's an edge case i came across:

```json
{
  "uri": "at://did:plc:wmawa5n2bfypw36plteupuzl/app.bsky.feed.post/3jvdg2kaydd2f",
  "cid": "bafyreigi7bc4duhsmcnxteifzdchuqinhyk5iw7u3z5243un3wnjrndghe",
  "value": {
    "text": "(https://test.com) \"(https://test.com)\"",
    "$type": "app.bsky.feed.post",
    "facets": [
      {
        "index": {
          "byteEnd": 17,
          "byteStart": 1
        },
        "features": [
          {
            "uri": "https://test.com",
            "$type": "app.bsky.richtext.facet#link"
          }
        ]
      },
      {
        "index": {
          "byteEnd": 39,
          "byteStart": 21
        },
        "features": [
          {
            "uri": "https://test.com)\"",
            "$type": "app.bsky.richtext.facet#link"
          }
        ]
      }
    ],
    "createdAt": "2023-05-09T23:20:41.499Z"
  }
}
```

in the case of `"(thing)"`, because the closing paren is not the very last character, it skips over it, and includes both the closing paren and the speech mark in the url

KevSlashNull:
@jesopo, yep good catch! I can take a look once this is merged.

The regex matches URL paths with `[\S]*` right now, which we need to narrow down. We can’t expect users to give us [RFC 3986-encoded URIs](https://www.rfc-editor.org/rfc/rfc3986#section-2.3) but we can make educated guesses, such as:

- `https://test.com)"` should obviously match out to `https://test.com` as link because `)"` aren’t part of the path.
- For links with parenthesis in the path, the parenthesis only belongs to a link if we find a pair, so that `https://test.com/(thing)` is a full link but `https://test.com/thing(` is not.
- A link never contains a double-quote, since it is much more likely to be used as punctuation in the post.

A boring solution would be to follow something like the [GFM extended autolinks](https://github.github.com/gfm/#extended-autolink-path-validation).

dholms:
bump @pfrazee

pfrazee:
(Tests certainly look right)

KevSlashNull:
Closing due to inactivity.

PR #917, opened 2023-04-27
Limit concurrency of pds background tasks

devinivy:


dholms (approved):
yup yup 👍

packages/pds/src/event-stream/background-queue.ts
```diff
@@ -5,10 +5,13 @@ import { dbLogger } from '../logger'
 // A simple queue for in-process, out-of-band/backgrounded work
 
 export class BackgroundQueue {
-  queue = new PQueue()
+  queue = new PQueue({ concurrency: 10 })
   constructor(public db: Database) {}
 
   add(task: Task) {
+    if (this.queue.isPaused) {
```
dholms, line 12: wait actually! can we log here?

dholms: just so we know when something is skipped during shutdown

devinivy: Will add this in soon with task names 👍

PR #918, opened 2023-04-27
Fix NSID name length error message typo

ryanwinchester:
The name length limit is 128 not 127.

### DIFF
--- a/packages/nsid/src/index.ts
+++ b/packages/nsid/src/index.ts
@@ -85,7 +85,7 @@ export const ensureValidNsid = (nsid: string): void => {
       throw new InvalidNsidError('NSID domain part too long (max 63 chars)')
     }
     if (l.length > 128 && i + 1 == labels.length) {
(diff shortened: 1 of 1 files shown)


bnewbold:
I feel like there is something subtle here with the length with/without a period. Or maybe the length should be 127 not 128? But also not sure it matters and maybe 128 is what we should return/document.

ryanwinchester:
> I feel like there is something subtle here with the length with/without a period. Or maybe the length should be 127 not 128? But also not sure it matters and maybe 128 is what we should return/document.

There are multiple places where it also checks `253 + 1 + 128` for max length.

Where: `253` is the [max length of a domain](https://en.wikipedia.org/wiki/Domain_name#Domain_name_syntax), `1` is the dot, and `128` is the max length of the name.

Screenshot 2023-04-29 at 11 44 18 PM

Since the name part is assumed to be `128` everywhere else in the code, except this one place, I think it's a safe bet that it's a typo.

devinivy (approved):
Thanks!

PR #919, opened 2023-04-27
Background notification creation and deletion

devinivy:
Notifications can be created out-of-band from record writes.

### DIFF
--- a/packages/pds/src/app-view/services/indexing/processor.ts
+++ b/packages/pds/src/app-view/services/indexing/processor.ts
@@ -199,17 +199,27 @@ export class RecordProcessor<T, S> {
         op.inserted ?? null,
       )
       if (forDelete.toDelete.length > 0) {
-        await this.db
-          .deleteFrom('user_notification')
-          .where('recordUri', 'in', forDelete.toDelete)
-          .execute()
+        // Notifs can be deleted in background: they are expensive to delete and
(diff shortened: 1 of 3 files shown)


packages/pds/src/app-view/services/indexing/processor.ts
```diff
@@ -199,17 +199,27 @@ export class RecordProcessor<T, S> {
...
       }
       notifs = forDelete.notifs
     } else if (op.inserted) {
       notifs = this.params.notifsForInsert(op.inserted)
     }
     if (notifs.length > 0) {
-      await this.db.insertInto('user_notification').values(notifs).execute()
+      this.appDb.onCommit(() => {
+        this.backgroundQueue.add(async (db) => {
+          await db.db.insertInto('user_notification').values(notifs).execute()
```
dholms, line 220: do we def wanna background inserts? 

they seem more important to not drop than deletes & I don't think they're that slow, are they?

devinivy: I'll update the queue so that the queued items should drain a bit more happily.

devinivy: I actually went back on this, and ended-up doing what you suggested 👍

dholms (approved):

dholms (approved):

PR #926, opened 2023-04-29
pds: load server key from file

zhuowei:
Add an environmental variable to the PDS demo server to load and store the server key in JSON format.


dholms:
Going to close this. 

We never use the `bin` file & it's gotten outdated. I will likely remove the file entirely as it does not have best practices for running a service. 

We encourage users that want to run a service directly from source to instead look at https://github.com/bluesky-social/atproto/tree/main/packages/pds/service

PR #928, opened 2023-04-29
Fix typo in NSID validation

MarshalX:
according to this if statement https://github.com/bluesky-social/atproto/blob/01b5971907dd63a04553c65dce3f334fc3ae8e5f/packages/nsid/src/index.ts#L87 and regular expression at https://github.com/bluesky-social/atproto/blob/01b5971907dd63a04553c65dce3f334fc3ae8e5f/packages/nsid/src/index.ts#L104

the `InvalidNsidError` exception should says that the max length is 128 chars instead of 127


MarshalX:
duplicate of https://github.com/bluesky-social/atproto/pull/918

PR #929, opened 2023-04-29
More aggressive backpressure on websocket writer

devinivy:
This adds some more aggressive backpressure on the websocket stream writer for xrpc subscriptions: we wait to drain each message before piling up additional messages, in case bad network causes a condition such as bad tcp buffering.

### DIFF
--- a/packages/pds/Dockerfile
+++ b/packages/pds/Dockerfile
@@ -46,7 +46,7 @@ ENV NODE_ENV=production
 
 # https://github.com/nodejs/docker-node/blob/master/docs/BestPractices.md#non-root-user
 USER node
-CMD ["node", "--enable-source-maps", "index.js"]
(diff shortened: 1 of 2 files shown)


packages/pds/Dockerfile
```diff
@@ -46,7 +46,7 @@ ENV NODE_ENV=production
 
 # https://github.com/nodejs/docker-node/blob/master/docs/BestPractices.md#non-root-user
 USER node
-CMD ["node", "--enable-source-maps", "index.js"]
```
dholms, line 49: this is temporary right? just so we can get better diagnostics?

dholms (approved):
sweet let's give it a go 👍

packages/pds/Dockerfile
devinivy: It doesn't have to stick around, but actually I don't know of any downsides of keeping it!

PR #931, opened 2023-04-30
Prevent app pass sessions deleting app passwords

KevSlashNull:
This fixes that sessions created with an app password can delete itself and other app passwords. Now only `Access` sessions (i.e. those authenticated with a user password) can delete them.

@dholms, this was one of the issues I reported via email but I saw it already [discussed on bsky](https://staging.bsky.app/profile/mattn.bsky.social/post/3jubuwn3kba2i), so yeah here’s a fix.

### DIFF
--- a/packages/pds/src/api/com/atproto/server/revokeAppPassword.ts
+++ b/packages/pds/src/api/com/atproto/server/revokeAppPassword.ts
@@ -3,7 +3,7 @@ import { Server } from '../../../../lexicon'
 
 export default function (server: Server, ctx: AppContext) {
   server.com.atproto.server.revokeAppPassword({
-    auth: ctx.accessVerifier,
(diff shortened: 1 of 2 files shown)


dholms:
As the discussion on bsky mentioned, there is additional scoping work to be done on auth in general, but also arguably on app password.

The purpose of app passwords was solely to allow users to log in to alternate clients without entering their account password.

Deleting other app passwords is inconvenient for sure, but it is actually one of the least inconvenient things that an app with your credentials can do (such as creating/deleting records on your behalf)

Right now we limit app passwords from possible auth loopholes (such as creating additional app passwords) & from very destructive actions (such as account deletion).

I'm going to close this for now, but may reopen when we get around to adding scopes to our auth system.

PR #941, opened 2023-05-01
pds: iterated Hive label parsing

bnewbold:
Tune "porn", "sexual", and "nudity" label categorization, including re-structuring of code paths.

This comes out of some image testing as well as review of Hive API docs.

### DIFF
--- a/packages/pds/src/labeler/hive.ts
+++ b/packages/pds/src/labeler/hive.ts
@@ -73,21 +73,74 @@ export const respToClasses = (res: HiveResp): HiveRespClass[] => {
   return classes
 }
 
-// sexual: https://docs.thehive.ai/docs/sexual-content
+// Matches only one (or none) of: porn, sexual, nudity
+//
+// porn: sexual and nudity. including both explicit activity or full-frontal and suggestive/intent
+// sexual: sexually suggestive, not explicit; may include some forms of nudity
+// nudity: non-sexual nudity (eg, artistic, possibly some photographic)
+//
+// hive docs/definitions: https://docs.thehive.ai/docs/sexual-content
+export const sexualLabels = (classes: HiveRespClass[]): string[] => {
+  const scores = {}
+
+  for (const cls of classes) {
+    scores[cls.class] = cls.score
+  }
+
+  // first check if porn...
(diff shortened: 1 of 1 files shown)


devinivy (approved):
💎 ace

packages/pds/src/labeler/hive.ts
```diff
@@ -73,21 +73,74 @@ export const respToClasses = (res: HiveResp): HiveRespClass[] => {
...
+    return ['sexual']
+  }
+  if (scores['yes_undressed'] >= 0.9) {
+    // special case for bondage examples
+    if (scores['yes_sex_toy'] > 0.75) {
+      return ['sexual']
+    }
+  }
+
+  // then finally non-sexual nudity...
```
dholms, line 122: wait this does imply that the order of "severity" is
porn > sexual > nudity

whereas i've been thinking of it as 
porn > nudity > sexual

it may be worthwhile for "sexual" to exist alongside "nudity"? instead of being exlclusive

dholms (approved):
besides the comment i left looks good!

probably okay to roll this out & if it leads to unexpected outcomes, reassess

packages/pds/src/labeler/hive.ts
bnewbold: yeah, I spent a bunch of time thinking about this last night, and comments at the top of that function.

basically `nudity` will be more narrowly "artistic nudity". meaning you should be able to block porn and sexual in config but just-show nudity and not have dick picks or horny poasting show up in feed as much.

porn: explicit stuff, and "sexual nudity" (most dick pics). includes animated sex stuff, and even Renaissance porn drawings (if actually porn-y)

sexual: mostly horny poasting, and porn-like stuff with cover-up (eg, nude with hands covering, if not artistic)

nudity: almost all just artistic nudity; potentially non-sexual photographic nudity like nudist colony stuff, or textbook nudity, but that usually actually gets flagged as "sexual" and ends up in other categories

bnewbold: this is mostly driven by what is possible with Hive categories, and a bit of real world experience now, not from first-principles or humanist values or anything.

PR #945, opened 2023-05-01
Tweak whats hot

dholms:
This tweaks what's hot
- formalizes a hot fix that increased like threshold to 12
- filters out content from spicier labels (mostly porn & gore)
- filters out `!no-promote` content. Does not forward `!no-promote` labels to client since client does not handle it yet

### DIFF
--- a/packages/bsky/src/services/label/index.ts
+++ b/packages/bsky/src/services/label/index.ts
@@ -6,6 +6,10 @@ import { sql } from 'kysely'
 
 export type Labels = Record<string, Label[]>
 
+// @TODO forward these labels to client
+// these are labels are currently only used server side & not sent to client since it does not handle them correctly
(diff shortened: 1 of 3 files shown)


packages/bsky/src/services/label/index.ts
```diff
@@ -6,6 +6,10 @@ import { sql } from 'kysely'
 
 export type Labels = Record<string, Label[]>
 
+// @TODO forward these labels to client
+// these are labels are currently only used server side & not sent to client since it does not handle them correctly
+const SERVER_SIDE_LABELS = ['!no-promote']
```
devinivy, line 11: Also may be worth a little note that `SERVER_SIDE_LABELS` cannot be empty (otherwise query below would fail).

devinivy (approved):

PR #950, opened 2023-05-01
test (api, common, common-web): add further tests

43081j:
just getting my head around the repo out of curiosity, so i filled in some test gaps while i was there

let me know if there's anything you want changing, and if you want this

fwiw too the agent tests could be better to purge the in-memory db between tests. you spin up a single db before the entire suite presumably to save time/resources. however, it means tests can pollute each others' state, and you end up with these ever-incrementing user IDs in your test cases to avoid conflicts.

would be much better if you could `afterEach(() => server.ctx.db.clear())` somehow

### DIFF
--- a/packages/api/tests/agent.test.ts
+++ b/packages/api/tests/agent.test.ts
@@ -398,4 +398,67 @@ describe('agent', () => {
     expect(sessions.length).toEqual(1)
     expect(sessions[0]?.accessJwt).toEqual(origAccessJwt)
   })
+
+  describe('setPersistSessionHandler', () => {
+    it('sets persist session handler', async () => {
+      let originalHandlerCallCount = 0
+      let newHandlerCallCount = 0
+
+      const persistSession = () => {
+        originalHandlerCallCount++
+      }
+      const newPersistSession = () => {
+        newHandlerCallCount++
+      }
+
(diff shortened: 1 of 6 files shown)


dholms (approved):
Sweet love it! 

Thanks for doing this. been meaning to get some `common` tests in for awhile, but haven't had the bandwidth

re the db setup & teardown - I think we'll leave it as is for now. We hope to refactor a lot of our tests soon to make them less-integrationy & more lightweight to run. But for now, I think we'll stick with the "1 db instance per suite" that we have going on

Gonna run the testsuite & see if @devinivy has any comments, but I think this is good to go in my book

dholms:
ah some linting errors. would you mind fixing?

think you should be able to fix with `yarn verify:fix`

devinivy:
Read through these and all looked good— appreciate it!

43081j:
Sure no worries, will try push the fix today

& Yeah I did notice the tests felt very integration-like. There's probably a fair amount of stuff you can test and just stub out the underlying server rather than launching a real one. Might be an option in future

43081j:
@dholms should be fixed now, can you approve the ci run again?

dholms:
lost track of this one - thanks for the fixes! merging now 👍

PR #951, opened 2023-05-01
Perform account deletion async

devinivy:
Account deletion is expensive to perform during the course of the request, so here we disable the account immediately while the hard-deletion process occurs in the background.  The deletion process has been updated to take place serially (e.g. no concurrent queries) to avoid contention, which works better now that the round-trip time is no longer in the critical path of the user's experience.


dholms (approved):
yup yup yup this looks spot on 🙌

PR #955, opened 2023-05-02
Disallow writes of bad records

devinivy:
The PDS can't reliably transform a record between lex and cbor, it will not allow the record to be written.


dholms (approved):
yup looks good 👍

PR #959, opened 2023-05-02
Delete dereferenced blobs

dholms:
When a blob is dereferenced (no more records from that user pointing to it) during a commit, we remove the blob, delete it from the blobstore (if not shared with another user) & invalidate the blob in the cache.

You'll note that the rebase code as changed - we no longer need to do blob deletes in rebases because we do them proactively in deletes now.

This PR should be coupled with an app migration that deletes all existing dereferenced blobs

### DIFF
--- a/packages/pds/src/services/repo/blobs.ts
+++ b/packages/pds/src/services/repo/blobs.ts
@@ -12,6 +12,7 @@ import Database from '../../db'
 import { Blob as BlobTable } from '../../db/tables/blob'
 import * as img from '../../image'
 import { BlobRef } from '@atproto/lexicon'
+import { PreparedDelete, PreparedUpdate } from '../../repo'
 
 export class RepoBlobs {
   constructor(public db: Database, public blobstore: BlobStore) {}
@@ -55,6 +56,8 @@ export class RepoBlobs {
   }
 
   async processWriteBlobs(did: string, commit: CID, writes: PreparedWrite[]) {
+    await this.deleteDereferencedBlobs(did, writes)
+
     const blobPromises: Promise<void>[] = []
     for (const write of writes) {
       if (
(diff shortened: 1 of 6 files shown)


packages/pds/src/services/repo/blobs.ts
```diff
@@ -70,6 +80,74 @@ export class RepoBlobs {
...
+    const updates = writes.filter(
+      (w) => w.action === WriteOpAction.Update,
+    ) as PreparedUpdate[]
+    const uris = [...deletes, ...updates].map((w) => w.uri.toString())
+    if (uris.length === 0) return
+
+    const deletedRepoBlobs = await this.db.db
+      .deleteFrom('repo_blob')
+      .where('did', '=', did)
+      .where('recordUri', 'in', uris)
```
devinivy, line 96: Based on the tests I am pretty sure this must be working, but what's the right way to think about updates that preserve a blob reference?  Naively it looks like the repo_blobs are being deleted for those updates here, pretty sure I am just missing something!

packages/pds/src/services/repo/blobs.ts
```diff
@@ -70,6 +80,74 @@ export class RepoBlobs {
...
+    await Promise.all([
+      ...blobsToDelete.map((cid) => this.blobstore.delete(CID.parse(cid))),
+      ...blobsToDelete.map((cid) => {
+        const paths = ImageUriBuilder.commonSignedUris.map((id) => {
+          const uri = this.imgUriBuilder.getCommonSignedUri(id, cid)
+          return uri.replace(this.imgUriBuilder.endpoint, '')
+        })
+        return this.imgInvalidator.invalidate(cid, paths)
+      }),
+    ])
```
devinivy, line 148: Do all these happen during the course of a transaction?  I ask because they may take a long time.  If so, maybe we could pull them out using `onCommit()` and plop them on the background queue.

packages/pds/src/services/repo/blobs.ts
dholms: oh wait think i misinterpreted question: this will delete the repo_blob linkage, (but not the blob itself) & then that repo_blob linkage gets set back up in the rest of `processWriteBlobs`

the commit value needs to be updated anyway, so we just delete & re-insert

packages/pds/src/services/repo/blobs.ts
dholms: that's a good call. i'm down to move to background queue

PR #963, opened 2023-05-03
Delete blobs app migration

dholms:
Building off of https://github.com/bluesky-social/atproto/pull/959

### DIFF
--- a/packages/pds/src/app-migrations/delete-derefed-blobs.ts
+++ b/packages/pds/src/app-migrations/delete-derefed-blobs.ts
@@ -1,0 +1,79 @@
+import AppContext from '../context'
+import { chunkArray } from '@atproto/common'
+import { CID } from 'multiformats/cid'
+import { ImageUriBuilder } from '../image/uri'
+
+async function main(ctx: AppContext) {
+  const db = ctx.db
+  const { ref } = db.db.dynamic
+  const deletedRepoBlobs = await db.db
+    .deleteFrom('repo_blob')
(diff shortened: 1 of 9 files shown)


packages/pds/src/app-migrations/delete-derefed-blobs.ts
```diff
@@ -0,0 +1,122 @@
...
+        .selectFrom('record')
+        .whereRef('record.uri', '=', ref('repo_blob.recordUri'))
+        .selectAll(),
+    )
+    .returningAll()
+    .execute()
+}
+
+async function derefedProfiles(dbTxn: Database) {
+  dbTxn.assertTransaction()
```
devinivy, line 81: We might find this is heavy for a single transaction.  If that's an issue, even just doing it in batches by did might do the trick.

devinivy, line 81: We might find this is heavy for a single transaction.  If that's an issue, even just doing it in batches by did might do the trick.

devinivy (approved):
Left one minor thought, but the approach looks solid 👍

packages/pds/src/app-migrations/delete-derefed-blobs.ts
dholms: Do you think we even need to use transactions here?

I actually don't see much that needs to be done transactionally in this 🤔

dholms:
successfully ran this

PR #967, opened 2023-05-03
Block lists

dholms:
This implements publishable & subscribe-able block lists.

We introduce three new record types - all in the `app.bsky.graph` namespace:
- `list`: a declaration of a list of users. includes name, description, & an avatar
- `listitem`: a user to be considered a member of some list, along with an optional reason for their inclusion
- `listblock`: a record that applies a block to all users in a given list

When a user creates a `listblock` record directed at a list of users, the impact is the same is if they had created a `block` record themselves on each user.

We define lists in the abstract so that they may be used for other purposes such as:
- mass follows
- views of some set of actors (a la twitter lists)
- inputs into an algorithm

Three new XRPC queries are defined - also in the `app.bsky.graph` namespace:
- `getList`: returns information about a list along with a paginated view of the items in the list
- `getLists`: returns information about an actor along with a paginated view of all the lists they've created & own
- `getListBlocks`: a method that, similar to `getBlocks` may only be called by a user about themselves. It returns a paginated view of all the lists that they are currently using for blocks


lexicons/app/bsky/graph/listitem.json
```diff
@@ -0,0 +1,20 @@
...
+    "main": {
+      "type": "record",
+      "description": "An item under a declared list of actors",
+      "key": "tid",
+      "record": {
+        "type": "object",
+        "required": ["subject", "list", "createdAt"],
+        "properties": {
+          "subject": {"type": "string", "format": "did"},
+          "list": {"type": "string", "format": "at-uri"},
```
pfrazee, line 14: Can we get an optional `reason` field here which is an object of a text string (max 1k chars) and a richtext facets? We want users to be able to explain why somebody is on list

dholms: yeah sure 👍

lexicons/app/bsky/actor/defs.json
```diff
@@ -85,6 +85,28 @@
         "following": {"type": "string", "format": "at-uri"},
         "followedBy": {"type": "string", "format": "at-uri"}
       }
+    },
+    "listView": {
+      "type": "object",
+      "required": ["name"],
+      "properties": {
+        "name": {"type": "string", "maxLength": 64, "minLength": 1},
+        "description": {"type": "string", "maxGraphemes": 300, "maxLength": 3000},
```
pfrazee, line 94: ah snap I just realized description needs facets too

dholms:
closing in favor of https://github.com/bluesky-social/atproto/pull/986

PR #974, opened 2023-05-04
Increase flexibility of paramaters accepted by DB_CREDS_JSON env var

mileszim:
## Summary

- Added default values for the `username`, `password`, `host`, `port`, and `database` parameters of the `pgUrl` function.
- Added an optional `sslmode` parameter to the `pgUrl` function.
- These default values are intended to ensure that the `pgUrl` function can still connect to the PostgreSQL database even if not all the parameters are provided.

### 1. With original options:

```json
{
  "username": "myuser",
  "password": "mypassword",
  "host": "localhost",
  "port": "5432"
}
```

Output:

```
postgresql://myuser:mypassword@localhost:5432/postgres
```

### 2. With new options including `sslmode`:

```json
{
  "username": "myuser",
  "password": "mypassword",
  "host": "localhost",
  "port": "5432",
  "database": "mydb",
  "sslmode": "require"
}
```

Output:

```
postgresql://myuser:mypassword@localhost:5432/mydb?sslmode=require
```

### DIFF
--- a/packages/pds/service/index.js
+++ b/packages/pds/service/index.js
@@ -78,9 +78,9 @@ const main = async () => {
   })
 }
 
-const pgUrl = ({ username, password, host, port }) => {
+const pgUrl = ({ username = "postgres", password = "postgres", host = "0.0.0.0", port = "5432", database = "postgres", sslmode }) => {
   const enc = encodeURIComponent
-  return `postgresql://${username}:${enc(password)}@${host}:${port}/postgres`
+  return `postgresql://${username}:${enc(password)}@${host}:${port}/${database}${sslmode ? `?sslmode=${enc(sslmode)}` : ''}`
 }
 
 const smtpUrl = ({ username, password, host }) => {


packages/pds/service/index.js
```diff
@@ -78,9 +78,9 @@ const main = async () => {
   })
 }
 
-const pgUrl = ({ username, password, host, port }) => {
+const pgUrl = ({ username = "postgres", password = "postgres", host = "0.0.0.0", port = "5432", database = "postgres", sslmode }) => {
```
devinivy, line 81: Is there a reason we should be using `0.0.0.0` rather than `localhost` or `127.0.0.1`?  I guess I usually think of `0.0.0.0` more as a source address than a destination.

devinivy:
I am down to land this!  Just a headsup that we'll probably also be tidying-up the PDS config in the coming weeks.

packages/pds/service/index.js
mileszim: no particular reason. localhost or 127.0.0.1 is probably the better default

devinivy: Cool!
```suggestion
const pgUrl = ({ username = "postgres", password = "postgres", host = "localhost", port = "5432", database = "postgres", sslmode }) => {
```

ianklatzco: It might not be relevant in this context but worth noting -- there's a meaningful semantic difference between 0.0.0.0 and 127.0.0.1. 

0.0.0.0 will bind to some definition of all interfaces - e.g. eth0, lo0.

127.0.0.1 will only bind to lo0/loopback.

i.e. it's possible to accidentally exposure services intended to only be accessible from loopback, if 0.0.0.0 is used where 127.0.0.1 should be.

It looks like localhost is just a DNS name for 127.0.0.1.

[sample stackoverflow answer](https://stackoverflow.com/a/20778887/1234621) w/ more info.

mileszim: whoa, good to know! thanks for taking the time to explain all of this

dholms (approved):
let's do it 👍

PR #977, opened 2023-05-04
convert punycode domain handles to ASCII

Evalprime:
punycode domains fail under the check in ensureValidHandle(). this pr converts them to ASCII in the normalization function first before validating them, using Mathias Bynens' RFC 3492 and RFC 5891 compliant library.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -9,3 +9,5 @@ ATProto receives so many contributions that we could never list everyone who des
 #### [ianklatzco](https://github.com/ianklatzco), Security disclosure, April 2023
 
 #### lily, Security disclosure, May 2023
+
+#### [april](https://github.com/april), Security disclosure, May 2023
--- a/lexicons/com/atproto/admin/defs.json
+++ b/lexicons/com/atproto/admin/defs.json
@@ -138,7 +138,8 @@
         "relatedRecords": {"type": "array", "items": {"type": "unknown"}},
         "indexedAt": {"type": "string", "format": "datetime"},
         "moderation": {"type": "ref", "ref": "#moderation"},
-        "invitedBy": {"type": "ref", "ref": "com.atproto.server.defs#inviteCode"}
+        "invitedBy": {"type": "ref", "ref": "com.atproto.server.defs#inviteCode"},
+        "invitesDisabled": {"type": "boolean"}
       }
     },
     "repoViewDetail": {
@@ -159,7 +160,8 @@
         "invites": {
           "type": "array",
           "items": {"type": "ref", "ref": "com.atproto.server.defs#inviteCode"}
-        }
+        },
+        "invitesDisabled": {"type": "boolean"}
       }
     },
     "repoRef": {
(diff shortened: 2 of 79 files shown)


devinivy:
Would you mind adding tests to help illustrate what cases this covers?

Also, is it possible that we can do this without an additional dependency using the WHATWG `URL` interface?  I believe this code ends-up running in web clients, so want to be a little careful with adding dependencies that may increase bundle size or not be web-compatible (cc @pfrazee in case you're interested).

roj1512:
Yes, the `URL` class can be used.

image

Evalprime:
I'll rewrite using URL, thanks!

Evalprime:
@roj1512 the URL class requires a valid URL, are the handles in ensureValidHandle() always going to be domains or could this potentially error?

roj1512:
@Evalprime I believe the values passed to `ensureValidHandle()` are always FQDNs (correct me if I’m wrong, @devinivy), so you could do this:

```ts
new URL(`https://${handle}`).hostname
```

roj1512 (approved):
Apparently LGTM.

Evalprime:
should be good now

Evalprime:
I'm not sure why tests are failing for a completely different part of the pds now

roj1512:
Consider merging with the upstream?

bnewbold:
I think we should hold off on this for a while. Clients that want to support IDNs can do so entirely client-side, both for creation and display. I don't think we should pull punycode parsing and semantics in to the protocol directly; the protocol (and API) should operate at the lower-level DNS record format (ASCII).

We should also do an audit of both security and UI/UX concerns before we do IDN display in our app and web interfaces, as well. IDNs are great, and we should be able to support them well in the long run, and we have laid the groundwork to do so in the protocol. But we are juggling a lot and should be careful with this one.

dholms:
Hmm these methods are only used on parameters/inputs into the PDS 🤔 

Since the handles are normalized & indexed as such, they'll be returned as ASCII in views. And this could cause confusion if, for instance, a DID doc handle is in punycode & the indexed handle on some service is in ASCII.

I agree with @bnewbold that handles should be kept as ASCII. This same normalization can be applied just as easily on client-side & punycode handles could also be rendered as such by a client that supports them.

bnewbold:
Tangentially, this thread made me wonder what the deal is with IDNs in `did:web`. It looks like the answer is "no Unicode in DIDs at all". One of the concerns with IDNs in DIDs would be having two representations of the same DID, which would require a bunch of logic to safely detect equality, uniqueness, etc.

https://w3c-ccg.github.io/did-method-web/#international-domain-names

Evalprime:
it might be better to represent domains as ASCII anyway to prevent someone using a punycode character to impersonate google.com or any other website, no?

devinivy:
It looks like there's some consensus here and this is good to close.  If that doesn't seem right just say the word and we can reopen.

roj1512:
Would be nice to have an issue/PR/discussion open/unresolved to keep track of the implementation of punycode support.

bnewbold:
@roj1512 we will be mentioning IDNs/punycode explicitly in our docs soon

PR #980, opened 2023-05-04
Tighten-up experience of blocks by third parties

devinivy:


packages/pds/src/app-view/api/app/bsky/feed/getPostThread.ts
```diff
@@ -91,6 +91,7 @@ const composeThread = (
   }
 
   if (post.author.viewer?.blocking || post.author.viewer?.blockedBy) {
+    // Globally blocked replies via post.replyBlocked are fully filtered out of results.
```
dholms, line 94: is it possible that we could still return as a `#blockedPost`?

seems like this could lead to unexpected behavior 🤔

packages/pds/src/db/util.ts
```diff
@@ -42,6 +42,10 @@ export const valuesList = (vals: unknown[]) => {
   return sql`(values (${sql.join(vals, sql`), (`)}))`
 }
 
+export const dbBool = (bool: boolean): 0 | 1 => {
```
dholms, line 45: nice

packages/pds/src/db/migrations/20230504T210229992Z-block-third-party.ts
```diff
@@ -0,0 +1,112 @@
...
+
+export async function down(db: Kysely<unknown>): Promise<void> {
+  await db.schema.alterTable('post').dropColumn('replyBlocked').execute()
+  await db.schema
+    .alterTable('post_embed_record')
+    .dropColumn('embedBlocked')
+    .execute()
+}
+
+export function dataMigrationQbs(db: Kysely<Schema>) {
```
dholms, line 22: where does this get run? is it an app migration?

packages/pds/src/app-view/api/app/bsky/feed/getPostThread.ts
dholms: i almost wonder if it's worth having two different defs for these like `blockedUserPost` & `blockNonComplyingPost`

(names are kinda longwinded but just to get the point across)

packages/pds/src/db/migrations/20230504T210229992Z-block-third-party.ts
devinivy: It is effectively an app migration, but it is simple enough to be run manually.  So I just documented it here for good measure.  I can also just wrap it into an app migration, maybe that's a good idea.

packages/pds/src/app-view/api/app/bsky/feed/getPostThread.ts
dholms: I really wanna get in that fix to #threadViewPost, we should coordinate with paul on how we might do it

dholms (approved):
alright you convinced me not to worry about the replies 😅 

lgtm 👍

dholms (approved):
alright you convinced me not to worry about the replies 😅 

lgtm 👍

packages/pds/src/db/migrations/20230504T210229992Z-block-third-party.ts
devinivy: Ahh I started pulling these out into an app migration but the more I think about it, I think it's best they just be run manually.  I am going to add a comment here to make it clearer what's going on.

devinivy:
Replaced by #1378

PR #986, opened 2023-05-05
Mute lists

dholms:
This implements publishable & subscribable mute lists.

We introduce three new record types - all in the `app.bsky.graph` namespace:
- `list`: a declaration of a list of users. includes name, description, & an avatar
- `listitem`: a user to be considered a member of some list, along with an optional reason for their inclusion

We define lists in the abstract so that they may be used for other purposes such as:
- blocks
- mass follows
- views of some set of actors (a la twitter lists)
- inputs into an algorithm

Five new XRPC queries are defined - also in the `app.bsky.graph` namespace:
- `getList`: returns information about a list along with a paginated view of the items in the list
- `getLists`: returns information about an actor along with a paginated view of all the lists they've created & own
- `getListMutes`: a method that, similar to `getMutes` may only be called by a user about themselves. It returns a paginated view of all the lists that they are currently using for mutes
- `subscribeMuteList`: mutes all users in a list
- `unsubscribeMuteList`: stops muting all users based on a list


lexicons/app/bsky/graph/defs.json
```diff
@@ -0,0 +1,52 @@
...
+        }
+      }
+    },
+    "listPurpose": {
+      "type": "string",
+      "knownValues": [
+        "app.bsky.graph.defs#modlist"
+      ]
+    },
+    "modlocklist": {
```
pfrazee, line 41: heh I think that's a typo

lexicons/app/bsky/graph/defs.json
```diff
@@ -0,0 +1,64 @@
...
+      "properties": {
+        "uri": {"type": "string", "format": "at-uri"},
+        "creator": {"type": "ref", "ref": "app.bsky.actor.defs#profileView"},
+        "name": {"type": "string", "maxLength": 64, "minLength": 1},
+        "purpose": {"type": "ref", "ref": "#listPurpose"},
+        "description": {"type": "string", "maxGraphemes": 300, "maxLength": 3000},
+          "descriptionFacets": {
+            "type": "array",
+            "items": {"type": "ref", "ref": "app.bsky.richtext.facet"}
+          },
```
devinivy, line 29:
```suggestion
        "descriptionFacets": {
          "type": "array",
          "items": {"type": "ref", "ref": "app.bsky.richtext.facet"}
        },
```

lexicons/app/bsky/graph/defs.json
```diff
@@ -0,0 +1,64 @@
...
+    },
+    "listPurpose": {
+      "type": "string",
+      "knownValues": [
+        "app.bsky.graph.defs#modlist"
+      ]
+    },
+    "modlist": {
+      "type": "token",
+      "description": "A list of actors to apply an aggregate moderation on"
```
devinivy, line 55: Re: "an aggregate moderation"— is the idea to keep this generic so that it could apply to either blocks or mutes?

lexicons/app/bsky/graph/defs.json
```diff
@@ -0,0 +1,64 @@
...
+        "name": {"type": "string", "maxLength": 64, "minLength": 1},
+        "purpose": {"type": "ref", "ref": "#listPurpose"},
+        "avatar": {"type": "string"},
+        "viewer": {"type": "ref", "ref": "#listViewerState"},
+        "indexedAt": {"type": "string", "format": "datetime"}
+      }
+    },
+    "listView": {
+      "type": "object",
+      "required": ["uri", "creator", "name", "purpose"],
```
devinivy, line 19: It's not uncommon for us to mark `indexedAt` as required too.

lexicons/app/bsky/graph/getLists.json
```diff
@@ -0,0 +1,33 @@
...
+  "id": "app.bsky.graph.getLists",
+  "defs": {
+    "main": {
+      "type": "query",
+      "description": "Fetch a list of lists that belong to an actor",
+      "parameters": {
+        "type": "params",
+        "required": ["actor"],
+        "properties": {
+          "actor": {"type": "string", "format": "did"},
```
devinivy, line 12: We've often been using the `at-identifier` format for this kind of parameter (e.g. see `getFollowers`, `getAuthorFeed`).

lexicons/app/bsky/graph/subscribeMuteList.json
```diff
@@ -0,0 +1,20 @@
+{
+  "lexicon": 1,
+  "id": "app.bsky.graph.subscribeMuteList",
```
devinivy, line 3: Not sure if it's a material issue, but flagging for good measure— `subscribe` also may end-up being a prefix for xrpc subscription methods which could be a naming conflict, i.e. `subscribeRepos`.

devinivy: One alternative could be to call these `addMuteList` and `removeMuteList`.

packages/pds/src/api/app/bsky/graph/subscribeMuteList.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export default function (server: Server, ctx: AppContext) {
+  server.app.bsky.graph.subscribeMuteList({
+    auth: ctx.accessVerifier,
+    handler: async ({ auth, input }) => {
+      const { list } = input.body
+      const requester = auth.credentials.did
+
+      await ctx.services.account(ctx.db).subscribeMuteList({
+        list,
```
devinivy, line 12: Do we want any special behavior if the users tries to use a weird `list` at-uri?  E.g. if they attempt to use a post as a list, or something nonsensical like that.

packages/pds/src/app-view/api/app/bsky/feed/getAuthorFeed.ts
```diff
@@ -41,20 +43,19 @@ export default function (server: Server, ctx: AppContext) {
...
-            .where('type', '!=', 'repost')
-            .orWhere('post.creator', 'not in', mutedDidsQb)
-        })
+          qb
+            .where('type', '=', 'post')
+            .orWhereNotExists(
+              accountService.mutedQb(requester, [ref('post.creator')]),
+            ),
+        )
+        .whereNotExists(graphService.blockQb(requester, [ref('post.creator')]))
```
devinivy, line 58: Is the block clause needed here?  I wonder if we may be able to pass on it, since a block between the two users is checked proactively before fetching the author feed.  Or maybe we only need to apply it to reposts, similar to mutes.

lexicons/app/bsky/actor/defs.json
```diff
@@ -80,6 +80,7 @@
       "type": "object",
       "properties": {
         "muted": {"type": "boolean"},
+        "mutedByList": {"type": "ref", "ref": "app.bsky.graph.defs#listViewBasic"},
```
devinivy, line 83: I assume this will be used to give viewers context about why the content was muted, does that sound right?

This _could_ be an array, but fine by me if we just want to surface one list responsible for the blocking.

packages/pds/src/app-view/services/graph/index.ts
```diff
@@ -0,0 +1,126 @@
...
+      .where((qb) =>
+        qb
+          .where('actor_block.creator', '=', requester)
+          .whereRef('actor_block.subjectDid', 'in', sql`(${subjectRefs})`),
+      )
+      .orWhere((qb) =>
+        qb
+          .where('actor_block.subjectDid', '=', requester)
+          .whereRef('actor_block.creator', 'in', sql`(${subjectRefs})`),
+      )
```
devinivy, line 59: These "or"s can create sneaky problems.  If there's a chance a consumer might add additional conditions to the where clause, it could be worth adding some parens around these two conditions.  If that's unlikely, then this should be fine 😎

packages/pds/src/app-view/services/indexing/plugins/list-item.ts
```diff
@@ -0,0 +1,99 @@
...
+  timestamp: string,
+): Promise<IndexedListItem | null> => {
+  const inserted = await db
+    .insertInto('list_item')
+    .values({
+      uri: uri.toString(),
+      cid: cid.toString(),
+      creator: uri.host,
+      subjectDid: obj.subject,
+      listUri: obj.list,
```
devinivy, line 30: I think we'll want to validate that the list item creator is the creator of the list.

packages/pds/src/app-view/services/indexing/plugins/list-item.ts
```diff
@@ -0,0 +1,99 @@
...
+
+const findDuplicate = async (
+  db: DatabaseSchema,
+  uri: AtUri,
+  obj: ListItem.Record,
+): Promise<AtUri | null> => {
+  const found = await db
+    .selectFrom('list_item')
+    .where('creator', '=', uri.host)
+    .where('listUri', '=', obj.list)
```
devinivy, line 52: If we validate that the creator of the list item is the creator of the list, then we can express that a subject should only appear in a list once by checking `(listUri, subjectDid)` rather than `(creator, listUri, subjectDid)`.

packages/pds/src/db/migrations/20230505T201944381Z-mute-lists.ts
```diff
@@ -0,0 +1,71 @@
...
+    .addColumn('uri', 'varchar', (col) => col.primaryKey())
+    .addColumn('cid', 'varchar', (col) => col.notNull())
+    .addColumn('creator', 'varchar', (col) => col.notNull())
+    .addColumn('subjectDid', 'varchar', (col) => col.notNull())
+    .addColumn('listUri', 'varchar', (col) => col.notNull())
+    .addColumn('reason', 'varchar')
+    .addColumn('reasonFacets', 'varchar')
+    .addColumn('createdAt', 'varchar', (col) => col.notNull())
+    .addColumn('indexedAt', 'varchar', (col) => col.notNull())
+    .addUniqueConstraint('list_item_unique_subect_in_list', [
```
devinivy, line 35:
```suggestion
    .addUniqueConstraint('list_item_unique_subject_in_list', [
```

packages/pds/src/db/migrations/20230505T201944381Z-mute-lists.ts
```diff
@@ -0,0 +1,71 @@
...
+      'creator',
+      'listUri',
+      'subjectDid',
```
devinivy, line 38: Made a similar comment in the indexing service— we might be able to narrow this down to `(listUri, subjectDid)` which seems like a natural constraint ("a subject can only be in a list once").

devinivy: (But we still may want an index on `creator` for deletion!)

packages/pds/src/db/migrations/20230505T201944381Z-mute-lists.ts
```diff
@@ -0,0 +1,71 @@
...
+    .addColumn('createdAt', 'varchar', (col) => col.notNull())
+    .addPrimaryKeyConstraint('list_mute_pkey', ['mutedByDid', 'listUri'])
+    .execute()
+
+  // missed index in `actor-block-init` migration
+  await db.schema
+    .createIndex('actor_block_creator_idx')
+    .on('actor_block')
+    .column('creator')
+    .execute()
```
devinivy, line 61: I think `actor_block`'s unique constraint on `(creator, subjectDid)` may actually be doing the trick for us.

packages/pds/src/services/account/index.ts
```diff
@@ -284,6 +290,52 @@ export class AccountService {
...
+  async unsubscribeMuteList(info: { list: string; mutedByDid: string }) {
+    const { list, mutedByDid } = info
+    await this.db.db
+      .deleteFrom('list_mute')
+      .where('listUri', '=', list)
+      .where('mutedByDid', '=', mutedByDid)
+      .execute()
+  }
+
+  mutedQb(requester: string, refs: NotEmptyArray<DbRef>) {
```
devinivy, line 318: Would be awesome to see how this plays on timelines, etc. with some real data.

packages/pds/src/services/account/index.ts
```diff
@@ -284,6 +290,52 @@ export class AccountService {
...
+      .innerJoin('list', 'list.uri', 'list_mute.listUri')
+      .innerJoin('list_item', (join) =>
+        join
+          .onRef('list_item.creator', '=', 'list.creator')
+          .onRef('list_item.listUri', '=', 'list.uri'),
+      )
+      .where('list_mute.mutedByDid', '=', requester)
+      .whereRef('list_item.subjectDid', 'in', sql`(${subjectRefs})`)
+      .select('list_item.subjectDid as muted')
+    return actorMute.union(listMute)
```
devinivy, line 336: If allowing duplicate results doesn't create any problems, using `unionAll` here might be a good look to save the cost of deduping.

packages/pds/src/services/account/index.ts
```diff
@@ -284,6 +290,52 @@ export class AccountService {
...
+      .selectFrom('list_mute')
+      .innerJoin('list', 'list.uri', 'list_mute.listUri')
+      .innerJoin('list_item', (join) =>
+        join
+          .onRef('list_item.creator', '=', 'list.creator')
+          .onRef('list_item.listUri', '=', 'list.uri'),
+      )
+      .where('list_mute.mutedByDid', '=', requester)
+      .whereRef('list_item.subjectDid', 'in', sql`(${subjectRefs})`)
+      .select('list_item.subjectDid as muted')
```
devinivy, line 335: I am really sorry if I keep mentioning this creator/listUri thing and it's way off for some reason that I overlooked 😆 

But I think this is another place it might simplify things.  Pretty sure this could become:
```ts
    const listMute = this.db.db
      .selectFrom('list_item')
      .innerJoin('list_mute', 'list_mute.listUri', 'list_item.listUri')
      .where('list_mute.mutedByDid', '=', requester)
      .whereRef('list_item.subjectDid', 'in', sql`(${subjectRefs})`)
      .select('list_item.subjectDid as muted')
```

lexicons/app/bsky/graph/defs.json
dholms: yup exactly

lexicons/app/bsky/graph/subscribeMuteList.json
dholms: that's a good point 👍  i'll change it

packages/pds/src/app-view/api/app/bsky/feed/getAuthorFeed.ts
dholms: this is actually for the case of a repost of a blocked user in another user's feed

lexicons/app/bsky/actor/defs.json
dholms: yup exactly

I thought about doing an array but just decided to do the one. this could end up getting arbitrarily large if we make it an array. & i imagine the UI will only ever show one reason at a time

packages/pds/src/app-view/services/graph/index.ts
dholms: yeah that's a good call. those are sneaky 👌

packages/pds/src/app-view/services/indexing/plugins/list-item.ts
dholms: ahhh yup yup okay i dig this approach 👍

packages/pds/src/db/migrations/20230505T201944381Z-mute-lists.ts
dholms: yup 👍

packages/pds/src/db/migrations/20230505T201944381Z-mute-lists.ts
dholms: oh yup yknow you're totally right

i coulda sworn i checked that 😅

packages/pds/src/services/account/index.ts
dholms: yup yup good call 👍

packages/pds/src/services/account/index.ts
dholms: nope this is right & there's another place to fix it as well 👌

packages/pds/src/api/app/bsky/graph/subscribeMuteList.ts
dholms: yeah can't hurt to give some extra user feedback 👍

packages/pds/src/services/account/index.ts
dholms: any thoughts on how to test it out?

PR #994, opened 2023-05-06
fix: updated repository field for packages

wesleytodd:
Swapped in the format which allows for specifying a subdirectory in a monorepo. This should help the npm website link to the right place.


devinivy (approved):
Thanks!

PR #995, opened 2023-05-06
pds: place-holder robots.txt, and a text base-level route (/)

bnewbold:
robots.txt: want to be explicit about allowing public crawling

/: a bit of dev-experience polish. in prod we already redirect this route, but may be helpful for self-hosting folks and people just exploring

### DIFF
--- a/packages/pds/src/index.ts
+++ b/packages/pds/src/index.ts
@@ -186,6 +186,18 @@ export class PDS {
     }
 
     app.use(health.createRouter(ctx))
+    app.get('/', function (req, res) {
+      res.type('text/plain')
+      res.send(
+        'This is an AT Protocol Personal Data Server (PDS): https://github.com/bluesky-social/atproto\n\nMost API routes are under /xrpc/',
+      )
+    })
+    app.get('/robots.txt', function (req, res) {
+      res.type('text/plain')
+      res.send(
+        '# Hello!\n\n# Crawling the public API is allowed\nUser-agent: *\nAllow: /',
+      )
+    })
     app.use(server.xrpc.router)
     app.use(error.handler)
 
--- a/packages/pds/src/index.ts
+++ b/packages/pds/src/index.ts
@@ -14,7 +14,8 @@ import { BlobStore } from '@atproto/repo'
 import { AppViewIndexer } from './app-view/indexer'
 import inProcessAppView from './app-view/api'
 import proxiedAppView from './app-view/proxied'
-import API, { health } from './api'
+import API from './api'
+import * as basicRoutes from './basic-routes'
 import Database from './db'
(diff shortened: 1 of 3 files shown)


packages/pds/src/index.ts
```diff
@@ -186,6 +186,18 @@ export class PDS {
...
+    app.get('/', function (req, res) {
+      res.type('text/plain')
+      res.send(
+        'This is an AT Protocol Personal Data Server (PDS): https://github.com/bluesky-social/atproto\n\nMost API routes are under /xrpc/',
+      )
+    })
+    app.get('/robots.txt', function (req, res) {
+      res.type('text/plain')
+      res.send(
+        '# Hello!\n\n# Crawling the public API is allowed\nUser-agent: *\nAllow: /',
```
Evalprime, line 198: does the pds need to be indexed, esp. when auth is required for some endpoints?

bnewbold: I'm not sure what you mean, but what this part of the PR is trying to make more explicit is that anything which is not auth'd can be crawled.

In the atproto/bluesky system, public repo content can be crawled publicly.

Evalprime: well Twitter's API isn't crawlable, actually explicitly. so it makes more sense for the actual client to be crawlable because it has the posts themselves. when I search "hellthread" or something, I expect to see an instance on there and not the JSON representation of that. also see #1000

bnewbold: As one example, some web clients are Javascript "single page apps", which then do HTTP API calls to this backend service directly from the browser, with no authentication. This is distinct from any service which does server-to-server HTTP API request.

In that case, if somebody wanted to use web.archive.org or similar to crawl, using a browser-style javascript-aware crawler, the crawler (bot) would be making HTTP requests to this service.

packages/pds/src/index.ts
```diff
@@ -186,6 +186,18 @@ export class PDS {
     }
 
     app.use(health.createRouter(ctx))
+    app.get('/', function (req, res) {
```
dholms, line 189: could we move these to a supplementary file so as not to clog up the entry file?

besides that, looks good!

Evalprime (approved):

packages/pds/src/index.ts
bnewbold: created a "basic-routes.ts" file, and pulled `/xrpc/_health` out of the `src/api/` folder as well. I guess we might want to keep all the `/xrpc/*` stuff together under `api/`, but this felt like the right thing to do because we are already attaching that router from `index.ts` and attaching two "housekeeping" routers felt wrong.

dholms (approved):
looks great 💎

PR #1001, opened 2023-05-08
Custom feeds

dholms:


roj1512:
Here’s a nicer emoji for WIP: 🚧

lexicons/app/bsky/feed/generator.json
```diff
@@ -0,0 +1,29 @@
+{
+  "lexicon": 1,
+  "id": "app.bsky.feed.generator",
+  "defs": {
+    "main": {
+      "type": "record",
+      "description": "A declaration of the existence of a feed generator",
+      "key": "named",
```
dholms, line 8: wasn't sure what to call this key type

other options I came up with: "custom", "provided", "user-provided", "unconstrained"

Note: I also opted to _not_ have a feed name in the record properties since it can be provided by the rkey

lexicons/app/bsky/feed/generator.json
```diff
@@ -0,0 +1,29 @@
...
+      "record": {
+        "type": "object",
+        "required": ["hostname", "createdAt"],
+        "properties": {
+          "hostname": {"type": "string"},
```
dholms, line 13: preference for the format here?

Do we include scheme? allow for port numbers? allow for additional pathing?

I think no on ports/pathing. maybe scheme?

- `feed.dholms.xyz`
- `https://feed.dholms.xyz`
- `http://feed.dholms.xyz:3000`
- `https://feed.dholms.xyz/api`

lexicons/app/bsky/feed/generator.json
pfrazee: not a huge question but some other options:

- string
- user-defined

lexicons/app/bsky/feed/generator.json
Matrix89: IMO the record should have a broad definition, then the server can decide if it's supports the given format or not

pfrazee: definitely no on pathing, but allowing ports might be nice. we could do "origin" and then override the scheme if we need to

devinivy: Truly don't want to get too big-brained about it, but I do wonder if referencing a service endpoint in the user's DID doc might not be a bad idea and answer some of these questions.  I'm just slightly uneasy about these endpoints changing via the repo without transparency and not being tied into identity at all.

dholms: I actually really like this^^ gonna take another pass at it

lexicons/app/bsky/feed/describeFeedGenerator.json
```diff
@@ -0,0 +1,32 @@
...
+        "encoding": "application/json",
+        "schema": {
+          "type": "object",
+          "required": ["did", "feeds"],
+          "properties": {
+            "did": {"type": "string", "format": "did"},
+            "feeds": {
+              "type": "array",
+              "items": {"type": "string", "format": "at-uri"}
+            },
```
devinivy, line 18: As a matter of future-proofing, would there be any value is making this an array of objects, e.g. `[{ uri }]`?  The thought is that if we ever want to add an optional field alongside the URI we'd have a place to put it.  If that seems unlikely, this is good!

lexicons/app/bsky/feed/getFeedGenerator.json
```diff
@@ -0,0 +1,29 @@
...
+      },
+      "output": {
+        "encoding": "application/json",
+        "schema": {
+          "type": "object",
+          "required": ["view", "isOnline", "isValid"],
+          "properties": {
+            "view": {"type": "ref", "ref": "app.bsky.feed.defs#generatorView"},
+            "isOnline": {"type": "boolean"},
+            "isValid": {"type": "boolean"}
```
devinivy, line 23: In general we haven't been prefixing booleans with `is` (e.g. `rebase`, tooBig`, `muted`, `blockedBy`, `disabled`, `resolved`).  There is one place we do use it, though, which is on notifications `isRead`.

devinivy: Just curious, what does `isValid` check for versus `isOnline`?

lexicons/app/bsky/feed/getFeedGenerator.json
```diff
@@ -0,0 +1,29 @@
+{
+  "lexicon": 1,
+  "id": "app.bsky.feed.getFeedGenerator",
+  "defs": {
+    "main": {
+      "type": "query",
+      "description": "Retrieve a list of feeds created by a given actor",
```
devinivy, line 7:
```suggestion
      "description": "Get information about a specific feed offered by a feed generator, such as its online status",
```

lexicons/app/bsky/feed/describeFeedGenerator.json
dholms: ah yeah that makes sense

I dig 👍

lexicons/app/bsky/feed/getFeedGenerator.json
dholms: i'm down to strip out the `is` 👌

dholms: as for the second part of this, I basically do the check against `describeFeedGenerator`, but maybe I should just be doing a sample call 🤔 

I have the code up for this now too if you want a look but basically:
- `isOnline` is "does the service respond at all"
- `isValid` is "is the did in the record correct & does this service declare that they support the given feed uri?"

packages/pds/src/feed-gen/whats-hot.ts
```diff
@@ -0,0 +1,109 @@
...
+  const { ref } = ctx.db.db.dynamic
+
+  // From: https://medium.com/hacking-and-gonzo/how-hacker-news-ranking-algorithm-works-1d9b0cf2c08d
+  // Score = (P-1) / (T+2)^G
+  // where,
+  // P = points of an item (and -1 is to negate submitters vote)
+  // T = time since submission (in hours)
+  // G = Gravity, defaults to 1.8 in news.arc
+
+  const computeScore = sql<number>`${ref(
```
aliceisjustplaying, line 41: ok can i just say i love this it's so obvious in hindsight and yet. i am so excited about this one

packages/pds/src/feed-gen/bsky-team.ts
```diff
@@ -0,0 +1,53 @@
...
+  const feedService = ctx.services.appView.feed(ctx.db)
+  const graphService = ctx.services.appView.graph(ctx.db)
+
+  const { ref } = ctx.db.db.dynamic
+
+  const postsQb = feedService
+    .selectPostQb()
+    .where('post.creator', 'in', BSKY_TEAM)
+    .whereNotExists(accountService.mutedQb(requester, [ref('post.creator')]))
+    .whereNotExists(graphService.blockQb(requester, [ref('post.creator')]))
```
devinivy, line 39: May just be missing a fine detail on this— but is the mute and block filtering necessary here?  I ask because I expected the user's PDS to do the work to filter-out bad results in `getFeed`.

packages/pds/src/feed-gen/whats-hot.ts
dholms: hahaha yes 😁

packages/pds/src/feed-gen/bsky-team.ts
dholms: not necessary, but you could get smaller pages without this

basically blocked/muted posts will count against the pagination

feed gens should probably be encouraged to take blocks into account (tho obvs not required)

packages/pds/src/feed-gen/with-friends.ts
```diff
@@ -0,0 +1,84 @@
+import AppContext from '../context'
+import { QueryParams as SkeletonParams } from '../lexicon/types/app/bsky/feed/getFeedSkeleton'
+import { paginate } from '../db/pagination'
+import { countAll } from '../db/util'
+import { AlgoHandler, AlgoResponse } from './types'
+import { FeedKeyset } from '../app-view/api/app/bsky/util/feed'
+
+const handler: AlgoHandler = async (
```
devinivy, line 8: There's a TODO on this one to sort out an issue it has when running with sqlite.

devinivy (approved):

PR #1002, opened 2023-05-08
Clean up what's hot

dholms:
Generally trying to reduce nsfw & raunchy material in what's hot, we apply 3 fixes:
- no longer serve replies
- ensure we filter quote posts that don't pass our filters
- label underwear photos & strip underwear & nudity from what's hot

I also updated our labeling service to accurately return `!no-promote` labels

### DIFF
--- a/packages/pds/src/app-view/api/app/bsky/unspecced.ts
+++ b/packages/pds/src/app-view/api/app/bsky/unspecced.ts
@@ -5,6 +5,7 @@ import AppContext from '../../../../context'
 import { FeedRow } from '../../../services/feed'
 import { FeedViewPost } from '../../../../lexicon/types/app/bsky/feed/defs'
 import { NotEmptyArray } from '@atproto/common'
+import { isViewRecord } from '../../../../lexicon/types/app/bsky/embed/record'
 
 const NO_WHATS_HOT_LABELS: NotEmptyArray<string> = [
   '!no-promote',
@@ -12,7 +13,7 @@ const NO_WHATS_HOT_LABELS: NotEmptyArray<string> = [
   'self-harm',
 ]
 
-const NSFW_LABELS = ['porn', 'sexual', 'nudity']
+const NSFW_LABELS = ['porn', 'sexual', 'nudity', 'underwear']
 
 // THIS IS A TEMPORARY UNSPECCED ROUTE
 export default function (server: Server, ctx: AppContext) {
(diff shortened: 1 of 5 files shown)


packages/pds/src/app-view/api/app/bsky/unspecced.ts
```diff
@@ -70,19 +77,10 @@ export default function (server: Server, ctx: AppContext) {
...
       )
 
-      const noRecordEmbeds = feed.map((post) => {
-        delete post.post.record['embed']
-        if (post.reply) {
-          delete post.reply.parent.record['embed']
-          delete post.reply.root.record['embed']
-        }
-        return post
-      })
```
devinivy, line 80: Records that contain image or external embeds will still show-up here— is this logic just not needed at all anymore?  I recall it goes back to some issues with lexicon validation and the client.

devinivy (approved):

packages/pds/tests/seeds/basic.ts
```diff
@@ -38,6 +38,7 @@ export default async (sc: SeedClient) => {
     [img1, img2], // Contains both images and a quote
     sc.posts[bob][0].ref,
   )
+  await sc.post(carol, posts.carol[0])
```
devinivy, line 41: I see `posts.carol[0]` used above—is this intended to be the newly-added post `posts.carol[1]`?

packages/pds/src/app-view/api/app/bsky/unspecced.ts
dholms: oh shit that's right... okay this does need to be back in my bad

packages/pds/tests/seeds/basic.ts
dholms: goddammit yes 😅

devinivy (approved):

devinivy (approved):

bnewbold:
If it was up to me:

- no replies I think will be the biggest impact, :+1:
- quote-post I feel like a more subtle thing, and a fair amount of positive commentary via quote post in What's Hot, but not against this change
- I really don't like doing expanding `!no-promote` because it is currently not transparent (we remove that label from hydration, b/c app was showing as a warning previously)

I haven't checked, but to date we have only had `!no-promote` on accounts, not posts, so should verify it works as expected when applied to posts. Eg, should not show pink warning, and should be removed from "What's Hot".

I know that "ignore unknown labels" is working it's way through the app pipeline, but I don't know how broadly it is deployed. If it was, we could rip out the "don't hydrate" logic.

I think expanding `sexual` would be better. I had been meaning to do that in a more nuanced way, but could make a more aggressive move first and then back it off a bit after. Also seems like we could do this as a second thing, eg, 24hr later, after seeing what impact the replies thing has on it's own.

devinivy (approved):

PR #1004, opened 2023-05-08
Disable account invites

dholms:
This provides an admin route for disabling _future_ account invite codes

❗ Important: hitting this route will not disable current codes (use `com.atproto.admin.disableInviteCodes` for that), it only prevents creation of _future codes

Implementation detail:
We create codes for a user dynamically when they hit `server.getAccountInviteCodes`. We still create those codes even for an account with disabled codes, we just immediately disable them. This aids in bookkeeping so that an account that was disabled for a long time & then is re-enabled does not get a windfall of all previous invite codes


packages/pds/src/db/migrations/20230508T204041032Z-disable-account-invites.ts
```diff
@@ -0,0 +1,15 @@
+import { Kysely } from 'kysely'
+
+export async function up(db: Kysely<unknown>): Promise<void> {
+  await db.schema
+    .alterTable('user_account')
+    .addColumn('invitesDisabled', 'int2', (col) => col.defaultTo(0))
```
devinivy, line 6: Might recommend marking this as `notNull()` since it has a default :)

bnewbold:
Is this an action that we can reverse? Looking just at the Lexicon I can't see how.

I sort of thought that an imperative account label (eg, `!no-new-invites`) would be a reasonable/transparent way to handle this, but maybe we don't want public, and too complex/intwined.

packages/pds/tests/views/admin/invites.test.ts
```diff
@@ -199,4 +217,39 @@ describe('pds admin invite views', () => {
...
     await expect(attemptCreateInvite).rejects.toThrow('Authentication Required')
   })
+
+  it('disables an account from getting additional invite codes', async () => {
+    await agent.api.com.atproto.admin.disableAccountInvites(
+      { account: carol },
+      { encoding: 'application/json', headers: { authorization: adminAuth() } },
+    )
+
+    // does not disable existing invites
```
devinivy, line 227: This comment might be out of date, or maybe I'm just totally understanding it (or maybe there's a bug?).  I would expect there to be some codes in the response below in order to show that existing invite codes are still enabled.

devinivy (approved):
I dig the approach of generating and disabling new invites 👍 👍

packages/pds/src/db/migrations/20230508T204041032Z-disable-account-invites.ts
dholms: ah good call 👍

dholms:
I don't think we should handle with invites since this is just private server state & i don't think a PDS will want to necessarily publish that info

However, I did add in a fix that allows re-enabling an invite code

packages/pds/src/services/account/index.ts
```diff
@@ -100,6 +100,7 @@ export class AccountService {
         did,
         passwordScrypt,
         createdAt: new Date().toISOString(),
+        invitesDisabled: 0,
```
devinivy, line 103: This is 💯 as-is, but just for sharing knowledge's sake (and good chance you already know this)— since there's a default on the `invitesDisabled` column, you can mark it in the model as `Generated<0 | 1>`, meaning if it's not set the db will generate it.  That allows you to omit the column from places such as this.

PR #1005, opened 2023-05-08
Temporarily disable blob api

dholms:
We're temporarily disabling the sync blob api until we wrap up deleting old blobs that have been marked for deletion. 

Current blobs can still be accessed through the CDN with signed urls returned in views

We'll be back soon :)


dholms:
Okay too many damn tests breaking on this one -- we'll just continue deploying off of the branch until we wrap up blob deletion

PR #1009, opened 2023-05-09
Fix indexed record datetimes

devinivy:
Datetimes in records are validated as ISO.  When we index these dates with the intent of lexically sorting them, we need to ensure that the dates are normalized to the same format in order to preserve temporal sorting.

Please approach with care and test this yourself if you go for it, but an approach such as this could be used to fix datetimes in the db retroactively.  Would need to be run on post, repost, like, follow, and block tables.
```sql
# dry run
select
  "createdAt" as before,
  to_char("createdAt"::timestamptz at time zone 'UTC', 'YYYY-MM-DD"T"HH24:MI:SS.MS"Z"') as after
from post
  where length("createdAt") != 24 or right("createdAt", 1) != 'Z';

# live run
update post
  set "createdAt" = to_char("createdAt"::timestamptz at time zone 'UTC', 'YYYY-MM-DD"T"HH24:MI:SS.MS"Z"') 
  where length("createdAt") != 24 or right("createdAt", 1) != 'Z';
```

### DIFF
--- a/packages/pds/tests/views/notifications.test.ts
+++ b/packages/pds/tests/views/notifications.test.ts
@@ -75,6 +75,7 @@ describe('pds notification views', () => {
       sc.replies[alice][0].ref,
       'indeed',
     )
+    await server.ctx.backgroundQueue.processAll()
 
     const notifCountAlice =
       await agent.api.app.bsky.notification.getUnreadCount(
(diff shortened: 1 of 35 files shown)


dholms (approved):
looks great 💪

packages/pds/tests/views/notifications.test.ts
```diff
@@ -75,6 +75,7 @@ describe('pds notification views', () => {
       sc.replies[alice][0].ref,
       'indeed',
     )
+    await server.ctx.backgroundQueue.processAll()
```
devinivy, line 78: Fixes a race in the tests.

PR #1010, opened 2023-05-09
block IPs from creating accounts

Evalprime:
this pr adds a comma delimited config variable with a list of IPs blocked from creating accounts. should cut down on potential spam

no CIDR range support yet, could be an area for improvement


dholms:
Thanks for the thoughtful PR!

Right now, we take care of blocking potential abuse at the firewall layer instead of within the application & I prefer to keep it there. It prevents the request from ever even hitting the application server. Even smaller PDS distributions can do so with something like nginx.

PR #1011, opened 2023-05-09
Bring your own DID

dholms:
This allows you to bring your own DID document on account creation.

When you bring your own DID document, we attempt to resolve it & ensure that the data within matches up with the intended values. Namely:
- handle matches requested handle
- pds service endpoint matches the service endpoint you're attempting account creation on
- atproto key matches the server's repoSigningKey
- if it is a PLC DID, then we ensure the server's plcRotationKey is included in the documents rotation keys

Buyer beware: we _do not_ do that last check for did:web. If you are using a did:web DID, it is your obligation to keep your DID document up to date with the service you're using - otherwise your data may no longer be indexed.

In service of: https://github.com/bluesky-social/atproto/pull/1001

### DIFF
--- a/lexicons/com/atproto/server/createAccount.json
+++ b/lexicons/com/atproto/server/createAccount.json
@@ -39,8 +39,8 @@
         {"name": "InvalidInviteCode"},
         {"name": "HandleNotAvailable"},
         {"name": "UnsupportedDomain"},
-        {"name": "CouldNotResolveDid"},
-        {"name": "PoorlyFormattedDidDoc"}
+        {"name": "UnresolvableDid"},
+        {"name": "IncompatibleDidDoc"}
       ]
     }
   }
--- a/lexicons/com/atproto/server/createAccount.json
+++ b/lexicons/com/atproto/server/createAccount.json
@@ -13,6 +13,7 @@
           "properties": {
             "email": {"type": "string"},
             "handle": {"type": "string", "format": "handle"},
+            "did": {"type": "string", "format": "did"},
             "inviteCode": {"type": "string"},
             "password": {"type": "string"},
             "recoveryKey": {"type": "string"}
--- a/lexicons/com/atproto/server/createAccount.json
+++ b/lexicons/com/atproto/server/createAccount.json
@@ -38,7 +38,9 @@
         {"name": "InvalidPassword"},
         {"name": "InvalidInviteCode"},
         {"name": "HandleNotAvailable"},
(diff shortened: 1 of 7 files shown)


lexicons/com/atproto/server/createAccount.json
```diff
@@ -37,7 +38,9 @@
         {"name": "InvalidPassword"},
         {"name": "InvalidInviteCode"},
         {"name": "HandleNotAvailable"},
-        {"name": "UnsupportedDomain"}
+        {"name": "UnsupportedDomain"},
+        {"name": "CouldNotResolveDid"},
+        {"name": "PoorlyFormattedDidDoc"}
```
devinivy, line 43: Sorry for the total nit, but I'm not feeling 100% sure about this error name.  My understanding is that this is used when there's a valid did doc but it's not compatible with the PDS that the account is being created on.  What if it were something like `IncompatibleDidDoc`?  Promise I'm not going to get hung-up on this if you prefer it as-is 😆

devinivy (approved):
Pretty sweet!

devinivy (approved):

packages/pds/src/api/com/atproto/server/createAccount.ts
```diff
@@ -158,3 +139,102 @@ export const ensureCodeIsAvailable = async (
...
+  } else if (atpData.signingKey !== ctx.repoSigningKey.did()) {
+    throw new InvalidRequestError(
+      'DID document signing key does not match service signing key',
+      'IncompatibleDidDoc',
+    )
+  }
+
+  if (input.did.startsWith('did:plc')) {
+    const data = await ctx.plcClient.getDocumentData(input.did)
+    if (!data.rotationKeys.includes(ctx.plcRotationKey.did())) {
```
AaronGoldman, line 231: This may not be a good idea.
If the reason a user is bringing their own did is that they don't want the PDS to be able to update their did:plc this will prevent them from using this PDS.

There is a real use case of I want to manage my own DID and just use the PDS to manage the repo.

dholms: Yeah we will loosen this in the future, but I think it's important for now

If the PDS does not have the ability to rotate documents, then
- it cannot help out with handle changes
- if it rotates it's repo signing key, it can no longer publish records on a repo's behalf

This should be allowed at some point, but for now, for usability's sake, we constrain the provided DID

PR #1012, opened 2023-05-09
use blob as fetch fallback

Evalprime:
for non-textual response bodies, blob is the safest fallback that should work for everything


Evalprime:
note: an ArrayBuffer could be used here but it lacks immutability, so blob is better

devinivy (approved):
Seems reasonable to me!  @pfrazee you're deep in RN-land, any qualms with this in the docs?

pfrazee:
I haven't tested whether this would work but it's just a suggestion so I'm +1

PR #1013, opened 2023-05-09
Add account preference APIs

pfrazee:
This PR is currently a proposal for adding two methods

- `com.atproto.server.getAccountPreferences`
- `com.atproto.server.putAccountPreferences`

They use an open-union array of preferences that are defined the server.defs. The settings should be stored privately under the requestor's account. The `put` should overwrite all current settings with the given input.

### DIFF
--- a/lexicons/com/atproto/server/defs.json
+++ b/lexicons/com/atproto/server/defs.json
@@ -25,6 +25,31 @@
         "usedBy": {"type": "string", "format": "did"},
         "usedAt": {"type": "string", "format": "datetime"}
       }
+    },
+    "preferences": {
+      "type": "array",
+      "items": {
+        "type": "union",
+        "refs": [
+          "#adultContentPref",
+          "#contentLabelPref"
+        ]
+      }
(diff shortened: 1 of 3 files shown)


bnewbold:
Some other pieces of private state relevant to inter-device sync or account migration. These already all "just work" I think, no changes needed:

- "last seen" notifications timestamp: feels pretty ephemeral, could just reset to "now" on account migration. already works fine for inter-device sync
- mutes: has separate `getMutes`/`muteActor` endpoints which could be used for migration
- invites: can "get" inter-device. does not make sense to migrate these
- app passwords: doesn't make sense for migration? and can already "get" from any device

Settings we *might* want to sync/include, but also maybe are device-local:

- dark mode (feels very device-local)
- language preferences

bnewbold:
Are you thinking that we would do all the labels individually, or at the toggle group/category level? I can see pros/cons to both approaches.

pfrazee:
Great! Lgtm!

PR #1014, opened 2023-05-09
Update how sequencer invalidation works

dholms:
Foreign key constraint was giving us a hard time & also this was requiring extra processing after INSERT

We weren't using the actual seq # contained in `invalidatedBy` anywhere, so we update it to just a boolean


packages/pds/src/db/migrations/20230509T192324175Z-seq-invalidated.ts
```diff
@@ -0,0 +1,43 @@
...
+  }
+  await db.schema
+    .alterTable('repo_seq')
+    .addForeignKeyConstraint(
+      'invalidated_by_fkey',
+      // @ts-ignore
+      ['invalidatedBy'],
+      'repo_seq',
+      ['seq'],
+    )
```
devinivy, line 41: I don't think this is supported by sqlite— probably fine to just skip in that case.

devinivy (approved):

PR #1016, opened 2023-05-09
disallow new registration of .arpa, .local, etc handles

bnewbold:
This is a "stop the bleeding" PR to halt creation and migration of new handles with `.arpa`, `.local`, and a couple other special "infrastructure" and "reserved" TLDs.

Rough plan would be to get this in, then ask existing `.arpa` tricksters to change handle (or better yet just delete those accounts), and then soon disallow at the protocol level (aka, even referencing such a domain would be invalid).

Also adds test coverage to ensure we don't allow "dotless" top-level handles, like "io". These are controversial and seem to be considered an anti-pattern by many IETF folk. I think they would be super duper cool but also super confusing for users, so we should continue to disallow at the protocol level.

cc: @devinivy @dholms for visibility

### DIFF
--- a/packages/identifier/src/handle.ts
+++ b/packages/identifier/src/handle.ts
@@ -2,6 +2,22 @@ import { reservedSubdomains } from './reserved'
 
 export * from './resolve'
 
+// Currently these are registration-time restrictions, not protocol-level
+// restrictions. We have a couple accounts in the wild that we need to clean up
+// before hard-disallow.
+// See also: https://en.wikipedia.org/wiki/Top-level_domain#Reserved_domains
+const DISALLOWED_TLDS = [
+  '.local',
+  '.arpa',
+  '.invalid',
+  '.localhost',
+  '.internal',
+  // policy could concievably change on ".onion" some day
+  '.onion',
+  // NOTE: .test is allowed in testing and devopment. In practical terms
+  // "should" "never" actually resolve and get registered in production
+]
+
 // Handle constraints, in English:
 //  - must be a possible domain name
 //    - RFC-1035 is commonly referenced, but has been updated. eg, RFC-3696,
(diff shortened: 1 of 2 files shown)


bnewbold:
I think the test failure is not due to this code, but some package/dependency thing

dholms (approved):
yup seems reasonable 👍 

thx for putting this together

ericsampson:
fwiw this comment didn't get updated:
> does not validate whether domain or TLD exists, or is a reserved or special TLD (eg, .onion or .local)

bnewbold:
@ericsampson that is intentional in this PR. What we have changed for now is preventing registration of these domains with our PDS implementation, but not yet banned them from the protocol itself. That will come as a second step.

caramelpony:
Not to beat a dead PR, but could we get clarification on the specific reasoning behind banning .arpa and dotless domains? While the documentation provides justification for other banned special-use TLDs, it doesn’t seem to address these two explicitly.

From a trust and ownership perspective, this restriction feels counterintuitive. Both .arpa and top level domains require a higher level of administrative control and are typically tied to infrastructure, making them verifiably unique and resistant to abuse. Given that ATProto uses domain-based handles as a trust mechanism, preventing the use of these domains seems like an unnecessary limitation. For instance, network operators might want to establish BSky accounts linked to .arpa subdomains for their IP ranges, providing an authoritative channel for network status updates or operational announcements. 

Similarly, top level "dotless" domains, while unconventional, could for example be useful to a corporate entity who owns a brand TLD and wishes to use that as their handle (i.e. protocol-wide `@apple` for Apple), or if a TLD registry operator wishes to establish proof of ownership over their BSky account. 

A protocol-wide ban on both of these prevents edge cases where they can be useful.

PR #1017, opened 2023-05-09
firehose tombstones

jesopo:
to be honest this change seems so simple that i assume the reason it wasn't implemented yet is less to do with spare time and more to do with technical details that need to be worked out, but i thought i'd try my luck anyway


jesopo:
(test on its way...)

packages/pds/tests/sync/subscribe-repos.test.ts
```diff
@@ -273,6 +289,41 @@ describe('repo subscribe repos', () => {
...
+    })).did
+    const baddie2 = (await sc.createAccount("baddie2.test", {
+      email: 'baddie2@test.com',
+      handle: 'baddie2.test',
+      password: 'baddie2-pass',
+      displayName: undefined,
+      description: undefined,
+    })).did
+
+    for (var did of [baddie1, baddie2]) {
```
ericvolp12, line 308: Should this be a `const` instead of a `var`? Looks like you used a `const` in the other test so this might have been a typo.

packages/pds/src/sequencer/events.ts
```diff
@@ -128,6 +128,20 @@ export const formatSeqHandleUpdate = async (
   }
 }
 
+export const formatSeqTombstone = async (
```
dholms, line 131: one thing is that these should invalidate all prior events from the user

take a look at `sequenceEvt` where "rebases" invalidate prior "rebases" & "appends" and "handles" invalidate prior "handles" 

(also merge in main first cause we just made some some changes to invalidation)

packages/pds/src/sequencer/events.ts
```diff
@@ -151,6 +167,10 @@ export const invalidatePrevHandleOps = async (db: Database, did: string) => {
   return invalidatePrevSeqEvts(db, did, ['handle'])
 }
 
+export const invalidatePrevOps = async (db: Database, did: string) => {
+  return invalidatePrevSeqEvts(db, did, ['append', 'handle', 'rebase'])
```
jesopo, line 171: enumerating all the different event types here feels sort of bad and prone to breaking next time an event type is added. i guess i could change the db query in `invalidatePrevSeqEvts` to just not have a `where eventType =` clause for this case? thoughts?

packages/pds/tests/sync/subscribe-repos.test.ts
```diff
@@ -104,6 +120,22 @@ describe('repo subscribe repos', () => {
...
   }
 
+  const getAllEvents = (userDid: string, frames: Frame[]) => {
+    const types = []
+    for (const frame of frames) {
+      if (frame instanceof MessageFrame) {
+        if (
+          (frame.header.t === "#commit" && (frame.body as CommitEvt).repo === userDid)
+          || (frame.header.t === "#handle" && (frame.body as HandleEvt).did === userDid)
+          || (frame.header.t === "#tombstone" && (frame.body as TombstoneEvt).did === userDid)
```
jesopo, line 130: same thing here wrt enumerating all the types

jesopo: though in this situation i don't think there's a better way to extract a did from body without doing this

Broken60:
> it actually totally had to do with time 😅
> 
> we just have a million small things to do that would only take 20 min a piece, but haven't gotten around to
> 
> Besides getting the invalidation logic in here, it looks good!
> 
> thanks for doing this!

H

dholms:
hey @jesopo wondering if you're planning on getting back to this or if i should get it over the line?

jesopo:
what's missing

jesopo:
ah yeah sorry missed that. I'll look at it shortly

jesopo:
@dholms how's that?

dholms (approved):
yup this looks good! thanks for putting this together

running tests then i'll merge & deploy soon

jesopo:
oh i see, your last round of comments were not actually submitted yet and now they are. i shall take a look

jesopo:
(also, i will fix tests)

jesopo:
> Oh yknow what I just realized?
> 
> We actually delete all the repo_seq events for a user when they delete their account, so there won't be any events to invalidate :sweat_smile: 
> 
> I think you're actually good the way it was. Sorry about that!

are you sure about this? `packages/pds/tests/sync/subscribe-repos.test.ts` `does not return invalidated events` is returning a handle event for a user deleted in the previous test (`account deletions invalidate all seq ops`)

jesopo:
https://github.com/bluesky-social/atproto/blob/95252ddb4fd10ac7e7035303b125a80ee8dd88f0/packages/pds/src/services/account/index.ts#L400-L419

I don't see `repo_seq` in here

jesopo:
race you :runner:

jesopo:
figured it out - my test was only calling account delete, not repo delete

dholms:
& then we'll finally get this in!

jesopo:
done(?)

jesopo:
i believe this won't send tombstones for account takedowns. i imagine that's desired behaviour anyway, but just wanted to check

jesopo:
I'll check this eve :)

dholms (approved):
:shipit:

PR #1018, opened 2023-05-09
Make BskyAgent instantiation 4 times faster

Matrix89:
By using `z.discriminatedUnion` instead of `z.union`
we can hint to zod at which field it should look when validating/parsing schemas.
You can read the docs on discriminated unions [here](https://github.com/colinhacks/zod#discriminated-unions)

This improves BskyAgent instantiation time around 4 times, when creating 10 agents using stuff from #1015,
the time went down from ~1200ms to 330ms

Issue #915, opened 2023-04-27
Agent Creation has a large CPU time


wesbos:
Not sure if this is a real issue, but I thought I'd open a discussion around how long creating an agent takes.

Creation of a BSkyAgent takes over 112ms on my Macbook M1. 

```js
console.time('start');
import { BskyAgent } from "@atproto/api";
const agent = new BskyAgent({ service: "https://bsky.social" });
console.timeEnd("start");
```

This code isn't hitting the network, it's simply the initialization of the class.

I ran into this issue trying to use this library on Cloudflare workers, which limits your requests to 10ms CPU time. 

They also offer 200ms script startup, and 50ms CPU time on pro plans. Both would go over.

Deno deploy has the same 10ms free, 50ms paid limitations.

So, is there anything that can be done to decrease this time? Or is the answer to just use a service that allows for longer run times?

roj1512:
The initialization doesn’t take anything for me on M2.

wesbos:
What is anything? Have you measured or deployed to Deno Deploy?

roj1512:
Oh, right. I was wrong.

```text
 $ time deno eval "import mod from \"npm:@atproto/api\";new mod.BskyAgent({ service: \"https://bsky.social\" });"
deno eval   0.65s user 0.06s system 134% cpu 0.528 total
```

wesbos:
I thought so, but the shipped version is compiled and has only 1 import. I think it's due to all the encrypt / decrypt, but also not my area of expertise

myConsciousness:
This is a common problem for API wrappers regardless of language, but the initialization cost tends to be heavy because of the high-end implementation. It would be interesting to see if there is a design pattern that could significantly improve this cost, because I also develop API wrapper in Dart.

Evalprime:
I can look into this this weekend

vincentriemer:
FWIW: I recorded a profile and it seems like almost the entirety of the CPU time is dedicated to the `addLexicons` method where it's parsing the lexicon docs at runtime. IMO this should be done ahead of time because the nearly 300ms of CPU time just to initialize an API client is kind of ridiculous.

<img width="605" alt="Screenshot 2023-05-08 at 14 59 00" src="https://user-images.githubusercontent.com/1398555/236947822-c13e216a-7485-45e5-8ad6-5ff889a8efe4.png">

vincentriemer:
[...]

PR #1021, opened 2023-05-10
Update zod version, some perf improvements

devinivy:
Zod has had a handful of improvements since we last bumped it, in particular a. it put `ZodError` creation behind a getter and b. compiler perf improvements in https://github.com/colinhacks/zod/pull/2107.

Ref: #915 #1018


dholms (approved):
😎 👌

Issue #1007, opened 2023-05-09
transport compression (gzip) for `com.atproto.sync.getRepo`


bnewbold:
Exports from this HTTP endpoint are fairly large, we should ensure they are compressed over the wire using HTTP transport compression.

It's possible this is already partially enabled and I've just missed it. I tried a browser request over HTTP/2, and the `accept-encoding` request header was sent as expected, and no compression on the response.

Not urgent, just filing to keep track.

Evalprime:
I think we should always use gzip whenever possible, but idk how that works when it's IPFS

dholms:
we're not using IPFS - only IPLD as a data format

We stream out blocks in a CAR file & gzip actually _really_ helps with these

would like to add ASAP 👌

bnewbold:
Didn't realize this had been requested so much!

Dupe: https://github.com/bluesky-social/atproto/issues/853
Dupe: https://github.com/bluesky-social/atproto/issues/620

PR #1022, opened 2023-05-10
Add compression in http res

dholms:
Add in compression middleware & explicitly tip it off that it should be compressing car files

Not implemented on repo subscription yet, we can discuss

Closes https://github.com/bluesky-social/atproto/issues/1007


packages/pds/package.json
```diff
@@ -7,7 +7,7 @@
     "url": "https://github.com/bluesky-social/atproto.git",
     "directory": "packages/pds"
   },
-  "main": "src/index.ts",
+  "main": "dist/index.js",
```
bnewbold, line 10: intentional fix, or accidental commit?

Matrix89: What's even up with that, when I do `yarn install` is does it automagically

bnewbold:
I dug through the docs of this package a bit, and we should verify if it is actually working.

It has a threshold of 1kbyte, below that it does not compress. Good! Only works if `Content-Length` is provided; should verify that is working for XRPC endpoints.

 It uses a MIME database to check if the payload is "compressible". Eg, `image/jpeg` won't be re-compressed (correct behavior). For CAR files, it isn't obvious if it will work or not. If not, one solution might be to just send a PR to the upstream project:
 
 https://www.npmjs.com/package/mime-db
 https://github.com/jshttp/mime-db/blob/master/src/custom-types.json

packages/pds/package.json
dholms: accidental

yeah I'm honestly not sure where it comes from. has bitten us a couple times

dholms:
ah yup it probably isn't working then 😅 

I threw this together in like 5 min just to get the ball rolling & didn't test it out at all yet

I'll give it some more love here soon

bnewbold:
Was just poking at bsky.app page sizes, and getTimeline is like 110 KBytes+. Would be good to get this in even if it only works for big JSON responses!

dholms:
@devinivy & I were just chatting about this!

gonna get back into it here soon

devinivy (approved):

devinivy (approved):
Ace!

PR #1024, opened 2023-05-10
Update sanitization.ts

speedbunny:
Improved sanitize function with commenting.


dholms:
We're not using this syntax to comment on our functions anywhere else in the codebase so don't think I'll add it here

If the param names + types do not adequetly describe a function, I'm down to add a comment, but I think they suffice in these cases

PR #1026, opened 2023-05-10
service <> service auth

dholms:
This tidies & formalizes service <> service auth.

Three primary changes:
- moves creation & validation of service jwts to `@atproto/xrpc-server` for easier reuse
- adds `aud` to service jwt. this is security consideration to prevent MITM attacks
- formalizes the concept that services should have DIDs & gives both the appview & PDS real DIDs (in their cfg & in test envs)

### DIFF
--- a/packages/pds/src/index.ts
+++ b/packages/pds/src/index.ts
@@ -179,7 +179,7 @@ export class PDS {
     })
 
     server = API(server, ctx)
-    if (ctx.cfg.bskyAppViewEndpoint) {
+    if (ctx.cfg.bskyAppViewEndpoint && ctx.cfg.bskyAppViewDid) {
       server = proxiedAppView(server, ctx)
     } else {
       server = inProcessAppView(server, ctx)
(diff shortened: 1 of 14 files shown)


packages/pds/src/index.ts
```diff
@@ -179,7 +179,7 @@ export class PDS {
     })
 
     server = API(server, ctx)
-    if (ctx.cfg.bskyAppViewEndpoint) {
+    if (ctx.cfg.bskyAppViewEndpoint && ctx.cfg.bskyAppViewDid) {
```
devinivy, line 182: I think a case with `bskyAppViewEndpoint` configured but not `bskyAppViewDid` should probably fail at server start if we can swing it easily.  Right now if I'm reading right, it will silently fallback to in-process.

devinivy (approved):

packages/pds/src/index.ts
dholms: yeah I'm down with that 👍

dholms: ah actually that check (& error throw) is done in the lex server for the proxy 👍  so i'll just delete this out of the if statement

devinivy (approved):

PR #1027, opened 2023-05-11
Add block record duplicate detection

### LINKED ISSUE #1025: Duplicate record detection on blocks
We detect duplicate likes/reposts/follows, but do not detect duplicate blocks yet

Matrix89:
Should close #1025 

Also a quick question, I had to run sqlite tests instead of the normal ones as some of those were failing with internal server errors. Im on node `v18.16` and postgres `15.2`


Matrix89:
Sorry for the workflow reruns, seems my editor didn't pick up neither `eslint` nor `prettier`, probably because there's no config files for them

packages/pds/src/services/record/index.ts
```diff
@@ -281,7 +281,10 @@ export class RecordService {
 // For now we just want to ensure we're tracking links from follows, likes, and reposts.
```
dholms, line 281:
```suggestion
// For now we just want to ensure we're tracking links from follows, blocks, likes, and reposts.
```

dholms (approved):
This looks great, thanks!

dholms:
huh that's strange 🤔 

any chance you caught the errors that were causing internal server error?

you can print a formatted tail of the logs with yarn test:log

dholms:
also there is an .eslintrc & .prettierrc, were those not getting picked up?

Matrix89:
> huh that's strange thinking
> 
> any chance you caught the errors that were causing internal server error?
> 
> you can print a formatted tail of the logs with yarn test:log

`grep "level\":50" test.log | yarn pino-pretty > errors.log`
Using `yarn test:log` was only giving me the ones that passed
https://gist.github.com/Matrix89/8ba2a32dee4063a5852cd25cd5c675e0

Matrix89:
> also there is an .eslintrc & .prettierrc, were those not getting picked up?

Yep, found them, problem with my vim config

PR #1028, opened 2023-05-11
Reports by active action type

devinivy:
Adds support for fetching moderation reports by active action type.  For example, if you filter reports by the action type "flag" then you will find all reports that were resolved with a flag action that hasn't been reversed.

### DIFF
--- a/lexicons/com/atproto/admin/getModerationReports.json
+++ b/lexicons/com/atproto/admin/getModerationReports.json
@@ -10,6 +10,15 @@
         "properties": {
           "subject": {"type": "string"},
           "resolved": {"type": "boolean"},
+          "actionType": {
+            "type": "string",
+            "knownValues": [
+              "com.atproto.admin.defs#takedown",
+              "com.atproto.admin.defs#flag",
+              "com.atproto.admin.defs#acknowledge",
+              "com.atproto.admin.defs#escalate"
+            ]
+          },
           "limit": {"type": "integer", "minimum": 1, "maximum": 100, "default": 50},
           "cursor": {"type": "string"}
         }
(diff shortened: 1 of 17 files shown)


lexicons/com/atproto/admin/getModerationReports.json
```diff
@@ -10,6 +10,15 @@
...
           "resolved": {"type": "boolean"},
+          "actionType": {
+            "type": "string",
+            "knownValues": [
+              "com.atproto.admin.defs#takedown",
+              "com.atproto.admin.defs#flag",
+              "com.atproto.admin.defs#acknowledge",
+              "com.atproto.admin.defs#escalate"
+            ]
+          },
```
bnewbold, line 21: It feels a bit of a bummer we can't transclude this list by ref. will have to update in both places if we add additional action types.

not that bad in this case though, probably not much more churn on this enum.

bnewbold:
This looks great to me! Thanks for the quick impl.

Looks ready to merge from my POV, will wait to see if @dholms can review for general code quality stuff. Would be great if we can get this merged and deployed by tomorrow and have the iterated redsky stuff building on top of this ready for the weekend.

lexicons/com/atproto/admin/getModerationReports.json
devinivy: I agree—I originally wrote it with a ref then realized it doesn't quite work.  I can imagine adding support for that to lexicon, though— the funny part is that you have to resolve the ref in order to know whether the parameter is valid, whereas right now the params can be validated statically.

devinivy: Logged an issue: https://github.com/bluesky-social/atproto/issues/1030

bnewbold:
also: a minor version bump and npm push of `@atproto/api` after this merges would be helpful for redsky integration. I can do prep dev work with a local reference, but can't, AFAIK, actually deploy without the package update.

dholms (approved):
sorry for the late review

this LGTM 👍

PR #1031, opened 2023-05-11
Revamp dev env

dholms:
This revamps the dev env to use the declarative & programmatic `test-env` that we were using in tests as opposed to the repl/cli-based approach before.

If I did everything right, this should feel no different to use

Gives us a better footing for introducing additional services such as Feed Generators


devinivy (approved):
No notes, this is great 🤤

PR #1032, opened 2023-05-11
WIP: Primitive refs in xrpc params

Matrix89:
Opening as a draft as I've realized that unions with all primitive types are probably also in scope

Testing lexicon gist: https://gist.github.com/Matrix89/9ef39c8ba5e3cc890bdf1755f608e66d


packages/lex-cli/src/codegen/lex-gen.ts
```diff
@@ -14,15 +14,18 @@ import {
   LexCidLink,
   LexBytes,
   LexIpldType,
+  LexXrpcParametersProperty,
+  LexRef,
 } from '@atproto/lexicon'
 import { toCamelCase, toTitleCase, toScreamingSnakeCase } from './util'
+import { PROPERTY_VALID_REF_TYPES } from '@atproto/lexicon/src/validators/xrpc'
```
dholms, line 21: I wanna avoid imports like this, so wherever this lands, we should properly export it so we can just import from `@atproto/lexicon`

Matrix89: Yeah, I've seen that before pushing, but I wasn't sure where it should even go, but having thought about it, it belongs in the xrpc validators file. I'll push an fix later

Matrix89: Should be good now

dholms:
Just a status update: we're having a bit of internal back & forth about when/if we should merge this, because it will require coordinating updates across other languages

Matrix89:
> because it will require coordinating updates across other languages

Let me know if I can help with other implementations

Matrix89:
FYI, rebase, the test failure seems to be unrelated

Issue #1037, opened 2023-05-13
Is `action` required in `com.atproto.repo.applyWrites`?


myConsciousness:
Hi amazing developers,

I was looking at the `com.atproto.repo.applyWrites` lexicon and noticed that `action` is there as required. But none of the `create`, `update`, or `delete` definitions have `action` in their properties.

https://github.com/bluesky-social/atproto/blob/bba1ba739eac67dcacbea873f76823e8a94c1891/lexicons/com/atproto/repo/applyWrites.json#L39-L66

I have confirmed that this `action` was used before the refactoring of the protocol. Maybe forgot to erase it, or maybe it was left there for some reason?

Matrix89:
> Maybe forgot to erase it

This seems to be the case, lexicon validator doesn't raise an error when there's a required property is missing from the properties definitions.

dholms:
yup that's right. Thanks for the catch! it didn't get properly cleaned up in the lex refactor

Matrix89:
@dholms #1040 Closes this one

PR #1040, opened 2023-05-14
Add lexicon LexXrpcParameters, Object required properties check

Matrix89:
Opening as a draft as there are two lexicons with issues
Addresses: #1037 
Now schemas like the one below won't validate, I also had to rewrite `lexUserType` validator as `z.discriminatedUnion` can't take `ZodEffects` as an argument and using just a simple `z.union` would be quite slow on such big object.

```json
...
{

  "type": "object",
  "required": ["foo"],
  "properties": {}
}
...
````

### DIFF
--- a/packages/lexicon/src/types.ts
+++ b/packages/lexicon/src/types.ts
@@ -234,6 +234,10 @@ export type LexRecord = z.infer<typeof lexRecord>
 // core
 // =
 
+// We need to use `z.custom` here because
+// lexXrpcProperty and lexObject are refined
+// `z.union` would work, but it's too slow
+// see #915 for details
 export const lexUserType = z.custom<
   | LexRecord
   | LexXrpcQuery
@@ -294,7 +298,26 @@ export const lexUserType = z.custom<
         return lexUnknown.parse(val)
     }
   },
-  { fatal: false },
+  (val) => {
+    if (!val || typeof val !== 'object') {
+      return {
+        message: 'Must be an object',
+        fatal: true,
+      }
+    }
+
+    if (val['type'] === undefined) {
+      return {
+        message: 'Must have a type',
+        fatal: true,
+      }
+    }
+
+    return {
+      message: `Invalid type: ${val['type']} must be one of: record, query, procedure, subscription, blob, array, token, object, boolean, integer, string, bytes, cid-link, unknown`,
+      fatal: true,
+    }
+  },
 )
 export type LexUserType = z.infer<typeof lexUserType>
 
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Comment by Matrix89** [comment]
Read for review, also, while running codegen I've noticed two things:
- Looks like codegen wasn't run in `packages/bsky` when mutes got introduced as it wanted to create a lot of files
- Is the order of imports/fields inconsistent between codegen runs?

File: packages/lexicon/src/types.ts [hunk 1]
```diff
@@ -229,26 +234,68 @@ export type LexRecord = z.infer<typeof lexRecord>
...
+    if (!val || typeof val !== 'object') {
+      return
+    }
 
-  lexArray,
-  lexToken,
-  lexObject,
+    if (val['type'] === undefined) {
+      return
+    }
```
devinivy, line 260: Just making sure I understand—these plain `return` statements get turned into errors by zod, is that right?

devinivy:
> Looks like codegen wasn't run in packages/bsky when mutes got introduced as it wanted to create a lot of files

Yeah, that's right— packages/bsky is a little out of sync right now, and we plan to bring it up to speed this week.

> Is the order of imports/fields inconsistent between codegen runs?
It's been consistent in my experience, but it may be relying on some implicit behavior.  I think it actually may be following the rules of bash glob expansion, which should be alphabetical.  Are you having issues with it?

Matrix89:
> It's been consistent in my experience, but it may be relying on some implicit behavior. I think it actually may be following the rules of bash glob expansion, which should be alphabetical. Are you having issues with it?

Yep, it changes the order of similarly named things, for example:
```diff
-import * as AppBskyFeedGetPostThread from './types/app/bsky/feed/getPostThread'
 import * as AppBskyFeedGetPosts from './types/app/bsky/feed/getPosts'
+import * as AppBskyFeedGetPostThread from './types/app/bsky/feed/getPostThread'
```

It's not only in imports, same goes for generated xrpc calls

packages/lexicon/src/types.ts
Matrix89: You're right, the error messages were less descriptive than before, I've changed it

packages/lexicon/src/types.ts
```diff
@@ -229,26 +234,68 @@ export type LexRecord = z.infer<typeof lexRecord>
...
 
-export const lexUserType = z.discriminatedUnion('type', [
-  lexRecord,
-
-  lexXrpcQuery,
-  lexXrpcProcedure,
-  lexXrpcSubscription,
-
-  lexBlob,
+export const lexUserType = z.custom<
```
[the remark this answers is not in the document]
Matrix89: Done

devinivy:
Thanks!

PR #1042, opened 2023-05-15
wip: create basic table for inputs/output json

Evalprime:
docs are unreadable, this pr adds a table for the inputs and provides the output in json


dholms:
mind giving an example of the output?

dholms:
Hmm just from looking at the code, kinda weird to put the parameters in a table but not the output...

This also doesn't handle inputs on procedures, records, or resolving refs/open unions.

Resolving refs/open unions is the tricky on & I think we hold off on pretty printing until we can do that

PR #1044, opened 2023-05-15
Count graphemes without Intl

### LINKED ISSUE #1043: strings.ts graphemeLen doesn't work in Firefox
just got a report from a user that the https://skeet.gotmystic.com login wasn't working and we grab a user profile after login which seems to use strings.ts/graphemeLen which I see was added with the fix for https://github.com/bluesky-social/atproto/issues/595 

only problem is, this doesn't work for Firefox as Intl.Segmenter doesn't exist there. Is it possible to use a polyfill here? or is there any fun way around this so this will work with Firefox?

Thank you!

Evalprime:
fixes #1043 using a special library

there is a polyfill for Intl.Segmenter() but it requires the entire stack to use async/await which I didn't want to deal with


dholms (approved):
yeah this probably makes sense since `Intl.Segmenter` is not available in React Native or Firefox

Any objections @devinivy @pfrazee?

Matrix89:
The library also looks reasonable, MIT, 0 dependencies, lots of weekly downloads and snyk doesn't show any known vulnerabilities.
The only problem I see is that there are issues with new emojis according to github issues, somebody recommended [this](https://github.com/flmnt/graphemer) as an alternative but I haven't looked into it

dholms:
actually, `graphemer` is what the client currently uses for a polyfill, so we should use that 👍 

(similarly MIT licensed, 0 deps)

dholms (changes requested):
Let's change this to `graphemer` & then I'm down to merge 👍

dholms:
hey @Evalprime, saw you re-requested a review - did you see my earlier comment about using `graphemer`?

pfrazee:
I'm good with this, the song-and-dance with the polyfill isn't really helping anybody

dholms (commented):
make sure you use yarn not npm

dholms:
lockfile looks like it got borked. can you reset & just run `yarn add graphemer`?

dholms (approved):
there we go 👍

PR #1045, opened 2023-05-15
Handling deleted content on admin/moderation methods

devinivy:
The admin views for moderation actions and reports were not adequately handling deleted repos and records.  This work ensures that deleted repos/records remain actionable, and that report and action detail views have a way to represent a missing repo/record.  The appview already had this covered in both cases, but tests have been added to back-up that behavior.


dholms (approved):
excellent 🙌

PR #1046, opened 2023-05-16
Starter bsky algos

dholms:
Some starter bsky algos:
- current `unspecced.getPopular` moved to `whats-hot-classic`
- `bsky-team` - feed of posts from bsky team members
- `skyline` - our custom algo that we'll evolve over time
  - currently: posts from in-network with >= threshold likes or posts from out-of-network with >= threshold likes from in-network accounts. Threshold likes is based on total follow count with a min of 2 & a max of 8.

I moved these algos to a new `feed-gen` section of the PDS. Similar to app-migrations, they are exported from the PDS package but they are actually mounted onto the PDS in the service file. 

I also added a new environment variable for providing the DID to be used in these feed gen uris (`@bsky.app`'s on production)

---

Also we need to give the PDS a proper did with a feed generator service entry before launching


Evalprime (commented):
should we hardcode five here? if the skyline algo is to be implemented across multiple pds, perhaps we could set five as a default but allow admins to change the algorithm.

the skyline algo might also be good for showing "here's what's on this pds" when posts become visible

dholms:
gonna make it dynamic based off of # of follow

skyline algo would likely only be by our appview

of course, others can implement & provide something similar however they want

would like to make all these composable & mountable, but haven't gotten there yet

Evalprime (approved):

Issue #969, opened 2023-05-04
Do not allow publicsuffix domains as handles


dholms:
https://publicsuffix.org/list/public_suffix_list.dat

dholms:
And maybe consider moving HTTP handle resolution to `.well-known`?

dholms:
Or maybe consider this:

> How about putting in DNS record which flags for URL verification support? So if the domains added that record then the client would proceed to check this path but if the record isn't there it ignores it. Simplest way for a domain hosting people's handles on subdomains, I think

SaphireLattice:
In addition to that, it really should not rely on subdomains to provide accurate info. Or well, subdomains of a domain under a public suffix. 

E.g. A handle of `subdomain.host.public-suffix` should cause the atproto to check with `host.public-suffix`, while `host.public-suffix` gets asked directly. 

> Could just limit public suffix list domains to only verify using DNS.
I guess you mean anything matching the suffix directly? Well, most of them are TLDs (or TLD-like, for when a country does not allow top-level registrations but has sub-TLDs), or services that allow almost anyone to get a subdomain... I guess it would work as a rule of thumb, but can honestly just deny anything matching directly.

april:
Checking the PSL never hurts and has little downside, although it's unlikely (if possible) that someone will have control of a TLD or `_atproto.tld`.

> And maybe consider moving HTTP handle resolution to `.well-known`?

In my experience, I've typically only seen three methods for verifying domain ownership via HTTP served files:

1. putting it inside `/.well-known`
2. uploading a file to the root directory of the domain (not a directory, as directories are often mapped all over the place)
3. editing the default HTML file at the root of the website, e.g. adding a `<meta>` tag

All of which are done over HTTPS (which I assume you're doing but just putting it out there.)

PR #1048, opened 2023-05-16
Move handle resolution to .well-known

dholms:
This moves handle resolution to `.well-known`

Specifically, a handle is resolved by calling `GET http(s)://${handle}/.well-known/atproto-did/${handle}`

The expected response is `text/plain` and simply contains the associated DID

Closes https://github.com/bluesky-social/atproto/issues/969


bnewbold:
It feels to me like this is generic to DIDs in general, not just atproto, and we could represent that in the path name. Eg, `/.well-known/did/${handle}` or `/.well-known/handle-did/${handle}`.

Is it a requirement that the handle in the hostname match the handle in the path?

bnewbold:
I actually really like the sub-directory thing, makes it quite easy to do virtual hosting with a bunch of domains and DIDs using something like nginx

dholms:
While it is generic to dids, I think it's useful for different ecosystems to be able to differentiate the dids/handles associated with each.

For instance, with the DNS records we do
`TXT.  _atproto.${handle}   "did=did:...."`

> Is it a requirement that the handle in the hostname match the handle in the path?

I don't have a super strong opinion on this. But I do think it's reasonable to be able to ask an authority what a subdomain resolves to. ie: asking `bsky.social` what `dholms.bsky.social` resolves to. 

Though canonically, you're encouraged to ask the hostname that matches the handle

dholms:
I was trying to match our DNS semantics with the pathing here. maybe `.well-known/atproto-did/${handle}`?

packages/identifier/tests/resolve.test.ts
```diff
@@ -48,12 +48,12 @@ describe('handle resolution', () => {
   })
 
   it('handles a bad DNS resolution', async () => {
-    await expect(resolveDns('bad.test')).rejects.toThrow(NoHandleRecordError)
+    const did = resolveDns('bad.test')
```
Evalprime, line 51: should this be an await statement? it's failing tests

Evalprime (commented):
line 51 in the resolve test needs an await statement

packages/identifier/tests/resolve.test.ts
dholms: yup 👍

dholms:
Yeah I think we'll need to wait until we have an actual spec before we can register there

Evalprime:
is the spec documented and will the lexicon work with the site?

lexicons/com/atproto/identity/resolveHandle.json
```diff
@@ -7,6 +7,7 @@
       "description": "Provides the DID of a repo.",
       "parameters": {
         "type": "params",
+        "required": ["handle"],
         "properties": {
           "handle": {
```
pfrazee, line 12: should probs drop the "if not specified" bit of the description (at least it was still in the comment)

pfrazee (approved):
Yep I'm good with this

pfrazee (approved):
Yep I'm good with this

packages/identity/src/handle/index.ts
```diff
@@ -0,0 +1,59 @@
...
+      return undefined
+    }
+    return found[0].slice(PREFIX.length)
+  }
+
+  async resolveHttp(
+    handle: string,
+    signal?: AbortSignal,
+  ): Promise<string | undefined> {
+    const url = `https://${handle}/.well-known/atproto-did`
```
devinivy, line 51: Not a must-have, but we might be able to squeeze a little more safety out of this by ensuring any funny business in `handle` gets ignored (i.e. ensure it is treated as a hostname, no additional path parts in there).  One approach to that:
```suggestion
    const url = new URL('/.well-known/atproto-did', `https://${handle}`).href
```

packages/pds/Dockerfile
```diff
@@ -10,8 +10,8 @@ COPY ./packages/aws ./packages/aws
 COPY ./packages/common ./packages/common
 COPY ./packages/common-web ./packages/common-web
 COPY ./packages/crypto ./packages/crypto
-COPY ./packages/did-resolver ./packages/did-resolver
 COPY ./packages/identifier ./packages/identifier
+COPY ./packages/identity ./packages/identity
```
devinivy, line 14: Thanks for remembering this 👍

devinivy (approved):
Good stuff!

packages/identity/src/handle/index.ts
dholms: nice i like 👍

AaronGoldman:
For some reason I was expecting
```
https://${handle}/.well-known/_atproto -> did=${did}
```
To match the DNS form as closely as possible 🤷‍♂️

devinivy (approved):

zfaizal2:
Curious, how are you configuring your domains to support wildcard subdomains (ex. `foobar.bsky.social/.well-known/atproto-did`)

PR #1051, opened 2023-05-16
fix: removed `action` from required (#1037)

myConsciousness:
Hi,

As I wrote recently in #1037, there is no longer `action` parameter in `com.atproto.repo.applyWrites` for `create`, `update` and `dalete`. But there was still an `action` in required, so I simply removed it.


myConsciousness:
Will be fixed in #1040

PR #1053, opened 2023-05-16
more labeling tweaks

bnewbold:
The "sex toy" and "hanging" changes are pretty simple.

Flagging all "underwear" in to "sexually suggestive" is more disruptive. I did some spot checks and posted examples in slack if you want to review. The motivation is to improve overall "NSFW" filtering using the existing toggles/config, weighed against over-sexualizing simple/neutral photos.


dholms (approved):
yeah i think this is reasonable until we get better classification tooling in

PR #1058, opened 2023-05-17
Fix dev env public url

pfrazee:
I need this URL to be correct for images in the dev env


dholms:
Since we're now using the same utils for dev-env + test environments, this broke a bunch of tests.

I moved the same config over to the setup of the network for the dev-env which should fix this for you & prevent it from breaking tests

PR #1060, opened 2023-05-17
Whats hot new

dholms:
Map out a new ranked version of what's hot that applies a similar algorithm to HN so that trending content stays near the top & gets pulled down by gravity


packages/pds/src/feed-gen/whats-hot.ts
```diff
@@ -0,0 +1,105 @@
...
+  if (cursor !== undefined) {
+    const [cursorScore, cursorCid] = cursor.split('::')
+    const cursorInt = parseInt(cursorScore)
+    if (isNaN(cursorInt)) {
+      throw new InvalidRequestError('Malformed cursor')
+    }
+    const maxScore = cursorInt / 1e10
+    builder = builder
+      .where('candidate.score', '<', maxScore)
+      .where('post.cid', '!=', cursorCid)
```
dholms, line 87: this is kinda a goofy check. but it was consistently getting off in the tests because of the minute gravity applied between requests. & in general I think this is a decent enough check to apply when paginating, basically just to stop immediately repeated posts between requests

packages/pds/src/feed-gen/whats-hot.ts
```diff
@@ -0,0 +1,109 @@
...
+  const { ref } = ctx.db.db.dynamic
+
+  // From: https://medium.com/hacking-and-gonzo/how-hacker-news-ranking-algorithm-works-1d9b0cf2c08d
+  // Score = (P-1) / (T+2)^G
+  // where,
+  // P = points of an item (and -1 is to negate submitters vote)
+  // T = time since submission (in hours)
+  // G = Gravity, defaults to 1.8 in news.arc
+
+  const computeScore = sql<number>`${ref(
```
dholms, line 41: not sure how crazy it is to compute this for every post in the last 24 hrs 😅 

we'll need to make it smarter over time

dholms:
Just gonna merge this one into bsky-algos for my ease 🙃 

Feel free to still review separately

PR #1062, opened 2023-05-17
Limit getPostThread recursion into parents

DavidBuchanan314:
This PR implements the changes required for https://github.com/bluesky-social/social-app/pull/723

It limits recursion into parent posts, in exactly the same way as recursion into child posts was previously limited.

Right now it uses the same depth limit value, perhaps there should be a secondary limit if you want to configure a different maximum parent depth vs child depth.

If it's going to be merged, https://github.com/bluesky-social/social-app/pull/723 should be merged first.

### DIFF
--- a/lexicons/app/bsky/feed/getPostThread.json
+++ b/lexicons/app/bsky/feed/getPostThread.json
@@ -9,8 +9,18 @@
         "required": ["uri"],
         "properties": {
           "uri": {"type": "string", "format": "at-uri"},
-          "depth": {"type": "integer"},
-          "parentHeight": {"type": "integer"}
+          "depth": {
+            "type": "integer",
+            "default": 6,
+            "minimum": 0,
+            "maximum": 1000
+          },
+          "parentHeight": {
+            "type": "integer",
+            "default": 80,
+            "minimum": 0,
+            "maximum": 1000
+          }
         }
       },
       "output": {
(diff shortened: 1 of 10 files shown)


dholms:
Nice, I like this & think we can roll it out at some high limit even before the client change releases (better to return it truncated than crash 😅 )

One think I would suggest is making it a separate value that is supplied in the method params (`parentHeight`?) & defaulted quite high for now, until we get the client change out. I forget at what depth the client crashes, but I think it's around ~100 or so

DavidBuchanan314:
Should the default values be specified in the lexicon?

DavidBuchanan314:
If you're prepared to merge this before the client changes, then it's now ready for review. The default `parentHeight` value is 80 (to give a bit of margin on the ~100 it takes to crash mobile clients)

DavidBuchanan314:
By the way, I notice there is also a `atproto/packages/bsky/src/api/app/bsky/feed/getPostThread.ts`, currently this PR only touches `atproto/packages/pds/src/app-view/api/app/bsky/feed/getPostThread.ts`. I assume the changes ought to be mirrored in that file too but I'm not confident I understand the architectural nuances here wrt appview proxying, so I guess I'll leave that for someone who knows what's going on heh

dholms (approved):
looks good 👍 

thanks for popping this back up on my radar

don't worry about the bsky package for now, it's on the back burner for a little bit longer

DavidBuchanan314:
I think this change might need mirroring into the bsky package now ;)

(On the basis that I'm getting server errors trying to view an "infinite" (self-replying) thread in the sandbox)

dholms:
hmm looks like it was mirrored to `bsky`

any idea what the error is?

dholms:
oh lol nvm, we got the change about 95% there but lost the actual check that would stop recursing 🙃

PR #670, opened 2023-03-15
Lex/xrpc methods app migration

dholms:
This is an app migration to go along with https://github.com/bluesky-social/atproto/pull/663

This just brings NSIDs & schema references that are stored in the DB up to date.

PR #1064, opened 2023-05-18
[WIP] Muted keywords init

parion:
This is a server side solution addressing [issue #670](https://github.com/bluesky-social/social-app/issues/670) in the blue-sky repo. Still very much a WIP.
The purpose of this PR is to propose functionality to add keyword muting functionality to AT. A query parameter array `mutedKeywords` is added to `app.bsky.feed.getTimeline`, allowing multiple words to be muted. From there, `composeFeed()` will check the record's text against every muted keyword, foregoing a push if a subtext match is found.

TODO:
- [ ] Exclude replies if they too include muted words
- [ ] Exclude parents elegantly if they include muted words (wondering if this will prove difficult)
- [x] Find a way to allow a `text` property to be retrieved from an unknown record type without the TS yelling at me
- [ ] Implement for `packages/bsky`
- [ ] Unit tests

### DIFF
--- a/lexicons/app/bsky/feed/getTimeline.json
+++ b/lexicons/app/bsky/feed/getTimeline.json
@@ -10,7 +10,11 @@
         "properties": {
           "algorithm": {"type": "string"},
           "limit": {"type": "integer", "minimum": 1, "maximum": 100, "default": 50},
-          "cursor": {"type": "string"}
+          "cursor": {"type": "string"},
+          "mutedKeywords": {
+            "type": "array", 
+            "items": { "type": "string" }
+          }
         }
       },
       "output": {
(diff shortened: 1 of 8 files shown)


Matrix89:
Shouldn't mutes be handled by something like `mutedKeywordsList`? Similarly to how account mutes are handled. That would allow for keyword mutes to be shareable. While passing an argument to `getTimeline` works I wouldn't expect it to change too often.

parion:
Great idea @Matrix89! I think I'll model off of `app.bsky.graph.block` instead as I don't really see the need for shareable lists of muted keywords. Since this will look incredibly different than this query param implementation, I'm going to close this draft.

Matrix89:
Also as per the new contributions guidelines, since this is quite a big feature let's first discuss it with the maintainers and get a sign off

PR #1066, opened 2023-05-18
Rkey len

dholms:
Set limits on rkey size & nsid constraints

### DIFF
--- a/lexicons/com/atproto/repo/applyWrites.json
+++ b/lexicons/com/atproto/repo/applyWrites.json
@@ -52,7 +52,7 @@
       "required": ["action", "collection", "rkey", "value"],
       "properties": {
         "collection": {"type": "string", "format": "nsid"},
-        "rkey": {"type": "string", "maxLength": 15},
+        "rkey": {"type": "string"},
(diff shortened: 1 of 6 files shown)


devinivy (commented):
Looks good!  A few small lingering questions/thoughts on this:
1. Should we be enforcing this again in @atproto/repo to avoid the ability to author repos that couldn't have been authored via the API?  Or at least document the constraint somewhere in the repo code.
2. The size here is fairly small!  I realize it's not the main use-case, but it is interesting that some common id formats such as UUID/GUIDs wouldn't fit here.
3. I wonder if we should loosen the constraint for updates just to support compatibility.  If there's already a record with a long rkey for whatever reason (e.g. migrated from another PDS with incompatible rkey sizes), I don't think there's much harm done to allow those records to be updated.

dholms:
1. yes, but that breaks some tests & I wanna get this in now without going test fixing 😅 
2. i grabbed this from some docs bryan put together with just a hypothesis. totally down to make it larger soon. rather be conservative now
3. that's a good call i'll update that 👍

devinivy (approved):

PR #1067, opened 2023-05-18
Label after write

dholms:
Ensure we're sending a record for labeling after the write occurs

### DIFF
--- a/packages/pds/src/services/repo/index.ts
+++ b/packages/pds/src/services/repo/index.ts
@@ -219,15 +219,13 @@ export class RepoService {
     await sequencer.sequenceEvt(this.db, seqEvt)
 
     // @TODO move to appview
-    this.db.onCommit(() => {
-      writes.map((write) => {
-        if (
-          write.action === WriteOpAction.Create ||
-          write.action === WriteOpAction.Update
-        ) {
-          this.labeler.processRecord(write.uri, write.record)
-        }
-      })
+    writes.forEach((write) => {
+      if (
+        write.action === WriteOpAction.Create ||
+        write.action === WriteOpAction.Update
+      ) {
+        this.labeler.processRecord(write.uri, write.record)
+      }
     })
   }
 
--- a/packages/pds/src/services/repo/index.ts
+++ b/packages/pds/src/services/repo/index.ts
@@ -215,6 +215,9 @@ export class RepoService {
     commitData: CommitData,
     writes: PreparedWrite[],
   ) {
+    const seqEvt = await sequencer.formatSeqCommit(did, commitData, writes)
+    await sequencer.sequenceEvt(this.db, seqEvt)
+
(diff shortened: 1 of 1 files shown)


packages/pds/src/services/repo/index.ts
```diff
@@ -215,18 +215,20 @@ export class RepoService {
...
-        write.action === WriteOpAction.Update
-      ) {
-        this.labeler.processRecord(write.uri, write.record)
-      }
-    })
-
     const seqEvt = await sequencer.formatSeqCommit(did, commitData, writes)
     await sequencer.sequenceEvt(this.db, seqEvt)
+
+    // @TODO move to appview
```
Evalprime, line 221: does this still apply and if so what needs to be done?

packages/pds/src/services/repo/index.ts
```diff
@@ -215,18 +215,20 @@ export class RepoService {
...
+    this.db.onCommit(() => {
```
dholms, line 222: oops this might overwrite notif processing & stuff

don't merge yet

devinivy: Oh dang!  Is it due to a possible bug in `onCommit()`?

devinivy (approved):
Aside from the possible issue you flagged, this looks good.

packages/pds/src/services/repo/index.ts
```diff
@@ -215,18 +215,20 @@ export class RepoService {
...
+      writes.map((write) => {
```
devinivy, line 223:
```suggestion
      writes.forEach((write) => {
```

packages/pds/src/services/repo/index.ts
dholms: yeah, but it's part of a larger architectural chnage. tracking it in a few locations with @TODOs

ericvolp12:
Will content be emitted via the firehose before labels are applied in this case? What happens for like, illegal content? Just checking because I'm sure some folks are storing the entire firehose and/or looking up all blobs and downloading them or something.

devinivy:
That's right, being labeled by the PDS isn't a prerequisite for being emitted from the firehose.  I believe it was that way before the PR as well, since labeling takes place in the background and may take some time to complete.  As soon as content is in a user's repository, it's fair game to emit over the firehose.

The network will have facilities for propagating flagging of illegal content.  In general, though, if you choose to rehost content it's your responsibility to determine if it's illegal in your jurisdiction, and you can't necessarily rely on the publisher of that content (in this case, a PDS) to determine that for you.  In my personal mental model, I think of this as having a lot of similarities to traditional web publishing (e.g. blogs).

dholms:
yup, to add onto that: this labeling in particular is on _images_ which are not & won't ever be transmitted out through the firehose

PR #1074, opened 2023-05-19
Db migration to prep for custom feeds

devinivy:
This is the same migration found in #1001.  Plan is to ship this first just to help things go smooth.


dholms (approved):
yessir!

PR #1078, opened 2023-05-22
Improve performance of whats-hot custom algo

devinivy:
This is an iterative step to improve performance of the main query powering the `whats-hot` custom algo.  By reorganizing the query logic, we're able to get the query planner to sort the result set in parallel which gives us an order of magnitude improvement.  We also proactively exclude results with low like counts, since they are unlikely to land in the feed.


dholms (approved):
nice great ideas 👌

PR #1081, opened 2023-05-23
Fix definitions in bsky

Matrix89:
Lexicon required properties check wasn't rebased onto main before merge which introduced invalid lexicon(creator field was not defined despite being in the required fields list)


devinivy (approved):
Good catch, thank you!

PR #1082, opened 2023-05-23
Tweaks to with-friends algo

devinivy:
After testing out the with-friends algo, it felt a bit too similar to the regular following feed.  We'd also like it to be a little snappier.  The update here surfaces popular posts that have been liked by any of your most active mutuals.

This work is complete, but putting it up as a draft since the behavior itself is possibly not finalized.


dholms (approved):
yup makes a lot of sense 👌

PR #1084, opened 2023-05-23
:sparkles: Allow searching repo by DID

foysalit:
This PR allows `admin.searchRepos` to take `did` as a search term along with handles.

### DIFF
--- a/packages/bsky/src/api/com/atproto/admin/searchRepos.ts
+++ b/packages/bsky/src/api/com/atproto/admin/searchRepos.ts
@@ -4,7 +4,6 @@ import AppContext from '../../../../context'
 import { adminVerifier } from '../../../auth'
 import { paginate } from '../../../../db/pagination'
 import { ListKeyset } from '../../../../services/actor'
-import { ensureValidDid } from '@atproto/identifier'
 
 export default function (server: Server, ctx: AppContext) {
   server.com.atproto.admin.searchRepos({
@@ -17,13 +16,7 @@ export default function (server: Server, ctx: AppContext) {
         throw new InvalidRequestError('The invitedBy parameter is unsupported')
       }
 
-      let searchField: 'did' | 'handle'
-      try {
-        ensureValidDid(term)
-        searchField = 'did'
-      } catch (err) {
-        searchField = 'handle'
-      }
+      const searchField = term.startsWith('did:') ? 'did' : 'handle'
 
       const { ref } = db.db.dynamic
       const keyset = new ListKeyset(ref('indexedAt'), ref('handle'))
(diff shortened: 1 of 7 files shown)


devinivy:
This seems like a reasonable approach to me 👍 I like that this does not have to affect the behavior of `accountService.search()` other places where it's used.

foysalit:
@devinivy ported over to bsky package along with tests. believe it's ready for review.

packages/bsky/src/api/com/atproto/admin/searchRepos.ts
```diff
@@ -16,9 +17,17 @@ export default function (server: Server, ctx: AppContext) {
         throw new InvalidRequestError('The invitedBy parameter is unsupported')
       }
 
+      let searchField: 'did' | 'handle'
+      try {
+        ensureValidDid(term)
+        searchField = 'did'
+      } catch (err) {
+        searchField = 'handle'
+      }
```
Matrix89, line 26: Throwing and then immediately catching is a bit of an anti-pattern IMO(hidden control flow, error creation overhead etc.), but there's no `isValidDid` method so I guess it's ok

dholms: Generally we handle this case with:

`switchField = term.startWith('did:') ? 'did' : 'handle'`

packages/pds/src/services/account/index.ts
```diff
@@ -333,11 +333,25 @@ export class AccountService {
   }
 
   async search(opts: {
+    searchField?: 'did' | 'handle'
     term: string
     limit: number
     cursor?: string
     includeSoftDeleted?: boolean
   }): Promise<(RepoRoot & DidHandle & { distance: number })[]> {
+    // TODO: Implement sqlite specific version?
```
dholms, line 342: I don't see anything sqlite specific here

packages/bsky/tests/views/admin/repo-search.test.ts
dholms, line 1: feels a bit weird to have this one admin test here & none of the others. but i guess it doesn't hurt 😅

packages/bsky/src/api/com/atproto/admin/searchRepos.ts
foysalit: Felt safer to use `ensureDid` but sure, I can replace this with the `startWith()` approach.

foysalit:
@dholms @devinivy addressed all comments and made necessary adjustments, I'd really appreciate another round of review and hopefully a ✅

devinivy (approved):
Looking good!

devinivy (approved):
Looking good!

bnewbold:
I think this is good to merge, and I could push the button? Don't want this to get stale and out of sync with `main`.

devinivy:
@bnewbold thanks for the ping 👍 merged!

PR #1085, opened 2023-05-24
identifier tweaks

bnewbold:
The motivation was to take some of the small proposed changes from identifier specification work and put them in PR form for review here.

The NSID "name" component length change (from 128 to 63): i'm actually now kind of ambivalent after implementing this. One of the motivations was realizing huge names would bloat AT-URIs. But maybe it just doesn't matter!

NSID other restrictions: I think tracking handles to domains very closely has been a good move. I'm not sure the same needs to apply to NSIDs, but it is weird to have extra rules around NSID *domain authority* section other than "valid hostname in reverse order". I am in favor of locking down the "name" section very harshly (to just ASCII letters, either case). We can allow trailing digits or whatever later if we want, once we verify which programming languages do/don't allow that in variable names.

### DIFF
--- a/packages/identifier/src/did.ts
+++ b/packages/identifier/src/did.ts
@@ -39,7 +39,7 @@ export const ensureValidDid = (did: string): void => {
   }
 
   if (did.length > 2 * 1024) {
-    throw new InvalidDidError('DID is far too long')
+    throw new InvalidDidError('DID is too long (2048 chars max)')
   }
 }
 
@@ -51,7 +51,7 @@ export const ensureValidDidRegex = (did: string): void => {
   }
 
   if (did.length > 2 * 1024) {
-    throw new InvalidDidError('DID is far too long')
+    throw new InvalidDidError('DID is too long (2048 chars max)')
   }
 }
 
--- a/packages/identifier/src/did.ts
+++ b/packages/identifier/src/did.ts
@@ -38,7 +38,7 @@ export const ensureValidDid = (did: string): void => {
     throw new InvalidDidError('DID can not end with ":" or "%"')
   }
 
-  if (did.length > 8 * 1024) {
+  if (did.length > 2 * 1024) {
     throw new InvalidDidError('DID is far too long')
   }
 }
@@ -50,7 +50,7 @@ export const ensureValidDidRegex = (did: string): void => {
     throw new InvalidDidError("DID didn't validate via regex")
   }
 
-  if (did.length > 8 * 1024) {
+  if (did.length > 2 * 1024) {
     throw new InvalidDidError('DID is far too long')
   }
 }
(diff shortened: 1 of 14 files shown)


bnewbold:
The tests are failing due to a test NSID named `io.example.stream1` (which would be disallowed by the NSID syntax changes).

Let's review/discuss the syntax changes first, I can clean up all the tests to match our decisions after.

packages/did-resolver/src/web-resolver.ts
```diff
@@ -20,7 +20,9 @@ export class DidWebResolver extends BaseResolver {
     } else if (parts.length === 1) {
       path = parts[0] + DOC_PATH
     } else {
-      path = parts.join('/') + '/did.json'
+      // how we *would* resolve a did:web with path, atproto supported it
+      //path = parts.join('/') + '/did.json'
+      throw new UnsupportedDidWebPathError(did)
```
dholms, line 25: yup 👍

packages/identifier/src/did.ts
```diff
@@ -38,7 +38,7 @@ export const ensureValidDid = (did: string): void => {
     throw new InvalidDidError('DID can not end with ":" or "%"')
   }
 
-  if (did.length > 8 * 1024) {
+  if (did.length > 2 * 1024) {
     throw new InvalidDidError('DID is far too long')
```
dholms, line 42:
```suggestion
    throw new InvalidDidError('DID is far long')
```

AaronGoldman:
ASCII can be combined with a recommendation for how to get Unicode if that is a thing you want to do.

https://en.wikipedia.org/wiki/Punycode

AaronGoldman:
ASCII can be combined with a recomendaton for 

https://en.wikipedia.org/wiki/Punycode

packages/did-resolver/src/web-resolver.ts
```diff
@@ -20,7 +20,9 @@ export class DidWebResolver extends BaseResolver {
     } else if (parts.length === 1) {
       path = parts[0] + DOC_PATH
     } else {
-      path = parts.join('/') + '/did.json'
+      // how we *would* resolve a did:web with path, atproto supported it
```
dholms, line 23:
```suggestion
      // how we *would* resolve a did:web with path, if atproto supported it
```

packages/identifier/src/did.ts
dholms, line 42: since we've limited to 2, can probably just change the error text to "too long" instead of "far too long" 😅

Matrix89: Since other errors return what's valid/invalid, shouldn't this error also mention the max length?

bnewbold:
@dholms made small changes to get CI to pass:

* rebased
* incorporated small review fixes
* fixed errors, including some from rebasing
* renamed NSIDs in a few package tests to not include digits (just spelled out numbers)

No behavioral changes. I plan to merge this later tonight or tomorrow.

dholms (approved):
yup looks good

feel free to merge whenevs 👌

PR #1086, opened 2023-05-24
Materialize candidates for whats-hot algo

devinivy:
The whats-hot algo does some expensive work of paginating over a dynamic score for all post over the past day (i.e. the feed candidates).

This work replaces the subquery to score the candidates with a materialized view that is refreshed [concurrently](https://www.postgresql.org/docs/current/sql-refreshmaterializedview.html#:~:text=Parameters-,CONCURRENTLY,-Refresh%20the%20materialized) every two minutes (configurable).  The materialized view is parameterized using a separate table so that we can alter the like threshold (default: 10) and the interval over which we grab candidates (default: 1day).

One complexity here is around how the refresh is scheduled.  This piggy-backs off the `Leader` abstraction to ensure there's no more than one node maintaining the materialized view at a time.  The creator of the view also needs to be the user who performs the refresh due to postgres limitations.


packages/pds/src/db/migrations/20230523T183902064Z-algo-whats-hot-view.ts
```diff
@@ -0,0 +1,92 @@
...
+        .where('name', '=', 'whats_hot_like_threshold')
+        .select(sql`value::integer`.as('val')),
+    )
+    .select(['post.uri as uri', 'post.cid as cid', computeScore.as('score')])
+
+  await db.schema
+    .createView('algo_whats_hot_view')
+    .materialized()
+    .as(viewQb)
+    .execute()
```
devinivy, line 69: Just wanted to flag this for you @dholms— this will lock `post` and `post_agg` for loading the initial data into the view.  Future refreshes will be done `concurrently` which does not require locking.

dholms: it's not possible to do the initial create concurrently is it?

devinivy: I tried to get that going by creating the view `with no data`, then running a concurrent refresh afterwards, but it doesn't work: apparently in order to perform a concurrent refresh the view already has to be populated.

packages/pds/src/db/migrations/20230523T183902064Z-algo-whats-hot-view.ts
```diff
@@ -0,0 +1,92 @@
+import { Kysely, sql } from 'kysely'
+import { Dialect } from '..'
+
+export async function up(db: Kysely<any>, dialect: Dialect): Promise<void> {
+  if (dialect !== 'pg') return
+
+  const { ref } = db.dynamic
+
+  // materialized views are difficult to change,
```
dholms, line 9: awesome

dholms (approved):
this looks great 🙌

PR #1087, opened 2023-05-24
Feed publishing tweaks

dholms:
Two things here:
- We need a feed generator did provided to the server so that it can properly respond to `describeFeedGenerator`
- We allow `putRecord` for generator records

Note: before, deploy we should add the following DIDs to our service configs:
```
staging: did:plc:go3w5oz6pk4a56zaaprn3lyr
production: did:plc:5fllqkujj6kqp5izd5jg7gox
```

### DIFF
--- a/packages/pds/src/api/com/atproto/repo/putRecord.ts
+++ b/packages/pds/src/api/com/atproto/repo/putRecord.ts
@@ -14,7 +14,11 @@ import {
 } from '../../../../repo'
 import { ConcurrentWriteError } from '../../../../services/repo'
 
-const ALLOWED_PUTS = [ids.AppBskyActorProfile, ids.AppBskyGraphList]
+const ALLOWED_PUTS = [
+  ids.AppBskyActorProfile,
+  ids.AppBskyGraphList,
+  ids.AppBskyFeedGenerator,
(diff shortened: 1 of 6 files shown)


packages/pds/src/api/com/atproto/repo/putRecord.ts
```diff
@@ -14,7 +14,11 @@ import {
 } from '../../../../repo'
 import { ConcurrentWriteError } from '../../../../services/repo'
 
-const ALLOWED_PUTS = [ids.AppBskyActorProfile, ids.AppBskyGraphList]
+const ALLOWED_PUTS = [
+  ids.AppBskyActorProfile,
+  ids.AppBskyGraphList,
+  ids.AppBskyFeedGenerator,
```
devinivy, line 20: Updates to this record time should get indexed just fine, might be worth a double-check though.

devinivy (approved):
Great!

packages/pds/src/api/com/atproto/repo/putRecord.ts
dholms: yup yup i added a quick tests

PR #1080, opened 2023-05-23
Fix lexicon. "format" instead of "ref" in type "string"

MarshalX:
my Python codegen failed on it. let me know pls if the lexicon has been changed and it's not a typo

PR #1088, opened 2023-05-24
Make lexicon validation strict

Matrix89:
This was a quick one and @devinivy supports the idea, so I've decided to do it quickly. 

This makes the whole type of bugs like #1080 impossible. It also fixes `app.bsky.feed.defs` in the bsky package as it was now throwing errors.

The test for this one is a bit meh, since it mostly tests zod's `strict` but I needed something to test with


devinivy:
Reproducing part of my comment here for good measure! https://github.com/bluesky-social/atproto/pull/1080#issuecomment-1560053698
> Re: making the lex document validation strict, I support that. @dholms sound reasonable to you? I suppose there's some thorny compatibility/upgradeability questions there if were actually part of the lexicon spec, but for the purposes of our tooling I think it's a good call.

dholms (approved):
yup i support as well 👍

PR #1090, opened 2023-05-24
Get popular feedgens

devinivy:
Endpoint for fetching top 50 popular feed generators, ranked by like count.


dholms (approved):
yup yup looks great 👍

Issue #1089, opened 2023-05-24
Setting a post thumb to a string instead of a blob ref permanently disables the user account


waxpancake:
Setting the image thumb for a post to a string (e.g. an image URL) via XRPC, instead of passing a blob ref, will permanently disable a user's account, corrupting the repo and leaving it in a read-only state where all writes fail with "500 Internal Server Errors" from both the web app and via XRPC. Posts, likes, replies, deletes, and profile edits all fail with 500 errors. (Logging in, reading posts, and creating app passwords all still function properly.)

Trying to rebase with an authenticated request to the com.atproto.repo.rebaseRepo endpoint will also fail with the same 500 error.

Some speculation from @DavidBuchanan314: "A commit seems to have been interrupted mid-commit (as far as I can tell), so the 'head' pointer is pointing at blocks that do not exist. Probably due to error handling happening at the wrong time." Dumping the repo results in an "an invalid CAR because the 'root' points to a block that is not contained within the CAR."

My account here is currently stuck in this state, if anyone with admin access wants to debug further:

DID: did:plc:l6vfwg5gvscvnmv5xfp4clol

**Update:** Edited the issue to better reflect the problem. See comments below for reproducible test case.

DavidBuchanan314:
`did:plc:xuhbtg3cju5cwz57gthjtyu6` is another example of an account in this state, partial details in this thread https://bsky.app/profile/spv.sh/post/3jvn3cw7x2g2r (it would appear that image uploads are a common factor here)

dholms:
Hey @waxpancake when was the request that you made that caused this?

We did have some issues with leaky transactions but fixed this a few weeks back. 

I'll look into these couple of particular cases here soon

waxpancake:
@dholms It was last Friday, May 19. I posted [this message](https://bsky.app/profile/waxy.org/post/3jw4kyuzafc2x) warning followers that they'd see some test posts from the account while I was debugging it, I was able to post/delete a couple of tests successfully, and then it was nothing but 500 messages after that. 

The last successful request I can see in my logs is a POST to the `com.atproto.repo.uploadBlob` endpoint that returned the link ref `bafkreihtupdlitnktf65zijpi63gn7cni7ltckbwdwta5yizhqbumgy2rq`, an image/png of size 5217, if that helps at all. Every attempted write after that fails with a 500.

waxpancake:
[...]

PR #1093, opened 2023-05-25
Fix incorrect union validation

Matrix89:
Fixes lexicon validation so incorrect types like in #1089 don't validate.
Now when a request like in #1089 is sent a proper 400 is returned :)

### DIFF
--- a/packages/lexicon/src/util.ts
+++ b/packages/lexicon/src/util.ts
@@ -10,6 +10,10 @@ import {
 import { z } from 'zod'
 
 export function toLexUri(str: string, baseUri?: string): string {
+  if (str.split('#').length > 2) {
+    throw new Error('Uri can only have one hash segment')
+  }
+
   if (str.startsWith('lex:')) {
     return str
   }
--- a/packages/lexicon/src/util.ts
+++ b/packages/lexicon/src/util.ts
@@ -41,7 +41,7 @@ export function validateOneOf(
         ),
       }
     }
-    if (!def.refs.includes(toLexUri(value.$type))) {
+    if (!refsContainType(def.refs, value.$type)) {
       if (def.closed) {
         return {
           success: false,
@@ -143,5 +143,20 @@ export function requiredPropertiesRefinement<
         message: `Required field "${field}" not defined`,
       })
     }
+  }
+}
+
+// to avoid bugs like #0189 this needs to handle both
+// explicit and implicit #main
+const refsContainType = (refs: string[], type: string) => {
+  const lexUri = toLexUri(type)
+  if (refs.includes(lexUri)) {
+    return true
+  }
+
+  if (lexUri.endsWith('#main')) {
+    return refs.includes(lexUri.replace('#main', ''))
+  } else {
+    return refs.includes(lexUri + '#main')
   }
 }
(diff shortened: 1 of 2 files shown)


packages/lexicon/src/util.ts
```diff
@@ -145,3 +145,18 @@ export function requiredPropertiesRefinement<
...
+// to avoid bugs like #0189 this needs to handle both
+// explicit and implicit #main
+const refsContainType = (refs: string[], type: string) => {
+  const lexUri = toLexUri(type)
+  if (refs.includes(lexUri)) {
+    return true
+  }
+
+  if (lexUri.endsWith('#main')) {
+    return refs.includes(lexUri.replace('#main', ''))
```
DavidBuchanan314, line 158: This would also accept strings like `com.example#main#main` and simplify to `com.example`

Matrix89: Yep, I took that into consideration, but AFAIK `com.example#main#main` is not a valid type

DavidBuchanan314: I would be inclined to use something like `lexUri.slice(0, -"#main".length)` instead

dholms: that check could also be moved up to `toLexUri`

dholms: there should never be more than one hash fragment in a lex uri

something like `if (str.split('#').length > 2) throw new Error()`

dholms (commented):
Thanks for tackling this! Solution looks great, altho I would like one bit of extra protection against `#main#main`. 

I think validation in lex uri would make sense here

packages/lexicon/src/util.ts
Matrix89: As @dholms suggested I've added a heck and tests to `toLexUri`

Matrix89:
One more thing I've noticed, it feels a bit icky that `validate` can throw because of `toLexUri`(and probably other places). IMO it shouldn't throw as it returns `ValidationResult`

dholms:
Hmmm yeah I agree with you there. I didn't realize that we weren't doing any error handling in `validate` Especially because this is a problem with the type that's being validated not with the lexicon document.

It looks like we do actually want errors to be thrown in other cases where `toLexUri` is used. So we could wrap it here in a try/catch that translates it to a `ValidationResult`

Thanks for pointing that out!

Matrix89:
:thinking: is looks like there are multiple instances where `validate` throws, let's maybe not care about that in this PR. I'm happy to work on that in the future

dholms:
ah kk didn't see those on first scan (i haven't been in this part of the codebase for a second)

dholms (approved):
thanks for the fix!

DavidBuchanan314:
A related thought, it feels like `#main`/"non-main" is a bit of an accident waiting to happen, in general (i.e. having two ways to encode the same thing). Would it be worth defining the "non-main" variant as canonical, and rejecting non-canonical encodings?

dholms:
Hmmm yeah I actually quite like that. We never use `#main` internally

I'd like to get the king of lexicon (@pfrazee)'s opinion on it

pfrazee:
Ah so we basically never allow `#main`, you just have to drop the hash segment to indicate main? I'm good with that

PR #1094, opened 2023-05-25
Catch missed txn rollbacks

dholms:
There is a race condition on transaction execution that looks something like:
- transaction is begun
- query is formatted, passed to executor & through plugins
- error is thrown
- transaction is rolled back
- query is processed by postgres _outside of_ txn

The LeakyTx plugin was meant to catch these, but apparently that plugin is not the absolute last thing done before sending the query.

The fix I applied here feels just a lil bit hacky but I can't see a better approach. We wait for 10ms after receiving an error for any queries to get sent off before throwing the error which will all it to rollback. During that 10ms, we stop any new queries from being sent.

You can see another approach I tried in the commit log where I tracked in-flight queries, but this was failing on queries that would throw (as they don't get passed back through the kysely plugin system)

Closes https://github.com/bluesky-social/atproto/issues/1089

### DIFF
--- a/packages/pds/src/db/index.ts
+++ b/packages/pds/src/db/index.ts
@@ -21,6 +21,7 @@ import { dummyDialect } from './util'
 import * as migrations from './migrations'
 import { CtxMigrationProvider } from './migrations/provider'
 import { dbLogger as log } from '../logger'
+import { wait } from '@atproto/common'
 
 export class Database {
   txEvt = new EventEmitter() as TxnEmitter
@@ -182,7 +183,7 @@ export class Database {
         const txRes = await fn(dbTxn)
           .catch(async (err) => {
             leakyTxPlugin.endTx()
-            await leakyTxPlugin.allSettled()
+            await wait(10)
             throw err
           })
           .finally(() => leakyTxPlugin.endTx())
@@ -302,9 +303,6 @@ type Channels = {
 class LeakyTxPlugin implements KyselyPlugin {
   private txOver: boolean
 
-  inFlightResolves: Record<string, () => void> = {}
-  inFlightPromises: Promise<void>[] = []
-
   endTx() {
     this.txOver = true
   }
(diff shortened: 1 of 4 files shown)


packages/pds/src/db/index.ts
```diff
@@ -179,7 +180,13 @@ export class Database {
       .transaction()
       .execute(async (txn) => {
         const dbTxn = new Database(txn, this.cfg, this.channels)
-        const txRes = await fn(dbTxn).finally(() => leakyTxPlugin.endTx())
+        const txRes = await fn(dbTxn)
+          .catch(async (err) => {
+            leakyTxPlugin.endTx()
+            await wait(10)
```
devinivy, line 186: Alright bear with me, having a thought!

Notice that in the test the query that leaks out is always the second of the two, despite the fact that both start at the same time.  The reason there is that the [`SingleConnectionProvider`](https://github.com/kysely-org/kysely/blob/a6d7bf358d5ee6552eddbd2cce6b4de5f8861a4b/src/driver/single-connection-provider.ts#L4) linearizes each query execution, essentially running a queue.  So the second query is delayed to begin until the first one completes.  You could imagine a longer chain of pending queries here, too.

I think this actually gives us a reliable way to flush the pending queries before kysely issues the rollback.  We could either perform one simple/non-effectful query like `select 1`, or (going lower-level) asking the executor to provide the connection one last time.  In either case, this completes only once other queries already on the queue have completed.  Examples:

```suggestion
            const flushQuery = sql`select 1`.execute(dbTxn.db)
            leakyTxPlugin.endTx()
            await flushQuery
```

or

```suggestion
            leakyTxPlugin.endTx()
            await dbTxn.db.getExecutor().provideConnection(async () => {})
```

dholms: This is great @devinivy 🙌 

I ended up opting for the latter fix. Seems cleaner here, doesn't involve an extra trip, & more clear what's going on. As you said, we can upgrade it if need be

dholms: Also buffed up the tests a bit, so there are ~20 inflight queries when the error is thrown

devinivy: Brilliant 🙏

devinivy (approved):
Yup yup!

PR #1095, opened 2023-05-25
Rebase rework

dholms:
This is a small rework to rebases

The primary change is that instead of walking the repo from the top, we reconstruct it from the records at the bottom.

This does a couple things:
- reduces likelihood of a bug that deletes records
- allows recoverability from an unwalkable repo (ie, in a bad state) 
- prevents records from becoming orphaned if there was bad bookkeeping

This should also entail less back and forth with the database so should be more efficient as well

This also includes a dummy check that counts the number of record blocks before & after the tx and rolls the tx back if the numbers are not the same. This check may be deleted eventually, but is useful now because this is a potentially destructive operation. As long as record blocks are not deleted, we can always reconstruct a repository

### DIFF
--- a/packages/pds/src/services/repo/index.ts
+++ b/packages/pds/src/services/repo/index.ts
@@ -234,32 +234,28 @@ export class RepoService {
     await sequencer.sequenceEvt(this.db, seqEvt)
   }
 
-  // rebases are expensive & should be done rarely, we don't try to re-process on concurrent writes
   async rebaseRepo(did: string, swapCommit?: CID) {
-    this.db.assertTransaction()
+    this.db.assertNotTransaction()
+    const rebaseData = await this.formatRebase(did, swapCommit)
+
+    // rebases are expensive & should be done rarely, we don't try to re-process on concurrent writes
+    await this.serviceTx(async (srvcTx) =>
+      srvcTx.processRebase(did, rebaseData),
+    )
+  }
+
+  async formatRebase(did: string, swapCommit?: CID): Promise<RebaseData> {
     const storage = new SqlRepoStorage(this.db, did, new Date().toISOString())
-    const currRoot = await storage.lockHead()
+    const currRoot = await storage.getHead()
     if (!currRoot) {
-      throw new ConcurrentWriteError()
+      throw new InvalidRequestError(
+        `${did} is not a registered repo on this server`,
+      )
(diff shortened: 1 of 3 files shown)


packages/pds/src/services/repo/index.ts
```diff
@@ -229,43 +234,76 @@ export class RepoService {
...
-      storage.applyRebase(rebaseData),
-      this.afterRebaseProcessing(did, rebaseData),
-    ])
+    const rebaseData = await this.formatRebase(did, currRoot)
+    const recordCountBefore = await this.countRecordBlocks(did)
+    await storage.applyRebase(rebaseData)
+    const recordCountAfter = await this.countRecordBlocks(did)
+    // This is purely a dummy check on a very sensitive operation
+    if (recordCountBefore !== recordCountAfter) {
+      throw new Error('Record blocks deleted during rebase. Rolling back')
```
devinivy, line 254: I wonder if it may be worth including the DID in this error message to ensure we can track it down.

dholms: yeah good call

packages/pds/src/services/repo/index.ts
```diff
@@ -229,43 +234,76 @@ export class RepoService {
...
-        `${did} is not a registered repo on this server`,
-      )
-    }
-    const repo = await Repo.load(storage, currRoot)
-    const rebaseData = await repo.formatRebase(this.repoSigningKey)
-
-    // rebases are expensive & should be done rarely, we don't try to re-process on concurrent writes
-    await this.serviceTx(async (srvcTx) =>
-      srvcTx.processRebase(did, rebaseData, swapCommit),
-    )
```
devinivy, line 247: Is it easy to explain why we switched to do the formatting during the transaction?

devinivy (approved):
👍 !

packages/pds/src/services/repo/index.ts
dholms: basically the lock was my intention. since we don't have any retry logic & want to lock the repo for the duration of the rebase.

that being said, we could still format the rebase ahead of time 🤔

dholms: actually yknow probably better to do this outside of a txn after all

PR #1096, opened 2023-05-25
Improved custom algo debugging

devinivy:
Hoping to make custom algorithms a bit more debuggable here through improved error messages and timing details.
 - In `getFeed` and `getFeedGenerator`, provide better error details for invalid feed DID docs and invalid feed output.
 - Support settings headers from handler in xrpc-server.  Update codegen to support this.
 - Implement helpers for working with the [server-timing](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Server-Timing) header.
 - Add server-timing header to `getFeed` response, breaking down time spent fetching the feed versus hydrating it.


packages/lex-cli/src/codegen/server.ts
```diff
@@ -446,6 +446,10 @@ function genServerXrpcMethod(
         type: 'Uint8Array | stream.Readable',
       })
     }
+    handlerSuccess.addProperty({
+      name: 'headers?',
+      type: '{ [key: string]: string }',
```
dholms, line 451: updated the `Record` type here because it was conflicting with the generated `Record` interface

dholms (approved):
looks great 👍

PR #1098, opened 2023-05-26
Implement quicker mute checks on pds

devinivy:
Previously the query snippet we'd use for excluding a user's mutes was roughly:
```
where not exists ((direct mutes) union all (mutes from lists))
```

Breaking-up the checks for direct mutes and list mutes turns out to be more flexible for the query planner, allowing it to speed some queries up.  So now we roughly do this:
```
where (not exists (direct mutes) and not exists (mutes from lists))
```


dholms (approved):
excellent 👌

PR #1100, opened 2023-05-26
Standalone migration for whats-hot algo materialized view

devinivy:
This is just the migration from #1086 to support release.


dholms (approved):
:shipit:

PR #1107, opened 2023-05-26
Update pool size for view maintainer

devinivy:
This should fix the view maintainer execution: the one connection was getting caught-up holding the lock that chooses the view-maintainer leader 🙃


dholms (approved):
🌻

PR #1110, opened 2023-05-29
Log backgrounded error when refreshing did cache

devinivy:


.github/workflows/build-and-push-pds-aws.yaml
```diff
@@ -3,6 +3,7 @@ on:
   push:
     branches:
       - main
+      - fix-did-cache
```
devinivy, line 6:
```suggestion
```

PR #1111, opened 2023-05-29
Minor perf tweaks for getTimeline and getRepo

devinivy:
- Avoid scanning back further than two weeks on timeline, to avoid worst-case degenerate cases for building feeds.
 - Increase commit chunk size when pulling blocks in `getRepo`, hoping for a positive tradeoff when halving the number of total queries.


dholms (approved):
beautiful 👌

PR #1112, opened 2023-05-29
Admin rebases

dholms:
This adds an admin route for kicking off rebases.

This allows us to perform on-demand rebases for purging user data or recovering a repo from a bad state.

Also gives us more flexibility for running instance-wide rebases


devinivy (approved):
Sweet 👍

PR #1114, opened 2023-05-29
Disallow scheduling repeated view refreshes

devinivy:
We haven't seen this issue actually occur, but in theory if PDSes are spinning then the view refreshes could happen many times in a row.  This avoids that by scheduling the first refresh rather than running it immediately.


dholms (approved):
yup yup 👍

PR #1116, opened 2023-05-29
Followup to getRepo and getTimeline perf tweaks

devinivy:
Followup to #1111, applying some learnings:
 - Changing the chunk size in `getRepo` didn't move the needle, may have even been a slight regression overall.  So this reverts that.
 - Adding a date threshold to `getTimeline` was useful, so applying the same change to with-friends and mutuals custom feeds.


packages/bsky/src/api/app/bsky/util/feed.ts
```diff
@@ -100,3 +100,10 @@ export class FeedKeyset extends TimeCidKeyset<FeedRow> {
     return { primary: result.sortAt, secondary: result.cid }
   }
 }
+
+// For users with sparse feeds, avoid scanning back further than two weeks
+export const getFeedDateThreshold = (days = 14) => {
```
dholms, line 105: did we want to consider 7 days instead?

or maybe within 7 days of the supplied cursor value?

dholms (approved):
perf 👌

packages/bsky/src/api/app/bsky/util/feed.ts
devinivy: Ooh the latter option seems solid: a single page can only get up to a week of content.  That way you could go pretty far back in time as long as your feed doesn't eventually blackhole.  Going to implement that 👍

devinivy: @dholms made those changes, wouldn't mind another peek.

dholms (approved):

PR #1117, opened 2023-05-29
Batch fetching of labels across posts and actors

devinivy:
We don't have to land this, but this was a pretty quick/simple change I wanted to at least consider.

Since labels pervade the whole app.bsky API, the query to fetch labels is consistently the most trafficked in the entire application by around a factor of 2.  The idea here is to nearly halve the number of label queries needed to satisfy feed and thread views.

The implementation allows opting-in to this behavior.
 - `getLabelsForSubjects()` now works for both aturis and dids, aggregating profile labels for dids.  It replaces `getLabelsForProfiles()`.
 - `getLabelsForUris()` is now the lower-level method to query for labels without the special profile/did behavior.
 - When you're hydrating posts and want this behavior:
    1.  tell `getActorViews()` to skip fetching labels with the `skipLabels` arg.
    2. use `getLabelsForSubjects()` to aggregate labels for both posts and actors.
    3. allow `formatPostView()` to assign the actor labels if they are not present (i.e. if they were skipped in `getActorViews()`).

### DIFF
--- a/packages/bsky/src/services/feed/index.ts
+++ b/packages/bsky/src/services/feed/index.ts
@@ -343,7 +343,8 @@ export class FeedService {
     const post = posts[uri]
     const author = actors[post?.creator]
     if (!post || !author) return undefined
-    // If the author labels are not hydrated yet, attempt to pull them from labels
+    // If the author labels are not hydrated yet, attempt to pull them
+    // from labels: e.g. compatible with composeFeed() batching label hydration.
     author.labels ??= labels[author.did] ?? []
     return {
       $type: 'app.bsky.feed.defs#postView',
--- a/packages/bsky/src/services/feed/index.ts
+++ b/packages/bsky/src/services/feed/index.ts
@@ -101,7 +101,7 @@ export class FeedService {
             .as('requesterFollowedBy'),
         ])
         .execute(),
-      this.services.label(this.db).getLabelsForProfiles(dids),
+      this.services.label(this.db).getLabelsForSubjects(dids),
     ])
     return actors.reduce((acc, cur) => {
       return {
--- a/packages/bsky/src/services/feed/index.ts
+++ b/packages/bsky/src/services/feed/index.ts
@@ -234,7 +234,7 @@ export class FeedService {
           viewer,
         ),
         this.embedsForPosts(nestedUris, viewer, _depth + 1),
-        this.services.label(this.db).getLabelsForSubjects(nestedUris),
+        this.services.label(this.db).getLabelsForUris(nestedUris),
       ])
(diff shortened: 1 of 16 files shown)


packages/bsky/src/services/label/index.ts
```diff
@@ -82,36 +82,50 @@ export class LabelService {
...
+      const uri = cur.startsWith('at://') ? new AtUri(cur) : null
+      if (
+        uri &&
+        uri.collection === ids.AppBskyActorProfile &&
+        uri.rkey === 'self'
+      ) {
+        // combine labels for profile + did
+        const did = uri.hostname
+        acc[did] ??= []
+        acc[did].push(...labels[cur])
```
devinivy, line 111:
```suggestion
        acc[did].push(...labels[did])
```

packages/bsky/src/services/feed/index.ts
```diff
@@ -70,6 +70,7 @@ export class FeedService {
   async getActorViews(
     dids: string[],
     viewer: string | null,
+    skipLabels?: boolean, // @NOTE used by composeFeed() to batch label hydration
```
dholms, line 73: for non-obvious boolean vals like this, I kinda like wrapping it in an object so it's clear from the caller what you're asking for like
```ts
getActorViews(dids, requester, { skipLabels: true })
```

totally unnecessary, but could be a good convention to establish. I started doing it in some of our services etc. thoughts?

also in cases where you have a bunch of string parameters that could get mis-arranged or something

packages/bsky/src/services/feed/index.ts
```diff
@@ -341,6 +343,9 @@ export class FeedService {
     const post = posts[uri]
     const author = actors[post?.creator]
     if (!post || !author) return undefined
+    // If the author labels are not hydrated yet, attempt to pull them
```
dholms, line 346: thx for the comment 👌

dholms (approved):
yup looks great 👍

packages/bsky/src/services/feed/index.ts
devinivy: I'm 100% down with that convention, really just reached for this because I thought it was more standard in our codebase 😆 I will tidy that up!

dholms: oops I forgot about this thread & merged the PR lol 🙃

dholms: put it in the quality of life queue lol

devinivy: Hahaha sounds good!

PR #1118, opened 2023-05-29
Return custom suggestions

dholms:
Allow a PDS to return a curated list of suggested follows

These suggested follows are stored in a simple database table.


packages/pds/src/db/migrations/20230529T222706121Z-suggested-follows.ts
```diff
@@ -0,0 +1,12 @@
+import { Kysely } from 'kysely'
+
+export async function up(db: Kysely<unknown>): Promise<void> {
+  await db.schema
+    .createTable('suggested_follow')
+    .addColumn('did', 'varchar', (col) => col.primaryKey())
```
devinivy, line 6: Any use for a column here to control the order?

devinivy (approved):

packages/pds/src/db/migrations/20230529T222706121Z-suggested-follows.ts
dholms: actually yeah that's probably smart 👍 otherwise we just get them in did-alphabetic order

PR #1120, opened 2023-05-30
Revert feed date threshold for with-friends

devinivy:
This optimization didn't turn out to be beneficial for the with-friends algo, so reverting.


dholms (approved):
yup 👍

PR #1124, opened 2023-05-31
Perf tweaks to with-friends, mutuals, label batching

devinivy:
A few changes here:
 - On with-friends algo I switched the filtering to use `originatorDid` rather than `post.creator`.  These are the same value, but the former comes from the base table rather than a joined table.  This seems to open up some options for the query planner, and based on my testing it lands in a much better place.
   - With that change I figured it made sense to reintroduce the date threshold.
 - Same with the mutuals algo.  Also constrained it only to posts, since it was including _all_ reposts of mutuals' posts.
 - Applied more label batching to other places it could easily fit-in.


dholms (approved):
nice yup!

PR #1125, opened 2023-05-31
Revert perf tweak to with-friends

devinivy:
This tweak to with-friends from #1124 worked well in testing but in practice it didn't move the needle for us, so reverting.


dholms (approved):
👍

PR #1126, opened 2023-05-31
Rebase indices

dholms:
This ended up being less than I thought at first:
- one missing index on repo_blob
- slightly adjust a rebase query to ensure we're making use of the index on repo_commit_block


packages/pds/src/sql-repo-storage.ts
```diff
@@ -158,6 +158,7 @@ export class SqlRepoStorage extends RepoStorage {
           .selectFrom('repo_commit_block')
           .selectAll()
           .where('repo_commit_block.creator', '=', this.did)
+          .where('repo_commit_block.commit', '=', rebase.commit.toString())
```
devinivy, line 161: I like that this also makes the query more specific and safer.  It should be the same resultset with or without this condition, but if for some reason it's not, this is more specific and would delete fewer blocks 👍

devinivy (approved):
Relaxed 🌴

packages/pds/src/sql-repo-storage.ts
dholms: no it would actually delete more blocks i believe - this is a clause in a WHERE NOT EXISTS

devinivy: Oh yup of course.  Still seems all good to me.

PR #1129, opened 2023-06-01
Fix pagination limit when not respected by feedgens

devinivy:
If a feedgen doesn't respect the pagination `limit` param, we now enforce it for them.  The `cursor` provided by the feed does still get passed through.


dholms (approved):
excellent 👍

PR #1135, opened 2023-06-02
:sparkles: Hydrate subject author's repo in moderation reports

foysalit:
Looking for initial feedback on the approach I've taken here based on the assumptions that
~~1. All moderation_report entries *always* contain the `subjectDid` which points to the author of the post if the report is for a post, otherwise, the user if the report is for a user.~~
~~2. Profiles may not always exist for a `subjectDid` value in moderation_report.~~
~~3. Hydrating the `subjectRepo` for all moderation reports may be redundant because for reports for users, we already include `repoView` where we're getting some duplicate data but it's ok to do because while there's some overlap in the data, there are some additional properties being exposed with `subjectRepo`~~

Changed the original approach based on initial feedback from reviewers. 
1. We are only exposing the `subjectRepoHandle` on moderation report list view. 

**If this approach looks good, I will go ahead and port it over to `bsky` repo.**


lexicons/com/atproto/admin/defs.json
```diff
@@ -127,144 +167,221 @@
...
+        "creator": { "type": "string", "format": "did" },
+        "handle": { "type": "string", "format": "handle" },
+        "displayName": { "type": "string" },
+        "description": { "type": "string" },
+        "banner": { "type": "string" },
+        "avatar": { "type": "string" },
+        "uri": { "type": "string" },
+        "cid": { "type": "string", "format": "cid" },
+        "indexedAt": { "type": "string", "format": "datetime" }
       }
```
devinivy, line 224: This looks like it is designed to be a view of the `app.bsky.actor.profile` record.  We might want to rethink that a little bit, as we're trying to keep the `com.atproto.*` namespace agnostic to what's happening in `app.bsky*`.

As a way to deal with this, on `#repoView` you'll see there's a `relatedRecords` property which we use to provide miscellaneous app-specific records that might provide context to the repository, including `app.bsky.actor.profile`.  Would that do the trick for us?  If so, perhaps we'd actually be just as well off using a `#repoView` rather than introducing a new kind of repo view.  What do you think?

lexicons/com/atproto/admin/defs.json
```diff
@@ -127,144 +167,221 @@
...
+        "invitedBy": {
+          "type": "ref",
+          "ref": "com.atproto.server.defs#inviteCode"
+        },
+        "invitesDisabled": { "type": "boolean" }
+      }
+    },
+    "subjectRepoView": {
+      "type": "object",
+      "nullable": [
```
devinivy, line 203: Since all of these properties are marked as optional, we have two ways of representing missing data (omitted or `null`), and it might be nice to reduce that down to just one possibility if we can.  It also feels as though we might want some minimal amount of information that should be able to be provided here, whereas now you could receive an object that is empty or only contain `null` values.

packages/pds/src/services/moderation/index.ts
```diff
@@ -486,7 +500,7 @@ export class ModerationService {
     subject: { did: string } | { uri: AtUri; cid?: CID }
     reportedBy: string
     createdAt?: Date
-  }): Promise<ModerationReportRow> {
+  }): Promise<ModerationReportRowWithSubjectRepo | undefined> {
```
devinivy, line 503: I think the contract here would ideally be that it always provides a new report, or it throws an error.  It would be nice if the existence of this new view didn't affect the ability of the report creation method to provide info about the newly-created report.

I notice the joins for `profile` and `did_handle` ended-up getting sprinkled around several places.  One option I could imagine to decouple all these things is to lookup that information in `ModerationView` only when building the `ReportView`, then this method (and some others) wouldn't require any changes.

lexicons/com/atproto/admin/defs.json
foysalit: After rethinking this a bit, I think I was trying to pack too much into this when in reality, we really just need the `handle` for display purposes. avatar was the only other thing I was using from here on the UI at least. So, I think if we just let go of profile entirely and only hydrate this with the handle, that solves the problem of mixing bsky stuff in here along with reducing the scope of this while achieving the primary goal. wdyt @devinivy ?

devinivy: That sounds pretty good!  I would also recommend keeping the `did` around, since that's the primary identifier.  Would there be any downsides to using `#repoView` here?  I think there's an upside to avoid having many subtly different representations of a repo (but only if it actually makes sense for us, of course).

packages/pds/src/services/moderation/index.ts
foysalit: yeah agreed. got rid of the join in almost all places except for the ones where we still need this.

packages/bsky/src/lexicon/lexicons.ts
```diff
@@ -980,6 +983,44 @@ export const schemaDict = {
       },
     },
   },
+  ComAtprotoAdminRebaseRepo: {
```
foysalit, line 986: Some of these changes, along with other bsky changes are from codegen, unrelated to the PR itself.

packages/pds/src/services/moderation/index.ts
```diff
@@ -170,7 +170,8 @@ export class ModerationService {
     }
 
     return await builder
-      .selectAll()
+      .innerJoin('did_handle', 'did_handle.did', 'moderation_report.subjectDid')
```
devinivy, line 173: I believe we may need to make this a left join so that we can serve reports for contents that have been deleted—the inner join will omit results where there's not a matching row in `did_handle`.  In those cases we might not have the handle of the account on hand anymore, and I think we'll just have to omit the optional `subjectRepoHandle` field.

packages/pds/src/services/moderation/views.ts
```diff
@@ -233,6 +234,11 @@ export class ModerationViews {
         .selectFrom('moderation_report')
         .where('subjectType', '=', 'com.atproto.repo.strongRef')
         .where('subjectUri', '=', result.uri)
+        .innerJoin(
+          'did_handle',
+          'did_handle.did',
+          'moderation_report.subjectDid',
+        )
```
devinivy, line 241: Same thing here—I believe that we want to be able to serve reports for subjects that have been deleted, so we'll need to switch to a left join, and omit the `subjectRepoHandle` field when we don't have a matching `did_handle`.

devinivy (commented):
I left one more comment re: still being able to serve reports for deleted subjects, then I think we should be good to go! 🙏

packages/bsky/src/lexicon/lexicons.ts
Matrix89: Yeah, when running codegen in bsky, you have to stage only the relevant changes before commiting, if you're using git from the terminal you can use `git add -p` if not, I think vscode allows you to stage individual hunks

foysalit: thanks but I was told to keep bsky in sync with the lexicons, at least for the time being, so I think these are safe to commit.

foysalit:
@devinivy moved to leftJoin and ported everything over to bsky package as well.

PR #1138, opened 2023-06-02
Add post options

alexmcroberts:
This idea came up as a result of a conversation on Bluesky with @jackmcasey - allowing the `…` menu of a Post View in the Feed to be populated by Post Options provided by 3rd Parties – https://bsky.app/profile/jik.wtf/post/3jwzzhwrsbc27

The screenshot below includes the image from the original post above:

Screenshot of the Bluesky app on iOS, showing a post by Paul Frazee, with the ellipsis highlighted

On a technical level, Post Options would be a simple webhook. This would allow a third party to provide a webhook that can be stored in an actor's repo as a reachable Post Option under the `…` on a Post.

When an actor selects the Post Option from the `…` menu of the Post View, the `callbackUri` of the Post Option would receive the payload, which should include at least the URI of the Post that the Post Option was executed on.

An example – a Zapier action that receives the URI to a Post, and after a certain time, forwards the URI of that Post by email.

To add a Post Option to the `…` menu of the Post View, a 3rd Party would provide a method to run a procedure (i.e. POST the payload of that Post Option (see the defs.json diff) to the actor's Bluesky account)

* The procedure to add a Post Option is defined in `addPostOption.json`
* The query to retrieve Post Options is defined in `getPostOptions.json`
* The definition of a `feedViewPostOption` is added to `defs.json`

### DIFF
--- a/lexicons/app/bsky/feed/addPostOption
+++ b/lexicons/app/bsky/feed/addPostOption
@@ -1,0 +1,20 @@
+{
+  "lexicon": 1,
+  "id": "app.bsky.graph.addPostOption",
+  "defs": {
+    "main": {
+      "type": "procedure",
+      "description": "Add a webhook to a list of post options.",
+      "input": {
+        "encoding": "application/json",
+        "schema": {
+          "type": "object",
+          "required": ["postOption"],
+          "properties": {
+            "postOption": { "type": "object", "ref": "app.bsky.feed.defs#feedViewPostOption" }
+          }
+        }
+      }
+    }
+  }
+}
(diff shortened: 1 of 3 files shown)


dholms:
Hey thanks for the interest & the proposal!

If we did this, I think we'd want to do it with private user preferences (similar to how we handle feed selection/pinning) instead storing it in the repo. Presumably these callback uris would need to include some sort of authentication which would not play nicely with a public repo.

And that flow is actually already supported through `app.bsky.actor.putPreferences` (tho if you make use of this, please ensure you properly namespace your preferences).

I don't think our client will be adding this any time soon (we have other stuff on our plate at the moment), but 3rd part clients are welcome to experiment.

Closing for now, but feel free to keep the discussion going

alexmcroberts:
Thanks for the feedback. The thought initially was for anonymous callback uris.

I fully agree, these should live in private user preferences. I hadn't considered where feed pinning lived, and having used the app, I should have thought of this behaviour.

The concept of namespacing preferences is very useful. 

Thanks for the feedback!

PR #1161, opened 2023-06-03
Prep migration for stream fixing

dholms:
This is a migration to run ahead of https://github.com/bluesky-social/atproto/pull/1134

Still need to write a test to ensure that inserts into the `repo_seq` view work as expected

I sketched out this rough plan. It's a bit tricky, but doable.
- wipe out old repo_seqs ahead of time, as far as we're comfortable to (likely the 24hr backfill period)
- run this migration/deploy
  - rename column & table of `repo_seq` -> `repo_event`
  - create `repo_seq` view for old instances still around
  - create new `outgoing_repo_seq` table
  - temporarily turn off the firehose route, don't error, just don't send events
- run an app migration that populates `outgoing_repo_seq` for all current `repo_event`s such that `seq === eventId` 
- [manually RESET the sequence value](https://stackoverflow.com/questions/16362930/change-the-starting-value-of-a-serial-postgresql) for `outgoing_repo_seq` such that it is hire than current `repo_seq` (with enough wiggle room for migration time)
- deploy new version from https://github.com/bluesky-social/atproto/pull/1134 where a Leader will take over & sequence any repo events that happened in the interim


dholms:
closing in favor of https://github.com/bluesky-social/atproto/pull/1167

PR #1165, opened 2023-06-05
Filter reports by reporter did

foysalit:
**Please don't review/merge**

This is building on top of another in-progress PR.


packages/bsky/src/services/moderation/index.ts
```diff
@@ -110,6 +112,12 @@ export class ModerationService {
       })
     }
 
+    if (reporters?.length) {
+      builder = builder.where((qb) => {
+        return qb.where('reportedByDid', 'in', reporters)
+      })
```
devinivy, line 118: I expect this can be simplified down to this:
```suggestion
      builder = builder.where('reportedByDid', 'in', reporters)
```

The way it's currently written would work fine too— the extra callback essentially just adds some parentheses around the expression in the SQL query.

packages/pds/src/services/moderation/index.ts
```diff
@@ -129,6 +131,12 @@ export class ModerationService {
       })
     }
 
+    if (reporters?.length) {
+      builder = builder.where((qb) => {
+        return qb.where('reportedByDid', 'in', reporters)
+      })
```
devinivy, line 137:
```suggestion
      builder = builder.where('reportedByDid', 'in', reporters)
```

devinivy (approved):
Sweet!

packages/bsky/src/services/moderation/index.ts
foysalit: ah yes, I was in copy-paste mode, didn't pay much attention here. thanks!

PR #1166, opened 2023-06-05
Disable stream

dholms:
In prep for migration, we temporarily disable stream

_Note: tests expected to fail because we've disabled the subscription stream_


devinivy (approved):

dholms:
migration done ✅

closing, actual feature is in https://github.com/bluesky-social/atproto/pull/1167

PR #1167, opened 2023-06-05
Fix dropped events in sync.subscribeRepos

dholms:
This fixes an issue we've had with dropped events while live-tailing the outgoing stream.

We were overly relying on Postgres to help with sequencing, & this instead puts the burden on an elected Leader from the set of instances.

Outgoing events are given an ID on write (from a Postgres Sequence). The elected leader then polls the DB (when notified) to find all unsequenced events. It then assigns them a sequence number from an in-memory clock & updates the relevant rows.

---

The migration path for this is a series of 3 deploys:
- first, disable `sync.subscribeRepos`: https://github.com/bluesky-social/atproto/pull/1166
  - Note: this will not throw an error, it will maintain an open websocket with no events on it
- then do a production-safe DB migration: https://github.com/bluesky-social/atproto/pull/1168 that renames `seq` -> `id` & introduces a new `seq` column
- then, within the DB, run `UPDATE TABLE "repo_seq" SET "seq" = "id";` to set default all currently events to their existing sequence number
- finally, deploy this branch which will, on startup, elect a leader, sequence any new events from the interim & re-enable the sequencer & subscription stream


dholms:
closing in favor of https://github.com/bluesky-social/atproto/pull/1134

packages/pds/src/sequencer/sequencer-leader.ts
```diff
@@ -0,0 +1,128 @@
...
+  // leader = new Leader(SQUENCER_LEADER_ID, this.db)
+  leader: Leader
+
+  destroyed = false
+  polling = false
+  queued = false
+  lastSeq: number
+
+  constructor(public db: Database) {
+    const lockId = this.db.getSchemaAdvisoryLockId(SEQUENCER_LEADER_ID)
```
dholms, line 19: had to add this because these advisory locks were conflicting across tests

packages/pds/tests/sync/subscribe-repos.test.ts
```diff
@@ -161,6 +161,7 @@ describe('repo subscribe repos', () => {
     const isDone = async (evt: any) => {
       if (evt === undefined) return false
       if (evt instanceof ErrorFrame) return true
+      await ctx.sequencerLeader.isCaughtUp()
```
devinivy, line 164: Can this line be removed?

packages/pds/src/api/com/atproto/sync/subscribeRepos.ts
```diff
@@ -27,27 +27,27 @@ export default function (server: Server, ctx: AppContext) {
           message: 'Requested cursor exceeded limit. Possibly missing events',
         }
       }
-      if (curr && cursor > curr.seq) {
+      if (curr && cursor > (curr.seq ?? 0)) {
```
devinivy, line 30: Just double-checking— cursors are always greater than `0`, sound right?

packages/pds/src/sequencer/sequencer-leader.ts
devinivy: We had to deal with this in the appview too and took a slightly different approach— checkout https://github.com/bluesky-social/atproto/blob/46e1f051e169f4a4608d18660ecb165d0cdbf5c8/packages/dev-env/src/bsky.ts#L42C1-L43

packages/pds/src/sequencer/sequencer-leader.ts
```diff
@@ -0,0 +1,127 @@
...
+      .selectFrom('repo_seq')
+      .select('seq')
+      .orderBy('seq', 'desc')
+      .limit(1)
+      .executeTakeFirst()
+    this.lastSeq = res?.seq ?? 0
+
+    while (!this.destroyed) {
+      try {
+        const { ran } = await this.leader.run(async ({ signal }) => {
```
devinivy, line 40: At some point I'll wrap this "forever leader" pattern into a proper util 😆

devinivy, line 40: At some point I'll wrap this "forever leader" pattern into a proper util 😆

packages/pds/src/sequencer/sequencer-leader.ts
```diff
@@ -0,0 +1,127 @@
...
+            if (this.polling) {
+              this.queued = true
+            } else {
+              this.polling = true
+              this.pollDb()
+            }
+          }
+          this.db.channels.new_repo_event.addListener('message', seqListener)
+          await new Promise<void>((resolve) => {
+            signal.addEventListener('abort', () => {
```
devinivy, line 51: If you want to be really defensive you could check if it's already aborted here with `signal.aborted` too.  Not a super practical issue, but could turn-up in tests.

packages/pds/src/sequencer/sequencer-leader.ts
```diff
@@ -0,0 +1,127 @@
...
+  async run() {
+    if (this.db.dialect === 'sqlite') return
+
+    const res = await this.db.db
+      .selectFrom('repo_seq')
+      .select('seq')
+      .orderBy('seq', 'desc')
+      .limit(1)
+      .executeTakeFirst()
+    this.lastSeq = res?.seq ?? 0
```
devinivy, line 36: The `lastSeq` could change by the time this node has become leader, in which case we'd probably move this down into the `leader.run()` callback.

packages/pds/src/sequencer/sequencer-leader.ts
```diff
@@ -0,0 +1,127 @@
...
+  async sequenceOutgoing() {
+    const unsequenced = await this.getUnsequenced()
+    for (const row of unsequenced) {
+      await this.db.db
+        .updateTable('repo_seq')
+        .set({ seq: this.nextSeqVal() })
+        .where('id', '=', row.id)
+        .execute()
+      await this.db.notify('outgoing_repo_seq')
+    }
```
devinivy, line 106: With a little elbow grease this could become a batch update, which we might find pretty worthwhile if we expect `unsequenced` to often contain multiple rows.  Something like:
```sql
UPDATE repo_seq
SET 
  seq = vals.seq
FROM (
  VALUES
    (1, nextSeqVal()),
    (2, nextSeqVal())
) AS vals (id, seq)
WHERE repo_seq.id = vals.id
```

packages/pds/src/sequencer/sequencer-leader.ts
devinivy: I think there's also totally a way to push the `lastSeq` state down into a postgres sequence if that simplifies the bookkeeping in the leader.

packages/pds/src/api/com/atproto/sync/subscribeRepos.ts
dholms: changed this around a bit to avoid possible footguns

`if (curr?.seq && cursor > curr.seq) {`

packages/pds/src/sequencer/sequencer-leader.ts
dholms: haha yeah i thought about doing it here but 🤷‍♂️

packages/pds/src/sequencer/sequencer-leader.ts
dholms: oh shit good call 👍

packages/pds/tests/sync/subscribe-repos.test.ts
devinivy: That makes sense to me in the other places this is checked, but I still feel like I may be missing something here.  Is it possible it should be something similar to this?
```js
const caughtUp = await ctx.sequencerLeader.isCaughtUp()
if (!caughtUp) return false
```

dholms: oh that's absolutely right ^^ good catch

devinivy (approved):

PR #1168, opened 2023-06-06
Seq stream migration

dholms:
This is the actual migration for the database. 

This should be safe to do in production.
For writes: 
- the `id`/`seq` column is not specified in writes, it is generated by the DB
- the _new_ `seq` column is optional & not used on reads

In the previous deploy (https://github.com/bluesky-social/atproto/pull/1166), we've turned off reads so will not have compatibility issues there

_Note: tests expected to fail because we've disabled the subscription stream_


packages/pds/src/db/migrations/20230605T235529700Z-outgoing-repo-seq.ts
```diff
@@ -0,0 +1,44 @@
...
+      .addColumn('id', 'integer', (col) => col.autoIncrement().primaryKey())
+      .addColumn('seq', 'integer', (col) => col.unique())
+      .addColumn('did', 'varchar', (col) => col.notNull())
+      .addColumn('eventType', 'varchar', (col) => col.notNull())
+      .addColumn('event', sql`blob`, (col) => col.notNull())
+      .addColumn('invalidated', 'int2', (col) => col.notNull().defaultTo(0))
+      .addColumn('sequencedAt', 'varchar', (col) => col.notNull())
+      .execute()
+  } else {
+    await db.schema.alterTable('repo_seq').renameColumn('seq', 'id').execute()
```
devinivy, line 18: We may also want to rename the `seq_pkey` constraint to`id_pkey`.

packages/pds/src/db/migrations/20230605T235529700Z-outgoing-repo-seq.ts
```diff
@@ -0,0 +1,44 @@
...
+      .addColumn('seq', 'integer', (col) => col.autoIncrement().primaryKey())
+      .addColumn('did', 'varchar', (col) => col.notNull())
+      .addColumn('eventType', 'varchar', (col) => col.notNull())
+      .addColumn('event', sql`blob`, (col) => col.notNull())
+      .addColumn('invalidated', 'int2', (col) => col.notNull().defaultTo(0))
+      .addColumn('sequencedAt', 'varchar', (col) => col.notNull())
+      .execute()
+  }
+  await db.schema.alterTable('repo_seq').dropColumn('seq').execute()
+  await db.schema.alterTable('repo_seq').renameColumn('id', 'seq').execute()
```
devinivy, line 43:
```suggestion
  } else {
    await db.schema.alterTable('repo_seq').dropColumn('seq').execute()
    await db.schema.alterTable('repo_seq').renameColumn('id', 'seq').execute()
  }
```

packages/pds/src/db/migrations/20230605T235529700Z-outgoing-repo-seq.ts
```diff
@@ -0,0 +1,44 @@
...
+      .addColumn('invalidated', 'int2', (col) => col.notNull().defaultTo(0))
+      .addColumn('sequencedAt', 'varchar', (col) => col.notNull())
+      .execute()
+  } else {
+    await db.schema.alterTable('repo_seq').renameColumn('seq', 'id').execute()
+    await db.schema
+      .alterTable('repo_seq')
+      .addColumn('seq', 'bigint', (col) => col.unique())
+      .execute()
+  }
```
devinivy, line 23: I wonder if this may be a convenient moment to update the values of `id` to that of `seq`.

devinivy (approved):
Dropped a few comments, but approach looks good 👍

devinivy (approved):
Dropped a few comments, but approach looks good 👍

packages/pds/src/db/migrations/20230605T235529700Z-outgoing-repo-seq.ts
dholms: yeah i wasn't sure if that was better to jam in the migration or if we should just run it as a command after. thoughts?

packages/pds/src/db/migrations/20230605T235529700Z-outgoing-repo-seq.ts
dholms: good call 👍 

kysely didn't have a fn for that so I had to hand write it

packages/pds/src/db/migrations/20230605T235529700Z-outgoing-repo-seq.ts
devinivy: We talked about this, seems like the move is to do it manually as originally planned 👍

dholms:
migration done ✅

closing, actual feature is in https://github.com/bluesky-social/atproto/pull/1167

PR #1170, opened 2023-06-06
all packages: prettier applies to tests/ in addition to src/

bnewbold:
Old branch, forgot to send a PR earlier.

This kept tripping me up when doing `yarn prettier:fix` on individual packages.

I don't love that it wanted to lint the JSON file as well, but was easier to just accept the format change instead of figuring out how to disable that beahvior.


dholms (approved):
yup seems fine

dholms (approved):
yup seems fine

PR #1171, opened 2023-06-06
lexicon: maximum report "reason" length of 500 chars (graphemes)

bnewbold:
Discussed integrating a reason text field in the client with Jay and Ansh, we settled on 500 chars, bit longer than a post. Presumably easier to extend to longer in the future.

I could somewhat go either way on this. From a mod team standpoint, we don't want folks writing a huge story in the report reason box. If they need a long conversation or a lot of context, that needs to be some out-of-band thing like an email or google doc or something like that.

On the other hand, any particular max feels kind of arbitrary and restrictive, and third party folks will probably want a different limit/behavior. This is also an atproto-wide mechanism, not restricted to app.bsky content.

So, basically, sending for review, but not sure this is urgent or even the correct values.

### DIFF
--- a/lexicons/com/atproto/moderation/createReport.json
+++ b/lexicons/com/atproto/moderation/createReport.json
@@ -31,7 +31,7 @@
           "properties": {
             "id": {"type": "integer"},
             "reasonType": {"type": "ref", "ref": "com.atproto.moderation.defs#reasonType"},
-            "reason": {"type": "string"},
+            "reason": {"type": "string", "maxGraphemes": 500, "maxLength": 5000},
             "subject": {
               "type": "union",
               "refs": [


dholms (commented):
Yea this is probs fine & we can become more liberal with it over time

Altho I think I would default to 1000

This also seems like a case where we don't care _that_ much about graphemes & can just say "maxLength" (utf8) 1000

dholms:
just a note: we'll want to run codegen before merging

pfrazee:
> Yea this is probs fine & we can become more liberal with it over time

SORT OF

bnewbold:
Agree on max length going to 1000, updated to that. Disagree on graphemes: I think this limit is reasonably important to be consistent (like post size).

Agree that changing this in the future would be a violation of lexicon semantics, in either direction (longer or shorter). Could add a secondary longer field if needed. Do expect that we'll probably also end up iterating on this again, maybe even before full federation, to add things like example reports in a post.

I do feel like we're going to have more folks "coming at us" and stuffing dozens of kbytes of text in random fields like this, so do want to get this in. But also not super high urgency and don't want to cause merge conflicts, so will let this linger until there is a slow day to rebase, codegen, and merge.

bnewbold:
Bumped the size even more, to 2000 graphemes. Thinking is that we do allow/encourage pasting a couple URLs in some cases, and that chews through character limit fast. 2k chars is still basically just a couple paragraphs.

devinivy (approved):

dholms (approved):
seems reasonable 👍 

really just need _some_ limit on this & 2k seems friendly enough

PR #1173, opened 2023-06-07
Sequencer leader error handling

dholms:
This updates error handling in the sequencer leader so that on failure of sequencing, it throws, thereby making the leader no longer the leader


dholms:
We don't need this anymore because we have a dedicated sequencer leader

PR #1174, opened 2023-06-07
Ensure keyset pagination makes use of indexes on timeline

devinivy:
The query snippet we use for keyset pagination wouldn't get picked-up by an index on `(primary-column, secondary-column)` due to its form using `or`.  We now use a "row value" comparison which is simpler to write and indexes know how to use it.  Currently just rolling out on `getTimeline`:  will eventually enable this other places once it's proven to work as intended, and hopefully remove the original implementation.


dholms (approved):
ah nice good find 👍

PR #1175, opened 2023-06-07
Fix lists pagination, ensure paginating on correct column

devinivy:
On `getLists` we were paginating over `indexedAt` but building the cursor based on `createdAt`, which was causing some bad pages to come through.  Now both are based on `createdAt`, which is consistent with how this is handled on similar endpoints such as `getFollows` and `getLikes`.


dholms (approved):
ah rip - good find 👍

PR #1177, opened 2023-06-07
Materialize post and profile aggregations on bsky appview

devinivy:
Port #858 and #859 into the appview, to materialize post and profile aggregations rather than compute them on the fly.  Also does some foundational setup here around:
 - the background queue.
 - putting the labeler on the background queue.
 - logging db pool and background queue metrics.
 - on-commit db hook.

> One possible todo outside of the one-to-one port is initializing post or profile aggregations when first seen, since they could have already existed and have some history prior to the appview seeing them.  I sort of lean towards doing that in a second pass, but open to doing that work here too.


dholms (approved):
looks great! 

i didn't step through individual queries - if you made any substantive changes from PDS impl, mind pointing them out so i can take a closer look?

devinivy:
@dholms no changes there, the aggregation query were able to be one-to-one ports 👍

PR #1183, opened 2023-06-08
Appview lists & listmutes

dholms:
This ports functionality to the appview for:
- indexing list records
- indexing listitem records
- muting/unmuting lists
- serving list views (getList, getLists, getListMutes)

This _does not_ yet include:
- muting/unmuting individuals
- applying mutes to views

---

Implementation notes:
I also ported the graph service & included some methods in the graph service that were on the account service in the pds (muteActorList & unmuteActorList)

PR #1184, opened 2023-06-08
Appview mutes

dholms:
Builds on top of https://github.com/bluesky-social/atproto/pull/1183 to apply mutes in the appview

Implementation notes:
- I removed `unspecced.getPopular` instead of implementing mutes there & bringing the implementation up to date with the PDS
- I moved all mute tests to be in `mutes.test.ts` instead of being spread out over the separate view tests
- I removed proxy tests in the PDS. They were all basically just testing mute application in the PDS. We'll want new tests for generic proxy functionality


packages/bsky/tests/seeds/client.ts
```diff
@@ -383,3 +394,8 @@ export class SeedClient {
     return { authorization: `Bearer ${jwt}` }
   }
 }
+
+// @TODO use real admin auth
+const adminAuth = () => {
+  return ''
+}
```
devinivy, line 401: I'm curious whatsup with this little stub.  I think we _shooould_ have real admin auth working on the appview already.

packages/bsky/tests/views/mutes.test.ts
```diff
@@ -0,0 +1,217 @@
...
+        [bob, carol].includes(notif.author.did),
+      ),
+    ).toBeFalsy()
+  })
+
+  it('flags muted accounts in get suggestions', async () => {
+    // unfollow so they _would_ show up in suggestions if not for mute
+    await sc.unfollow(alice, bob)
+    await sc.unfollow(alice, carol)
+
```
devinivy, line 127: We may need to `processAll()` here.

devinivy (approved):
Solid 💎

Not sure the cause of that last flaky test, but it looks like something minor.

devinivy:
Ah, that flaky test is actually not a flaky test 😆 the failure was suggesting the correct snapshot, so I've updated.

Last thing we might want to do here is apply mutes to the custom algos, which are now merged in here.  We could also hit that when get factored-in.

packages/bsky/tests/views/mutes.test.ts
dholms: yup yup you right 👍

packages/bsky/tests/seeds/client.ts
dholms: yeah it was just a broken import & I stubbed it out

meant to get back to it

dholms: moved it to a param

we should probably move the whole seed client over to dev-env

PR #1187, opened 2023-06-09
Appview blocks

dholms:
Implements blocks on the appview


packages/dev-env/src/network.ts
```diff
@@ -75,6 +75,13 @@ export class TestNetwork extends TestNetworkNoAppView {
     throw new Error(`Sequence was not processed within ${timeout}ms`)
   }
 
+  async processAll(timeout?: number) {
+    await this.pds.ctx.backgroundQueue.processAll()
+    if (!this.bsky) return
+    await this.processFullSubscription(timeout)
+    await this.bsky.ctx.backgroundQueue.processAll()
+  }
```
devinivy, line 83: Yupyup, been wanting to do exactly this 💪

packages/bsky/src/services/feed/index.ts
```diff
@@ -454,3 +462,44 @@ export class FeedService {
...
+  if (!str) return str
+  const encoder = new TextEncoder()
+  const utf8 = encoder.encode(str)
+  if (utf8.length > length) {
+    const decoder = new TextDecoder('utf-8', { fatal: false })
+    const truncated = utf8.slice(0, length)
+    return decoder.decode(truncated).replace(/\uFFFD$/, '')
+  }
+  return str
+}
```
devinivy, line 476: I _think_ we can get rid of this now that all clients are compatible with the lex refactor.

packages/bsky/src/services/indexing/index.ts
```diff
@@ -54,6 +56,7 @@ export class IndexingService {
       profile: Profile.makePlugin(this.db, backgroundQueue),
       list: List.makePlugin(this.db, backgroundQueue),
       listItem: ListItem.makePlugin(this.db, backgroundQueue),
+      block: Block.makePlugin(this.db, backgroundQueue),
```
devinivy, line 59: I think we should remove records from the `actor_block` table during `unindexActor()` too.

packages/bsky/tests/views/blocks.test.ts
```diff
@@ -0,0 +1,329 @@
...
+    const resDan = await agent.api.app.bsky.feed.getTimeline(
+      { limit: 100 },
+      { headers: await network.serviceHeaders(dan) },
+    )
+    expect(
+      resDan.data.feed.some((post) => post.post.author.did === carol),
+    ).toBeFalsy()
+  })
+
+  // it('strips blocked users out of getPopular', async () => {
```
devinivy, line 138: Does this remain commented out intentionally?

devinivy (approved):
Had a few minor thoughts, but this is looking great 🙏

packages/bsky/src/services/feed/index.ts
dholms: ah cool cool checks out 👍 

i couldn't quite remember when we'd added this in & erred on the side of caution

packages/bsky/src/services/indexing/index.ts
dholms: oh yup totally. good call 👍

packages/bsky/tests/views/blocks.test.ts
dholms: yea removed that method from the appview - i'll just remove this test

PR #1188, opened 2023-06-09
Add ability to embed lists

ansh:
Hydrate & format embedded list info in post views. Allows lists to be embedded as records in a post.

Also cleanup a missing `cid` param on some of our list views.

Added to both PDS & AppView.


lexicons/app/bsky/graph/defs.json
```diff
@@ -4,9 +4,10 @@
   "defs": {
     "listViewBasic": {
       "type": "object",
-      "required": ["uri", "name", "purpose"],
+      "required": ["uri", "cid", "name", "purpose"],
       "properties": {
         "uri": {"type": "string", "format": "at-uri"},
+        "cid": {"type": "string", "format": "cid"},
```
devinivy, line 10: Just a note, we'll need to rollout the backend changes here ahead of client changes for compatibility.

ansh: yes, aware, causing type errors on frontend so we arent gonna merge until it is released

cc:@pfrazee

devinivy (approved):
Looks solid! 💎

PR #1189, opened 2023-06-09
Misc. syncing-up bsky appview with on-pds appview

devinivy:
- Custom suggestions from `getSuggestions`.
 - Optimization to `getSuggestions`, making use of aggregation tables.
 - Put notification creation on background queue.
 - Sync-up current profile labeling behavior.
 - Make deletions during actor unindexing serialized, include some missing indexes.
 - Add db indexes to better support actor unindexing.
 - Prevent notifications from deeply-nested threads w/ hellthread exception.


dholms:
ah heads up `getSuggestions` goes off of a hardcoded (for now) `suggested_accounts` table

dholms (approved):
yup yup 👍

PR #1190, opened 2023-06-11
Misc. syncing-up bsky appview with on-pds appview (pt. 2)

devinivy:
Some more odds and ends after #1189:

 - Ensure post and profile aggregations can be maintained out-of-order.
 - Proxy PDS getPopular to appview hot-classic custom feed.
 - Proxy PDS getPopularFeedGenerators to appview.
 - Make `did` field required on feed generator views.


dholms (approved):
looks great 🙌

PR #1191, opened 2023-06-11
Pds proxy to appview based on headers

dholms:
Simple proxying to appview based on existence of headers

We can make this more generic as we remove appview logic from the PDS, but I felt like it might be too premature to do atm 🤔


packages/pds/src/app-view/api/app/bsky/unspecced.ts
```diff
@@ -16,13 +16,39 @@ const NO_WHATS_HOT_LABELS: NotEmptyArray<string> = [
...
+    handler: async ({ req, params, auth }) => {
+      const hotClassicUri = Object.keys(ctx.algos).find((uri) =>
+        uri.endsWith('/hot-classic'),
+      )
+      if (!hotClassicUri) {
+        return {
+          encoding: 'application/json',
+          body: { feed: [] },
+        }
+      }
```
devinivy, line 33: I think this belongs down in the `canProxy()` condition block.

devinivy:
Did a pass on this, looking good to me 👍

packages/pds/tests/_util.ts
```diff
@@ -188,7 +188,7 @@ const basicAuth = (username: string, password: string) => {
 
 // Swap out identifiers and dates with stable
 // values for the purpose of snapshot testing
-export const forSnapshot = (obj: unknown) => {
+export const forSnapshot = (obj: unknown, collapseImgUris?: boolean) => {
```
devinivy, line 191: I see this new param mirroring the change in bsky— think should we remove it or implement it here?

packages/pds/tests/proxied/__snapshots__/feedgen.test.ts.snap
```diff
@@ -780,3 +771,27 @@ Object {
...
+    "uri": "record(0)",
+  },
+}
+`;
+
+exports[`feedgen proxy view proxies getFeedGenerator 2`] = `
+Object {
+  "feeds": Array [],
+}
+`;
```
devinivy, line 797: Doesn't seem like this should be empty, I'm taking a look!

packages/pds/tests/proxied/procedures.test.ts
```diff
@@ -0,0 +1,170 @@
+import AtpAgent from '@atproto/api'
+import { TestNetwork } from '@atproto/dev-env'
+import { SeedClient } from '../seeds/client'
+import basicSeed from '../seeds/basic'
+// import { forSnapshot } from '../_util'
```
devinivy, line 5:
```suggestion
```

packages/pds/tests/_util.ts
dholms: ah we can remove in both places

devinivy (approved):

PR #1192, opened 2023-06-12
Bsky appview: resolve long transactions, tombstone behavior

devinivy:
Some more cleanup in the bsky appview:
 - Now we process each record indexing transactionally, but other indexing operations are not necessarily transactional.  The goal here is to avoid prohibitively long transactions that could affect indexing of other repositories.  A handful of http requests that previously held-up a transaction now occur outside of one.
 - The behavior for indexing tombstones is now loosened from "did doc no longer resolves" to "there is no pds hosting this user's content".
 - The `sync.getHead` endpoint now has an error type to explicitly indicate that the head cannot be found.

I expect it will be easiest to review this [ignoring whitespace](https://github.com/bluesky-social/atproto/pull/1192/files?w=1).

### DIFF
--- a/lexicons/com/atproto/sync/getHead.json
+++ b/lexicons/com/atproto/sync/getHead.json
@@ -25,7 +25,10 @@
             "root": {"type": "string", "format": "cid"}
           }
         }
-      }
+      },
+      "errors": [
+        {"name": "HeadNotFound"}
+      ]
     }
   }
 }
--- a/packages/api/src/client/lexicons.ts
+++ b/packages/api/src/client/lexicons.ts
@@ -3059,6 +3059,11 @@ export const schemaDict = {
             },
           },
         },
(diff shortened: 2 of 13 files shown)


dholms (approved):
sweet looks good 👍

lexicons/com/atproto/sync/getHead.json
```diff
@@ -25,7 +25,10 @@
             "root": {"type": "string", "format": "cid"}
           }
         }
-      }
+      },
+      "errors": [
+        {"name": "HeadNotFound"}
```
dholms, line 30: wondering if we should take this chance to change the name of this route to `getRoot`?

the git metaphor of HEAD makes less sense if we're getting rid of history & we've gotten some.... jokes... about the current name 😅

PR #1194, opened 2023-06-12
Notify crawlers on write

dholms:
Notify crawlers on write.

On write, ping all configured crawlers that a new write is available.

Only do this maximum once every 20 minutes.

### DIFF
--- a/packages/pds/src/crawlers.ts
+++ b/packages/pds/src/crawlers.ts
@@ -26,7 +26,7 @@ export class Crawlers {
           })
         } catch (err) {
           log.warn(
-            { err, service: agent.service.toString() },
+            { err, cralwer: agent.service.toString() },
             'failed to request crawl',
           )
         }
--- a/packages/pds/src/crawlers.ts
+++ b/packages/pds/src/crawlers.ts
@@ -1,0 +1,37 @@
+import { AtpAgent } from '@atproto/api'
+import { crawlerLogger as log } from './logger'
+import { MINUTE } from '@atproto/common'
+
+const NOTIFY_THRESHOLD = 20 * MINUTE
+
+export class Crawlers {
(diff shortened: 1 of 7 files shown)


packages/pds/src/crawlers.ts
```diff
@@ -0,0 +1,37 @@
...
+
+    await Promise.all(
+      this.agents.map(async (agent) => {
+        try {
+          await agent.api.com.atproto.sync.notifyOfUpdate({
+            hostname: this.hostname,
+          })
+        } catch (err) {
+          log.warn(
+            { err, service: agent.service.toString() },
```
devinivy, line 29: The `service` key gets picked-up by some logging software (e.g. datadog) as a special name indicating the source service.  If we could rename to something like `crawler`, that might be a good look.  In the repo subscription logs we had a similar thing happening and ended-up changing it to `provider`.

packages/pds/src/services/repo/index.ts
```diff
@@ -220,6 +225,10 @@ export class RepoService {
     commitData: CommitData,
     writes: PreparedWrite[],
   ) {
+    this.db.onCommit(() => {
+      this.crawlers.notifyOfUpdate()
```
devinivy, line 229: Looks like `notifyOfUpdate()` is designed not to throw, but if I didn't know that I'd probably recommend adding a `catch` here to avoid an unhandled rejection. We could add it anyway if we're feeling defensive.  I suppose I'm not too worried about it either way!

devinivy (approved):
💎

packages/pds/src/services/repo/index.ts
dholms: yup I'll add it in just for peace of mind 👍

PR #1198, opened 2023-06-12
Finalize PDS in-process AppView removal

dholms:
Finalizing the work of removing the AppView from the PDS & syncing this branch (which is being used for the current PDS distribution in sandbox) up with `main`

- Refactor config
- Wipe migrations
- Tidying/clean up


packages/pds/service/index.js
```diff
@@ -12,130 +12,42 @@ require('dd-trace/init') // Only works with commonjs
...
-    keyId: env.plcRotationKeyId,
-  })
-  let recoveryKey
-  if (env.recoveryKeyId.startsWith('did:')) {
-    recoveryKey = env.recoveryKeyId
-  } else {
-    const recoveryKeypair = await KmsKeypair.load({
-      keyId: env.recoveryKeyId,
+  const env = readEnv()
+  env.version ||= pkg.version
```
Matrix89, line 26:
```suggestion
  env.version ??= pkg.version
```

Matrix89 (commented):
Hi, I know this is not ready for a full review, but some semantic stuff I've found.
Also since all the functions in `packages/pds/src/config/util.ts` are prefixed with `env` maybe those could read the variables from the environment, that would make `packages/pds/src/config/env.ts` easier on the eyes

packages/pds/.env.example
```diff
@@ -0,0 +1,24 @@
+# See more env options in src/config/env.ts
```
bnewbold, line 1: Should we merge this file with the existing `packages/pds/example.dev.env`? which gets auto-copied for use with `make run-dev-pds`. I'd otherwise be worried this `.env.example` will be hard to find (default hidden dotfile) and get stale.

If we did that, would need to ensure that the config actually works for local dev, instead of sandbox. I think just commenting out partial strings would work.

packages/pds/src/config/env.ts
```diff
@@ -0,0 +1,173 @@
...
+
+    // blobstore: one required
+    // s3
+    blobstoreS3Bucket: envStr('PDS_BLOBSTORE_S3_BUCKET'),
+    // disk
+    blobstoreDiskLocation: envStr('PDS_BLOBSTORE_DISK_LOCATION'),
+    blobstoreDiskTmpLocation: envStr('PDS_BLOBSTORE_DISK_TMP_LOCATION'),
+
+    // identity
+    didPlcUrl: envStr('PDS_DID_PLC_URL'),
```
bnewbold, line 32: in all the golang code we call this `ATP_PLC_HOST`; fine if they are different or service-specific though.

bnewbold (commented):
Left minor devex-related comments. I would lean towards just merging this huge effort and then address any of those in a follow-on, instead of trying to resolve them first.

In that category: `make run-dev-pds` doesn't work on this branch; I think (and `make run-dev-bsky`) should not maybe point to `services/pds/` and `services/bsky/`?

packages/pds/src/config/env.ts
dholms: yeah we'll probably just keep it server-specific

especially since we already have the `PDS_` prefix on all our env vars

packages/pds/.env.example
dholms: Yup yup that makes sense 👍 

i actually forgot we already had an .env.example

packages/pds/src/db/migrations/20230613T164932261Z-init.ts
```diff
@@ -0,0 +1,375 @@
...
+
+export async function up(db: Kysely<unknown>, dialect: Dialect): Promise<void> {
+  const binaryDatatype = dialect === 'sqlite' ? 'blob' : sql`bytea`
+
+  await db.schema
+    .createTable('app_migration')
+    .addColumn('id', 'varchar', (col) => col.primaryKey())
+    .addColumn('success', 'int2', (col) => col.notNull().defaultTo(0))
+    .addColumn('completedAt', 'varchar', (col) => col)
+    .execute()
```
devinivy, line 14: I wonder if we can get rid of this table and the app migration system.

packages/pds/src/db/migrations/20230613T164932261Z-init.ts
```diff
@@ -0,0 +1,375 @@
...
+    .addColumn('prev', 'varchar')
+    .addPrimaryKeyConstraint('repo_commit_history_pkey', ['creator', 'commit'])
+    .execute()
+
+  await db.schema
+    .createTable('repo_root')
+    .addColumn('did', 'varchar', (col) => col.primaryKey())
+    .addColumn('root', 'varchar', (col) => col.notNull())
+    .addColumn('indexedAt', 'varchar', (col) => col.notNull())
+    .addColumn('takedownId', 'varchar')
```
devinivy, line 268: I think we may need to run a manual migration for the live service to change these cols to varchars.

packages/pds/tests/proxied/views.test.ts
```diff
@@ -167,14 +181,46 @@ describe('proxies view requests', () => {
...
         cursor: pt1.data.cursor,
       },
+      {
+        headers: { ...sc.getHeaders(alice) },
+      },
+    )
+    expect([...pt1.data.feed, ...pt2.data.feed]).toEqual(res.data.feed)
+  })
+
+  it('feed.getListFeed', async () => {
```
devinivy, line 199: I think this test may have gotten duped.

packages/pds/src/storage/disk-blobstore.ts
```diff
@@ -24,14 +24,14 @@ export class DiskBlobStore implements BlobStore {
     this.quarantineLocation = quarantineLocation
   }
 
+  // @TODO move quarantine out of temp by default
```
devinivy, line 27: I think this TODO may be resolved now.

packages/pds/tests/_util.ts
```diff
@@ -33,120 +32,84 @@ export type TestServerOpts = {
...
     ...params,
-  })
+  }
 
-  const db =
-    cfg.dbPostgresUrl !== undefined
-      ? Database.postgres({
-          url: cfg.dbPostgresUrl,
-          schema: cfg.dbPostgresSchema,
-          txLockNonce: cfg.dbTxLockNonce,
```
devinivy, line 115: I can't think of anything that allows us to drop this—is it possible we need to bring it back?  Or maybe I missed the change!

services/pds/index.js
```diff
@@ -12,135 +12,60 @@ require('dd-trace') // Only works with commonjs
...
+  const secrets = envToSecrets(env)
+  const pds = await PDS.create(cfg, secrets)
+  if (cfg.db.dialect === 'pg') {
+    // Migrate using credentialed user
+    const migrateDb = Database.postgres({
+      url: cfg.db.migrationUrl,
+      schema: cfg.db.schema,
     })
-    recoveryKey = recoveryKeypair.did()
+    await migrateDb.migrateToLatestOrThrow()
```
devinivy, line 38: We may not want to run migrations automatically.  Should we keep a separate service entrypoint, or maybe introduce a new config for it?

packages/pds/src/config/config.ts
```diff
@@ -90,6 +90,7 @@ export const envToCfg = (env: ServerEnvironment): ServerConfig => {
     resolverTimeout: env.resolverTimeout ?? 3 * SECOND,
     recoveryDidKey: env.recoveryDidKey ?? null,
     serviceHandleDomains,
+    handleBackupNameservers: env.handleBackupNameservers,
```
devinivy, line 93: Ah, good call.

devinivy (approved):
Looking pretty solid, pretty much just left a few misc questions and notes 💎

packages/dev-env/src/network-no-appview.ts
```diff
@@ -15,12 +15,16 @@ export class TestNetworkNoAppView {
...
       params.dbPostgresSchema || process.env.DB_POSTGRES_SCHEMA
+    const dbSqliteLocation =
+      dbPostgresUrl === undefined ? ':memory:' : undefined
 
     const plc = await TestPlc.create(params.plc ?? {})
     const pds = await TestPds.create({
       dbPostgresUrl,
       dbPostgresSchema,
-      plcUrl: plc.url,
+      dbPostgresPoolSize: 5,
```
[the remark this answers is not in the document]
dholms: nice 👍

packages/pds/src/api/app/bsky/preferences/searchActorsTypeahead.ts
```diff
@@ -0,0 +1,20 @@
...
+      const requester = auth.credentials.did
+      const res =
+        await ctx.appViewAgent.api.app.bsky.actor.searchActorsTypeahead(
+          params,
+          await ctx.serviceAuthHeaders(requester),
+        )
+      return {
+        encoding: 'application/json',
+        body: res.data,
+      }
```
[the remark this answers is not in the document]
dholms: I agree 🤔 

Could see about trying to wire something simple up - would be nice

packages/pds/src/config/config.ts
```diff
@@ -0,0 +1,292 @@
...
+  const subscriptionCfg: ServerConfig['subscription'] = {
+    maxBuffer: env.maxSubscriptionBuffer ?? 500,
+    repoBackfillLimitMs: env.repoBackfillLimitMs ?? DAY,
+    sequencerLeaderEnabled: env.sequencerLeaderEnabled ?? true,
+    sequencerLeaderLockId: env.sequencerLeaderLockId ?? 1100,
+  }
+
+  const bskyAppViewCfg: ServerConfig['bskyAppView'] = {
+    url: env.bskyAppViewUrl ?? 'https://api.bsky-sandbox.dev',
+    did: env.bskyAppViewDid ?? 'did:plc:abc', // get real did
```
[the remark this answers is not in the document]
dholms: I actually may just get rid of the defaults & throw if not included 🤔 

This was added really just for ease of running in sandbox

packages/pds/src/db/migrations/20230613T164932261Z-init.ts
dholms: maybe? it might be helpful with some automatic migrations for the pds distribution in the future tho 🤔 

I think the runtime flag table is asking for it more tbh

services/pds/index.js
dholms: Yup good call - i removed it 👍

packages/pds/tests/_util.ts
dholms: It got moved to `AppContext` 👍

packages/pds/src/db/migrations/20230613T164932261Z-init.ts
dholms: Oh good snag 🤔 

Would we rather these be varchars or integers? Does it matter?

dholms: I can't quite remember why we changed it to a varchar back in the day

packages/pds/.env.example
dholms: alright I actually removed the `make run-dev-pds` & `make run-dev-bsky` commands as well as their env files

Right now, the best way to run these services in dev is in the dev-env the packages themselves aren't set up very well with service entry files. So I also moved over the `dotenv` code to dev-env so that it's env can be adjusted

I don't think we were using those Make commands at all during development, but lmk if this affects your ability to develop.

In the future, it might be worth taking another pass at refactoring running each of theses services in dev mode

packages/pds/src/db/migrations/20230613T164932261Z-init.ts
devinivy: I think it was so that you can store any identifier there, i.e. if the takedown came from some external system that didn't use integers to identify their actions.  Pretty sure it was foreseeing removing some of the moderation machinery from the PDS, making that field less coupled to the current implementation.

Overall it doesn't matter a ton either way, though 👍

packages/pds/.env.example
bnewbold: I do use those make commands fairly often, for the PDS and appview, distinct from `dev-env`. my workflow is basically to run `make run-dev-*` in a bunch of different terminals and a bunch of repos, while editing 1-2 code bases.

but I can try to get all that working in a follow-up PR, don't need to block this one on that side feature

PR #1201, opened 2023-06-13
lexicon: remove 'enum' from string and integer validators

bnewbold:
DO NOT MERGE

This PR is an elaborate way of asking the question: why do we have both 'enum' and 'knownValues' constraints on strings in Lexicon? And should we remove 'enum' in integers, given that we don't support it.

There doesn't seem to be test coverage for 'knownValues'? Though maybe i'm misunderstanding what 'knownValues' actually means.

### DIFF
--- a/packages/lexicon/src/types.ts
+++ b/packages/lexicon/src/types.ts
@@ -22,7 +22,6 @@ export const lexInteger = z
     default: z.number().int().optional(),
     minimum: z.number().int().optional(),
     maximum: z.number().int().optional(),
-    enum: z.number().int().array().optional(),
     const: z.number().int().optional(),
   })
   .strict()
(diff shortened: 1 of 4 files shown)


devinivy:
In my personal opinion we can live without `enum`.  We make good use of `knownValues`, though.  The idea is that it's a list of known values, and it's considered open to extension.  So if the known values are `"red"` and `"green"`, the string `"blue"` is still considered valid. 
 By the same token, `"blue"` can be added as a known value in the future while the lexicon remains both backward- and forward-compatible.

So it's kind of similar to the `union` type without `closed: true`, but for strings.  Maybe there's an argument that `enum` could be like the `union` type _with_ `closed: true`, but for primitives.

bnewbold:
Ah, ok, thanks for clarifying, I didn't understand what `knownValues` meant. I've updated my lexicon spec draft.

Enums do seem to have sharp edges if they are always "closed", and maybe redundant with "union". I'll mention in the spec that we might yank them, but don't see particular urgency resolving thing right now.

bnewbold:
After thinking more, and feedback from Dan, don't think we'll be removing enum, so closing this PR.

PR #1202, opened 2023-06-13
Remove what seems to be cruft from Lexicon types

bnewbold:
Two Lexicon language allowed fields, which don't seem to be used by actual real lexicons, or referenced by any code:

- `infos` field on subscriptions (moved to regular messages)
- `revision` field on top-level Lexicon files

I'm not totally sure about the "revision" value, but suspect it is vestigial.


devinivy:
The `infos` field is definitely gone.  I'm not as sure about `revision`, I think it's worth checking with @pfrazee.

pfrazee:
Revision isn't vestigial -- it's basically present as an advisory notice for when non breaking changes are made to a schema. We haven't used it yet because none of the schemas have been considered stable yet.

bnewbold:
Added a note about revisions to the specs (yesterday, already deployed).

Rebased this PR to just remove the `infos`. Ready to review+merge.

devinivy (approved):

PR #1203, opened 2023-06-13
Makefile: API codegen helper

bnewbold:
Quick helper so I don't need to dig around in multiple packages to remember how to do this.


dholms (commented):
maybe worth adding the equivalent for pds & bsky 

as well as maybe codegen-all to do all three?

looks good tho 👍

dholms (approved):

bnewbold:
Ah! That was exactly what I was forgetting.

I updated to have it do all three packages, and be called `codegen` instead of `codegen-api`.

dholms (approved):
excellent 🙌 

i'm definitely gonna be using this & can't believe i didn't think to do it myself earlier 😅

PR #1205, opened 2023-06-13
Invite epochs

dholms:
Adds an inviteEpoch config variable

This allows us to change the invite interval without giving a big invite drop to old accounts

### DIFF
--- a/packages/pds/tests/account.test.ts
+++ b/packages/pds/tests/account.test.ts
@@ -791,20 +791,3 @@ describe('account', () => {
     }
   })
 })
-
-// describe('invite epochs', () => {
-//   let agent: AtpAgent
-//   let sc: SeedClient
-//   let server: TestServerInfo
-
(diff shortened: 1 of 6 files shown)


packages/pds/tests/account.test.ts
```diff
@@ -765,3 +791,20 @@ describe('account', () => {
...
+//     server = await runTestServer({
+//       dbPostgresSchema: 'invite_epochs',
+//       inviteRequired: true,
+//       userInviteInterval: 1,
+//       userInviteEpoch: Date.now(),
+//     })
+//     agent = new AtpAgent({ service: server.url })
+//     sc = new SeedClient(agent)
+//   })
+// })
```
devinivy, line 810: Looks like this can likely be removed.

devinivy:
```suggestion
```

devinivy (approved):

PR #1215, opened 2023-06-19
Allow optional auth for reporting on bsky appview

devinivy:
The bsky appview will temporarily allow anonymous reports to support early usage in sandbox.


dholms (approved):
yup 👍

Issue #1207, opened 2023-06-15
admin+mod token authz iteration


bnewbold:
We want to do a small short-term iteration on admin tokens to enable better tiered moderation.

Currently there is `ADMIN_PASSWORD`, which gives full admin token privileges, and `MODERATOR_PASSWORD` which can resolve reports with an initial action, but can not reverse actions or do take-downs. Want to end up with three tiers with the following permissions:

* "Triage Moderation" (`TRIAGE_PASSWORD`): can do initial moderation, but only limited publicly-visible actions
    - can not view account emails
    - can create or reverse actions with the actions "acknowledge" and "escalate"
    - can "resolve with current action" for any action type
    - can not create or remove any labels
    - can not change takedown status
    - no control over invite codes
* "Full Moderation" (`MODERATOR_PASSWORD`): can do most moderation actions
    - can view account emails
    - can create and remove any labels
    - can do record take-downs
    - can not do full account take-downs (?)
    - no control over invite codes (?)
* "Admin" (`ADMIN_PASSWORD`): same as current token, full privileges

The idea is that triage moderators do a first pass to escalate reports, then full moderators do both  escalation queue review and regular triage, and then admin is required for full account take-down. This should get us through the next couple months.

Email visibility might require a small lexicon tweak to make that field optional. If we want to bypass the need for redsky work, we could just stuff a fake token in that field to start (eg, `REDACTED`).

The parts that might change would be letting full moderators do account take-downs and invite code control for accounts; or letting triage so some limited labeling. But likely we would just start giving triage agents full moderation power sooner if needed. We will also want "temporary suspension" power for full mods soon, but that doesn't exist yet.

As a heads up, after this is done we would probably want to cycle *all* of the tokens. As a minor UX thing we should ensure our tokens don't have meta-special characters like quotes (single or double); better to just make them quite long with boring characters to improve strength, but 128 bits (UUID size) should be totally strong enough.

cc: @devinivy @dholms

PR #1216, opened 2023-06-19
Add administrative triage role, update moderator role

devinivy:
> ⚠️ please hold off on merge, even once approved

This updates the moderator role and introduces the triage role per the requirements outlined in #1207.  Also updates the basic auth username for the moderator role, which would invalidate existing tokens out there for it (seems appropriate, given that it now has more privileges).  Finally, this introduces the moderator and triage roles to the bsky appview.


packages/bsky/src/api/com/atproto/admin/getRecord.ts
```diff
@@ -1,11 +1,10 @@
 import { InvalidRequestError } from '@atproto/xrpc-server'
 import { Server } from '../../../../lexicon'
 import AppContext from '../../../../context'
-import { adminVerifier } from '../../../auth'
 
 export default function (server: Server, ctx: AppContext) {
   server.com.atproto.admin.getRecord({
-    auth: adminVerifier(ctx.cfg.adminPassword),
+    auth: ctx.adminVerifier,
```
dholms, line 7: we call this `triageVerifier` in the pds & `adminVerifier` here. maybe we sync up the terminology?

It looks like all `triageVerifier` does is also ensure that `credentials.triage` is true (which it always is if `credentials.valid` is true) so probably not a change in logic

I think I like the appview semantics better here 🤔 

altho i wish we had a word to differentiate between "admin as in basic auth triage/moderation/admin" & "admin as in basic auth with admin priveleges"

dholms (approved):
all looks great besides syncing up our terminology on triage v admin

packages/bsky/src/api/com/atproto/admin/getRecord.ts
devinivy: Yup yup 👍 I should be able to work something sensible out there.

devinivy:
@dholms I ended-up reframing this all as "role-based" auth in both pds and appview, and then moving the role checks into the handlers where necessary.  Let me know what you think.

devinivy:
Cool!  @bnewbold @dholms this is good to go, but holding off on the merge since it does change the moderator token credentials and will require a little coordination to roll-out.

PR #1217, opened 2023-06-19
Make requestCrawl & notifyOfUpdate procedures

dholms:
re: https://github.com/bluesky-social/atproto/discussions/1214

Because these are intended to have side effects they really should be POSTs instead of GETs.

Rollout plan: we're intending to support both POSTs & GETs on BGS for some period - likely a week - as this change goes out, then we will be deprecating GETs


devinivy (approved):
Yup, looks good!

PR #1218, opened 2023-06-19
Cleanup @atproto/crypto

dholms:
Cleaned up/revamped `@atproto/crypto` in a few ways:
- upgraded `@noble/secp256k1` to `@noble/curves`
- removed webcrypto dependency in favor of `@noble/curves/p256`
- removed multiformats dependency in favor of `@noble/hashes` (which is a dependency we were already importing for `@noble/curves`)
- Removed AES impl (we weren't using it)
- ensured low-s signatures & verification on P256
- Renamed `EcdsaKeypair` to `P256Keypair`
- Used `@noble/curves` public key compression logic instead of handwritten


devinivy (approved):
Looking good to me!  I wonder if it might be a good moment to dust-off #737 so that we have extra solid confidence that signing and verification isn't only internally-consistent, but also generating identical signatures to before.

packages/crypto/tests/signature-fixtures.json
```diff
@@ -0,0 +1,16 @@
+[
+  {
+    "algorithm": "ES256",
+    "publicKeyDid": "did:key:zDnaeSQxcWyYLQQkQm6kdofoQYZ3FTeBrVDQ7ATsH5x7jDtan",
+    "publicKeyMultibase": "zdSdJswiffwCx8wRyw9h2tEd1Qq4iiqsyDcZqX8Xor7Xg",
+    "messageBase64": "oWVoZWxsb2V3b3JsZA",
+    "signatureBase64": "4x9P4AMMUjMBazqFS+97XQeb1K2bilrbjOwqhUl8d+lykX/UQfT0U3QBePkX0JU0cx7PPrJ46UCU5bA3YNCmYQ"
```
dholms, line 7: actually had to replace this test vector because the previous signature had a `lowS` which we no longer allow

bnewbold:
I'm trying to think if the low-S changes here could cause immediate problems with the golang codebase, and I don't think they should. I haven't verified how the libraries we use in golang handle high-S/low-S.

We don't have golang code doing signatures in our prod systems right now, only verifying signatures, so should not have a situation where typescript starts rejecting sigs.

I can put following up on the golang side and adding interop tests for both key types as a task on my backburner.

cc: @whyrusleeping

dholms:
Yup that's a good call 👍 

Go codebase should be able to handle our signatures either way & since it's not producing signatures, should not break anything.

That being said, should get this into the Go codebase sooner rather than later because we want a consistent understanding of "good" signatures.

Note: I added some test vectors for low v high s: https://github.com/bluesky-social/atproto/pull/1218/files#diff-21f10dec2cf78d8bf828a214cc6eee8a37ca27279062ef032ba5fdaae602e13f

packages/crypto/tests/signature-fixtures.json
devinivy: Oh funny!  I bet you already did this, but I double-checked that the new vectors work just fine without the changes in this PR.

PR #1221, opened 2023-06-20
Surface sequencer leader pg client errors in the logs

devinivy:
The sequencer leader may be aborted on a postgres error.  This ensures that the details about that error make their way into the logs.


dholms (approved):
👍

PR #1222, opened 2023-06-20
Proxy getPopularFeedGenerators on simplified pds

devinivy:
Adds a proxy route for `app.bsky.unspecced.getPopularFeedGenerators` to go from the simplified PDS to the appview.  @dholms just a note I think we'll want to do the same on `main` at some point: #1223 .


dholms (approved):
yup lg 👍

PR #1224, opened 2023-06-21
Appview backfill refactor

dholms:
Building on top of https://github.com/bluesky-social/atproto/pull/898

This provides a methods that will peek at a given stream, determine if backfill is necessary, and if so backfill all repos from that service.

This method is expected to be called similar to an app migration - either in the service entry point or run as a script prior to deploy

---

Note: pointed this at main so that we could tell the diff more easily, but can merge into the original branch if we like


packages/bsky/src/subscription/backfill.ts
```diff
@@ -0,0 +1,149 @@
...
+    .execute()
+}
+
+export const peekStream = async (ctx: AppContext): Promise<number | null> => {
+  const repoProvider = ctx.cfg.repoProvider
+  if (!repoProvider) {
+    throw new Error('No repo provider for backfill')
+  }
+
+  const sub = new Subscription({
```
devinivy, line 48: I can't quite tell if this subscription will get cleaned-up properly.  One way to guarantee it is to pass it a `signal` option, then abort the signal before returning from `peekStream()`.  Alternately could break from a `for await...of` rather than use the raw async iterator.

packages/bsky/src/subscription/backfill.ts
```diff
@@ -0,0 +1,149 @@
...
+        .executeTakeFirst()
+      return lastSeenCursor ? JSON.parse(lastSeenCursor.state) : { cursor: 0 }
+    },
+  })
+  const evts = sub[Symbol.asyncIterator]()
+  const first = await evts.next()
+  // first message should be an OutdatedCursor info msg
+  if (first.done || first.value.name !== 'OutdatedCursor') {
+    return null
+  }
```
devinivy, line 69: Alternately we could ensure to never use params with the subscription here, and take the sequence number of the first event that comes through.

packages/bsky/src/subscription/backfill.ts
```diff
@@ -0,0 +1,149 @@
...
+            { err, provider: repoProvider, repo },
+            'repo subscription backfill failed on a repository',
+          )
+        })
+    })
+    cursor = page.cursor
+    await queue.onEmpty() // Remaining items are in progress
+  } while (cursor)
+
+  // Wait until final batch finishes processing then update cursor.
```
devinivy, line 138:
```suggestion
  // Wait until final batch finishes processing.
```

devinivy (approved):
💪 yup!

dholms:
closing - did this out of band

PR #1225, opened 2023-06-21
Run test DB for full monorepo tests

dholms:
I believe the CI failures in https://github.com/bluesky-social/atproto/pull/1198 come from the following issue:

- Either the bsky or the pds test suite starts the test db
- Both test suites take advantage of it
- the first to finish closes it down
- the second test suite may not be finished with it yet

This moves startup & of the test suite to the monorepo root.

It also alters the pg script file to check if the container is already running. If not, then only the script in charge of starting up the container will tear it down.

Unrelated to this, but I also moved the `on_sigint` handler a bit higher up, so that we don't hit the awkward case where a sigint will kill the container but keep running all the tests.

---

I also noticed a merge conflict on `LOG_ENABLED=false`. Not sure which way we want that flag toggled, but I'd assume off during tests, right? I'd make for pretty noisy output unless we send all logs to some file

### DIFF
--- a/packages/pds/tests/_util.ts
+++ b/packages/pds/tests/_util.ts
@@ -6,6 +6,7 @@ import * as plc from '@did-plc/lib'
 import { PlcServer, Database as PlcDatabase } from '@did-plc/server'
 import { AtUri } from '@atproto/uri'
 import { randomStr } from '@atproto/crypto'
+import { uniqueLockId } from '@atproto/dev-env'
 import { CID } from 'multiformats/cid'
 import * as uint8arrays from 'uint8arrays'
 import { PDS, ServerConfig, Database, MemoryBlobStore } from '../src/index'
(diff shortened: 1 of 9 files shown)


packages/pds/tests/_util.ts
```diff
@@ -159,16 +160,6 @@ export const runTestServer = async (
   }
 }
 
-const usedLockIds = new Set()
-const uniqueLockId = () => {
```
dholms, line 163: snuck this in - I wonder if it was the source of some of our pds test failures on leader - different tests could get the same lock id because we were tracking two separate sets

devinivy: Haha, oh lord!

devinivy (approved):
Good stuff 💖  If this continues to cause headaches for any reason, one alternative approach might be to allow the script to create separate pg containers for each test suite.  But I like this and it seems to be doing the trick.

dholms:
yknow that's not a bad idea & maybe would've been a better option to explore 🤔 would help prevent namespacing collisions between tests as well

ah well let's do this for now & we can come back to it

PR #1229, opened 2023-06-22
Weaker repo lock

dholms:
We were having lock contention between repo writes & an auto-vacuum on the repo_root table.

Use the slightly weaker `SELECT FOR NO KEY UPDATE` instead of `SELECT FOR UPDATE` when locking a repo_root

This lock is the weakest row lock that still conflicts with itself which is the important quality here.

https://www.postgresql.org/docs/current/explicit-locking.html


dholms:
closing in favor of https://github.com/bluesky-social/atproto/pull/1230

PR #1233, opened 2023-06-23
Support proxying moderation through to appview

devinivy:
This adds support to the PDS for proxying moderation reports and actions to an appview.
 - Takedowns and their reversals are in turn applied on the PDS.
 - Resulting labels and their reversals are applied on the PDS.
 - PDS-specific details such as invite status are hydrated onto appview responses.


packages/pds/src/api/com/atproto/admin/takeModerationAction.ts
```diff
@@ -1,16 +1,74 @@
...
+          }
+          if (isStrongRef(result.subject)) {
+            await applyLabels(result.subject.uri, result.subject.cid)
+          }
+        })
+
+        try {
+          await transact
+        } catch (err) {
+          req.log.error(
```
dholms, line 60: what's the recovery from this?

now redsky will show that something is taken down but that isn't reflected on the PDS, right?

this is probably rare, but we should have a plan in mind for how to deal with it 🤔 
would it make sense to fire of a reversal if the txn fails on the PDS? to attempt to keep state in sync? (of course that request can fail as well 😅

packages/pds/tests/seeds/users.ts
```diff
@@ -1,10 +1,10 @@
 import { SeedClient } from './client'
 
-export default async (sc: SeedClient) => {
-  await sc.createAccount('alice', users.alice)
-  await sc.createAccount('bob', users.bob)
-  await sc.createAccount('carol', users.carol)
-  await sc.createAccount('dan', users.dan)
+export default async (sc: SeedClient, invite?: { code: string }) => {
+  await sc.createAccount('alice', { ...users.alice, inviteCode: invite?.code })
```
dholms, line 4: nice 👌

dholms (approved):
code looks good - i'm a bit nervous about half-applied actions

Ideally we don't get into a bad state, but that might be unrealistic & I don't want to overengineer this just for this transition period.

I think at the least we need a good story about how the client will be notified of something going haywire & how we can recover from that state 🤔 

I supposed one option would just be to somewhat rigorously monitor that error log so that we can manually patch it up if it fails (should basically never happen)

packages/pds/src/api/com/atproto/admin/takeModerationAction.ts
devinivy: It's a good question. I was imagining we'd monitor these errors closely and handle them in a high-touch way.  My hunch is that if this happens it's probably not a temporary issue, and something needs fixing.

PR #1234, opened 2023-06-23
Improve invite code counting

dholms:
We no longer count admin-gifted invites against a user routine interval invites

I also took the time to split invites out into their own test suite & made it a bit more robust

### DIFF
--- a/packages/pds/src/api/com/atproto/server/getAccountInviteCodes.ts
+++ b/packages/pds/src/api/com/atproto/server/getAccountInviteCodes.ts
@@ -66,13 +66,16 @@ export default function (server: Server, ctx: AppContext) {
           }))
           await ctx.db.transaction(async (dbTxn) => {
             await dbTxn.db.insertInto('invite_code').values(rows).execute()
-            const forUser = await dbTxn.db
+            const finalRoutineInviteCodes = await dbTxn.db
               .selectFrom('invite_code')
               .where('forUser', '=', requester)
               .where('createdBy', '!=', 'admin') // dont count admin-gifted codes aginast the user
               .selectAll()
               .execute()
-            if (forUser.length > routineCodes.length + toCreate) {
+            if (
+              finalRoutineInviteCodes.length >
+              routineCodes.length + toCreate
+            ) {
               throw new InvalidRequestError(
                 'attempted to create additional codes in another request',
                 'DuplicateCreate',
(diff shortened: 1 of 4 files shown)


packages/pds/src/api/com/atproto/server/getAccountInviteCodes.ts
```diff
@@ -62,9 +69,10 @@ export default function (server: Server, ctx: AppContext) {
             const forUser = await dbTxn.db
               .selectFrom('invite_code')
               .where('forUser', '=', requester)
+              .where('createdBy', '!=', 'admin') // dont count admin-gifted codes aginast the user
               .selectAll()
               .execute()
-            if (forUser.length > userCodes.length + toCreate) {
+            if (forUser.length > routineCodes.length + toCreate) {
```
devinivy, line 75: When I read through this again with the additional "routine" idea in the mix, the `forUser` name tripped me up a little bit.  If it were something like `finalRoutineCodes` it might be a little easier to pick-up what's happening here.

dholms: a good call, I dig 👍

packages/pds/src/api/com/atproto/server/getAccountInviteCodes.ts
```diff
@@ -33,18 +33,25 @@ export default function (server: Server, ctx: AppContext) {
       // note: even if a user is disabled from future invites, we still create the invites for bookkeeping, we just immediately disable them as well
       const now = new Date().toISOString()
       if (createAvailable && ctx.cfg.userInviteInterval !== null) {
+        // for the sake of generating routine interval codes, we do not count explicitly gifted admin codes
+        const routineCodes = userCodes.filter(
+          (code) => code.createdBy !== 'admin',
+        )
+        const unusedRoutineCodes = unusedCodes.filter(
+          (code) => code.createdBy !== 'admin',
+        )
```
devinivy, line 42: Not really a suggestion, but I realized we have all the types we need kicking around to write both the query builder parts and js logic in a similar style if it ever proves useful.  An example for a case like we have here:
```ts
const isRoutine = (code: InviteCode) =>
  code.createdBy !== 'admin'

const whereIsRoutine = (qb: Where<'invite_code'>) =>
  qb.where('createdBy', '!=', 'admin')

// type util to make it a bit prettier
type Where<T extends keyof DatabaseSchemaType> = WhereInterface<DatabaseSchemaType, T>
```

devinivy (approved):
🙏

PR #1235, opened 2023-06-23
Appview limit thread recursion

dholms:
Just a small fix to limit thread recursion in parents in the appview. 

As was pointed out [here](https://github.com/bluesky-social/atproto/pull/1062#issuecomment-1601833585), we aren't limiting thread recursion in the appview.


devinivy (approved):
Thanks for nabbing this 👍

PR #1232, opened 2023-06-22
Optimize reverse chron feeds

dholms:
Some fixes for reverse chron feeds that help out the query analyzer, namely:
- don't check post/account takedowns in feed construction, but instead in feed hydration
- build `with-friends` off of the posts table instead of the feed_items table

PR #1239, opened 2023-06-25
Miscellaneous perf improvements and bsky sync-up

devinivy:
Just some odds and ends!
 - Sync-up the bsky package with changes from #1232.
 - Sync-up the bsky package, applying blocks and mutes on custom algos.
 - Paginate with-friends feed using a condition that hits the post order-by index.
 - Update invite codes base query to be more performant (avoids materializing counts of all codes during the query).
 - Use simpler approach for actor typeahead search query.
   - For full actor search we need to fully support pagination by `(distance, handle)`.  But for the typeahead version we only need to produce one page of results.  That allows us to drop the need to sub-sort profile and account results by `handle`, which is quite a bit easier since it can be done during the trigram index scan.
   - This one lead to a little refactor of the code for the search query.

### DIFF
--- a/packages/bsky/src/feed-gen/with-friends.ts
+++ b/packages/bsky/src/feed-gen/with-friends.ts
@@ -2,7 +2,7 @@ import AppContext from '../context'
 import { QueryParams as SkeletonParams } from '../lexicon/types/app/bsky/feed/getFeedSkeleton'
 import { paginate } from '../db/pagination'
 import { AlgoHandler, AlgoResponse } from './types'
-import { FeedKeyset } from '../api/app/bsky/util/feed'
+import { FeedKeyset, getFeedDateThreshold } from '../api/app/bsky/util/feed'
 
 const handler: AlgoHandler = async (
   ctx: AppContext,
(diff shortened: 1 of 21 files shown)


packages/bsky/src/feed-gen/with-friends.ts
```diff
@@ -11,26 +11,30 @@ const handler: AlgoHandler = async (
...
+        .whereRef('follow.subjectDid', '=', 'post.creator'),
     )
+    .where((qb) =>
+      graphService.whereNotMuted(qb, requester, [ref('post.creator')]),
+    )
+    .whereNotExists(graphService.blockQb(requester, [ref('post.creator')]))
+    .where('post.indexedAt', '>', getFeedDateThreshold(sortFrom))
 
-  postsQb = paginate(postsQb, { limit, cursor, keyset })
+  postsQb = paginate(postsQb, { limit, cursor, keyset, tryIndex: true })
```
dholms, line 37: Can we make `tryIndex` the default now?

dholms (approved):
looks great 👍

packages/bsky/src/feed-gen/with-friends.ts
devinivy: Hmm, that would be nice!  It's still not used in most places.  Maybe we flip a bunch more on soon, and then after that make it the default?

PR #1240, opened 2023-06-26
Unbless hellthread.

alimony:


dholms:
Hellthread is an important part of the culture of the app right now. Many users really love it and would be sad to see it go so we don't want to just remove it.

I would also like to remove the special carveout for hellthread, but I think the solution is to get proper thread mutes on the server for _all_ threads. This would allow us to get rid of the special carveout for hellthread, but also get rid of the arbitrary "5 post depth" rule in general.

alimony:
Good points. I opened this because I’ve seen the tide turn a bit on the fun/annoying balance of the hellthread but, of course, it’s not very important.

bnewbold:
From my perspective: there are pros and cons and ambiguities with hellthread, and they are changing over time, the right path isn't super clear and may or may not be urgent at this moment. But all that said, it is **important**, both to participants and to us.

devinivy:
Replaced by #1289

PR #1242, opened 2023-06-26
Appview - use sortAt

dholms:
Return `sortAt` in appview queries instead of `indexedAt`

This is a composite timestamp that is `min(createdAt, indexedAt)` & will stop a newly spun up appview service from naively returning the time it encountered a post

### DIFF
--- a/packages/pds/src/app-view/api/app/bsky/actor/getProfile.ts
+++ b/packages/pds/src/app-view/api/app/bsky/actor/getProfile.ts
@@ -9,7 +9,6 @@ export default function (server: Server, ctx: AppContext) {
     handler: async ({ req, auth, params }) => {
       const requester = auth.credentials.did
       if (ctx.canProxy(req)) {
-        console.log('HERE')
         const res = await ctx.appviewAgent.api.app.bsky.actor.getProfile(
           params,
           await ctx.serviceAuthHeaders(requester),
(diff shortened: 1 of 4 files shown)


packages/pds/src/app-view/api/app/bsky/actor/getProfile.ts
```diff
@@ -18,6 +19,7 @@ export default function (server: Server, ctx: AppContext) {
           body: res.data,
         }
       }
+      console.log('NO PROXY')
```
devinivy, line 22:
```suggestion
```

packages/pds/src/app-view/api/app/bsky/actor/getProfile.ts
```diff
@@ -9,6 +9,7 @@ export default function (server: Server, ctx: AppContext) {
     handler: async ({ req, auth, params }) => {
       const requester = auth.credentials.did
       if (ctx.canProxy(req)) {
+        console.log('HERE')
```
devinivy, line 12:
```suggestion
```

devinivy (approved):

packages/pds/src/app-view/api/app/bsky/actor/getProfile.ts
dholms: lol oops

PR #1244, opened 2023-06-27
Appview - Invalid handles

dholms:
Deal with invalid handles on the appview
- make handles nullable in actor table
- if `null`, return `"handle.invalid"` as handle
- if revalidating & a handle is no longer valid, then set to `null`
- if another user validly claims an existing handle, invalidate the handle of the previous user (set to `null`)


packages/dev-env/src/util.ts
```diff
@@ -1,13 +1,22 @@
-import { DidResolver, HandleResolver } from '@atproto/identity'
+import { DidResolver, HandleResolver, IdResolver } from '@atproto/identity'
 import { TestPds } from './pds'
+import { TestBsky } from './bsky'
 
-export const mockNetworkUtilities = (pds: TestPds) => {
+export const mockNetworkUtilities = async (pds: TestPds, bsky?: TestBsky) => {
```
dholms, line 5: this change snuck in here because I used the approach of mocking the function on the instance (instead of the prototype) in the handle invalidation test

Do you see any reason why this approach doesn't work?

packages/bsky/src/services/indexing/index.ts
```diff
@@ -127,25 +127,38 @@ export class IndexingService {
...
+    const handleToDid = await this.idResolver.handle.resolve(atpData.handle)
+
+    const handle: string | null = did === handleToDid ? atpData.handle : null
+
+    if (actor && actor.handle !== handle) {
+      const actorWithHandle =
+        handle !== null
+          ? await this.db.db
+              .selectFrom('actor')
+              .where('handle', '=', handle)
```
devinivy, line 140: Just double-checking that `handle` should be normalized here (e.g. not going to miss a match due to casing differences).

packages/bsky/src/services/util/search.ts
```diff
@@ -22,7 +22,7 @@ export const getUserSearchQuery = (
     limit,
     cursor,
     direction: 'asc',
-    keyset: new SearchKeyset(distanceAccount, ref('handle')),
+    keyset: new SearchKeyset(distanceAccount, ref('actor.did')),
```
devinivy, line 25: Not necessarily a bad thing, but this will affect the feel of search results in some cases, since it was sometimes pretty clear they were alphabetical by handle.  It's not uncommon for the distance to be `0` for many matches, especially with short/common `term`s.

In any case, I don't think we need to address it, just a note!

packages/dev-env/src/util.ts
devinivy: Seems like it should be fine 👍

devinivy (approved):
Hey, those changes didn't end-up being so bad! 👍!

packages/bsky/src/services/indexing/index.ts
dholms: that's a good point - the handle comes from the did doc which could contain capitals. I added an extra normalization step on the incoming handle 👍 

so any handle in the db should be all lower

packages/bsky/src/services/util/search.ts
dholms: I had to adjust this to make the type system happy, but I'd be happy to adjust another way if you think that'd be happy

I feel like that's okay tho 🤔 we need to fix up search in general so that it takes into account other ranking factors - proximity, activity level, etc. so results won't always be sorted alphabetical by handle

PR #1246, opened 2023-06-28
PDS - Invalid handles

dholms:
This builds on https://github.com/bluesky-social/atproto/pull/1244

- use similar pagination for PDS implementations of routes
- make `handle` on `did_handle` nullable
- allow for handle contention for external domain handles
- sends an email to a user when their handle is invalidated


packages/pds/src/db/migrations/20230724T194042238Z-optional-handle.ts
```diff
@@ -0,0 +1,51 @@
+import { Kysely, sql } from 'kysely'
+import { Dialect } from '..'
+
+export async function up(db: Kysely<unknown>, dialect: Dialect): Promise<void> {
+  if (dialect === 'pg') {
```
devinivy, line 5: The paranoid part of me wants to flip this condition so that it's more likely to go down the `'pg'` path by mistake than the `'sqlite'` path, since the latter is quite destructive.
```suggestion
  if (dialect !== 'sqlite') {
```

packages/pds/tests/handle-invalidation.test.ts
```diff
@@ -0,0 +1,111 @@
...
+    expect(aliceRes.data.handle).toEqual('handle.invalid')
+
+    const bobRes = await agent.api.app.bsky.actor.getProfile(
+      { actor: bob },
+      { headers: sc.getHeaders(alice) },
+    )
+    expect(bobRes.data.handle).toEqual(HANDLE)
+  })
+
+  it('no longer allows a user to login with a handle they lost due to contention', async () => {
```
devinivy, line 97: As I recall we want to send the user an email in this case.  I realize that's not intended to land in this PR, but just wanted to resurface that in case it's blocking.

devinivy (approved):
A couple misc thoughts in here, but this is looking good 👍

PR #1248, opened 2023-06-29
Appview preset image URLs

devinivy:
This switches the appview from using signed image URLs to fixed, named presets. The purpose is to allow services other than the appview to build valid image URLs without opening the door to cache-busting.

### DIFF
--- a/packages/bsky/src/image/uri.ts
+++ b/packages/bsky/src/image/uri.ts
@@ -1,199 +1,67 @@
-import { createHmac } from 'crypto'
-import * as uint8arrays from 'uint8arrays'
 import { CID } from 'multiformats/cid'
 import { Options } from './util'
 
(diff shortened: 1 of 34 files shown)


dholms (approved):
sweet this looks great. so much slicker 👌

packages/bsky/src/image/uri.ts
```diff
@@ -202,13 +70,33 @@ type BlobLocation = { cid: CID; did: string }
...
+    fit: 'cover',
+    height: 1000,
+    width: 3000,
+    min: true,
+  },
+  feed_thumbnail: {
+    format: 'jpeg',
+    fit: 'inside',
+    height: 2000,
+    width: 2000,
```
davidfraser, line 92: This inadvertently swapped the sizes of thumbnails and fullsize images, so that fullsize images are now a quarter the size of thumbnails. See #3922 and a simple fix in #4077

PR #1250, opened 2023-06-29
Make sequencer leader behavior optional on pds

devinivy:
By making the sequencer leader behavior optional on the pds, it allows deployments to setup a sole, dedicated sequencer leader.


dholms (approved):
nice 👌

PR #1251, opened 2023-06-29
:bug: Only ignore reports for specific at-uri when ignoreSubject contains at-uri

foysalit:
Due to the way postgres handles null values, when we try to ignore reports for a specific `subjectUri`, the previous query was also ignoring all reports with `subjectUri=null` (so basically, all account reports).


devinivy (approved):
👍 !

Issue #1255, opened 2023-06-29
Duplicate blocks in getCheckout


dholms:
We're readding some blocks to the car file for a checkout. Either trees or leaves, need to investigate a bit more.

PR #1256, opened 2023-06-29
Prevent duplicate blocks in checkout

dholms:
We were sending some duplicate blocks in checkout.

This was because we were starting with a basecase of the root node & all it's subtrees before iterating. This small change ended up propagating down each layer of the tree

Closes https://github.com/bluesky-social/atproto/issues/1255


devinivy (approved):
Funky, good find!

PR #1257, opened 2023-06-30
Remove post hierarchy indexing from bsky appview

devinivy:
Here we move away from post hierarchy indexing (supports `getPostThread`) in the bsky appview:
 - It was complex to manage out-of-order indexing of threads in this way.
 - In a world where hellthreads exist, this index quickly becomes quite large.

The change here is to use a recursive query to walk post threads, with some facilities for avoiding cycles.  I made the `post_replyparent_idx` index cover `uri` so that fetching post descendants can be done with an index-only scan, making it pretty snappy.

Also fixes what I believe was a bug causing duplicate notifications when indexing replies seen out-of-order.


dholms (approved):
nice this is excellent 🙌

PR #1262, opened 2023-07-03
Rework record embeds

dholms:
Cleans up logic around record embeds

Also dramatically cuts down on database hits for formatting embeds


devinivy (approved):
💎

PR #1264, opened 2023-07-03
Proxy timeline skeleton construction

dholms:
This is a partial measure to start proxying timeline construction to the appview

The unspecced route will likely only be staying around for a week or two


devinivy (approved):

packages/bsky/src/api/app/bsky/unspecced/getTimelineSkeleton.ts
```diff
@@ -0,0 +1,80 @@
...
+            ref('originatorDid'),
+          ]),
+        )
+        .whereNotExists(
+          graphService.blockQb(viewer, [
+            ref('post.creator'),
+            ref('originatorDid'),
+          ]),
+        )
+        .where('feed_item.sortAt', '>', getFeedDateThreshold(sortFrom))
```
devinivy, line 51: When we launch this we may want to ensure the threshold is set to 1.  Doesn't necessarily have to happen in this PR 👍

PR #1268, opened 2023-07-03
backpressure on bsky backfill indexing

devinivy:


dholms (approved):

packages/bsky/src/background.ts
```diff
@@ -5,7 +5,7 @@ import { dbLogger } from './logger'
 // A simple queue for in-process, out-of-band/backgrounded work
 
 export class BackgroundQueue {
-  queue = new PQueue({ concurrency: 10 })
+  queue = new PQueue({ concurrency: 100 })
```
devinivy, line 8: We may want to roll this change back.

PR #1281, opened 2023-07-04
Explicit dns servers

dholms:
Add ability to set up explicit dns servers to query for handle dns resolution

### DIFF
--- a/packages/identity/src/handle/index.ts
+++ b/packages/identity/src/handle/index.ts
@@ -61,9 +61,9 @@ export class HandleResolver {
   }
 
   async resolveDnsBackup(handle: string): Promise<string | undefined> {
-    const resolver = new dns.Resolver()
     const backupIps = await this.getBackupNameserverIps()
     if (!backupIps || backupIps.length < 1) return undefined
+    const resolver = new dns.Resolver()
     resolver.setServers(backupIps)
     let chunkedResults: string[][]
     try {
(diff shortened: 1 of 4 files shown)


packages/identity/src/handle/index.ts
```diff
@@ -57,4 +59,39 @@ export class HandleResolver {
       return undefined
     }
   }
+
+  async resolveDnsBackup(handle: string): Promise<string | undefined> {
```
devinivy, line 63: I think that if any errors happen here that we want to `return undefined`, similar to the other resolver methods.

packages/identity/src/handle/index.ts
```diff
@@ -57,4 +59,39 @@ export class HandleResolver {
...
+    return found[0].slice(PREFIX.length)
+  }
+
+  private async getBackupNameserverIps(): Promise<string[] | undefined> {
+    if (!this.backupNameservers) {
+      return undefined
+    } else if (!this.backupNameserverIps) {
+      const res = await Promise.all(
+        this.backupNameservers.map((h) => dns.lookup(h)),
+      )
```
devinivy, line 92: Could also do an `allSettled()` here, take the passing ones, and log the failing ones.  If you have a bad one I think that it just causes the dns lookups to occur again and `resolveDnsBackup()` to throw each time it's called.

devinivy (commented):
Just a small note on error handling, beyond that lookin good 👍

packages/identity/src/handle/index.ts
dholms: ah yeah probs smart

PR #1283, opened 2023-07-05
:sparkles: Allow searching reports by moderator did

foysalit:
Allow fetching reports by a moderator did so that only reports that a certain moderator actioned are returned.

### DIFF
--- a/lexicons/com/atproto/admin/getModerationReports.json
+++ b/lexicons/com/atproto/admin/getModerationReports.json
@@ -10,6 +10,11 @@
         "properties": {
           "subject": { "type": "string" },
           "ignoreSubjects": { "type": "array", "items": { "type": "string" } },
+          "actionedBy": {
+            "type": "string",
+            "format": "did",
+            "description": "Get all reports that were actioned by a specific moderator"
+          },
           "reporters": {
             "type": "array",
             "items": { "type": "string" },
--- a/packages/pds/tests/views/admin/get-moderation-reports.test.ts
+++ b/packages/pds/tests/views/admin/get-moderation-reports.test.ts
@@ -243,19 +243,37 @@ describe('pds admin get moderation reports view', () => {
   })
 
   it('gets all moderation reports actioned by a certain moderator.', async () => {
+    const adminDidOne = 'did:example:admin0'
+    const adminDidTwo = 'did:example:admin2'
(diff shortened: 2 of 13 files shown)


packages/pds/tests/views/admin/get-moderation-reports.test.ts
```diff
@@ -241,6 +242,22 @@ describe('pds admin get moderation reports view', () => {
...
+        { headers: { authorization: adminAuth() } },
+      ),
+      agent.api.com.atproto.admin.getModerationReports(
+        { actionedBy: 'did:example:admin2' },
+        { headers: { authorization: adminAuth() } },
+      ),
+    ])
+
+    expect(forSnapshot(actionedByAdminOne.data.reports)).toMatchSnapshot()
+    expect(forSnapshot(actionedByAdminTwo.data.reports)).toMatchSnapshot()
```
devinivy, line 258: Since `forSnapshot` normalizes the did values away, it might just be worth an additional check that each item in `actionedByAdminOne.data.reports` has the expected did in `createdBy`.

devinivy (approved):
Great!

lexicons/com/atproto/admin/getModerationReports.json
```diff
@@ -10,6 +10,11 @@
         "properties": {
           "subject": { "type": "string" },
           "ignoreSubjects": { "type": "array", "items": { "type": "string" } },
+          "actionedBy": {
+            "type": "string",
+            "format": "did",
+            "description": "Get all reports that were actioned by a specific moderator"
+          },
```
devinivy, line 17: Think there's any chance we may want this to be an array in the future?  I don't have a strong opinion on it, but if it seems likely to you, now might be a good time to preemptively accept a list of dids.

foysalit: I thought about it and decided that it probably won't be a use-case any time soon. Can't think of cases where we would want to see reports actioned by multiple folks but LMK if I'm missing anything obvious there.

packages/pds/tests/views/admin/get-moderation-reports.test.ts
foysalit: we don't include `createdBy` for actions in `getModerationReports` so I had to call `getModerationReport` with individual report id to check this. not sure if that breaks some kind of convention we follow for these tests @devinivy

PR #1284, opened 2023-07-05
Misc scaling

dholms:
This was a bit of a catchall branch for dealing with scaling the PDS.

Included in here:
- move labels to an in-memory cache that is routinely refreshed. this abstraction will likely stick around but become more mature over time
- inline mute list queries into post hydration instead of as a separate post
- do sequencing of multiple firehose evts within a tx
- cache blocks from most recent commit before mutating a repo
- increase pagination sizes for sequencer polling & sequencer outbox
- temporarily disable Popular with Friends
- adjust feed date threshold for getTimeline from 3 -> 1

### DIFF
--- a/packages/pds/src/label-cache.ts
+++ b/packages/pds/src/label-cache.ts
@@ -1,0 +1,65 @@
+import { createDeferrable, wait } from '@atproto/common'
+import Database from './db'
+import { Label } from './db/tables/label'
+
+export class LabelCache {
+  bySubject: Record<string, Label[]> = {}
+  latestLabel = ''
+  defer = createDeferrable()
+
+  destroyed = false
+
+  constructor(public db: Database) {}
+
+  async start() {
+    const allLabels = await this.db.db.selectFrom('label').selectAll().execute()
+    this.processLabels(allLabels)
+    this.poll()
+  }
+
+  async poll() {
+    if (this.destroyed) return
+    const labels = await this.db.db
+      .selectFrom('label')
+      .selectAll()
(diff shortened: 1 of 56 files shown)


packages/pds/src/label-cache.ts
```diff
@@ -0,0 +1,90 @@
...
+  defer = createDeferrable()
+  refreshes = 0
+
+  destroyed = false
+
+  constructor(public db: Database) {}
+
+  async start() {
+    await this.fullRefresh()
+    this.poll()
```
devinivy, line 17: This seems unlikely to throw, but for good measure it might make sense to add a `catch()` and log to avoid an unhandled rejection.

packages/pds/src/label-cache.ts
```diff
@@ -0,0 +1,90 @@
...
+    if (this.refreshes >= 120) {
+      await this.fullRefresh()
+      this.refreshes = 0
+    } else {
+      await this.partialRefresh()
+      this.refreshes++
+    }
+    this.defer = createDeferrable()
+    await wait(500)
+    this.poll()
```
devinivy, line 44: Same down here, a `catch()` might be nice to avoid an unhandled rejection (admittedly unlikely).

packages/pds/src/label-cache.ts
dholms: sweet, I ended up just wrapping the execution of `poll` in a try/catch similar to what we do in the sequencer

devinivy (approved):

PR #1294, opened 2023-07-06
Add websocket heartbeat

dholms:
Add a recurring heartbeat to the websocket

Every interval (default 10s) a "ping" is sent
If a "pong" is not sent back within the interval, the websocket is terminated & restarted

Also refactored the code such that we have a `WebSocketKeepAlive` class with all of our heartbeat & retry logic and then a `Subscription` class with the smarts around xrpc.


packages/xrpc-server/src/stream/websocket-keepalive.ts
```diff
@@ -0,0 +1,146 @@
...
+        this.clearHeartbeatInterval()
+      })
+
+      // send a ping on an interval and check to ensure that a pong has come back in that time
+      this.clearHeartbeatInterval() // clean up any previous interval that's stuck around
+      this.isAlive = true
+      // send a ping every x seconds
+      this.heartbeatInterval = setInterval(() => {
+        if (!this.isAlive) {
+          this.ws?.terminate()
```
devinivy, line 62:
```suggestion
          return this.ws?.terminate()
```

packages/xrpc-server/src/stream/websocket-keepalive.ts
```diff
@@ -0,0 +1,146 @@
...
+        } else {
+          throw err
+        }
+      }
+      break // Other side cleanly ended stream and disconnected
+    }
+  }
+
+  clearHeartbeatInterval() {
+    if (this.heartbeatInterval === null) return
```
devinivy, line 100: `heartbeatInterval` isn't initialized to `null`, so this condition might not get triggered when it's intended to.  We could initialize `heartbeatInterval` or make this check looser, but I also don't really think the check is needed if we want to just nix it.

packages/xrpc-server/src/stream/websocket-keepalive.ts
```diff
@@ -0,0 +1,146 @@
...
+  public reconnects: number | null = null
+  public isAlive = false
+  public heartbeatInterval: NodeJS.Timer | null
+
+  constructor(
+    public opts: ClientOptions & {
+      getUrl: () => Promise<string>
+      maxReconnectSeconds?: number
+      signal?: AbortSignal
+      heartbeatInterval?: number
```
devinivy, line 18: Kind of funny naming conundrum here, where "interval" is sort of used in different senses in the option (a duration) versus the class property (a node timer).  Maybe just making the option `heartbeatIntervalMs` or something like that would be a way to distinguish them.

packages/xrpc-server/src/stream/websocket-keepalive.ts
dholms: ah good catch 🙏

packages/xrpc-server/src/stream/websocket-keepalive.ts
dholms: nice i like it 👍

packages/xrpc-server/tests/subscriptions.test.ts
```diff
@@ -344,4 +345,59 @@ describe('Subscriptions', () => {
       ])
     })
   })
+
+  it.only('uses a heartbeat to reconnect if a connection is dropped', async () => {
```
devinivy, line 349:
```suggestion
  it('uses a heartbeat to reconnect if a connection is dropped', async () => {
```

bnewbold:
Flagging that we should do dev comms/outreach on this if it is going to apply to the main prod firehose.

dholms:
> Flagging that we should do dev comms/outreach on this if it is going to apply to the main prod firehose.

Thx for taking a look 🙏 

This is only on the client & is basically an improvement of the keepalive/retry logic. Doesn't affect server implementations (since ping/pong is part of the websockets spec) & not _required_ in client implementations. Just a helpful tool to detect dead connections

devinivy (approved):

PR #1298, opened 2023-07-07
Batch outgoing sequencing

devinivy:
We do batch updates to sequence outgoing events.

We can do this statelessly in a single query using a Postgres sequence - with better guarantees of monotonicity.

We also adjust how the sequencer does polling - it will now queue polls continually as long as it polls more than 0 results.


packages/pds/src/db/migrations/20230718T170914772Z-sequencer-leader-sequence.ts
```diff
@@ -0,0 +1,25 @@
...
+export async function up(db: Kysely<any>, dialect: Dialect): Promise<void> {
+  if (dialect === 'sqlite') return
+  const res = await db
+    .selectFrom('repo_seq')
+    .select('seq')
+    .where('seq', 'is not', null)
+    .orderBy('seq', 'desc')
+    .limit(1)
+    .executeTakeFirst()
+  const startAt = res?.seq ? res.seq + 10000 : 1
```
devinivy, line 13: I think that 10k works for a dedicated sequencer using the quick-reload, but might not have enough wiggle room for a rolling update.  During normal operations 10k events might occur in around 4mins, being slightly conservative.

dholms, line 13: opinion on doing this in the migration vs by hand?

there's some slight worry that during deploy the following happens:
- sequence is created in migration
- old leader steps down
- new leader (with new code) steps up
- new leader crashes/steps down for some reason
- old instance grabs leader  & throws off our bookkeeping

devinivy: (So we could just do it with a quick-reload, or up this number.)

dholms: wow yeah just did the math and... events are coming out fast lol

i'll bump it up to 50k

devinivy: I made a similar comment above, ha!  I agree, it might be smart to do manualy.  I also think that quick-reloads on the dedicated sequencer could help quite a bit.

dholms: really no downside to overshooting a bit

Issue #565, opened 2023-02-17
Provide the dimensions of images in presented views


pfrazee:
The UI has some layout jank with images because it has to guess at a size/aspect-ratio, then set the real sizing once the image has loaded. If the server could provide the dimensions of the images it serves in the presented views, I could eliminate that jank entirely.

pfrazee:
We decided this isnt possible

Issue #427, opened 2022-12-16
Include image aspect ratio in embed views


dholms:

redsolver:
I would love for the aspect ratio to be included in embed views:
- Makes it possible to show images with their native aspect ratios in feeds without the layout jumping when loading images
- Makes it possible to render a gallery view of images (similar to Google Photos) for image-only feeds

It would also be nice to have something like https://evanw.github.io/thumbhash/ for images to show during loading or when blurring out images (without needing to load them).

aliceisjustplaying:
ohmygod yes please please thank you.

JAManfredi:
Aspect ratio will work, but what about just adding height and width to be even more granular? Similar to the Twitter API (https://developer.twitter.com/en/docs/twitter-api/data-dictionary/object-model/media)

redsolver:
Is it ok if I create a PR to add optional width and height `integer` fields to the embed lexicon? That would already make it possible for third-party clients to include that data in a standardized way.

bnewbold:
This issue is quite old (Dec 2022), and i'm going to close it.

We spent a lot of time in Winter 2023 discussing how and whether to include image metadata like height, width, or aspect ratio in embed records, or even in the `blob` built-in type. I can't remember all the reasoning why we didn't, but I think it probably had to do with safely and accurately extracting this metadata. There are a lot of security issues parsing binary file formats and I don't think we wanted to do that in-process; it is also possible for malicious clients/PDS/etc to create incorrect metadata.

Don't want to prevent further conversation; please feel free to create a new issue or discussion about this!

bnewbold:
Cross-reference to related issue: https://github.com/bluesky-social/atproto/issues/565

redsolver:
@bnewbold you are already parsing all image files for thumbnail generation / image resizing for the cdn, could you maybe add the width/height data to the CDN image URLs?

DavidBuchanan314:
IMHO this is all trivially solvable if you let the client figure out the dimensions, and embed them in the post record (specifically, an optional field of `app.bsky.embed.images.image`).

The client can "lie" (or be wrong), but the impact of that would be limited to subtle UI jank for anyone viewing their posts (no more than the current levels of pop-in UI jank!)

PR #1306, opened 2023-07-09
RFC: Support image aspect ratio metadata

DavidBuchanan314:
An optional `aspectRatio` field has been added to the `app.bsky.embed.images.image` and `app.bsky.embed.images.view`  lexicons.

The latter seems a bit redundant, but since the `alt` field was already "mirrored" into the view, I thought it made sense to duplicate the `aspectRatio` field there too.

The `aspectRatio` object has two fields, width and height integers, which encode the aspect ratio together as a pair. 

The aspect ratio of an image is (optionally) detected by clients when a post is being made, and embedded into the post record. This could be encoded as a "standard" aspect ratio like 16:9 (width=16, height=9), or more likely, the client will just give the raw pixel dimensions of the original image, e.g. 1920:1080, because that's the most convenient logic to use. To be totally clear, it doesn't *have* to give the pixel dimensions, it can use any pair of integers that encode the desired ratio (ideally one close to the truth!)

The rationale behind encoding a *ratio* as opposed to the *resolution* is that the ratio stays the same after the image has been scaled/thumbnailed. Furthermore, some supported image formats (like SVG) don't necessarily use pixel units, but ratios are unitless.

A possible edge-case to consider is what happens when an image contains orientation metadata. I'm inclined to say that clients ought to normalise image orientation prior to upload. (but if they don't, the aspect ratio should be stored according to the expected on-screen orientation of the image)

There is no need for the PDS or AppView to parse media files to verify the reported aspect ratio, it can be taken at face value without much consequence. However, clients should still have logic to prevent overly tall or overly wide images from causing UI/UX breakage (but similar logic ought to be there anyway, for images that really do have extreme aspect ratios). If an incorrect aspect ratio is specified, the worst that can happen is just "it looks a bit weird in the UI".

Thanks to @redsolver, @swgws, and others for earlier discussion related to this feature.

Relevant issues:

https://github.com/bluesky-social/atproto/issues/565

https://github.com/bluesky-social/atproto/issues/427


packages/pds/src/app-view/services/feed/views.ts
```diff
@@ -223,6 +223,7 @@ export class FeedViews {
         img.image.ref,
       ),
       alt: img.alt,
+      ...(img.aspectRatio && { aspectRatio: img.aspectRatio }),
```
DavidBuchanan314, line 226: This could probably done more cleanly with a ternary operator or something? I'm bad at JS...

mozzius: I think that’s probably the cleanest way to do it in one line

DavidBuchanan314: In fact, would a simple `aspectRatio: img.aspectRatio,` do the right thing? Looking at the other parts of the same source file, I think an `undefined` value results in the field being omitted from the final record output.

mozzius: I’d have to check the code but the JSON stringifier removes undefined fields so yeah it would almost certainly be fine. The only issue typically is if you do something like `“aspectRatio” in object` but I highly doubt that’s the case here

pfrazee:
Just popping in to say I'm +1 to this idea

packages/pds/src/app-view/services/feed/views.ts
dholms: yeah i think just `img.aspectRatio` would be fine here

dholms:
Sorry for letting this sit so long! I'm +1 on the ideas as well.

Two quick things: 
- Mind merging in main, rebuilding `lex-cli` and rerunning codegen? we've made some tweaks to codegen & I just want to ensure we don't get any strange artifacts
- Can you port the logic from the pds views service to the bsky views service of the same name
  - the code in both should look basically exactly the same, but we're actively deprecating the pds appview logic

devinivy:
My only concern is that there's no guarantee aspect ratio of the original is preserved by resized images, and in particular we don't always preserve it.  E.g. profile banner images and avatars are served resized w/ rules akin to the to `background-size: cover` property, which doesn't preserve the aspect ratio.

mozzius:
I think since this only applies to embed images, images used for other purposes (profiles etc) don't need to worry about this. Just like how pfps don't have alt text

Do you foresee resizing potentially happening to embed images too?

dholms:
yea agreed - I think it's important that this is an attribute on posts not images at large

DavidBuchanan314:
> Sorry for letting this sit so long! I'm +1 on the ideas as well.
> 
> Two quick things:
> 
>     * Mind merging in main, rebuilding `lex-cli` and rerunning codegen? we've made some tweaks to codegen & I just want to ensure we don't get any strange artifacts
> 
>     * Can you port the logic from the pds views service to the bsky views service of the same name
>       
>       * the code in both should look basically exactly the same, but we're actively deprecating the pds appview logic

all done, I think

dholms (approved):
looks good 👍 

i'll get this merged & rockin tmrw 👀

PR #1308, opened 2023-07-10
Refactor appview repo subscription for memleak

dholms:
Possible fix for memory leak in repo subscription. We move error handling to to the handler so that it doesn't use a closure in the for loop

### DIFF
--- a/packages/bsky/src/subscription/repo.ts
+++ b/packages/bsky/src/subscription/repo.ts
@@ -19,7 +19,12 @@ import AppContext from '../context'
 import { Leader } from '../db/leader'
 import { IndexingService } from '../services/indexing'
 import { subLogger } from '../logger'
-import { ConsecutiveList, LatestQueue, PartitionedQueue } from './util'
+import {
+  ConsecutiveItem,
+  ConsecutiveList,
+  LatestQueue,
+  PartitionedQueue,
+} from './util'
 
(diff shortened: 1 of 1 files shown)


packages/bsky/src/subscription/repo.ts
```diff
@@ -123,18 +98,45 @@ export class RepoSubscription {
...
-      await this.handleUpdateHandle(msg)
-    } else if (message.isTombstone(msg)) {
-      await this.handleTombstone(msg)
-    } else if (message.isMigrate(msg)) {
-      // Ignore migrations
-    } else {
-      const exhaustiveCheck: never = msg
-      throw new Error(`Unhandled message type: ${exhaustiveCheck['$type']}`)
+  private async handleMessage(seq: number, msg: ProcessableMessage) {
+    const item = this.consecutive.push(seq)
```
devinivy, line 102: One material change in this PR is to track the consecutive item during `handleMessage()` (i.e. _during_ queue processing) rather than ahead of the item on the queue being processed.  This might actually be something to revisit, since items can begin processing out-of-order, which I believe could cause reordering of the consecutive items.

devinivy (commented):
I dig 👍! One tweak might be needed—pretty sure we want to hoist the `this.consecutive.push(...)` back out into the loop to ensure the consecutive items are always added in order.

devinivy (approved):

PR #1312, opened 2023-07-10
:sparkles: Added new procedure for sending admin email

foysalit:


packages/pds/src/api/com/atproto/admin/sendEmail.ts
```diff
@@ -0,0 +1,46 @@
...
+        return {
+          encoding: 'application/json',
+          body: { sent: true },
+        }
+      } catch (err) {
+        return {
+          encoding: 'application/json',
+          // TODO: add more details about the error via message?
+          body: { sent: false },
+        }
```
devinivy, line 42: Is there a special reason why we want to use a 200/success response here rather than an error response?  For example, the error thrown above `throw new InvalidRequestError('Recipient not found')` will end-up being turned into a 200 rather than remaining 400 with some error details.

packages/pds/src/api/com/atproto/admin/sendEmail.ts
```diff
@@ -0,0 +1,46 @@
...
+          throw new InvalidRequestError('Recipient not found')
+        }
+
+        await ctx.moderationMailer.send(
+          { content },
+          { subject, to: userInfo.email },
+        )
+        return {
+          encoding: 'application/json',
+          body: { sent: true },
```
devinivy, line 35: I noticed in lexicon the response also has a `message` key.  Will that be used here in any situations?

packages/pds/src/index.ts
```diff
@@ -111,12 +112,28 @@ export class PDS {
...
+    const moderationMailTransport =
+      config.moderationEmailUser !== undefined &&
+      config.moderationEmailPassword !== undefined
+        ? createTransport({
+            service: 'gmail',
+            auth: {
+              user: config.moderationEmailUser,
+              pass: config.moderationEmailPassword,
+            },
+          })
```
devinivy, line 129: If we don't need to use the special gmail transport here, could we configure it with an SMTP url?  This would be convenient as it's pretty flexible and consistent with the other mail transport.

packages/pds/src/mailer/index.ts
```diff
@@ -51,6 +51,19 @@ export class ServerMailer {
     })
   }
 
+  // TODO: Implement gmail sending here?
```
devinivy, line 54: I think this TODO may be resolved :)

devinivy: Or maybe the method is no longer used?

packages/pds/src/mailer/moderation.ts
```diff
@@ -0,0 +1,36 @@
...
+    this.transporter = transporter
+  }
+
+  async send({ content }: { content: string }, mailOpts: Mail.Options) {
+    const res = await this.transporter.sendMail({
+      ...mailOpts,
+      text: content,
+      from:
+        mailOpts.from ??
+        (this.config.moderationEmailUser || this.config.emailNoReplyAddress),
```
devinivy, line 25: Is `moderationEmailUser` an email address, or just a username?

devinivy: (If it's just a username, I'm not sure if it can be used here.)

packages/pds/src/mailer/moderation.ts
```diff
@@ -0,0 +1,36 @@
...
+    })
+
+    if (!this.config.moderationEmailUser || !this.config.moderationEmailUser) {
```
devinivy, line 28: This looks like it has the same condition twice, maybe once needs to be replaced or removed.

packages/pds/src/mailer/index.ts
foysalit: yep no longer used, leftover from early attempt at implementation.

packages/pds/src/mailer/moderation.ts
foysalit: It's an email address, that makes me think the name should probably reflect that. Would `moderationEmailAddress` have been clearer?

packages/pds/src/api/com/atproto/admin/sendEmail.ts
foysalit: ahh good catch! I originally wrapped it this way to capture any potential error from the actual send email request and return it to the client with the `message` property but you're right, 200 status code wouldn't be right in that case either. I'll get rid of `message` prop from the response as well.

packages/pds/src/mailer/moderation.ts
foysalit: good eye! thanks!

packages/pds/src/api/com/atproto/admin/sendEmail.ts
foysalit: see this please https://github.com/bluesky-social/atproto/pull/1312#discussion_r1264212549

packages/pds/src/index.ts
foysalit: reading the docs it looks like this is the only way to use app specific password and otherwise, we would have to turn on something called `less secure apps` on the google account which isn't recommended. but I could have missed something so lmk if you know a way to use the stmp transport.

foysalit:
Thanks for the detailed review @devinivy sorry there were some very basic mistakes and thanks for catching those!

packages/pds/src/index.ts
devinivy: The best information I've been able to find about it is [here](https://wpmailsmtp.com/gmail-less-secure-apps/), perhaps you have a better resource (I really struggled to find official google documentation!).

My understanding is that now you _must_ turn "less secure apps" off, and _must_ use app passwords to send SMTP email.  Luckily this is what we want to do anyway, so I think that the SMTP option should work based on that information.  Let me know what info you're seeing and what you think!

foysalit: yeah that adds up with what I've found but may be I'm misunderstanding what you're asking here? I've swapped out the `service: 'gmail'` prop with smtp host config in this commit https://github.com/bluesky-social/atproto/pull/1312/commits/0ca129669a5641153a16ca55d5f6a51e60bc8447 LMK if this isn't what you wanted. honestly, I'm not sure what's the difference between the 2 configs because I'd assume nodemailer just use the smtp config behind the scene when service is set to 'gmail'?

packages/pds/src/mailer/moderation.ts
devinivy: Seem alright either way to me, just wanted to ensure it was an email address being used here 👍

packages/pds/src/index.ts
devinivy: Ah I'm sorry for the confusion!  The idea is just to make the implementation not tied to gmail if possible.  Then both the moderation and non-moderation mailers would have the same kind of configuration: each just needs a single SMTP url.  When we configure the PDS we'll include `smtp.gmail.com` in the SMTP url ourselves, rather than it being part of the PDS's implementation.

If we go this direction my recommendation would be to replace `moderationEmailUser` and `moderationEmailPassword` config with a single `moderationEmailSmtpUrl`.  Does that seem alright?

foysalit: ahhh I got it. thanks for explaining. made some adjustments to use a singular smtp url lmk if that works?

devinivy: Sounds good!  If you made the change I'm not seeing it yet, may just need a `push`!

foysalit: 🤦🏽‍♂️

foysalit: sorry, I consistently forget to push to my forked repo haha

devinivy (approved):
Awesome!

PR #1314, opened 2023-07-13
add slurs to reserved words

jesslark:
**Task:**
Added slurs to the reserved words file. 

**Purpose:**
Prevent users from using slurs as handles.

**Challenges:**
It is not a perfect solution but to prioritize the safety of marginalized users it is a good first step. This list will only prevent these exact words from being used.

**Next Steps:**
- Needs additional work to add any other words that should be caught.
- Needs regex added to catch common variations and attempts to circumvent. 
- Does not address the issue of using slurs in posts, only handles. A separate ticket and PR, potentially an entire project/epic, is required to censor slurs while allowing appropriate expression. 

**Recommendation:**
Merge as-is and follow up with additional functionality.


Bossett:
@jesslark can you have a look at https://bsky.app/profile/cwonus.org/post/3k2er3cftha2c? I think worth removing a couple more and leaving as worst-of-the-worst?

jesslark:
> @jesslark can you have a look at https://bsky.app/profile/cwonus.org/post/3k2er3cftha2c? I think worth removing a couple more and leaving as worst-of-the-worst?

Thank you for the link! I took a look at the list I have here and I think it's pretty much worst of the worst. Please point out if there are any I have left on that I shouldn't. There is another PR with a lot more words, this one is pretty short.

dholms (approved):
Thanks for putting this together. Will merge & deploy once CI runs and follow up with more tightening

jesslark:
> Thanks for putting this together. Will merge & deploy once CI runs and follow up with more tightening

Thanks so much!

packages/identifier/src/reserved.ts
jessamynwest, line 1: caught?

mattfg:
I'm in favor of this as a put-out-the-fire measure but as we have already seen from the discussion around *which* words to include, the bigger problem is much less simple. Long-term, I'm not sure this logic belongs at the protocol level, and should instead be handled either by the client, or by the server instance (so, it could be a rule at english-speaking bsky.social that "retard" is banned but on a french-speaking server it's allowed). Once it's figured out how to implement it at the instance level, this code should probably be removed.

Sominemo:
Dictionary filtering is a very basic but also very flawed technique:
- It has a lot of false positives: filtered words inside of legitimate words; words from other languages; junctions between two words, — and combinations of these. Even English has a word which contains the hard R inside itself. Trying to consider all the other languages and the junctions, and it becomes unmanageable.
- It's easy to bypass: no matter how sophisticated your anti-bypass filters are, humans always find ways to put something offensive in all sorts of places like usernames

This way of things is "good enough" for the current state of the project when moderation is not in priority and is unaffordable, but I think it shouldn't be cemented into the foundation of the software and expected to be improved.

In my opinion, it's better to consider this solution to be temporary instead of thinking how to improve it.

For the future in far perspective, I'd suggest relying on human moderation and heuristics instead. For example, a warning system for moderators when an account with a handle where a dictionary word is detected signs up or a few reports are sent about such handle.

PR #1315, opened 2023-07-13
Added badWords list

Bossett:
Pulled list from https://github.com/dariusk/wordfilter/blob/master/lib/badwords.json and added new badWords list for general bans, slurs, etc.

Obviously not the long-term solution - but hopefully something that can be quickly checked and implemented.

May be worth pluralising terms, extending matching, doing cleverness to match Unicode, etc. etc.; but this is the quick-and-dirty. 

Source is https://github.com/dariusk/wordfilter/blob/master/lib/badwords.json from https://github.com/dariusk/wordfilter/


Bossett:
Obviously not the long-term solution - but hopefully something that can be quickly checked and implemented.

May be worth pluralising terms, extending matching, doing cleverness to match Unicode, etc. etc.; but this is the quick-and-dirty. 

Source is https://github.com/dariusk/wordfilter/blob/master/lib/badwords.json from https://github.com/dariusk/wordfilter/

azigler:
Would close #1317

jesslark:
> Would close #1317

I specifically removed some words. As a fat person I see no reason to block those handles. Some I removed might be something a SW uses. I'm fine with whatever PR gets merged to address, at the very least, the most common racial slurs. We can work the rest out as we move forward.

Bossett:
Closed in favour of @jesslark's [PR 1314](https://github.com/bluesky-social/atproto/pull/1314) which already did some curation.

PR #1316, opened 2023-07-13
Augment reservedSubdomains with English Banwords

jcsalterego:
Add English banwords to `reservedSubdomains`, sourced from: https://github.com/dariusk/wordfilter/blob/master/lib/badwords.json

It's stored in ROT13'd form and ROT13'd upon inclusion in the `reservedSubdomains` constant.


azigler:
Would close #1317

jcsalterego:
@azigler For sure, this is a short-term fix at the level of "hardcoding values in a constant" and a more sophisticated and long-term solution likely involves the database and much more sophisticated detection.

Bossett:
I think this has the same 'too many words' issue as mine had, and is addressed in https://github.com/bluesky-social/atproto/pull/1314 - and may be worth also adding a pluralise step in the ROT13 to just tack 's' on the end?

jcsalterego:
@Bossett yeah, I'll close this one as there seems to be growing consensus for #1314. It's a slippery slope to start curating this list though, so I'll bow out here.

PR #1318, opened 2023-07-13
add slurs to reserved words

dholms:
Just a tweak on https://github.com/bluesky-social/atproto/pull/1314 so that we pass CI linting


packages/identifier/src/reserved.ts
```diff
@@ -1047,10 +1047,61 @@ const famousAccounts = [
   'zerohora',
 ]
 
+// Needs additional work to add more words that should be cought and regex added to catch common variations
```
jasonflorentino, line 1050: nice. glad this is open. just a typo in the comment. should be "caught" not "cought"

Separately, first time in here and im on my phone, TODOs okay in comment like this or should we open a separate issue to cover this?

jasonflorentino: just saw the previous PR. probs not worth a change right now. disregard! ty!

dholms: yeah TODOs in comments are totally fine

gay-frogs: hi, trying to figure out how this all works & hope it's okay to leave this here.  i've appended the list with an additional set of 25-30 terms and variations, each of which i believe falls under the category of "absolutely bad and is never used in a reclaimed context."  viewable here: https://gist.github.com/gay-frogs/aaa6ead819b62a3538d95b1ded6e571a

ariehkovler: >  additional set of 25-30 terms and variations, each of which i believe falls under the category of "absolutely bad and is never used in a reclaimed context." viewable here: https://gist.github.com/gay-frogs/aaa6ead819b62a3538d95b1ded6e571a

The term Yid is literally just the Yiddish word for Jew. It is used as a slur by non-Jews but it's routinely used by Jews, both Yiddish and English speakers, in its original context.

I believe there may be similar concerns with some of the other entries.

holyspiritomb: > >  additional set of 25-30 terms and variations, each of which i believe falls under the category of "absolutely bad and is never used in a reclaimed context." viewable here: https://gist.github.com/gay-frogs/aaa6ead819b62a3538d95b1ded6e571a
> 
> I believe there may be similar concerns with some of the other entries. 

"Goyim" jumps out at me in that list because it's legitimately used neutrally by Jews to mean "non-Jews," especially when talking about antisemitism. It's used in Yiddish and in modern Hebrew with the same meaning.

gay-frogs: thanks for these.  i was not fully aware of those definitions and appreciate you all for the corrections

PR #1320, opened 2023-07-13
adding reserved handle validation to domains outside of supported user domains

SlickDomique:
I was able to set myself handle taylorswift.domi.zip which contains a word from reserved handles list. I noticed that com.atproto.identity.updateHandle while failed for unsupported domain error skipped other validation steps and gone directly to checking dns records. This allowed adding slurs or reserved names to your custom handle.

### DIFF
--- a/packages/identifier/src/reserved.ts
+++ b/packages/identifier/src/reserved.ts
@@ -1135,6 +1135,14 @@ const slurs = [
 export const reservedSubdomains: Record<string, boolean> = [
   ...atpSpecific,
   ...commonlyReserved,
+].reduce((acc, cur) => {
+  return {
+    ...acc,
+    [cur]: true,
+  }
+}, {})
+
+export const dangerousUnallowedSubdomains: Record<string, boolean> = [
   ...famousAccounts,
   ...slurs,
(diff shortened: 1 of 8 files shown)


SlickDomique:
after latest change the checker checks if slurs are contained in domain minus dots but leaves checking for celeb names in direct matches of url parts

packages/identifier/src/reserved.ts
```diff
@@ -1128,18 +1131,37 @@ const slurs = [
...
+
+  return (
+    !!unallowedSubdomains.lowCheck.find((unallowedString) =>
+      handleParts.includes(unallowedString),
+    ) ||
+    !!unallowedSubdomains.highCheck.find((unallowedString) =>
+      normalizedInput.toLowerCase().includes(unallowedString),
+    )
+  )
+}
```
HarryGogonis, line 1167: Minor feedback: This part could maybe be broken up to be clearer. It seems like this function is checking 2 different things?
* normalized strings that contain slurs
* subdomain parts w/ reserved names

dholms:
Hey thanks for the PR - we handled this case in https://github.com/bluesky-social/atproto/pull/1336

PR #1323, opened 2023-07-13
feat: add regex support for banned words

cparedes:
I'm a little rusty at TS/JS, but hoping this helps solve some issues with having usernames that might have bad words embedded within them.


packages/identifier/src/handle.ts
```diff
@@ -139,15 +141,23 @@ export const ensureHandleServiceConstraints = (
   if (reserved[front]) {
     throw new ReservedHandleError('Reserved handle')
   }
+  
+  for (const bannedWord of banned) {
+    const regex = new RegExp(`.*${bannedWord}.*`)
```
HarryGogonis, line 146: does the regex need to be case insensitive?
```suggestion
    const regex = new RegExp(`.*${bannedWord}.*`, "i")
```

cparedes: Looks like it’s handled here: https://github.com/bluesky-social/atproto/blob/0bac4e775242b1916ed4216ab113b9553facf8fa/packages/identifier/src/handle.ts#L90

packages/identifier/src/banned.ts
```diff
@@ -0,0 +1,98 @@
...
+
+// naive, incomplete list of slurs and unacceptable words
+const slurs = [
+  'bluegum',
+  'chink',
+  'chinks',
+  'coolie',
+  'coolies',
+  'coon',
+  'coons',
```
wolfteeth, line 14: With the change of dealing with this via regex, the word "raccoon" will be blocked as well. I don't disagree with the basic premise of the PR, but there should be a way to allow innocuous words that otherwise might get caught in the slur filter.

cparedes: Agreed - though, seems the bluesky team’s got their own solution that looks like it’ll solve that issue https://github.com/bluesky-social/atproto/pull/1336

dholms:
Hey thanks for the PR! We got regex checks & more sophisticated checks for l33tspeak in here: https://github.com/bluesky-social/atproto/pull/1336

PR #1324, opened 2023-07-13
Improve check for slurs

arriven:
Check for inclusion, not exact matching

NOTE: only addresses `fucks.bsky.social`, not `fuck.custom.domain`


arriven:
A better option would be to go with https://github.com/bluesky-social/atproto/pull/1323 and https://github.com/bluesky-social/atproto/pull/1320 but leaving this open until at least one of those is merged

dholms:
Thanks for the PR. We added inclusion of unacceptable words here: https://github.com/bluesky-social/atproto/pull/1336

PR #1326, opened 2023-07-13
Update reserved.ts to include hexpeak and permutations of slurs

simonblack:
- Added dynamic slur generation to ensure no slurs are missed/bypassed using hexspeak
- removed trailing space after 'nazi'
- reintroduced pajeet, added jeet, its a bastardization of the traditional name Paaji; jeet/pajeet is used almost exclusively to exhibit racist undertones
- reintroduced 'retard' for what should be clear reasons

apologies in advance if this breaks your code style, implementation speed was more important!

cheers

### DIFF
--- a/packages/identifier/src/reserved.ts
+++ b/packages/identifier/src/reserved.ts
@@ -1047,53 +1047,42 @@ const famousAccounts = [
   'zerohora',
 ]
 
+
 // naive, incomplete list of slurs and unacceptable words
-const slurs = [
+const slurs_template = [
   'bluegum',
   'chink',
-  'chinks',
   'coolie',
-  'coolies',
   'coon',
-  'coons',
   'coont',
   'golliwog',
-  'golliwogs',
   'gook',
-  'gooks',
   'gyp',
-  'gyps',
   'half-breed',
   'halfbreed',
-  'half-breeds',
-  'halfbreeds',
   'heeb',
-  'heebs',
   'hitler',
+  'jeet',
   'kaffer',
-  'kaffers',
   'kaffir',
-  'kaffirs',
   'kaffre',
-  'kaffres',
   'kafir',
-  'kafirs',
   'kike',
-  'kikes',
   'kkk',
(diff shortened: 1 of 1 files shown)


ixtli:
this is a good pr because it is a template for others to use to add more checks for obfuscated slurs. 

one thing to note, though, its been shown elsewhere that given the size and complexity of unicode rendering its not really possible to detect all permutations of characters. ultimately there may be a need to render into a buffer and do some computer vision on it to see if the form resembles the form of a slur :(

packages/identifier/src/reserved.ts
```diff
@@ -1047,91 +1047,107 @@ const famousAccounts = [
...
+                hex_chars.push([char]);
+            }
+        }
+
+        let permutations: string[][] = cartesian(hex_chars);
+        for (let item of permutations) {
+            let hex_word = item.join('');
+            hex_words.push(hex_word);
+            hex_words.push(hex_word + 's');
+            hex_words.push(hex_word + '5');
```
ixtli, line 1130: to whomever reviews this: this is likely the most important part that really should have been there before.

simonblack:
> this is a good pr because it is a template for others to use to add more checks for obfuscated slurs.
> 
> one thing to note, though, its been shown elsewhere that given the size and complexity of unicode rendering its not really possible to detect all permutations of characters. ultimately there may be a need to render into a buffer and do some computer vision on it to see if the form resembles the form of a slur :(

totally! my thoughts behind this implementation was to quickly patch it the way they did but to make it far more robust that what they did initially. I am certain that if this is adopted there will have to be a better design decision made when integrating though, imo, anything involving machine learning would be something to consider for the longterm, right now we just need a patch for people to feel as safe as they did a week ago (hopefully a bit better) while a far more robust solution is implemented

dholms:
Hey thanks for the PR! We added support for hexpeak & permutations of slurs here: https://github.com/bluesky-social/atproto/pull/1336

PR #1330, opened 2023-07-13
Allow moderators to take and reverse actor takedowns

devinivy:
Just tweaking some of the role-based auth so that moderators can take and reverse actor takedowns.


dholms (approved):
yup yup 👍

PR #1332, opened 2023-07-13
Partially revert removal of some contentious and community discussed terms for slur list

sapphonie:
Bandaid, 0 effort fix that readds a few words back to the "banned handles" list.

"jap" / "japs" could be removed, if this is a fuzzy search, which I think it is, in which case, also 0 effort.

Thanks


sapphonie:
Superseded by #1333

PR #1333, opened 2023-07-13
Update reserved.ts to add back in some reserved words and fix a space

jesslark:
**Task:**
Added and updated slurs in the reserved words file.

**Purpose:**
Prevent users from using slurs as handles. The re*t*rd slur and related, and the j*p slur were removed. I believe in both of these cases these words should have been retained in the reserved words list.

**Challenges:**
This list will only prevent these exact words from being used.

**Next Steps:**
- Bluesky should create a process for modifying this list to avoid removing relevant words
- Next steps in my original PR are still relevant: https://github.com/bluesky-social/atproto/pull/1314


packages/identifier/src/reserved.ts
```diff
@@ -1070,6 +1070,8 @@ const slurs = [
   'heeb',
   'heebs',
```
gay-frogs, line 1071: I'm not an expert, but from quick research we should be probably be censoring "hebe" and not "heeb," since the latter seems to be part of a reclamation effort?

packages/identifier/src/reserved.ts
```diff
@@ -1083,9 +1085,9 @@ const slurs = [
   'kkk',
   'klukluxklan',
```
gay-frogs, line 1086: also "klu" seems to be a typo here

dholms:
Hey thanks again for the PRs! We added in a more sophisticated flagging system here: https://github.com/bluesky-social/atproto/pull/1336

jesslark:
> Hey thanks again for the PRs! We added in a more sophisticated flagging system here: #1336

Thank you! Awesome!

PR #1335, opened 2023-07-13
Switch backup handle nameservers to backup xrpc resolver

devinivy:
Gives a little more control to sidestep backup DNS nameservers that are resolving handshake domains.


dholms (approved):
💎

devinivy:
This got stale—we should just take a separate pass at this if it's still needed.

PR #1336, opened 2023-07-13
Handle validation improvements

dholms:
- move handle service constraints to PDS codebase
- do explicit slur check on handle to prevent creation
- do a fuzzier check on handles for handles that may require additional context or human intervention. On flag, we create a report from the PDS
- move fuzzy check list to config to prevent bad actor exploits and allow for fast adhoc changes


packages/pds/src/handle/index.ts
```diff
@@ -0,0 +1,95 @@
...
+
+export const ensureHandleServiceConstraints = (
+  handle: string,
+  availableUserDomains: string[],
+  allowReserved = false,
+): void => {
+  const supportedDomain =
+    availableUserDomains.find((domain) => handle.endsWith(domain)) ?? ''
+  const front = handle.slice(0, handle.length - supportedDomain.length)
+  if (front.indexOf('.') > -1) {
```
devinivy, line 80: Pretty inconsequential but `includes()` can be nice for this!

devinivy (approved):

WilliamRoyNelson:
Glad to see you're all working on stuff! This is a really tough challenge, but it's worth doing! Thanks!

packages/pds/src/config.ts
```diff
@@ -50,6 +50,8 @@ export interface ServerConfigValues {
   hiveApiKey?: string
   labelerDid: string
   labelerKeywords: Record<string, string>
+  unacceptableHandleWordsB64?: string
```
IAmJSD, line 53: nitpick: Loading this file would be better for transparency. Keep the env var though, the file should be a fallback used in prod!

packages/pds/src/config.ts
```diff
@@ -50,6 +50,8 @@ export interface ServerConfigValues {
...
+  falsePositiveHandleWordsB64?: string
```
IAmJSD, line 54: nitpick: Loading this file would be better for transparency. Keep the env var though, the file should be a fallback used in prod!

packages/pds/src/handle/moderation/index.ts
```diff
@@ -0,0 +1,66 @@
+import * as ui8 from 'uint8arrays'
+import AppContext from '../../context'
+import { REASONOTHER } from '../../lexicon/types/com/atproto/moderation/defs'
+import { UnacceptableHandleValidator } from './validator'
+
+// regexes taken from: https://github.com/Blank-Cheque/Slurs
+/* eslint-disable no-misleading-character-class */
+const explicitSlurRegexes = [
```
IAmJSD, line 8: Feels like this should be moved to its own set of files

IAmJSD (commented):
Mostly LGTM!

packages/pds/src/config.ts
conancat: yeah I think it'll be great if we know what are the words that are considered as unacceptable words here for transparency

nobember11:
So, does this mean that you guys have hidden the slur word list in the environmental variables without taking any action or providing any explanation about [this issue](https://github.com/bluesky-social/atproto/pull/1319#issuecomment-1634390882)? Please correct me if my understanding is wrong. Or could it be that those derogatory terms for Asians has been added back to the list?

IAmJSD:
@nobember11 I really doubt that's going to happen. People will be watching for this now, so the thousand eyes should stop that. I hope eventually the word list is open for transparency, but I don't believe there is malicious intent here.

packages/pds/src/handle/moderation/index.ts
samwightt, line 8: There is no license in the repo you stole this code from, meaning this code is copyrighted and you do not have permission to use it. I would expect devs working on an open source project not to do something as basic as literally ripping code from copyrighted material but oh well.

IAmJSD: > I'm no lawyer, but I'd kindly ask that you leave the comment included for credit.
This is a license. The implied agreement is to use it, it'd be nice if you left a comment.

samwightt: > This is a license. The implied agreement is to use it, it'd be nice if you left a comment.

That is not a license. I'd encourage you to read some of the licenses here: https://opensource.org/licenses/

Furthermore *if* that was even a license, it would only be for use for Reddit Automod use, as is made clear in the README.

IAmJSD: Not every license has to be in a license file. It's not a formal license sure, but a license is just a way of handing over rights when you are initially given none. The readme which is the only place rights are specified does not say it is for Reddit only, and only says that "but I'd kindly ask that you leave the comment included for credit". This implies you can use it due to the lack of restrictions and that one condition being specified in what is effectively a block of the readme that is a pseudo-license.

CatDevz: > > This is a license. The implied agreement is to use it, it'd be nice if you left a comment.
> 
> That is not a license. I'd encourage you to read some of the licenses here: https://opensource.org/licenses/
> 
> Furthermore _if_ that was even a license, it would only be for use for Reddit Automod use, as is made clear in the README.

The license is clearly stated in the README being that you can use it if you give credit. Licenses don't have to be in a specific file nor do they have to be super formal. You're just unnecessarily trying to cause issues.

packages/pds/src/config.ts
ligerzero459: Not just for transparency sake, but for future instances as well. A list contained solely in an env var means no other created instances get access

packages/pds/src/handle/moderation/index.ts
ZASMan: There's only regexes available for a handful of slurs though, which means this might actually regress some of the previously banned words?

Wouldn't it make sense to import a library that has a more robust list, and then an additional constant or database call to pull out Bluesky custom implementations?

PR #1339, opened 2023-07-14
Add ethnic slur for ukrainians to explicitSlurRegexes

imposeren:
New attempt to add ethnic slur for Ukrainians, after https://github.com/bluesky-social/atproto/pull/1321 was rejected (just because code for slurs changed dramatically)

Wiki as a proof: https://en.wikipedia.org/wiki/Oseledets#Ethnic_slur


dholms:
Thank you for the PR.

Most slurs we do not handle in the explicit slur check and instead use a fuzzier more exhaustive check that will report such that human moderators can take context into account before applying the action.

We've added this word, both singular & plural to the fuzzier check.

devinivy:
@arturovm the fuzzy checking logic can be found here: https://github.com/bluesky-social/atproto/blob/main/packages/pds/src/content-reporter/validator.ts

arturovm:
@devinivy @dholms Thanks! I can see that the list of banned words is loaded from an env variable. I'm assuming that this comes from an env file or something of the sort. Is this file (or anything else in which these words live) checked into this repo? I searched for the env var names but could find nothing other than the code that reads the config.

PR #1350, opened 2023-07-17
:sparkles: Disable signing up with invite code from takendown account

foysalit:


devinivy (approved):
Looks good!  @dholms mind taking a quick final peek?  As simple as the change is, also a little sensitive.  A couple considerations on my mind:
 - When `forUser` is set to `'admin'` (i.e. not in `repo_root` table).
 - When `for update` is added to the query, any locking concerns?  I don't think `repo_root` would get locked in the subquery, but that's the kind of thing that was crossing my mind.
 - General approach of keeping invites enabled and just disallowing their usage based on the takedown.

Again, all these seemed alright to me, just hoping for another set of eyes.

dholms (approved):
> When forUser is set to 'admin' (i.e. not in repo_root table).

yup that seems fine to me - this logic shouldn't affect "admin" codes

> When for update is added to the query, any locking concerns? I don't think repo_root would get locked in the subquery, but that's the kind of thing that was crossing my mind.

I'm like >99% sure that this won't cause any locking concerns. 

> General approach of keeping invites enabled and just disallowing their usage based on the takedown.
There is something nice about disabling invites when an account is taken down instead of doing this check. But I see two problems with that:
- flagging them for "disabled invites" only affects _future_ invites that are created (not ones that currently exist)
- if we disable their current codes, then presumably we want to be able to re-enable them due to bad take downs & that bookkeeping starts getting pretty complex

All that to say: I think the current approach makes sense 👍 

it _may_ be worth toggling `user_account.invitesDisabled` to `true` on account takedown tho 🤔

PR #1353, opened 2023-07-17
Check rkey contents just for non-tids

devinivy:


packages/pds/src/content-reporter/index.ts
```diff
@@ -42,12 +42,13 @@ export class ContentReporter {
 
   checkRecord(opts: { record: RepoRecord; uri: AtUri; cid: CID }) {
     const { record, uri, cid } = opts
-    let content = uri.rkey
+    let content = ''
     if (isProfile(record)) {
```
bnewbold, line 46: we *could* add profile description here, though that could create too much noise. I think this is a good place to start and see how this goes, don't want to swap the mod queue.

bnewbold (approved):
looks like this fixes the urgent issue (triggering on many random TIDs).

A fancy longer-term solution might include checking the record Lexicon to see if the rkey type is `any`; or maybe having a distinct check for slurs in rkey (and other fields) where the slur shouldn't be possible at all (eg, not base32), because repos can inject such records manually anyways).

PR #1354, opened 2023-07-17
AtpAgent: RefreshSession once if resumeSession fails

neko-shinshi:
Attempts to refresh the session once when resuming session fails if a refreshJwt exists.


dholms:
This seems reasonable to me & it takes advantage of the internal session resumption logic. But I also know there's some subtlety to this. Any thoughts @pfrazee?

pfrazee:
I'm 99% sure we don't want to do this. If you look at the `_fetch()` handler you can see that the session-refresh logic lives there. The `getSession()` call on line 152 uses the fetch handler and so _should_ run this logic already.

PR #1357, opened 2023-07-18
Stop using post hierarchy in pds

devinivy:
Following suit with #1257 but on the pds, we stop using the post_hierarchy table to power `getPostThread`.  There will be a followup PR that actually stops indexing into the table.


dholms (approved):
✨ amazing ✨

PR #1359, opened 2023-07-18
Sequencer benching

dholms:
Add some simple benchmarks for sequencing & a couple small performance improvements:
- don't use zod in the sequencer path
- increase page size for both backfill & live tail

### DIFF
--- a/packages/pds/src/sequencer/outbox.ts
+++ b/packages/pds/src/sequencer/outbox.ts
@@ -104,19 +104,42 @@ export class Outbox {
 
   // yields only historical events
   async *getBackfill(backfillCursor: number, backfillTime?: string) {
+    const PAGE_SIZE = 200
+    const firstEvt = await this.sequencer.requestSeqRange({
+      earliestTime: backfillTime,
+      earliestSeq: this.lastSeen > -1 ? this.lastSeen : backfillCursor,
+      limit: 1,
+    })
+    for (const evt of firstEvt) {
+      yield evt
+    }
+
+    let nextPage = this.sequencer.requestSeqRange({
+      earliestSeq: this.lastSeen,
+      limit: PAGE_SIZE,
+    })
+    let queued = this.sequencer.requestSeqRange({
+      earliestSeq: this.lastSeen + PAGE_SIZE,
+      limit: PAGE_SIZE,
+    })
+
     while (true) {
-      const evts = await this.sequencer.requestSeqRange({
-        earliestTime: backfillTime,
-        earliestSeq: this.lastSeen > -1 ? this.lastSeen : backfillCursor,
-        limit: 100,
-      })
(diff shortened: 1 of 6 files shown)


packages/pds/src/sequencer/sequencer-leader.ts
```diff
@@ -111,38 +114,42 @@ export class SequencerLeader {
...
+            'id as update_id',
+            sql<number>`nextval(${sql.literal(REPO_SEQ_SEQUENCE)})`.as(
+              'update_seq',
+            ),
+          ])
+          .where('seq', 'is', null)
+          .orderBy('id', 'asc')
+          .as('update'),
+      )
+      .set({ seq: sql`update_seq::bigint` })
```
devinivy, line 132: I have a feeling we still have to make an exception for sqlite here (probably wont like the `::bigint`).

packages/pds/src/sequencer/sequencer-leader.ts
```diff
@@ -111,38 +114,42 @@ export class SequencerLeader {
...
-    return this.db.db
+  async getUnsequencedCount() {
+    const res = await this.db.db
       .selectFrom('repo_seq')
       .where('seq', 'is', null)
-      .select('id')
-      .orderBy('id', 'asc')
-      .execute()
+      .select(countAll.as('count'))
+      .limit(1)
```
devinivy, line 144: I don't think the limit is necessary, but can't hurt!

packages/pds/src/sequencer/outbox.ts
```diff
@@ -104,19 +104,44 @@ export class Outbox {
...
       const seqCursor = this.sequencer.lastSeen ?? -1
-      if (seqCursor - this.lastSeen < 100) break
+      if (seqCursor - this.lastSeen < PAGE_SIZE / 2) break
       if (evts.length < 1) break
+
+      nextPage = queued
+      queued = this.sequencer.requestSeqRange({
+        earliestSeq: this.lastSeen + PAGE_SIZE,
+        limit: PAGE_SIZE,
+      })
```
devinivy, line 144: I am currently a little fuzzy on exactly how this will pan out, i.e. what kinds of cases could arise if there's a bug.  I think the thing that makes it trickier for me is that `this.lastSeen` is core to the pagination happening here but it's set elsewhere.

For example, I see the defensive `evt.seq > this.lastSeen` check to ensure we don't put the same event out on the stream twice (e.g. when there's a gap on one of the pages and both `nextPage` and `queued` contain some of the same events), and correctness there depends on `lastSeen` being incremented elsewhere.  Another case that might happen if `lastSeen` isn't incremented would be getting stuck in a loop for the same page.

Don't think we need to block on this, but I bet if we chat it through I'll have a better sense for it.

PR #1361, opened 2023-07-19
Include takendown posts for admins (feature branch)

devinivy:
Builds off of #1301, also looking ahead to #1233 and unifying the pds a bit with the bsky appview implementation.

### DIFF
--- a/packages/pds/src/app-view/api/app/bsky/feed/getAuthorFeed.ts
+++ b/packages/pds/src/app-view/api/app/bsky/feed/getAuthorFeed.ts
@@ -30,12 +30,32 @@ export default function (server: Server, ctx: AppContext) {
       }
 
       const { actor, limit, cursor } = params
-      const { ref } = ctx.db.db.dynamic
-      const feedService = ctx.services.appView.feed(ctx.db)
 
-      let feedItemsQb = requester
-        ? await getFeedItemsQbWithAccess(ctx, { requester, actor })
-        : getFeedItemsQbBasic(ctx, { actor })
+      const { ref } = ctx.db.db.dynamic
+      const accountService = ctx.services.account(ctx.db)
+      const feedService = ctx.services.appView.feed(ctx.db)
+      const graphService = ctx.services.appView.graph(ctx.db)
+
+      let feedItemsQb = getFeedItemsQb(ctx, { actor })
+
+      // for access-based auth, enforce blocks and mutes
+      if (requester) {
+        await assertNoBlocks(ctx, { requester, actor })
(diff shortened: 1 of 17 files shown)


packages/pds/src/app-view/api/app/bsky/feed/getAuthorFeed.ts
```diff
@@ -26,19 +34,21 @@ export default function (server: Server, ctx: AppContext) {
...
-        throw new InvalidRequestError(
-          `Requester has blocked actor: ${actor}`,
-          'BlockedActor',
-        )
-      } else if (blocks.blockedBy) {
-        throw new InvalidRequestError(
-          `Requester is blocked by actor: $${actor}`,
-          'BlockedByActor',
-        )
+      if (requester) {
```
dholms, line 37: a part of me just wants to two separate functions for "user getAuthorFeed" & "admin getAuthFeed" - not necessarily two different routes, just two functions that can call based on auth 🤔 

but also think this is totally fine & not that confusing. just depends on how much of a pattern we see this being going forward

dholms: basically just that functions with a bunch of logic inside if statements are kinda hard to grok

dholms (approved):
look good 👍 

i'd just like to change around the requester parsing in getAuthorFeed a bit

packages/pds/src/app-view/api/app/bsky/feed/getAuthorFeed.ts
devinivy: Yupyup!  I took a quick pass at this to extract some logic to help with readability, and then combined the two conditional blocks for access-based auth rules into a single block.

packages/pds/src/app-view/api/app/bsky/feed/getAuthorFeed.ts
```diff
@@ -1,19 +1,27 @@
...
 import { FeedRow } from '../../../../services/feed'
-import { InvalidRequestError } from '@atproto/xrpc-server'
 
 export default function (server: Server, ctx: AppContext) {
   server.app.bsky.feed.getAuthorFeed({
-    auth: ctx.accessVerifier,
+    auth: ctx.accessOrRoleVerifier,
     handler: async ({ req, params, auth }) => {
-      const requester = auth.credentials.did
+      const requester = 'did' in auth.credentials ? auth.credentials.did : null
```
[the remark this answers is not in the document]
devinivy: Got a `type` discriminator in there 👍 feels much better.

PR #1363, opened 2023-07-19
Allow setting a note when enabling/disabling invite codes

foysalit:


packages/pds/src/db/migrations/20230719T170511044Z-invite-note.ts
```diff
@@ -0,0 +1,12 @@
+import { Kysely } from 'kysely'
+
+export async function up(db: Kysely<unknown>): Promise<void> {
+  await db.schema
+    .alterTable('user_account')
+    .addColumn('inviteNote', 'text')
```
dholms, line 6: we use `varchar` everywhere else for text columns (i believe there's no difference under the hood between varchar & text), but I'd like to keep it all `varchar` just for consistencies sake

lexicons/com/atproto/admin/enableAccountInvites.json
```diff
@@ -11,7 +11,11 @@
           "type": "object",
           "required": ["account"],
           "properties": {
-            "account": {"type": "string", "format": "did"}
+            "account": { "type": "string", "format": "did" },
+            "note": {
```
dholms, line 15: do we need a note on enable? or should this just unset the existing note?

packages/pds/src/db/tables/user-account.ts
```diff
@@ -8,6 +8,7 @@ export interface UserAccount {
   passwordResetToken: string | null
   passwordResetGrantedAt: string | null
   invitesDisabled: Generated<0 | 1>
+  inviteNote: string | null
```
dholms, line 11: this tables starting to get pretty bloated/multi-purpose, may be worth splitting it up soon

probably doesn't need to land in this PR, altho I'm a bit worried about adding in an unconstrained text column to a table that is in pretty low latency queries (like search). Probably worth putting a max length on the note field

packages/pds/src/db/migrations/20230719T170511044Z-invite-note.ts
foysalit: I noticed that but looks like for moderation reports, we're using `text` column so I figured the content for these fields would probably be similar. moving to varchar if there's no difference

lexicons/com/atproto/admin/enableAccountInvites.json
foysalit: In my mind, if someones invites were disabled, we would probably want to explain why we are re-enabling them but not too sure if that adds up with our actual use-case. @bnewbold said we're tracking all of this in an excel right now, do we also track reasons for re-enabling invite codes?

packages/pds/src/db/tables/user-account.ts
foysalit: agreed. I wasn't sure if adding it here was wise but it made sense to keep this along with `invitesDisabled` column.

foysalit:
@devinivy / @dholms how do we normally resolve migration conflicts? imagine the order of import is what determines which migration runs first?

devinivy:
@foysalit the name of the import is what determines the order: lexical sorting of the migration names will be the order in which they run.

Since the migration names use ISO timestamps which preserve the lexical ordering by time, those timestamps ultimately determine the order.  We also try to keep the imports in order just to keep things clear :)

foysalit:
Thanks @devinivy would love to get this merged, lmk if you see anything else required here

lexicons/com/atproto/admin/enableAccountInvites.json
```diff
@@ -11,7 +11,11 @@
...
+              "type": "string",
+              "description": "Additionally add a note describing why the invites were disabled"
```
devinivy, line 17:
```suggestion
              "description": "Additionally add a note describing why the invites were enabled"
```

packages/pds/src/db/migrations/index.ts
```diff
@@ -55,3 +55,4 @@ export * as _20230530T213530067Z from './20230530T213530067Z-rebase-indices'
 export * as _20230605T235529700Z from './20230605T235529700Z-outgoing-repo-seq'
 export * as _20230703T044601833Z from './20230703T044601833Z-feed-and-label-indices'
 export * as _20230718T170914772Z from './20230718T170914772Z-sequencer-leader-sequence'
+export * as _20230719T170511044Z from './20230719T170511044Z-invite-note'
```
devinivy, line 58: I see this has a conflict with `20230727T172043676Z-user-account-cursor-idx`.  Since that migration has already run, we'll need to rename the migration for `invite-code` so that it comes after `user-account-cursor-idx`.  The simplest way to do this is probably to use the `migration:create` command again to get a new file, and move the contents of the migration into that file.

PR #1365, opened 2023-07-19
add optional `ILIKE` query to `getPopularFeedGenerators`

estrattonbailey:
Adds an optional `query` query param to `getPopularFeedGenerators` in order to optionally apply an `ILIKE` operator on `displayName` and `description` to match Ansh's [current client-side search](https://github.com/bluesky-social/social-app/pull/1031/files#diff-b85a98a45a6674a667a38278b8fb39914d1c5316959c5dc5476d73c1d842ea81R89-R90).

### DIFF
--- a/packages/pds/src/app-view/api/app/bsky/unspecced.ts
+++ b/packages/pds/src/app-view/api/app/bsky/unspecced.ts
@@ -148,11 +148,13 @@ export default function (server: Server, ctx: AppContext) {
         ])
 
       if (query) {
-        inner = inner.where((qb) =>
+        // like is case-insensitive is sqlite, and ilike is not supported
+        const operator = ctx.db.dialect === 'pg' ? 'ilike' : 'like'
+        inner = inner.where(qb => (
           qb
-            .where('feed_generator.displayName', 'ilike', `%${query}%`)
-            .orWhere('feed_generator.description', 'ilike', `%${query}%`),
-        )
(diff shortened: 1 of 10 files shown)


packages/pds/src/app-view/api/app/bsky/unspecced.ts
```diff
@@ -147,6 +147,14 @@ export default function (server: Server, ctx: AppContext) {
             .as('likeCount'),
         ])
 
+      if (query) {
+        inner = inner.where((qb) =>
+          qb
+            .where('feed_generator.displayName', 'ilike', `%${query}%`)
```
dholms, line 153: heads up: when you're working the pds package, we need to support both sqlite & postgres

I believe `ilike` is postgres only so this won't work with the sqlite impl :/

Kyesely abstracts over a lot of this, but this one, you'll probably need to do an if/else on `db.dialect`



Couple thoughts:
- this route will move to the appview soon so probably isn't _that_ important that we don't support both
- this would make search fail when running the dev-env without postgres (which I think is the general workflow for the frontend team)
- should be quick to add in so i lean towards doing so

dholms (commented):
This approach looks good to me 👍 

We'll probably want these to be different endpoints at some point (listing vs search)

But it's unspecced right now & probably going to change anyway so I'm not terribly worried about that & would rather do this small change than introduce a second search route

dholms (approved):
yup perfect

LGTM ✨

PR #1366, opened 2023-07-20
Labeling & moderation updates to SDK

pfrazee:
- [x] Add canonical definitions for labels and label groups
- [x] Generate label documentation from canonical definitions
- [x] Generate typescript from canonical definitions
- [x] Implement the `Moderation` API for determining client moderation behaviors on user accounts and content
- [x] Document the new APIs
- [x] Tests
  - [x] Post moderation
  - [x] Profile moderation

Future tasks:
- Label review and finalization
- Userlist moderation
- Feedgen moderation


packages/api/README.md
```diff
@@ -160,6 +160,87 @@ console.log(rt3.length) // => 25
 console.log(rt3.graphemeLength) // => 1
 ```
 
+### Moderation
```
dholms, line 163: is there an argument for putting this in a separate package?

we use `@atproto/api` in a the pds/appview for instance and may not want to pull in the whole labeling sdk

don't think this is block but could imagine an `sdk` package that is comprised of `api` & `labels` or similar

packages/api/definitions/moderation-behaviors.d.ts
```diff
@@ -0,0 +1,48 @@
...
+  cause?: string
+  filter?: boolean
+  blur?: boolean
+  alert?: boolean
+  noOverride?: boolean
+}
+
+export interface ModerationBehaviorScenario {
+  cfg: string
+  subject: 'post' | 'profile' | 'userlist' | 'feedgen'
```
devinivy, line 13: Would it be equally valid to use collection names here, and this is just a more convenient interface?  Or would using collections end-up being constraining/incorrect in some way?

packages/api/tests/util/moderation-behavior.ts
```diff
@@ -0,0 +1,177 @@
+import { ModerationUI, ModerationOpts, ComAtprotoLabelDefs } from '../../src'
+import type {
+  ModerationBehaviors,
+  ModerationBehaviorScenario,
+  ModerationBehaviorResult,
+} from '../../definitions/moderation-behaviors'
+import { mock as m } from './index'
+
+expect.extend({
```
dholms, line 9: 🆒

packages/api/tests/util/moderation-behavior.ts
```diff
@@ -0,0 +1,177 @@
...
+      }
+      if (!!actual.noOverride !== !!expected.noOverride) {
+        return fail(
+          `${context} expected to be noOverride=${
+            expected.noOverride || false
+          }, got ${actual.noOverride || false}`,
+        )
+      }
+    }
+    return { pass: true }
```
dholms, line 78: my compiler is telling me this needs a `message: string` or `message: () => string` on it

packages/api/tests/post-moderation.test.ts
```diff
@@ -0,0 +1,46 @@
...
+describe('Post moderation behaviors', () => {
+  const scenarios = Array.from(Object.entries(suite.scenarios))
+  it.each(scenarios)(
+    '%s',
+    (_name: string, scenario: ModerationBehaviorScenario) => {
+      const res = moderatePost(
+        suiteRunner.postScenario(scenario),
+        suiteRunner.moderationOpts(scenario),
+      )
+      expect(res.content).toBeModerationResult(
```
dholms, line 28: my compiler isn't picking up on the availability of this msg. any ideas?

packages/api/src/moderation/subjects/feed-generator.ts
```diff
@@ -0,0 +1,13 @@
...
+  ModerationSubjectFeedGenerator,
+  ModerationDecision,
+  ModerationOpts,
+} from '../types'
+
+export function decideFeedGenerator(
+  subject: ModerationSubjectFeedGenerator,
+  opts: ModerationOpts,
+): ModerationDecision {
+  // TODO handle labels applied on the feed generator itself
```
devinivy, line 11: Just confirming this is intended to be left as a TODO for now!

packages/api/src/moderation/subjects/user-list.ts
```diff
@@ -0,0 +1,13 @@
...
+  ModerationSubjectUserList,
+  ModerationOpts,
+  ModerationDecision,
+} from '../types'
+
+export function decideUserList(
+  subject: ModerationSubjectUserList,
+  opts: ModerationOpts,
+): ModerationDecision {
+  // TODO handle labels applied on the list itself
```
devinivy, line 11: Just confirming this is intended to be left as a TODO for now 👍

packages/api/tests/util/moderation-behavior.ts
devinivy, line 78: Just making some types happy.
```suggestion
    return { pass: true, message: () => '' }
```

packages/api/tests/util/index.ts
```diff
@@ -0,0 +1,176 @@
...
+    author: AppBskyActorDefs.ProfileViewBasic
+    embed?: AppBskyFeedDefs.PostView['embed']
+    replyCount?: number
+    repostCount?: number
+    likeCount?: number
+    viewer?: AppBskyFeedDefs.ViewerState
+    labels?: ComAtprotoLabelDefs.Label[]
+  }): AppBskyFeedDefs.PostView {
+    return {
+      uri: `at://${author.did}/app.bsky.post/fake`,
```
devinivy, line 76: I noticed we use the real collection name `app.bsky.feed.post` above in `post()`, but not here.  Just flagging in case you think it's important for those to match for any reason.

packages/api/tests/post-moderation.test.ts
devinivy, line 28: The types for the extended `expect` matcher are not making it through here for me, so `toBeModerationResult()` appears not to exist.  I did a brief search to see if it's easy to extend the types on jest's `expect`, but nothing immediately jumped out at me.  I don't think it needs to block this work at all, but would be nice for us to eventually resolve so that the types in the test suite all validate.

packages/api/tests/profile-moderation.test.ts
```diff
@@ -0,0 +1,46 @@
...
+      expect(res.account).toBeModerationResult(
+        scenario.behaviors.account,
+        'account',
+        JSON.stringify(res, null, 2),
+      )
+      expect(res.profile).toBeModerationResult(
+        scenario.behaviors.profile,
+        'profile content',
+        JSON.stringify(res, null, 2),
+      )
```
devinivy, line 37: The type of `behaviors` here comes from `ModerationBehaviorScenario`, which doesn't have they keys `behaviors.account` or `behaviors.profile`.  It just has the following: `behaviors.content`, `behaviors.avatar`, and `behaviors.embed`.  Not 100% sure whether the tests or types need updating, but I expect there's a little something to tidy here.

devinivy: Taking a closer look, I expect it's just that the `ModerationBehaviorScenario` type needs to be updated 👍

packages/api/src/moderation/subjects/profile.ts
```diff
@@ -0,0 +1,31 @@
+import { ModerationCauseAccumulator } from '../accumulator'
+import {
+  Label,
+  ModerationSubjectProfile,
+  ModerationOpts,
+  ModerationDecision,
+} from '../types'
+
+export function decideProfile(
```
dholms, line 9: Are there times that we want to decide on the profile as something different from the account? 

I'm fine with having them split out for now but it's just one extra function to juggle & it seems like we'll want to consider the profile any time we're looking at an account (since it provides the display name & avatar).
It seem

packages/api/src/moderation/subjects/user-list.ts
dholms: yeah pretty sure we're not returning labels on lists or feed gens yet

packages/api/src/moderation/util.ts
```diff
@@ -0,0 +1,98 @@
...
+export function takeHighestPriorityDecision(
+  ...decisions: (ModerationDecision | undefined)[]
+): ModerationDecision {
+  // remove undefined decisions
+  const filtered = decisions.filter((d) => !!d) as ModerationDecision[]
+  if (filtered.length === 0) {
+    return ModerationDecision.noop()
+  }
+
+  // sort by highest priority
```
devinivy, line 17: Could be worth a note somewhere around here that "highest priority" means "least priority number".  I.e. that we want to sort lowest to highest.

packages/api/src/moderation/subjects/profile.ts
devinivy: I believe @bnewbold surfaced some fairly common scenarios where this comes-up, and that it's designed this way to support them.

dholms (approved):
helluva chunk of work

this looks solid to me - well structured, easy to understand & lots of test scenarios

I didn't follow through all the logic or test scenarios but I spot checked them & they seemed solid to me. I'm sure we'll have some slight tweaks etc over time

Just to summarize things I noticed:
- types for that jest extension aren't being friendly to me
- may want to consider a separate package for all this at some point
- I think combining profile & actor into a single "subject" may make it easier as we add new datatypes - seems that it could be easy to forget one or the other there

packages/api/src/moderation/subjects/profile.ts
dholms: ah cool cool 👍

packages/api/src/moderation/index.ts
```diff
@@ -0,0 +1,343 @@
...
+    profile: ModerationDecision
+    quote?: ModerationDecision
+    quotedAccount?: ModerationDecision
+  }
+  content: ModerationUI
+  avatar: ModerationUI
+  embed: ModerationUI
+}
+
+export function moderatePost(
```
dholms, line 131: ah just thought of something here: do we ever want to make moderation decisions based on "repostedBy"?

I believe we strip out reposts from blocked/muted accounts, but what about reposts from a labeled account?

bnewbold: that is a good question. to date we have not taken any moderation actions on reposts (and also not on like or follow records).

I think that yeah, we *should* probably consider both the re-posting *account* as well labels on the account and post record of the content itself. I'm not sure how common this is... reposts show up mostly in author feeds and following feeds (I think), so only if you follow or investigate the account. The user story would be something like, you followed somebody a long time ago, they have been labeled in a way your client would be filtering their posts, they re-post other similarly unwanted content, maybe bulk-reposting, that content ends up in your following feed.

packages/api/definitions/moderation-behaviors.d.ts
bnewbold: yeah, this seems like it would be "more correct" and future-proof, if a bit more verbose.

packages/api/docs/labels.md
```diff
@@ -0,0 +1,826 @@
...
+  ### Configurable?
+
+  Non-configurable labels cannot have their preference changed by the user.
+
+  ### Flags
+
+  Additional behaviors which a label can adopt.
+
+  - <code>no-override</code> The user cannot click through any covering of content created by the label.
+  - <code>adult</code> The user must have adult content enabled to configure the label. If adult content is not enabled, the label must adopt the strictest preference.
```
bnewbold, line 30: Not blocking, and might not be worth changing at this point, but we might want to use a different keyword here to distinguish between "account is/isn't in show-adult-content mode" from "account is an adult". It is at least conceivable that this SDK will specify different behaviors for minors vs. for adults, in a way that can't be overridden by configuration.

packages/api/src/moderation/accumulator.ts
```diff
@@ -0,0 +1,181 @@
...
+  addLabel(label: Label, opts: ModerationOpts) {
+    // look up the label definition
+    const labelDef = LABELS[label.val]
+    if (!labelDef) {
+      // ignore labels we don't understand
+      return
+    }
+
+    // look up the label preference
+    // TODO use the find() when 3P labelers support lands
```
bnewbold, line 50: I came here to say "oh wait will it be a heavy lift to add 3P labeler support" and the API signature is already in place :kissing_heart:

packages/api/README.md
pfrazee: Maybe but I'm not inclined to do it yet

packages/api/definitions/moderation-behaviors.d.ts
pfrazee: This `d.ts` is just used informally at the moment and is only used by internal code. We can expand it in the future if it becomes more official

packages/api/src/moderation/subjects/feed-generator.ts
pfrazee: yep!

packages/api/src/moderation/subjects/profile.ts
pfrazee: Yep this is actually a very specific choice that's used actively

packages/api/src/moderation/subjects/user-list.ts
pfrazee: yep

packages/api/tests/util/index.ts
pfrazee: Whoops yep, that's a mistake

packages/api/tests/util/moderation-behavior.ts
pfrazee: +1

pfrazee:
Published `@atproto/api@0.5.0`

PR #1367, opened 2023-07-20
Tweak timeline construction

dholms:
Some tweaks for timeline & popular with friends constructions:
- split into two separate queries that we UNION ALL
  - posts from follows
  - posts from self
- base follows query off of a JOIN instead of WHERE EXISTS which frees up the query planner a bit more
- increase feed date threshold from 1 -> 7
- add a idx on (creator, sortAt, cid) to post table (similar idx to the one that already exists on feed_item)

We should pair this with increasing the stats target on post, feed_item, and follow to 1000 & re-analyzing.


packages/bsky/src/api/app/bsky/util/feed.ts
```diff
@@ -12,7 +12,7 @@ export class FeedKeyset extends TimeCidKeyset<FeedRow> {
 }
 
 // For users with sparse feeds, avoid scanning more than one week for a single page
-export const getFeedDateThreshold = (from: string | undefined, days = 1) => {
+export const getFeedDateThreshold = (from: string | undefined, days = 7) => {
```
devinivy, line 15: Just double-checking that this number is intended for both timeline and with-friends.

packages/bsky/src/api/app/bsky/unspecced/getTimelineSkeleton.ts
```diff
@@ -14,27 +14,19 @@ export default function (server: Server, ctx: AppContext) {
...
-      let feedItemsQb = feedService
-        .selectFeedItemQb()
-        .where((qb) =>
-          qb
-            .where('originatorDid', '=', viewer)
-            .orWhere('originatorDid', 'in', followingIdsSubquery),
-        )
+      const followBuilder = ctx.db.db
+        .selectFrom('follow')
+        .innerJoin('feed_item', 'feed_item.originatorDid', 'follow.subjectDid')
```
devinivy, line 27: Out of curiosity, do you know if switching the base table versus the join table makes any difference to postgres?  I think of this query as "selecting feed items" though of course the queries are equivalent 👍

packages/bsky/src/api/app/bsky/unspecced/getTimelineSkeleton.ts
```diff
@@ -48,7 +40,23 @@ export default function (server: Server, ctx: AppContext) {
             ref('originatorDid'),
           ]),
         )
+        .selectAll('feed_item')
+
+      const selfBuilder = ctx.db.db
+        .selectFrom('feed_item')
+        .where('feed_item.originatorDid', '=', viewer)
+        .selectAll('feed_item')
```
devinivy, line 48: Only possible (?) snag I could see here is viewing your own reposts of posts by someone who blocked you.

packages/bsky/src/api/app/bsky/unspecced/getTimelineSkeleton.ts
```diff
@@ -48,7 +40,23 @@ export default function (server: Server, ctx: AppContext) {
...
+      let feedItemsQb = ctx.db.db
+        .selectFrom(followBuilder.unionAll(selfBuilder).as('feed_item'))
+        .innerJoin('post', 'post.uri', 'feed_item.postUri')
         .where('feed_item.sortAt', '>', getFeedDateThreshold(sortFrom))
+        .selectAll('feed_item')
+        .select([
+          'post.replyRoot',
+          'post.replyParent',
+          'post.creator as postAuthorDid',
+        ])
```
devinivy, line 59: Just double-checking that this all works fine with pagination.  I only bring it up because I recall the way kysely places parentheses (or really, _doesn't_ place parentheses) with `union all` has given me some grief in the past.  I assume the tests would catch this, though.

devinivy (approved):
Had a couple questions/notes, but this looks solid, excited to give it a whirl 👍

packages/bsky/src/api/app/bsky/unspecced/getTimelineSkeleton.ts
dholms: eh good call - i'll just add in all the mute & block checks for consistencies sake. they're pretty cheap

packages/bsky/src/api/app/bsky/unspecced/getTimelineSkeleton.ts
dholms: I don't believe it does i'll switch this back to being off of feed item 👍

packages/bsky/src/api/app/bsky/util/feed.ts
dholms: yup - optimistic that with these changes, with-friends will be fast as well

PR #1368, opened 2023-07-20
Partitioning in bsky indexer

devinivy:
This introduces partitioning in the bsky appview indexer:
 - One ingester places repo subscription messages onto partitions.
 - Multiple indexers each consume from one or more partitions.
 - Fixed number of partitions, but dynamic number of consumers.  Each partition has a single consumer.

Outstanding items:
 - [x] Tests of indexing pipeline.
 - [x] Fix existing tests that relied on in-process indexing pipeline.
 - [x] Indexer teardown is timing out.
 - [x] Backpressure on ingester.
 - [x] Service entrypoints for indexer and ingester.
 - [x] Do we want to keep around leader/locking?
 - [x] Related to above, work-out how repartitioning will work.
 - [x] General tidying, maybe wrapping some of the redis APIs to make them easier to read.
 - [x] Do we want to use consumer groups and xack, or do we like the xdel approach?
   - _using xtrim as tail of partition is fully processed_
 - [x] Taking handle revalidation out of main indexing path.
   - _not yet implemented_ 
 - [x] Docs for `with-test-redis-and-db.sh` script.
 - [x] The `with-test-redis-and-db.sh` script has been shoehorned into the `packages/pg/` directory.  Perhaps we want to move and/or rename some things?
 - [x] Observability, particularly of each partition.
 - [x] HTTP health checks?
   - _not yet implemented_ 
 - [x] Misc. TODOs documented in comments.


packages/bsky/src/ingester/subscription.ts
```diff
@@ -0,0 +1,248 @@
...
+  }
+  return { info: null }
+}
+
+function ui8ToBuffer(bytes: Uint8Array) {
+  return Buffer.from(bytes.buffer, bytes.byteOffset, bytes.byteLength)
+}
+
+async function getPartition(did: string, n: number) {
+  const partition = await randomIntFromSeed(did, n)
```
dholms, line 202: ayy this fn

may be worth a quick test on this to make sure it gives a nice uniform distribution on a low range like this - I can't see why it wouldn't but just for peace of mind

packages/bsky/src/ingester/subscription.ts
```diff
@@ -0,0 +1,248 @@
...
+  } else if (message.isTombstone(msg)) {
+    return { seq: msg.seq, repo: msg.did, message: msg }
+  } else if (message.isInfo(msg)) {
+    return { info: msg }
+  }
+  return { info: null }
+}
+
+function ui8ToBuffer(bytes: Uint8Array) {
+  return Buffer.from(bytes.buffer, bytes.byteOffset, bytes.byteLength)
```
dholms, line 198: ah nice! this could be a good function for `@atproto/common` - i think i've run into similar in the past

packages/bsky/src/indexer/services.ts
```diff
@@ -0,0 +1,25 @@
...
+  backgroundQueue: BackgroundQueue
+}): Services {
+  const { idResolver, labeler, backgroundQueue } = resources
+  return {
+    indexing: IndexingService.creator(idResolver, labeler, backgroundQueue),
+    label: LabelService.creator(),
+  }
+}
+
+export type Services = {
```
dholms, line 20: may be cool to have the actual services live in this dir as well

def not a blocking thing tho

packages/bsky/tests/views/mute-lists.test.ts
```diff
@@ -81,7 +81,7 @@ describe('bsky views with mutes from mute lists', () => {
       sc.getHeaders(alice),
     )
     await network.processAll()
-    await network.bsky.ctx.backgroundQueue.processAll()
+    await network.bsky.processAll()
```
dholms, line 84: `network.processAll` actually already handles this call

packages/bsky/src/indexer/services.ts
devinivy: That would def be ideal.  I think there are some cross-dependencies with the label service, which is also used in the api server.  If we can resolve those then I think we're good to move some services around.  I was thinking of this as an "all or nothing" kind of thing just to avoid confusion.

dholms: yeah i think it's totally fine the way it is

maybe we could split the labeling service into a creation & reading

PR #1378, opened 2023-07-24
Apply block rules for third parties on replies and embeds

### LINKED ISSUE #1374: Can still interact with an account that has blocked you via the API
**Describe the bug**

Right now, if an account has blocked you on Bluesky, you can still interact with that account's posts if you have the post URI. Tested all types of record creation (like, reply, repost, quote post) and they all posted successfully and are visible on the timeline. While the first three are still not visible to the account that did the blocking, I am worried that the quote post functioning is a viable harassment vector.

**Expected behavior**

Ideally, the block should prevent these functions from resolving. I think either the PDS should have logic that prevents these records from being created, or the AppView can prevent these records from being displayed.

devinivy:
Record embeds are marked as blocked to third party actors when there's a block between the poster and author of the embed. Replies are not surfaced to third party actors when there's a block between the poster and author of the reply.

These are applied lazily during post hydration using a lookup on `actor_block` for an arbitrary list of actor pairs.  In this case those pairs are calculated from the post contents—i.e. between post author and reply parent author, between post author and quoted post author.  If we like this approach, we could eventually move checks for block relationships with the requester/viewer into here too.

Resolves #1374


packages/pds/src/app-view/services/feed/index.ts
```diff
@@ -341,19 +353,93 @@ export class FeedService {
...
+      // skip posts that we can't process or appear to already have been processed
+      if (!isPostRecord(post.record)) continue
+      if (byPost[post.uri]) continue
+      byPost[post.uri] = {}
+      // 3p block for replier
+      const replyUri = post.record.reply?.parent.uri
+      const replyToDid = replyUri ? didFromUri(replyUri) : null
+      // 3p block for record embeds
+      const embedUris = nestedRecordUris([post.record])
+      // gather blocks
```
dholms, line 385:
```suggestion
      // gather potential blocks
```

packages/pds/src/app-view/services/feed/index.ts
```diff
@@ -341,19 +353,93 @@ export class FeedService {
...
+      blocks,
       usePostViewUnion,
     )
   }
 
-  async embedsForPosts(postInfos: PostInfoMap, requester: string, depth = 0) {
+  // applies blocks for visibility to third-parties (i.e. based on post content)
+  async blocksForPosts(posts: PostInfoMap): Promise<PostBlockState> {
+    const allPairs: BlockPair[] = []
+    const byPost: Record<string, { reply?: BlockPair; embed?: BlockPair }> = {}
```
dholms, line 373:
```suggestion
    const byPost: PostBlockState = {}
```

packages/pds/src/app-view/services/feed/index.ts
```diff
@@ -341,19 +353,93 @@ export class FeedService {
...
       embeds,
       labels,
+      blocks,
       usePostViewUnion,
     )
   }
 
-  async embedsForPosts(postInfos: PostInfoMap, requester: string, depth = 0) {
+  // applies blocks for visibility to third-parties (i.e. based on post content)
+  async blocksForPosts(posts: PostInfoMap): Promise<PostBlockState> {
```
dholms, line 371: this fn is a little bit hard to follow altho the logic all seems sound

I think two things might clear it up
- finding a new word for "potential block" other than "block" - it's a bit overloaded & I have to look for context to see if a "block" is a potential or actual
- using "parent" instead of reply. I kinda wish our lexicons reflected this as well - but the thing that is "blocking" a given post is it's "parent" not it's "reply". put another way: it's the post in question is not "blocking a reply" it's "blocked in the context of being a reply"

packages/pds/src/app-view/services/feed/index.ts
```diff
@@ -354,13 +364,89 @@ export class FeedService {
...
+    const byPost: Record<string, PostRealtionships> = {}
+    const didFromUri = (uri) => new AtUri(uri).host
+    for (const post of Object.values(posts)) {
+      // skip posts that we can't process or appear to already have been processed
+      if (!isPostRecord(post.record)) continue
+      if (byPost[post.uri]) continue
+      byPost[post.uri] = {}
+      // 3p block for replies
+      const parentUri = post.record.reply?.parent.uri
+      const parentToDid = parentUri ? didFromUri(parentUri) : null
```
devinivy, line 393: I will rename this variable on the final pass to `parentDid`.

packages/pds/src/app-view/services/feed/index.ts
```diff
@@ -519,3 +610,48 @@ const nestedRecordUris = (posts: PostRecord[]): string[] => {
   }
   return uris
 }
+
+type PostRealtionships = { reply?: RelationshipPair; embed?: RelationshipPair }
```
devinivy, line 614: Will fix this typo on final pass (should be `PostRelationships`).

dholms (approved):
looks excellent 🙏 

might just porting tests over to `bsky` as well?

dholms (approved):

PR #1382, opened 2023-07-24
Application ratelimits

dholms:
This introduces the concept of rate limits to `@atproto/xrpc-server`

One reason to include rate limits in `xrpc-server` is to start encouraging a standardized approach to rate-limiting across services.

Rate limits can be monitored by any RateLimiter that matches our interface, though we supply implementations for:
- memory
- redis

There are two types of rate limits:
- route
  - a rate limit applied to a particular route - only requests against that route count against your rate limit
  - this rate limit is defined on the route that it limits
- shared
  - a "named" rate limit that is "consumed" by many routes
  - this rate limit is defined on xrpc-server instantiation and then specified on each route that uses it
- global
  - a shared rate limit that affects every route on the server
  - this route is defined on xrpc-server instantiation and does not need to be specified on any other route

Rate limiting can also be configured with a `bypassSecret` which will allow you to bypass any rate limits.

Rate limiting is very flexible. For any given rate limit, you can provide a function to calculate the key (defaults to IP, but you we may wish to use DIDs or a composite key in some cases) as well as the 'cost' of the route (applyWrites for instance "consumes" as many points as the number of creates/updates it contains).

Rate limiting is also composable, many different limits can be added to each route. Limits will be processed in parallel and if any limit hits a threshold, the request will fail. When reporting limits from a route with multiple limits, we report the limit with the lowest remaining points.

---

🚲 Bike shed on HTTP headers 🏠 

I used the headers specified in the [new IETF draft](https://www.ietf.org/archive/id/draft-polli-ratelimit-headers-02.html) for rate limit headers.

This is still a draft but it seems very likely to get approved & I think it makes our rate limiting approach a little bit more concrete.

This includes `RateLimit-Policy` which is a location where we add in arbitrary comments. Is there any atproto-specific information that we would like to include there? One idea that I had is that we could encode the rate limit key in that field (for instance `'did'` or `'ip'`). Thoughts?

Finally, I used `X-RateLimit-Bypass` as the header name where you include the bypass secret. Any considerations on that name?


bnewbold:
This looks pretty nifty and powerful! A couple high-level notes:

- using https://www.npmjs.com/package/rate-limiter-flexible, nice
- should `RateLimitExceededError` always return HTTP 429? I assume so, but don't see that tested or wired up specifically from a quick skim
- given that we are computing it already, could we have the error JSON include the same info we are putting in the HTTP headers? kind of extra work, maybe we could just mention the HTTP header in the error message? I bet a bunch of devs will only see (and log!) the response body and not think to look in headers. this could be future work/iteration
- I didn't read through recursively to verify, but assume: hitting a limit and getting a 429 response doesn't consume additional points (?); and most errors consume limits (eg, a failed login attempt consumes limit)
- XRPC endpoints are either GET or POST, and not either, so didn't need to do per-method checks
- I can see us wanting to rate limit repo ops by collection. eg, limit creation of `app.bsky.feed.post` differently from `app.bsky.graph.follow`. for createRecord that wouldn't be hard to pull out, but harder for batch ops. can't think of a simple way to resolve that
- it could potentially be nice to have multiple limits per endpoint, though that might be too big a complexity lift. eg, for logins, would want to limit login attempts by IP and *also* limit login attempts for the same DID or handle

Mixed feelings about the bypass secret. It is simple, so probably should go ahead with it now and not get too hung up on it. But we sure have a lot of special auth mechanisms and secrets at this point. The pattern I have usually used is to have separate HTTP routes/endpoints for internal services to bypass this sort of limit, eg coming in from a different domain name or direct port connection, or even a separate instance pool with the limit disabled.

packages/pds/src/api/com/atproto/server/createSession.ts
```diff
@@ -3,57 +3,68 @@ import AppContext from '../../../../context'
...
-  server.com.atproto.server.createSession(async ({ input }) => {
-    const { password } = input.body
-    const identifier = input.body.identifier.toLowerCase()
-    const authService = ctx.services.auth(ctx.db)
-    const actorService = ctx.services.account(ctx.db)
+  server.com.atproto.server.createSession({
+    rateLimit: {
+      durationMs: 5 * MINUTE,
+      points: 10,
+      calcKey: ({ req, input }) => `${req.ip}-${input.body.identifier}`,
```
devinivy, line 13: We'll want to double-check `req.ip` is set correctly, since if it's not it could represent some piece of infra like a load balancer.  It looks like the behavior there depends on an express setting.

packages/pds/src/index.ts
```diff
@@ -242,14 +248,39 @@ export class PDS {
...
       },
-    })
+    }
+    if (config.rateLimitsEnabled) {
+      xrpcOpts['rateLimits'] = {
+        creator: (opts: RateLimiterOpts) =>
+          RateLimiter.memory({
+            bypassSecret: config.rateLimitBypassKey,
+            ...opts,
+          }),
```
devinivy, line 265: I assume we'll end-up swapping this with redis when applicable.

packages/pds/src/index.ts
```diff
@@ -242,14 +248,39 @@ export class PDS {
...
+            points: 1000,
+          },
+        ],
+        shared: [
+          {
+            name: 'repo-write',
+            durationMs: DAY,
+            points: 5000,
+          },
+        ],
```
devinivy, line 279: Just to confirm my understanding: this will amount to allowing 5000 create/update ops per day, and 1000 requests every 5mins by IP.  I take it on the consumer pds these limits are likely to be configurable as part of the deployment, but not necessary needed at the moment.

packages/pds/src/index.ts
```diff
@@ -242,14 +248,39 @@ export class PDS {
...
+        creator: (opts: RateLimiterOpts) =>
+          RateLimiter.memory({
+            bypassSecret: config.rateLimitBypassKey,
+            ...opts,
+          }),
+        global: [
+          {
+            name: 'global-ip',
+            durationMs: 5 * MINUTE,
+            points: 1000,
```
devinivy, line 270: Will `OPTIONS` requests get counted against this limit?

packages/pds/tests/_util.ts
```diff
@@ -93,6 +93,7 @@ export const runTestServer = async (
     didCacheStaleTTL: HOUR,
     jwtSecret: 'jwt-secret',
     availableUserDomains: ['.test'],
+    rateLimitsEnabled: false,
```
devinivy, line 96: Since the rate limits are in the critical path, it might be nice to get at least one simple test going with rate limiting enabled, i.e. basic gutcheck to ensure they generally allow requests through.

packages/xrpc-server/src/rate-limiter.ts
```diff
@@ -0,0 +1,158 @@
...
+  for (const resp of resps) {
+    if (resp === null) continue
+    if (lowest === null || resp.remainingPoints < lowest.remainingPoints) {
+      lowest = resp
+    }
+  }
+  return lowest
+}
+
+const defaultKey: CalcKeyFn = (ctx: XRPCReqContext) => ctx.req.ip
```
devinivy, line 157: It came-up before, but I think we may have to do something tricky to account for untrusted headers and express's handling of x-forwarded-for.

If the server knows there's one hop setting x-forwarded-for then it can take `req.ips.at(-1)`.  That should be true for us, but in general it doesn't necessarily hold.  A couple possible takeaways:
 - xrpc-server may support setting a custom default key
 - xrpc-server might expose a utility for safely determining the client ip, or bailing when it can't be determined.
 - the `calcKey()` function may support returning `null` or a specific symbol (e.g. `xrpc.kSkipRateLimit` to avoid rate-limiting when a legit key can't be determined.  That would cover our case of `req.ips.at(-1)` potentially missing.

To be honest I'm not sure there's a safe default implementation for an intermediary that appends but doesn't replace x-forwarded for 🙃 So we'll just have to choose our tradeoff and document it.  I think the ability to bail on rate limiting based on result of `calcKey` will be useful here.

devinivy: (The behavior of a missing key could also be based on whether rate limiting is configured to fail open or closed.)

packages/xrpc-server/src/rate-limiter.ts
```diff
@@ -0,0 +1,158 @@
...
+  status: RateLimiterStatus,
+) => {
+  ctx.res.setHeader('RateLimit-Limit', status.limit)
+  ctx.res.setHeader('RateLimit-Remaining', status.remainingPoints)
+  ctx.res.setHeader(
+    'RateLimit-Reset',
+    Math.floor((Date.now() + status.msBeforeNext) / 1000),
+  )
+  ctx.res.setHeader('RateLimit-Policy', `${status.limit};w=${status.duration}`)
+}
```
devinivy, line 142: 🤤 this will be nice!

devinivy (commented):
Just gotta figure-out how we want to handle determining the client ip, then I think we're good to go!

packages/pds/src/index.ts
dholms: yup that's exactly right 👍

packages/pds/src/index.ts
dholms: no they shouldn't be - the rate limit consumption occurs in xrpc-server directly before the handler

packages/pds/tests/_util.ts
dholms: yup yup - that was totally on my TODOs & then forgot that i never got it done 😛

packages/xrpc-server/src/rate-limiter.ts
estrattonbailey: A lot of HTTP client libraries use `Retry-After`. Should we set that header too in order to be backwards compatible?

From the new spec: `The behavior of RateLimit-Reset is compatible with the delta-seconds notation of Retry-After.`

PR #1388, opened 2023-07-26
app.bsky.feed.getAuthorFeeds and GraphService.getBlocksMulti

jcsalterego:
_Please consider this an extremely specific suggestion as it's scored a D+ according to the Guidelines section 😬. I fully accept the risk that the first & only reply is "you're wasting everyone's time" followed by an immediate close._

This PR introduces a new endpoint called `app.bsky.feed.getAuthorFeeds` which is similar to `getAuthorFeed` but takes in a list of actor handles/DIDs. For this effort, I've also added a `getBlocksMulti` method to the Graph service.

Refinement:

* Tests are really a copy and slight extension of what exists for `getAuthorFeed`, so it probably needs work.
* The endpoint can't actually return blocked / blockedBy errors, because it simply drops those posts.

**This would be a great boon to feed generators, which could pass in a list of users to fetch a combined timeline.**

### DIFF
--- a/lexicons/app/bsky/feed/getAuthorFeeds.json
+++ b/lexicons/app/bsky/feed/getAuthorFeeds.json
@@ -1,0 +1,40 @@
+{
+  "lexicon": 1,
+  "id": "app.bsky.feed.getAuthorFeeds",
+  "defs": {
+    "main": {
+      "type": "query",
+      "description": "A view of actors' feeds.",
+      "parameters": {
+        "type": "params",
+        "required": ["actors"],
+        "properties": {
+          "actors": {
+            "type": "array",
+            "items": {"type": "string", "format": "at-identifier"}
+          },
+          "limit": {"type": "integer", "minimum": 1, "maximum": 100, "default": 50},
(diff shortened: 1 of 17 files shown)


dholms:
Yup unfortunately, I'm gonna turn this one down.

The rationale being that this doesn't correspond to any in-app views & I believe there are better ways for a feed generator to get this info.

Generally, I would say it's an anti-pattern for a feed generator to satisfy skeleton requests by sending out subsequent RPC requests, it should ideally have the content ready when it receives the request. This route basically serves only to allow a feed generate to function as a pass-through list.

We have intentions of getting lists in somewhat soon which should provide similar functionality but more straightforward.

That being said, thanks for the work that you put into this - the code looks clean & I appreciate that you added tests & replicated the work across PDS & appview.

For future big pieces of work like this, I'd encourage you to put a PR up with just the lexicon & get feedback on that before doing the implementation details

jcsalterego:
@dholms fair enough, all that makes sense. Thanks for taking a look!

PR #1391, opened 2023-07-27
Optimize com.atproto.sync.listRepos

devinivy:
Adds an index to support `listRepos`, and reworks the query to make use of the index.


dholms (approved):
beautiful 👌

packages/pds/src/db/migrations/20230727T172043676Z-user-account-cursor-idx.ts
```diff
@@ -0,0 +1,13 @@
+import { Kysely } from 'kysely'
+
+export async function up(db: Kysely<unknown>): Promise<void> {
+  // await db.schema
```
jcsalterego, line 4: Did you mean to leave the migration up/down commented out?

dholms: Yea, we have to run these migrations CONCURRENTLY on our prod DB which is not supported in kysely's migration system. so we will often do a deploy with the migration commented out so that the migration exists in our DB's migration history while we apply the migration manually

jcsalterego: Ahh makes sense, thanks. I checked `db/migrations` git history and couldn't find previous commenting like this, but perhaps I didn't look far back enough.

dholms: we uncomment it before merging into main 👌

PR #1397, opened 2023-07-28
Support primary vs secondary/replica db in appview

devinivy:
To support both primary and secondary/replica dbs in the appview, I've added a `Primary` typescript interface that we can sprinkle around wherever a primary is needed.  There are also some runtime type guards that have been added to support the few areas in the codebase that mix usage of primary and non-primary dbs.

The primary is used in the following areas:
 - setting mutes
 - setting notification last-seen state
 - labelers
 - all moderation functionality
 - all indexing functionality
 - db leader

Right now the appview effectively requires a primary be configured since it's coupled to the labeler, but with the services starting to be split in #1368 we should be able to start loosening that requirement.

From a configuration standpoint it looks like this:
  - Only db primary configured: services all queries, 1 pool.
  - Only db non-primary configured: services any query that doesn't require a primary, 1 pool.
    - At the moment this isn't supported in practice, see note above re: #1368.
  - Both db primary and non-primary configured: services all queries only using primary when needed, 2 pools.
  - Neither db primary nor non-primary configured: not allowed, wont startup.

> [!NOTE]  
> ~~On launch this will require setting `DB_PRIMARY_POSTGRES_URL` (and if there's not a secondary yet, removing `DB_POSTGRES_URL`).~~


packages/bsky/src/context.ts
```diff
@@ -13,6 +14,7 @@ export class AppContext {
   constructor(
     private opts: {
       db: Database
+      dbPrimary?: Database & Primary
```
dholms: is there a reason this is optional? looks like it should always be set based on the service entry file?

dholms (approved):
looks great - just the question about dbPrimary being optional on context

packages/bsky/src/context.ts
devinivy: That was leftover from my first pass, I think I must have tidied this up while you were reviewing!

dholms: ah nice 👍

dholms:
closing in favor of https://github.com/bluesky-social/atproto/pull/1459

PR #1398, opened 2023-07-28
Read after write

dholms:
This is a basic implementation of read stick on the pds which gives us read-after-write semantics for some operations.

The basic approach here is:
- add a optional logical clock to each commit in the repo
- alongside each record, notate the logical clock at which it was added/updated
- in appview responses, set a header (`atproto-clock`) which includes it's last seen logical clock for the requesting user
- after each proxied request to the appview, check which of the user's records, if any, have not made it into the appview's indexes
- if a relevant record has not made it in yet, add that data into the view locally

Importantly, while this includes application-specific logic, it does not include any application-specific indexes and may be done generically off of the record table.

The application of this read-sticky approach is not exhaustive. However, this applies it to some of the most common and impactful occurrences. Namely:
- profiles in getProfile
- profiles in getProfiles
- posts in getPostThread
- posts in getTimeline
- posts and profiles in getAuthorFeed

Note: this currently builds a PDS url for any local images that get added in, however after landing https://github.com/bluesky-social/atproto/pull/1248, we can deterministically build appview image urls


bnewbold:
I was a bit worried we had specified ourself in to a corner, but thankfully:

> Adding optional fields to commit and MST node objects may or may not result in a repo format version change.

@ericvolp12 had the idea to use `rev` or `revision` as the field name instead of `clock`. I'm fine with `clock` after having used it a bunch, but we might be in a bubble and `rev` might be clearer and be back-pressure against folks assuming the TID value is a safe, trusted, general-purpose timestamp for the commit (eg, if they aren't familiar with Lamport clocks).

packages/bsky/src/api/app/bsky/actor/getProfile.ts
```diff
@@ -31,6 +34,10 @@ export default function (server: Server, ctx: AppContext) {
         throw new InvalidRequestError('Profile not found')
       }
 
+      if (actorClock !== null) {
+        res.setHeader('atproto-clock', actorClock)
```
bnewbold, line 38: In HTTP/2, header will get lower-cased anyways, but shouldn't we `Atproto-Clock` for anybody looking at HTTP/1?

:heavy_plus_sign: to not having an `X-` prefix, as no longer encouraged as a best practice even for experimental stuff.

I would maybe bikeshed to `Atproto-Repo-Clock` or `Atproto-Self-Clock` to clarify semantics, but not a strong opinion, short+simple is nice as well.

packages/common-web/src/tid.ts
```diff
@@ -36,11 +36,15 @@ export class TID {
     if (clockid === null) {
       clockid = Math.floor(Math.random() * 32)
     }
-    return TID.fromTime(timestamp, clockid)
+    const tid = TID.fromTime(timestamp, clockid)
+    if (!prev || tid.newerThan(prev)) {
+      return tid
+    }
+    return TID.fromTime(prev.timestamp() + 1, clockid)
```
devinivy, line 43: I think we also have some leftover bits in `clockid` if that's useful.  One question I have is—would we take the same approach here of incrementing the timestamp if our timestamps already had microsecond precision?

packages/pds/src/app-view/api/app/bsky/actor/getProfiles.ts
```diff
@@ -35,3 +38,28 @@ export default function (server: Server, ctx: AppContext) {
...
+const ensureReadAfterWrite = async (
+  ctx: AppContext,
+  requester: string,
+  res: ApiRes<OutputSchema>,
+): Promise<OutputSchema> => {
+  const clock = getClock(res.headers)
+  if (!clock) return res.data
+  const hasSelf = res.data.profiles.some((prof) => prof.did === requester)
+  if (!hasSelf) return res.data
+  const localSrvc = await ctx.services.local(ctx.db)
```
devinivy, line 51:
```suggestion
  const localSrvc = ctx.services.local(ctx.db)
```

packages/pds/src/app-view/api/app/bsky/feed/getAuthorFeed.ts
```diff
@@ -118,3 +124,50 @@ async function assertNoBlocks(
...
+      return item
+    }
+  })
+  return {
+    ...res.data,
+    feed,
+  }
+}
+
+const getAuthor = (res: ApiRes<OutputSchema>): string | undefined => {
```
devinivy, line 165: I take it this is a little workaround so that we don't have to potentially resolve a handle, does that seem right?

packages/pds/src/app-view/api/app/bsky/feed/getTimeline.ts
```diff
@@ -111,3 +118,56 @@ export default function (server: Server, ctx: AppContext) {
...
+        inserted = true
+        break
+      }
+    }
+    if (!inserted) {
+      feed.push({ post })
+    }
+  }
+  return {
+    cursor: res.data.cursor,
```
devinivy, line 155: I do like these passthroughs since they're a little more relaxed about future changes to the response lexicons:
```suggestion
    ...res.data,
```

packages/pds/src/app-view/api/app/bsky/feed/getTimeline.ts
```diff
@@ -111,3 +118,56 @@ export default function (server: Server, ctx: AppContext) {
...
+  ])
+  const formatted = await findAndFormatPostsToInsert(
+    localSrvc,
+    local.posts,
+    res.data.feed,
+  )
+  if (formatted.length === 0) return res.data
+  const feed = [...res.data.feed]
+  for (const post of formatted) {
+    if (post === null) continue
```
devinivy, line 141: If I'm understanding correctly, I don't think `null`s should be able to show-up here 👍 Maybe the filtering of nulls was added late to `findAndFormatPostsToInsert()`.

packages/pds/src/index.ts
```diff
@@ -45,6 +45,7 @@ import { Crawlers } from './crawlers'
 import { LabelCache } from './label-cache'
 import { ContentReporter } from './content-reporter'
 import { ModerationService } from './services/moderation'
+import { AtpAgent } from '@atproto/api'
```
devinivy, line 48: Tiny baby nit: could be nice to hoist this up near the other `@atproto/*` deps.

packages/pds/src/services/index.ts
```diff
@@ -19,6 +19,8 @@ import { BackgroundQueue } from '../event-stream/background-queue'
 import { Crawlers } from '../crawlers'
 import { LabelCache } from '../label-cache'
 import { ContentReporter } from '../content-reporter'
+import { LocalService } from './local'
+import { AtpAgent } from '@atproto/api'
```
devinivy, line 23: Same tiny baby nit: could be nice to hoist this up near the other `@atproto/*` deps.

packages/bsky/src/api/app/bsky/actor/getProfile.ts
dholms: good call - upper cased it 👍 

hmm i could see us possibly sending multiple clocks down so not opposed to being a bit more specific in the name 🤔 
lemme think on it

packages/common-web/src/tid.ts
dholms: Yea I think we would - likelihood is not that timestamps collide but one has a lower clockid. likelihood is some clockdrift so we'll want to increment the timestamp itself

the clockid is actually just mean to for a multiwriter world to prevent id collision (inspired by twitter snowflakes)

the tid (in JS at least) is actually make of three parts

unix epoch millisecond + count to prevent millisecond collisions + clockid (unique value for machine generating TIDs)

packages/pds/src/app-view/api/app/bsky/feed/getAuthorFeed.ts
dholms: yea basically 👍

imo we basically want to work off of the response + local records as much as possible

packages/pds/src/services/local/index.ts
```diff
@@ -0,0 +1,337 @@
...
+      throw new Error('Could not find bsky appview did')
+    }
+    return createServiceAuthHeaders({
+      iss: did,
+      aud: this.appviewDid,
+      keypair: this.signingKey,
+    })
+  }
+
+  async getRecordsSinceClock(
```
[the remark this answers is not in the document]
dholms: hmmm yup i think that makes sense 👍

packages/pds/src/app-view/api/app/bsky/feed/getPostThread.ts
```diff
@@ -265,3 +271,76 @@ class ParentNotFoundError extends Error {
...
+    }
+  }
+  if (!view.replies) return view
+  const replies = await Promise.all(
+    view.replies.map(async (reply) =>
+      isThreadViewPost(reply)
+        ? await insertIntoThreadReplies(localSrvc, reply, descript)
+        : reply,
+    ),
+  )
```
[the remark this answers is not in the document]
dholms: hmmm i feel fairly comfortable with this & generally it won't even _try_ to do anything deeper than one, because the app will just reload the immediate parent

maybe woulda been worth it if I'd thought to do that before writing this, but now it's good to go 😛

packages/pds/src/app-view/api/app/bsky/feed/getTimeline.ts
dholms: ah yup good call 👍

packages/pds/src/app-view/api/app/bsky/feed/getTimeline.ts
dholms: ah yup that's right 👍

packages/pds/src/index.ts
dholms: yup yup 👍

packages/pds/tests/proxied/read-after-write.test.ts
```diff
@@ -123,4 +164,15 @@ describe('proxy read after write', () => {
...
     expect(res.data.feed[0].post.uri).toEqual(postRes.uri)
   })
+
+  it('returns lag headers', async () => {
+    const res = await agent.api.app.bsky.feed.getTimeline(
+      {},
+      { headers: { ...sc.getHeaders(alice), 'x-appview-proxy': 'true' } },
+    )
+    const lag = res.headers['atproto-upstream-lag']
+    expect(lag).toBeDefined
```
devinivy, line 174:
```suggestion
    expect(lag).toBeDefined()
```

packages/pds/src/services/local/index.ts
```diff
@@ -140,11 +141,41 @@ export class LocalService {
...
+      newestToOldest.map((p) => this.getPost(p)),
+    )
+    const formatted = maybedFormatted.filter((p) => p !== null) as PostView[]
+    for (const post of formatted) {
+      let inserted = false
+      for (let i = 0; i < feed.length; i++) {
+        if (feed[i].post.indexedAt < post.indexedAt) {
+          feed.splice(i, 0, { post })
+          inserted = true
+          break
```
devinivy, line 169: Spidey senses tingled a little seeing a loop over `feed` that also mutates `feed`, but this should be fine since it causes a break.

packages/pds/src/services/local/index.ts
```diff
@@ -140,11 +141,41 @@ export class LocalService {
...
+    if (posts.length === 0) {
+      return feed
+    }
+    const lastTime = feed.at(-1)?.post.indexedAt ?? new Date(0).toISOString()
+    const inFeed = posts.filter((p) => p.indexedAt > lastTime)
+    const newestToOldest = inFeed.reverse()
+    const maybedFormatted = await Promise.all(
+      newestToOldest.map((p) => this.getPost(p)),
+    )
+    const formatted = maybedFormatted.filter((p) => p !== null) as PostView[]
```
devinivy, line 162:
```suggestion
    const maybeFormatted = await Promise.all(
      newestToOldest.map((p) => this.getPost(p)),
    )
    const formatted = maybeFormatted.filter((p) => p !== null) as PostView[]
```

packages/pds/src/app-view/api/app/bsky/feed/getPostThread.ts
```diff
@@ -275,32 +316,39 @@ class ParentNotFoundError extends Error {
...
-  const local = await localSrvc.getRecordsSinceRev(requester, rev, [
-    ids.AppBskyFeedPost,
-  ])
-  const inThread = findPostsInThread(thread, local.posts)
-  if (inThread.length < 0) return res.data
-  for (const record of inThread) {
-    thread = await insertIntoThreadReplies(localSrvc, thread, record)
+  if (!isThreadViewPost(original.thread)) {
+    return original
   }
```
devinivy, line 326: I suppose technically if `original.thread` is a `NotFoundPost` we might consider handling it similarly to a not-found error.  But I am comfortable just leaving a note about that and moving on, since in practice it doesn't happen on our service right now.

packages/pds/src/app-view/api/app/bsky/feed/getPostThread.ts
```diff
@@ -344,3 +388,70 @@ const insertIntoThreadReplies = async (
...
+    data: {
+      thread,
+    },
+    lag: getLocalLag(local),
+  }
+}
+
+const getHighestParent = (
+  thread: ThreadViewPost,
+  height = 0,
```
devinivy, line 450: May be able to get rid of this argument, at least I don't see it getting any use at the moment.

devinivy (approved):

packages/pds/src/services/local/index.ts
dholms: cleaned it up a bit with a `findIndex` instead 👌

packages/pds/src/app-view/api/app/bsky/feed/getPostThread.ts
dholms: yup good call - left a note

PR #1400, opened 2023-07-28
app.bsky.feed.getTimeline gets optional filtering parameters

jcsalterego:
This mirrors functionality from the Home Feed content preferences settings in social-app https://github.com/bluesky-social/social-app/pull/885 by adding the following optional parameters to `app.bsky.feed.getTimeline`:

```
minReplyLikeCount?: number
includeQuotePosts?: boolean
includeReplies?: boolean
includeReposts?: boolean
```

For starters, all default values are backwards-compatible with the present implementation.

Next, social-app could call `app.bsky.feed.getTimeline` with these parameters and push the filtering closer to the DB.

There's probably room for optimization around reply like counts, but I couldn't see a straightforward way to add it to the query builder.

### DIFF
--- a/lexicons/app/bsky/feed/getTimeline.json
+++ b/lexicons/app/bsky/feed/getTimeline.json
@@ -10,7 +10,7 @@
         "properties": {
           "algorithm": {"type": "string"},
           "limit": {"type": "integer", "minimum": 1, "maximum": 100, "default": 50},
-          "minReplyLikeCount": {"type": "integer", "minimum": 0, "maximum": 25, "default": 2},
+          "minReplyLikeCount": {"type": "integer", "minimum": 0, "maximum": 25, "default": 0},
           "showQuotePosts": {"type": "boolean", "default": true},
           "showReplies": {"type": "boolean", "default": true},
           "showReposts": {"type": "boolean", "default": true},
--- a/lexicons/app/bsky/feed/getTimeline.json
+++ b/lexicons/app/bsky/feed/getTimeline.json
@@ -11,9 +11,9 @@
           "algorithm": {"type": "string"},
           "limit": {"type": "integer", "minimum": 1, "maximum": 100, "default": 50},
           "minReplyLikeCount": {"type": "integer", "minimum": 0, "maximum": 25, "default": 0},
-          "showQuotePosts": {"type": "boolean", "default": true},
-          "showReplies": {"type": "boolean", "default": true},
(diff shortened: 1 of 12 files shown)


dholms:
I'm not exactly sure how I feel about this one - it's not totally dissimilar to some filter params that we're adding to `getAuthorFeed` (https://github.com/bluesky-social/atproto/pull/1439)

However, if we add parameters to this, then it would prevent us from abstracting the timeline out into the current feed generator interface. As well, I think there's something nice about `getTimeline` being the simplest possible interface. Clients, of course, are welcome to do any additional filtering on top of that. 

We plan to add params to the feed generator interface at some point, but don't have that quite specced out yet. 

I'll leave this open for a bit for some back and forth, but my instinct is to close for now.

dholms:
(oops meant to comment not close 😅)

jcsalterego:
> I'm not exactly sure how I feel about this one - it's not totally dissimilar to some filter params that we're adding to `getAuthorFeed` (#1439)
> 
> However, if we add parameters to this, then it would prevent us from abstracting the timeline out into the current feed generator interface. As well, I think there's something nice about `getTimeline` being the simplest possible interface. Clients, of course, are welcome to do any additional filtering on top of that.

I would definitely echo that it's similar to `getAuthorFeed` modifications in #1439.

IMHO the main difference is this PR is intending to shift existing complexity and Home Timeline-specific dials and knobs closer to the AppView/PDS. I wasn't aiming for a generalizable and reusable pattern for timeline querying, such as inclusions/exclusions, or minimum/maximum thresholds.

Ultimately my take is: this is probably ok to add and won't hinder whatever the unification towards whatever the feed generator interface ends up being, _assuming_ that 1) `getTimeline` will be deprecated for feed generator 2.0 and 2) you are looking to migrate other endpoints like `getAuthorFeed` to this new(er) world.

> We plan to add params to the feed generator interface at some point, but don't have that quite specced out yet.

This is beyond the scope of this PR, but I think the straightforward part will be "this is how you define and pass parameters to the feed generator interface," and the interesting portion will be how it might surface in the client/UX. I would personally look to Slack Workflow Builder and iOS Shortcuts for inspiration there. The other path (and maybe this is an additional, advanced method) would basically be something that's eventually equivalent to SQL, Lucene or ES Queries.

jcsalterego:
All that said, I'm starting to see where the friction will occur with `getTimelineSkeleton`.

jcsalterego:
Moving discussion to #1462!

Issue #1411, opened 2023-07-31
Factor out separate mute list queries in appview


dholms:
These got factored out in PDS, but that change didn't make it to appview. Reduces read load

PR #1414, opened 2023-08-01
Actor service fix

dholms:
Fixing some bugs introduced in https://github.com/bluesky-social/atproto/pull/1412

Refactors the appview to return a map of records. Introduce separate methods for returning either an individual record, or an ordered hydrated list (which skips over non-existant actors). 

Also refactor requesterMutedBy in appview actor service (Closes https://github.com/bluesky-social/atproto/issues/1411)

### DIFF
--- a/packages/bsky/src/api/app/bsky/feed/getLikes.ts
+++ b/packages/bsky/src/api/app/bsky/feed/getLikes.ts
@@ -1,3 +1,4 @@
+import { mapDefined } from '@atproto/common'
 import { Server } from '../../../../lexicon'
 import { paginate, TimeCidKeyset } from '../../../../db/pagination'
 import AppContext from '../../../../context'
@@ -41,17 +42,23 @@ export default function (server: Server, ctx: AppContext) {
         .actor(db)
         .views.profiles(likesRes, requester)
 
+      const likes = mapDefined(likesRes, (row) =>
+        actors[row.did]
+          ? {
+              createdAt: row.createdAt,
+              indexedAt: row.indexedAt,
+              actor: actors[row.did],
+            }
+          : undefined,
+      )
+
       return {
         encoding: 'application/json',
         body: {
           uri,
           cid,
           cursor: keyset.packFromResult(likesRes),
-          likes: likesRes.map((row) => ({
-            createdAt: row.createdAt,
-            indexedAt: row.indexedAt,
-            actor: actors[row.did],
-          })),
+          likes,
         },
       }
     },
--- a/packages/bsky/src/api/app/bsky/feed/getLikes.ts
+++ b/packages/bsky/src/api/app/bsky/feed/getLikes.ts
@@ -37,7 +37,9 @@ export default function (server: Server, ctx: AppContext) {
       })
 
       const likesRes = await builder.execute()
-      const actors = await services.actor(db).views.profile(likesRes, requester)
+      const actors = await services
+        .actor(db)
+        .views.profiles(likesRes, requester)
 
       return {
         encoding: 'application/json',
@@ -45,10 +47,10 @@ export default function (server: Server, ctx: AppContext) {
           uri,
           cid,
           cursor: keyset.packFromResult(likesRes),
-          likes: likesRes.map((row, i) => ({
+          likes: likesRes.map((row) => ({
             createdAt: row.createdAt,
             indexedAt: row.indexedAt,
-            actor: actors[i],
+            actor: actors[row.did],
           })),
         },
       }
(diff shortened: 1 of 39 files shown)


packages/bsky/src/api/app/bsky/feed/getLikes.ts
```diff
@@ -37,18 +37,20 @@ export default function (server: Server, ctx: AppContext) {
...
           uri,
           cid,
           cursor: keyset.packFromResult(likesRes),
-          likes: likesRes.map((row, i) => ({
+          likes: likesRes.map((row) => ({
             createdAt: row.createdAt,
             indexedAt: row.indexedAt,
-            actor: actors[i],
+            actor: actors[row.did],
           })),
```
devinivy, line 54: I don't think it's really a new issue here, but we may want to skip over likes that don't have a given `actors[row.did]` rather than return a potentially invalid response.

packages/bsky/src/api/app/bsky/notification/listNotifications.ts
```diff
@@ -90,10 +90,10 @@ export default function (server: Server, ctx: AppContext) {
         labelService.getLabelsForUris(recordUris),
       ])
 
-      const notifications = notifs.map((notif, i) => ({
+      const notifications = notifs.map((notif) => ({
         uri: notif.uri,
         cid: notif.cid,
-        author: authors[i],
+        author: authors[notif.authorDid],
```
devinivy, line 96: Similar here, we might consider skipping notifications that don't have an author.

Totally understandable if we want to do these in a separate pass, since I wouldn't anticipate an actual regression.

packages/bsky/src/services/actor/views.ts
```diff
@@ -95,79 +85,100 @@ export class ActorViews {
...
-        labels: labels[result.did] ?? [],
+        labels: labels[cur.did] ?? [],
       }
-    })
-
-    return Array.isArray(result) ? views : views[0]
+      return {
+        ...acc,
+        [cur.did]: profile,
+      }
```
devinivy, line 154: I generally don't focus too much on the style of accumulation (e.g. mutating vs immutable), but I wonder if we should actually consider mutating in hot paths like these view methods.  It does add-up to quite a few assignments and thrown-away objects that I think even today may not get optimized away by V8.

devinivy: In case it wasn't clear what I was getting at, e.g.:
```suggestion
      acc[cur.did] = profile
      return acc
```

packages/bsky/src/services/actor/views.ts
```diff
@@ -10,32 +9,22 @@ import { noMatch, notSoftDeletedClause } from '../../db/util'
...
     opts?: { skipLabels?: boolean; includeSoftDeleted?: boolean },
-  ): Promise<ProfileViewDetailed[]>
-  async profileDetailed(
-    result: ActorResult | ActorResult[],
-    viewer: string | null,
-    opts?: { skipLabels?: boolean; includeSoftDeleted?: boolean },
-  ): Promise<ProfileViewDetailed | ProfileViewDetailed[]> {
-    const results = Array.isArray(result) ? result : [result]
-    if (results.length === 0) return []
+  ): Promise<Record<string, ProfileViewDetailed>> {
```
devinivy, line 26: I really don't think we need to do it here/now, but I don't think it would be out of control to eventually start returning `Map`s from these methods rather than `Record`s.  The main thing I worry about with `Record`s is that `record[key]` is always typed to exist, whereas `map.get(key)` is more honest about the `key` possibly not existing in the map.  Maps are also just a little better suited to dynamic keys under the hood.  Of course there's something nice/simple/familiar about using the object notation, that's the main tradeoff.

packages/bsky/src/services/actor/views.ts
```diff
@@ -95,79 +85,100 @@ export class ActorViews {
...
-              muted: !!profileInfo?.requesterMuted || !!listMutes[result.did],
-              mutedByList: listMutes[result.did],
-              blockedBy: !!profileInfo.requesterBlockedBy,
-              blocking: profileInfo.requesterBlocking || undefined,
+              following: cur?.requesterFollowing || undefined,
+              followedBy: cur?.requesterFollowedBy || undefined,
+              muted: !!cur?.requesterMuted || !!cur.requesterMutedByList,
+              mutedByList: cur.requesterMutedByList
+                ? this.services.graph.formatListViewBasic(
+                    listViews[cur.requesterMutedByList],
```
devinivy, line 142: It's technically possible this list view might not exist to pass for formatting.

devinivy (approved):
Had a few general thoughts about how we work with all the "info maps" that come out of the view services, but nothing actually pressing.  Looks solid, the little reorg feels good 👍

packages/bsky/src/api/app/bsky/notification/listNotifications.ts
dholms: yup 👍

packages/bsky/src/services/actor/views.ts
dholms: i'm game 👍

packages/bsky/src/services/actor/views.ts
dholms: yeah I'd support - we've had some gotchas related to that & it'd be nice to be forced to reckon with it

PR #1420, opened 2023-08-02
Codegen `ErrorName` enum for each resource

estrattonbailey:
Small quality-of-life update. Updates `lex-cli` to generate a `ErrorName` enum that maps possible error values as specified by the lexicon configs.

Usage example:

```ts
import { AppBskyFeedGetAuthorFeed } from '@atproto/api'

  `Requester is blocked by actor: ${actor}`,
  AppBskyFeedGetAuthorFeed.ErrorName.BlockedByActor // instead of a string 'BlockedByActor'
)
```

Is `ErrorName` naming OK?

### DIFF
--- a/packages/api/src/client/lexicons.ts
+++ b/packages/api/src/client/lexicons.ts
@@ -530,7 +530,6 @@ export const schemaDict = {
       },
       moderation: {
         type: 'object',
-        required: [],
         properties: {
           currentAction: {
             type: 'ref',
@@ -1652,7 +1651,7 @@ export const schemaDict = {
       },
       reasonSexual: {
         type: 'token',
-        description: 'Unwanted or mis-labeled sexual content',
+        description: 'Unwanted or mislabeled sexual content',
       },
       reasonRude: {
         type: 'token',
--- a/packages/api/src/client/types/app/bsky/actor/getPreferences.ts
+++ b/packages/api/src/client/types/app/bsky/actor/getPreferences.ts
@@ -27,6 +27,8 @@ export interface Response {
   data: OutputSchema
 }
 
+export enum ErrorName {}
+
 export function toKnownErr(e: any) {
   if (e instanceof XRPCError) {
   }
--- a/packages/api/src/client/types/app/bsky/actor/getProfile.ts
+++ b/packages/api/src/client/types/app/bsky/actor/getProfile.ts
@@ -25,6 +25,8 @@ export interface Response {
   data: OutputSchema
 }
 
+export enum ErrorName {}
+
 export function toKnownErr(e: any) {
   if (e instanceof XRPCError) {
   }
--- a/packages/api/src/client/types/app/bsky/actor/getProfiles.ts
+++ b/packages/api/src/client/types/app/bsky/actor/getProfiles.ts
@@ -29,6 +29,8 @@ export interface Response {
   data: OutputSchema
 }
 
+export enum ErrorName {}
+
 export function toKnownErr(e: any) {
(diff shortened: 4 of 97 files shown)


dholms:
Definitely in support of this direction 👍 

The client package actually already has classes for it's known errors. So while I'm not opposed to generating these enums in the api types, I think they'd be better suited for the server types. And we could reference that enum in `HandlerError` instead of the string union.

Something I've wanted to do for awhile is generate actual error classes for each method. So you could just do something like `throw new ActorNotFoundError('blah')` instead of `throw new InvalidRequestError('blah', 'ActorNotFound')`. But that's probably too much of a side quest & would probably require some wiring up in xrpc-server 😛

estrattonbailey:
Let me try to summarize this better.

### Overview
In order for API consumers to effectively differentiate between error responses, we generate `<Name>Error` subclasses of `XRPCError`, which are hydrated from error responses like `{ error: 'UnknownFeed', message: 'Blah blah' }` so that the consumers can do comparisons like `if (e instanceof UnknownFeedError)` and handle specific errors.

To expand the possible error modes, we simply add more error types to Lexicons as needed.

In order to respond to an API consumer with a error type they can expect and handle, we must _currently_ pass in string values that correspond to the Lexicon-defined error types and code-generated error subclasses e.g.:

```typescript
throw new InvalidResponseError('message', 'UnknownFeed')
```

### Problem

Strings are prone to typing errors, and since we _need_ to use our common server error subclasses like `InvalidResponseError` in order to control the HTTP status code, we can't effectively type-check the strings we pass as the `customErrorName`.

In fact, we already have an invalid usage like this in the codebase within `getFeed`: `InvalidFeedResponse`. `InvalidFeedResponse` doesn't exist in the Lexicon defs, and therefore an API consumer would not be able to effectively handle that error.

A lesser problem is also that for API consumers, checking `instanceof` against more than a few error modes is a little more cumbersome that it maybe needs to be.

### Solution

Using the same Lexicon defs, generate an TypeScript `enum` that both API consumers and we internally can use to ensure a thrown error can be effectively mapped to an error response on the client-side.

Usage on the server looks like this:
```typescript
import { AppBskyFeedGetAuthorFeed } from '@atproto/api'

  `Requester is blocked by actor: ${actor}`,
  AppBskyFeedGetAuthorFeed.ErrorName.BlockedByActor // instead of a string 'BlockedByActor'
)
```

And usage on the client would look like this, where you can do exhaustive checking with a `switch`:

```typescript
import { AppBskyFeedGetAuthorFeed } from '@atproto/api'

switch(error.name) {
  case AppBskyFeedGetAuthorFeed.ErrorName.BlockedByActor:
    // handle error
    break;
  default:
    // handle default
}
```

We can of course continue to generate subclasses like `BlockedByActorError` as well, and give end-users the choice.

estrattonbailey:
>Something I've wanted to do for awhile is generate actual error classes for each method.

Yeah I dig that! Somehow we'd need to config the HTTP codes too. Maybe that could go in the Lexicon?

dholms:
> Somehow we'd need to config the HTTP codes too. Maybe that could go in the Lexicon?
ah good point - I'd definitely support getting those into the lexicon.

Okay sweet I understand your reasoning a bit better here 👌 
As well, checking instances of classes can be fickle if you accidentally have two versions of the same package installed due to subdependencies etc

I support this approach 👍 The one other piece that I'd ask for is code-genning those same error enums on the server side so that we can import from the server types instead of from `@atproto/api`

PR #1422, opened 2023-08-02
Return Like URIs when fetching a page of likes

ericvolp12:
The `getLikes` endpoint returns a page of likes but it doesn't provide the URI of the like records which makes it impossible to lookup the like record from the response. This change adds the like URI to the paginated response.

### DIFF
--- a/lexicons/app/bsky/feed/getLikes.json
+++ b/lexicons/app/bsky/feed/getLikes.json
@@ -33,10 +33,11 @@
     },
     "like": {
       "type": "object",
-      "required": ["indexedAt", "createdAt", "actor"],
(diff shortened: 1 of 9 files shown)


JoelOtter:
Thank you :) Would it be possible to have URIs in the `getRepostedBy` response as well?

ericvolp12:
> Thank you :) Would it be possible to have URIs in the `getRepostedBy` response as well? I'd be happy to attempt making this change myself if given the maintainers' blessing

Yeah I think we need them in `getLikes`, `getRepostedBy`, `getFollows`, and `getFollowers` but it'll be a breaking change for some of those since they currently just return lists of author references.

PR #1423, opened 2023-08-02
Push labels to PDS

devinivy:
Allows the appview to push labels to the PDS using an unspecced route, as a temporary measure.  Also adds an http retry to the hive labeler.

### DIFF
--- a/lexicons/app/bsky/unspecced/applyLabels.json
+++ b/lexicons/app/bsky/unspecced/applyLabels.json
@@ -4,27 +4,16 @@
   "defs": {
     "main": {
       "type": "procedure",
-      "description": "Apply labels to a subject directly.",
+      "description": "Allow a labeler to apply labels directly.",
       "input": {
         "encoding": "application/json",
         "schema": {
           "type": "object",
-          "required": ["subject"],
+          "required": ["labels"],
           "properties": {
-            "subject": {
-              "type": "union",
-              "refs": [
-                "com.atproto.admin.defs#repoRef",
-                "com.atproto.repo.strongRef"
-              ]
-            },
-            "createLabelVals": {
+            "labels": {
               "type": "array",
-              "items": { "type": "string" }
-            },
-            "negateLabelVals": {
-              "type": "array",
-              "items": { "type": "string" }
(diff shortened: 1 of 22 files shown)


lexicons/app/bsky/unspecced/applyLabels.json
```diff
@@ -0,0 +1,34 @@
...
+          "required": ["subject"],
+          "properties": {
+            "subject": {
+              "type": "union",
+              "refs": [
+                "com.atproto.admin.defs#repoRef",
+                "com.atproto.repo.strongRef"
+              ]
+            },
+            "createLabelVals": {
```
dholms, line 21: i kinda want this to just take already formatted labels 🤔 

so all the pds has to do is dumb shuffling of them off to the database, no formatting or anything & nowhere that the labels could get out of sync

packages/bsky/src/labeler/hive.ts
```diff
@@ -51,13 +52,16 @@ export const makeHiveReq = async (
...
   form.append('media', blob)
-  const res = await axios.post(HIVE_ENDPOINT, form, {
-    headers: {
-      'Content-Type': 'multipart/form-data',
-      authorization: `token ${hiveApiKey}`,
-      accept: 'application/json',
-    },
-  })
+
+  const res = await retryHttp(() =>
```
dholms, line 56: this does max of 3 right? seems good 👍

dholms (approved):
seems solid 👍 

as i mentioned, i'm inclined to do this with full label objects instead of `create/deleteLabelVals`. Just seems simpler & less prone to small bugs. But also not married to it & think this is totally fine

lexicons/app/bsky/unspecced/applyLabels.json
devinivy: Yeah, I like that more—on it 👍

packages/bsky/src/labeler/hive.ts
devinivy: Yup!

dholms (approved):
:shipit:

PR #1426, opened 2023-08-02
remove unused/unspecified label

bnewbold:
NOTE: this commit contains no change to mod policy, app behavior, etc.

We have been creating redundant "underwear" labels along with "sexual" for many months. "sexual" is the actual label that gets actioned in mod preferences, etc. The un-specified "underwear" labels are confusing and take up resources on-disk and in API responses.

cc: @pfrazee for visibility


pfrazee:
ack

dholms (approved):
this slipped by me, but seems good 👍 

need to reflect this change in the appview labeler before merge - i'll do that now

Issue #1399, opened 2023-07-28
Make it so users don't (or can can choose not to) see followers they have blocked


jesslark:
**Is your feature request related to a problem? Please describe.**

It can feel really bad to have to see people on the followers list who have been blocked for significant reasons. For instance, if an abusive ex, or a child predator, has followed a user, they may actively want to avoid see those names. Right now, following someone can be abused as part of harassing that person.

**Describe the solution you'd like**

I think it would be ideal for a block to remove all follows between accounts. I understand that this would mean userA actually impacts userB and that that might be something your team is currently avoiding. Other suggestions to follow.

**Describe alternatives you've considered**

Wide impact solution
- make it so blocked followers are hidden from the following list for all users, this means if userA has blocked userB, no users would be able to see that userB follows userA 
- optionally, add a toggle so the above suggestion only applies when the blocking account opts in

Isolated to the blocking account solution
- make it so blocked followers are hidden from the following list for the users who blocked them, this means if userA has blocked userB, userC would be able to see that userB follows userA, but userA would not see userB in their followers list
- optionally, add a toggle so the above suggestion only applies when the blocking account opts in

khizka:
as it is a temporary solution, I think this shouldn't be protocol level

a better protocol level solution might be a two way follow agreement

I think this would enable follow requests and removing followers

khizka:
just saw #1352

but I think this could work as a temporary solution in https://github.com/bluesky-social/social-app/

jesslark:
Good point! The sole reason I proposed it here was that I interacted here before. I do not have a technical opinion on where or how this feature should or could be implemented. Would you like to copy this and paste it over on the social app? If not, I can do it.

khizka:
I think @pfrazee should just transfer the issue

jesslark:
Fine with that! @pfrazee please move this issue to where it goes if this is not the best place for it. Thanks so much! :)

pfrazee:
It would be a bit slow to do in the client. We have an optimized route for fetching the follows with pagination. To do this we'd also have to directly query all of the block records when looking at follows, which we'd probably want to do async so you'd see the follows weirdly disappear after.

Likely the best move here is to resolve the discussion about causing unfollows into one of three states (yes, no, blocked) and if no/blocked we can look at applying this filtering logic at the server (cc @dholms)

dholms:
Looks like a mixed up the creator/subject columns on the block check for these. Sorry yall!

Will be getting this in soon
[...]

PR #1427, opened 2023-08-02
Fix blocks in follow views

dholms:
Pretty sure I had subject/creator swapped in these queries.

Also looks like blocks got missed in these routes in the appview code (though that code is not yet in service), so I added them in here.

This probably needs a bit more work (or a quick followup) so that follows that violate a block relationship are not showed to any viewers - similar to how we handle blocked replies.

Closes https://github.com/bluesky-social/atproto/issues/1399

### DIFF
--- a/packages/pds/src/app-view/api/app/bsky/graph/getFollowers.ts
+++ b/packages/pds/src/app-view/api/app/bsky/graph/getFollowers.ts
@@ -43,7 +43,7 @@ export default function (server: Server, ctx: AppContext) {
         )
         .where(notSoftDeletedClause(ref('creator_repo')))
         .whereNotExists(
-          graphService.blockQb(requester, [ref('follow.subjectDid')]),
+          graphService.blockQb(requester, [ref('follow.creator')]),
         )
         .selectAll('creator')
         .select(['follow.cid as cid', 'follow.createdAt as createdAt'])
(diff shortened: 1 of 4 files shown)


devinivy (approved):

packages/pds/src/app-view/api/app/bsky/graph/getFollowers.ts
```diff
@@ -43,7 +43,13 @@ export default function (server: Server, ctx: AppContext) {
...
         .whereNotExists(
-          graphService.blockQb(requester, [ref('follow.subjectDid')]),
+          graphService.blockQb(requester, [ref('follow.creator')]),
+        )
+        .whereNotExists(
+          graphService.blockRefQb(
+            ref('follow.subjectDid'),
+            ref('follow.creator'),
+          ),
         )
```
devinivy, line 53: If this ends-up being too heady of a query, we can always use the same approach as `blocksForPosts()`/`getBlockSet()`.

dholms: yeah i think we feel it out - i think this is a bit easier to model in-line than as a separate query (in comparison to replies). but if we're having trickery with it, we can totally split it out

PR #1428, opened 2023-08-03
Native Notifications

ansh:
Types of notifications handled
- [x] like
- [x] comment/reply
- [x] quote
- [x] repost
- [x] follow
- [x] custom feed like

Misc items

- [x] Separate AppView and PDS concerns
- [x] Move notification sending out of `processor.ts` into indexing service
- [x] Duplicate code in AppView
- [x] Add tests
- [x] Rate limiting and queueing

Future:
- [ ] Show image previews in notification (need to do some CDN resizing/compression stuff for this) --  not very important
- [ ] Handle notifs for web (gorush does not support web notifs)


packages/bsky/src/services/indexing/processor.ts
```diff
@@ -246,6 +247,30 @@ export class RecordProcessor<T, S> {
     }
   }
 
+  async sendNotifsToUser(notifs: Notif[]) {
```
dholms, line 250: heads up, we'll probably want to do the actual _sending_ of notifications from within the indexing service (not the record processor). the `insert/update/deleteRecord` functions on the processor would probably want to return some information about notifications that the indexing service could then send off through the gorush client. We'll want to wire the gorush client up through the `AppContext` and give the indexing service access to it

ansh: Done! Thanks for the heads up

packages/bsky/src/notifications.ts
```diff
@@ -0,0 +1,285 @@
...
+      }
+
+      for (const t of userTokens) {
+        const { appId, platform, token } = t
+        // get title and message of notification
+        const attr = await this.getNotificationDisplayAttributes(notif)
+        // if no title or body, skip
+        if (!attr) {
+          continue
+        }
```
estrattonbailey, line 59: Looks like `attr` will always be defined, unless `getNotificationDisplayAttributes` throws, in which case this loop will exit, right?

Do we want to catch and filter out invalid notifications so we can send valid ones, or just exit the whole notification process?

packages/bsky/src/notifications.ts
```diff
@@ -0,0 +1,285 @@
...
+            title: title,
+            message: body,
+            topic: appId,
+            data: {
+              reason: notif.reason,
+              recordUri: notif.recordUri,
+              recordCid: notif.recordCid,
+            },
+          })
+        } else if (platform === 'android') {
```
estrattonbailey, line 74: Nit: could use a ternary for the value of `platform` prop here and dry it up a bit

packages/bsky/src/notifications.ts
```diff
@@ -0,0 +1,285 @@
...
+        {
+          headers: {
+            'Content-Type': 'application/json',
+            accept: 'application/json',
+          },
+        },
+      )
+    } catch (error) {
+      throw new Error('Failed to send push notification')
+    }
```
estrattonbailey, line 139: Maybe silly question: does gorush already give us enough logs, or should we send along more info from `error` here?

packages/bsky/src/notifications.ts
```diff
@@ -0,0 +1,285 @@
...
+  async registerDeviceForPushNotifications(
+    did: string,
+    platform: Platform,
+    token: string,
+    endpoint = PUSH_NOTIF_SERVER_URL,
+    appId = BSKY_APP_ID,
+  ) {
+    try {
+      // check if token did pair already exists
+      const existing = await this.db.db
```
estrattonbailey, line 151: I'm no expert on this but I think we might be able to do something like an insert and `on conflict do nothing` in a single query instead of two 🤔

packages/bsky/src/services/indexing/index.ts
```diff
@@ -93,7 +97,20 @@ export class IndexingService {
...
-        await indexer.insertRecord(uri, cid, obj, timestamp)
+        const insertedRecords = await indexer.insertRecord(
+          uri,
+          cid,
+          obj,
+          timestamp,
+        )
+        // THIS WILL SEND THE NOTIFICATIONS
+        if (insertedRecords) {
+          /* don't await */ this.notifServer
```
estrattonbailey, line 108: Nice comment 👍

packages/pds/src/app-view/api/app/bsky/unspecced.ts
```diff
@@ -192,6 +192,69 @@ export default function (server: Server, ctx: AppContext) {
...
+        await notificationServer.registerDeviceForPushNotifications(
+          did,
+          platform,
+          token,
+          endpoint,
+          appId,
+        )
+      } catch (error) {
+        throw new Error('Failed to register device push notification token')
+      }
```
estrattonbailey, line 213: Is this gonna swallow earlier errors? I'm honestly unclear on how our error logs are reported, I'll look into that 🤔

estrattonbailey (commented):
Sick, looking good to me 😎

packages/pds/src/app-view/api/app/bsky/unspecced.ts
ansh: I am unclear too. @dholms can give some guidance

packages/bsky/src/services/indexing/index.ts
ansh: Lol

packages/bsky/src/notifications.ts
ansh: Good callout! Fixes.

packages/bsky/src/notifications.ts
ansh: `gorush` basically gives no logs unfortunately

packages/bsky/src/notifications.ts
ansh: We will need to create a new interface for web notifs so that is why I made it more verbose.

packages/pds/src/notifications.ts
```diff
@@ -0,0 +1,283 @@
+import axios from 'axios'
```
dholms, line 1: I think we can totally remove the notification server from the pds - only needs to run in the appview

lexicons/app/bsky/unspecced/registerPushNotificationEndpoint.json
```diff
@@ -0,0 +1,20 @@
+{
+  "lexicon": 1,
+  "id": "app.bsky.unspecced.registerPushNotificationEndpoint",
```
dholms, line 3: I'm not sure the different between these routes, but I like this name better - altho I would just do `registerPushNotifications`

packages/bsky/src/notifications.ts
ansh: There were times I was getting no `attr` like it was an empty object `{}`. Not 100% sure why

packages/pds/src/notifications.ts
ansh: Okay. I can remove it completely.

packages/pds/src/app-view/api/app/bsky/unspecced.ts
```diff
@@ -192,6 +192,69 @@ export default function (server: Server, ctx: AppContext) {
...
+          //   {
+          //     token,
+          //     platform,
+          //     appId,
+          //     endpoint,
+          //   },
+          //   headers,
+          // )
+          const res =
+            await appviewAgent.api.app.bsky.unspecced.putNotificationPushToken(
```
dholms, line 242: unless i'm misunderstanding something, I think this should forward the request to the method of the same name on the appview

packages/pds/src/app-view/api/app/bsky/unspecced.ts
```diff
@@ -192,6 +192,69 @@ export default function (server: Server, ctx: AppContext) {
...
+              {
+                token,
+                platform,
+                appId,
+                endpoint,
+              },
+              await ctx.serviceAuthHeaders(did),
+            )
+        } catch (error) {
+          throw new Error('Failed to register device push notification token')
```
dholms, line 252: Errors like this will get transformed into an opaque internal server error - which won't report anything that went wrong to the client.

Throwing one of the errors from `@atproto/xrpc-server` will do better reporting

packages/pds/src/app-view/db/tables/notification-push-token.ts
```diff
@@ -0,0 +1,11 @@
+export const tableName = 'notification_push_token'
```
dholms, line 1: we don't need to store these on the pds either i think

packages/bsky/src/notifications.ts
```diff
@@ -0,0 +1,285 @@
+import axios from 'axios'
+import Database from './db'
+import { Notification } from './db/tables/notification'
+import { AppBskyEmbedImages, AtUri } from '@atproto/api'
+import { Insertable } from 'kysely'
+
+const PUSH_NOTIF_SERVER_BASE_URL = 'https://push.bsky.app'
```
dholms, line 7: we'll need to wire these up through the config so they can be set per environment

packages/pds/src/notifications.ts
ansh: Done.

packages/bsky/src/notifications.ts
```diff
@@ -0,0 +1,285 @@
...
+  topic: string
+  data?: {
+    [key: string]: string
+  }
+}
+type InsertableNotif = Insertable<Notification>
+
+export class NotificationServer {
+  constructor(public db: Database) {}
+  static creator() {
```
dholms, line 27: is this creator function being used anywhere?

packages/pds/src/app-view/db/tables/notification-push-token.ts
ansh: Yup, I'll remove

ansh: removed.

packages/bsky/src/notifications.ts
```diff
@@ -0,0 +1,285 @@
...
+      const userTokens = await this.getUserTokens(userDid)
+      // if user has no tokens, skip
+      if (!userTokens || userTokens.length === 0) {
+        continue
+      }
+
+      for (const t of userTokens) {
+        const { appId, platform, token } = t
+        // get title and message of notification
+        const attr = await this.getNotificationDisplayAttributes(notif)
```
dholms, line 55: does this need to be done inside of the loop? looks like it only depends on the notif not the token

packages/pds/src/app-view/api/app/bsky/unspecced.ts
ansh: Bad naming on my part, will remove this function

packages/bsky/src/notifications.ts
dholms: yeah kinda looks like this will just eat whatever the axios error is - I'd probably just let axios throw, or at least preserve the message for the new error

packages/bsky/src/notifications.ts
dholms: yup yup^^

looks like `.onConflict((oc) => oc.donothing())`

packages/bsky/src/notifications.ts
```diff
@@ -0,0 +1,285 @@
...
+        .values({
+          did,
+          token,
+          platform,
+          endpoint,
+          appId,
+        })
+        .execute()
+    } catch (error) {
+      throw new Error('Failed to insert notification token')
```
dholms, line 163: here's another spot where we should avoid eating the error & just let kysely throw so that we can see it in logs

packages/bsky/src/notifications.ts
ansh: Nope, removed.

packages/bsky/src/notifications.ts
ansh: Good callout

packages/bsky/src/notifications.ts
```diff
@@ -0,0 +1,285 @@
...
+        .executeTakeFirst()
+      body = mentionedPostData?.text || ''
+      return { title, body }
+    }
+
+    if (!subjectUri) {
+      throw new Error('Failed to get subject URI')
+    }
+
+    const postData = await this.db.db
```
dholms, line 215: since the indexing service already has access to the record in question, it may be better to just get the relevant data from there instead of following it up with a bunch of selects

especially a Create with a wide fanout - like a reply in a thread - could end up hitting the database a lot

packages/bsky/src/notifications.ts
ansh: Alright, just gonna let axios throw!

packages/bsky/src/notifications.ts
```diff
@@ -0,0 +1,285 @@
...
+    }
+
+    if (title === '' && body === '') {
+      throw new Error('Failed to get title and body')
+    }
+
+    return { title: title, body: body }
+  }
+
+  static sanitizeDisplayName(str: string): string {
```
dholms, line 261: how necessary do you think this is? genuinely asking. I don't love it being in the server since we don't do much other sanitization :/ but also push notifications are kinda like "client in the server" lol

packages/pds/src/app-view/api/app/bsky/unspecced.ts
ansh: Will let `kysley` handle that

packages/bsky/src/services/indexing/index.ts
dholms, line 108: for things we don't want to do synchronously, we normally throw them on the `backgroundQueue` (which is available in this service)

that let's us wait for backgrounded tasks to finish before tearing down an instance, puts a cap on concurrency of backgrounded tasks & it handles error handling + logging

lexicons/app/bsky/unspecced/registerPushNotificationEndpoint.json
ansh: Changed it to `registerPushNotifications`

packages/bsky/src/notifications.ts
ansh: Done!

packages/bsky/src/services/indexing/processor.ts
```diff
@@ -84,7 +84,7 @@ export class RecordProcessor<T, S> {
     if (inserted) {
       this.aggregateOnCommit(inserted)
       await this.handleNotifs({ inserted })
-      return
+      return this.params.notifsForInsert(inserted)
```
dholms, line 87: we may want to return something else here that has more details on the record - so that the notification server doesn't have to do a bunch of re-querying stuff out of the db

packages/bsky/src/db/migrations/20230802T232331476Z-native-notifications.ts
```diff
@@ -0,0 +1,17 @@
+import { Kysely } from 'kysely'
+
+export async function up(db: Kysely<unknown>): Promise<void> {
+  await db.schema
+    .createTable('notification_push_token')
```
dholms, line 5: we'll probably want to split endpoints & tokens into different tables at some point - I think I'm inclined to do so now to avoid a future migration 🤔

dholms: hmmm actually, i don't think we even need endpoint yet if we're not doing 3rd party push notifs

packages/bsky/src/db/migrations/index.ts
```diff
@@ -21,4 +21,5 @@ export * as _20230627T212437895Z from './20230627T212437895Z-optional-handle'
 export * as _20230629T220835893Z from './20230629T220835893Z-remove-post-hierarchy'
 export * as _20230703T045536691Z from './20230703T045536691Z-feed-and-label-indices'
 export * as _20230720T164800037Z from './20230720T164800037Z-posts-cursor-idx'
+export * as _20230802T232331476Z from './20230802T232331476Z-native-notifications'
```
dholms, line 24: (before merge, we need to re-create the migration & ensure that it's the highest value in this list)

packages/bsky/src/notifications.ts
ansh: We do this sanitizing on the client so that is why I moved it to the sever. I think it is for people who tried to "act" as verified by putting a checkmark in their name. Upto you and @pfrazee if we should remove it.

packages/bsky/src/notifications.ts
dholms: ah just realized - i guess it doesn't have access to the subject which would be the big saver 🤔

packages/bsky/src/notifications.ts
```diff
@@ -0,0 +1,285 @@
...
+    if (reason === 'like') {
+      title = `${author} liked your post`
+      body = postData?.text || ''
+      // custom feed like
+      if (subjectUri?.includes('feed.generator')) {
+        title = `${author} liked your custom feed`
+        body = `${new AtUri(subjectUri).rkey}`
+      }
+    } else if (reason === 'reply') {
+      title = `${author} replied to your post`
```
dholms, line 223: does it make more sense to include the reply text here?

packages/bsky/src/notifications.ts
dholms: still let's try to do as few selects as possible for the notif fanout

ansh: yeah... it would...

packages/pds/src/app-view/api/app/bsky/unspecced.ts
ansh: done

packages/pds/src/app-view/api/app/bsky/unspecced.ts
dholms: yeah this swallows earlier errors both from our perspective & from the clients perspective

if we throw errors from `@atproto/xrpc-server` those will get reported back to the client, everything else will get translated to an internal server error & logged for our sake

packages/pds/tests/feed-generation.test.ts
```diff
@@ -347,6 +347,18 @@ describe('feed generation', () => {
     })
   })
 
+  describe('putNotifications', () => {
+    it('updates notifications', async () => {
```
dholms, line 351: guessing this was just for internal testing but will wanna clean up before merge

ansh: yes. I am still moving my tests over to a new file!

packages/bsky/src/notifications.ts
ansh: I can probably remove it since it doesn't seem like people are doing that anymore but I don't know if its still an issue

dholms: I think it's fine to remove tbh - when they open the app, they'd see the sanitized name

ansh: Removed

packages/bsky/src/notifications.ts
ansh: Done

packages/pds/src/app-view/api/app/bsky/unspecced.ts
```diff
@@ -192,6 +192,43 @@ export default function (server: Server, ctx: AppContext) {
...
+      const { token, platform, appId, endpoint } = params
+      const {
+        credentials: { did },
+      } = auth
+
+      const { appviewAgent } = ctx
+
+      if (ctx.canProxyWrite()) {
+        // TODO: hit the endpoint here so that 3rd party devs can run their own notification servers
+        // const headers = await ctx.serviceAuthHeaders(did)
```
dholms, line 207: mind cleaning up this comment - leaving the todo is great 👌 

(altho such a nit but can we write it as `@TODO`)

packages/bsky/tests/notification-server.test.ts
```diff
@@ -0,0 +1,74 @@
+import AtpAgent, { AtUri } from '@atproto/api'
+import { TestNetwork } from '@atproto/dev-env'
+import { SeedClient } from './seeds/client'
+import basicSeed from './seeds/basic'
+import { NotificationServer } from '../src/notifications'
+import { Database } from '../src'
+import { sql } from 'kysely'
+
+describe('notification views', () => {
```
dholms, line 9: it'd be cool to see some tests of it actually sending the notifications

you could wire up a very simple mock gorush service here (just an express app with one route) that makes sure that the notifs get sent properly formatted

packages/bsky/src/services/indexing/index.ts
dholms: you can see an example of this in, for instance, the `Labeler` class

it generally looks something like this:
```ts
    this.backgroundQueue.add(() =>
      this.createAndStoreLabels(uri, obj).catch((err) => {
        log.error(
          { err, uri: uri.toString(), record: obj },
          'failed to label record',
        )
      }),
    )
```

packages/bsky/src/notifications.ts
```diff
@@ -0,0 +1,253 @@
...
+      body = postData?.text || ''
+    } else if (reason === 'quote') {
+      title = `${author} quoted your post`
+      body = postData?.text || ''
+    } else if (reason === 'repost') {
+      title = `${author} reposted your post`
+      body = postData?.text || ''
+
+      // TODO: handle images
```
dholms, line 235: just curious, is this gonna make it in in this PR or save for later?

if not, let's delete the commented coded but can preserve the TODO (but again change to `@TODO`)

packages/bsky/src/notifications.ts
```diff
@@ -0,0 +1,253 @@
...
+      //     notification.additionalPost?.thread?.post.embed,
+      //   ) &&
+      //   notification.additionalPost?.thread?.post.embed.images[0]?.thumb
+      // ) {
+      //   image = notification.additionalPost.thread.post.embed.images[0].thumb
+      // }
+    }
+
+    if (title === '' && body === '') {
+      throw new Error('Failed to get title and body')
```
dholms, line 248: it's actually totally fine to throw in here, if we have this on the background queue & are catching errors

that being said, these errors aren't gonna give us a ton of info & it might be nice to stick some additional fields on them (like the record it had a hard time processing for instance)

alternately, we can not throw & just log then return

dholms: maybe a custom error class that allows you to stick some additional info on

packages/bsky/src/notifications.ts
```diff
@@ -0,0 +1,253 @@
...
+        token,
+        platform,
+        appId,
+        endpoint,
+      })
+      .onConflict((oc) => oc.doNothing())
+      .execute()
+  }
+
+  async getNotificationDisplayAttributes(notif: InsertableNotif) {
```
dholms, line 158: a return type on this one might be helpful to clarify that it can return `undefined`

packages/bsky/src/notifications.ts
dholms: sry real quick to circle back on this - how different is the web interface? 

i'm also inclined to simplify this function just a bit

honestly even something like
```ts
if(platform === 'ios' || platform === 'android'){
  // do the format w a ternary
} else {
  // handle web
}
```

packages/bsky/src/notifications.ts
ansh: Nah I don't think it is necessary. I'll remove it for now.

packages/bsky/src/notifications.ts
ansh: The web interface will be completely different. Happy to clean this up a bit like you suggested but I was optimizing for readability.

packages/bsky/src/services/indexing/index.ts
ansh: Done

packages/bsky/src/notifications.ts
ansh: I think logging them is probably fine as they are not critical.

packages/bsky/src/notifications.ts
ansh: Yeah that is probably better. Fixed.

packages/bsky/tests/notification-server.test.ts
ansh: I added a test to check the notif format!

ansh: Let me know if you still want me to mock `gorush` but it really doesn't return anything useful. Just successes every time.

packages/bsky/src/db/migrations/index.ts
ansh: Sounds good. I'll let you handle that since you will be the one merging this.

packages/dev-env/src/bsky.ts
```diff
@@ -88,6 +88,7 @@ export class TestBsky {
       indexerPartitionIds: [0],
       indexerNamespace: `ns${ns}`,
       indexerSubLockId: uniqueLockId(),
+      pushNotificationEndpoint: 'https://push.bsky.app/api/push',
```
dholms, line 91: this should be some local url that just no-ops - we don't want to push test notifs to our actual notif service

packages/bsky/src/notifications.ts
dholms: I'd really like to eliminate the DB hits here 🤔 

can we just use `handle` instead of `displayName` and thread the record through to this method so we don't have to hit the db?

packages/bsky/src/notifications.ts
```diff
@@ -0,0 +1,239 @@
...
+
+    return userTokens
+  }
+
+  async prepareNotifsToSend(notifications: InsertableNotif[]) {
+    const notifsToSend: PushNotification[] = []
+
+    for (const notif of notifications) {
+      const { did: userDid } = notif
+      const userTokens = await this.getUserTokens(userDid)
```
dholms, line 39: would be cool to get notifs for all relevant users in a single request instead of multiple requests

`.where('did', 'in', dids)`

packages/bsky/src/services/indexing/index.ts
```diff
@@ -93,7 +97,22 @@ export class IndexingService {
...
+          obj,
+          timestamp,
+        )
+        if (insertedRecords) {
+          // SEND NOTIFICATIONS as part of background queue process
+          this.backgroundQueue.add(async () => {
+            await this.notifServer
+              .prepareNotifsToSend(insertedRecords)
+              .then((preparedNotifs) => {
+                this.notifServer.sendPushNotifications(preparedNotifs)
```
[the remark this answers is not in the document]
devinivy: Though if we're making that change, might recommend writing it with awaits which doesn't have the same footgun:
```ts
const preparedNotifs = await this.notifServer.prepareNotifsToSend(insertedRecords)
await this.notifServer.sendPushNotifications(preparedNotifs)
```

packages/bsky/src/notifications.ts
```diff
@@ -0,0 +1,239 @@
...
+        "image": string, (Android only)
+        "expiration": number, (iOS only)
+        "badge": number, (iOS only)
+      }
+    ]
+    3. `gorush` will send notification to APNS or FCM
+    4.  store response from `gorush` which contains the ID of the notification
+    5. If notification needs to be updated or deleted, find the ID of the notification from the database and send a new notification to `gorush` with the ID (repeat step 2)
+  */
+  async sendPushNotifications(notifications: PushNotification[]) {
```
dholms, line 101: are we gonna do any batching or rate limiting? worried this is gonna be super noisy to send a push notification on every action

packages/bsky/src/services/indexing/index.ts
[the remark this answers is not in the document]
ansh: done

packages/bsky/src/services/indexing/index.ts
```diff
@@ -93,7 +97,22 @@ export class IndexingService {
...
-        await indexer.insertRecord(uri, cid, obj, timestamp)
+        const insertedRecords = await indexer.insertRecord(
+          uri,
+          cid,
+          obj,
+          timestamp,
+        )
+        if (insertedRecords) {
+          // SEND NOTIFICATIONS as part of background queue process
+          this.backgroundQueue.add(async () => {
```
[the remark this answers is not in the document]
ansh: done

packages/bsky/src/notifications.ts
ansh: Added some in-memory rate-limiting and batching. However, I think I should utilize redis for that. Let me know if that sounds good.

packages/dev-env/src/bsky.ts
ansh: I asked Jake about this but he said its fine to do so. Let me know if not.

ansh: `gorush` has a binary we can ship with `dev-env` but will probably need @Jacob2161 to set that up

packages/bsky/src/notifications.ts
ansh: I assume by this you mean changing the `getUserTokens` function to have `.where('did', 'in', dids)` so we can grab all the dids we need in one request instead of many?

packages/bsky/src/notifications.ts
ansh: Added Redis too/

packages/bsky/src/notifications.ts
```diff
@@ -0,0 +1,311 @@
...
+        // Set expiration for rate limit data to 25 minutes just in case
+        await this.redis.expire(key, 25 * 60)
+
+        // BATCHING
+        // Add to batch
+        this.notificationBatch.push(notif)
+        // If batch size is 20, add task to background queue
+        if (this.notificationBatch.length >= this.notificationBatchSize) {
+          try {
+            this.backgroundQueue.add(async () => {
```
devinivy, line 136: I think we need this `await` a. in order to catch any error below, and b. in order to maintain `this.notificationBatch`:
```suggestion
            await this.backgroundQueue.add(async () => {
```

Is the purpose of using the background queue here to limit concurrency of sending notifications, or some other reason?

dholms: I think we want to put the entire notification flow on the background queue (out in the indexing service) so that it's non-blocking on write

dholms: similar to labels

ansh: This part doesn't have to be in the background queue, as the resource intensive portion `prepareNotifs` is already in the background queue.

I just put it in the background queue because I thought we wanted to limit concurrency, as you said, as much as possible. But I do not think it is required because the function is already called in a backgroundQueue.

packages/bsky/src/services/indexing/processor.ts
```diff
@@ -230,6 +233,17 @@ export class RecordProcessor<T, S> {
       notifs = this.params.notifsForInsert(op.inserted)
     }
     for (const chunk of chunkArray(notifs, 500)) {
+      if (this.notifServer) {
```
dholms, line 236: seems nicer to do this out in the indexing service instead of within the processor which I view as a "raw record -> DB indexes" translator 🤔 

any reason to prefer it here?

packages/bsky/src/notifications.ts
```diff
@@ -0,0 +1,334 @@
...
+    notifs: InsertableNotif[],
+  ): Promise<NotifDisplay[]> {
+    const { ref } = this.db.db.dynamic
+    const authorDids = notifs.map((n) => n.author)
+    const subjectUris = notifs.flatMap((n) => n.reasonSubject ?? [])
+    const recordUris = notifs.map((n) => n.recordUri)
+    const allUris = [...subjectUris, ...recordUris]
+
+    // gather potential display data for notifications in batch
+    const [authors, posts] = await Promise.all([
```
dholms, line 223: i still wonder if we can avoid most of these db hits by passing the record into the notification service?

i will see - these bulk gets give a decent justification for processing notifs in batches

packages/bsky/src/services/indexing/processor.ts
devinivy: I put it here in order mostly to make use of the existing code paths for notification handling, make it simpler to answer "where are notifications sent?", ensure the same logic was being applied no matter the destination of the notifications.  Previously notifications that may be generated from a deletion weren't being handled, and the added `notifsForInsert()` calls down in `insertRecord()`/`updateRecord()` felt a bit leaky/special-cased to me.

I do see what you're saying, though.  I may just have a slightly different perspective on it—that the destination of the notifications (in the db or off to some push service) isn't all that significant.

My guess is that we could get the best of both worlds with a little refactor!  But now might not be the time for it.  Happy to switch it back to the way it was if you think that's the move 👍

packages/bsky/src/notifications.ts
devinivy: I believe that does help with the post record lookup for mentions and replies, which do show the text from the post that was just indexed.  The author display name lookups for all notifs and post record lookups for likes, quotes, reposts will still need to happen at some point.  The changes I made here tried to batch work together into two queries rather than a few queries per notification.  Batching more notifications together as you're suggesting will probably be what gives the biggest win here.

packages/bsky/src/services/indexing/processor.ts
dholms: nah that's cool I dig the justification 👍 

if we end up maturing this system, that might be a good time to move it around a bit

syui:
I'm getting an error in rust, Not sure if this change is relevant.

```sh
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error("trailing characters")', src/main.rs:3:1
```

> src/main.rs
```rust
let str = at_notify_limit::get_request(20);
let notify: Notify = serde_json::from_str(&str.await).unwrap();
let n = notify.notifications;
let reason = &n[1].reason;
println!("{}", reason);
```

PR #1431, opened 2023-08-03
:sparkles: Allow creating moderation action with a duration

foysalit:
This PR paves the way for creating moderation actions with duration. The current use-case for this is to takedown a post/account temporarily, essentially creating a way to temporarily suspend an account or a post.


lexicons/com/atproto/admin/defs.json
```diff
@@ -17,6 +17,10 @@
       "properties": {
         "id": { "type": "integer" },
         "action": { "type": "ref", "ref": "#actionType" },
+        "actionDurationInHours": {
```
foysalit, line 20: Using `Hours` because realistically, this would be smallest unit of time that we would use for reverting actions

packages/bsky/service/api.js
```diff
@@ -65,12 +66,22 @@ const main = async () => {
   })
   const viewMaintainer = new ViewMaintainer(migrateDb)
   const viewMaintainerRunning = viewMaintainer.run()
+  
+  const periodicModerationActionReversal = new PeriodicModerationActionReversal(
+    db,
```
foysalit, line 71: should we just pass `pds` here and grab the db instance from pds inside the `PeriodicModerationActionReversal` class?

packages/pds/service/index.js
```diff
@@ -84,12 +85,28 @@ const main = async () => {
   })
   const viewMaintainer = new ViewMaintainer(migrateDb)
   const viewMaintainerRunning = viewMaintainer.run()
+
+  const periodicModerationActionReversal = new PeriodicModerationActionReversal(
+    db,
+    pds.ctx,
```
devinivy, line 91: Same here, I'd recommend maybe just passing `pds.ctx`.

packages/pds/service/index.js
```diff
@@ -84,12 +85,28 @@ const main = async () => {
...
   process.on('SIGTERM', async () => {
     await pds.destroy()
+
+    // Gracefully shutdown view-maintainer
     viewMaintainer.destroy()
     await viewMaintainerRunning
+
+    // Gracefully shutdown periodic-moderation-action-reversal
+    periodicModerationActionReversal.destroy()
+    await periodicModerationActionReversalRunning
```
devinivy, line 107: We might consider calling this before `pds.destroy()`.  Since it uses the `db` which is closed by `pds`, it could start to fail if it's stopped before `pds`.

packages/pds/src/db/migrations/20230804T224916736Z-action-duration.ts
```diff
@@ -0,0 +1,27 @@
...
+  await Promise.all([
+    db.schema
+      .alterTable('moderation_action')
+      .addColumn('actionDurationInHours', 'integer')
+      .execute(),
+    db.schema
+      .alterTable('moderation_action')
+      .addColumn('actionExpiresAt', 'varchar')
+      .execute(),
+  ])
```
devinivy, line 13: We tend to run schema updates serially, just to keep things simpler and deterministic.  The amount of time the migration takes to run isn't much of a concern.

devinivy, line 13: How would you feel about dropping the `action*` prefix here?  I think `durationInHours` and `expiresAt` would be clear enough, and maybe a bit more natural to write e.g. `action.expiresAt`.  I don't think we have similar prefixes elsewhere, so it could be good for consistency too.

packages/pds/tests/moderation.test.ts
```diff
@@ -1005,6 +1006,47 @@ describe('moderation', () => {
       await reverse(action.id)
     })
 
+    it('automatically reverses actions marked with duration', async () => {
```
devinivy, line 1009: Nice!

devinivy (commented):
Good stuff, I like how this turned-out!  Just a handful of small questions and comments we can work through.

lexicons/com/atproto/admin/defs.json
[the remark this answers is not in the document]
foysalit: yeah I like the brevity and like you mentioned, the existing `action` naming made me think the prefix would be helpful here.

packages/pds/service/index.js
foysalit: I see so viewmaintainer uses migrateDb which is why we can wait for it to destroy after I guess?

packages/pds/src/db/migrations/20230804T224916736Z-action-duration.ts
foysalit: agreed!

packages/bsky/src/services/moderation/index.ts
```diff
@@ -249,6 +252,8 @@ export class ModerationService {
       reason,
       subject,
       subjectBlobCids,
+      actionDurationInHours,
+      actionExpiresAt,
       createdAt = new Date(),
```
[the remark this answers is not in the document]
foysalit: as in, compute `expiresAt` based on `createdAt + durationInHours`  within this method instead of passing it in from the caller of this method?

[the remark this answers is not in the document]
devinivy: Yeah, that is what I was thinking 👍

packages/bsky/service/api.js
```diff
@@ -65,9 +66,19 @@ const main = async () => {
   })
   const viewMaintainer = new ViewMaintainer(migrateDb)
   const viewMaintainerRunning = viewMaintainer.run()
+
+  const periodicModerationActionReversal = new PeriodicModerationActionReversal(
+    pds.ctx,
```
devinivy, line 71: I think we want this to be:
```suggestion
    bsky.ctx,
```

packages/bsky/src/db/periodic-moderation-action-reversal.ts
```diff
@@ -0,0 +1,78 @@
...
+import Database from '.'
+import AppContext from '../context'
+
+export const MODERATION_ACTION_REVERSAL_ID = 1011
+
+export class PeriodicModerationActionReversal {
+  leader = new Leader(MODERATION_ACTION_REVERSAL_ID, this.db)
+  destroyed = false
+
+  constructor(public db: Database, private appContext: AppContext) {}
```
devinivy, line 13: I think this may have been intended to change to:
```suggestion
  constructor(private appContext: AppContext) {}
```

foysalit: 🤦🏽‍♂️

devinivy (approved):
Nice!

packages/bsky/src/db/periodic-moderation-action-reversal.ts
```diff
@@ -2,24 +2,42 @@ import { wait } from '@atproto/common'
...
       const moderationTxn = this.appContext.services.moderation(dbTxn)
-      await moderationTxn.revertAction({
+      const reverseAction = {
         id,
         createdBy,
         createdAt: new Date(),
         reason: `[SCHEDULED_REVERSAL] Reverting action as originally scheduled`,
-      })
+      }
+      await moderationTxn.revertAction(reverseAction)
```
devinivy, line 35: We can actually get rid of this call here: roundtripping to the PDS will ensure the reversal is synced between the appview and PDS, just as it does when a moderator takes an action in the app.

devinivy: Or really, if there's a `pushAgent` then we use that, otherwise we just make the change locally.  Something like:
```ts
if (this.pushAgent) {
  await this.pushAgent.com.atproto.admin.reverseModerationAction(
    reverseAction,
  )
} else {
  await moderationTxn.revertAction(reverseAction)
}
```

packages/bsky/src/db/periodic-moderation-action-reversal.ts
```diff
@@ -2,24 +2,42 @@ import { wait } from '@atproto/common'
...
 import AppContext from '../context'
+import AtpAgent from '@atproto/api'
+import { buildBasicAuth } from '../auth'
 
 export const MODERATION_ACTION_REVERSAL_ID = 1011
 
 export class PeriodicModerationActionReversal {
   leader = new Leader(MODERATION_ACTION_REVERSAL_ID, this.appContext.db)
   destroyed = false
+  pushAgent: AtpAgent
```
devinivy, line 13:
```suggestion
  pushAgent?: AtpAgent
```

packages/pds/service/index.js
```diff
@@ -89,8 +89,13 @@ const main = async () => {
   const periodicModerationActionReversal = new PeriodicModerationActionReversal(
     pds.ctx,
   )
+
+  // If the PDS is configured to proxy moderation, we don't need to run the automatic reversal
+  // since the app-view will take care of that and let the PDS know when it's done.
   const periodicModerationActionReversalRunning =
-    periodicModerationActionReversal.run()
+    pds.ctx.shouldProxyModeration()
+      ? new Promise((resolve) => resolve())
```
devinivy, line 97: A nice short-form for this:
```suggestion
      ? Promise.resolve()
```

devinivy (approved):

PR #1434, opened 2023-08-03
Hive retry logic & logging

dholms:
Improve retry logic so that we don't hand the Hive API req an empty string.

As well, we log on each hive api failure & we add in backgroundQueue logging as a bonus

### DIFF
--- a/.github/workflows/build-and-push-bsky-aws.yaml
+++ b/.github/workflows/build-and-push-bsky-aws.yaml
@@ -3,7 +3,6 @@ on:
   push:
     branches:
       - main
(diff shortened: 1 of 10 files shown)


devinivy (approved):

bnewbold:
while this is getting touched: we previously discussed logging the entire raw Hive API response, to make it easier to debug unexpected label situations. eg, can grab the response from logs and turn the response in to a new test case when tuning scores-to-label logic.

any change that already got enabled, or could be slipped in to this PR?

dholms:
yup i'll get that in @bnewbold 👍

PR #1436, opened 2023-08-07
Misc perf improvements on actor search, post deletion, actor invite codes

devinivy:
Three separate optimizations here:
 - On actor search we were casting a wide net with strict word similarity using the low bar of `.1` similarity.  This just makes too many results eligible when there's a large dataset.  Now we use non-strict word similarity, which is looser, but up the bar to `.4` similarity.  This will give comparable results but should be faster and less resource intensive, which especially benefits actor typeahead search.
 - Added an index for `postUri` on `feed_item`, which makes post deletions from `feed_item` more palatable.
 - Added an index for `forUser` on `invite_code`, which makes it simple to list all invites for a user, which we commonly compute for `server.getAccountInviteCodes`.


dholms (approved):
awesome 💪

packages/bsky/src/db/migrations/20230807T035309811Z-feed-item-delete-invite-for-user-idx.ts
```diff
@@ -0,0 +1,14 @@
...
+  // supports post deletion
+  await db.schema
+    .createIndex('feed_item_post_uri_idx ')
+    .on('feed_item')
+    .column('postUri')
+    .execute()
+}
+
+export async function down(db: Kysely<unknown>): Promise<void> {
+  await db.schema.dropIndex('feed_item_post_uri_idx ').execute()
```
devinivy, line 13:
```suggestion
  await db.schema
    .createIndex('feed_item_post_uri_idx')
    .on('feed_item')
    .column('postUri')
    .execute()
}

export async function down(db: Kysely<unknown>): Promise<void> {
  await db.schema.dropIndex('feed_item_post_uri_idx').execute()
```

packages/pds/src/db/migrations/20230807T035309811Z-feed-item-delete-invite-for-user-idx.ts
```diff
@@ -0,0 +1,21 @@
...
+  // supports listing user invites
+  await db.schema
+    .createIndex('invite_code_for_user_idx')
+    .on('invite_code')
+    .column('forUser')
+    .execute()
+}
+
+export async function down(db: Kysely<unknown>): Promise<void> {
+  await db.schema.dropIndex('feed_item_post_uri_idx ').execute()
```
devinivy, line 19:
```suggestion
    .createIndex('feed_item_post_uri_idx')
    .on('feed_item')
    .column('postUri')
    .execute()
  // supports listing user invites
  await db.schema
    .createIndex('invite_code_for_user_idx')
    .on('invite_code')
    .column('forUser')
    .execute()
}

export async function down(db: Kysely<unknown>): Promise<void> {
  await db.schema.dropIndex('feed_item_post_uri_idx').execute()
```

PR #1439, opened 2023-08-07
add filter param to getAuthorFeed

estrattonbailey:
Adds a `filter` parameter to the `getAuthorFeed` query so that we can return `posts` (the default, includes replies), `posts_no_replies`, and only `posts_with_media`.

### DIFF
--- a/lexicons/app/bsky/feed/getAuthorFeed.json
+++ b/lexicons/app/bsky/feed/getAuthorFeed.json
@@ -12,7 +12,7 @@
           "actor": {"type": "string", "format": "at-identifier"},
           "limit": {"type": "integer", "minimum": 1, "maximum": 100, "default": 50},
           "cursor": {"type": "string"},
-          "filter": {"type": "string", "enum": ["post_and_replies", "posts", "media"], "default": "posts_and_replies"}
+          "filter": {"type": "string", "enum": ["posts_and_replies", "posts", "media"], "default": "posts_and_replies"}
         }
       },
       "output": {
--- a/lexicons/app/bsky/feed/getAuthorFeed.json
+++ b/lexicons/app/bsky/feed/getAuthorFeed.json
@@ -12,7 +12,7 @@
           "actor": {"type": "string", "format": "at-identifier"},
           "limit": {"type": "integer", "minimum": 1, "maximum": 100, "default": 50},
           "cursor": {"type": "string"},
-          "filter": {"type": "string", "knownValues": ["posts_and_replies", "posts", "media"], "default": "posts_and_replies"}
+          "filter": {"type": "string", "knownValues": ["posts_and_replies", "posts_only", "posts_with_media"], "default": "posts_and_replies"}
         }
       },
       "output": {
--- a/lexicons/app/bsky/feed/getAuthorFeed.json
+++ b/lexicons/app/bsky/feed/getAuthorFeed.json
@@ -12,7 +12,7 @@
           "actor": {"type": "string", "format": "at-identifier"},
           "limit": {"type": "integer", "minimum": 1, "maximum": 100, "default": 50},
           "cursor": {"type": "string"},
-          "filter": {"type": "string", "enum": ["posts", "post_and_replies", "media"], "default": "posts"}
+          "filter": {"type": "string", "enum": ["post_and_replies", "posts", "media"], "default": "posts_and_replies"}
         }
       },
       "output": {
--- a/lexicons/app/bsky/feed/getAuthorFeed.json
+++ b/lexicons/app/bsky/feed/getAuthorFeed.json
@@ -12,7 +12,7 @@
           "actor": {"type": "string", "format": "at-identifier"},
           "limit": {"type": "integer", "minimum": 1, "maximum": 100, "default": 50},
           "cursor": {"type": "string"},
-          "filter": {"type": "string", "knownValues": ["posts_and_replies", "posts_only", "posts_with_media"], "default": "posts_and_replies"}
+          "filter": {"type": "string", "knownValues": ["posts", "posts_no_replies", "posts_with_media"], "default": "posts"}
         }
       },
       "output": {
--- a/lexicons/app/bsky/feed/getAuthorFeed.json
+++ b/lexicons/app/bsky/feed/getAuthorFeed.json
@@ -11,7 +11,8 @@
         "properties": {
           "actor": {"type": "string", "format": "at-identifier"},
           "limit": {"type": "integer", "minimum": 1, "maximum": 100, "default": 50},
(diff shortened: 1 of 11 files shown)


lexicons/app/bsky/feed/getAuthorFeed.json
```diff
@@ -11,7 +11,8 @@
         "properties": {
           "actor": {"type": "string", "format": "at-identifier"},
           "limit": {"type": "integer", "minimum": 1, "maximum": 100, "default": 50},
-          "cursor": {"type": "string"}
+          "cursor": {"type": "string"},
+          "filter": {"type": "string", "enum": ["posts_and_replies", "posts", "media"], "default": "posts_and_replies"}
```
dholms, line 15: I think I'd do `knownValues` here instead of `enum` since it allows us to add more going forward. An `enum` can never change :/

estrattonbailey: Roger roger

packages/api/src/client/lexicons.ts
```diff
@@ -4752,6 +4752,15 @@ export const schemaDict = {
             cursor: {
               type: 'string',
             },
+            filter: {
+              type: 'string',
+              knownValues: [
+                'posts_and_replies',
+                'posts_only',
+                'posts_with_media',
+              ],
```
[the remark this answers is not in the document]
dholms: yup i like it 👍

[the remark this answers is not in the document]
dholms: actually on second through "replies" are technically "posts"

so maybe we change "posts_only" to "posts_no_replies" or "root_posts_only"

packages/bsky/src/api/app/bsky/feed/getAuthorFeed.ts
```diff
@@ -46,10 +46,23 @@ export default function (server: Server, ctx: AppContext) {
...
       }
 
+      // defaults to posts, reposts, and replies
       let feedItemsQb = feedService
         .selectFeedItemQb()
         .where('originatorDid', '=', did)
 
+      if (filter === 'posts_with_media') {
+        // only posts with media
+        feedItemsQb = feedItemsQb.innerJoin(
```
dholms, line 56: first thought was to do a `whereExists` here but I actually think the query planner does the same thing either way & the join actually gives it a bit more flexibility

long-winded way of saying: looks good 😛

packages/bsky/tests/views/author-feed.test.ts
```diff
@@ -232,4 +234,42 @@ describe('pds author feed views', () => {
...
+    ).toBeTruthy()
+
+    const { data: mediaFeed } = await agent.api.app.bsky.feed.getAuthorFeed({
+      actor: carol,
+      filter: 'posts_with_media',
+    })
+
+    expect(
+      mediaFeed.feed.every(({ post }) => {
+        return isEmbedRecordWithMedia(post.record)
```
dholms, line 261: this doesn't seem quite right & I'm honestly not sure why it's not throwing 🤔 

when I try to do the same checkout outside of the `.every` it throws an error

dholms: I think you want to import `isMain` instead of `validateMain` from both embed/images & embed/recordWithMedia & do 

`return isEmbedImages(post.record.embed) || isEmbedRecordWithMedia(post.record.embed)`

packages/bsky/src/api/app/bsky/feed/getAuthorFeed.ts
dholms: oh actually this does need to be a `whereExists` because the inner join is causing duplicate entries for posts with more than one image embedded

dholms:
approach looks solid 👍 few pieces of feedback

packages/bsky/src/api/app/bsky/feed/getAuthorFeed.ts
estrattonbailey: Yeah I only did this because it was shorter. Do you think an exists would make the intention a little more obvious though? I kinda do now that I think about it.

packages/bsky/tests/views/author-feed.test.ts
estrattonbailey: Oh, false positive, was just returning an error, not throwing, good catch

packages/api/src/client/lexicons.ts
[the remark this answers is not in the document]
estrattonbailey: Went with `posts`, `posts_no_replies` and `posts_with_media`

packages/bsky/src/api/app/bsky/feed/getAuthorFeed.ts
estrattonbailey: Changed it to the subquery, I do think it's a little more obvious what's happening for future generations. Cool?

packages/bsky/tests/views/author-feed.test.ts
```diff
@@ -232,4 +235,46 @@ describe('pds author feed views', () => {
...
+
+    const { data: mediaFeed } = await agent.api.app.bsky.feed.getAuthorFeed({
+      actor: carol,
+      filter: 'posts_with_media',
+    })
+
+    expect(
+      mediaFeed.feed.every(({ post }) => {
+        return (
+          isEmbedRecordWithMedia(post.embed) && isImageEmbed(post.embed?.media)
```
dholms, line 265: ah good call on the `&&` here 👍 

and carol doesn't have any posts with just an "image embed", only posts with a "record with media embed"
but just to make the tests a bit more self documenting, I think this should be 

```
(isEmbedRecordWithMedia(post.embed) && isImageEmbed(post.embed?.media)) || isImageEmbed(post.embed)
```

dholms: the embeds logic is a bit heady so lmk if you want me to walk you through it at all

dholms (approved):
sweet this looks great - just the last piece of feedback on the test & this should be good to go

packages/bsky/tests/views/author-feed.test.ts
estrattonbailey: Ok I messed around with some more seed data and I think I grok this. Pry wouldn't hurt to walk through it some time but no rush. Updated this assertion, as well as added an images only post for Carol (and updated the snapshot).

dholms:
think this broke some other snaps but it's on track 👍 
other option is to just test both carol & bobs `getAuthorFeed` filtered by media only - bob has a reply with an image

dholms (approved):
looks good 🙌

Zero3K:
When will the media tab be showing up?

PR #1440, opened 2023-08-07
add getActorLikes

estrattonbailey:
We use the term actor in the errors/params, and since a Like could originate from a bot or human, decided `actor` the better term.


dholms (commented):
yup schema looks right 👍

packages/bsky/src/api/app/bsky/feed/getActorLikes.ts
```diff
@@ -0,0 +1,95 @@
...
+
+      if (viewer !== null) {
+        feedItemsQb = feedItemsQb
+          .where((qb) =>
+            qb.where((qb) =>
+              graphService.whereNotMuted(qb, viewer, [ref('post.creator')]),
+            ),
+          )
+          // TODO do we want this? was missing here
+          .whereNotExists(graphService.blockQb(viewer, [ref('post.creator')]))
```
estrattonbailey, line 64: This clause was present in the PDS but not here. Was there a reason we didn't want to filter out posts from blocked users?

packages/bsky/src/api/app/bsky/feed/getActorLikes.ts
```diff
@@ -0,0 +1,95 @@
...
+        keyset,
+      })
+
+      const [feedItems, repoRev] = await Promise.all([
+        feedItemsQb.execute(),
+        actorService.getRepoRev(viewer),
+      ])
+      setRepoRev(res, repoRev)
+
+      const feed = await feedService.hydrateFeed(feedItems, viewer)
```
estrattonbailey, line 84: This assembles a feed that includes replies/threads. I'm assuming we don't want that for this view. Do we have an existing hydrator for this, or should I just add some options to `hydrateFeed`?

(Same question in PDS)

packages/bsky/src/api/app/bsky/feed/getActorLikes.ts
```diff
@@ -0,0 +1,75 @@
...
+      }
+
+      let feedItemsQb = feedService
+        .selectFeedItemQb()
+        .innerJoin('like', 'like.subject', 'feed_item.uri')
+        .where('like.creator', '=', actorDid)
+
+      if (viewer !== null) {
+        feedItemsQb = feedItemsQb
+          .where((qb) =>
```
dholms, line 39: don't think this nested where clause is necessary - you probably had an OR here at some point or something

packages/bsky/src/api/app/bsky/feed/getActorLikes.ts
dholms: I think this is fine 👍 

It's the same hydration that occurs on `getAuthorFeed`

The "reply" property is a bit misleading, it should really be "replyTo" and just provides some context about the parent post

packages/bsky/src/api/app/bsky/feed/getActorLikes.ts
dholms: (same thing in PDS)

packages/pds/src/app-view/api/app/bsky/feed/getActorLikes.ts
```diff
@@ -0,0 +1,135 @@
...
+import { paginate } from '../../../../../db/pagination'
+import AppContext from '../../../../../context'
+import { FeedRow } from '../../../../services/feed'
+import { OutputSchema } from '../../../../../lexicon/types/app/bsky/feed/getAuthorFeed'
+import { handleReadAfterWrite } from '../util/read-after-write'
+import { authPassthru } from '../../../../../api/com/atproto/admin/util'
+import { LocalRecords } from '../../../../../services/local'
+
+export default function (server: Server, ctx: AppContext) {
+  server.app.bsky.feed.getActorLikes({
```
dholms, line 13: good news: we're incredibly close to not having to re-implement these on the PDS 👀

dholms (commented):
couple small things but overall looking good!

packages/pds/src/app-view/api/app/bsky/feed/getActorLikes.ts
```diff
@@ -0,0 +1,135 @@
...
+              localProf.record,
+            ),
+          },
+        }
+      } else {
+        return item
+      }
+    })
+  }
+  feed = await localSrvc.formatAndInsertPostsInFeed(feed, local.posts)
```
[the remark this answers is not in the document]
estrattonbailey: Cool yeah wasn't sure on this

gildaswise:
Hey, what's the reason for `9b1f5f0` which disallows seeing other profile's likes? Shouldn't that be something each client decides? The statement was "Those likes can still be accessed by APIs, so other apps may show them."

estrattonbailey:
Good question, and totally understand the confusion. The distinction here is that Like _records_ are public i.e. they are publicized via our BGS, but we chose not to provide access to that data from our App View.

Clients can opt to interact with our App View APIs, or choose make their own set of APIs if they want to make different decisions that we have. Some users have already done this but ingesting Like records from our BGS and providing them as custom feeds.

PR #1441, opened 2023-08-07
Include pds account info on mod report and action details

devinivy:
Resolves an issue where moderation report and action details were not reflecting PDS state (e.g. invites and email) when proxying to the appview.


dholms (approved):
yup 👍

PR #1442, opened 2023-08-07
Filter out blocked posts on getPosts

dholms:
Do not return any posts that violate a blocking relationship


packages/bsky/src/api/app/bsky/feed/getPosts.ts
```diff
@@ -18,7 +18,10 @@ export default function (server: Server, ctx: AppContext) {
       const posts: PostView[] = []
       for (const uri of uris) {
         const post = postViews[uri]
-        if (post) {
+        const isBlocked =
+          post.author.viewer?.blockedBy === true ||
+          typeof post.author.viewer?.blocking === 'string'
```
devinivy, line 23: This will fail if `post` doesn't exist.  I think something like this would do:
```suggestion
        const isBlocked =
          post?.author.viewer?.blockedBy === true ||
          typeof post?.author.viewer?.blocking === 'string'
```

PR #1443, opened 2023-08-07
Tighten feed filtering

dholms:
This strengthens the feed cleaning abstraction & adds it to the feed service.

We also rearrange the timeline skeleton code so that we don't write it twice & don't check for mutes/blocks in the main construction query


devinivy (approved):
Killer

PR #1452, opened 2023-08-08
Fix open handles in appview tests

dholms:
Fix open handles in appview tests but waiting for the indexer to be destroyed in dev-env


devinivy (approved):
Solid find!

PR #1455, opened 2023-08-09
lexicons: Add `"minimum": 0` to count integers.

aDotInTheVoid:
There's no case when these count's could be negative. By encoding this in the lexicons, it allows generators for languages with support for unsigned integers to use them here.

### DIFF
--- a/lexicons/app/bsky/actor/defs.json
+++ b/lexicons/app/bsky/actor/defs.json
@@ -65,9 +65,9 @@
         },
         "avatar": { "type": "string" },
         "banner": { "type": "string" },
-        "followersCount": {"type": "integer"},
-        "followsCount": {"type": "integer"},
-        "postsCount": {"type": "integer"},
+        "followersCount": {"type": "integer", "minimum": 0},
+        "followsCount": {"type": "integer", "minimum": 0},
+        "postsCount": {"type": "integer", "minimum": 0},
         "indexedAt": {"type": "string", "format": "datetime"},
         "viewer": {"type": "ref", "ref": "#viewerState"},
         "labels": {
--- a/lexicons/app/bsky/feed/defs.json
+++ b/lexicons/app/bsky/feed/defs.json
@@ -19,9 +19,9 @@
             "app.bsky.embed.recordWithMedia#view"
           ]
         },
-        "replyCount": {"type": "integer"},
-        "repostCount": {"type": "integer"},
-        "likeCount": {"type": "integer"},
+        "replyCount": {"type": "integer", "minimum": 0},
+        "repostCount": {"type": "integer", "minimum": 0},
+        "likeCount": {"type": "integer", "minimum": 0},
         "indexedAt": {"type": "string", "format": "datetime"},
         "viewer": {"type": "ref", "ref": "#viewerState"},
         "labels": {
--- a/packages/api/src/client/lexicons.ts
+++ b/packages/api/src/client/lexicons.ts
@@ -3652,12 +3652,15 @@ export const schemaDict = {
           },
           followersCount: {
             type: 'integer',
+            minimum: 0,
           },
           followsCount: {
             type: 'integer',
+            minimum: 0,
           },
           postsCount: {
             type: 'integer',
+            minimum: 0,
           },
           indexedAt: {
             type: 'string',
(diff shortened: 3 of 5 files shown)


dholms (approved):
Yup this seems reasonable to me. Any issues with this @pfrazee?

pfrazee:
sure no prob w/this

bnewbold (approved):
Seems like a reasonable pattern! There shouldn't ever be negative numbers in these fields.

As noted in the specs (https://atproto.com/specs/data-model#data-types):

> As a best practice to ensure Javascript compatibility with default types, integer should be limited to 53 bits of precision. Note that JSON numbers can have an arbitrary number of digits, but integer is limited to 64 bits even ignoring Javascript.

So i'm assuming you want to map to unsigned integer types in code for reasons other than capacity (maximum value).

aDotInTheVoid:
> So i'm assuming you want to map to unsigned integer types in code for reasons other than capacity (maximum value).

Yeah, it's 50% because it makes the programing more natural in languages like with strict integers like rust (you can't convert a `list.len()` into an `i64`, only `u64`), and 50% because I've had Make Illegal States Unrepresentable drilled into my head so many times that it feels sad to allow a user to have -1 follower.

pfrazee:
@aDotInTheVoid it's JSON so the data type can't be determined by this. The minimum will establish valid values but you should make sure to interpret the input as signed to validate

aDotInTheVoid:
@pfrazee I'm not sure I understand what you mean. In the json library I'm using [`serde_json`](https://docs.rs/serde_json/), you tell it what type you're expecting. This let's you do validation and deserialization at the same time. If the JSON isn't valid, it won't deserialize.

pfrazee:
Valid JSON is going to give you a signed number in all cases. If I give you this JSON:

```json
{
  "postsCount": -1
}
```

That will pass the JSON validation but then fail our record (lexicon) validation. However you need to start by assuming the number is signed

- Read the JSON `postsCount` into a signed number type (probably a float)
- Check if it's >= 0
- *Then* you could potentially put it into an unsigned number type

However if you assume the response is always going to be correct, you'll end up reading -1 into an unsigned type and wind up with MAX_UNSIGNED. It's not the end of the world for this, but it won't be correct

bnewbold:
I think for application-specific, known-schema processing of records, with native types, it is reasonable to use, eg, `serde_json` with a `u64` and have deserialization fail on a negative number (or float) in the input doc. This kind of failure should be record-specific, and not invalidate anything else in the repository.

This is of course different from schema-agnostic processing of records, where you don't have this additional context on hand.

There is also the situation where you might need to round-trip a record, in whole or in part. For example, if you wanted to update a specific field in a record (read-modify-write), your software should probably preserve all other fields as-is, both known and unknown. THis is important both for schema evolution (don't want old software to clobber new optional fields) and possible third party extensions (don't want to clobber off-spec "extra" fields an account may have intentionally added to records). This can be tricky with something like `serde_json` because it will normally discard any unknown fields. And depending on how it is implemented, it might also normalize fields (eg, datetimes with different sub-second precision, lower-casing of fields which are case-insensitive).

There are plenty of totally acceptable use cases for typed serialization (like a client app doing presentation of records, indexing in feed generators, etc), just need to be careful about the use case.

aDotInTheVoid:
> However if you assume the response is always going to be correct, you'll end up reading -1 into an unsigned type and wind up with MAX_UNSIGNED. It's not the end of the world for this, but it won't be correct

Just because you attempt to deserialze into an unsigned type doesn't mean you assume the number is unsigned. Rather you say "if this can't be losslessly deserialized to a u64, return an error. Eg for this case you'd get [``"invalid value: integer `-1`, expected u64"``](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=42a61815a36fb2d7959cc133a29be7e7)

> There is also the situation where you might need to round-trip a record, in whole or in part. 

Yeah, the initial goal is to target client/feedgen like use-cases, where round-tripping isn't done. It adds a load of extra complexity (eg field ordering) that arn't necessary for most use cases.

For the PDS use case where you need to be able to round-trip (and forward fields/lexicons that you arn't aware of), the goal is to offer a different library that stores the extra data, but not make the clients/feedgen pay for this complexity. (These libraries would be related, and share as much as possible)

PR #1456, opened 2023-08-09
✨ Expose takendown profile, their follows and followers to mods

foysalit:
This PR bypasses soft delete checks when profile, their follows and followers are viewed by moderators.


packages/bsky/src/auth.ts
```diff
@@ -30,6 +30,45 @@ export const authOptionalVerifier =
     return authVerifier(idResolver, opts)(reqCtx)
   }
 
+export const authOptionalAccessOrRoleVerifier = (
```
foysalit, line 33: This is basically a ported over version of the existing role/access check from pds but it makes the check optional

packages/bsky/src/api/app/bsky/actor/getProfile.ts
```diff
@@ -5,10 +5,12 @@ import AppContext from '../../../../context'
...
 export default function (server: Server, ctx: AppContext) {
   server.app.bsky.actor.getProfile({
-    auth: ctx.authOptionalVerifier,
+    auth: ctx.authOptionalAccessOrRoleVerifier,
     handler: async ({ auth, params }) => {
       const { actor } = params
-      const requester = auth.credentials.did
+      const requester = 'did' in auth.credentials ? auth.credentials.did : null
+      const canViewTakendownProfile =
+        'triage' in auth.credentials && auth.credentials.triage
```
devinivy, line 13: I think we can use the `type` field on the credentials now for these checks:
```suggestion
      const requester = auth.credentials.type === 'access' ? auth.credentials.did : null
      const canViewTakendownProfile = auth.credentials.type === 'role'
```

packages/bsky/src/auth.ts
```diff
@@ -30,6 +30,45 @@ export const authOptionalVerifier =
...
+        ...result,
+        credentials: {
+          type: 'access' as const,
+          ...result.credentials,
+        },
+      }
+      // If access check fails, we can just return unauthorized credentials since this verifier is optional
+    } catch (err) {
+      return defaultUnAuthorizedCredentials
+    }
```
devinivy, line 68: If it's no big deal, I lean towards allowing this to fail if auth is present but invalid.

packages/bsky/src/auth.ts
```diff
@@ -30,6 +30,45 @@ export const authOptionalVerifier =
...
   }
 
+export const authOptionalAccessOrRoleVerifier = (
+  idResolver: IdResolver,
+  cfg: ServerConfig,
+) => {
+  const verifyAccess = authVerifier(idResolver, { aud: cfg.serverDid })
+  const verifyRole = roleVerifier(cfg)
+  return async (ctx: { req: express.Request; res: express.Response }) => {
+    const defaultUnAuthorizedCredentials = { credentials: { did: null } }
```
devinivy, line 40: We could add a `type` to this so that all the resulting credentials (role, access, and unauthed) can be easily distinguished from each other.

packages/bsky/src/api/app/bsky/actor/getProfile.ts
foysalit: hmm... I was thinking that at some point, we may only want to give access to these content to mods with specific level of permission in which case, we would want to check for specific level of access which is why I used this approach.

devinivy (approved):

PR #1457, opened 2023-08-09
Include limited viewer info on blocked post

dholms:
includes limited viewer info (blocking & blockedBy) on blocked posts. This helps the UI display the correct prompt.

One little point of friction: the `viewer` here is different  from the `viewer` on post (post v actor). Do we want to adjust the property name at all? `viewerActor` feels kinda sloppy


pfrazee:
🙏 muchos gracias this is perf

> One little point of friction: the viewer here is different from the viewer on post (post v actor). Do we want to adjust the property name at all? viewerActor feels kinda sloppy

I'm not overly bothered by this, but one thing we could consider doing is

```
uri: string
blocked: true
author: {
  did: string
  viewer: Viewer
}
```

We know the did already from the URL so it doesn't violate anything, and it resolves this awkwardness.

dholms:
yeah I like that 👍

pfrazee:
do you intend to make author optional? (we can get away with a breaking change here, I think). Otherwise +1

dholms:
good call - I think we want author to be required but viewer to be optional

pfrazee:
Thanks!

PR #1459, opened 2023-08-09
Appview database replicas

dholms:
This provides an application mechanism for managing a primary database and replicas

This centers the concept of a DatabaseCoordinator which manages connections for the primary database and any number of (optionally tagged) replicas.

The coordinator is passed around in the AppContext & each handler can request either the primary or a replica.

When a replica is requested:
- if no tag is specified, we round robin a random untagged replica (or a replica tagged `*`)
- if a tag is specified, we round robin a replica that is tagged with that tag

The intention of requesting a replica is to use it to service an entire request. Therefore with tagging, we may tag replicas in verticals such as `timeline` or `thread`. The intention is not to tag them horizontally (ie as `construction` or `hydration`)


packages/bsky/src/db/coordinator.ts
```diff
@@ -0,0 +1,80 @@
...
+      dbs: [],
+      roundRobinIdx: 0,
+    }
+    for (const cfg of opts.replicas) {
+      const db = new Database({
+        schema: opts.schema,
+        ...cfg,
+      })
+      this.allReplicas.push(db)
+      if (cfg.tag) {
```
devinivy, line 46: If we're going to get the tagging system in, I think it might be worth allowing multiple tags per replica.  Then you can say something like _this replica handles timelines and author feeds_ without also needing to muck around with the application code too much.

packages/bsky/src/db/coordinator.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  getReplica(tag?: string): Database {
+    if (tag && this.tagged[tag]) {
+      this.tagged[tag].roundRobinIdx =
+        (this.tagged[tag].roundRobinIdx + 1) % this.tagged[tag].dbs.length
+      return this.tagged[tag].dbs[this.tagged[tag].roundRobinIdx]
+    } else {
+      this.untagged.roundRobinIdx =
+        (this.untagged.roundRobinIdx + 1) % this.untagged.dbs.length
+      return this.untagged.dbs[this.untagged.roundRobinIdx]
```
devinivy, line 70: I see we already have a fallback for when a tag doesn't exist.  When there are no untagged replicas I think we may want another fallback, just to ensure there's no chance of not having a db to work with.

packages/bsky/src/db/coordinator.ts
```diff
@@ -0,0 +1,80 @@
...
+      }
+    }
+  }
+
+  getPrimary(): PrimaryDatabase {
+    return this.primary
+  }
+
+  getReplica(tag?: string): Database {
+    if (tag && this.tagged[tag]) {
```
devinivy, line 63: We might consider issuing a `log.warn()` the first time the app tries to use a tag that doesn't have a replica.

devinivy (approved):

PR #1460, opened 2023-08-09
Require bluesky-social/atproto for .github/workflows/build-and-push-*

jcsalterego:
The build and push images actions fail for forks which in turn create a lot of failure noise.

This PR runs the actions only for the original `bluesky-social/atproto` repository on the `main` branch.

Do not merge if you wish to build and push main branches from other repositories.

### DIFF
--- a/.github/workflows/build-and-push-bsky-aws.yaml
+++ b/.github/workflows/build-and-push-bsky-aws.yaml
@@ -11,6 +11,7 @@ env:
 
 jobs:
   bsky-container-aws:
+    if: github.repository == 'bluesky-social/atproto'
     runs-on: ubuntu-latest
     permissions:
       contents: read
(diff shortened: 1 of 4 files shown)


jcsalterego:
Test failures appear to be unrelated.

bnewbold (approved):
This looks good to me, seems to be the idiomatic way to accomplish this, and should reduce friction/annoyance for third party devs.

:thinking: we should probably also do this for `social-app`, `indigo`, and `did-method-plc`

cc: @Jacob2161 for a second set of eyes in case there are unintended consequences on our build pipeline.

jcsalterego:
One thought is it still might be useful to build the images (but not push) in all branches except main, so there aren't any surprises after PRs are merged. Those should probably exist as a separate set of workflows (possibly in a separate PR as well).

Jacob2161:
Seems smart, thanks!

dholms:
sweet thanks for this!

PR #1462, opened 2023-08-10
getTimeline and getTimelineSkeleton gain excludePostTypes parameter

jcsalterego:
This is a successor PR to #1400.

`app.bsky.feed.getTimeline` and `app.bsky.unspecced.getTimelineSkeleton` gain an `excludePostTypes` array parameter which can contain any combination of the following strings: `quote`, `reply`, `repost`, defaulting to an empty array (and thus, no exclusions).

This represents the middle ground between adding a parameter that's modeled after the client changes in https://github.com/bluesky-social/social-app/pull/885, while taking into account a world where `getTimelineSkeleton` exists. There would still be client-side filtering needed for the minimum like count for replies filter.

IMHO proposing just a lexicon change wouldn't have been enough to see how this could play out, so the implementation is provided as a useful discussion point.

### DIFF
--- a/lexicons/app/bsky/feed/getTimeline.json
+++ b/lexicons/app/bsky/feed/getTimeline.json
@@ -9,6 +9,11 @@
         "type": "params",
         "properties": {
           "algorithm": {"type": "string"},
+          "excludePostTypes": {
+            "type": "array",
+            "items": {"type": "string"},
+            "description": "Exclude post types: quote, reply, repost"
+          },
           "limit": {"type": "integer", "minimum": 1, "maximum": 100, "default": 50},
           "cursor": {"type": "string"}
         }
--- a/packages/bsky/src/api/app/bsky/feed/getTimeline.ts
+++ b/packages/bsky/src/api/app/bsky/feed/getTimeline.ts
@@ -5,19 +5,27 @@ import { paginate } from '../../../../db/pagination'
 import AppContext from '../../../../context'
 import Database from '../../../../db'
 import { SkeletonFeedPost } from '../../../../lexicon/types/app/bsky/feed/defs'
+import { isView } from '../../../../lexicon/types/app/bsky/embed/record'
+import { PostTypeSet } from '../util/postTypeSet'
 
 export default function (server: Server, ctx: AppContext) {
   server.app.bsky.feed.getTimeline({
     auth: ctx.authVerifier,
     handler: async ({ params, auth }) => {
-      const { algorithm, limit, cursor } = params
+      const { algorithm, excludePostTypes, limit, cursor } = params
       const viewer = auth.credentials.did
 
       if (algorithm && algorithm !== FeedAlgorithm.ReverseChronological) {
         throw new InvalidRequestError(`Unsupported algorithm: ${algorithm}`)
       }
 
-      const skeleton = await getTimelineSkeleton(ctx.db, viewer, limit, cursor)
+      const skeleton = await getTimelineSkeleton(
+        ctx.db,
+        viewer,
+        limit,
+        excludePostTypes,
+        cursor,
+      )
 
       const feedService = ctx.services.feed(ctx.db)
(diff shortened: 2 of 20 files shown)


lexicons/app/bsky/feed/getTimeline.json
```diff
@@ -9,6 +9,11 @@
         "type": "params",
         "properties": {
           "algorithm": {"type": "string"},
+          "excludePostTypes": {
+            "type": "array",
+            "items": {"type": "string"},
+            "description": "Exclude post types: quote, reply, repost"
```
jcsalterego, line 15: Current codegen doesn't support `default: []` here with array, AFAICT.

lexicons/app/bsky/feed/getTimeline.json
```diff
@@ -9,6 +9,11 @@
         "type": "params",
         "properties": {
           "algorithm": {"type": "string"},
+          "excludePostTypes": {
+            "type": "array",
+            "items": {"type": "string"},
```
jcsalterego, line 14: Current codegen doesn't support `knownValues` here, for a nested `string[]`. But, it might have been a skill issue.

packages/bsky/src/api/app/bsky/feed/getTimeline.ts
```diff
@@ -5,27 +5,40 @@ import { paginate } from '../../../../db/pagination'
...
       const feedService = ctx.services.feed(ctx.db)
       const feedItems = await feedService.cleanFeedSkeleton(
         skeleton.feed,
         limit,
         viewer,
       )
-      const feed = await feedService.hydrateFeed(feedItems, viewer)
+      let feed = await feedService.hydrateFeed(feedItems, viewer)
+
+      const excludePostTypeSet = new PostTypeSet(excludePostTypes)
```
jcsalterego, line 38: Had to do quote exclusion after hydration as I could not figure out how to add it to the query builder.

packages/bsky/src/api/app/bsky/util/postTypeSet.ts
```diff
@@ -0,0 +1,37 @@
+enum PostType {
+  QUOTE = 'quote',
+  REPLY = 'reply',
+  REPOST = 'repost',
+}
+
+export class PostTypeSet {
+  static readonly POST_TYPE_VALUES = Object.values(PostType).map((v) =>
```
jcsalterego, line 8: My Java might be showing. Not sure how this is done canonically in TypeScript.

packages/pds/src/app-view/api/app/bsky/util/postTypeSet.ts
```diff
@@ -0,0 +1,37 @@
+enum PostType {
```
jcsalterego, line 1: This file & class exist in two packages and perhaps that means it needs to be in one upstream package.

packages/pds/tests/proxied/timeline-skeleton.test.ts
```diff
@@ -69,6 +69,18 @@ describe('proxies timeline skeleton', () => {
...
+  it('timeline skeleton construction with excludePostTypes', async () => {
+    const res = await agent.api.app.bsky.feed.getTimeline(
+      {
+        excludePostTypes: ['reply', 'repost', 'query'],
+      },
+      {
+        headers: { ...sc.getHeaders(alice), 'x-appview-proxy': 'true' },
+      },
+    )
+    expect(forSnapshot(res.data)).toMatchSnapshot()
```
jcsalterego, line 81: There's a slight chance my local snapshot generation is weird around labels. I'm not sure.

jcsalterego (commented):
Some misc notes.

jcsalterego:
Test failure appears to be unrelated:
```
FAIL Bsky App View tests/handle-invalidation.test.ts (5.509 s)
```

PR #1464, opened 2023-08-11
`@atproto/syntax`: merge of `@atproto/identifier`, `uri`, and `nsid`

bnewbold:
A bit of tidying towards keeping SDKs simple and consistent across languages. I can't remember who I discussed this with ahead of time, but I think Devin+Daniel.

Keeps the old deprecated packages around via transcluding the relevant exports from `@atproto/syntax`, to minimize potential for breakage, while we deprecate those packages on npm.

Updated all the packages in this repo to import from `@atproto/syntax`; will need to do a pass on our other repos after a package release gets cut.

Also want to iterate on READMEs, docs, etc, but this PR touches a ton of files already.


dholms (approved):
This all looks great to me!

One bikeshed: I don't love the package name `identifier` (even tho I was the one that named it that 😛). It's a bit clunky & longwinded to me. It's also a bit to similarly named to our other package `identity`

Any ideas for something else?

The only things that come to mind are:
- roll these into `@atproto/common` 
- maybe something like `@atproto/strings` or `@atproto/formats`

bnewbold:
Yeah, I also don't really like `identifier`, mostly because of the common `identi*` prefix with `identity`, and I do think that `identity` is a good name for the other package.

What I picked for the golang SDK is `syntax`, and the scope is simple verification of identifiers and other formats against the identifier *syntax* rules and other Lexicon string formats. That mostly aligns with now `@atproto/identifier` works. Here is the PR for that golang work:

https://github.com/bluesky-social/indigo/pull/276

and my current overall thinking for the golang SDK shape: https://github.com/bluesky-social/indigo/pull/266

There is not a `common` in the golang SDK to roll things up in. This package will be a dependency of almost all other packages, so I wanted to keep it very small and stdlib-only.

dholms:
cool cool - I like `syntax` well enough 👍 

let's just roll with that to keep the two SDKs lined up

bnewbold:
cool so just to confirm:

* will move actual code from `@atproto/identifier` to `@atproto/syntax`
* will have all of `@atproto/identifier`, `@atproto/nsid`, and `@atproto/uri` be temporary shims that just transclude from `@atproto/syntax`, as a deprecation/migration step
* will update all the code in this repo to import from `@atproto/syntax` instead of the others

then after enough time has passed to update third party code (and our own code in other repos), can actually drop the old packages. This is a bit of extra ceremony over just dropping the old package names immediately, but not much, I lean towards this gradual approach.

bnewbold:
Rebased on main and switched to `@atproto/syntax`, with the others transcluding from there.

This PR is a giant magnet for merge conflicts. After merging, will need to publish the `@atproto/syntax` package to npm or downstream things will break (I don't have perms/skills to do this).

Plan is to iterate on READMEs for things like `@atproto/syntax` in the near future.

dholms (approved):

Issue #1473, opened 2023-08-14
AppView proxy can mark handles as "invalid" despite DNS records pointing correctly to user's DID


94tx:
**Describe the bug**
When running with the experimental AppView proxy enabled, some users will appear with an invalid handle, despite the DNS records for their handles pointing correctly to the user's DID:

As an example, with user `@wolf.observer` (`did:plc:o7kc3e22v7cxms67f4gbpryi`):

![image](https://github.com/bluesky-social/atproto/assets/113658863/e5861aef-2f12-4a4c-b719-7b7ca418c0f2)

![image](https://github.com/bluesky-social/atproto/assets/113658863/31a69555-afb8-4efb-a954-defd7f51a4fb)

**To Reproduce**

I do not have any steps to reproduce it as the bug is largely found on a case-by-case basis.

**Expected behavior**

One would expect the AppView proxy to not mark their handles as invalid as the DNS records point correctly to the users' DIDs

dholms:
Is this your account? A couple things I'm curious about: Was the handle incorrect and then fixed today? Were you changed DNS records at all today? 

We revalidate handles every 24 hours and any event you do (like/post/etc) should trigger that. Our application code is saying that this handle is valid so I'm a bit surprised to see it marked as invalid unless the DNS records were changing today.

That being said, I think we'll get 2 fixes through to get folks back on the right track more easily:
- revalidate handles more often than once every 24 hrs when they have an invalid handle (to get them back on track faster)
- make updateHandle idempotent _except_ it pushes out an event on the firehose. this would give users an in-app mechnism to prod the appview (or any downstream service) to revalidate their handle

94tx:
Not my account, but from what I know, the owner hasn't changed their handle lately, and has been liking/posting while having their handle appear as invalid

PR #1474, opened 2023-08-14
Handle revalidation

dholms:
Makes it easier to get back on track if your account has an "invalid handle"
- makes update handle idempotent - except for publishing a handle update event - as a way to prompt downstream services to revalidate your handle
- the same on the admin route so that we can aid users in getting back on track
- while we generally revalidate handles every 24hrs, revalidate "invalid" handles every hour to more quickly get users back on track

Closes https://github.com/bluesky-social/atproto/issues/1473

### DIFF
--- a/packages/pds/src/api/com/atproto/identity/updateHandle.ts
+++ b/packages/pds/src/api/com/atproto/identity/updateHandle.ts
@@ -27,11 +27,7 @@ export default function (server: Server, ctx: AppContext) {
 
       let seqHandleTok: HandleSequenceToken
       if (handleDid) {
-        if (handleDid !== requester) {
-          throw new InvalidRequestError(`Handle already taken: ${handle}`)
-        } else {
-          seqHandleTok = { did: requester, handle: handle }
-        }
+        seqHandleTok = { did: requester, handle: handle }
       } else {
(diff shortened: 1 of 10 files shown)


packages/pds/src/api/com/atproto/identity/updateHandle.ts
```diff
@@ -20,28 +20,39 @@ export default function (server: Server, ctx: AppContext) {
...
-            .updateHandle(requester, handle)
-        } catch (err) {
-          if (err instanceof UserAlreadyExistsError) {
-            throw new InvalidRequestError(`Handle already taken: ${handle}`)
-          }
-          throw err
+      let seqHandleTok: HandleSequenceToken
+      if (handleDid) {
+        if (handleDid !== requester) {
+          throw new InvalidRequestError(`Handle already taken: ${handle}`)
```
estrattonbailey, line 31: This is the same check as above, right?

estrattonbailey (approved):
LGTM

packages/pds/src/api/com/atproto/identity/updateHandle.ts
dholms: oh oops - i meant to only have one of these

PR #1475, opened 2023-08-15
identity: parse support for newer DID document format (Multikey)

bnewbold:
Adds support for parsing DID Documents with newer changes:

- `Multikey` `verificationMethod` type, in which case the key type is inferred from `publicKeyMultibase`, which includes multicodec prefix (like `did:key` does)
- support for `verificationMethod` `id` being bare `#atproto` or full DID URL plus fragement (eg, `did:plc:yk4dd2qkboz2yv6tpubpc6co#atproto`)

Implementation is a bit weird here, hacking in use of `did:key` for parsing. We might want to expose a raw "parseMulticodecMultibase" or something in the crypto library.

cc: @dholms

### DIFF
--- a/packages/identity/src/did/atproto-data.ts
+++ b/packages/identity/src/did/atproto-data.ts
@@ -10,7 +10,7 @@ export const getDid = (doc: DidDocument): string => {
 }
 
 export const getKey = (doc: DidDocument): string | undefined => {
-  let did = getDid(doc)
+  const did = getDid(doc)
   let keys = doc.verificationMethod
   if (!keys) return undefined
   if (typeof keys !== 'object') return undefined
--- a/packages/identity/src/did/atproto-data.ts
+++ b/packages/identity/src/did/atproto-data.ts
@@ -32,8 +32,7 @@ export const getKey = (doc: DidDocument): string | undefined => {
   } else if (found.type === 'EcdsaSecp256k1VerificationKey2019') {
     didKey = crypto.formatDidKey(crypto.SECP256K1_JWT_ALG, keyBytes)
   } else if (found.type === 'Multikey') {
-    const foundKey = `did:key:${found.publicKeyMultibase}`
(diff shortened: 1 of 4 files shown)


dholms:
> We might want to expose a raw "parseMulticodecMultibase" or something in the crypto library.

Yup I support that 👍 

One gotcha is that `did:key` uses a multibase prefix but still requires base58. My understanding is that Multikey allows for other bases?

Besides that this looks great ✨

packages/identity/src/did/atproto-data.ts
```diff
@@ -10,13 +10,16 @@ export const getDid = (doc: DidDocument): string => {
...
+  const did = getDid(doc)
   let keys = doc.verificationMethod
   if (!keys) return undefined
   if (typeof keys !== 'object') return undefined
   if (!Array.isArray(keys)) {
     keys = [keys]
   }
-  const found = keys.find((key) => key.id === '#atproto')
+  const found = keys.find(
+    (key) => key.id === '#atproto' || key.id === `${did}#atproto`,
```
dholms, line 21: we may want to update this check for all ids - including the ids found in `services`

bnewbold: make sense, added. the service `id` fields don't *need* to have the full DID included (unlike for `verificationMethod`), but it is *allowed* by the spec, so we should parse them correctly.

dholms:
pushed a small tweak to factor out the parse/format multikey methods

once CI passes, we can ship this 👌

bnewbold:
Forgot that this hadn't gotten merged!

There was a conflict; I rebased on current `main` and resolved it. I'm inclined to merge this ASAP, but won't do that myself as I don't want to create any conflicts or operational issues with other code going out.

Also need to publish and updated version of the `@atproto/identity` package in npm, which I don't have perms (or skills) to do.

dholms (approved):

PR #1481, opened 2023-08-16
Handle db client errors on appview

devinivy:
Avoids potential crash on certain kinds of connection errors on appview.


packages/bsky/src/db/db.ts
```diff
@@ -83,3 +85,5 @@ export class Database {
 }
 
 export default Database
+
+const onClientError = (err: Error) => dbLogger.error({ err }, 'db client error')
```
dholms, line 89: should we close the connection here? 🤔

dholms (approved):
yup yup 👍

PR #1487, opened 2023-08-17
include reposted replies in posts_no_replies filter

estrattonbailey:
Fixes https://github.com/bluesky-social/social-app/issues/1197

Ensures that if a user reposts a reply, that repost also shows up in their Posts/Posts & Replies tabs on their profile.


packages/pds/tests/views/author-feed.test.ts
```diff
@@ -334,14 +334,31 @@ describe('pds author feed views', () => {
...
+      }),
+    ).toBeTruthy()
+
+    const { data: danFeed } = await agent.api.app.bsky.feed.getAuthorFeed(
+      { actor: dan, filter: 'posts_no_replies' },
+      { headers: sc.getHeaders(alice) },
+    )
+
+    expect(
+      danFeed.feed.every(({ post }) => {
```
estrattonbailey, line 357: Only problem here is that our seeds don't currently have any reposted replies, so these tests aren't actually confirming this works.

If y'all are cool with it I can add a reply repost and update snaps everywhere.

packages/pds/src/app-view/api/app/bsky/feed/getAuthorFeed.ts
```diff
@@ -64,8 +63,15 @@ export default function (server: Server, ctx: AppContext) {
...
-        feedItemsQb = feedItemsQb.where('post.replyParent', 'is', null)
+        feedItemsQb = feedItemsQb
+          // only posts, no replies
+          .where('post.replyParent', 'is', null)
+          // or any reposted replies
+          .orWhere((qb) => {
+            return qb
+              .where('originatorDid', '=', actorDid)
+              .where('type', '=', 'repost')
+          })
```
estrattonbailey, line 74: These `where` clauses are cumulative, so could make this kinda brittle if someone needs to edit in the future. Is there a better way to structure this to be more clear to future generations?

packages/pds/src/app-view/api/app/bsky/feed/getAuthorFeed.ts
```diff
@@ -64,8 +63,15 @@ export default function (server: Server, ctx: AppContext) {
...
       } else if (params.filter === 'posts_no_replies') {
-        // only posts, no replies
-        feedItemsQb = feedItemsQb.where('post.replyParent', 'is', null)
+        feedItemsQb = feedItemsQb
+          // only posts, no replies
+          .where('post.replyParent', 'is', null)
+          // or any reposted replies
+          .orWhere((qb) => {
+            return qb
+              .where('originatorDid', '=', actorDid)
```
dholms, line 72: i actually don't think this line is necessary, because we're already only selecting items originating from the actor

packages/pds/src/app-view/api/app/bsky/feed/getAuthorFeed.ts
dholms: ugh yeah I think these have to be grouped together under a single `where`

(a new kysely version has a better api around wheres btw, could be worth upgrading soon)

but it'll be simpler to express if we remove the line i mentioned above:
```ts
.where((qb) => 
  qb.where('post.replyParent', 'is', null)
  .orWhere('type', '=', 'repost')
)
```

packages/pds/tests/views/author-feed.test.ts
dholms: that sounds good 👍 

probably a useful thing to have in our snaps anyways

packages/pds/src/app-view/api/app/bsky/feed/getAuthorFeed.ts
estrattonbailey: Oh ok yeah I see the distinction there, good call

packages/bsky/src/api/app/bsky/feed/getAuthorFeed.ts
```diff
@@ -59,8 +59,11 @@ export default function (server: Server, ctx: AppContext) {
               .whereRef('post_embed_image.postUri', '=', 'feed_item.postUri'),
           )
       } else if (filter === 'posts_no_replies') {
-        // only posts, no replies
-        feedItemsQb = feedItemsQb.where('post.replyParent', 'is', null)
+        feedItemsQb = feedItemsQb.where((qb) => {
```
dholms, line 62: such a nit, but for these internal qb, we usually leave out the brackets & the return

dholms (approved):

PR #1491, opened 2023-08-18
Runtime flags in PDS, appview-proxy flags

devinivy:
This implements a simple little system for runtime flags in the pds.  Flags are all pulled and cached every 5 seconds for in-mem lookups.

Also implements one set of flags to support appview proxy switchover.  Flag names take the form `appview-proxy:app.bsky.feed.getTimeline` including the endpoint nsid.  The values are integers from `0` to `10` inclusive, which controls how much traffic (`0` none -> `10` all) is proxied.  The assignments are by requester did, which are each assigned a number from `0` to `9`.  A given did is proxied if its assigned number is less than the flag value for the given endpoint.

### DIFF
--- a/packages/pds/package.json
+++ b/packages/pds/package.json
@@ -56,6 +56,7 @@
     "http-terminator": "^3.2.0",
     "jsonwebtoken": "^8.5.1",
     "kysely": "^0.22.0",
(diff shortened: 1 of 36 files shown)


dholms (approved):
looks great 👌

PR #1496, opened 2023-08-21
Revert labels on takedown reversal

foysalit:


packages/bsky/tests/moderation.test.ts
```diff
@@ -1003,6 +1004,53 @@ describe('moderation', () => {
         'Must be a full moderator to perform an account takedown',
       )
     })
+    it('automatically reverses actions marked with duration', async () => {
```
foysalit, line 1007: Mostly just copied over test from pds package.

packages/bsky/src/db/periodic-moderation-action-reversal.ts
```diff
@@ -26,12 +28,34 @@ export class PeriodicModerationActionReversal {
...
+        ? actionRow.negateLabelVals.split(' ')
+        : undefined,
+      negate: actionRow.createLabelVals
+        ? actionRow.createLabelVals.split(' ')
+        : undefined,
+    })
+  }
+
+  async revertAction(actionRow: ModerationActionRow) {
     return this.appContext.db.getPrimary().transaction(async (dbTxn) => {
```
devinivy, line 54: While we're in here could we also move the `pushAgent` usage outside of the db transaction?  The issue is relatively small—it just takes-up a database connection that doesn't get used for the duration of `pushAgent.com.atproto.admin.reverseModerationAction()`.

devinivy (approved):
Added a comment for a nice-to-have, but the changes here look good 👍

PR #1497, opened 2023-08-21
Fix runtime flags open handles

dholms:
Fixes some open handles that were introduced in runtime flags


devinivy (approved):
Thanks for cleaning-up my slop 🙏

PR #1498, opened 2023-08-21
make sure to await rejected promise expectations

estrattonbailey:
Just noticed I had missed these — tests were passing but we'd probably see some unexpected results here at some point.


devinivy (approved):
Good call!

PR #1499, opened 2023-08-21
Don't notify user of self-likes and self-reposts

estrattonbailey:
Should fix https://github.com/bluesky-social/social-app/issues/1201

Filters out notifications for Likes and Reposts if the subject is the same as the content creator.


dholms (approved):
looks good 👍

PR #1503, opened 2023-08-22
com.atproto.sync.subscribeRepos lexicon: add #repoOp description

snarfed:
specifically, note that it's null for deletes. for https://github.com/bluesky-social/atproto/discussions/1252 , cc @DavidBuchanan314

thanks in advance!


dholms (approved):
yup thanks for the clarification 👍

PR #1504, opened 2023-08-22
Indexer request reprocess

dholms:
Provides a simple http endpoint on indexers through which you can request that a given repo be reprocessed


devinivy (approved):
Excellent 👍

PR #1507, opened 2023-08-23
Improve matching in results from `searchRepos`

foysalit:
This PR ports over existing search queries for matching profiles by handle and `displayName` into the `searchRepos` endpoint in order to provide better results when the search term is not a did but rather a plain string.


packages/bsky/src/services/actor/index.ts
```diff
@@ -77,33 +83,57 @@ export class ActorService {
...
+      // Matching profiles based on display name
+      const profilesQb = getMatchingProfilesQb(this.db, { term })
+        .orderBy('distance', 'asc')
+        .limit(limit)
+      // Combine and paginate result set
+      builder = combineAccountsAndProfilesQb(
+        this.db,
+        accountsQb,
+        profilesQb,
+      ).select('distance')
```
devinivy, line 120: I think this will have a pagination issue: since `accountQb` and `profilesQb` apply a limit but not any kind of offset, we wont be able to get deeper pages of results.  If you check out the `getUserSearchQuery()` utility, you can see how we handled this in `app.bsky.actors.searchActors`.  I wonder if we may even be able to use `getUserSearchQuery()` directly here, though!

foysalit: I see what you mean.
Adjusted the implementation to re-use `getUserSearchQuery` and a test to assert that pagination works.

PR #1511, opened 2023-08-24
:bug: Add labels caching as a config for pds and enable it when running dev-env

foysalit:


packages/dev-env/src/network-no-appview.ts
```diff
@@ -21,6 +21,7 @@ export class TestNetworkNoAppView {
       dbPostgresUrl,
       dbPostgresSchema,
       plcUrl: plc.url,
+      enableLabelsCache: true,
```
devinivy, line 24: I think we want to set this param in `src/bin.ts` and `src/bin-network.ts` rather than in `src/network.ts` and `src/network-no-appview.ts`.  That will allow us to enable the labels when running dev-env via CLI, but not when running it in the tests.

devinivy (approved):
Nice, thanks for cleaning that up!

PR #1512, opened 2023-08-24
Notifications improvements

pfrazee:
Trying some improvements to notifications

- Use the collapse key/id so the OS can group together similar kinds of notifications
- Possible fix to the repeat notif for replies


dholms (approved):
:shipit:

PR #1513, opened 2023-08-24
Add Changesets

estrattonbailey:
This PR adds [Changesets](https://github.com/changesets/changesets) into the mix. Changesets handles versioning, publishing, the creation of `CHANGELOG.md` documents, git tags and GitHub releases.

The basic process is this:
1. You edit some files
2. If these file edits should result in a version change to the package, add a changeset **along with the other files in the commit**
    1. You can add as many changesets (for as many packages as needed) to a PR as you want, [they will be flattened](https://github.com/changesets/changesets/blob/main/docs/decisions.md#how-changesets-are-combined)
3. Commit your edits and changeset(s)
4. Push up your PR

This PR contains an example changeset atm.

From this point:

5. You would merge your PR
6. Changesets gathers all `*.md` files from `.changeset` dir, and creates and maintains a separate PR with the version bumps required
7. Other PRs with other changesets may be merged in the meantime
8. When ready to publish, we merge the Changesets maintained PR
9. Changesets publishes all updated files to npm
10. Changesets updates changelog files and tags
11. Changesets commits changelogs, tags, and a release back to the repo

**TODO**
- [x] don't build images on merge to main
- [x] add `NPM_TOKEN` to this repo's secrets
- [x] confirm there's nothing we need to do to ensure changesets starts at our current versions (we have no git tag history or releases rn)
- [ ] format changelogs after versioning (optional)
- [x] update-main-to-dist

PR #1514, opened 2023-08-24
Clean up prettier/eslint scripts

estrattonbailey:
This PR hoists prettier/eslint scripts to root `package.json` and removes scripts and config from individual `packages` in order to simplify things a bit. It also makes sure that all packages are linted (api wasn't).

Main motivation here originally was the Changesets PR #1513 where I'd like to run prettier over changelogs as they're generated. This makes that a little cleaner/easier.

prettier go brrrr


packages/api/docs/labels.md
```diff
@@ -1,44 +1,44 @@
 <!-- this doc is generated by ./scripts/docs/labels.mjs -->
 
-  # Labels
-  
-  This document is a reference for the labels used in the SDK.
+# Labels
```
estrattonbailey, line 3: Kinda double work, but seems OK to me.
1. Codegen runs
2. On/before commit, run prettier to tidy it up

ericvolp12:
If you're gonna do this, make sure to put it in a precommit hook or something so it stays clean.

estrattonbailey:
> If you're gonna do this, make sure to put it in a precommit hook or something so it stays clean.

Oh yes 😏

packages/api/docs/labels.md
dholms: it'd be nice if codegen genned code that passed prettier 😛

dholms:
I'm down for this 👍 

Altho I'm tempted to just exclude genned code for now

I'm not particularly fond of linting running on commit :/

But I'm down to make the CI linting check (`verify`) lint the entire repo, including schemas/codegen

bnewbold:
Seems nice!

I would run prettier as part of the codegen process itself, specifically over output code.

Would be great to have the Makefile invoke this for `make fmt` / `make lint`.

Like dholms I prefer to check in CI and not do linting in a pre-commit hook, but i'm sort of guest in this particular repo so :shrug:. It does look like we don't actually enforce a lint in CI on this repo though!

dholms:
oh yeah doing it in the codegen step makes a lot of sense 👍

packages/api/docs/labels.md
estrattonbailey: FWIW codegen does do some prettier, but apparently not well enough.

My feeling is that if it's code we check in and/or ship to npm, it should be formatted. We can improve the codegen CLI to handle this, but for now it doesn't hurt to do it here.

.eslintignore
```diff
@@ -0,0 +1,3 @@
+packages/api/src/client
+packages/bsky/src/lexicon
+packages/pds/src/lexicon
```
estrattonbailey, line 3: Copied from the individual packages (except API which wasn't linted)

.eslintrc
```diff
@@ -26,7 +26,11 @@
   "rules": {
     "no-var": "error",
     "prefer-const": "warn",
-    "@typescript-eslint/no-unused-vars": ["warn", { "argsIgnorePattern": "^_" }],
+    "no-misleading-character-class": "warn",
```
estrattonbailey, line 29: There was a regex in the rich text code that this didn't like, but didn't seem worth fixing atm.

package.json
```diff
@@ -9,8 +9,10 @@
     "node": ">=18"
   },
   "scripts": {
-    "verify": "pnpm -r --stream lint && pnpm -r --stream prettier",
-    "format": "pnpm -r --stream prettier:fix",
+    "lint:fix": "pnpm lint --fix",
+    "lint": "eslint . --ext .ts,.tsx",
+    "verify": "prettier --check . && pnpm lint",
+    "format": "prettier --write .",
```
estrattonbailey, line 15: All run from the root now instead of the individual packages

packages/aws/package.json
```diff
@@ -13,12 +13,6 @@
     "directory": "packages/aws"
   },
   "scripts": {
-    "prettier": "prettier --check src/",
-    "prettier:fix": "prettier --write src/",
-    "lint": "eslint . --ext .ts,.tsx",
-    "lint:fix": "pnpm lint --fix",
-    "verify": "run-p prettier lint",
-    "verify:fix": "pnpm prettier:fix && pnpm lint:fix",
```
estrattonbailey, line 21: Was able to remove a bunch of scripts in the packages since everything is run from the root now

PR #1515, opened 2023-08-24
Add missing labels that are available on the moderation tool

foysalit:


devinivy:
@pfrazee mind taking a quick peek at this when you find a moment?

pfrazee (commented):
+1 to the change but you're modifying generated code. You need to place these definitions in the `.json` files under `./definitions` and then run `yarn codegen`

foysalit:
> +1 to the change but you're modifying generated code. You need to place these definitions in the `.json` files under `./definitions` and then run `yarn codegen`

definition of 🤦🏽‍♂️

Moved everything over to the json definition now @pfrazee

pfrazee (approved):
All good! Looks good tme

PR #1518, opened 2023-08-25
Speed-up a few hot queries on pds and bsky

devinivy:
A few little perf-oriented query touchups here.

 - Add index to `blob.tempKey` which is in a hot path or two on the PDS.
 - When counting unread notifications, ensure we can use the `(did, sortAt)` index to only inspect unread notifs for the user.  We were missing it on the appview due to the `lastSeenNotifs is null or ...` check which wasn't allowing us to use the second part of the index.
 - The most mysterious one is that on account lookups by handle on the PDS, postgres was opting to use the handle's trigram index rather than its much simpler/faster uniqueness index.  I compromised on getting postgres to use the lowercase-handle index.


dholms (approved):
incredible 👍

Issue #1523, opened 2023-08-27
`make run-dev-env` is failed by sqlite syntax error


KingYoSun:
Exec `make run-dev-env`, then it is failed by sqlite syntax error.

```
[ created by Bluesky ]

/home/kingyosun/atproto/node_modules/better-sqlite3/lib/methods/wrappers.js:5
        return this[cppdb].prepare(sql, this, false);
                           ^
SqliteError: near "include": syntax error
    at Database.prepare (/home/kingyosun/atproto/node_modules/better-sqlite3/lib/methods/wrappers.js:5:21)
    at SqliteConnection.executeQuery (/home/kingyosun/atproto/packages/dev-env/dist/bin.js:241985:27)
    at /home/kingyosun/atproto/packages/dev-env/dist/bin.js:237602:39
    at #run (/home/kingyosun/atproto/packages/dev-env/dist/bin.js:240554:18)
    at SingleConnectionProvider.provideConnection (/home/kingyosun/atproto/packages/dev-env/dist/bin.js:240545:30)
    at DefaultQueryExecutor.provideConnection (/home/kingyosun/atproto/packages/dev-env/dist/bin.js:240415:37)
    at DefaultQueryExecutor.executeQuery (/home/kingyosun/atproto/packages/dev-env/dist/bin.js:237601:23)
    at RawBuilder.execute (/home/kingyosun/atproto/packages/dev-env/dist/bin.js:237683:21)
    at Object.up61 (/home/kingyosun/atproto/packages/dev-env/dist/bin.js:268678:96)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
  code: 'SQLITE_ERROR'
}
```

Is is probably caused by this line. `create index ... include` can't exec by sqlite.

https://github.com/bluesky-social/atproto/blob/895a21fc25be29c6bd5e84281878e18808b68859/packages/pds/src/db/migrations/20230824T182048120Z-remove-post-hierarchy.ts#L7

little-buddy:
same issue.

little-buddy:
Just delete this line. 
Depending on the context, 
up means clean and down means create, 
so it should be an accident that this line appears here.

PR #1526, opened 2023-08-28
Fix broken sqlite migration

dholms:
Doesn't _really_ matter cause this table is going away soon anyway. But for now, it's breaking some local dev-envs

Closes https://github.com/bluesky-social/atproto/issues/1523


devinivy (approved):
Thanks for cleaning that up!

PR #1533, opened 2023-08-29
Add APIs for getting and setting account owner birth age

pfrazee:
For moderation purposes, we need to store the user's birth date and tell the application their age (in order to enable/disable adult content). We want to ask for the birth date but only reveal the age via the APIs, and at this stage the age shouldn't be published to other users. (In the future users may want to put their age in their profiles but that's not in the scope of this task.)

Tasks: 

- [ ] Finalize lexicons
- [ ] Update database to store birth age
- [ ] Implement API routes

### DIFF
--- a/lexicons/com/atproto/server/createAccount.json
+++ b/lexicons/com/atproto/server/createAccount.json
@@ -16,7 +16,8 @@
             "did": {"type": "string", "format": "did"},
             "inviteCode": {"type": "string"},
             "password": {"type": "string"},
-            "recoveryKey": {"type": "string"}
+            "recoveryKey": {"type": "string"},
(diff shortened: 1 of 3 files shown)


pfrazee:
I've implemented some proposed lexicons for this. Some notes:

- I chose to put these items under `com.atproto` because they have to do with account management and is intended to be universal across applications (age gating for moderation)
- I had to add an `updateAccount` method for this
- I decided the simplest option for fetching the age was to add a `describeAccount` method rather than stuffing it in get/create session, since it involves a date calculation and we might want to reduce access to the information

Open to alternative approaches, but this is my first idea

pfrazee:
We're going to take a different approach

PR #1539, opened 2023-08-30
Legacy checkouts

dholms:


devinivy (approved):

packages/pds/src/db/index.ts
```diff
@@ -193,8 +193,17 @@ export class Database {
...
     this.assertTransaction()
+    return this.txAdvisorylock(name)
+  }
+
+  async checkTxAdvisoryLock(name: string): Promise<boolean> {
+    this.assertNotTransaction()
+    return this.txAdvisorylock(name)
+  }
+
+  private async txAdvisorylock(name: string): Promise<boolean> {
```
devinivy, line 206: Just a tiny consistency thing.
```suggestion
    return this.txAdvisoryLock(name)
  }

  async checkTxAdvisoryLock(name: string): Promise<boolean> {
    this.assertNotTransaction()
    return this.txAdvisoryLock(name)
  }

  private async txAdvisoryLock(name: string): Promise<boolean> {
```

dholms: yup good catch

packages/pds/src/db/index.ts
```diff
@@ -193,8 +193,17 @@ export class Database {
...
 
-  async txAdvisoryLock(name: string): Promise<boolean> {
+  async takeTxAdvisoryLock(name: string): Promise<boolean> {
     this.assertTransaction()
+    return this.txAdvisorylock(name)
+  }
+
+  async checkTxAdvisoryLock(name: string): Promise<boolean> {
+    this.assertNotTransaction()
+    return this.txAdvisorylock(name)
```
devinivy: Clever/interesting—I assume it takes and releases the lock during the course of the statement?  I wonder if there's a simple way to check without taking it at all. 🤔 May not really matter!

packages/pds/src/services/repo/index.ts
```diff
@@ -159,6 +164,12 @@ export class RepoService {
     writes: PreparedWrite[],
     swapCommit?: CID,
   ): Promise<CommitData> {
+    // this is not in a txn, so this won't actually hold the lock,
+    // we just check if it is currently held by another txn
+    const available = await storage.lockAvailable()
+    if (!available) {
+      throw new ConcurrentWriteError()
+    }
```
devinivy, line 172: I take it this is defensive, i.e. not really required but allows us to fail early?

dholms: yup exactly

devinivy (approved):

PR #1545, opened 2023-09-01
Abyss Integration

dholms:
Expands the labeler abstraction to serve as an automoderator generally & introduces the ability to takedown records through abyss


packages/bsky/src/auto-moderator/abyss.ts
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  async makeReq(did: string, cid: CID): Promise<ScannerResp> {
+    const { stream, contentType } = await resolveBlob(
+      did,
+      cid,
+      this.ctx.db,
+      this.ctx.idResolver,
+    )
+    const { data } = await axios.post(this.getReqUrl({ did }), stream, {
```
bnewbold, line 47: Things we might want to do at this HTTP request call site:

- log response status code and duration (ms). we're also doing logging+metrics in abyss, but good to do it on this side of the loadbalancer as well
- have a timeout. 10sec?

bnewbold (approved):
One small note but looks good otherwise

packages/bsky/src/auto-moderator/abyss.ts
dholms: yup both seem reasonable 👍

packages/bsky/src/indexer/config.ts
```diff
@@ -86,6 +92,9 @@ export class IndexerConfig {
     assert(dbPostgresUrl)
     assert(redisHost || (redisSentinelName && redisSentinelHosts?.length))
     assert(indexerPartitionIds.length > 0)
+    assert(imgUriEndpoint)
```
dholms, line 95: btw i think i'm gonna make these not required & if they're not setup, then just no abyss integration

packages/bsky/src/auto-moderator/index.ts
```diff
@@ -0,0 +1,184 @@
...
+            'failed to check img for takedown',
+          )
+        }),
+      ])
+    })
+  }
+
+  async labelRecord(uri: AtUri, recordCid: CID, text: string[], imgs: CID[]) {
+    if (uri.collection === ids.AppBskyActorProfile) {
+      // @TODO label profiles
```
devinivy, line 103: This all good to leave as a `@TODO`?

packages/bsky/src/auto-moderator/index.ts
```diff
@@ -0,0 +1,184 @@
...
+        subject: { uri, cid: recordCid },
+        subjectBlobCids: takedownCids,
+        reason: `automated takedown for labels: ${results.flat().join(', ')}`,
+        createdBy: this.ctx.cfg.labelerDid,
+      })
+      await modSrvc.takedownRecord({
+        takedownId: action.id,
+        uri: uri,
+        blobCids: takedownCids,
+      })
```
devinivy, line 150: For this to be synced with the PDS we want to actually hit the PDS and let it propagate to the appview—a similar approach can be found in `packages/bsky/src/db/periodic-moderation-action-reversal.ts`.

devinivy (approved):
Left one note re: syncing takedowns which may be important, but generally everything here looks solid!

packages/bsky/src/auto-moderator/index.ts
dholms: yup 👍 will work by just deleting that check. just need client signoff that there won't be adverse side effects

packages/bsky/src/auto-moderator/index.ts
dholms: ah good looks 👍

PR #1547, opened 2023-09-04
Refactor hydration / presentation to unify application of block and mute state

devinivy:
This is a long-time-coming refactor of how we wrangle all the state that goes into presenting feeds and actor profiles in the appview.  In particular, we'd like to look-up all the state for blocks and mutes a single time both for a. applying the blocks and mutes to remove unwanted content from feeds, and b. presenting the `viewer` state (e.g. `post.author.viewer.mutedByList`) that appears in responses.

In order to achieve this, we got a bit stricter about how responses are constructed.  We are gradually moving it to four steps which are referred in the code as a "pipeline":
 - Skeleton: this step determines the result set, often as simple as a list of dids or uris.
 - Hydration: this step fetches all data that will be necessary to construct the result set in the response.  For example, on `getPosts` the skeleton is a list of post URIs, and the hydration step pulls all the full details of posts, actors, labels, block state, etc. that will be used to construct the response.
 - Rules: this step modifies the hydration state to apply some rules, e.g. removing results that violate the viewer's blocks or mutes.
 - Presentation: this maps the hydration state (with rules applied) down into the actual response, generally matching lexicon output.

This has allowed us to remove some awkward hacks around label hydration (see: removed `kSelfLabels` and `skipLabels`).

For the reviewer:
 - There are two separate areas of review: reorganizing services to support the pipeline, and reorganizing endpoints to use the pipeline.
 - On service reorganization could use eyes on:
   - Removal of some duplicate logic, particularly for a. presenting nested embeds, and b. for presenting basic profiles (previously used different but similar methods logic in feed and actor services).
   - Ensure successfully avoiding duplicate roundtrips for labels and block/mute state.  The way this works now is that `getBlockAndMuteState` and `getLabelsForSubjects` can accept the current state and use it to avoid looking-up data that has already been fetched.  That should end in a no-op whenever possible.
 - On endpoint reorganization could use eyes on:
   - Ensure no change to application of mutes and blocks.  Some of this logic has been moved from the query into the "rules" step.  Each endpoint has slightly different rules about how blocks/mutes are applied, so it does require some care to check each one.
   - Generally ensure no important logic has changed or is missing.
   - Sometimes there are little judgement calls around how to split-up steps of the pipeline, certainly open to feedback.

Future work:
 - DRY-up pipeline steps.  There are some repeated across endpoints that we can probably collapse down.
 - Apply pipeline to the rest of the endpoints.  Most endpoints (and in particular, all that touched application of blocks/mutes) have been moved over to a pipeline, but there are some more to hit.  These should not require any/significant changes in services to accomplish.
 - We have specific hydration/presentation steps in our services for profiles and posts, but it's not as baked for feedgens, lists, etc.  We can gradually move these over too.


packages/bsky/src/api/app/bsky/feed/getPostThread.ts
```diff
@@ -1,97 +1,104 @@
...
-  replies?: PostThread[]
-}
+import { createPipeline, noRules } from '../../../../pipeline'
+import { ActorService } from '../../../../services/actor'
 
 export default function (server: Server, ctx: AppContext) {
+  const getPostThread = createPipeline(
+    skeleton,
+    hydration,
+    noRules,
```
dholms, line 29: this should have rules right? around no blocks?

packages/bsky/src/services/feed/index.ts
```diff
@@ -280,29 +204,51 @@ export class FeedService {
...
-    ])
-    const blocks = await this.blocksForPosts(posts)
-    const embeds = await this.embedsForPosts(posts, blocks, viewer)
+    return { dids: actorDids, uris: postUris }
+  }
 
-    return this.views.formatFeed(
-      items,
-      actors,
+  async feedHydration(
```
dholms, line 210: not to muck this up even more, but we want to treat block violations different in different contexts. for instance, a block violating reply is:
- not shown in thread
- shown in author feed

block violating quote posts are still shown but detach the quoted post

packages/bsky/src/services/feed/views.ts
```diff
@@ -83,13 +84,18 @@ export class FeedViews {
 
   formatFeed(
```
dholms, line 85: the 3rd party block check in here kinda seems like it should be in the rules fn instead of in the presentation fn 🤔

dholms: altho i guess the problem there is that it technically stays in the skeleton but is "presented" differently (with `#notFoundPost`)

packages/bsky/src/api/app/bsky/actor/getSuggestions.ts
```diff
@@ -1,75 +1,124 @@
...
+import { QueryParams } from '../../../../lexicon/types/app/bsky/actor/getSuggestions'
+import { createPipeline } from '../../../../pipeline'
+import { ActorInfoMap, ActorService } from '../../../../services/actor'
+import { BlockAndMuteState, GraphService } from '../../../../services/graph'
 
 export default function (server: Server, ctx: AppContext) {
+  const getSuggestions = createPipeline(
+    skeleton,
+    hydration,
+    noBlocks,
```
dholms, line 16: i guess it's diff behavior than we have no, but seems like we should remove mutes form this as well

packages/bsky/src/services/feed/views.ts
devinivy: I don't disagree at all, but it's tricky to make it happen without some more in-depth changes—same goes for the `noRules` on `getPostThread`.  As far as block state goes, luckily I think those are really the only two loose ends.  My primary goal here was to unify application and presentation of the viewer's block state.

Another similar situation is the presentation logic that happens during hydration of embeds.  Ideally all that presentation logic would be hoisted up too.

So you can think of this refactor as sort of going "one level deep", but there are some things in here that are deeply enough coupled to hydration that I chose not to dismantle.  Happy to hop back in though 😈

packages/bsky/src/api/app/bsky/feed/getAuthorFeed.ts
```diff
@@ -1,112 +1,171 @@
...
+
+export const skeleton = async (
+  params: Params,
+  ctx: Context,
+): Promise<SkeletonState> => {
+  const { cursor, limit, actor, filter, viewer } = params
+  const { db, actorService, feedService, graphService } = ctx
+  const { ref } = db.db.dynamic
+
+  // maybe resolve did first
```
dholms, line 61: if we wanted to, we could simplify this logic by just not returning block violating posts - ie an empty feed 🤷‍♂️ 

Tradeoff is:
- trimming off a couple of DB hits that we do on every req
- some extra composition/hydration for reqs on a blocked actor

packages/bsky/src/services/feed/index.ts
devinivy: Fixing that behavior seems like a good chunk of work on which to hinge some further dismantling of feed hydration.  I am pretty sure it would be "additive" onto the work done so far here, so I think we have leeway about whether to bite it off now or handle separately.

packages/bsky/src/api/app/bsky/feed/getAuthorFeed.ts
devinivy: I'm totally down, would just need to be on the lookout for whether the client uses this error as a signal in any way.

packages/bsky/src/api/app/bsky/feed/getFeed.ts
```diff
@@ -137,3 +219,26 @@ async function skeletonFromFeedGen(
     feed: skeleton.feed.slice(0, params.limit), // enforce limit
   }
 }
+
+const cleanFeedSkeleton = async (
```
dholms, line 223: this isn't really "cleaning" anymore right? it's more of a "pre-hydration hydration" step 🤔 

mainly just the naming thing

dholms: but also if we're feeling adventurous, seems that we could cut out some duplicate work if we moved this to the skeleton step & only did it for 3rd party feeds

packages/bsky/src/api/app/bsky/feed/getPostThread.ts
devinivy: Yeah, we could pretty easily yank 3p block-violating replies.  Blocks violating the viewer state are sort of part of the presentation, though, since they turn into `#blockedPost`s.

packages/bsky/src/services/feed/views.ts
devinivy: Looking at this again, it might not bee too bad—I will give it a quick whirl!  Unraveling the embed presentation still looks like it may be for another day, though 🙃

packages/bsky/src/api/app/bsky/feed/getLikes.ts
```diff
@@ -1,78 +1,127 @@
...
 import AppContext from '../../../../context'
 import { notSoftDeletedClause } from '../../../../db/util'
+import { BlockAndMuteState, GraphService } from '../../../../services/graph'
+import { ActorInfoMap, ActorService } from '../../../../services/actor'
+import { Actor } from '../../../../db/tables/actor'
+import { Database } from '../../../../db'
+import { createPipeline } from '../../../../pipeline'
 
 export default function (server: Server, ctx: AppContext) {
+  const getLikes = createPipeline(skeleton, hydration, noBlocks, presentation)
```
dholms, line 14: looks like we were originally filtering mutes here - which tbh i think i like it better just blocks 🤔

packages/bsky/src/api/app/bsky/feed/getPostThread.ts
dholms: ah right i see that 👍 

mer making me wish more and more that the response was flattened...

dholms: maybe a comment here next to `noRules` saying that it's applied in presentation?

packages/bsky/tests/views/__snapshots__/follows.test.ts.snap
```diff
@@ -5,9 +5,9 @@ Object {
   "cursor": "0000000000000::bafycid",
   "followers": Array [
     Object {
-      "avatar": "https://bsky.public.url/img/avatar/plain/user(3)/cids(0)@jpeg",
+      "avatar": "https://bsky.public.url/img/avatar/plain/user(1)/cids(0)@jpeg",
```
dholms, line 8: any idea why these snaps updated?  they look totally the same to me, just the numbers changed 🧐

devinivy: Oh yeah, I meant to make a note of that.  I was able to verify it's because the response fields now come in a different order.  Jest canonicalizes the output for the snapshots, but the util that replaces cids with `cid(x)`, etc. depends on the order of `Object.entries()`.  Surprised we haven't run into this more often!

packages/bsky/src/api/app/bsky/feed/getLikes.ts
devinivy: If that's based off of the name `filterBlocksAndMutes()` below, it's a little confusing but it was only applying blocks, since it was only being passed an implementation for `getBlockPairs()` and not `getMutePairs()`.  In any case, I'll leave it as-is 👍 👍

packages/bsky/src/services/feed/index.ts
dholms: yeah makes sense - i'm down to tackle it as a separate chunk of work 👌

packages/bsky/src/api/app/bsky/feed/getLikes.ts
dholms: ah right on - yup 👍

packages/bsky/tests/views/__snapshots__/follows.test.ts.snap
dholms: ah nice 👍

dholms (approved):
killer work on this - really happy how it came together 👌

packages/bsky/src/api/app/bsky/feed/getAuthorFeed.ts
devinivy: Ended-up punting on this, since the client does make use of it: https://github.com/bluesky-social/social-app/blob/565c1a0510b5fb516cc23db4f4e2ec903ad66730/src/state/models/feeds/posts.ts#L330-L331

packages/bsky/src/services/feed/views.ts
devinivy: Dove into this, and sadly I think it makes sense to leave as-is for the time being.  It is actually straightforward to get rid of that one 3p block check for replies.  However, in `formatMaybePostView()` a similar check is made and in that case it applies to the presentation, possibly resulting in a `#blockedPost` for `post.reply.parent` rather than being omitted entirely.  Since these 3p block checks are so similar and appear nearby each other, I think it may be best to leave both of them colocated for now.

My takeaways:
 - it probably makes sense to blow this up when we do a deeper fix for handling 3p blocks (ref: https://github.com/bluesky-social/atproto/pull/1547#discussion_r1316434594) rather than this one-off.
 - may need to think through pipeline `rules()` more, to better account for situations where what we think of as rules end-up affecting the presentation beyond just omitting some data.

packages/bsky/src/api/app/bsky/feed/getPostThread.ts
devinivy: Done!

packages/bsky/src/api/app/bsky/actor/getSuggestions.ts
devinivy: Done!

packages/bsky/src/api/app/bsky/feed/getFeed.ts
devinivy: Done!

devinivy:
@dholms I made some final changes, final comments, and merged.  If there's anything leftover in here you think we should take another look at, no problem!

PR #1532, opened 2023-08-29
pnpm setup

estrattonbailey:
Replaces `yarn` with `pnpm` and fixes a few dozen type errors, dependency mismatches, and missing dependencies.

To test, checkout this branch and run through all commands:
```
pnpm i

pnpm build

pnpm verify

pnpm format

pnpm test
```

When publishing, the process should be essentially the same and we should be able to use any flavor of `publish` we want e.g. `npm` `yarn` or `pnpm`. My plan is to add Changesets to automate this in a future PR, at which point we won't need to worry about manually publishing anymore.

Todo:
- [x] test publishing with `publishConfig`
- [x] test Docker builds
- [x] test run Docker image

PR #1548, opened 2023-09-04
Add missing dependencies

aleclarson:
I'm using PNPM locally, which doesn't allow implicit dependency sharing between packages, so I found these missing dependencies as a result. I had to update the `multiformats` version everywhere to keep a single version in the lockfile.


dholms:
We actually just migrated this repo to use pnpm 😆 

https://github.com/bluesky-social/atproto/pull/1532

Thanks for putting up this PR, but I think we got all these covered in the one we just landed 🙏

PR #1549, opened 2023-09-05
Indexing and applying block lists

devinivy:
Building on #1547, indexes list-blocks and applies them throughout views.  Includes implementation of `getListBlocks` and tests.  I also noticed that `getList` could make use of services for hydration/presentation, so DRY'ed that up.


packages/bsky/src/services/graph/index.ts
```diff
@@ -116,12 +124,28 @@ export class GraphService {
...
           .select('uri')
           .as('blockedBy'),
+        this.db.db
+          .selectFrom('list_item')
+          .innerJoin('list_block', 'list_block.subjectUri', 'list_item.listUri')
+          .whereRef('list_block.creator', '=', targetRef)
+          .whereRef('list_item.subjectDid', '=', sourceRef)
+          .select('list_item.listUri')
+          .limit(1)
+          .as('blockedByViaList'),
```
dholms, line 148: took me a second to parse this but makes sense 😆

dholms: (i actually like the wording here, it's just a funny concept to name)

dholms (approved):
beautiful - wow that flowed out super nicely from the prev work 👌

PR #1553, opened 2023-09-05
add getSuggestedFollowsByActor

estrattonbailey:
The intention here is to provide additional suggested accounts on an account's profile view _after_ a user follows that account.

Simple logic to ease migration to v2 arch:
- get 1000 likes of `actor`, get accounts of the liked posts, rank by most common accounts, select top 10
- if commonly liked accounts is less than 10, get `actor`'s `follow`s sorted by # of followers (popularity), limit to 10, and append to result
- in _either case_, if result is still < 10, fill result with `suggested_follow` table values
- hydrate profiles for results, filtering out blocks/mutes/etc
- return at most 10 results to the frontend


lexicons/app/bsky/graph/getSuggestedFollowsByActor.json
```diff
@@ -0,0 +1,33 @@
...
+  "defs": {
+    "main": {
+      "type": "query",
+      "description": "Get suggested follows related to a given actor.",
+      "parameters": {
+        "type": "params",
+        "required": ["actor"],
+        "properties": {
+          "actor": {"type": "string", "format": "at-identifier"},
+          "limit": {"type": "integer", "minimum": 1, "maximum": 100, "default": 50},
```
dholms, line 13: do we need pagination on this? it becomes even easier if we remove the need for an ordered result set & just send back results with no promise of order or completeness

packages/bsky/src/api/app/bsky/graph/getSuggestedFollowsByActor.ts
```diff
@@ -0,0 +1,122 @@
...
+
+      const db = ctx.db.getReplica()
+      const actorService = ctx.services.actor(db)
+      const actorDid = await actorService.getActorDid(actor)
+
+      if (!actorDid) {
+        throw new InvalidRequestError('Actor not found')
+      }
+
+      const likes = await db.db
```
dholms, line 22: i think it'd be better to do this aggregation in the db query itself to prevent 1000 rows from being sent back

ie
```ts
db.selectFrom(subquery).groupBy('subject').select('subject', 'count(*').orderBy('count', 'desc').limit(20)`
```

or similar, i'm sure my syntax is off

where the `subquery` is the same query you have here

packages/bsky/src/api/app/bsky/graph/getSuggestedFollowsByActor.ts
```diff
@@ -0,0 +1,122 @@
...
+            .selectFrom('suggested_follow')
+            .select('did')
+            .where(
+              'suggested_follow.did',
+              'not in',
+              // exclude any we already have
+              authorDIDsExcludingActorAndViewer.concat([actorDid, viewer]),
+            )
+            .execute()
+          const additionalActors = await db.db
```
dholms, line 89: may be nice to just merge this db hit with the previous one for getting actor info

packages/bsky/src/api/app/bsky/graph/getSuggestedFollowsByActor.ts
```diff
@@ -0,0 +1,122 @@
...
+            )
+            .selectAll()
+            .execute()
+
+          actors.push(...additionalActors)
+        }
+
+        // this handles blocks/mutes etc
+        suggestions = (
+          await actorService.views.hydrateProfiles(actors, viewer)
```
dholms, line 104: mer really gotta make this just accept `did`s so we don't need to do the pre-hydration 🙃 

i actually think some of the hydration rewrite stuff devin has been working on in https://github.com/bluesky-social/atproto/pull/1531 might do this

anyways no need to worry about it, just talking lol

packages/bsky/src/api/app/bsky/graph/getSuggestedFollowsByActor.ts
estrattonbailey: yeah sg, leaned too hard into the nosql vibes

packages/bsky/src/api/app/bsky/graph/getSuggestedFollowsByActor.ts
estrattonbailey: sick, love it, thanks

estrattonbailey: So I tried this, but realized that we can't group directly by `subject` since it's a URI. We need to group by the `did` portion of the URI to find most commonly liked accounts.

We could maybe use a substring helper to do this in the db query itself, looks like that differes by postgres/sqlite 🤔

estrattonbailey: Just posting that the query that I think should work is basically:

```ts
const mostLikedAccounts: { subject }[] = await db.db
  .selectFrom('like')
  .select('subject')
  .select(qb => qb.fn.count('subject').as('count')) // problem is subject is a URI
  .where('creator', '=', actorDid)
  .groupBy('subject')  // problem is subject is a URI
  orderBy('count', 'desc')
```

dholms: (no need to worry about sqlite in the appview btw)

a small note on your query - i think it's still worth doing this in a subquery because then we're cutting off the total aggregation (only looking at 1000 rows max not the entire likeset)

but yeah that's a good point... lemme think

dholms: off the cuff, a substring helper sounds like a pain - but at the same time, it should be pretty lightweight if it's wrapped in a subquery 🤔 

basically:
- grab 1000 most recent likes
- substring each of those uris to the subjectDid
- then run aggregation over that filtered set

dholms: but this also all might just be too galaxy brained for the problem at hand - really 1000 rows (of just 2 dids)  isn't _that_ big

dholms: okay final thoughts:

i think we can make this work fine with scylla (or at least something similar to it work fine)

if it's easy enough to write the substring in postgres, I think I would do that using the subquery (limited to 1000 still)

if not, then let's do it the old way and we _may_ have to drop the query size depending on performance. Is there a certain level at which you feel it loses signal? Like 500 likes seems fine? 200?

packages/bsky/src/api/app/bsky/graph/getSuggestedFollowsByActor.ts
```diff
@@ -0,0 +1,152 @@
...
+          )
+          .selectAll()
+          .execute()
+        const sortedAuthors = authorsSortedByMostCommon
+          .map(([did]) => authors.find((a) => a.did === did))
+          .filter(Boolean) as typeof authors
+
+        allActors = sortedAuthors
+      } else {
+        const popularFollows = await db.db
```
dholms, line 84: could we split each of these subqueries into it's own fn outside of the route handler (like just at the bottom of the file)?

it may make the control flow of the handler a bit easier to follow.

my initial thought when seeing this was "wow that's a lot of queries we're doing" before realizing that most of these queries aren't actually called because they're in if/elses

estrattonbailey: Cleaned this up a bit, lmk what ya think

packages/bsky/src/api/app/bsky/graph/getSuggestedFollowsByActor.ts
```diff
@@ -0,0 +1,132 @@
...
+          db.db
+            .selectFrom('like')
+            .where('creator', '=', actorDid)
+            .select(sql`split_part(subject, '/', 3)`.as('subjectDid'))
+            .as('likes')
+        )
+        .select('likes.subjectDid as did')
+        .select(qb => qb.fn.count('likes.subjectDid').as('count'))
+        .where('likes.subjectDid', 'not in', actorsViewerFollows)
+        .where('likes.subjectDid', 'not in', [actorDid, viewer])
```
estrattonbailey, line 43: Have to do this kind of thing in a few places, maybe could combine? Just don't want to overfetch, cuz if we have to filter out "actors viewer already follows" later, it could cut down on the # of results substantially (if the viewer follows a lot of accounts in common with the actor).

Doing it in each query gives us as many good results as possible each time 🤔

packages/bsky/src/api/app/bsky/graph/getSuggestedFollowsByActor.ts
```diff
@@ -0,0 +1,135 @@
...
+       * 20 most liked accounts that aren't already followed by the viewer, ARE
+       * the viewer, or are the actor
+       */
+      const mostLikedAccounts = db.db
+        .selectFrom(
+          db.db
+            .selectFrom('like')
+            .where('creator', '=', actorDid)
+            .select(sql`split_part(subject, '/', 3)`.as('subjectDid'))
+            .as('likes'),
```
dholms, line 38: can we put a limit of 1000 on this subquery? otherwise it'll end up iterating over every like (which will be pretty expensive for some accounts

packages/bsky/src/api/app/bsky/graph/getSuggestedFollowsByActor.ts
```diff
@@ -0,0 +1,135 @@
...
+      /**
+       * 20 most liked accounts that aren't already followed by the viewer, ARE
+       * the viewer, or are the actor
+       */
+      const mostLikedAccounts = db.db
+        .selectFrom(
+          db.db
+            .selectFrom('like')
+            .where('creator', '=', actorDid)
+            .select(sql`split_part(subject, '/', 3)`.as('subjectDid'))
```
dholms, line 37: 💪 😎 

thx for working through this with me - turned out nicely

packages/bsky/src/api/app/bsky/graph/getSuggestedFollowsByActor.ts
```diff
@@ -0,0 +1,135 @@
...
+            qb.where(
+              'subjectDid',
+              'not in',
+              actors.map((a) => a.did),
+            ),
+          )
+        const mostPopularAccountsActorFollows = db.db
+          .selectFrom('profile_agg')
+          .select(['did', 'followersCount'])
+          .innerJoin(
```
dholms, line 76: i think you can do this without the subquery & it'll give postgres more flexibility to find the right path

right now i _think_ i'll try to materialize all of `actorsActorFollow` before adding in post agg content.

basically:
- INNER JOIN follow & profile_agg
- apply all the same WHEREs/ORDER BY/LIMIT you already have

packages/bsky/src/api/app/bsky/graph/getSuggestedFollowsByActor.ts
```diff
@@ -0,0 +1,135 @@
...
+            'follows.subjectDid',
+            'profile_agg.did',
+          )
+          .orderBy('followersCount', 'desc')
+          .limit(20)
+        const mostPopularActors = await db.db
+          .selectFrom('actor')
+          .selectAll()
+          .innerJoin(
+            mostPopularAccountsActorFollows.as('popularFollows'),
```
dholms, line 87: Joining on a subquery scares me just a bit (it hamstrings postgres a bit)

I actually think this forces it into an okay query path and it'd probably be fine, but we can probably aovid it easily.

I'd ditch the "actor wrappers" you have around these three queries & just construct a list of dids - hydrate that into an actor after constructing the full list

dholms: mer `hydrateProfiles` should really take a list of dids so you didn't have to worry about this at all...

packages/bsky/src/api/app/bsky/graph/getSuggestedFollowsByActor.ts
```diff
@@ -0,0 +1,135 @@
...
+          .selectAll()
+          .execute()
+
+        actors.push(...additional)
+      }
+
+      // resolve all profiles, this handles blocks/mutes etc
+      const suggestions = (
+        await actorService.views.hydrateProfiles(actors, viewer)
+      ).filter((account) => {
```
dholms, line 119: excellent 👌

packages/bsky/src/api/app/bsky/graph/getSuggestedFollowsByActor.ts
```diff
@@ -0,0 +1,135 @@
...
+        .select('likes.subjectDid as did')
+        .select((qb) => qb.fn.count('likes.subjectDid').as('count'))
+        .where('likes.subjectDid', 'not in', actorsViewerFollows)
+        .where('likes.subjectDid', 'not in', [actorDid, viewer])
+        .groupBy('likes.subjectDid')
+        .orderBy('count', 'desc')
+        .limit(20)
+
+      /**
+       * 20 most liked accounts as actor results
```
dholms, line 49: is there a reason we get 20 when the max is 10?

packages/bsky/src/api/app/bsky/graph/getSuggestedFollowsByActor.ts
dholms: yeah this seems fine 👍 it's a _bit_ wordy but i prefer wordy over unclear

packages/bsky/src/api/app/bsky/graph/getSuggestedFollowsByActor.ts
dholms: ah sorry i phrased this wrong

this did clean up quite a bit & i can follow it well!

I should have said "queries" not "subqueries.

I just mean something like:
```ts
const suggestions = await getSuggestionsFromLikes(...)
if(suggestions.length < MAX) {
  const fromFollows = await getSuggestsionFromFollow(...)
  // append
} 
if(suggestions.length < MAX) {
  const fromSuggestedFollows = await getSuggestionsFromSuggestedFollows(...)
  // append
}
```

where each of those function impls are what you currently have just split out into a named function.

then the route handler itself can be like 15 lines & really easy to follow what it's doing

packages/bsky/src/api/app/bsky/graph/getSuggestedFollowsByActor.ts
estrattonbailey: Thanks for your thoughts!

packages/bsky/src/api/app/bsky/graph/getSuggestedFollowsByActor.ts
estrattonbailey: Oh great call 👍

packages/bsky/src/api/app/bsky/graph/getSuggestedFollowsByActor.ts
estrattonbailey: Heh yeah that'd be cool, but u right, can do as the last step 👍

packages/bsky/src/api/app/bsky/graph/getSuggestedFollowsByActor.ts
estrattonbailey: Oh just to allow for some of the results to be blocked/muted by the viewer. I'm assuming that the # of users that block > 50% of the result set is pretty rare.

Basically I want it to be able to fill 10 results as much as possible. We could filter by blocks in each of the queries too, though 🤔

estrattonbailey: Answered my own question here, one sec

dholms (approved):
nice this really shaped up 💎

PR #1559, opened 2023-09-06
maintain feed order

estrattonbailey:
The `where in` clause in `getFeedGeneratorInfos` doesn't maintain order, so doing it manually. There might be a way to do this with postgres but we're only doing this with 26 rows rn.

### DIFF
--- a/packages/bsky/src/api/app/bsky/feed/getSuggestedFeeds.ts
+++ b/packages/bsky/src/api/app/bsky/feed/getSuggestedFeeds.ts
@@ -19,7 +19,11 @@ export default function (server: Server, ctx: AppContext) {
         feedsRes.map((r) => r.uri),
         viewer,
       )
-      const genList = Object.values(genInfos)
+      const genList = Object.values(genInfos).sort((a, b) => {
+        const aOrder = feedsRes.find((r) => r.uri === a.uri)?.order || 1
+        const bOrder = feedsRes.find((r) => r.uri === b.uri)?.order || 1
+        return aOrder - bOrder
+      })
 
       const creators = genList.map((gen) => gen.creator)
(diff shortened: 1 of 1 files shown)


packages/bsky/src/api/app/bsky/feed/getSuggestedFeeds.ts
```diff
@@ -19,7 +19,11 @@ export default function (server: Server, ctx: AppContext) {
         feedsRes.map((r) => r.uri),
         viewer,
       )
-      const genList = Object.values(genInfos)
+      const genList = Object.values(genInfos).sort((a, b) => {
```
dholms, line 22: what about ordering this just based on `feedsRes`?

something like:
```ts
const genList = feedsRes.map(r => genInfos[r.uri])
```

estrattonbailey: Ah yeah I shouldn't rely entirely on TS type huh. Copy that, fixing 👍

packages/bsky/src/api/app/bsky/feed/getSuggestedFeeds.ts
```diff
@@ -14,16 +14,13 @@ export default function (server: Server, ctx: AppContext) {
...
         .selectAll()
         .execute()
-
       const genInfos = await feedService.getFeedGeneratorInfos(
         feedsRes.map((r) => r.uri),
         viewer,
       )
-      const genList = Object.values(genInfos)
-
+      const genList = feedsRes.map((r) => genInfos[r.uri]).filter(Boolean)
```
dholms, line 21: whoa never seen `filter(Boolean)` before. makes sense tho 👍

dholms (approved):
looks good :thum

PR #1561, opened 2023-09-07
Interaction Gating

dholms:
Adds interaction gating to threads

This takes the form a new record type `app.bsky.feed.threadgate`

A `threadgate` lives _alongside_ a `post` with the same `rkey`. It references the post in question as well as a set of up to 5 rules about who many reply within a given thread.

These rules can include:
- mentioned in post
- followed by user
- member of a list

The interaction rules setup in the root of the thread apply to the entire thread -  each reply inherits it's rules from the root.

For interaction gating, unlike blocklists, we materialize wether a post was in violation of the interaction gate at write time.
This is for several reasons:
- it dramatically simplifies thread queries
- while blocklists are global, threads are local. They are often "active" for a short duration of time and they are self-contained
- applying gating at write time more closely mirrors user expectation


packages/bsky/src/services/indexing/plugins/post.ts
```diff
@@ -96,6 +103,37 @@ const insertFn = async (
...
+      .selectFrom('post')
+      .innerJoin('record', 'record.uri', 'post.uri')
+      .where('post.uri', 'in', [replyRoot, replyParent])
+      .selectAll('post')
+      .select('record.json')
+      .execute()
+    const root = replyRefs.find((ref) => ref.uri === replyRoot)
+    const parent = replyRefs.find((ref) => ref.uri === replyParent)
+    const isInvalidReply = !parent || checkInvalidReply(obj.reply, parent)
+    const isInvalidInteraction =
```
dholms, line 119: may be nice to split up this logic a bit - have a hard time following it

i also think that this invalidates any replies after a root is deleted?

lexicons/app/bsky/feed/defs.json
```diff
@@ -139,6 +157,36 @@
       "properties": {
         "repost": {"type": "string", "format": "at-uri"}
       }
+    },
+    "mentionInteractionView": {
+      "type": "object",
+      "description": "The view counterpart to app.bsky.feed.post#mentionInteraction"
```
dholms, line 163: kinndaa inclined to have these live next to one another 🤔 

but then again, we have `post` & `#postView` separate from one another 🧐

devinivy: Yeah I was a little wishy-washy about this, but that was my thinking on it.  If we were to change this, would you lean towards moving them all into `app.bsky.feed.defs` or all into `app.bsky.feed.post`?

packages/bsky/src/services/feed/views.ts
```diff
@@ -342,4 +375,39 @@ export class FeedViews {
       media: mediaEmbed,
     }
   }
+
+  formatGate(gate: GateRecord, lists: ListInfoMap) {
```
devinivy, line 379: Do we want to include the original record or uri of the gate here?

packages/bsky/src/services/indexing/plugins/gate.ts
```diff
@@ -0,0 +1,95 @@
+import { AtUri } from '@atproto/syntax'
```
devinivy, line 1: Awkwardly I found we don't really make use of the new `gate` index—we always grab the gate record and just work with it that way.  We can do this mostly because the gate uri can be derived from the post uri.  But it also feels a little odd not to index them at all.

lexicons/app/bsky/feed/gate.json
```diff
@@ -0,0 +1,46 @@
...
+    "followingRule": {
+      "type": "object",
+      "description": "Allow replies from actors you follow."
+    },
+    "listRule": {
+      "type": "object",
+      "description": "Allow replies from actors on a list.",
+      "required": ["list"],
+      "properties": {
+        "list": {"type": "ref", "ref": "com.atproto.repo.strongRef"}
```
devinivy, line 42: We could alternately make this a weak ref.  I went back and forth on this one.

packages/bsky/src/db/tables/post.ts
```diff
@@ -12,6 +12,8 @@ export interface Post {
   replyParent: string | null
   replyParentCid: string | null
   langs: string[] | null
+  isInvalidReply: Generated<boolean>
+  isInvalidInteraction: Generated<boolean>
```
devinivy, line 16: This "interaction" terminology is a bit leftover from the last pass on this, I wonder if we should change it.  The `isInvalidReply` means that the contents of `post.reply` just doesn't look accurate, e.g. the reply parent and root aren't compatible.  The `isInvalidInteraction` incorporates the idea of violating a thread's reply/interaction gate. 🤔

lexicons/app/bsky/feed/gate.json
dholms: yeah I think a weak ref 🤔 similar to a `listitem`, we only provide the `listUri` not the cid. 

the list cid also doesn't relate to the contents of the list. so it would really only be an integrity check against the list displayname/avatar/etc

packages/bsky/src/db/tables/post.ts
dholms: i agree - it's not immediately clear from reading...

maybe just making it super clear

- `invalidReplyRoot`
- `violatesThreadGate`

packages/bsky/src/services/indexing/plugins/gate.ts
dholms: yeah i think we just index for now - there are other reasons that it's useful to have this table around - for instance introspecting it ourselves. we may find we need it down the line as well

packages/bsky/src/services/indexing/plugins/post.ts
```diff
@@ -96,6 +106,18 @@ const insertFn = async (
     return null // Post already indexed
   }
 
+  if (obj.reply) {
```
dholms, line 109: could we do this in the initial insert? no biggie if not

packages/bsky/src/services/feed/util.ts
```diff
@@ -0,0 +1,111 @@
...
+  }
+  // replying to root post: ensure the root looks correct
+  if (replyParent === replyRoot) {
+    return !!parent.record.reply
+  }
+  // replying to a reply: ensure the parent is a reply for the same root post
+  return parent.record.reply?.root.uri !== replyRoot
+}
+
+export const checkInvalidInteractions = async (
```
dholms, line 36: nit, but this is a single interaction

maybe `isInvalidInteraction`?

dholms: similarly to `checkInvalidReplyParent` the "check" suggests to me that this returns more info than just a boolean

sorry this is such a nit 😅 i just had to check the fn impl to figure out what these were doing cause the name wasn't immediatley clear to me

packages/bsky/src/services/feed/index.ts
```diff
@@ -216,31 +225,35 @@ export class FeedService {
...
     // profileState for labels and bam handled above, profileHydration() shouldn't fetch additional
-    const [profileState, blocks] = await Promise.all([
+    const [profileState, blocks, lists] = await Promise.all([
       this.services.actor.views.profileHydration(
         Array.from(dids),
         { viewer },
         { bam, labels },
       ),
       this.blocksForPosts(posts, bam),
+      this.listsForPostGates(gates, viewer),
```
dholms, line 245: mer seems like a shame to have to hydrate the lists for every post :/

similar to viewer `canReply` state - is this actually needed in feeds?

may be a paul question

packages/bsky/src/services/feed/views.ts
dholms: hmm the uri is deterministic based on post uri so not necessary, but does seem kinda nice....

in general, i like the idea of including primary content where it doesn't add a bunch of overhead

maybe a gateview could look something like: 
```
{
  record: Record
  lists: ListView[]
}
```
or
```
{
  record: Record
  view: #gateView
}
```
the latter is similar to how we do post embeds, but is more duplicate data

packages/bsky/tests/views/gating.test.ts
```diff
@@ -0,0 +1,575 @@
...
+    )
+    expect(aliceThread.post.gate).toMatchSnapshot()
+    expect(aliceThread.viewer).toEqual({ canReply: true })
+    const [reply, ...otherReplies] = aliceThread.replies ?? []
+    assert(isThreadViewPost(reply))
+    expect(otherReplies.length).toEqual(0)
+    expect(reply.post.uri).toEqual(aliceReply.ref.uriStr)
+  })
+
+  it('does not apply gate to original poster.', async () => {
```
dholms, line 466: good test - i woulda forgetten about this one 😅

dholms:
can't approve cuz i created the PR

few small things but overall this came together super nicely 👌

packages/bsky/src/services/feed/index.ts
```diff
@@ -216,31 +225,35 @@ export class FeedService {
     depth = 0,
   ): Promise<FeedHydrationState> {
     const { viewer, dids, uris } = refs
-    const [posts, labels, bam] = await Promise.all([
+    const [posts, gates, labels, bam] = await Promise.all([
       this.getPostInfos(Array.from(uris), viewer),
+      this.gatesByPostUri(Array.from(uris)),
```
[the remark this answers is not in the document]
devinivy: One thing I considered here is to move the `allow` rules into a single column `gate` table, and then join it over as part of `getPostInfos()`.  At one point I did something like that in `getPostInfos()` joining all the way over to `record` to pull out the full gate record, but adding two joins there seemed a bit heady.  Wouldn't be too worried about adding one, though.  Tradeoffs are a. more space, to duplicate storage of the rules, b. adding a join when fetching post info.  Maybe worth it?

packages/bsky/src/services/feed/index.ts
devinivy: Spoke with Paul, sounds like we're going to keep this one for now 👍

lexicons/app/bsky/feed/gate.json
```diff
@@ -0,0 +1,46 @@
+{
+  "lexicon": 1,
+  "id": "app.bsky.feed.gate",
```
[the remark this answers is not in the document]
devinivy: Agree 👍 I sort of forgot that using two words is fine, and we can just keep them lowercased (similar to `listitem`).

packages/bsky/src/services/indexing/plugins/post.ts
devinivy: We totally can—my thinking was that it should be rare enough that we need to perform the write that it shouldn't matter a ton either way.

lexicons/app/bsky/feed/gate.json
```diff
@@ -0,0 +1,46 @@
...
+  "defs": {
+    "main": {
+      "type": "record",
+      "key": "tid",
```
dholms, line 7: also think it's wroth adding a note about the rkey constraint here (being the same as the post)

maybe in `description` or something

packages/bsky/src/services/indexing/plugins/post.ts
dholms: actually yeah that's a good point - seems fine 👍

packages/bsky/src/services/feed/index.ts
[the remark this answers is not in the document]
dholms: yeah that seems like a good story for what we'll do if this turns out to be outsized pain. let's keep it simple for now & we can add it in if we realize we need it 👍

packages/bsky/src/services/feed/views.ts
devinivy: Ended-up going with the former, seems like it should be just fine here.  Also allowed me to avoid needing to present not-founds.  I added in `{ uri, cid, record }` since that seems like it might be sort of standard-issue record info.

packages/bsky/src/services/indexing/index.ts
```diff
@@ -363,6 +363,10 @@ export class IndexingService {
       .where('post_embed_record.postUri', 'in', postByUser)
       .execute()
     await this.db.db.deleteFrom('post').where('creator', '=', did).execute()
+    await this.db.db
```
dholms, line 366: ah good catch 👍

dholms:
looks great 💪

lexicons/app/bsky/feed/threadgate.json
```diff
@@ -0,0 +1,43 @@
...
+      }
+    },
+    "mentionRule": {
+      "type": "object",
+      "description": "Allow replies from actors mentioned in your post."
+    },
+    "followingRule": {
+      "type": "object",
+      "description": "Allow replies from actors you follow."
+    },
```
christiandeange, line 33: Aren't `object` types supposed to have 1 or more `properties`? I maintain a lexicon codegen plugin and I noticed it's choking on these new types

christiandeange: Ah looks like it was already reported: https://github.com/bluesky-social/atproto/issues/1615

PR #1563, opened 2023-09-07
enable granular perms for publish action

estrattonbailey:
Enables these perms only for this job. Based [on docs](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs) I think we only need these two, but I'll keep an eye on it and can open it up only as much as needed.


dholms (approved):
excellent 💎

PR #1564, opened 2023-09-07
mark dev/service packages as private

estrattonbailey:
This should prevent these from being published, simply because we don't need them to be. [See docs](https://github.com/changesets/changesets/blob/main/docs/config-file-options.md#ignore-array-of-packages).

This **will not** prevent these packages from being referenced in changesets, and their versions will be bumped as well. Just never published to npm.

### DIFF
--- a/packages/dev-env/package.json
+++ b/packages/dev-env/package.json
@@ -1,4 +1,5 @@
 {
+  "private": true,
   "name": "@atproto/dev-env",
   "version": "0.2.3",
   "main": "src/index.ts",
--- a/packages/pds/package.json
+++ b/packages/pds/package.json
@@ -1,4 +1,5 @@
 {
+  "private": true,
   "name": "@atproto/pds",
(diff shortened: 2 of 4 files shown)


packages/pds/package.json
```diff
@@ -1,4 +1,5 @@
 {
+  "private": true,
```
dholms, line 2: ah we actually do publish this one for the pds distribution 

https://github.com/bluesky-social/pds

packages/dev-env/package.json
```diff
@@ -1,4 +1,5 @@
 {
+  "private": true,
   "name": "@atproto/dev-env",
```
dholms, line 3: doesn't the client team use this lib?

estrattonbailey:
Ah yeah you're right. Ok this is kinda pointless then, might as well publish everything for now and avoid any resolution weirdness. Maybe later 🖖

PR #1565, opened 2023-09-08
Add personal-details user preference with birth date

pfrazee:
Replaces https://github.com/bluesky-social/atproto/pull/1533

For moderation purposes, we need to store the user's birth date and tell the application their age (in order to enable/disable adult content). When we introduce OAuth and permissioning to the system, we'll be able to control access to this set of preferences using its NSID. For now, it will be available to any app that the user signs into with a main password or app password.

### DIFF
--- a/lexicons/app/bsky/actor/defs.json
+++ b/lexicons/app/bsky/actor/defs.json
@@ -134,6 +134,16 @@
           }
         }
       }
+    },
+    "personalDetailsPref": {
+      "type": "object",
+      "properties": {
+        "birthDate": {
+          "type": "string",
+          "format": "datetime",
+          "description": "The birth date of the owner of the account."
+        }
+      }
     }
   }
 }
--- a/lexicons/app/bsky/actor/defs.json
+++ b/lexicons/app/bsky/actor/defs.json
@@ -94,7 +94,12 @@
       "type": "array",
       "items": {
         "type": "union",
-        "refs": ["#adultContentPref", "#contentLabelPref", "#savedFeedsPref"]
+        "refs": [
+          "#adultContentPref",
+          "#contentLabelPref",
+          "#savedFeedsPref",
+          "#personalDetailsPref"
+        ]
       }
     },
     "adultContentPref": {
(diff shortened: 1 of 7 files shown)


dholms (approved):

dholms:
looks good! doesn't require any actual backend changes right?

pfrazee:
I don't _think_ so

lexicons/app/bsky/actor/defs.json
```diff
@@ -134,6 +134,16 @@
           }
         }
       }
+    },
+    "personalDetailsPref": {
```
devinivy, line 138: I think we may want to add this to the `#preferences` union a few dozen lines above.

lexicons/app/bsky/actor/defs.json
```diff
@@ -134,6 +134,16 @@
...
+      "type": "object",
+      "properties": {
+        "birthDate": {
+          "type": "string",
+          "format": "datetime",
+          "description": "The birth date of the owner of the account."
+        }
```
devinivy, line 145: What will our convention be for using datetimes to store a date?

lexicons/app/bsky/actor/defs.json
pfrazee: Oh shoot, right

lexicons/app/bsky/actor/defs.json
pfrazee: 🤷 do we need to worry about it? Time can be anything

devinivy: Yeah, I do think it's worth considering, have totally run into issues around treating dates as datetimes.  Mostly because folks truly mean a date, which specifically throws away time and timezone.

For example, if my birthday is `1980-05-30`, there's no situation where it should be interpreted as `1980-05-29`, even though the moment I was born could have been `1980-05-29` in some areas of the world.  (Maybe underlined even more by the fact that folks wont be inputting their exact birthtime.)

Two options here are a. represent it at midnight GMT, b. always ignore time information, take the `yyyy-mm-dd` part of the datetime verbatim.

devinivy (approved):

lexicons/app/bsky/actor/defs.json
pfrazee: Oh yeah that's all legit, my point is I'm not sure we need to be super precise with this because we just need to calculate user age

pfrazee:
Published `@atproto/api@0.6.11`

PR #1566, opened 2023-09-08
Move fuzzy matcher to appview

dholms:
This moves our fuzzy matcher for flagging possible slurs to the appview auto-moderator, instead of running in PDS


devinivy (approved):
👍 smooth

Issue #479, opened 2023-01-24
Email confirmation during signup? Some questions


pfrazee:
Email confirmation is meant to solve two problems:

1. Give the PDS admin confidence the user owns the email
2. Ensure the user gave the correct email so that we can offer support (eg password reset)

It's unclear that proving ownership (1) is a priority since we don't make any trust decisions based on user email. Ensuring we can provide support (2) however is a priority.

Here are the options in from of us:

- A. Include a step during signup where a confirmation email is sent. Wait for the user to enter the code in the email (or click the link inside) before we allow signup to finish.
- B. Send a confirmation email with a code/link during signup but don't block signup for confirmation. Afterwards, ask the user to confirm. Give an option to fix their email.
- C. Send a confirmation email with **no** code during signup. Afterwards, ask the user if they received the email at the given address but don't bother with a confirmation code. Give an option to fix their email.

Given that proving ownership (1) is not a priority but providing support (2) is, I'm inclined to say options B or C are suitable, and that option C may be enough. All we're trying to do is help users avoid mistakes. It seems like that should be as low friction as possible.

Thoughts?

devinivy:
Another reason I'd toss in there is to ensure emails aren't sent to non-existent addresses, which can affect email reputation.  Also would be nice to have some kind of paper trail for users getting signed-up without their permission.  So if we are requiring an email for signup already, I think we formally confirm the email address but allow the user to login and use the app without confirmation.  So that's a vote for B!

arcalinea:
Right now, and up through the phase where we're pulling in people off the email list, we already know we have a valid email for them. So think this is eventually necessary once we move into open access phase, but we can table it for now. It adds extra friction to a signup process that is already made complicated by invite codes!

dholms:
Worth pointing out that even though we know we have valid emails for them, there is no requirement that they signed up with the same email that the invite went out to

Mike-E-angelo:
I am currently locked out of my account as the email that I thought I used is not working to get a code sent to me.  Is there somewhere that I can go to get this resolved?  Thank you for any assistance.

bnewbold:
@Mike-E-angelo email <support@bsky.app>; it might take a while to get a response.

PR #1568, opened 2023-09-08
Email confirmation/update

dholms:
Adds two new flows that both rely on single-use tokens delivered via email.

Email confirmation
- request a token
- input the token and email in order to verify your email

Email Update
- if you have a currently verified email, then no token is required
- if you're email is verified, then a token is sent to your current email
- that token can then be used, alongside a current access token, to update your email
- when you do so, your new email will be un-verified

Closes https://github.com/bluesky-social/atproto/issues/479


packages/api/docs/labels.md
estrattonbailey, line 1: Ah, didn't think about this: code-genned and overwrote what was prettier-ified in my other PR 😬 

If you run `pnpm format` from the root I think this file will be dropped from this PR.

estrattonbailey: We could address this by updating lex-cli and fixing our prettier integration, or using a hook to lint files during commits, see #1569

dholms: yup yup no worries 👍

lexicons/com/atproto/server/updateEmail.json
```diff
@@ -0,0 +1,29 @@
...
+      "input": {
+        "encoding": "application/json",
+        "schema": {
+          "type": "object",
+          "required": ["email"],
+          "properties": {
+            "email": { "type": "string" },
+            "token": {
+              "type": "string",
+              "description": " Requires a token from com.atproto.sever.requestEmailUpdate if the account's email has been confirmed."
```
devinivy, line 17:
```suggestion
              "description": "Requires a token from com.atproto.sever.requestEmailUpdate if the account's email has been confirmed."
```

devinivy, line 17: I wonder if we'll run into issues here when a user loses access to their former, confirmed email address.  It's not uncommon when work-specific or school-specific email addresses are used to create accounts, and then the user leaves their job or school.

Just to toss out one unflesh-out alternative—the user could receive a token by providing their credentials (e.g. username and password, or in the future 2fa).  This is similar to how e.g. github has you enter "sudo mode" before adding a new email address or performing other sensitive actions.

packages/pds/src/db/migrations/20230908T224408678Z-email-tokens.ts
```diff
@@ -0,0 +1,27 @@
...
+import { Dialect } from '..'
+
+export async function up(db: Kysely<unknown>, dialect: Dialect): Promise<void> {
+  const timestamp = dialect === 'sqlite' ? 'datetime' : 'timestamptz'
+  await db.schema
+    .createTable('email_token')
+    .addColumn('purpose', 'varchar', (col) => col.notNull())
+    .addColumn('did', 'varchar', (col) => col.notNull())
+    .addColumn('token', 'varchar', (col) => col.notNull())
+    .addColumn('requestedAt', timestamp, (col) => col.notNull())
```
devinivy, line 11: We haven't really used this in the pds yet, but nice!  May be worth a quick test if you haven't already that inserting a date then selecting it out results in two equal `Date`s, particularly if local server time is not GMT.  Essentially that kysely or the pg driver doesn't muck with the times in some unanticipated way.

devinivy (approved):
Smooooth!

lexicons/com/atproto/server/confirmEmail.json
```diff
@@ -0,0 +1,27 @@
...
+          "type": "object",
+          "required": ["email", "token"],
+          "properties": {
+            "email": { "type": "string" },
+            "token": { "type": "string" }
+          }
+        }
+      },
+      "errors": [
+        { "name": "UserNotFound" },
```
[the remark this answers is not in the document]
dholms: yup good call 👍

lexicons/com/atproto/server/confirmEmail.json
```diff
@@ -0,0 +1,27 @@
...
+    "main": {
+      "type": "procedure",
+      "description": "Confirm an email using a token from com.atproto.server.requestEmailConfirmation.",
+      "input": {
+        "encoding": "application/json",
+        "schema": {
+          "type": "object",
+          "required": ["email", "token"],
+          "properties": {
+            "email": { "type": "string" },
```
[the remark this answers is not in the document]
dholms: My thinking here is that it's an extra integrity check on email.

Imagine the following flow:
- user requests email confirmation with `real-email@example.com`
- attacker updates email to `attacker-email@example.com`
- user verifies email with token
- user account now has verified email `attacker-email@example.com`

This is very unlikely. I'm not sure we need to defend against it and we may have a better mechanism by which to do it (storing that state in the db?)

packages/pds/src/db/migrations/20230908T224408678Z-email-tokens.ts
dholms: good point 👍 i'll run some tests

I went back and forth on if this should be a timestamp or string. I'm not opposed to just sticking with stirngs in the PDS (since we support both sqlite & postgres and it smooths over the difference). Thinking was "strings for records and timestamps for other stuff". But I suppose this leaves it in a pretty inconsistent state

dholms: hmmm i think i'm talking myself into making this a string 😅

lexicons/com/atproto/server/updateEmail.json
dholms: I think they'll need to go through an out-of-band support flow then

I like the alternative 👍 I don't see the current appraoch as being backwards incompatible with it tho. May be something we do when we get in 2fa.

lexicons/com/atproto/server/createSession.json
```diff
@@ -29,7 +29,8 @@
             "refreshJwt": { "type": "string" },
             "handle": { "type": "string", "format": "handle" },
             "did": { "type": "string", "format": "did" },
-            "email": { "type": "string" }
+            "email": { "type": "string" },
+            "emailConfirmed": { "type": "boolean" }
```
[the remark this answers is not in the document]
dholms: Yeah. I see what you're saying 🤔

This would be fairly easy to switch over and may save us from some future deprecation

devinivy:
Resolves #479

dholms:
note to self: make sure we reflect email validation in `updateEmail`: https://github.com/bluesky-social/atproto/pull/1658

PR #1571, opened 2023-09-08
Lower applyWrites limit

dholms:
We had a higher limit on `applyWrites` when repos had history because batch writes would prevent repo history increase.

However, now that repos are a-historical, that benefit isn't needed. So `applyWrites` is really meant for applying transactional writes.

Business logic that requires > 10 transactional records is rare


devinivy (approved):

dholms:
Decided to hold off on this, because rate limits give us the desired effect. May drop to 100 or so to keep event size down

PR #1572, opened 2023-09-09
Clean up other email tokens

dholms:
Optional addition on top of https://github.com/bluesky-social/atproto/pull/1568

Move account delete & password reset tokens to the new `email_token` table & logic.

This cleans up the routes substantially, removes password reset state from the `user_account` table and gets some of this sensitive logic all co-located.

The migration is a _little_ bit tricky here because we'll end up breaking some folks' deletes/password resets. But it will just be for a short time and it might be worth it. We can do it in off-peak hours to mitigate the impact


packages/pds/src/db/migrations/20230908T224408678Z-email-tokens.ts
```diff
@@ -12,16 +12,49 @@ export async function up(db: Kysely<unknown>, dialect: Dialect): Promise<void> {
     .addPrimaryKeyConstraint('email_token_pkey', ['purpose', 'did'])
     .execute()
 
+  // for lookups where we don't know the did
+  await db.schema.createIndex('email_token_token_idx').on('token').execute()
```
devinivy, line 16: This may need to be unique when paired with `purpose` to ensure two dids don't end-up with the same token.  I believe the `passwordResetToken` column dropped below did have this uniqueness constraint on it for that purpose.  

(Relatedly, I forget if that index/unique-constraint automatically gets dropped when the column is dropped.  If so, I wonder if the `DROP INDEX` is not done `CONCURRENTLY` and can cause table locking.)

devinivy (approved):
Worth taking a peek at that one comment re: indexing by `token`, but by and large this looks great and feels real nice to have cleaned-up.

packages/pds/src/db/migrations/20230908T224408678Z-email-tokens.ts
dholms: took care of this in a previous PR with a unique index on `(purpose, token)` 👍

devinivy (approved):

PR #1573, opened 2023-09-11
Increase timeline threshold

dholms:


devinivy (approved):
@dholms this seems fine for merge 👍 we can probably remove the build config.

PR #1575, opened 2023-09-11
Fixes and updates to the preferences API

pfrazee:
- Guards against duplicate entries in preferences and ensures consistent reads of the state when setting and getting
- Adds the "personal details" preference with the birth date that was recently added


pfrazee:
Had to add a `findLast()` declaration. None of the es version targets that our tsconfig supports includes that, but support is pretty universal in the ES ecosystem (according to mdn; I also checked RN hermes) so I feel good w/this

dholms (approved):
all looks good 👍 

one small thing: i'm not sure the order the preferences are returned is a part of the formal api contract - tho it is our our implementation does it (we explicitly order in the query)

that being said - this relieves the short term pain & fixing up the writes on the PDS should be the actual fix (and will work with either `find` or `findLast`)

pfrazee:
Published `@atproto/api@0.6.12`

PR #1576, opened 2023-09-11
Prevent user pref races

dholms:
We were getting racing updates to the `putPreferences` route that would leave the table in an inconsistent state.

This applies a row lock on the user's row in the `user_account` table to ensure that these updates get linearized. This should be a short-held lock and will block other writes while it is held.

---

This seemed like the simplest implementation to prevent conflicting writes. Other options include:
- advisory lock (similar to repo writes)
- skipping on lock and returning an error instead of waiting & overwriting
- skipping the lock & retrying
- doing integrity checks at the end of the transaction to ensure we didn't leave it in a bad state & throwing if we did
- reworking the data modeling of the user-pref table so that it's a simple atomic `UPDATE`

### DIFF
--- a/packages/pds/src/services/account/index.ts
+++ b/packages/pds/src/services/account/index.ts
@@ -590,6 +590,15 @@ export class AccountService {
         `Some preferences are not in the ${namespace} namespace`,
       )
     }
+    // short-held row lock to prevent races
+    if (this.db.dialect === 'pg') {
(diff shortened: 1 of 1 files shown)


devinivy (approved):
These should be rare enough that it's a non-issue 👍 May be worth keeping an eye on, but I think it should be good.

PR #1577, opened 2023-09-11
Get rate limit ip correctly

dholms:
To get the ip of the client for rate limiting correctly we need to do 2 things:
- tell express to trust the proxy that it's receiving the req from (in this case the loadbalancer)
- grab the _leftmost_ (not _rightmost_) ip from the XFF header. `req.ip` does this

for reference: https://expressjs.com/en/guide/behind-proxies.html

### DIFF
--- a/packages/xrpc-server/src/rate-limiter.ts
+++ b/packages/xrpc-server/src/rate-limiter.ts
@@ -14,7 +14,6 @@ import {
   RateLimiterStatus,
   XRPCReqContext,
 } from './types'
-import { getReqIp } from './util'
 
 export type RateLimiterOpts = {
   keyPrefix: string
@@ -155,5 +154,5 @@ export const getTightestLimit = (
   return lowest
 }
 
-const defaultKey: CalcKeyFn = (ctx: XRPCReqContext) => getReqIp(ctx.req)
+const defaultKey: CalcKeyFn = (ctx: XRPCReqContext) => ctx.req.ip
 const defaultPoints: CalcPointsFn = () => 1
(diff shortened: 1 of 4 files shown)


packages/xrpc-server/src/rate-limiter.ts
```diff
@@ -155,5 +154,5 @@ export const getTightestLimit = (
   return lowest
 }
 
-const defaultKey: CalcKeyFn = (ctx: XRPCReqContext) => getReqIp(ctx.req)
+const defaultKey: CalcKeyFn = (ctx: XRPCReqContext) => ctx.req.ip
```
devinivy, line 157: Perhaps we drop a note nearby here re: `app.set('trust proxy', true)`?

devinivy (approved):

packages/xrpc-server/src/util.ts
```diff
@@ -268,10 +268,6 @@ function decodeBodyStream(
   return stream
 }
 
-export const getReqIp = (req: express.Request): string => {
-  return req.ips.at(-1) ?? req.ip
```
devinivy, line 272: Part of the issue here is that `req.ips` was empty since `'trust proxy'` was `false`, so fell back to `req.ip`, which was the raw socket address (again since `'trust proxy'` was `false`).

packages/xrpc-server/src/rate-limiter.ts
dholms: yeah good call 👍

packages/xrpc-server/src/util.ts
dholms: yup totally 👍 

altho this would still need to grab the first ip in the array i believe (`req.at(0)`) which is exactly what `req.ip` does so i figured i'd just simplify it

PR #1589, opened 2023-09-12
:sparkles: Send HTML content in email

foysalit:
Using a very simple regex, check if the email content contains any html tags in it and send it as html, otherwise, fallback to plain-text content

### DIFF
--- a/packages/pds/src/mailer/moderation.ts
+++ b/packages/pds/src/mailer/moderation.ts
@@ -16,12 +16,23 @@ export class ModerationMailer {
     this.transporter = transporter
   }
 
+  isContentHtml(content: string): boolean {
+    return /<\/?[a-z][\s\S]*>/i.test(content)
+  }
+
   async send({ content }: { content: string }, mailOpts: Mail.Options) {
-    const res = await this.transporter.sendMail({
+    const mail = {
       ...mailOpts,
-      text: content,
       from: this.config.moderationEmailAddress,
-    })
+    }
+
+    if (this.isContentHtml(content)) {
+      mail.html = content
+    } else {
+      mail.text = content
+    }
+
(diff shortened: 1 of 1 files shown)


bnewbold:
As a chronic user of text email (instead of HTML), it is nice when automated emails like this get sent as both `text/plain` and `text/html`. Markdown can go in `text/plain` and that would be what is expected in the content of the API request, with the server doing Markdown-to-HTML.

Alternatively, the Lexicon endpoint could have text and html fields to pass-through from client.

Or, have a `contentType` field in the Lexicon which can indicate if text or HTML (or markdown?); that wouldn't make sending both easy, but it might be better than auto-detecting HTML, which feels brittle.

All the above are ideas and could be done later; I wouldn't block merging this sniffing PR as-is.

devinivy:
There is also a plugin for nodemailer, used on the server, that automatically generates the plaintext version of an email given an html email.  I pretty much always use it when I'm using nodemailer!  https://www.npmjs.com/package/nodemailer-html-to-text

devinivy:
Ohp, just saw the PR to author emails in markdown on the client!  In which case yeah, I think the move might be to:

 - author all emails in markdown on the client.
 - compile email markdown to html on the client, and treat all emails as html on the server (or explicitly indicate the content type, as @bnewbold suggested).
 - use the `nodemailer-html-to-text` plugin to provide a plaintext version of all emails.

foysalit:
@devinivy ok so I think treating all incoming content as html is the easiest option and I am not seeing a case where we would need to explicitly send plaintext content so adding `contentType` param just to facilitate that feels like unnecessary bloat right now.

foysalit:
@devinivy can I get one more look at this please? thanks!

devinivy (approved):

PR #1592, opened 2023-09-12
order by `like.indexedAt` in app view

### LINKED ISSUE #1574: getActorLikes should include like timestamp and should sort by that

Imagine the situation:

* one of the devs posts a real banger that I somehow miss
* in the year of our Lord 2027, I'm browsing Bluesky on my brand new Apple Vision Pro Series 3 and I come across someone reposting it; I press "like"
* the next morning, I remember about the post and I want to send it to a friend
* I open my "likes" tab, however, since the posts there are sorted by post creation date, not by like date, and this one is from 4 years ago, it will be positioned thousands of pages down, so I will never find it there

* the items in the JSON returned from `getActorLikes` should include the `createdAt` timestamp of the *like* record
* the list should be sorted by that timestamp, so something that I liked a moment ago should be at the top of the list, even if it's a post from 4 years ago

estrattonbailey:
Fixes #1574

Sort by `like.sortAt` instead of `feed_item.sortAt` **in app view only**.

### DIFF
--- a/packages/bsky/src/api/app/bsky/feed/getActorLikes.ts
+++ b/packages/bsky/src/api/app/bsky/feed/getActorLikes.ts
@@ -1,4 +1,3 @@
-import { sql } from 'kysely'
 import { InvalidRequestError } from '@atproto/xrpc-server'
 import { Server } from '../../../../lexicon'
 import { QueryParams } from '../../../../lexicon/types/app/bsky/feed/getActorLikes'
@@ -71,7 +70,6 @@ const skeleton = async (
   let feedItemsQb = feedService
     .selectFeedItemQb()
     .innerJoin('like', 'like.subject', 'feed_item.uri')
-    .select('like.indexedAt as sortAt')
     .where('like.creator', '=', actorDid)
 
   const keyset = new FeedKeyset(ref('like.sortAt'), ref('feed_item.cid'))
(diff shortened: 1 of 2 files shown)


packages/bsky/src/api/app/bsky/feed/getActorLikes.ts
```diff
@@ -70,9 +71,10 @@ const skeleton = async (
   let feedItemsQb = feedService
     .selectFeedItemQb()
     .innerJoin('like', 'like.subject', 'feed_item.uri')
+    .select('like.indexedAt as sortAt')
```
dholms, line 74: ah all the indexed record tables in the appview actually already have a sortAt column!

https://github.com/bluesky-social/atproto/blob/main/packages/bsky/src/db/tables/like.ts

estrattonbailey: Smh ooooh I see

packages/bsky/src/api/app/bsky/feed/getActorLikes.ts
```diff
@@ -72,7 +72,7 @@ const skeleton = async (
     .innerJoin('like', 'like.subject', 'feed_item.uri')
     .where('like.creator', '=', actorDid)
 
-  const keyset = new FeedKeyset(ref('feed_item.sortAt'), ref('feed_item.cid'))
+  const keyset = new FeedKeyset(ref('like.sortAt'), ref('feed_item.cid'))
```
dholms, line 75: this probably works, but keeping the secondary cursor on the same table can make pagination faster (it can stay in the index):

```suggestion
  const keyset = new FeedKeyset(ref('like.sortAt'), ref('like.cid'))
```

dholms (approved):
nice lookin good 👍

PR #1593, opened 2023-09-13
Including reposted replies in `getAuthorFeed?posts_no_replies`

estrattonbailey:
Helps https://github.com/bluesky-social/social-app/issues/1371

However, self-replies that you repost can sometimes still get threaded by frontend logic. So this PR will need a FE counterpart to fully fix this issue.


estrattonbailey:
We can do this entirely on the frontend, see https://github.com/bluesky-social/social-app/pull/1450

PR #1594, opened 2023-09-13
proposed new search lexicons

bnewbold:
First pass at post search and backend skeleton Lexicons to support search iteration.

The existing searchActors and searchActorsTypeahead are slightly tweaked (default result size, descriptions), in a way I think is safe and backwards compatible.

A bit of tension around how similar to keep searchActors and searchPosts. I think having the query param be `q` for search is really ingrained in API design, so I went with an inconsistent value there. Could do searchProfiles instead? Or maybe sticking with `terms` (like searchActors) is the right move.

I didn't end up using the exact `app.bsky.feed.defs#skeletonFeedPost` from feed skeletons, just a list of ATURIs, because we don't have a "reason" in search.

I'd like us to stick with limit/offset queries in to opensearch (elasticsearch), and not do full scroll cursors. This can be done with the existing cursor setup, by having the search service stick an offset+limit number in the cursor string, i'm just mentioning it as a "what are we trying to provide with this searchPosts endpoint".

It is relatively cheap for opensearch to do limit/offset up to a result set of a few thousands hits. The cursor/scroll mode allows fast scrolling through the entire index (eg, billions of docs), but has per-cursor overhead in most situations and we don't want that. Basically, this API should support common search cases, and not be a defacto public API for enumerating our full index (folks who want to do deep investigation/research should run their own mirror cluster/index; we can access our opensearch cluster directly internally if we want that).

I'm guessing that we will add additional params to these over time, but want to start relatively simple.


lexicons/app/bsky/unspecced/searchActorsSkeleton.json
```diff
@@ -0,0 +1,56 @@
...
+      "description": "Backend Actors (profile) search, returning only skeleton",
+      "parameters": {
+        "type": "params",
+        "required": ["q"],
+        "properties": {
+          "q": {
+            "type": "string",
+            "description": "search query string; syntax, phrase, boolean, and faceting is unspecified, but Lucene query syntax is recommended. For typeahead search, only simple term match is supported, not full syntax"
+          },
+          "typeahead": {
```
dholms, line 16: would it make sense to use a non-normative parameter for this? something like `simple` or `quick`?

lexicons/app/bsky/actor/searchActorsTypeahead.json
```diff
@@ -8,12 +8,15 @@
...
+          "term": {
+            "type": "string",
+            "description": "search query prefix; not a full query string"
+          },
           "limit": {
             "type": "integer",
             "minimum": 1,
             "maximum": 100,
-            "default": 50
+            "default": 10
```
dholms, line 19: nice i dig both of these new defaults 👍

lexicons/app/bsky/actor/searchActors.json
```diff
@@ -4,16 +4,19 @@
   "defs": {
     "main": {
       "type": "query",
-      "description": "Find actors matching search criteria.",
+      "description": "Find actors (profiles) matching search criteria.",
       "parameters": {
         "type": "params",
         "properties": {
-          "term": { "type": "string" },
+          "term": {
```
dholms, line 11: I would support phasing out `term` and phasing in `q`

I think "term" makes sense when it's simple search on a word & no additional syntax is allowed

`q` fits nicely when we allow for special query syntax (like Lucene syntax)

dholms:
These look good and make a lot of sense to me 👍 

Just chatting about a few things from the PR description:

> A bit of tension around how similar to keep searchActors and searchPosts

I like the idea of keeping these relatively similar - I'm down with moving to `q` over `term`. 

> I didn't end up using the exact app.bsky.feed.defs#skeletonFeedPost from feed skeletons, just a list of ATURIs, because we don't have a "reason" in search.

One question is: will we ever want any extra metadata on an object for why it turned up in search? something like "trending down" or "you follow this user"? If so, we may not want to use `#skeletonFeedPost` but we could define a new container schema like `#skeletonSearchPost` that would give us the leeway to add additional metadata

> I'd like us to stick with limit/offset queries in to opensearch (elasticsearch), and not do full scroll cursors. This can be done with the existing cursor setup, by having the search service stick an offset+limit number in the cursor string, 

Makes sense to me 👍  And strong agree that we do not want to support full index pagination

lexicons/app/bsky/unspecced/searchActorsSkeleton.json
bnewbold: hmmm, I feel like typeahead is pretty term-of-art for this query type (not just the term we are using). different sub-set of fields are queried, and only really works for prefix of a single token or two, character-by-character.

"simple" or "quick" would be confusing to me, i'd assume that they would return un-hydrated responses or something (responses are always hydrated).

lexicons/app/bsky/actor/searchActors.json
bnewbold: cool, I updated all the instances of 'term' as a search query with 'q', and marked the old 'term' fields as "DEPRECATED" in description. This impacted an admin route as well. Idea is that for a short transition we'll fall back to "term" if "q" is empty, and maybe eventually nuke them (if/when we are feeling comfortable breaking query params in Lexicons in a small way, like a v1.0 release)

bnewbold:
w/r/t `#skeletonSearchPost` or similar: that does make sense as a good Lexicon design pattern, but I can't really think of any reason we would want that for search specifically in the skeleton stage. The views that get returned will have stuff like "do I follow this account" or "have I liked post" hydrated just like a feed, and also things like blocks and mutes will get enforced. The AppView has all that extra context, the search index is pretty dumb.

I'd lean towards keeping it simple (makes for simpler code), but don't feel super strongly and could add wrapper objects if you think we should be consistent.

lexicons/app/bsky/unspecced/searchActorsSkeleton.json
dholms: cool cool - i'm game for "typeahead" 👍

dholms:
Makes sense - I think it's fine as is then 👍

bnewbold:
Ran codegen, and naively prefer new `q` over `term` for older endpoints (but fallback to `term` if `q` is not set).

cc: @devinivy if you might have any last-minute Lexicon style vibes

lexicons/app/bsky/unspecced/searchActorsSkeleton.json
```diff
@@ -0,0 +1,56 @@
...
+          "limit": {
+            "type": "integer",
+            "minimum": 1,
+            "maximum": 100,
+            "default": 25
+          },
+          "cursor": {
+            "type": "string",
+            "description": "optional pagination mechanism; may not necessarily allow scrolling through entire result set"
+          }
```
devinivy, line 29: In the past it's proven useful to not support a cursor for typeahead search.  Just wanted to note that in case it's relevant here, since both `typeahead` and `cursor` parameters can appear together.

lexicons/app/bsky/unspecced/searchActorsSkeleton.json
```diff
@@ -0,0 +1,56 @@
...
+              "type": "integer",
+              "description": "count of search hits. optional, may be rounded/truncated, and may not be possible to paginate through all hits"
+            },
+            "actors": {
+              "type": "array",
+              "items": {
+                "type": "string",
+                "format": "did"
+              }
+            }
```
devinivy, line 49: Not terribly worried about it, but there could be some value in wrapping the did in an object, just so that there's an upgrade path in case we want to provide any additional/meta info with each result in the future.

lexicons/app/bsky/unspecced/searchPostsSkeleton.json
```diff
@@ -0,0 +1,52 @@
...
+              "type": "integer",
+              "description": "count of search hits. optional, may be rounded/truncated, and may not be possible to paginate through all hits"
+            },
+            "posts": {
+              "type": "array",
+              "items": {
+                "type": "string",
+                "format": "at-uri"
+              }
+            }
```
devinivy, line 45: Same goes here, there could be some value in wrapping the uri in an object so that we leave open an upgrade path to add additional/meta info about each result in the future.

lexicons/app/bsky/unspecced/searchActorsSkeleton.json
bnewbold: Dan had the same thoughts in earlier review, which is strong enough signal that i'll add that to these endpoints.

lexicons/app/bsky/unspecced/searchActorsSkeleton.json
bnewbold: Would the alternative be a second endpoint just for typeahead? That feels expansive to me. I think in typeahead mode we can safely ignore the `cursor` param and never populate that field in the output.

bnewbold:
updated with skeletal objects, and re-codegen

viksit:
@bnewbold how are you running the search in the backend for posts? are you using pgsql FTS or something else?

bnewbold:
@viksit the backend search service is `palomar`, implemented in golang: https://github.com/bluesky-social/indigo/tree/main/cmd/palomar

The actual index/engine is opensearch, the AWS fork of elasticsearch.

PR #1606, opened 2023-09-15
remove cruft packages (uri, nsid, identifier)

bnewbold:
We merged all these in to `@atproto/syntax` a few weeks back, and I think we can/should just rip these out now to prevent confusion.

Some related ecosystem cleanups:

- [x] publish updated packages to npm for things which used to depend on these removed packages
- [x] mark the deleted packages as deprecated in npm, with a note indicating the new packages
- [ ] maybe other ecosystem deprecation best practices i'm not aware of

Looking at the dependency tree in, eg, `social-app`, I think the crucial packages to publish updates for is `@atproto/lexicon`. `lex-cli` would also be good to hit if easy.

@estrattonbailey, having done some recent packaging improvements, could you review this plan and help with these later bits?

I'm trying to think if we need to do any dev comms or outreach. If we do, should have the package dependency tree updated and clean first, I think. As long as we don't actually yank the old packages I don't think we'll break anybody.


estrattonbailey:
Nice! Yeah sounds like a plan.

I think a good move would be to update readmes for these packages first to link out to the `syntax` package, maybe even a "migration" section of the `syntax` readme. Then we can patch the deprecated package versions as well as `syntax` to publish the packages with the updated docs. Then deprecation is a manual step in the npm dashboard for each package, pretty easy.

Then come back here and remove all these files as a last step.

In terms of updating downstream packages, that seems reasonable if there aren't to many. If there are no breaking changes resulting from our reorg of this code then it's probably not critical, but some might consider it the neighborly thing to do if we have bandwidth.

estrattonbailey:
Update: realized that we can just specify a message to show on the npm package page from the npm dashboard, so no real need to edit those readmes, nvm! Closed #1610 that was for doing that.

estrattonbailey:
>Looking at the dependency tree

Maybe we should just do a publish to get us to square-one with any packages that haven't been published recently 🤔

dholms (approved):
looks good 👍 

happy to get these out of here

Feel free to merge whenever - @estrattonbailey are you up for ensuring downstream packages are published correctly? Happy to hop in on any npm stuff that needs me

estrattonbailey:
Wanted to double check some things related to pnpm/changesets. While I was at it I fixed the conflicts, hence the rebase and force push 👍

estrattonbailey (approved):
Sweet, I think we're good to go here. Just need to mark the deleted packages as deprecated in npm. Can chat in Slack about what message we should include with the deprecations.

PR #1610, opened 2023-09-15
deprecate packages and update docs

estrattonbailey:
This will patch `nsid`, `uri` and `identifier`, as well as `syntax`, and bump any packages that depend on `syntax` as well.

Once these are published with the deprecation notice, we can deprecate them in npm as well.


estrattonbailey:
Oh I just realized that we can specify a deprecation _message_ in npm too. So no real reason to publish new docs 👍

PR #1617, opened 2023-09-16
Evented architecture for moderation system

foysalit:
This PR is work-in-progress.

### High level summary
The goal of this PR is to migrate the moderation system to a more signal/event based info architecture. In the initial implementation, we opted for a report driven UX where users/automated systems can "report" a piece of content and a human/automated moderator would take an "action" that determines the final "moderated state" of the content where there can only be one action for a content at a given time. This PR, allows an incoming stream of events on a content and an event can have a side-effect which can change the "moderation state" of the content. A few examples to more concretely describe this model

1. a user report on a post/account flags the content as "requires review" for moderators.
2. a moderator taking an "escalate" action emits an event which flags the content as "escalated review" for moderators.
3. a moderator taking a "takedown" action emits an event which flags the content as "reviewed" and marks the content as taken down.
4. a moderator emitting a "comment" event just adds a comment to the content without impacting the state of the content.

### Primary lexicon changes
This PR introduces 3 new xrpc methods

#### `getModerationStatuses()`
Fetches a list of "current moderation state" for all "subjects" (subject being any record) with a wide array of input params that helps filter out the list to get a certain set of subject/s. The data itself is retrieved from a new `moderation_subject_status` table. Few notes on this
1. Right now, the input params are super flexible and not everything is used through ozone. We can/will probaly tighten up the schema a bit down the road. For instance, when we need to fetch the moderation status for just an account or a post, we are calling this method with the `subject` param when a more ideal setup would be a more dedicated `getModerationStatus` xrpc method that only takes a subject and returns the state of that subject alone.
2. We could consider hydrating the response for this with repoView/recordView to make it a bit more "complete" but so far, it hasn't been necessary.

#### `emitModerationEvent()`
Logs a moderation event on a subject and applies the necessary side-effects. Mostly a modified version of `takeModerationAction` method The event itself is stored in a new table `moderation_event` which is quite similar to the previous `moderation_action` table.  Both reports and actions are stored in this table. Side effects like labels and takedowns continue to work just like they used to through previous `takeModerationAction` xrpc method.

#### `getModerationEvents()`
Fetches a list of moderation events on a given subject. Allows filtering through a couple of handy params.

### New features
As part of this new architecture, this PR also adds a couple of new features that have been requested over time.

1. Mute a subject globally. Up until now, we worked around this with a client side hack that allowed a moderator to "snooze" a subject but the "snooze" was scoped to browser only.
2. Keep history of sent emails.
3. Subject based moderation queue instead of reports.

TODO:
- [x] Migration script to take current actions and reports and consolidate them into the new event table and generate the subject status
- [x] Implement proxying to appview through PDS for new xrpc methods
- [x] Resolve conflicts with main


packages/bsky/src/db/migrations/20230927T202833377Z-create-moderation-subject-status.ts
```diff
@@ -0,0 +1,55 @@
...
+    .execute()
+  await db.schema
+    .createTable('moderation_subject_status')
+    .addColumn('id', 'serial', (col) => col.primaryKey())
+    .addColumn('subjectType', 'varchar', (col) => col.notNull())
+    .addColumn('subjectDid', 'varchar', (col) => col.notNull())
+    .addColumn('subjectUri', 'varchar')
+    .addColumn('subjectCid', 'varchar')
+    .addColumn('createdAt', 'varchar', (col) => col.notNull())
+    .addColumn('updatedAt', 'varchar', (col) => col.notNull())
```
bnewbold, line 29: will probably want an index on `updatedAt` also (to have a view/query of "recent mod actions", and possibly for pagination), but could wait until we have an actual need to create it

packages/bsky/src/db/migrations/20230927T202833377Z-create-moderation-subject-status.ts
```diff
@@ -0,0 +1,55 @@
...
+  await db.schema
+    .alterTable('moderation_action')
+    .addColumn('meta', 'jsonb')
+    .execute()
+  await db.schema
+    .createTable('moderation_subject_status')
+    .addColumn('id', 'serial', (col) => col.primaryKey())
+    .addColumn('subjectType', 'varchar', (col) => col.notNull())
+    .addColumn('subjectDid', 'varchar', (col) => col.notNull())
+    .addColumn('subjectUri', 'varchar')
```
bnewbold, line 26: it feels to me like these columns are a bit verbose in terms of database disk (and, in particular, index and pagecache) use.

eg, i'd probably break out URI in to separate collection and rkey columns (or just a combined `path` column), so that the DID doesn't need to be stored again. subjectType could be skipped/inferred based on whether `collection` is null or not (if null, it's a report on a repo, not a record). or an even bigger refactor would be separate tables for status-on-repo and status-on-record.

this is all kind of fiddly optimization and could wait till later. and defer to devin on more local style/norms in this repo.

bnewbold (commented):
In this iteration, the current `status` of the subject still rolls up a lot of state into a single named state. What if we split it out more like this:

```text
# informal SQL table sketch
mod_state
  # primary key: (did, recordPath, recordCid)
  did (string or foreign key to identity table, required)
  recordPath (string, nullable): "<collection>/<rkey>"
  recordCid (string, nullable)

  # report state
  lastReportedAt (timestamp, nullable)

  # human review team state
  reviewedAt (timestamp, nullable)
  reviewState (enum, default=open): open, escalated, closed
  muteUntil (timestamp, nullable)
  summaryNote (string, nullable): persistent account note (different from per-action comments; eg "looks suspicious but not actually impersonation" or "legal hold on account, talk to XYZ")

  # visibility/intervention state
  takendown (bool, default=false)
  suspendUntil (timestamp, nullable)

  # more columns in the future? speculative
  flags (array strings): like mod-internal hashtags or labels
  modAnnotation (string): separate from labels, a way to put a public annotation next to a post or account? maybe not.
  extra (jsonb, nullable): for other free-form info?
```

Basically we have several distinct pieces of state: "where is this in the review flow diagram" (possibly re-opened, etc) is distinct from "what was the most recent decision taken". Most events will only impact some of the columns, not all of them. Eg, a new report would probably touch the `lastReportedAt`, but never `takedown`.

The more I think about it, the more I would lean towards renaming "action" to "event". Things like a new report aren't an action, and I think that could be confusing.

packages/bsky/src/db/migrations/20231003T202833377Z-create-moderation-subject-status.ts
```diff
@@ -0,0 +1,72 @@
...
+    .addColumn('lastReportedAt', 'varchar')
+
+    // visibility/intervention state
+    .addColumn('takendown', 'boolean', (col) => col.defaultTo(false).notNull())
+    .addColumn('suspendUntil', 'varchar')
+
+    // timestamps
+    .addColumn('createdAt', 'varchar', (col) => col.notNull())
+    .addColumn('updatedAt', 'varchar', (col) => col.notNull())
+    .addUniqueConstraint('did_record_path_unique_idx', ['did', 'recordPath'])
```
devinivy, line 51: Since constraints all share a namespace in postgres, I think we've been prefixing them with the table name.  You'll find some examples if you search for `addUniqueConstraint` in the codebase.

packages/bsky/src/db/migrations/20231003T202833377Z-create-moderation-subject-status.ts
```diff
@@ -0,0 +1,72 @@
...
+    .execute()
+}
+
+export async function down(db: Kysely<unknown>): Promise<void> {
+  await db.schema.alterTable('moderation_event').renameTo('moderation_action')
```
devinivy, line 56: This one needs an `execute()` on it.

lexicons/com/atproto/admin/defs.json
```diff
@@ -93,16 +89,35 @@
...
+        "#takedown",
+        "#flag",
+        "#acknowledge",
+        "#escalate",
+        "#comment",
+        "#label",
+        "#revert",
+        "#mute"
+      ]
     },
```
devinivy, line 121: Does the action type imply what content is allowed in the action?  E.g. can you include labels in a `#mute` event even though there's also a `#label` event?  If so, we might consider turning actions into an open union.

lexicons/com/atproto/admin/defs.json
```diff
@@ -93,16 +89,35 @@
...
     "actionType": {
       "type": "string",
-      "knownValues": ["#takedown", "#flag", "#acknowledge", "#escalate"]
+      "knownValues": [
+        "#takedown",
+        "#flag",
+        "#acknowledge",
+        "#escalate",
+        "#comment",
+        "#label",
```
devinivy, line 117: As an example, if a moderator adds a comment and a label at the same time then clicks "save", what event(s) are generated, and are those two events bundled together in any way?

lexicons/com/atproto/admin/defs.json
```diff
@@ -93,16 +89,35 @@
...
       }
     },
+    "actionMeta": {
+      "type": "object",
+      "properties": {
+        "resolveReportIds": { "type": "array", "items": { "type": "integer" } },
+        "reportType": {
+          "type": "ref",
+          "ref": "com.atproto.moderation.defs#reasonType"
+        }
```
devinivy, line 106: I am interested to hear a little more about how to think about these two fields as a part of action metadata.

lexicons/com/atproto/admin/defs.json
```diff
@@ -93,16 +89,35 @@
...
+      }
+    },
     "actionType": {
       "type": "string",
-      "knownValues": ["#takedown", "#flag", "#acknowledge", "#escalate"]
+      "knownValues": [
```
devinivy, line 111: In some of the codegen I see a `#report` type—should it be listed here too?

packages/pds/src/services/moderation/index.ts
```diff
@@ -420,9 +418,9 @@ export class ModerationService {
       .updateTable('moderation_action')
       .where('id', '=', id)
       .set({
-        reversedAt: createdAt.toISOString(),
-        reversedBy: createdBy,
-        reversedReason: reason,
+        // reversedAt: createdAt.toISOString(),
+        // reversedBy: createdBy,
+        // reversedReason: reason,
```
devinivy, line 423: What's the fate of these lines?  Ignore me if you already have this tracked/planned 😏

packages/bsky/src/services/moderation/status.ts
```diff
@@ -0,0 +1,272 @@
...
+  REPORT,
+  ESCALATE,
+  REVERT,
+  TAKEDOWN,
+  MUTE,
+]
+
+// TODO: How do we handle revert? for "revert" event we will have a reference event id that is being reversed
+// We will probably need a helper that can take a list of events and compute the final state of the subject
+// That helper will have to be invoked here with all events up until the point where the reverted event was created
```
devinivy, line 33: Is the idea that history would be replayed with the reverted event missing?  I wonder if we should think on this a bit and see if we can make our lives a little simpler for reverts if we can.

packages/bsky/src/services/moderation/status.ts
```diff
@@ -0,0 +1,272 @@
...
+      return {
+        takendown: true,
+        reviewState: REVIEWCLOSED,
+        lastReviewedAt: new Date().toISOString(),
+      }
+    case MUTE:
+      return {
+        // By default, mute for 24hrs
+        muteUntil: new Date(
+          Date.now() + (durationInHours || 24) * 60 * 60 * 1000,
```
devinivy, line 74: We should have some constants kicking around in the @atproto/common package `DAY`/`HOUR` for the number of milliseconds in a day/hour.

devinivy (commented):
Not a full review, just dropping some comments and questions on the in-progress work 👍 Good stuff.

packages/bsky/src/db/migrations/20231003T202833377Z-create-moderation-subject-status.ts
foysalit: I vaguely remember there was a length limit on constraint keys so probably was subconsciously trying to keep it brief. Set a new name with the table prefix, seems fine :)

lexicons/com/atproto/admin/defs.json
foysalit: I want to keep this flexible for now and not apply too much validation based on action types just yet. Some of these props are kinda composable. for instance, I could #acknowledge for a certain period of time and the action will be reverted after that period or #mute AND apply a label. some are not so composable for instance a #revert with a expiry probably doesn't make too much sense.
so overall, there should be some specific combinations that are not allowed but I haven't thought too much into it yet.

packages/bsky/src/services/moderation/status.ts
foysalit: yeah I'm not too clear on this either. The current implementation carries the concept of `status impacting event`which helps us find the last event that could have changed the status of the subject BEFORE the reverted event and "replays" that event to produce the side-effect. but it gets really hairy if we think about potential multiple reverts and removing effects of events since the reverted event. 
Maybe we can limit this to not allow reverting specific event and ONLY allow reverting "the last status impacting event" which seems practical but I need to think through the potential edge cases around this.

packages/pds/src/services/moderation/index.ts
foysalit: I was just trying to shut the linter up here. this whole thing is gonna go away since we won't be updating any existing event but rather add a new `#revert` event in the stream negating the effects of the event being reverted.

lexicons/com/atproto/admin/defs.json
foysalit: `resolveReportIds` will go away since we don't need to keep track of individual reports being resolved.
`reportType` will stay and store the metadata for `#report` events only where the user specifies the `reportType` on the client.

lexicons/com/atproto/admin/sendEmail.json
```diff
@@ -9,11 +9,12 @@
...
         "schema": {
           "type": "object",
-          "required": ["recipientDid", "content"],
+          "required": ["recipientDid", "content", "senderDid"],
           "properties": {
             "recipientDid": { "type": "string", "format": "did" },
             "content": { "type": "string" },
-            "subject": { "type": "string" }
+            "subject": { "type": "string" },
+            "senderDid": { "type": "string", "format": "did" }
```
bnewbold, line 17: maybe `creatorDid` or `requesterDid` is clearer? this don't impact who the email appears to come from for the recipient, just what is recorded in the mod event (I assume)

lexicons/com/atproto/admin/defs.json
```diff
@@ -368,6 +380,145 @@
...
+    },
+    "modEventComment": {
+      "type": "object",
+      "description": "Add a comment to a subject",
+      "required": ["comment"],
+      "properties": {
+        "comment": {
+          "type": "string"
+        },
+        "persistNote": {
```
bnewbold, line 433: maybe `sticky` as the name for this flag?

lexicons/com/atproto/admin/defs.json
```diff
@@ -368,6 +380,145 @@
...
+          "description": "Indicates how long the takedown should be in effect before automatically expiring."
+        },
+        "expiresAt": {
+          "type": "string",
+          "format": "datetime",
+          "description": "Indicates at what time the subject's takedown will be/was reverted."
+        }
+      }
+    },
+    "modEventReverseTakedown": {
```
bnewbold, line 415: It kind of feels like there could be just a single `modEventTakedown` with a boolean flag. reversing would be emitting that even with takedown=false. Same with mute. This would align with how we negate labels.

I could go either way on it though, it's ok how it is now as well.

bnewbold (commented):
I did a review pass over the lexicons. Overall looking pretty good!

Left some polish comments with possible renaming and a couple places that descriptions would be good in the lexicon. Hopefully it isn't too much churn to search/replace names if you agree with the wording, or just skip if you disagree.

Something about "emitModerationEvent" feels weird from a wording standpoint, feels a bit inside baseball (eg, could be confusing to a dev just trying to write a quick bot). But I haven't thought of a good alternative. Maybe just "create"?

One area the API doesn't feel super consistent is comments when creating events. It feels like it would be simplest if every event could have a comment on it, but only the `#comment` event could turn it in to a persistent note/annotation on the overall subject? Currently, eg, "mute" doesn't take a comment but "unmute" does.

lexicons/com/atproto/admin/defs.json
```diff
@@ -368,6 +380,145 @@
...
+      "properties": {
+        "comment": { "type": "string" }
+      }
+    },
+    "modEventMute": {
+      "type": "object",
+      "description": "Mute incoming reports on a subject",
+      "required": ["durationInHours"],
+      "properties": {
+        "durationInHours": {
```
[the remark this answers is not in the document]
foysalit: we only use these for takedowns and mutes, I'm not convinced that we need the "second" level granular control for those? 
I think the `expiresAt` thing was a leftover, will clean it up, good catch!

lexicons/com/atproto/admin/getModerationEvents.json
```diff
@@ -1,13 +1,25 @@
...
         "type": "params",
         "properties": {
+          "type": {
+            "type": "string"
+          },
+          "createdBy": {
+            "type": "string",
+            "format": "did"
+          },
+          "sortDirection": {
```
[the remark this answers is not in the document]
foysalit: Correct. added the description.

lexicons/com/atproto/admin/getModerationEvents.json
```diff
@@ -1,13 +1,25 @@
...
   "defs": {
     "main": {
       "type": "query",
-      "description": "List moderation actions related to a subject.",
+      "description": "List moderation events related to a subject.",
       "parameters": {
         "type": "params",
         "properties": {
+          "type": {
+            "type": "string"
```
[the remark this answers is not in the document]
foysalit: added description. agreed that `type` is a bit generic but also `eventType` feels a bit redundant here. I was considering `name` in both lexicon and db column, wdyt?

lexicons/com/atproto/admin/getModerationEvents.json
```diff
@@ -1,13 +1,25 @@
...
+          "createdBy": {
+            "type": "string",
+            "format": "did"
+          },
+          "sortDirection": {
+            "type": "string",
+            "default": "desc",
+            "enum": ["asc", "desc"]
+          },
           "subject": { "type": "string" },
```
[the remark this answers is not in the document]
foysalit: yeah it can be at-uri or a did. 
wait so `"format": "uri"` can also take in dids? or does it have to be fully qualified at-uri?

lexicons/com/atproto/admin/sendEmail.json
foysalit: hmm... idk, neither of those sound better/clearer than `senderDid` in the context tbh.

packages/bsky/src/services/moderation/status.ts
```diff
@@ -0,0 +1,219 @@
+// This may require better organization but for now, just dumping functions here containing DB queries for moderation status
```
foysalit, line 1: I know we're not using functional style in most places other than util functions so may be this needs to be converted to a class to align with other services?

packages/bsky/tests/admin/get-record.test.ts
```diff
@@ -69,15 +64,15 @@ describe('admin get record view', () => {
     })
   })
 
-  it('gets a record by uri, even when taken down.', async () => {
+  it.skip('gets a record by uri, even when taken down.', async () => {
```
devinivy, line 67: I can't account for why this issue would turn-up all of a sudden, but try adding a `await network.processAll()` after the `basicSeed()` call above.  That ensures that writes have propagated from the pds through the appview—I would expect it to already have been there!

foysalit: goddamn that did the trick!

lexicons/com/atproto/admin/getModerationEvents.json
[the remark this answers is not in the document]
bnewbold: `format: at-uri` specifies that the field must be an AT-URI. `format: uri` can be an AT-URI, DID, or even `https://` (pretty flexible, but some restrictions vs "any string")

lexicons/com/atproto/admin/defs.json
```diff
@@ -368,6 +380,145 @@
...
+          "type": "array",
+          "items": { "type": "string" }
+        },
+        "negateLabelVals": {
+          "type": "array",
+          "items": { "type": "string" }
+        }
+      }
+    },
+    "modEventFlag": {
```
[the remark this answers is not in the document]
devinivy: I'm all for removing it until we've nailed the use-case, personally!  I originally added it so that PDS operators would have an action that is a negative signal but has no effect so they could use to implement their own complex policies (e.g. a three-strikes policy).

packages/bsky/tests/__snapshots__/feed-generation.test.ts.snap
```diff
@@ -552,96 +552,6 @@ Array [
       "uri": "record(9)",
       "viewer": Object {},
     },
-    "reply": Object {
-      "parent": Object {
-        "$type": "app.bsky.feed.defs#postView",
```
devinivy, line 557: Any idea why these snaps changed?

packages/bsky/tests/auto-moderator/fuzzy-matcher.test.ts
```diff
@@ -37,7 +37,7 @@ describe('fuzzy matcher', () => {
   const getAllReports = () => {
     return network.bsky.ctx.db
       .getPrimary()
-      .db.selectFrom('moderation_report')
+      .db.selectFrom('moderation_event')
```
devinivy, line 40: Can we make it so that this function `getAllReports()` returns only reports rather than all events?

packages/pds/src/services/moderation/views.ts
```diff
@@ -0,0 +1,477 @@
+import { Selectable } from 'kysely'
```
devinivy, line 1: I don't think moderation views are used anywhere on the pds, so I believe we should be able to leave this file removed 👍

packages/pds/src/api/com/atproto/admin/sendEmail.ts
```diff
@@ -29,6 +31,20 @@ export default function (server: Server, ctx: AppContext) {
...
+            subjectLine: subject,
+          },
+          subject: {
+            $type: 'com.atproto.admin.defs#repoRef',
+            did: recipientDid,
+          },
+          createdBy: senderDid,
+        },
+        { ...authPassthru(req), encoding: 'application/json' },
+      )
```
devinivy, line 47: I think we might want to make the email flow more uniform with how takedowns, etc. work: a `#modEventEmail` is dispatched to the appview using the `emitModerationEvent` endpoint, then the appview reaches out to `com.atproto.admin.sendEmail` on the PDS in order to push the email out to the user.

packages/bsky/src/api/blob-resolver.ts
```diff
@@ -83,20 +83,16 @@ export async function resolveBlob(
...
-        'moderation_action',
-        'moderation_action.id',
-        'moderation_action_subject_blob.actionId',
-      )
-      .where('cid', '=', cidStr)
-      .where('action', '=', TAKEDOWN)
-      .where('reversedAt', 'is', null)
+      .selectFrom('moderation_subject_status')
+      .select('id')
+      .where(sql<string>`${ref('blobCids')} @> ${JSON.stringify([cidStr])}`)
```
devinivy, line 94: I think we may be able to take more advantage of kysely here since it knows about the `@>` operator.
```ts
      .where('blobCids', '@>', json([cidStr]))
```

Where `json` is a helper like:
```ts
function json<T>(value: T) {
  return sql`CAST(${JSON.stringify(value)} AS JSONB)`
}
```

packages/bsky/src/api/blob-resolver.ts
```diff
@@ -83,20 +83,16 @@ export async function resolveBlob(
...
+      .where('takendown', 'is', true)
```
devinivy, line 95: I am having a little trouble thinking through how blobs are modeled here.  I think the main concern is that we want these lookups to be quick and cheap for the blob resolver.  But we also might consider tracking blob takedowns separately similar to how they are managed with `updateSubjectStatus` and `getSubjectStatus`: it might make it easier to make this lookup quick/cheap, and also might make blobs easier to work with.

lexicons/com/atproto/admin/defs.json
```diff
@@ -158,6 +121,59 @@
...
+          "type": "string",
+          "format": "datetime"
+        },
+        "lastReportedAt": {
+          "type": "string",
+          "format": "datetime"
+        },
+        "takendown": {
+          "type": "boolean"
+        },
```
devinivy, line 170: Might be worth taking a peek at `#statusAttr` and `getSubjectStatus`, `updateSubjectStatus` lexicons— we may be able to make these more consistent with each other.

devinivy (commented):
Dropping a first round of high-level review, some more incoming.  This PR is a fun read!

packages/bsky/src/api/blob-resolver.ts
foysalit: ah TIL!
avoiding the helper for now and using direct sql helper.

packages/bsky/tests/__snapshots__/feed-generation.test.ts.snap
foysalit: idk but these seem to fail intermittently on my local on main. I tried to debug for a bit and found out that somehow `invalidReplyRoot` comes out to be true for some replies which causes the tests to fail. however, it's not reliably reproduced. so I ended up disabling all failing tests and enabling 1 at a time and somehow ended up getting all green.

packages/bsky/src/api/blob-resolver.ts
foysalit: So, previously, when an action would contain an array of `subjectBlobCids` it would simply create a row for each item in the array where each row references back to the action entry itself. now, instead of storing these in a separate table, we're just dumping them all in this column.

For perf concerns, I am sure a separate table would be faster for lookups but also, I don't think we have many rows with blob references (speaking purely from observing moderators taking actions, we should verify this from prod db perhaps?) so maybe with an index on this array column, we would have fast enough lookups for now? Happy to take your recommendation on this since I don't have a good idea of the scale we're at right now vs. where we will be before this becomes an issue.

on `getSubjectStatus` it looks like we are using `did`, `cid` and `recordUri` which is basically the same as other records/strongRefs except that for those we're not using explicit `did`, right? so, if wanted to start treating blobs as individual subjects, wouldn't we also need some kind of flag column in this table that tells us which entry is for a blob and which one is for a generic record, since both of them would have a `did`, `recordPath` and a `cid` ?

lexicons/com/atproto/admin/defs.json
foysalit: not sure if these can be same though? `statusAttr` is defined as 

```
{
    "applied": { "type": "boolean" },
    "ref": { "type": "string" }
}
```

but in the appview, `takendown: true` always means `applied: true` and `ref` would just be a reference to the entry's id itself so it's kinda redundant unless we map this ref to reference to the event that changed the status?

devinivy: They do also have different names in `getSubjectStatus` / `updateSubjectStatus`: `takedown` vs. `takendown`.

bnewbold (approved):
I think the important earlier notes I left have been resolved. This is looking great, really looking forward to getting it deployed!

lexicons/com/atproto/admin/defs.json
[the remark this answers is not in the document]
bnewbold: The argument for seconds is that it is just a standard "how to express duration" thing. In languages like golang, there are duration types like `time.Day` and those are basically constants expressed as a number of seconds. To convert to "hours" would end up not using those constants or types.

For mutes it also feels pretty likely that we'd want to "mute for 10 minutes" or whatever. Numbers in Lexicons are integers, not floats, so fraction of hour wouldn't be possible.

I wouldn't hold up merging/deploying on just this though.

lexicons/com/atproto/admin/defs.json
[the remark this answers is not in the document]
bnewbold: I'd be ok with mapping `#flag` to `#modEventAcknowledge`.

packages/bsky/src/db/migrations/20231003T202833377Z-create-moderation-subject-status.ts
```diff
@@ -0,0 +1,116 @@
...
+    .execute()
+  await db.schema
+    .createTable('moderation_subject_status')
+    .addColumn('id', 'serial', (col) => col.primaryKey())
+
+    // Identifiers
+    .addColumn('did', 'varchar', (col) => col.notNull())
+    // Default to '' so that we can apply unique constraints on did and recordPath columns
+    .addColumn('recordPath', 'varchar', (col) => col.notNull().defaultTo(''))
+    .addColumn('blobCids', 'jsonb')
```
devinivy, line 33: I think I recall you saying there was an index that we needed on `blobCids`.

packages/bsky/src/db/migrations/20231003T202833377Z-create-moderation-subject-status.ts
```diff
@@ -0,0 +1,116 @@
...
+    .execute()
+  await db.schema
+    .alterTable('record')
+    .addForeignKeyConstraint(
+      'record_takedown_id_fkey',
+      ['takedownId'],
+      'moderation_event',
+      ['id'],
+    )
+    .execute()
```
devinivy, line 82: I didn't see anything in the migration that would adjust these columns on `record` and `actor` while the `takedownId` will change, so this foreign key will become invalid.

One idea I have is to preserve the action->event ids, but not bother maintaining the report ids.  The process would be something like:
1. Insert all actions into the event table explicitly setting the `id` columns to match.
2. Update the next autoincrement event `id` value to be higher than that last id (maybe plus some gap) using `setval()`.
3. Insert reports into the event table using the autoincrementing id.

packages/bsky/src/migrate-moderation-data.ts
```diff
@@ -0,0 +1,280 @@
...
+
+type ModerationActionRow = Omit<ModerationEventRow, 'comment' | 'meta'> & {
+  reason: string | null
+}
+
+const getEnv = () => ({
+  DB_URL:
+    process.env.MODERATION_MIGRATION_DB_URL ||
+    'postgresql://pg:password@127.0.0.1:5433/postgres',
+  DB_POOL_SIZE: Number(process.env.MODERATION_MIGRATION_DB_URL) || 10,
```
devinivy, line 14:
```suggestion
  DB_POOL_SIZE: Number(process.env.MODERATION_MIGRATION_DB_POOL_SIZE) || 10,
```

packages/bsky/src/migrate-moderation-data.ts
```diff
@@ -0,0 +1,280 @@
...
+        .limit(chunkSize)
+        .selectAll()
+      const events = (await eventsQuery.execute()) as ModerationActionRow[]
+
+      for (const event of events) {
+        // Remap action to event data type
+        const actionParts = event.action.split('#')
+        await adjustModerationSubjectStatus(tx, {
+          ...event,
+          action: `${actionParts[0]}#modEvent${actionParts[1]
```
devinivy, line 201: There may be a few legacy actions where `actionParts[0]` is `com.atproto.admin.moderationAction` rather than `com.atproto.admin.defs`.  Since we know we want the first part to be `com.atproto.admin.defs`, perhaps we can hardcode it to that value.

packages/bsky/src/migrate-moderation-data.ts
```diff
@@ -0,0 +1,280 @@
...
+          )}` as ModerationEventRow['action'],
+          comment: event.reason,
+          meta: null,
+        })
+      }
+
+      console.log(`Processed events chunk ${currentChunk} of ${totalChunks}`)
+      lastProcessedId = events.at(-1)?.id ?? 0
+      currentChunk++
+    } while (currentChunk < totalChunks)
```
devinivy, line 214: We start at chunk `1` and remain less than `totalChunks`—at a glance it seems like this may lose one chunk.  For example, if there are two chunks then we might process chunk `1`, increment it to `2`, then stop before processing chunk `2` due to the `while ()` condition.

If this seems like we could get rid of chunk-counting altogether, and instead paginate through the result set as we're currently already doing until there are no results left.  The condition might look something like:
```suggestion
      lastProcessedId = events.at(-1)?.id
    } while (lastProcessedId !== undefined)
```

packages/bsky/src/migrate-moderation-data.ts
foysalit: hmm... yeah I did see those even in the staging dump but thought we would wanna leave them alone.

packages/bsky/src/db/migrations/20231003T202833377Z-create-moderation-subject-status.ts
foysalit: Problem with re-sequencing report ids is that the application logic expects the ids to be sequential and pagination depends on that too so if we were to keep action ids matching their original ids but not reports, when looking at these data, it will always look like some actions were taken before reports came in since the sorting uses ids.

I am wondering if we even need this reference to events? isn't the subject_status table maintaining the real state anyways? if we really do need the reference, how about we update the takedownId in these tables to match the newly created event ids during migration?

devinivy: > Problem with re-sequencing report ids is that the application logic expects the ids to be sequential 

Ohh, I didn't notice that!  Where does that happen?

> I am wondering if we even need this reference to events? isn't the subject_status table maintaining the real state anyways?

It's a good point, `subject_status` is a new table to maintain this state and the `takedownId` column wasn't designed with it in mind.  However, I don't think that we can afford to join over to `subject_status` everywhere that we need to check for takedowns, so I expect that some state will remain on the actor/record rows to keep these checks very cheap.  If that is the case I think it's nearly all upside to maintain not just a boolean but a full id, which gives us better debuggability and helps rather than hurts with data integrity issues that can arise from having duplicated state.

> if we really do need the reference, how about we update the takedownId in these tables to match the newly created event ids during migration?

That works!  I think it would be ideal if the ids didn't change, but if they must change then we can update the `takedownId` column.  Part of the reason it would be nice not to touch that column is that it's quite sensitive, so we should just take some care to ensure that there's not a way to unset it (e.g. if we can't find a matching id for a given subject we should rollback rather than null it out).  On the pds we moved `takedownId` to a new column named `takedownRef`, and we could totally do the same thing here if it's easier to update the ids in a new column rather than in-place.


Overall I'd be interested to see where the id ordering matters in the application.  If it's not difficult to loosen that constraint it might be a nice outcome.  I am sure that we can handle the pagination issue by paginating on `(createdAt, id)` as we do many places.  If this ends-up not being very easy to do then we should look at updating the `takedownId` state.

lexicons/com/atproto/admin/defs.json
foysalit: Referencing our previous conversations on this: we decided that this not blocking at the moment and we are settling for 1 event at a time model for now. Soon after release, if we see this being a common case, we could either end up adjusting the client to dispatch multiple requests through unified UI/UX or allow this endpoint to receive multiple events as array and process all in one go.

packages/bsky/src/migrate-moderation-data.ts
```diff
@@ -0,0 +1,280 @@
...
+                ON moderation_action.id = moderation_action_subject_blob."actionId"
+          WHERE moderation_action."reversedAt" is NULL
+          GROUP by moderation_action."subjectUri", moderation_action."subjectDid"
+    ) as blob_action
+    WHERE did = "subjectDid" AND position("recordPath" IN "subjectUri") > 0
+  `.execute(db.db)
+  console.log(`Updated blob cids on ${results.numUpdatedOrDeletedRows} rows`)
+}
+
+async function main() {
```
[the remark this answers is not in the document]
foysalit: @devinivy do you mean the bsky package's js build? not sure why this would concern the pds?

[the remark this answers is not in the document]
devinivy: Yep yep, typo on my part!

packages/bsky/src/migrate-moderation-data.ts
```diff
@@ -272,9 +285,29 @@ export async function MigrateModerationData() {
...
+
+  // If there are existing events in the moderation_event table, we assume that the migration has already been run
+  // so we just bring over any new reports since last run
+  if (existingEventsCount) {
+    console.log(
+      `Found ${existingEventsCount} existing events. Migrating ${counts.reportsCount} reports only, ignoring actions`,
+    )
+    const reportMigrationStartedAt = Date.now()
+    // TODO: Make sure to set the appropriate last report id to ensure continuity before running the script
+    await createEvents(primaryDb, { onlyReportsAboveId: 1000000 })
```
foysalit, line 301: @devinivy this is hardcoded so you'd have to set the id before running the script. I could make it a cli arg but not really sure how you're planning on running this.

Issue #1615, opened 2023-09-16
The properties field is missing for threadgate lexicon


MarshalX:
**Describe the bug**

Recently added [threadgate lexicon](https://github.com/bluesky-social/atproto/blob/main/lexicons/app/bsky/feed/threadgate.json) uses object without properties.

**Expected behavior**

Instead of 
https://github.com/bluesky-social/atproto/blob/17ce804ffaaafb1b1efb61174217a1da275feda8/lexicons/app/bsky/feed/threadgate.json#L26-L33

I except something valid ;) 

mb this (it works for me)? 

```json
    "mentionRule": {
      "type": "object",
      "description": "Allow replies from actors mentioned in your post.",
      "properties": {}
    },
    "followingRule": {
      "type": "object",
      "description": "Allow replies from actors you follow.",
      "properties": {}
    },
```

because we have this already:
https://github.com/bluesky-social/atproto/blob/17ce804ffaaafb1b1efb61174217a1da275feda8/lexicons/app/bsky/actor/getPreferences.json#L8-L11

**Additional context**

My Python lexicon parser failed to parse the latest lexicon from the main branch.

According to documentation, this field is still required: https://atproto.com/specs/lexicon#object

<img width="710" alt="image" src="https://github.com/bluesky-social/atproto/assets/15520314/9d92cab5-f865-41e8-9b71-fc81ec82510a">

Pls let me know if the docs is outdated

MarshalX:
hi @dholms and @devinivy! summon you as code owners of this lexicon

intrnl:
seems like an unintentional mistake that didn't get caught due to how JS processes values in a for..in loop or something similar, it skips the loop if you pass `undefined` in as the object to iterate,

when this issue was brought up I was curious why my lexicon to d.ts wasn't crashing 👀

intrnl:
okay so I've looked further into the lexicon code and found out that well... the `properties` field in the schema is actually marked as optional

https://github.com/bluesky-social/atproto/blob/17ce804ffaaafb1b1efb61174217a1da275feda8/packages/lexicon/src/types.ts#L170-L183

so there's two issues at play, the lexicon schema validation didn't throw because `properties` is marked as optional, the lexicon codegen didn't throw because of for..in

bnewbold:
[...]

PR #1628, opened 2023-09-18
Enforce properties field on lexicon object schemas

dholms:
Re: https://github.com/bluesky-social/atproto/issues/1615

We should enforce the semantic that objects have a `properties` field declared


bnewbold (approved):
It's a bit verbose this way, but I think it is actually good for it to be explicit and obvious that there are no properties.

PR #1629, opened 2023-09-18
Bugfix: handle contention

dholms:
There was an case in handle contention that we weren't correctly handling: a _new_ account is created that correctly uses a handle for a previously indexed account


devinivy (approved):
Good catch 👍

PR #1637, opened 2023-09-19
add `tags` to posts

estrattonbailey:
Adds tags fields to rich text facets and posts lexicons, and implements tags in the PDS and App View codebases.

The `maxLength` and `maxGraphemes` values minic `displayName` on other lexicons. Though unrelated, that seemed like a pretty reasonable length considering overall post max of 300 graphemes.


lexicons/app/bsky/feed/post.json
```diff
@@ -38,6 +38,10 @@
             "type": "union",
             "refs": ["com.atproto.label.defs#selfLabels"]
           },
+          "tags": {
+            "type": "array",
+            "items": { "type": "string", "maxLength": 640, "maxGraphemes": 64 }
```
bnewbold, line 43: include a brief description, eg: "Additional non-inline hashtags describing this post". The description should clarify whether the `#` prefix should be included in these strings or not (i'm not sure). both here and for the in-line (facet) string field, should be consistent about expectation.

at some point we might want a string format like "slug" or "keyword" which has some generic character restrictions. that's a larger protocol tangent and we don't have that today and it should not block this work.

I assume we *will* apply some restrictions server-side when indexing, beyond just length, like "trim whitespace". we could update the description with those details later.

lexicons/app/bsky/feed/post.json
```diff
@@ -38,6 +38,10 @@
             "type": "union",
             "refs": ["com.atproto.label.defs#selfLabels"]
           },
+          "tags": {
+            "type": "array",
```
bnewbold, line 42: hashtags are notoriously stuffed for SEO/broadcast purposes, both as simple spam as as annoyingly "loud" promotion of otherwise legitimate content.

We *could* impost an array length limit here (8?). A quick search says that Tumblr has no limit but only indexes the first 5, which is maybe a simpler/better solution.

I'd be fine with a limit but could go either way. Not a blocking thing for PR, just in case it hadn't been considered.

lexicons/app/bsky/feed/post.json
bnewbold, line 43: at some point we might want a string format like "slug" or "keyword" which has some generic character restrictions. that's a larger protocol tangent and we don't have that today and it shouldn't block this work.

bnewbold (approved):
Overall not controversial, Lexicon looks good.

I'm not sure the overall plan/semantics, looks like we'll support both in-line (facet) hashtags which will get parsed out client-side, plus additional tags.

There was a decision to be made about whether to stuff hashtags in labels, as special labels with `#` prefix, and this is probably that decision point. There could be some advantages having them flow through the regular moderation SDK, in terms of mutes and filtering. Keeping them separate and application-specific is fine with me.

Probably out of scope for this PR but might want a `suggestHashtags` typeahead search endpoint (Lexicon). I think we'll be able to do this with the new/current search index pretty quick+easy, sorted by usage. This would be used in-client to list full hashtag strings when adding to a post (similar to mention typeahead).

lexicons/app/bsky/feed/post.json
estrattonbailey: Ah good call on description, will add 👍 

As for adding an additional format and other restrictions, @pfrazee and I talked in circles about this last night and we ended up back at the starting point: restrict length/graphemes, but otherwise allow anything.

We _will_ be restrictive with detection of facets in the rich text editor, but when adding additional tags non-inline you'd be able to include things like spaces, etc that would otherwise gum up the rich text editor.

Users may do some wacky stuff, which might make searching for a tag a little difficult, but otherwise handling the tag strings doesn't seem like it should present serious difficulties.

Thoughts?

lexicons/app/bsky/feed/post.json
estrattonbailey: Not sure what's best here, but I would guess we _should_ add a limit e.g. we limit post length, it would be kinda silly if someone could add 100 tags or something.

Yeah maybe 8, or 10? Search can index differently if need be, right?

CC @pfrazee for thoughts

estrattonbailey: Updated with a maxLength of 8

pfrazee: 8 sgtm

pfrazee:
> There was a decision to be made about whether to stuff hashtags in labels, as special labels with `#` prefix, and this is probably that decision point. There could be some advantages having them flow through the regular moderation SDK, in terms of mutes and filtering. Keeping them separate and application-specific is fine with me.

The main advantage would be giving labelers the ability to tag items. I'm really not sure that's something we want. It introduces complexity with a very minimal upside.

The other advantage of tags == labels would be to have tags drive self-labeling. If you `#porn` then it's the equivalent of self-labeling as porn. Again, I'm not sure this is a huge win. 

> Probably out of scope for this PR but might want a suggestHashtags typeahead search endpoint (Lexicon). I think we'll be able to do this with the new/current search index pretty quick+easy, sorted by usage. This would be used in-client to list full hashtag strings when adding to a post (similar to mention typeahead).

Agree on all points here

estrattonbailey:
>If you #porn

We're gonna want to filter some of these out of the typeahead hashtag search

pfrazee:
> We're gonna want to filter some of these out of the typeahead hashtag search

Some, though "porn" should get filtered client side based on user mod prefs

lexicons/app/bsky/feed/post.json
```diff
@@ -38,6 +38,12 @@
...
+            "maxLength": 8,
+            "items": { "type": "string", "maxLength": 640, "maxGraphemes": 64 },
```
dholms, line 44: maxLength v maxGraphemes is basically to allow for unicode stuff

i'm not sure we want to allow hashtags that are 64 emojis 😅

Which is to say: I think we'd be fine with a smaller max length here

dholms: (altho tbh i'm not sure how much it really matters, what matters is that there is _some_ limit)

lexicons/app/bsky/feed/post.json
```diff
@@ -38,6 +38,12 @@
             "type": "union",
             "refs": ["com.atproto.label.defs#selfLabels"]
           },
+          "tags": {
```
dholms, line 41: btw i like that we used `tags` & not `hashtags`

lexicons/app/bsky/feed/post.json
estrattonbailey: I don't have an educated opinion on this, but keeping the max kinda roomy might be good for flexibility and creativity. Unlikely that 64 emoji hashtags will be used all that often.

.changeset/three-snakes-turn.md
```diff
@@ -0,0 +1,7 @@
+---
+'@atproto/bsky': patch
+'@atproto/api': patch
+'@atproto/pds': patch
```
estrattonbailey, line 4: Was debating doing a minor release 🤔

lexicons/app/bsky/feed/defs.json
```diff
@@ -31,7 +31,11 @@
           "type": "array",
           "items": { "type": "ref", "ref": "com.atproto.label.defs#label" }
         },
-        "threadgate": { "type": "ref", "ref": "#threadgateView" }
+        "threadgate": { "type": "ref", "ref": "#threadgateView" },
+        "tags": {
```
dholms, line 35: just curious - is there a reason to elevate this to the postView instead of just getting the data from within the record (similar to how we handle text/facets/etc)?

i'm not opposed to this, but most other fields that we elevate to the view have some additional info/hydration that go along with them

lexicons/app/bsky/feed/post.json
dholms: eh yeah seems fine 👍

dholms (approved):
looks good! 👍  just the one question about tags in postView

lexicons/app/bsky/feed/defs.json
dholms: for instance, we don't have `langs` on `postView`

estrattonbailey: Ah ok yeah, I suppose not 🤔 I'll remove this functionality

dholms: we can always add it in later if we realize we need it - it's non-breaking

PR #1638, opened 2023-09-20
Add feed-vew and thread-view preferences

pfrazee:
Adds some preference fields for feeds and threads. For feeds, it has support for multiple different feeds; it seems pretty likely that people will want to set per-feed settings so that seems like a good place to start.

### DIFF
--- a/packages/api/src/bsky-agent.ts
+++ b/packages/api/src/bsky-agent.ts
@@ -8,6 +8,18 @@ import {
 } from './client'
 import { BskyPreferences, BskyLabelPreference } from './types'
 
+const FEED_VIEW_PREF_DEFAULTS = {
+  hideReplies: false,
+  hideRepliesByUnfollowed: false,
+  hideRepliesByLikeCount: 0,
+  hideReposts: false,
+  hideQuotePosts: false,
+}
+const THREAD_VIEW_PREF_DEFAULTS = {
+  sort: 'oldest',
+  prioritizeFollowedUsers: true,
+}
+
 declare global {
   interface Array<T> {
     findLast(
@@ -254,6 +266,12 @@ export class BskyAgent extends AtpAgent {
         saved: undefined,
         pinned: undefined,
       },
+      feedViewPrefs: {
+        home: {
+          ...FEED_VIEW_PREF_DEFAULTS,
+        },
+      },
+      threadViewPrefs: { ...THREAD_VIEW_PREF_DEFAULTS },
       adultContentEnabled: false,
       contentLabels: {},
       birthDate: undefined,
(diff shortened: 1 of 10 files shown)


packages/api/src/bsky-agent.ts
```diff
@@ -289,6 +307,18 @@ export class BskyAgent extends AtpAgent {
...
+        AppBskyActorDefs.isFeedViewPref(pref) &&
+        AppBskyActorDefs.validateFeedViewPref(pref).success
+      ) {
+        const { $type, feed, ...v } = pref
+        prefs.feedViewPrefs[pref.feed] = { ...FEED_VIEW_PREF_DEFAULTS, ...v }
+      } else if (
+        AppBskyActorDefs.isThreadViewPref(pref) &&
+        AppBskyActorDefs.validateThreadViewPref(pref).success
+      ) {
+        const { $type, ...v } = pref
```
bnewbold, line 320: could do this `$type` removal for `feedViewPrefs` values also (couple lines above), or not bother

packages/api/src/bsky-agent.ts
```diff
@@ -406,6 +436,46 @@ export class BskyAgent extends AtpAgent {
...
+  ) {
+    await updatePreferences(this, (prefs: AppBskyActorDefs.Preferences) => {
+      const existing = prefs.findLast(
+        (pref) =>
+          AppBskyActorDefs.isFeedViewPref(pref) &&
+          AppBskyActorDefs.validateFeedViewPref(pref).success &&
+          pref.feed === feed,
+      )
+      if (existing) {
+        pref = { ...existing, ...pref }
```
bnewbold, line 452: I'm not super strong with typescript, or our use of schema objects, but i'd worry that this might not successfully remove/nullify preferences in some situations.

Eg, if you had `hideRepliesByLikeCount = 10`, then submitted a request where `hideRepliesByLikeCount` was cleared (but not specifically set to `0`), would the old value be retained?

bnewbold (commented):
A couple comments but generally looks good to me here on the client API side.

On the PDS server-side, it looks like having multiple preference entries with the same `$type` (eg, with different `feed` value) might not work? The database table uses the `$type` as a column and it looks like it is probably a unique key.

packages/api/src/bsky-agent.ts
pfrazee: Good looking out but it actually does exactly what's intended

```
setFeedViewPrefs({hideRepliesByLikeCount: 5}) -> 5
setFeedViewPrefs({hideRepliesByLikeCount: undefined}) -> undefined
setFeedViewPrefs({}) -> no change
```

packages/api/src/bsky-agent.ts
pfrazee: wdym?

devinivy (approved):

estrattonbailey (approved):
LGTM!

dholms (approved):

packages/api/src/bsky-agent.ts
bnewbold: I could have sworn that `const { $type, feed, ...v } = pref` on line 314 was not there when I reviewed, but I see it now!

PR #1644, opened 2023-09-21
Disable pds appview routes

dholms:
In preperation of merging https://github.com/bluesky-social/atproto/pull/1198 which contains many changes to pds logic including config changes, we'll merge & deploy several piecemeal steps

This PR removes all appview routes & view logic from the PDS.

PR #1645, opened 2023-09-21
Disable pds appview indexing

dholms:
Follow up to https://github.com/bluesky-social/atproto/pull/1644 & precursor to merging https://github.com/bluesky-social/atproto/pull/1198 

This PR removes all appview indexing logic from the PDS.


packages/dev-env/src/mock/index.ts
```diff
@@ -186,30 +186,30 @@ export async function generateMockSetup(env: TestNetworkNoAppView) {
...
-          cts: new Date().toISOString(),
-        },
-      ])
-      .execute()
-  }
+  // const ctx = env.pds.ctx
+  // if (ctx) {
+  //   await ctx.db.db
+  //     .insertInto('label')
+  //     .values([
```
devinivy, line 193: Can we nix this code?

packages/pds/src/app-view/api/app/bsky/unspecced.ts
```diff
@@ -43,18 +43,6 @@ export default function (server: Server, ctx: AppContext) {
       }
     },
   })
-
-  server.app.bsky.unspecced.applyLabels({
```
devinivy, line 47: I expect we'll want to remove the use of this endpoint in `packages/bsky/src/auto-moderator/index.ts`.

packages/pds/src/db/migrations/20230922T033938477Z-remove-appview.ts
```diff
@@ -0,0 +1,29 @@
...
+  await db.schema.dropTable('post').execute()
+  await db.schema.dropTable('profile_agg').execute()
+  await db.schema.dropTable('profile').execute()
+  await db.schema.dropTable('repost').execute()
+  await db.schema.dropTable('subscription').execute()
+  await db.schema.dropTable('suggested_follow').execute()
+  await db.schema.dropTable('view_param').execute()
+}
+
+export async function down(db: Kysely<unknown>): Promise<void> {
```
devinivy, line 27: I think this check warning may continue to bug us if we don't handle it here 🙃 
```suggestion
export async function down(_db: Kysely<unknown>): Promise<void> {
```

devinivy (approved):
Looks solid! 👋 goodbye in-process appview, it's been real.

packages/dev-env/src/mock/index.ts
dholms: nixed the pds-only dev-env & switched this code to insert the labels into the appview 👌

packages/pds/src/app-view/api/app/bsky/unspecced.ts
dholms: yup - wiped out the whole schema 👌

PR #1647, opened 2023-09-21
indicate that getPopular is deprecated

bnewbold:
As discussed, a quick first step to ripping this out soon.


devinivy (approved):
Looks good!  It's possible a codegen needs to run to incorporate the updated description into API and type docs.

PR #1651, opened 2023-09-21
Add hashtag detection to richtext

estrattonbailey:
Updates our rich text detection to include hashtags.

### DIFF
--- a/packages/api/src/rich-text/detection.ts
+++ b/packages/api/src/rich-text/detection.ts
@@ -72,19 +72,20 @@ export function detectFacets(text: UnicodeString): Facet[] | undefined {
   {
     const re = /(?:^|\s)(#[^\d\s]\S*)(?=\s)?/g
     while ((match = re.exec(text.utf16))) {
-      let [, tag] = match
+      let [tag] = match
+      const hasLeadingSpace = /^\s/.test(tag)
 
       tag = tag.trim().replace(/\p{P}+$/gu, '') // strip ending punctuation
 
       // inclusive of #, max of 64 chars
       if (tag.length > 66) continue
 
-      const index = match.index
+      const index = match.index + (hasLeadingSpace ? 1 : 0)
 
       facets.push({
         index: {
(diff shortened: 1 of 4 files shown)


packages/api/tests/rich-text-detection.test.ts
```diff
@@ -208,6 +209,57 @@ describe('detectFacets', () => {
...
+        [],
+      ],
+      ['its a #double#rainbow', ['#double#rainbow']],
+      ['some #n0n3s@n5e!', ['#n0n3s@n5e']],
+      ['works #with,punctuation', ['#with,punctuation']],
+      [
+        'strips trailing #punctuation, #like, #this.',
+        ['#punctuation', '#like', '#this'],
+      ],
+      ['strips #multi_trailing___...', ['#multi_trailing']],
```
estrattonbailey, line 241: We don't want to detect trailing punctuation e.g. `I love #turtles!` should match `#turtles`. However, we should also strip multiple ending punctuation marks, like `I love #turtles!!!!!`.

packages/api/src/rich-text/detection.ts
```diff
@@ -69,6 +69,33 @@ export function detectFacets(text: UnicodeString): Facet[] | undefined {
       })
     }
   }
+  {
+    const re = /(?:^|\s)(#[^\d\s]\S*)(?=\s)?/g
+    while ((match = re.exec(text.utf16))) {
+      let [tag] = match
+      const hasLeadingSpace = /^\s/.test(tag)
```
estrattonbailey, line 76: Matches a leading space rn because I need to make sure that something like `https://esb.lol/page#hash` doesn't match, but `I love #turtles` does. So this is needed rn to bump the index start so that we ignore the leading space.

Lmk if y'all have regex ideas and I can provide some text to test with.

pfrazee: Might be possible with a lookahead or lookbehind but not sure it's worth bothering with

pfrazee (approved):
LGTM!

PR #1659, opened 2023-09-22
Clear follow viewer state when blocking

devinivy:
When there's a block between two actors, ignore any follow state between them in profile view `viewer` state.  This case shows-up on `getProfile` and `getBlocks` in particular, two of the rare cases where a profile view is served between two blocking users.  Conceptually the idea is that follows are invalid in the face of blocks: blocking users cannot be following each other.


dholms (approved):
looks good 👍

PR #1660, opened 2023-09-23
Makefile: run code formatting after codegen

bnewbold:
little quality-of-life thing for me


devinivy (approved):
I think that the codegen routines should already be doing this, but perhaps they aren't.  In any case, I don't think this can hurt!

dholms (approved):
yup 👍 

i actually had this change in an earlier PR but it somehow didn't make it into main

thx for adding!

PR #1668, opened 2023-09-25
Idempotent creates

dholms:
When duplicate record backlinks are created (ie 2 likes of the same post from the same user), just make the second idempotent and return the uri/cid of the original.

This is in opposition to our previous approach of deleting the original and re-creating the second.

### DIFF
--- a/packages/pds/src/api/com/atproto/repo/createRecord.ts
+++ b/packages/pds/src/api/com/atproto/repo/createRecord.ts
@@ -6,12 +6,12 @@ import {
   BadCommitSwapError,
   InvalidRecordError,
   PreparedCreate,
-  prepareDelete,
 } from '../../../../repo'
 import AppContext from '../../../../context'
 import { ids } from '../../../../lexicon/lexicons'
 import Database from '../../../../db'
 import { ConcurrentWriteError } from '../../../../services/repo'
+import { AtUri } from '@atproto/syntax'
 
 export default function (server: Server, ctx: AppContext) {
   server.com.atproto.repo.createRecord({
@@ -62,16 +62,20 @@ export default function (server: Server, ctx: AppContext) {
         throw err
       }
 
-      const backlinkDeletions = validate
-        ? await getBacklinkDeletions(ctx.db, ctx, write)
-        : []
-
-      const writes = [...backlinkDeletions, write]
+      const existing = validate
+        ? await getExistingBacklink(ctx.db, ctx, write)
+        : null
+      if (existing) {
+        return {
+          encoding: 'application/json',
+          body: { uri: existing.uri.toString(), cid: existing.cid.toString() },
+        }
+      }
 
       try {
         await ctx.services
           .repo(ctx.db)
-          .processWrites({ did, writes, swapCommitCid }, 10)
+          .processWrites({ did, writes: [write], swapCommitCid }, 10)
       } catch (err) {
         if (err instanceof BadCommitSwapError) {
           throw new InvalidRequestError(err.message, 'InvalidSwap')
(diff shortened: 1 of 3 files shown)


pfrazee:
Oh interesting, I wasn't aware we were doing this at all. I'm fine with this change, but I'm curious how we expect this to evolve with lexicon generality

dholms:
If your PDS doesn't understand a given lexicon, then it just won't do this (similar to how it won't do read-after-write or schema validation)

The backlinks are stored generically, so it doesn't require any schema-specific indexing - only schemas-specific logic

pfrazee:
ok good to know

packages/pds/src/api/com/atproto/repo/createRecord.ts
```diff
@@ -62,16 +62,20 @@ export default function (server: Server, ctx: AppContext) {
...
-      const writes = [...backlinkDeletions, write]
+      const existing = validate
+        ? await getExistingBacklink(ctx.db, ctx, write)
+        : null
+      if (existing) {
+        return {
+          encoding: 'application/json',
+          body: { uri: existing.uri.toString(), cid: existing.cid.toString() },
+        }
+      }
```
devinivy, line 73: To be honest the idempotent create a little weird once you see it!  I understand again why we may have originally reached for the deletion + creation rather than idempotency in the original implementation.  In short, it would be natural for the post-condition of `createRecord` for the record to exist with the specified content and uri.  If we want the no-op (I definitely see the upsides of no-oping) another way to get it could be to raise an error that clients are expected to understand and usually ignore.

packages/pds/src/api/com/atproto/repo/createRecord.ts
```diff
@@ -6,12 +6,12 @@ import {
   BadCommitSwapError,
   InvalidRecordError,
   PreparedCreate,
-  prepareDelete,
 } from '../../../../repo'
 import AppContext from '../../../../context'
 import { ids } from '../../../../lexicon/lexicons'
 import Database from '../../../../db'
 import { ConcurrentWriteError } from '../../../../services/repo'
+import { AtUri } from '@atproto/syntax'
```
devinivy, line 14: Tiiiny nit, might be cool to hoist this import up.

devinivy (approved):
Code looks great!  Having 🤔s about allowing creates that are idempotent relative to application logic (i.e. backlinks), left a little note about it.

packages/pds/src/api/com/atproto/repo/createRecord.ts
devinivy: I think what I'm getting at is just that it makes the spec for `createRecord` very loose—the baseline guarantees of a successful creation would be something like "some (any) record exists in the provided repo and collection."  I think the difficulty may be that its behavior in case of success isn't super legible to its inputs/outputs.  Raising an error (e.g. 409 conflict) that is a carve-out for the PDS enforcing application logic seems alright, but forces clients to know that this happens.  The current behavior deleting the "duped" record is also fairly weird, but it does the nice thing of maintaining the post-condition that the input record exists.  At the end of the day this is just the friction of building an app on top of CRUD 🫠

pfrazee: yeah it'll register as a failure and roll back the attempted like. it won't throw up an error message and it's a rare case, so we could merge this and then update the app if we really need to do this

tpaulshippy: Is the consensus to throw a 409 then? Makes more sense to me than the current behavior. I realize it is over a year later, but are we just waiting for someone to implement the 409?

Or, if we are fine with the current behavior, should this be closed?

PR #1677, opened 2023-09-26
Better handle 3rd party blocks in feeds

dholms:
Previously, we were stripping block violating replies entirely from feeds.

This adjusts the logic such that we strip them from threads still but leave them in feeds. We still strip the hydrated reply parent/root on the `feedPostView`.

It also adjusts `getPostThread` such that a blocked reply can be rendered as the anchor of the thread (the requested URI), however it is still untethered from it's parent in that context.


packages/bsky/src/services/feed/views.ts
```diff
@@ -246,7 +249,7 @@ export class FeedViews {
     if (
       post.author.viewer?.blockedBy ||
       post.author.viewer?.blocking ||
-      blocks[uri]?.reply
+      (replyUri !== null && blocks[replyUri]?.reply)
```
devinivy, line 252: Just want to make sure I understand the intended behavior here: replies to block-violating replies are blocked, is that right?

devinivy (commented):
Looks good!  To round this out I think we may want to allow these posts to appear in `getPostThread` specifically when they are the anchor post (i.e. at depth 0).  Otherwise the posts will appear in feeds but wont have a detail view, which I expect could be confusing.

packages/bsky/src/services/feed/views.ts
dholms: so a feed view post includes post views for it's replyParent & replyRoot

this basically says, the reply violates a block, then don't hydrated in the replyParent/Root

another downside of not naming this field `replyTo` 😅

devinivy (approved):

PR #1678, opened 2023-09-26
:sparkles: Let admins see post thread even when the author blocks them

foysalit:


packages/pds/src/api/app/bsky/feed/getPostThread.ts
```diff
@@ -22,25 +22,41 @@ import {
   getRepoRev,
   handleReadAfterWrite,
 } from '../util/read-after-write'
+import { authPassthru } from '../../../../../api/com/atproto/admin/util'
```
devinivy, line 25: I noticed a build error that might be related to this—is it possible there's one too many `../`?

packages/pds/src/api/app/bsky/feed/getPostThread.ts
```diff
@@ -22,25 +22,41 @@ import {
...
+          encoding: 'application/json',
+          body: res.data,
+        }
       } catch (err) {
-        if (err instanceof AppBskyFeedGetPostThread.NotFoundError) {
+        // TODO: is this right? checking for requester and error type is safe enough?
+        if (
+          err instanceof AppBskyFeedGetPostThread.NotFoundError &&
+          requester
+        ) {
```
devinivy, line 59: Yeah, I believe that should be fine 👍 Maybe there's a way to split the two cases-up more rather than have the `requester` checks sprinkled throughout.  For example, maybe the case could be offloaded up front since it's relatively simple on its own:

```ts
      const requester =
        auth.credentials.type === 'access' ? auth.credentials.did : null

      if (!requester) {
        const res = await ctx.appviewAgent.api.app.bsky.feed.getPostThread(
          params,
          authPassthru(req),
        )
        return {
          encoding: 'application/json',
          body: res.data,
        }
      }

      // ...
      // rest of code continues as originally written
      // ...
```

foysalit: ah yes, looks much better. just wasn't sure if the whole `handleReadAfterWrite` is needed for mods too so didn't spend any time thinking about the implementation. cleaned it up as suggested, thanks!

packages/pds/src/api/app/bsky/feed/getPostThread.ts
foysalit: hmm.. no idea how this happened though since it was auto-imported by vscode lol.

devinivy (approved):
Nice!

PR #1680, opened 2023-09-27
Check recent likes for suggested follows

dholms:
When calculating suggested follows, do so off of recent likes.

Previously, we were not ordering, and my quick testing showed that it was grabbing old likes which is not representative of the account current interests


estrattonbailey (approved):
Nice one!

PR #1684, opened 2023-09-27
improve tag validation

estrattonbailey:


packages/pds/src/repo/prepare.ts
```diff
@@ -301,6 +302,14 @@ function assertNoExplicitSlurs(rkey: string, record: RepoRecord) {
     toCheck += ' ' + record.displayName
   } else if (isPost(record)) {
     toCheck += record.tags?.join(' ')
+
```
devinivy, line 305: Not related to your work, but just noticed there may be a tiiiny baby bug on the line above this one: I expect it will append the string `undefined` when the post has no tags.

devinivy (approved):
Ace!

dholms (approved):
nice 👌

packages/pds/src/repo/prepare.ts
estrattonbailey: Ah, but that was my _previous_ work. Fixed!

devinivy (approved):

PR #1689, opened 2023-09-28
Add support for a bunny CDN img invalidator

devinivy:
Adds support to the appview to use an image invalidator to purge from a bunny.net CDN.


dholms (approved):
looks great 💪

dholms (approved):
👍

PR #1690, opened 2023-09-28
Refactor tests to make better use of dev-env

dholms:
- Use `TestNetworkNoAppview` in pds tests instead of the similar code in the pds test utils
- Combine the seed clients in bsky & pds into a single seed client class in dev-env


devinivy (approved):
Amazing, so pumped to have this cleaned-up!

PR #1694, opened 2023-09-29
Switch takedown ids back to ints on pds distribution

devinivy:
The column type for `takedownId` the only incompatible difference in the db model between the prod pds and distribution of the pds.  By aligning them we can avoid a tricky rollout of the pds distribution to prod without sacrificing very much.


packages/pds/src/db/migrations/20230929T213219699Z-takedown-id-as-int.ts
```diff
@@ -0,0 +1,17 @@
+import { Kysely, sql } from 'kysely'
+
+export async function up(db: Kysely<unknown>): Promise<void> {
+  await sql`
```
dholms, line 4: does this work on sqlite?

dholms (approved):

PR #1695, opened 2023-09-30
Rework logging

dholms:


packages/pds/src/logger.ts
```diff
@@ -1,49 +1,70 @@
-import pino from 'pino'
-import pinoHttp from 'pino-http'
-import { subsystemLogger } from '@atproto/common'
+import * as pino from 'pino'
+import * as pinoHttp from 'pino-http'
 import * as jwt from 'jsonwebtoken'
 import { parseBasicAuth } from './auth'
+import * as xrpc from '@atproto/xrpc-server'
```
devinivy, line 5:
```suggestion
import * as xrpc from '@atproto/xrpc-server'
import { parseBasicAuth } from './auth'
```

packages/xrpc-server/src/types.ts
```diff
@@ -2,23 +2,29 @@ import { IncomingMessage } from 'http'
...
+  logger?: LoggerOpts | pino.Logger
   validateResponse?: boolean
   payload?: {
     jsonLimit?: number
     blobLimit?: number
     textLimit?: number
   }
   rateLimits?: {
-    creator: RateLimiterCreator
+    enabled: true
```
devinivy, line 22: If `enabled` is always `true`, can we safely nix it altogether?

devinivy (approved):

PR #1699, opened 2023-10-01
Fix writing duplicate blocks in sync.getRepo

devinivy:
I think this was just a little hiccup from recent code shuffling around `getCarStream()`.  The blocks are written once from the call to `writeRows()` but were also getting written again inline.


dholms (approved):
oh duh - thx dude 🙇‍♂️

PR #1702, opened 2023-10-03
harden datetime verification

bnewbold:
This adds stronger datetime verification in the syntax package, including a bunch of interop test cases, and swaps that in during Lexicon verification.

The prior verification was pretty lax, so I suspect this might cause some problems to other devs, or with existing records, when we roll it out.


bnewbold:
Some updates after discussing with @devinivy earlier in the week.

This keeps a proper ensureValid, and that is what we should use for, eg, new record creation.

It also has `normalizeDatetime` which has the dual role of normalizing even valid datetimes for sorting in a database column; and being somewhat more lenient for existing createdAt records, but will still throw on totally bogus stuff (eg, empty string or "asfd"). And then `normalizeDatetimeAlways` if we need a drop-in for `toSimplifiedISOSafe`, which returns UNIX epoch time if `normalizeDatetime` would have thrown.

Also rebased on top of `main`.

The review or feedback i'm still not sure about is whether the full strict validation will only happen on record creation. If that is true, great! Otherwise, if it is also validated when deserializing existing records, maybe we need special logic to distinguish between "new record creation" and "existing record parsing" for this corner-case. A test for this would probably be good.

packages/lexicon/src/validators/formats.ts
```diff
@@ -1,24 +1,22 @@
...
+  ensureValidDatetime,
 } from '@atproto/syntax'
 import { validateLanguage } from '@atproto/common-web'
 
 export function datetime(path: string, value: string): ValidationResult {
   try {
-    if (!isValidISODateString(value)) {
-      throw new Error()
-    }
+    ensureValidDatetime(value)
```
dholms, line 14: yeah this will happen everywhere - on new record creations & when validating existing records

I think we leave this code path as-is for now, and I can add in the more rigorous validation into the pds indexing path

dholms (commented):
Looks great! 

Left a note about where the validation happens & what I think we should do there. I'm happy to tackle that

bnewbold:
If you could tackle that, would be much appreciated!

Noticed tests failed on this branch, I think just flakey, trying a re-run of failed.

packages/pds/src/repo/prepare.ts
```diff
@@ -127,6 +129,22 @@ export const assertValidRecord = (record: Record<string, unknown>) => {
...
 
+// additional more rigorous check on datetimes
+// this check will eventually be in the lex sdk, but this will stop the bleed until then
+export const assertValidCreatedAt = (record: Record<string, unknown>) => {
+  const createdAt = record['createdAt']
+  if (typeof createdAt !== 'string') {
+    return
+  }
+  try {
+    ensureValidDatetime(createdAt)
```
dholms, line 140: added the extra check here

This means records with bad timestamps will still pass "lexicon validation" (which we want so that old records don't break application views), but they will no longer be able to be created on the pds

this specific approach only works as is because the only datetimes on our records are `createdAt` times & every `createdAt` property is a datetime. if we get other datetimes, we'll need to tweak this logic

packages/syntax/tests/datetime.test.ts
```diff
@@ -0,0 +1,106 @@
...
+      if (line.startsWith('#') || line.length == 0) {
+        return
+      }
+      expectInvalid(line)
+    })
+  })
+})
+
+describe('normalization', () => {
+  it('normalizes datetimes', () => {
```
devinivy, line 71: I believe this fun vector originally seen from retr0.id still makes it through, where the normalized date ends-up before year zero, which is invalid:
```js
> normalizeDatetime('0000-01-01T00:00:00+01:00')
'-000001-12-31T23:00:00.000Z'
```

In `toSimplifiedISOSafe()` you'll find an extra validation check on the normalized output, which is what currently catches this.

dholms: just added a test for that, lmk how it looks 👍

devinivy (approved):
Feels nice to have this nailed down.

bnewbold:
I didn't see @dholms small patch, and added a bigger patch which verifies that the output of the normalization function is itself valid. And also just disallows dates starting `000` (aka, before year 0010) to ward off more trickery. The specs somewhat disclaim the ability to do very old dates already.

If Dan's fix is simpler/better we can just revert my patches. I resolved a merge conflict.

dholms:
Yup yup I think that looks fine! finally gonna get this in 💪

PR #1705, opened 2023-10-03
Pds sqlite refactor

dholms:
This refactors the PDS to use a single-tenant SQLite datastore.

Each user has their own SQLite file that stores their repo and private account state.
User databases are stored hierarchically as such: `/${dbDirectory}/${sha256Hex(did).slice(0,2)}/${did}`

Repo signing keys for each repo are stored alongside the SQLite file.

We also switch out the abstraction for interacting with user data to an `ActorStore`. The primary difference between the `ActorStore` and our previous "services" is that the `ActorStore` has different classes for reading & writing. Since SQLite does not support concurrent transactions, the `ActorStore` requires that you clearly "transact" with the store in order to make a write.

We also maintain an LRUCache for signing keys and databases. We allow 30k open file handles and 30k keys held in memory. When a database falls out of the cache, we ensure that we close the file handle.

We also introduce 3 separate SQLite databases for managing service state:
- service DB for account info, invite codes, refresh tokens etc
- did cache DB with only one table for caching DID resolutions
- sequencer DB with only one table for sequencing repo updates across all repos on a service.

Each of these SQLite files in run in WAL mode to allow for concurrent reads and streaming replication.

We intend to ship the PDS distribution with [Litestream](https://litestream.io/) or something similar


packages/pds/src/actor-store/index.ts
```diff
@@ -0,0 +1,220 @@
...
+  }
+
+  async close() {
+    const promises: Promise<void>[] = []
+    for (const key of this.dbCache.keys()) {
+      const got = this.dbCache.get(key)
+      this.dbCache.delete(key)
+      if (got) {
+        promises.push(got.close())
+      }
```
devinivy, line 160: I think that the `delete()` also attempts to dispose of the db.  Maybe the close is idempotent and a non-issue.  If it is an issue, I think the dispose function takes a `reason` parameter, and you could explicitly skip the disposal on deletes.

packages/pds/src/actor-store/record/reader.ts
```diff
@@ -213,56 +130,67 @@ export class RecordService {
...
-        q.where('backlink.linkToDid', '=', linkTo),
-      )
-      .where('record.did', '=', did)
+      .where('backlink.linkTo', '=', linkTo)
       .where('record.collection', '=', collection)
       .selectAll('record')
       .execute()
   }
+
+  // @NOTE this logic a placeholder until we allow users to specify these constraints themselves.
```
devinivy, line 170:
```suggestion
  // @NOTE this logic is a placeholder until we allow users to specify these constraints themselves.
```

packages/pds/src/db/util.ts
```diff
@@ -1,22 +1,15 @@
...
-import DatabaseSchema from './database-schema'
-
-export const actorWhereClause = (actor: string) => {
-  if (actor.startsWith('did:')) {
-    return sql<0 | 1>`"did_handle"."did" = ${actor}`
-  } else {
-    return sql<0 | 1>`"did_handle"."handle" = ${actor}`
-  }
-}
+import { DynamicReferenceBuilder } from 'kysely/dist/cjs/dynamic/dynamic-reference-builder'
```
devinivy, line 12: I wonder if you can get away with `ReferenceExpression`, which I think is public.

packages/pds/src/db/util.ts
```diff
@@ -1,22 +1,15 @@
 import {
   DummyDriver,
   DynamicModule,
+  Kysely,
```
devinivy, line 4: Now might be a good opportunity to update to the latest kysely.

packages/pds/src/did-cache/index.ts
```diff
@@ -1,17 +1,19 @@
...
+  db: DidCacheDb
   public pQueue: PQueue | null //null during teardown
 
   constructor(
-    public db: Database,
+    dbLocation: string,
     public staleTTL: number,
     public maxTTL: number,
   ) {
+    this.db = getDb(dbLocation)
```
devinivy, line 16: Having the storage encapsulated here makes me happy.

packages/pds/src/crawlers.ts
```diff
@@ -18,20 +23,22 @@ export class Crawlers {
...
-            { err, cralwer: agent.service.toString() },
-            'failed to request crawl',
-          )
-        }
-      }),
-    )
-    this.lastNotified = now
+    await this.backgroundQueue.add(async () => {
+      await Promise.all(
+        this.agents.map(async (agent) => {
```
devinivy, line 28: I notice the background queue task is awaited here, so it's not backgrounded relative to `notifyOfUpdate()`.  Is the purpose of putting this on the queue to limit concurrency or something along those lines?

packages/pds/src/db/db.ts
```diff
@@ -0,0 +1,113 @@
...
+  RootOperationNode,
+  QueryResult,
+  UnknownRow,
+} from 'kysely'
+import SqliteDB from 'better-sqlite3'
+
+const DEFAULT_PRAGMAS = {
+  journal_mode: 'WAL',
+  busy_timeout: '5000',
+}
```
devinivy, line 17: We might consider `strict: 'ON'` here too https://sqlite.org/src/wiki/StrictMode

packages/pds/src/service-db/migrations/index.ts
```diff
@@ -0,0 +1,5 @@
+import * as init from './001-init'
+
+export default {
+  ['001']: init,
```
devinivy, line 4: Basically nothing, just a tiny tweak for consistency with the other migration files.
```suggestion
  '001': init,
```

packages/pds/src/services/auth.ts
```diff
@@ -91,7 +91,6 @@ export class AuthService {
     this.db.assertTransaction()
     const token = await this.db.db
       .selectFrom('refresh_token')
-      .if(this.db.dialect !== 'sqlite', (qb) => qb.forUpdate())
```
devinivy, line 94: To mitigate this, we might shorten the distance between this select and the update below.  Or perform the update with an additional match on the selected `expiredAt` time to wrangle concurrent refreshes.

packages/pds/src/basic-routes.ts
```diff
@@ -25,7 +25,8 @@ export const createRouter = (ctx: AppContext): express.Router => {
       await sql`select 1`.execute(ctx.db.db)
     } catch (err) {
       req.log.error(err, 'failed health check')
-      return res.status(503).send({ version, error: 'Service Unavailable' })
+      res.status(500).send({ version, error: 'Service Unavailable' })
```
devinivy, line 28: Any special reason for the status code change?

packages/pds/tests/account-deletion.test.ts
```diff
@@ -135,49 +136,53 @@ describe('account deletion', () => {
...
 
+  it('deletes the users actor store', async () => {
+    const sqliteDir = network.pds.ctx.cfg.db.directory
+    const dbExists = await fileExists(path.join(sqliteDir, carol.did))
+    expect(dbExists).toBe(false)
+    const walExists = await fileExists(path.join(sqliteDir, `${carol.did}-wal`))
+    expect(walExists).toBe(false)
+    const shmExists = await fileExists(path.join(sqliteDir, `${carol.did}-shm`))
+    expect(shmExists).toBe(false)
+  })
```
devinivy, line 175: If I understand correctly these may not exist because the actual file lives at `${sqliteDir}/{hash}/${did}...`.

packages/pds/tests/sync/subscribe-repos.test.ts
```diff
@@ -366,93 +330,93 @@ describe('repo subscribe repos', () => {
...
-    )
-    const frames = await readTillCaughtUp(byFrame(ws))
-    ws.terminate()
-    expect(frames.length).toBe(1)
-    if (!(frames[0] instanceof ErrorFrame)) {
-      throw new Error('Expected ErrorFrame')
-    }
-    expect(frames[0].body.error).toBe('FutureCursor')
-  })
+  // it('account deletions invalidate all seq ops', async () => {
```
devinivy, line 333: Just a note that this remains commented.  I think we want to ensure tombstones stick around now.

packages/pds/tests/seeds/basic.ts
```diff
@@ -107,6 +107,8 @@ export default async (
     'tests/sample-img/key-landscape-small.jpg',
     'image/jpeg',
   )
+  // must ensure ordering of replies in indexing
+  await sc.network.processAll()
```
devinivy, line 111: Can you remind me why/where the ordering of replies is significant?  I think the appview at least should be permissive of this.

packages/pds/tests/races.test.ts
```diff
@@ -4,14 +4,14 @@ import { TestNetworkNoAppView } from '@atproto/dev-env'
 import { CommitData, readCarWithRoot, verifyRepo } from '@atproto/repo'
 import AppContext from '../src/context'
 import { PreparedWrite, prepareCreate } from '../src/repo'
-import SqlRepoStorage from '../src/sql-repo-storage'
-import { ConcurrentWriteError } from '../src/services/repo'
+import { Keypair } from '@atproto/crypto'
 
 describe('crud operations', () => {
```
devinivy, line 9:
```suggestion
describe('races', () => {
```

packages/pds/tests/invites-admin.test.ts
```diff
@@ -2,7 +2,7 @@ import { TestNetworkNoAppView, SeedClient } from '@atproto/dev-env'
 import AtpAgent from '@atproto/api'
 import { randomStr } from '@atproto/crypto'
 
-describe('pds admin invite views', () => {
+describe.skip('pds admin invite views', () => {
```
devinivy, line 5: Maybe we can un-skip now?  If not, might just be worth a note.

packages/pds/tests/file-uploads.test.ts
```diff
@@ -1,70 +1,51 @@
...
   let smallFile: Uint8Array
 
   it('handles client abort', async () => {
     const abortController = new AbortController()
-    const _putTemp = network.pds.ctx.blobstore.putTemp
-    network.pds.ctx.blobstore.putTemp = function (...args) {
+    const _putTemp = DiskBlobStore.prototype.putTemp
+    // const _putTemp = network.pds.ctx.blobstore.putTemp
+    DiskBlobStore.prototype.putTemp = function (...args) {
+      // network.pds.ctx.blobstore.putTemp = function (...args) {
```
devinivy, line 48:
```suggestion
    DiskBlobStore.prototype.putTemp = function (...args) {
```

packages/pds/src/actor-store/repo/sql-repo-transactor.ts
```diff
@@ -0,0 +1,109 @@
...
+    await Promise.all(
+      chunkArray(blocks, 500).map((batch) =>
+        this.db.db
+          .insertInto('repo_block')
+          .values(batch)
+          .onConflict((oc) => oc.doNothing())
+          .execute(),
+      ),
+    )
+  }
```
devinivy, line 65: It might be worth checking if 500 is actually possible here.  I think that sqlite maxes out at 999 bound parameters, and I could imagine there being more than that in this query.

packages/pds/package.json
```diff
@@ -24,10 +24,11 @@
     "build": "node ./build.js",
     "postbuild": "tsc --build tsconfig.build.json",
     "test": "../dev-infra/with-test-redis-and-db.sh jest",
+    "test:sqlite": "jest",
+    "test:sqlite-only": "jest --testPathIgnorePatterns /tests/proxied/*",
```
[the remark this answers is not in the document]
dholms: eh i had a hard time naming these commands. i basically wanted a command that I could use to run a test in sqlite mode: `pnpm test:sqlite tests/crud.test.ts` for instance. This command would only be used as "sqlite one-off"

and then `pnpm test:sqlite` will just run _all_ sqlite-compatible tests

happy to rename these if you have an idea

packages/dev-env/src/network.ts
```diff
@@ -61,26 +58,17 @@ export class TestNetwork extends TestNetworkNoAppView {
...
     const start = Date.now()
+    const lastSeq = await this.pds.ctx.sequencer.curr()
+    if (!lastSeq) return
     while (Date.now() - start < timeout) {
-      await wait(50)
-      if (!this.pds.ctx.sequencerLeader) {
-        throw new Error('Sequencer leader not configured on the pds')
-      }
-      const caughtUp = await this.pds.ctx.sequencerLeader.isCaughtUp()
-      if (!caughtUp) continue
```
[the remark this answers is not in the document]
dholms: yess 💪

packages/pds/src/actor-store/index.ts
```diff
@@ -0,0 +1,220 @@
...
+  backgroundQueue: BackgroundQueue
+}
+
+export class ActorStore {
+  dbCache: LRUCache<string, ActorDb>
+  keyCache: LRUCache<string, Keypair>
+
+  constructor(public resources: ActorStoreResources) {
+    this.dbCache = new LRUCache<string, ActorDb>({
+      max: 30000,
```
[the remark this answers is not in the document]
dholms: yeah good points - i'm not opposed to getting the size into config in this PR def seems nice to have

Agreed on long-inactive db handles, but also think that just decreasing the size of the cache can do quite a bit for that

packages/pds/src/actor-store/index.ts
```diff
@@ -0,0 +1,220 @@
...
+    })
+  }
+
+  private async getLocation(did: string) {
+    const didHash = await crypto.sha256Hex(did)
+    const subdir = path.join(this.resources.dbDirectory, didHash.slice(0, 2))
+    const dbLocation = path.join(subdir, `${did}.sqlite`)
+    const keyLocation = path.join(subdir, `${did}.key`)
+    return { subdir, dbLocation, keyLocation }
+  }
```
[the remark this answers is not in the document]
dholms: Ah yeah good point, tbh kinda forgot we planned that. I'll get that in 👍

packages/pds/src/actor-store/index.ts
```diff
@@ -0,0 +1,220 @@
...
+    await Promise.all([
+      await migrator.migrateToLatestOrThrow(),
+      await fs.writeFile(keyLocation, privKey),
+    ])
+
+    const result = await db.transaction((dbTxn) => {
+      const store = createActorTransactor(did, dbTxn, keypair, this.resources)
+      return fn(store)
+    })
+    this.dbCache.set(did, db)
```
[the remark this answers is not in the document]
dholms: good catch. i'll clean that up 👍

packages/pds/src/actor-store/index.ts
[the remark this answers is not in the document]
dholms: kk revamped cfg for this

packages/pds/src/actor-store/index.ts
```diff
@@ -0,0 +1,220 @@
...
+  async close() {
+    const promises: Promise<void>[] = []
+    for (const key of this.dbCache.keys()) {
+      const got = this.dbCache.get(key)
+      this.dbCache.delete(key)
+      if (got) {
+        promises.push(got.close())
+      }
+    }
+    await Promise.all(promises)
```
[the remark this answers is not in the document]
dholms: yup good call 👍

packages/pds/src/actor-store/index.ts
dholms: the close should be idempotent

I haven't noticed any issues from this in tests. One downside of relying on the cache to handle the dispose is that the `.delete()` function is synchronous & I don't see a way to track when it has finished disposing of the item

packages/pds/src/db/util.ts
dholms: switched it out & it seems to work fine 👌

packages/pds/src/did-cache/index.ts
dholms: yeah it's so much better

perfect fit tbh

packages/pds/src/db/util.ts
dholms: oohhh i'd really like to do that

a bit worried that the api mighta changed a decent bit tho. i can try upgrading on another branch & checkout how much work it'd be

packages/pds/src/crawlers.ts
dholms: ah hmmm yeah this should probably not be awaited 👍

packages/pds/src/db/db.ts
dholms: oh yup good call 👍

packages/pds/src/services/auth.ts
dholms: this gets tweaked in the account manager code. so i'll give it a look over there to make sure it all makes sense 👌

packages/pds/src/basic-routes.ts
dholms: nope this was from a strange failure in one of the tests & i think i forgot to change it back

packages/pds/tests/account-deletion.test.ts
dholms: yup missed this update 😅 

fixed it when i updated the fs location a few min ago 👌

packages/pds/tests/seeds/basic.ts
dholms: I think it's less permissive of it now that we check to ensure that reply parents are valid

basically if we don't have the parent yet, I think we mark it as invalid 🤔

packages/pds/tests/db.test.ts
```diff
@@ -20,372 +16,149 @@ describe('db', () => {
...
-        leader.session?.abortController.abort(new Error('Oops!'))
-      }
-      const result = await leader.run(task)
-      expect(result).toEqual({ ran: true, result: new Error('Oops!') })
+    await db.transaction(async (dbTxn) => {
+      expect(() => dbTxn.assertTransaction()).not.toThrow()
     })
   })
 
-  describe('appMigration()', () => {
```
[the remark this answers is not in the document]
dholms: yeah i'm actually wondering if we might want to delete it 🤔 

if not, agreed we should bring these back

packages/pds/src/actor-store/repo/sql-repo-transactor.ts
dholms: yeah good call. also i think sqlite might be happier with smaller chunk sizes anyway

packages/pds/src/db/util.ts
devinivy: Yeah, no biggie if it's not the right time!

packages/pds/tests/invites-admin.test.ts
dholms: yup we can 👍 forgot why this was skipped originally

packages/pds/src/api/com/atproto/sync/getRecord.ts
```diff
@@ -20,17 +20,22 @@ export default function (server: Server, ctx: AppContext) {
...
         ? CID.parse(params.commit)
         : await storage.getRoot()
+
       if (!commit) {
         throw new InvalidRequestError(`Could not find repo for DID: ${did}`)
       }
-      const proof = repo.getRecords(storage, commit, [{ collection, rkey }])
+      const carIter = repo.getRecords(storage, commit, [{ collection, rkey }])
+      const carStream = byteIterableToStream(carIter)
+      carStream.on('close', actorDb.close)
```
devinivy, line 35: I wonder if this db close may not be necessary.  If we keep it around my concern would be that it could stick in the db cache but be closed.

packages/pds/src/actor-store/db/migrations/001-init.ts
```diff
@@ -0,0 +1,104 @@
+import { Kysely } from 'kysely'
+
+export async function up(db: Kysely<unknown>): Promise<void> {
+  await db.schema
+    .createTable('repo_root')
+    .addColumn('rev', 'varchar', (col) => col.primaryKey())
+    .addColumn('cid', 'varchar', (col) => col.notNull())
+    .addColumn('indexedAt', 'varchar', (col) => col.notNull())
+    .execute()
```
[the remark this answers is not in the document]
dholms: kk decided to throw it in there 👍

packages/pds/src/sequencer/sequencer.ts
```diff
@@ -133,35 +156,65 @@ export class Sequencer extends (EventEmitter as new () => SequencerEmitter) {
...
         this.lastSeen = evts.at(-1)?.seq ?? this.lastSeen
+      } else {
+        this.triesWithNoResults++
+        // when no results, exponential backoff on pulling, with a max of a second wait
+        const waitTime = Math.min(Math.pow(2, this.triesWithNoResults), SECOND)
+        await wait(waitTime)
       }
+      this.pollPromise = this.pollDb()
     } catch (err) {
       log.error({ err, lastSeen: this.lastSeen }, 'sequencer failed to poll db')
```
devinivy, line 179: I could imagine treating the error case as being a "try with no results."  It might be a little safety mechanism in case it's constantly failing and spinning on this.

packages/pds/src/api/com/atproto/server/createAccount.ts
```diff
@@ -52,28 +53,42 @@ export default function (server: Server, ctx: AppContext) {
...
       // if the provided did document is poorly setup, we throw
-      const { did, plcOp } = await getDidAndPlcOp(ctx, handle, input.body)
+      const signingKey = await Secp256k1Keypair.create({ exportable: true })
+      const { did, plcOp } = await getDidAndPlcOp(
+        ctx,
+        handle,
+        input.body,
+        signingKey,
+      )
+      const commit = await ctx.actorStore.create(
```
devinivy, line 63: I think I may have mentioned it on the account manager PR, but it could be nice to clean this new repo/keypair up in the case that registration fails in a subsequent step.

devinivy (approved):
Left a good handful of notes and comments, but seriously wow  💎  tons of great simplifications in here and things feel very tidy.

packages/pds/tests/db.test.ts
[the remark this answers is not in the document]
devinivy: I'm sort of pro deleting it— Idon't think it's too hard to bring back if it ends-up being useful.

packages/pds/src/actor-store/index.ts
```diff
@@ -0,0 +1,220 @@
...
+
+  async destroy(did: string) {
+    const blobstore = this.resources.blobstore(did)
+    if (blobstore instanceof DiskBlobStore) {
+      await blobstore.deleteAll()
+    } else {
+      const db = await this.db(did)
+      const blobRows = await db.db.selectFrom('blob').select('cid').execute()
+      const cids = blobRows.map((row) => CID.parse(row.cid))
+      await Promise.allSettled(cids.map((cid) => blobstore.delete(cid)))
```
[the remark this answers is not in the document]
dholms: I agree that'd be really nice, but I don't think that s3 has a mechanism for deleting all based on a prefix :/

Ended up adding a `deleteMany` method and doing chunks of 500

packages/pds/src/services/auth.ts
dholms: okay got this into the account manager PR

you can check it out here: https://github.com/bluesky-social/atproto/pull/1769/files#r1378166057

packages/pds/src/actor-store/index.ts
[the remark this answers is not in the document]
devinivy: Yeah, I'm with ya— it would involve paginating over all items with that prefix using `ListObjects`.  The thought is just that we don't necessarily need the `blob` table to clear out all blobs being stored for that did.  In some ways it might be nice to just paginate over all the items and remove them from the bucket, no need to worry about any discrepancy between what's tracked in the `blob` table and what's actually stored for the user.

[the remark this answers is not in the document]
devinivy: The new approach you have in there seems great, this is small peas 👍

packages/pds/src/api/com/atproto/server/createAccount.ts
dholms: sweet handled it in the account manager PR since the route is a bit simpler there 👌

packages/pds/src/sequencer/sequencer.ts
dholms: good call switched that out 👍

packages/pds/tests/db.test.ts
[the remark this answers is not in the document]
dholms: cool cool deleted 👌

packages/pds/src/actor-store/index.ts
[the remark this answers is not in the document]
dholms: kk cool 👍 

let's keep that in our back pocket tho 👌

packages/pds/src/actor-store/index.ts
```diff
@@ -0,0 +1,220 @@
...
+    this.dbCache.delete(did)
+    if (got) {
+      await got.close()
+    }
+
+    const { dbLocation, keyLocation } = await this.getLocation(did)
+    await rmIfExists(dbLocation)
+    await rmIfExists(`${dbLocation}-wal`)
+    await rmIfExists(`${dbLocation}-shm`)
+    await rmIfExists(keyLocation)
```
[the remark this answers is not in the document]
dholms: I think they should just error out on the next query

but ideally they won't ever get a next query cause the account will be "deleted" so I don't think we need to handle that super gracefully in a user-facing way. would be worth some tests though 👍

packages/pds/src/actor-store/index.ts
devinivy: If we end-up needing to address that, we could keep a p-queue for disposing dbs and add to it during lrucache's `dispose()`.

npetrangelo:
Hey, I'm trying to research this PR because it seems like a large change; would any of you at Bluesky be willing to summarize the tradeoffs between the Postgres architecture before, and the SQLite architecture introduced by this PR?

PR #1718, opened 2023-10-07
makefile: fix 'logged' variant of run env

bnewbold:
Little ittybitty Makefile fix. This is a common gotcha (need to update PHONY and the actual command), not sure how to avoid the footgun though.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ run-dev-env: ## Run a "development environment" shell
 	cd packages/dev-env; pnpm run start
 
 .PHONY: run-dev-env-logged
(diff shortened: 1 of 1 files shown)


estrattonbailey (approved):
Looks right. What's the footgun here?

devinivy (approved):
💎

dholms (approved):

bnewbold:
@estrattonbailey the footgun is adding a new `make` target via copy/paste, and only updating the `PHONY` line, or only updating the name of the target, and not updating both together.

PR #1723, opened 2023-10-09
Simplify PDS moderation

dholms:
This removes the moderation info architecture (actions, reports, labels, etc) from the PDS and puts it solely in the appview.

This prepares the moderation system for multi-PDS federation.

The appview tracks canonical moderation state and fans out any needed actions (for instance takedowns) to the PDS of the relevant user.

It prompts a PDS of a takedown by calling `admin.updateSubjectState`. Similarly, it hydrates in relevant PDS state (such as emails/invites/etc) by calling `admin.getAccountInfo`.

For both of these routes, it uses an asymmetrically signed service auth JWT, issued by the service's DID.
Currently, these tokens are addressed to the did that they pertain to as we don't publicize the PDS's did anywhere. But happy to reconsider that if we have ideas 🤔 

Builds off of https://github.com/bluesky-social/atproto/pull/1725


packages/bsky/tests/admin/__snapshots__/get-moderation-action.test.ts.snap
```diff
@@ -0,0 +1,343 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
dholms, line 1: these tests are all basically exactly the same, but since I removed "pds" from the test title, it rewrote them all. 

only actual difference is that some action ids are changed because there's no admin action in the seed

packages/dev-env/src/const.ts
```diff
@@ -0,0 +1,3 @@
+export const ADMIN_PASSWORD = 'admin-pass'
```
dholms, line 1: split these out to keep consistent between pds & bsky

packages/bsky/src/context.ts
```diff
@@ -91,6 +98,23 @@ export class AppContext {
...
+      iss,
+      aud,
+      keypair: this.signingKey,
+    })
+  }
+
+  async pdsAdminAgent(did: string): Promise<AtpAgent> {
+    const data = await this.idResolver.did.resolveAtprotoData(did)
+    const agent = new AtpAgent({ service: data.pds })
+    const jwt = await this.serviceAuthJwt(did)
```
dholms, line 113: unsure of having the audience be the did in question, but seemed like the easiest thing for now 🤔 

we'll need to be careful to not have the appview sign tokens for other purposes in the meantime

packages/bsky/tests/admin/__snapshots__/get-record.test.ts.snap
```diff
@@ -1,18 +1,28 @@
...
+  "labels": Array [
+    Object {
+      "cid": "cids(0)",
+      "cts": "1970-01-01T00:00:00.000Z",
+      "neg": false,
+      "src": "user(0)",
+      "uri": "record(0)",
+      "val": "self-label",
+    },
+  ],
```
devinivy, line 18: Is it easy for us to account for this self-label now appearing here?

packages/bsky/tests/admin/repo-search.test.ts
```diff
@@ -74,7 +74,7 @@ describe('pds admin repo search view', () => {
     expect(res.data.repos[0].did).toEqual(term)
   })
 
-  it('finds repo by email', async () => {
+  it.skip('finds repo by email', async () => {
```
devinivy, line 77: Intending to keep this skipped?  If so, might just be worth a little note.

packages/bsky/src/context.ts
devinivy: Hmmm!  Bit of a thinker 🤔

packages/bsky/src/api/com/atproto/admin/reverseModerationAction.ts
```diff
@@ -77,12 +77,27 @@ export default function (server: Server, ctx: AppContext) {
...
+        const { did, subjects } = restored
+        const agent = await ctx.pdsAdminAgent(did)
+        await Promise.all(
+          subjects.map((subject) =>
+            agent.api.com.atproto.admin.updateSubjectStatus({
+              subject,
+              takedown: {
+                applied: false,
+              },
+            }),
```
devinivy, line 93: We have a helper `retryHttp()` for retrying http requests—might recommend plopping it in here and anywhere else we want to ensure a status gets pushed out to the PDS.

packages/bsky/src/api/com/atproto/admin/takeModerationAction.ts
```diff
@@ -98,12 +103,30 @@ export default function (server: Server, ctx: AppContext) {
...
 
-        return result
+        return { result, takenDown }
       })
 
+      if (takenDown) {
+        const { did, subjects } = takenDown
+        if (did && subjects.length > 0) {
+          const agent = await ctx.pdsAdminAgent(did)
+          await Promise.all(
```
devinivy, line 113: Total nit, but `allSettled()` might be nice here so that if one fails we still wait for others to succeed.

packages/bsky/tests/admin/moderation.test.ts
```diff
@@ -960,6 +962,86 @@ describe('moderation', () => {
...
+            headers: network.bsky.adminAuthHeaders(),
+          },
+        )
+
+      const res1 = await pdsAgent.api.com.atproto.admin.getSubjectStatus(
+        {
+          uri,
+        },
+        { headers: network.pds.adminAuthHeaders() },
+      )
```
devinivy, line 1030: Truly no problem here, but I bet you could get away with this if you like 😏 
```suggestion
      const res1 = await pdsAgent.api.com.atproto.admin.getSubjectStatus(
        { uri },
        { headers: network.pds.adminAuthHeaders() },
      )
```

packages/pds/src/api/com/atproto/admin/updateSubjectStatus.ts
```diff
@@ -35,11 +35,16 @@ export default function (server: Server, ctx: AppContext) {
...
+          try {
+            await modSrvc.updateBlobTakedownState(
+              subject.did,
+              CID.parse(subject.cid),
+              takedown,
+            )
+          } catch (err) {
+            console.log(err)
+            throw err
+          }
```
devinivy, line 47: Looks like this may be leftover from some debugging.

packages/pds/src/auth.ts
```diff
@@ -1,10 +1,15 @@
...
+  AuthRequiredError,
+  InvalidRequestError,
+  verifyJwt,
+} from '@atproto/xrpc-server'
 import * as ui8 from 'uint8arrays'
 import express from 'express'
 import * as jwt from 'jsonwebtoken'
 import AppContext from './context'
 import { softDeleted } from './db/util'
+import { IdResolver } from '@atproto/identity'
```
devinivy, line 12:
```suggestion
import { IdResolver } from '@atproto/identity'
import * as ui8 from 'uint8arrays'
import express from 'express'
import * as jwt from 'jsonwebtoken'
import AppContext from './context'
import { softDeleted } from './db/util'
```

devinivy: Ooh there's an ugly diff.  Just moved `@atproto/identity` up near its other `@atproto/` pals.

packages/pds/src/db/migrations/20231011T155513453Z-takedown-id-as-str.ts
```diff
@@ -0,0 +1,47 @@
+import { Kysely, sql } from 'kysely'
+import { Dialect } from '..'
+
+export async function up(db: Kysely<unknown>, dialect: Dialect): Promise<void> {
+  if (dialect === 'pg') {
+    await sql`
+      alter table "repo_root" alter column "takedownId" type varchar;
+      alter table "repo_blob" alter column "takedownId" type varchar;
+      alter table "record" alter column "takedownId" type varchar;
```
devinivy, line 9: I think for the rollout we'll need a slightly different approach, but shouldn't be too, too bad.

devinivy (approved):
Made note of a few odds and ends, but this looks great!

packages/pds/src/api/com/atproto/admin/updateSubjectStatus.ts
dholms: yup thx 👍

packages/bsky/src/context.ts
dholms: kk i left this with the did of the user as the `aud` & I explicitly verify the `aud` matches the request in question on the PDS (which dramatically reduces blast-radius of one of these leaking)

packages/bsky/tests/admin/__snapshots__/get-record.test.ts.snap
dholms: Yup the view is fulfilled by the appview now. The moderation views service on the PDS didn't hydrate in self labels 👍

packages/bsky/tests/admin/repo-search.test.ts
dholms: yeah it's not clear to me how this will function with multi-pds

it's also not that clear that me that search by email is useful. Altho we may want to add some way to resolve email -> did on the entryway

dholms: i'll add a @NOTE tho 👍

dholms: was gonna leave a ntoe, but I just decided to delete it instead. These are appview tests & nothing in the API contract suggests you need to be able to search on email

If this is an important feature, we can find another way of implementing

packages/bsky/src/api/com/atproto/admin/reverseModerationAction.ts
dholms: yeah good call 👍

packages/bsky/src/api/com/atproto/admin/takeModerationAction.ts
dholms: yup good point 👍 

do you think we should report an error if one does not succeed? or should we just log?

packages/bsky/tests/admin/moderation.test.ts
dholms: lol yup good looks - leftover from refactoring 😛

packages/pds/src/db/migrations/20231011T155513453Z-takedown-id-as-str.ts
dholms: yeah we can chat through what exactly this should look like 🤔

packages/bsky/src/api/com/atproto/admin/takeModerationAction.ts
devinivy: It's a hard call, but I guess I lean a little towards reporting an error that tells the user that it was partially applied.

devinivy (approved):

packages/bsky/src/api/com/atproto/admin/takeModerationAction.ts
dholms: cool cool 👍 

i ended up doing that with an `UpstreamFailureError`

PR #1724, opened 2023-10-09
Proxy for accounts across PDSes

devinivy:
A few things going on here:
 - Switch from `jsonwebtoken` to `jose` library to support secp256k1 keys for asymmetric signing of JWTs.
 - Setup rudimentary proxy to route user to their home pds when applicable.
   - Will do a second pass on this in a separate PR to make the proxying lighterweight and more faithful (e.g. proper/full header passthrough).
 - Serve token expiration errors when account migrates across pdses.

Ref: #1742 for some practical usage.


packages/pds/src/api/proxy.ts
```diff
@@ -0,0 +1,89 @@
...
+): Promise<T | null> => {
+  if (isThisPds(ctx, pdsDid)) {
+    return null // skip proxying
+  }
+  const accountService = ctx.services.account(ctx.db)
+  const pds = pdsDid && (await accountService.getPds(pdsDid))
+  if (!pds) {
+    throw new UpstreamFailureError('unknown pds')
+  }
+  // @TODO reuse agents
```
dholms, line 20: yeah unfortunately this really is necessary

packages/pds/src/auth.ts
```diff
@@ -27,92 +33,118 @@ export type AuthToken = {
...
+  aud?: string
 }
 
-export type RefreshToken = AuthToken & { jti: string }
+export type RefreshToken = AuthToken & { jti: string; aud: string }
 
 export class ServerAuth {
-  private _secret: string
+  private _signingSecret: KeyObject
+  private _signingKeyPromise?: Promise<KeyObject>
```
dholms, line 43: i'm inclined to put this behind an async static constructor so we don't have to manage the promise

packages/pds/src/auth.ts
```diff
@@ -1,16 +1,22 @@
...
 import { softDeleted } from './db/util'
 
 const BEARER = 'Bearer '
 const BASIC = 'Basic '
+const SECP256K1_JWT = 'ES256K'
+const HMACSHA256_JWT = 'HS256'
 
 export type ServerAuthOpts = {
   jwtSecret: string
+  jwtSigningKey?: crypto.Secp256k1Keypair
```
dholms, line 19: is the intention to always support either symmetric or asymmetric? or just as a cutover period thing?

packages/pds/src/auth.ts
```diff
@@ -324,3 +361,13 @@ export enum AuthStatus {
   Invalid,
   Missing,
 }
+
+const createPrivateKeyObject = async (
```
dholms, line 365: mer it's a shame we have to convert between these :/

If we need to use their key format, I don't think it'd be wrong to just set the config value as a pem/der formatted key 🤔

packages/pds/src/config/env.ts
```diff
@@ -74,6 +74,9 @@ export const readEnv = (): ServerEnvironment => {
 
     // secrets
     jwtSecret: envStr('PDS_JWT_SECRET'),
+    jwtSigningKeyK256PrivateKeyHex: envStr(
+      'PDS_JWT_SIGNING_KEY_K256_PRIVATE_KEY_HEX',
```
dholms, line 78: heh or we could do what bryan was pushing for & use multicodecs for our private keys to avoid stuff like this

could jam some multikey <> der utilities in @atproto/crypto

packages/pds/src/services/account/index.ts
```diff
@@ -588,6 +595,15 @@ export class AccountService {
       await this.db.db.insertInto('user_pref').values(putPrefs).execute()
     }
   }
+
+  // @TODO cache w/ in-mem lookup
+  async getPds(pdsDid: string) {
```
dholms, line 600: okay this is inspiring me to actually use the pdsDid in the mod refactor i did 😛

dholms (approved):
lookin good 👍

packages/pds/src/auth.ts
devinivy: In my view it would be great to support both, but it's only for cutover that we need to support them at the same time / on the same server.

packages/pds/src/api/proxy.ts
devinivy: My plan is for this to go away when we incorporate #1730 😈

packages/pds/src/auth.ts
devinivy: Going to comment below in context of the multicodecs idea 👍

packages/pds/src/config/env.ts
devinivy: I have a smattering of general thoughts on this, not sure how they work out together yet 🥴 😆 

 - For the purposes of configuration I think it would be pretty nice to support PEMs since so much tooling works well with it, and it's familiar to people.  Ideally we could accept something that openssl can spit out.  It's possible that these env vars could become file paths to keys.  Right now here's how keys are generated with common tools—not the end of the world, but feels like it could be better: https://github.com/bluesky-social/pds/blob/01dbd27728754dc765316e5e4a768380238bddc0/installer.sh#L16.
 - For the @atproto/crypto interfaces, I think it would be great in general to be able to convert to node's `KeyObject` interface and/or the webby `CryptoKey` / `CryptoKeyPair` interfaces (I lean towards the latter).  Seems very desirable to me for interop purposes!
 - Just an observation: we'll also want to configure public keys on their own, e.g. the jwt verify key in #1742.  Maybe we'll want a standard way to represent those in @atproto/crypto?

The multicodecs direction seems great for a ton of use-cases, but I'm not sure whether it's ideal for these app configurations.  I might need to learn a little more about how it would work, though.

PR #1725, opened 2023-10-09
Admin account state endpoints

dholms:
Some simple routes for toggling various subject status attributes.

This only includes the `takedown` attribute, but the basic idea is that other types of state (such as ability to login, mutate, download repo, etc) can included in a similar manner.

I also made only one route for every type of subject (repo, record, blob) and unbundled record & blob takedowns.

There's also a bit of a weird implementation detail where we translate between the lexicon represenation of state (boolean & optional ref) and our current database schema (arbitrary string where presence means the takedown is applied). We can clean this up in pds v2

Precursor to https://github.com/bluesky-social/atproto/pull/1723 & we'll probably merge/deploy both together. Just kept as a separate PR for ease of review


lexicons/com/atproto/admin/getSubjectState.json
```diff
@@ -0,0 +1,39 @@
...
+  "defs": {
+    "main": {
+      "type": "query",
+      "description": "Fetch the service-specific the admin state of a subject (account, record, or blob)",
+      "parameters": {
+        "type": "params",
+        "properties": {
+          "did": { "type": "string", "format": "did" },
+          "uri": { "type": "string", "format": "at-uri" },
+          "blob": { "type": "string", "format": "cid" }
```
devinivy, line 13: If you want the status for a blob do you end-up passing `(did, blob)`?

dholms: yeah that's what i was thinking

packages/pds/src/api/com/atproto/admin/getSubjectState.ts
```diff
@@ -0,0 +1,45 @@
...
+import AppContext from '../../../../context'
+import { OutputSchema } from '../../../../lexicon/types/com/atproto/admin/getSubjectState'
+import { AuthRequiredError, InvalidRequestError } from '@atproto/xrpc-server'
+
+export default function (server: Server, ctx: AppContext) {
+  server.com.atproto.admin.getSubjectState({
+    auth: ctx.roleVerifier,
+    handler: async ({ params, auth }) => {
+      const access = auth.credentials
+      // if less than admin access then cannot perform a takedown
```
devinivy, line 13:
```suggestion
```

packages/pds/src/api/com/atproto/admin/getSubjectState.ts
```diff
@@ -0,0 +1,45 @@
...
+      if (!access.moderator) {
+        throw new AuthRequiredError(
+          'Must be a full moderator to update subject state',
+        )
+      }
```
devinivy, line 18: There's no update happening here, so I think we can safely skip this check.

lexicons/com/atproto/admin/getSubjectState.json
```diff
@@ -0,0 +1,39 @@
+{
+  "lexicon": 1,
+  "id": "com.atproto.admin.getSubjectState",
```
devinivy, line 3: Do you envision having a separate endpoint for getting the subject state specific to a particular service?  Or beyond that, fetching the subject state for multiple services at once?

packages/pds/src/api/com/atproto/admin/updateSubjectState.ts
```diff
@@ -0,0 +1,56 @@
...
+} from '../../../../lexicon/types/com/atproto/admin/defs'
+import { isMain as isStrongRef } from '../../../../lexicon/types/com/atproto/repo/strongRef'
+import { AuthRequiredError, InvalidRequestError } from '@atproto/xrpc-server'
+
+export default function (server: Server, ctx: AppContext) {
+  server.com.atproto.admin.updateSubjectState({
+    auth: ctx.roleVerifier,
+    handler: async ({ input, auth }) => {
+      const access = auth.credentials
+      // if less than admin access then cannot perform a takedown
```
devinivy, line 17: This comment might be out-of-date since this check below permits non-admin moderators. Or maybe we need a more stringent check within the `if (takedown) {}` block below.

lexicons/com/atproto/admin/getSubjectState.json
devinivy, line 3: I see what you mean about naming being hard here—"subject state" is a little loaded.  Some other ideas to replace: "state": flags, statuses, notices, advisories.

packages/pds/src/db/tables/record.ts
```diff
@@ -7,7 +7,7 @@ export interface Record {
   rkey: string
   repoRev: string | null
   indexedAt: string
-  takedownId: number | null
+  takedownId: string | null
```
devinivy, line 10: I have a feeling you're already on it, but just a note to ensure we don't forget the db migration path for these.

packages/pds/src/api/com/atproto/admin/getSubjectState.ts
dholms: ah yup good call 👍

lexicons/com/atproto/admin/getSubjectState.json
dholms: yeah - i view this as just being "what is this particular service's state for this subject?" then the moderation service could have some higher-level end points for collating state from across services

packages/pds/src/api/com/atproto/admin/updateSubjectState.ts
dholms: ah good call 👍 

i updated the error text but not the comment

packages/pds/src/db/tables/record.ts
dholms: yup yup 👍

lexicons/com/atproto/admin/defs.json
```diff
@@ -2,6 +2,14 @@
   "lexicon": 1,
   "id": "com.atproto.admin.defs",
   "defs": {
+    "statusAttr": {
```
dholms, line 5: Swapped out "state" for "status" (inspired by Foysal's PR https://github.com/bluesky-social/atproto/pull/1617)

As well, differentiated between the names of these two:
"status attribute": an individual attribute of a subject's status. For instance, it's current takedown status
"status": all attributes of a subject's status

devinivy (approved):

dholms:
Closing because this is all contained in in https://github.com/bluesky-social/atproto/pull/1723

PR #1726, opened 2023-10-09
Fix typo in bsky test: purpose is #modlist not #blocklist

pfrazee:
(Correct me if I'm wrong) The purpose of a blocklist should be `#modlist` so I think this test is just a typo. Doesn't affect behaviors but might be unclear to folks.


devinivy (approved):
Yup, good call!

PR #1727, opened 2023-10-09
Facilitate authing w/ PDS based on DID doc

devinivy:
- [x] Add `didDoc` output to lexicons used to create or refresh a session.
 - [x] Allow configuring PDS to include `didDoc` on those lexicon methods.
 - [x] Update API to use the returned pds endpoint


packages/api/package.json
```diff
@@ -36,7 +36,8 @@
     "@atproto/xrpc": "workspace:^",
     "multiformats": "^9.9.0",
     "tlds": "^1.234.0",
-    "typed-emitter": "^2.1.0"
+    "typed-emitter": "^2.1.0",
+    "zod": "^3.21.4"
```
pfrazee, line 40: This isn't an addition to the api because its dependencies already depended on it

packages/api/src/did/did-doc.ts
```diff
@@ -0,0 +1,40 @@
...
+export type DidDocument = z.infer<typeof didDocument>
+
+export function isValidDidDoc(doc: unknown): doc is DidDocument {
+  return didDocument.safeParse(doc).success
+}
+
+export function getPdsEndpoint(doc: unknown): URL | undefined {
+  if (isValidDidDoc(doc)) {
+    const pds = doc.service?.find((s) => s.type === 'AtprotoPersonalDataServer')
+    if (pds && typeof pds.serviceEndpoint === 'string') {
```
pfrazee, line 32: Note: this is currently taking the first PDS service entry and only handling the string value. I don't know what a non-string value would look like.

pfrazee:
@devinivy I tested by fudging some responses but I don't have a great way to test this meaningfully or in the test suite, and I'm only somewhat confident that it's correct. We either need to cook up good automated tests or be really aggressive about testing in staging scenarios before deploying.

packages/api/src/did/did-doc.ts
```diff
@@ -0,0 +1,40 @@
+import { z } from 'zod'
+
+export const verificationMethod = z.object({
```
dholms, line 3: hmmm we already have something very similar this over in `@atproto/identity` - we may want to just use the code there instead of duplicating it over to the api package 🤔

dholms: or if the client doesn't agree with the dependencies of that package, porting just these types & verification methods over to `@atproto/common-web`

packages/api/src/did/did-doc.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export type DidDocument = z.infer<typeof didDocument>
+
+export function isValidDidDoc(doc: unknown): doc is DidDocument {
+  return didDocument.safeParse(doc).success
+}
+
+export function getPdsEndpoint(doc: unknown): URL | undefined {
+  if (isValidDidDoc(doc)) {
+    const pds = doc.service?.find((s) => s.type === 'AtprotoPersonalDataServer')
```
dholms, line 31: in `@atproto/identity`, we find the first service entry with id `#atproto_pds` &  then check the type

packages/api/src/did/did-doc.ts
dholms: yeah that'd be considered a valid json-ld did document, but not supported by atproto rn

this is similar to how we validate services in our identity resolver 👍

dholms (approved):
Sweet implementation seems sound 👌 

Only thing I'd push for is moving the did doc logic to a common location - either `@atproto/identity` or `@atproto/common-web`

devinivy:
@dholms took a quick swing at addressing your feedback.

devinivy:
@pfrazee yup I got it from here!

packages/api/src/did/did-doc.ts
```diff
@@ -0,0 +1,24 @@
+import { DidDocument, didDocument } from '@atproto/common-web'
+
+export function isValidDidDoc(doc: unknown): doc is DidDocument {
```
dholms, line 3: not required, but I think we could take this one step further & move this to common-web as well

Issue #1583, opened 2023-09-07
Support other S3 object storage providers


UpcraftLP:
It would be nice to be able to use other S3-compatible providers for object storage, such as [minio](https://github.com/minio/minio).

Here is a list of relevant S3 configuration options and whether or not the PDS currently supports them
| Property | Type | Supported |
|---|---|---|
| Endpoint URL | string | ❌ |
| Path-Style Bucket Access | boolean | ❌ |
| Bucket Name | string | ✔ |
| Access Key ID | string| ✔ |
| Secret Access Key | string |✔ |

Seeing as the PDS uses the aws sdk which explicitly supports this, all that is left would be exposing those configuration options to the server operator [here](https://github.com/bluesky-social/atproto/blob/main/services/pds/index.js#L139-L156), and passing them along [here](https://github.com/bluesky-social/atproto/blob/a7c42cfe3981e8f1491f8d7a09fde041a9e83afe/services/pds/index.js#L48).

bnewbold:
This is a reasonable request and would make self-hosting easier.

As context for others looking at this, the default setup for self-hosted PDS in the sandbox right now is to store blobs in a configurable directory on disk.

PR #1729, opened 2023-10-10
PDS: Allow configuring non-AWS S3 blob storage.

YOCKOW:
This PR is just intended to allow configuring non-AWS S3 blob storage.

See https://github.com/bluesky-social/atproto/issues/1583

### DIFF
--- a/packages/pds/src/config/config.ts
+++ b/packages/pds/src/config/config.ts
@@ -55,6 +55,15 @@ export const envToCfg = (env: ServerEnvironment): ServerConfig => {
   }
   if (env.blobstoreS3Bucket) {
     blobstoreCfg = { provider: 's3', bucket: env.blobstoreS3Bucket }
+    if (env.blobstoreS3Region) {
+      blobstoreCfg.region = env.blobstoreS3Region
+    }
+    if (env.blobstoreS3Endpoint) {
+      blobstoreCfg.endpoint = env.blobstoreS3Endpoint
+    }
+    if (env.blobstoreS3ForcePathStyle !== void 0) {
+      blobstoreCfg.forcePathStyle = env.blobstoreS3ForcePathStyle
+    }
   } else if (env.blobstoreDiskLocation) {
     blobstoreCfg = {
       provider: 'disk',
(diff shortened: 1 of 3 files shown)


UpcraftLP:
Hey o/
might I trouble you to also add a config for path style naming?
AWS prefers `bucket.region.endpoint/path` (with path-style being supported but not recommended), while some other providers only support path style naming, i.e. `region.endpoint/bucket/path`.

As I mentioned in the issue already, the AWS SDK fully supports this, it is only a matter of letting users enable it.

YOCKOW:
@UpcraftLP

Ah, I missed `forcePathStyle`. Added it.
Can we make this PR ready for review?

packages/pds/src/config/config.ts
```diff
@@ -55,6 +55,15 @@ export const envToCfg = (env: ServerEnvironment): ServerConfig => {
   }
   if (env.blobstoreS3Bucket) {
     blobstoreCfg = { provider: 's3', bucket: env.blobstoreS3Bucket }
+    if (env.blobstoreS3Region) {
+      blobstoreCfg.region = env.blobstoreS3Region
+    }
+    if (env.blobstoreS3Endpoint) {
+      blobstoreCfg.endpoint = env.blobstoreS3Endpoint
+    }
+    if (env.blobstoreS3ForcePathStyle !== void 0) {
```
UpcraftLP, line 64: any reason for using `void(0)` here instead of `undefined`?

YOCKOW: That's a kind of my habit because `undefined` was once writable and even nowadays it can be shadowed.
(Ref.: https://eslint.org/docs/latest/rules/no-undefined)

Alternative is

```typescript
    if (tyepof(env.blobstoreS3ForcePathStyle) == 'boolean') {
```

YOCKOW:
@bnewbold What do you think of this PR since you mentioned allowing non-AWS storage is reasonable?

packages/pds/src/config/config.ts
devinivy, line 64: Updating just for consistency with the rest of the codebase.
```suggestion
    if (env.blobstoreS3ForcePathStyle !== undefined) {
```

devinivy (approved):
Appreciate the contribution!  Resolves #1729

(Just a note— I know what's up with the CI failure, and it's not a result of this work.  The same issue should not happen once this is merged to main.)

PR #1730, opened 2023-10-10
Passthrough proxying on xrpc-server

devinivy:
This implements http passthroughs on xrpc-server and a proxy handler for forwarding the active request to another service.  It's also now used on the relevant proxied app.bsky routes on the PDS to the appview.  Some notes:
 - Can return `{ passthru: http.IncomingRequest }` from xrpc-server handlers, and the request will be faithfully passed through.
   - Updated server codegen to reflect this.
 - Implemented `xrpcServer.proxy()` which proxies the active request using a passthrough.
 - Fixes a header normalization issue in the xrpc client.  One "soft" breaking change is that `defaultFetchHandler()` is no longer responsible for header normalization, and assumes the headers passed to it have already been normalized to lower-case.
 - Where possible PDS routes for `app.bsky` are now proxied using `xrpcServer.proxy()`.  The exceptions are generally read-after-write endpoints.
   - These routes will now start passing through response headers from appview.


dholms (approved):
this is beautiful 😍 🙇‍♂️

devinivy:
Closing this but keeping the branch around—will probably redo this in the future and hit all proxying services.

PR #1733, opened 2023-10-11
Fix test seed & snap

dholms:
We were using the wrong selfLabels for bob (we were using the right one in the appview codebase!) & it was leading to weird hard to explain discrepancies in the snaps


devinivy (approved):
Nice catch 💎

PR #1738, opened 2023-10-12
helpers for rkey and tid syntax; validate rkey at record creation time

bnewbold:
More syntax validators, with interop tests.

cc: @ericvolp12 who has a popular feedgen record with a colon (`:`) in the rkey, which is disallowed by the spec (but wasn't being enforced by PDS), and is a character not in the RFC 3986 "Unreserved Characters" list, which is what the rkey syntax rules are based on. Colon is in `pchar` and we could expand to support it in theory. I'd kind of rather find a migration path for the current record... create a new record w/o colon and update all the subscriptions as a on-time fix? Regardless, my understanding is that this current PR would not break the existing feed generator, only prevent new records with `:` from being created.


ericvolp12:
Would this change prevent users from liking/adding-to-list a FeedGenerator with a `:` in the rkey?

packages/pds/src/repo/prepare.ts
```diff
@@ -170,6 +170,8 @@ export const prepareCreate = async (opts: {
   }
 
   const rkey = opts.rkey || nextRkey.toString()
+  // @TODO: validate against Lexicon record 'key' type, not just overall recordkey syntax
+  ensureValidRecordKey(rkey)
```
devinivy, line 174: Since this is coarse and lexicon-unaware, I wonder if we should push this check down into the repo code.  If the PDS is lexicon-aware, it could enforce the stricter key type check as a part of record validation.

bnewbold:
@ericvolp12 this PR as written does not change validation of AT URI fields, so it doesn't directly impact liking or add-to-list of existing records (like a feedgen record with `:`). The way the spec is currently written, we *should* be more restrictive on AT URIs as well, but that is a broader change I would not push through right now.

packages/pds/src/repo/prepare.ts
bnewbold: That makes sense to me in terms of correctness.

`repo`: I'm not super confident what the blast radius of impact would be of verifying NSID+RecordKey would be in, eg, `repo/src/util.ts` `parseDataKey()`. That seems to run both on writes and reads? So could break things like the existing with-`:` record. We could initially check only in the create/write path, maybe creating a separate `parseDataKeyPermissive` as a migration/transition thing.

`PDS`: I expect PDS to be lexicon-aware enough to do generic Lexicon validation, possibly by fetching/resolving (and caching) new Lexicon files in the future. For that, passing `rkey` to the "validate record" function would make sense to me.

devinivy: @dholms any feels on this?

packages/syntax/src/recordkey.ts
```diff
@@ -0,0 +1,26 @@
+export const ensureValidRecordKey = (rkey: string): void => {
+  if (rkey.length > 512 || rkey.length < 1) {
```
dholms, line 2: i'm pretty tempted to limit this to less than 512

I think we talked about 16 at one point (tho i think that's too low). but maybe 64?

packages/pds/src/repo/prepare.ts
dholms: I think it's a little bit more comfortable living here for the time being.

We don't do NSID validation on collection names in the repo code & that strikes me as similar to this.

So far the repo/MST is pretty unaware of these syntatic things & is primarily concerned with the tree structure & cbor encoding. Though I think there's a good argument for pushing some more lexicon/syntax awareness into it.

In short, I'd leave it here for now but consider adjusting where we draw the abstraction boundary

dholms (approved):
yup yup this looks good 👍

packages/syntax/src/recordkey.ts
bnewbold: I'm not totally against that, but there are a couple things to discuss. I opened an issue in the specs repo: https://github.com/bluesky-social/atproto-website/issues/250

dholms: cool cool - that doesn't have to hold this up 👍

PR #1739, opened 2023-10-12
Lexicons to support bring-your-own-did w/ per-repo signing keys

devinivy:
If a user wants to bring their own DID to account creation with a per-repo signing key, they may follow this process:

1. The client hits the PDS at `com.atproto.server.reserveSigningKey`.  The response contains `signingKey` as a string in the form of a did key (i.e. public).  The PDS stores this signing key with its generation time in order to ensure it is single-use and claimed in a reasonable amount of time.
2. The client signs a PLC op containing their repo signing key and their new PDS, possibly generating their DID for the first time.
3. The client hits the PDS at `com.atproto.server.createAccount` passing their signed `plcOp` as CBOR (using lexicon’s `bytes` type) in addition to their `handle`, etc.
4. The PDS validates the contents of the `plcOp` (PDS service endpoint, handle, repo signing key, etc.).  To validate the signing key, it ensures that it’s the public key for one of the signing keys that it has temporarily reserved via `com.atproto.server.reserveSigningKey`, that it has been used once in a reasonable time period.  Once validated, PLC op is submitted to PLC by the PDS and if accepted, the signing key is associated with the new account.


lexicons/com/atproto/server/getSigningKey.json
```diff
@@ -0,0 +1,23 @@
+{
+  "lexicon": 1,
+  "id": "com.atproto.server.getSigningKey",
```
dholms, line 3: I think I'd prefer an indication in the method name that this is for account creation.

Even if we go the verbose route and say `server.getSigningKeyForNewAccount` or  similar. Otherwise I feel like this looks like it just returns _the server's_ signing key

dholms (approved):

PR #1741, opened 2023-10-12
Support accounts on entryway PDS without a local repo

devinivy:
This service will manage accounts that may be hosted by other PDSes.  Here we're just moving some things around so that it's not assumed that each account has a repo hosted locally.


dholms (approved):
looks great 👍 

did something similar in my v2 work 👌

PR #1742, opened 2023-10-12
Misc tests and fixes for entryway PDS

devinivy:
Writing some tests on auth flows around the entryway PDS.  To make this work the test suite mocks now support multiple PDSes, and each PDS service can be marked as an entryway or not.  If the PDS is not an entryway it will not provide certain proxying functionality.  We also allow the PDS to be configured with a public key that can verify creds but not generate them itself.

PDS assignment has also gotten a bit more configurable, now supporting weighted assignment (which also allows us to enable/disable assignment to a given PDS).


packages/pds/src/config/secrets.ts
```diff
@@ -82,3 +91,8 @@ export type SigningKeyMemory = {
   provider: 'memory'
   privateKeyHex: string
 }
+
+export type VerifyKeyMemory = {
```
dholms, line 95: i think the verify key can always be memory 🤔

dholms (approved):
sickkk 💪

PR #1743, opened 2023-10-12
Split apart auth grant & verification

dholms:
This reworks the existing pds `ServerAuth` class into an `AuthVerifier`.

Along with this we move granting of refresh/access tokens to the auth service & wrap those operations in a higher level abstraction.

This is a step towards having a swappable auth grant abstraction. As every local PDS will be doing auth verification, but only some will be granting auth tokens (others will reach out to an identity provider).

This also cleans up the somewhat awkward mounting of the auth verifier methods onto `ctx`.


packages/pds/tests/app-passwords.test.ts
```diff
@@ -66,7 +66,7 @@ describe('app_passwords', () => {
     const attempt = appAgent.api.com.atproto.server.createAppPassword({
       name: 'another-one',
     })
-    await expect(attempt).rejects.toThrow('Token could not be verified')
+    await expect(attempt).rejects.toThrow('Bad token scope')
```
devinivy, line 69: Ha, I recognize this change!  Had a similar fix in #1724 :)

devinivy (approved):
Yeahhh, that's pretty nice 👌

PR #1744, opened 2023-10-13
Account creation flow between PDS and entryway

devinivy:
Implements the account creation flow between PDS and entryway, using the process described in #1739.

There may be some final changes to the lexicons in #1739, but everything is expected to keep the same shape.  Will let that work land before updating here.

Feature branch: #1722


packages/pds/src/api/com/atproto/server/createAccount.ts
```diff
@@ -262,12 +323,19 @@ const getDidAndPlcOp = async (
...
 const assignPds = async (ctx: AppContext) => {
+  if (!ctx.cfg.service.isEntryway) return
   const pdses = await ctx.db.db.selectFrom('pds').selectAll().execute()
   const idx = randomIndexByWeight(pdses.map((pds) => pds.weight))
   if (idx === -1) return
   return pdses.at(idx)
 }
 
+const reserveSigningKey = async (host: string) => {
+  const agent = new AtpAgent({ service: getPdsEndpoint(host) })
```
dholms, line 334: it'd be better if we only instantiate the agent once since it can be kinda expensive

but not the end of the world rn

dholms (approved):
sweet looks good 👍

PR #1749, opened 2023-10-14
Fix read after write on threads when posting a media embed

dholms:
We weren't wrapping it in an `external` container.

Added some tests for external & image embeds


devinivy (approved):
Nice find, that was a real brain itch!

PR #1757, opened 2023-10-20
:sparkles: Expose emailConfirmedAt field from admin getRepo

foysalit:
This will help us view which users have verified their email address and make informed moderation decisions.


packages/pds/src/db/tables/repo-root.ts
```diff
@@ -5,6 +5,7 @@ export interface RepoRoot {
   rev: string | null
   indexedAt: string
   takedownId: number | null
+  emailConfirmedAt: string | null
```
devinivy, line 8: I believe this column comes from the `user_account` table rather than the `repo_root` table.

packages/pds/tests/admin/get-repo.test.ts
```diff
@@ -91,6 +91,40 @@ describe('pds admin get repo view', () => {
...
+    const verificationToken = await network.pds.ctx.services
+      .account(network.pds.ctx.db)
+      .createEmailToken(sc.dids.bob, 'confirm_email')
+    await agent.api.com.atproto.server.confirmEmail(
+      { email: bobsAccount.email, token: verificationToken },
+      {
+        encoding: 'application/json',
+        headers: {
+          authorization: `Bearer ${bobsAccount.accessJwt}`,
+        },
```
devinivy, line 113: I think you can use `sc.getHeaders(sc.dids.bob)` here rather than manually grabbing Bob's token and building the header.

foysalit:
Addressed both @devinivy thanks!

packages/pds/src/services/moderation/views.ts
```diff
@@ -605,7 +606,10 @@ export class ModerationViews {
   }
 }
 
-type RepoResult = DidHandle & RepoRoot
+type RepoResult = DidHandle &
+  RepoRoot & {
+    emailConfirmedAt?: string | null
```
devinivy, line 611: Is there a reason this field should be optional?  My concern is that it could create inconsistencies in the resulting views, where some responses contain `emailConfirmedAt` and others don't.

packages/pds/src/services/moderation/views.ts
```diff
@@ -21,6 +21,7 @@ import { AccountService } from '../account'
 import { RecordService } from '../record'
 import { ModerationReportRowWithHandle } from '.'
 import { ids } from '../../lexicon/lexicons'
+import { UserAccount } from '../../db/tables/user-account'
```
devinivy, line 24: I think we can nix this now that it isn't used 👍

packages/pds/src/services/moderation/views.ts
```diff
@@ -179,8 +181,7 @@ export class ModerationViews {
           'in',
           results.map((r) => didFromUri(r.uri)),
         )
-        .selectAll('repo_root')
-        .selectAll('did_handle')
+        .selectAll(['repo_root', 'did_handle'])
```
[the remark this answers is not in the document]
foysalit: so, the idea here is to only expose this field through `admin.getRepo()` endpoint and the lexicon property `com.atproto.admin.defs#repoViewDetail`. the primary use-case for this right now is to just show email verified accounts with a special flag on ozone's repository view page so trying to avoid exposing this on every `repoView` slice.
which is also why the field itself is defined as optional since the only time we're actually fetching it from the db before passing it along to the view is from the `account().getAccount()` [call](https://github.com/bluesky-social/atproto/blob/main/packages/pds/src/api/com/atproto/admin/getRepo.ts#L13)

LMK if that tracks.

packages/pds/src/services/moderation/views.ts
foysalit: clarified in the [above comment](https://github.com/bluesky-social/atproto/pull/1757#discussion_r1376151206)

packages/pds/src/services/moderation/views.ts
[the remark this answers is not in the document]
devinivy: Ahh, I am getting it now, thanks!

Since `emailConfirmedAt` is part of `RepoViewDetail` but not `RepoView`, I think it would be sort of nice/tidy for that difference to be reflected in `repoDetail()` the same way the other differences between those two views are.  With this change there's a bit of coupling between how the result set is generated versus whether it's suitable to be passed to `repo()` versus `repoDetail()`, and I wonder if it might be a little simpler to always select `emailConfirmedAt` then place the logic to include the field in the view inside `repoDetail()`.

All that said this is quite a small thing, I certainly wouldn't block on this 👍

devinivy (commented):
Looks good!  However, I think post-merge you'll find that many of the existing PDS changes can go away and that `admin.getAccountInfo` will need to be updated.  The appview's `admin.getRepo` looks like it now calls out to the PDS's `admin.getAccountInfo` for invite/email details.

foysalit:
@devinivy brought in some more changes, mind giving it one more look before I hit merge?

devinivy (approved):
Great!

Issue #1622, opened 2023-09-18
ListItemView doesn't have its own uri


gildaswise:
**Is your feature request related to a problem? Please describe.**

I've implemented curated lists in [deck.blue](https://deck.blue) and it's mostly working fine, but for the removal process of a list item, I need to query the `app.bsky.graph.listitem` repo, check if any item matches the list I'm trying to manage, and then filter the related records just to get their ListItem uris to use with deleteRecord.

**Describe the solution you'd like**

This whole process would be easier and more performant if [ListItemView](https://github.com/bluesky-social/atproto/blob/17ce804ffaaafb1b1efb61174217a1da275feda8/lexicons/app/bsky/graph/defs.json#L41C6-L41C18) just had the uri property as its [Record](https://github.com/bluesky-social/atproto/blob/main/lexicons/app/bsky/graph/listitem.json) version does when returned by [listRecords](https://github.com/bluesky-social/atproto/blob/main/lexicons/com/atproto/repo/listRecords.json).

intrnl:
feel like it might also be worthwhile for the getList endpoint to also *not* hide list items where the subject user no longer exists...

we can turn the ListItemView into the following
```
export interface ListItemView {
  uri: AtUri;
  did: DID;
  subject: ProfileView | null;
}
```

PR #1758, opened 2023-10-20
Attach record URI to listItemView

intrnl:
Fixes https://github.com/bluesky-social/atproto/issues/1622

### DIFF
--- a/lexicons/app/bsky/graph/defs.json
+++ b/lexicons/app/bsky/graph/defs.json
@@ -40,8 +40,9 @@
     },
     "listItemView": {
       "type": "object",
-      "required": ["subject"],
+      "required": ["uri", "subject"],
       "properties": {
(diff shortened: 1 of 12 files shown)


intrnl:
Not really sure how the tests are being done, it seems like running `pnpm test:withFlags --updateSnapshot` is resulting in a lot of diffs being generated, with the notable one being the invalid handles that all of the users now has

I don't think I've strayed from the instructions, even starting the dev environment leads to all the users having invalid handle.

gildaswise:
Is there any issue with this one? It would really help managing actor removal on lists

dholms (approved):
This looks good to me & I agree the uri of the item should come down on the view

This is technically a breaking change in the lexicon, but it should be fine to roll out while we're not federated yet. I'll double check with @pfrazee on this to make sure it's good

pfrazee:
Yeah this is safe for now

PR #1759, opened 2023-10-20
Fix CI issues

dholms:
- Upgrade express & ws (turned out not to be necessary)
- Use axios instead of fetch for network mocks so we can mock the host header
- Normalize output from docker for dev-infra


dholms:
Closed in favor of https://github.com/bluesky-social/atproto/pull/1763

PR #1764, opened 2023-10-23
Prevent clashing redis namespaces in tests

dholms:
We got bday paradoxed 🙃 

Turns out both `appview_views_admin_get_moderation_reports` & `bsky_views_posts` hash to `5312`

I bumped this from 10k to 1m. Altho I'm curious if there's a reason that we require going to an int here instead of just using the string from dbPostgresSchema @devinivy?

### DIFF
--- a/packages/dev-env/src/bsky.ts
+++ b/packages/dev-env/src/bsky.ts
@@ -81,7 +81,7 @@ export class TestBsky {
     })
     // indexer
     const ns = cfg.dbPostgresSchema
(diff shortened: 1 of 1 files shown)


devinivy (approved):
Looks good!  If I recall correctly, the reason an integer is used here is mainly to avoid the redis commands getting very long.  For example, when the indexers pull from multiple partitions, each partition will be listed in the command with its namespace.  It also probably sidesteps issues with invalid key names since integers are always valid.  But it's not a major deal either way.

PR #1765, opened 2023-10-24
Respect updated service auth keys

devinivy:
When a service auth key is updated, the appview was not immediately recognizing the new key due to caching.  With this change, we now attempt to update the cache when verification fails so that we immediately respect updated keys.


packages/bsky/tests/auth.test.ts
```diff
@@ -0,0 +1,64 @@
+import AtpAgent from '@atproto/api'
+import { SeedClient, TestNetwork } from '@atproto/dev-env'
+import usersSeed from './seeds/users'
+import { createServiceJwt } from '@atproto/xrpc-server'
+import { Keypair, Secp256k1Keypair } from '@atproto/crypto'
+
+describe('auth', () => {
```
dholms, line 7: these tests _might_ make more sense on `xrpc-server`. thoughts?

dholms (approved):
looks good 👌 thx for hitting this!

packages/bsky/tests/auth.test.ts
devinivy: Yeah, I can see that.  I am inclined to leave this test here since it's an important feature of the appview (and it could easily be implemented incorrectly), but I will add unit tests to xrpc-server for `verifyJwt()` 👍

dholms (approved):
yup yup looks good 👍

PR #1766, opened 2023-10-24
Miscellaneous query optimizations on appview

devinivy:
Three miscellaneous query optimizations here:
 - On the author feed `posts_with_media` filter, we simplify the condition that ensures only the user's own posts are displayed (i.e. not media in their reposts).  By not referencing the `post` table here the query loosens, allowing it to do more filtering early in the query plan and make better use of the cursor index.  Hoping to see lower IO on this query.
 - The query for notification skeleton was a bit bloated with joins and returned columns that went unused.  By fetching notification record contents separate from the skeleton, the intent is to make simpler queries that postgres is able to cache more efficiently.  Hoping to see lower IO on this query.
 - Fetching a single actor by handle occurs on many routes, and was hitting a degenerate case.  Postgres would choose the gist trigram index originally used for typeahead search rather than the unique btree index on `handle`.  The btree index is much quicker, so we've employed a little trick to ensure it's chosen instead of the trigram index.

### DIFF
--- a/packages/bsky/src/api/app/bsky/feed/getAuthorFeed.ts
+++ b/packages/bsky/src/api/app/bsky/feed/getAuthorFeed.ts
@@ -89,8 +89,8 @@ export const skeleton = async (
 
   if (filter === 'posts_with_media') {
     feedItemsQb = feedItemsQb
-      // and only your own posts/reposts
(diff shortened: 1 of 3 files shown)


dholms (approved):
💎  these looks great!

PR #1770, opened 2023-10-25
Provide proper error for updating to an email that's already in use

devinivy:
Previously users would receive a 500 when attempting to change their email to one that is already in use.  With this change they should receive a 400 with an actionable error message.


packages/pds/src/db/util.ts
```diff
@@ -57,3 +57,11 @@ export const dummyDialect = {
 export type DbRef = RawBuilder | ReturnType<DynamicModule['ref']>
 
 export type AnyQb = SelectQueryBuilder<any, any, any>
+
+export const isErrUniqueViolation = (err: unknown) => {
```
dholms, line 61: amazing 💎

dholms (approved):
looks great 🙌

PR #1772, opened 2023-10-25
Insert to did cache using excluded column

devinivy:
Rather than passing the whole did doc twice when upserting into the did cache, we update it using the excluded column.


dholms (approved):
beautiful

PR #1776, opened 2023-10-25
Include own replies to own post roots via new filter `posts_and_author_threads`

estrattonbailey:
Update here: to avoid breaking any other clients, we've opted to keep `posts_no_replies` as-is, and add a new `posts_and_author_threads` filter that includes the logic needed by the frontend. I'll make another corresponding frontend PR to update the filter param.

Original post below.

***

Since we moved `posts_no_replies` logic to the backend, we've lacked self-threads on the default "Posts" view of profiles.

This PR expands that query to include replies, but ONLY if those replies are (1) by the author you're viewing and (2) their reply root is a post by the author you're viewing.

As noted in https://github.com/bluesky-social/social-app/issues/1166#issuecomment-1679276475, this results in returning replies that are "detached" from the root by a post by another user. That part will be handled on the frontend for now, at least until we move to our new backend architecture.


packages/bsky/src/api/app/bsky/feed/getAuthorFeed.ts
```diff
@@ -99,9 +99,21 @@ export const skeleton = async (
...
       )
   } else if (filter === 'posts_no_replies') {
-    feedItemsQb = feedItemsQb.where((qb) =>
-      qb.where('post.replyParent', 'is', null).orWhere('type', '=', 'repost'),
-    )
+    feedItemsQb = feedItemsQb
+      .where((qb) =>
+        qb.where('post.replyParent', 'is', null).orWhere('type', '=', 'repost'),
+      )
+      .orWhere((qb) =>
```
dholms, line 106: i think this `.orWhere` needs to make it's way into a `.where` wrapper. otherwise it might conflict with some other where statements (like limiting originator did to the author)

dholms: actually the `where('originatorDid', '=', actorDid)` is outside of this whole if block

so you might just be able to tag that whereExists into the previous where clause as an `orWhereExists`

devinivy: Just to toss out another option, here avoiding the subquery/join:
```ts
qb.where('post.replyParent', 'is', null)
  .orWhere('post.replyRoot', 'like', `at://${actorDid}/%`)
  .orWhere('type', '=', 'repost'),
```

estrattonbailey: Yep good call, can consolidate these. And good idea on the `like` check!

estrattonbailey:
Hey @mary-ext 👋 

>could there be an additional condition

If I understand what you're asking, yeah, a replyParent could have been written by the same actor and could match too. But that replyParent would also have a replyRoot that would match, so this check _should_ capture both.

>could be catastrophic for profiles

Could you elaborate on what would happen in this case? For `A -> B -> A` reply trees, this PR should flatten them to only return `A -> A`. The logic here is (intended to be, anyway): get all posts and reposts for an actor, PLUS any replies for an actor that are written by the actor in response to a "root" post by the same actor.

The frontend will then filter out any "detached" threads, as mentioned in https://github.com/bluesky-social/social-app/issues/1166#issuecomment-1679276475. The result should be that a user should see only posts, reposts, and their own _unbroken_ threads.

I'll double check on this stuff! Thanks for the comment, lmk if I'm off-base or not making sense 😄

estrattonbailey:
@mary-ext yep exactly, you're correct. This PR will include all replies by Alice and send them down to the frontend.

The frontend then does the rest of the work. For now, we're going filter out any posts that are not part of an un-broken chain back to a root post by the user. So for the edge case here https://github.com/bluesky-social/social-app/issues/1166#issuecomment-1679276475, the reply by Bob and the two replies to Bob from Alice will not appear on Alice's `posts_no_replies` tab.

packages/api/src/moderation/const/labels.ts
```diff
@@ -99,23 +99,23 @@ export const LABELS: LabelDefinitionMap = {
     strings: {
       settings: {
         en: {
-          name: 'Requested Hidden to Logged-out Users',
+          name: 'Sign-in Required',
```
dholms, line 102: guess this codegen just didn't quite make it into a previous PR?

wanna make sure this PR contains the correct content & not main

dholms (approved):
sweet this looks great 👌

PR #1784, opened 2023-10-30
Support running pds-in-a-box with an entryway service

devinivy:
When running pds-in-a-box with an entryway service, the entryway governs: a. auth, b. identity/handles, c. the PLC rotation key.  The PDS does the actual repo hosting and holds the repo signing keys.  Here we allow the PDS to delegate to an entryway for the relevant functionality, and setup a testing environment that uses pds-in-a-box with a real entryway service.


packages/pds/src/api/com/atproto/admin/updateAccountHandle.ts
```diff
@@ -27,7 +28,19 @@ export default function (server: Server, ctx: AppContext) {
           throw new InvalidRequestError(`Handle already taken: ${handle}`)
         }
       } else {
-        await ctx.plcClient.updateHandle(did, ctx.plcRotationKey, handle)
+        if (ctx.cfg.entryway) {
+          // the pds defers to the entryway for updating the handle in the user's did doc.
+          // here was just check that the handle is already bidirectionally confirmed.
+          // @TODO if handle is taken according to this PDS, should we force-update?
```
dholms, line 34: i think probably yes because the entryway should be canonically deciding who gets which handle and the PDS should just inherit its decision

dholms (approved):
this looks great 🙌

packages/pds/src/account-manager/helpers/auth.ts
```diff
@@ -10,70 +12,61 @@ export type AuthToken = {
...
-  return {
-    payload: payload as RefreshToken, // exp set by sign()
-    jwt: jwt.sign(payload, jwtSecret, {
-      expiresIn: expiresIn,
-      mutatePayload: true,
-    }),
-  }
+}): Promise<string> => {
+  const { did, jwtKey, jti = getRefreshTokenId(), expiresIn = '90days' } = opts
+  // @TODO set alg header? audience?
```
devinivy, line 55: I lean towards yes, and to validate the audience assuming `cfg.service.did` is set.
```suggestion
  // @TODO set audience to pds did?
```

dholms: we chatted about this, but I agree 👍

PR #1785, opened 2023-10-30
Entryway update handle flow, pds and agent caching

devinivy:
A few misc fixes, improvements, and tweaks to the entryway setup:
 - Fix `getAccountByEmail()` which was giving a `null` for the user's did.
 - Cache `getPds(did)` results which is in the hot path of proxying.
 - Cache `AtpAgent`s for the entryway's pdses.
 - Update entryway flow for updating account handles: should use pds's `admin.updateAccountHandle` endpoint to avoid entryway and pds `identity.updateHandle` routes depending on each other.
 - Only include `didDoc` alongside credentials when the pds endpoint is one of the entryway's pdses.

### DIFF
--- a/packages/pds/package.json
+++ b/packages/pds/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@atproto/pds",
-  "version": "0.3.0-entryway.0",
+  "version": "0.3.0-entryway.1",
   "license": "MIT",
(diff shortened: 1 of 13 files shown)


dholms (approved):
looks great 💪

PR #1787, opened 2023-10-30
Support S3-compatible credentials for pds blobstore

devinivy:
As followup to #1729, also enabling configuration of S3-compatible credentials for the PDS blobstore.  This can be used with some services such as R2.

### DIFF
--- a/packages/pds/src/config/config.ts
+++ b/packages/pds/src/config/config.ts
@@ -1,3 +1,4 @@
+import assert from 'node:assert'
 import os from 'node:os'
 import path from 'node:path'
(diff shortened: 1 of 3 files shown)


packages/pds/src/config/config.ts
```diff
@@ -1,3 +1,4 @@
+import assert from 'node:assert'
```
dholms, line 1: looks like this isn't ever used

dholms (approved):
looks good 👍

packages/pds/src/config/config.ts
devinivy: Oops!
```suggestion
```

PR #1793, opened 2023-11-01
Set and validate token audiences on pds v2

devinivy:
Here we start setting and validating audience claims on access and refresh tokens on the pds.  This is particularly useful when using the pds with an entryway service, which sets the audience claim in the credentials that it generates.

### DIFF
--- a/packages/pds/src/auth-verifier.ts
+++ b/packages/pds/src/auth-verifier.ts
@@ -83,7 +83,11 @@ export type AuthVerifierOpts = {
   adminPass: string
   moderatorPass: string
   triagePass: string
-  adminServiceDid: string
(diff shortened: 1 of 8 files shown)


packages/pds/src/auth-verifier.ts
```diff
@@ -205,14 +212,14 @@ export class AuthVerifier {
     const payload = await verifyServiceJwt(
       jwtStr,
       null,
-      async (did: string) => {
-        if (did !== this.adminServiceDid) {
+      async (did, forceRefresh) => {
```
dholms, line 215: nice

dholms (approved):
yup looks great 👌

PR #1794, opened 2023-11-01
Delegate create account from entryway

dholms:


packages/pds/src/api/com/atproto/server/createAccount.ts
```diff
@@ -17,60 +17,10 @@ export default function (server: Server, ctx: AppContext) {
...
-      const { did, plcOp } = await getDidAndPlcOp(
-        ctx,
-        handle,
-        input.body,
-        signingKey,
-      )
+      const { did, handle, email, password, inviteCode, signingKey, plcOp } =
+        ctx.entrywayAgent
+          ? await validateInputsForEntrywayPds(ctx, input.body)
+          : await validateInputsForLocalPds(ctx, input.body)
```
DavidBuchanan314, line 23: This ternary is backwards! Or rather, it logically does the right thing, but the names of the two functions are backwards. Unless I'm mistaken, `validateInputsForEntrywayPds` contains the Local PDS validation logic, and vice versa. The functions should perhaps have their names swapped?

devinivy: You're right, it's confusing! Will take a look at fixing the names, adding better comments 👍 There are three different setups kicking around, and this package handles two of them:

 - Being an entryway (not covered here)
 - Being a pds behind an entryway, e.g. the mushroom servers (validateInputsForEntrywayPds())
 - Being a pds that doesn't use an entryway, e.g. a self-hoster (validateInputsForLocalPds())

DavidBuchanan314: Ahh that makes sense, thanks

PR #1800, opened 2023-11-02
Pds v2 migrate script

dholms:


packages/pds/src/migrate-script/index.ts
```diff
@@ -0,0 +1,417 @@
...
+    .selectAll()
+    .execute()
+  let pdsCounter = 0
+  for (const status of todo) {
+    let pdsId = status.pdsId
+    if (!pdsId) {
+      pdsId = pdsCounter % pdsInfos.length
+      pdsCounter++
+    }
+    const pdsInfo = pdsInfos[pdsId]
```
devinivy, line 48: This currently uses the pds id and the index of the pds in the `pdsInfos` array interchangeably, thought they could differ from each other.

packages/pds/src/migrate-script/db.ts
```diff
@@ -0,0 +1,96 @@
...
+  async getMigrations() {
+    return {
+      '1': {
+        async up(db: Kysely<unknown>) {
+          await db.schema
+            .createTable('status')
+            .addColumn('did', 'varchar', (col) => col.primaryKey())
+            .addColumn('pdsId', 'integer')
+            .addColumn('signingKey', 'varchar')
+            .addColumn('phase', 'integer')
```
devinivy, line 26: We might make this a not-null to match the types below.

packages/pds/src/migrate-script/index.ts
```diff
@@ -0,0 +1,417 @@
...
+      await db
+        .updateTable('status')
+        .set({ failed: 1 })
+        .where('did', '=', status.did)
+        .execute()
+      console.error(`failed to migrate: ${status.did}`, err)
+    }
+    pdsCounter++
+  }
+  console.log('DONE WITH ALL')
```
devinivy, line 62: Haaa!!

packages/pds/src/migrate-script/index.ts
```diff
@@ -0,0 +1,417 @@
...
+      console.log(`completed migrating: ${status.did}`)
+    } catch (err) {
+      await db
+        .updateTable('status')
+        .set({ failed: 1 })
+        .where('did', '=', status.did)
+        .execute()
+      console.error(`failed to migrate: ${status.did}`, err)
+    }
+    pdsCounter++
```
devinivy, line 60: I think this may be handled up in the `!pdsId` block?

packages/pds/src/migrate-script/index.ts
```diff
@@ -0,0 +1,417 @@
...
+    did: row.did,
+    url: `https://${row.host}`,
+    agent: new AtpAgent({ service: `https://${row.host}` }),
+  }))
+  const todo = await db
+    .selectFrom('status')
+    .where('status.phase', '<', 7)
+    .where('failed', '!=', 1)
+    .selectAll()
+    .execute()
```
devinivy, line 40: Should this have a well-defined order?

packages/pds/src/migrate-script/index.ts
```diff
@@ -0,0 +1,417 @@
...
+    const importedRev = await lockAndTransfer(ctx, db, pds, status)
+    status.importedRev = importedRev
+    status.phase = TransferPhase.transferred
+    await updateStatus(db, status)
+  }
+
+  console.log(`prefs (${status.did}, ${status.phase})`)
+  if (status.phase < TransferPhase.preferences) {
+    try {
+      await transferPreferences(ctx, pds, status.did)
```
devinivy, line 101: May want to remove the entry from `failed_pref` if it succeeds here on the second go.

devinivy: Ah, actually I suppose we'll want a separate routine for handling the failures since `status.phase` still moves along during a failure 👍

packages/pds/src/migrate-script/index.ts
```diff
@@ -0,0 +1,417 @@
...
+      headers: { 'content-type': 'application/vnd.ipld.car' },
+      decompress: true,
+      responseType: 'stream',
+    },
+  )
+
+  for await (const log of importRes.data) {
+    if (typeof log === 'string') {
+      const lines = log.split('\n')
+      for (const line of lines) {
```
devinivy, line 170: I am not sure if we get guarantees that the chunks will be full lines, or if they could stop in the middle of a line.  If it's an issue, could pipe through [binary-split](https://www.npmjs.com/package/binary-split).

packages/pds/src/migrate-script/index.ts
```diff
@@ -0,0 +1,417 @@
...
+        if (line.indexOf('failed to import blob') > 0) {
```
devinivy, line 171: If this string is at the beginning of the line (i.e. at index `0`) we'd actually miss it!  I like to use `includes()` to avoid the footgun:
```suggestion
        if (line.includes('failed to import blob')) {
```

packages/pds/src/migrate-script/index.ts
```diff
@@ -0,0 +1,417 @@
...
+  const defer = createDeferrable()
+  ctx.db
+    .transaction(async (dbTxn) => {
+      const storage = new SqlRepoStorage(dbTxn, status.did)
+      await storage.lockRepo()
+      await defer.complete
+    })
+    .catch((err) => {
+      console.error(`error in repo lock tx for did: ${status.did}`, err)
+    })
```
devinivy, line 197: If this does fail there may be some important moments below where we don't want to continue.  Could we set a variable in the `catch()` and then assert that it's not set whenever we want below?

packages/pds/src/migrate-script/index.ts
```diff
@@ -0,0 +1,417 @@
...
+
+  let importedRev
+  try {
+    importedRev = await doImport(
```
devinivy, line 201: Do we want to ensure the lock is already taken by the time this starts?  Not sure if it's overly paranoid.

packages/pds/src/migrate-script/index.ts
```diff
@@ -0,0 +1,417 @@
...
+    await ctx.db.db
+      .updateTable('user_account')
+      .where('did', '=', status.did)
+      .set({ pdsId: pds.id })
+      .execute()
+    await ctx.db.db
+      .updateTable('repo_root')
+      .where('did', '=', status.did)
+      .set({ did: `migrated-${status.did}` })
+      .execute()
```
devinivy, line 253: These are unlikely to fail, but it's tempting to wrap these in a transaction.

packages/pds/src/migrate-script/index.ts
dholms: oh lol could call

packages/pds/src/migrate-script/db.ts
dholms: yup yup

packages/pds/src/migrate-script/index.ts
dholms: oh duh good catch

packages/pds/src/migrate-script/index.ts
dholms: yup yup good looks 👍

packages/pds/src/migrate-script/index.ts
dholms: easy enough to add in 👍

packages/pds/src/migrate-script/index.ts
dholms: yeah good call - also attempted to track the state where these fail but the rest of the transfer succeeds since that'd be a pretty subtle failure case

PR #1801, opened 2023-11-02
Add changeset for #1790

gaearon:
I want to release https://github.com/bluesky-social/atproto/pull/1790 so we can update the client and get the perf win.

I tried to word the change from the perspective of someone updating (they might have to explicitly declare the type).

### DIFF
--- a/.changeset/dry-cameras-hang.md
+++ b/.changeset/dry-cameras-hang.md
@@ -1,0 +1,7 @@
+---
+'@atproto/lexicon': minor
+'@atproto/xrpc': minor
+'@atproto/xrpc-server': minor
+---
+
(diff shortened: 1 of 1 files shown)


gaearon:
Do we need to update this example as well? https://github.com/bluesky-social/atproto/blob/main/packages/xrpc-server/README.md#usage

I think it might need an explicit `LexiconDoc` type definition now.

dholms:
Yup looks like that README could use the explicit type def. As well, most of the tests in xrpc-server could use the typedef as well. 

You up for hitting that?

gaearon:
https://github.com/bluesky-social/atproto/pull/1807

PR #1804, opened 2023-11-02
Simplify moderation fanout to configured url

devinivy:
The moderation functionality on the appview would fan out status changes to the affected user's pds.  We may want to do something like that eventually, but for now we simplify down to a single configured url.


dholms (approved):
this is a really nice simplification. shoulda gone with this earlier 👌 

we may want to tighten auth on pds routes (to not allow for the service auth) in a future PR

PR #1805, opened 2023-11-02
Entryway proxying for admin subject status endpoints

devinivy:
Ensures entryway proxies status updates to the user's pds, pairs with #1804.


dholms (approved):
nice! dig the simplification in parsing too 👌

PR #1807, opened 2023-11-02
Adjust types to use LexiconDoc

gaearon:
Follow-up to https://github.com/bluesky-social/atproto/pull/1790, as discussed in https://github.com/bluesky-social/atproto/pull/1801.


gaearon:
Alternatively I'm down to re-export `LexiconDoc` from `/xrpc-server` — which would take care of potential versioning issues if those packages aren't updated in sync.

dholms (approved):
looks good! i think this is fine 👍

PR #1810, opened 2023-11-02
Add a test for links that end in .php

estrattonbailey:
Noticed when debugging a ticket that we support `.php` links but didn't have a test.


devinivy (approved):
Good call!

PR #1812, opened 2023-11-03
feat: infer definition type from "types" argument in getDefOrThrow

matthieusieben:
This change ensures that:
1) the `types` argument contains only correct value types (no typo)
2) the returned value type is properly narrowed when a `types` argument is provided, removing the need to type cast the returned value.

### DIFF
--- a/packages/lex-cli/src/codegen/lex-gen.ts
+++ b/packages/lex-cli/src/codegen/lex-gen.ts
@@ -7,10 +7,7 @@ import {
   LexArray,
   LexPrimitive,
   LexBlob,
-  LexXrpcProcedure,
-  LexXrpcQuery,
   LexToken,
-  LexXrpcSubscription,
   LexCidLink,
   LexBytes,
   LexIpldType,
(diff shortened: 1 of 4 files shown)


packages/lex-cli/src/codegen/lex-gen.ts
```diff
@@ -306,12 +303,9 @@ export function genXrpcInput(
   lexUri: string,
   defaultsArePresent = true,
 ) {
-  const def = lexicons.getDefOrThrow(lexUri, [
-    'query',
-    'procedure',
-  ]) as LexXrpcProcedure
+  const def = lexicons.getDefOrThrow(lexUri, ['query', 'procedure'])
```
matthieusieben, line 306: There might be a bug in the code here as the type casting did not match the expected types. Not sure what is expected here.

devinivy (approved):
This is very useful (and slick!), thanks for the contribution.

matthieusieben:
rebased on main (and adapted)

PR #1818, opened 2023-11-04
Test entryway session handling

devinivy:
- Test `AtpAgent` handles switch from entryway to pds based on `didDoc` served with sessions.
 - Test transparent session refresh when user is assigned a new pds.


dholms (approved):
yup yup looks good 👍

PR #1819, opened 2023-11-05
PDS v2 entryway account deletion

devinivy:
Implements a new `com.atproto.admin.deleteAccount` endpoint on pds v2.  This supports account deletion across a pds and its entryway.  The proposed flow for account deletion  across a pds and its entryway is:
```
pds(server.deleteAccount)            [dumb proxy to entryway]
  -> entryway(server.deleteAccount)  [performs acct deletion on entryway then calls to pds]
  -> pds(admin.deleteAccount)        [performs acct deletion on pds]
```


dholms (approved):
yup looks great 👍

PR #1822, opened 2023-11-06
perf(bsky): avoid re-creating auth functions on every request

matthieusieben:
This is a minor change that will impact performances of the auth utilities


dholms (approved):
Looks great 👍 

Fixed a lil merge conflict in it

PR #1823, opened 2023-11-06
fix(bsky): cf image invalidator not taken into account when bunny used

matthieusieben:
The current logic prevents using two image invalidators. Indeed, the `if (bunnyInvalidator && imgInvalidator)` check is never `true` since `imgInvalidator` was never initialized at this point. The "simple" fix is to change `imgInvalidator` to `cfInvalidator` at this place. The approach I took is less error prone by confining the initialization of individual invalidators to a single place (`if (has config for invalidator) imgInvalidators.push(new invalidator)`).

### DIFF
--- a/services/bsky/api.js
+++ b/services/bsky/api.js
@@ -79,31 +79,31 @@ const main = async () => {
 
   const signingKey = await Secp256k1Keypair.import(env.serviceSigningKey)
 
-  // configure zero, one, or both image invalidators
-  let imgInvalidator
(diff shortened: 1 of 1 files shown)


dholms (commented):
This looks good to me, but just checking @devinivy do we even use the CloudFront CDN/invalidator anymore?

devinivy (approved):
Thanks!

devinivy:
@dholms I think it makes sense to keep the cloudfront invalidator in here for a little while longer 👍

PR #1824, opened 2023-11-06
fix(pds): include aspectRatio on read-sticky posts

mary-ext:
- adds missing `aspectRatio` property when providing read-sticky posts
- adds missing test case for alt text


dholms (approved):
Thanks for the catch - looks great!

PR #1825, opened 2023-11-06
chore(dev): prevent tests from running if a command fails to run

matthieusieben:
Running tests (`make test`) without `docker` installed result in the script silently failing to start docker, followed by a lot of failing tests (because no database is available). This change makes it more obvious that docker should be installed for the tests to run.

This also allows to run tests (and the dev env) without actually needing docker, provided that a Postgres URI is provided. The tests relying heavily on timing of operations, being able to run them in a non virtualized environment (esp. from MacOS), makes their execution much more reliable and fast.

### Testing `packages/pds` alone
With Docker (notice the 2.5Gb of RAM used by the VM):
Capture d’écran 2023-11-07 à 12 54 03
The failing test is due to a race condition (they do run successfully from time to time).

Without Docker (pg & redis installed on the host):
Capture d’écran 2023-11-07 à 12 46 20


packages/dev-infra/_common.sh
```diff
@@ -36,8 +39,80 @@ export_redis_env() {
   export REDIS_HOST="127.0.0.1:6380"
 }
 
-# Main entry point
-main() {
+pg_clear() {
```
devinivy, line 42: Since this action is quite destructive, I wonder if it would be worthwhile to enforce that the db name or user includes "test", or something along those lines.  Let me know what you think.

matthieusieben: This is already the case. `pg_clear` is called from `cleanup` only if the "service" used is `db_test` which must come from the `DB_TEST_POSTGRES_URL` env var (notice that the name contains TEST here)

matthieusieben: Capture d’écran 2023-12-05 à 11 57 04

The following appears in the console.

matthieusieben:
Capture d’écran 2023-12-05 à 11 57 48

Running tests without docker requires that environment variables `DB_TEST_POSTGRES_URL` and `REDIS_TEST_HOST` are set.

packages/dev-infra/_common.sh
devinivy: That should be good 👍

PR #1826, opened 2023-11-06
style(xrpc-server): avoid un-neccessary "if" statement

matthieusieben:
The `if` is always "truthy" here.


dholms (approved):
yup yup good find :thum

PR #1827, opened 2023-11-06
Security: prevent timing attack

matthieusieben:
In the current implementation; a timing attack can be used to deduce if a particular email or identifier is being used on a particulat PDS. This PR mitigates this attack by ensuring that the `createSession` handler always takes the same amount of time.


matthieusieben:
WIP (still need tests)

PR #1833, opened 2023-11-07
pre-compile handlebar templates at build time

matthieusieben:
- any compilation error will occur at buildtime instead of runtime
- makes builds smaller (10.1mb + 96k templates => 10.0mb) because only the handlebars runtime is needed (not the compiler)
- makes the server start faster (removed 4 `fs.readFileSync` every time the `ServerMailer` is instantiated)
- future proof (stacking `fs.readFileSync` calls at boot time is a bad idea) as more hbs file might be coming (oauth?)


dholms (approved):
love it! thanks for knocking that out 🙏

matthieusieben:
Fixed build and rebased on main (after pds-v2)

PR #1835, opened 2023-11-07
Handle quarantining/unquarantining a blob that does not exist

dholms:
Update subject status should be idempotent, but right now we throw if you try to quarantine a blob that either does not exist or is already quarantined

### DIFF
--- a/packages/aws/src/s3.ts
+++ b/packages/aws/src/s3.ts
@@ -187,7 +187,7 @@ export class S3BlobStore implements BlobStore {
 }
 
 const handleErr = (err: unknown) => {
-  if (err && typeof err === 'object' && err['Code'] === 'NoSuchKey') {
+  if (err?.['Code'] === 'NoSuchKey') {
     throw new BlobNotFoundError()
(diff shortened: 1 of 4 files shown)


packages/aws/src/s3.ts
```diff
@@ -170,15 +170,27 @@ export class S3BlobStore implements BlobStore {
...
+        Key: keys.from,
+      })
+    } catch (err) {
+      handleErr(err)
+    }
+  }
+}
+
+const handleErr = (err: unknown) => {
+  if (err && typeof err === 'object' && err['Code'] === 'NoSuchKey') {
```
devinivy, line 190: I believe this is totally typesafe:
```suggestion
  if (err?.['Code'] === 'NoSuchKey') {
```

devinivy (approved):
Lookin good 👍

packages/aws/src/s3.ts
dholms: typescript yelled at me for that :/

PR #1836, opened 2023-11-08
Use TTL cache for ActorStore

dholms:
Replace LRUCache on ActorStore with a TTLCache to prevent open handles from sticking around for an indeterminate amount of time.


packages/pds/src/actor-store/index.ts
```diff
@@ -31,44 +32,60 @@ type ActorStoreResources = {
...
 
+  private async fetchCacheKey(did: string) {
+    const got = this.keyCache.get(did)
+    if (got) return got
+    const { keyLocation } = await this.getLocation(did)
+    const privKey = await fs.readFile(keyLocation)
+    const keypair = await crypto.Secp256k1Keypair.import(privKey)
+    this.keyCache.set(did, keypair)
+    return keypair
+  }
```
devinivy, line 69: Without some kind of coalescing or abort mechanism, it's may not be too difficult for multiple callers to get past the `if (got) return got` line in parallel and each end-up generating the cached item.  I think the cache handles the disposal on `set()`, so hopefully it at least wouldn't result in a leak 👍

dholms:
closing in favor of https://github.com/bluesky-social/atproto/pull/1837

PR #1837, opened 2023-11-08
No cache actor store

dholms:
Refactor the actor store to close databases immediately after use instead of holding them in a cache


packages/pds/src/actor-store/index.ts
```diff
@@ -83,44 +53,46 @@ export class ActorStore {
...
+    return crypto.Secp256k1Keypair.import(privKey)
   }
 
-  async db(did: string): Promise<ActorDb> {
-    const got = await this.dbCache.fetch(did)
-    if (!got) {
+  async openDb(did: string): Promise<ActorDb> {
+    const { dbLocation } = await this.getLocation(did)
+    const exists = await fileExists(dbLocation)
+    if (!exists) {
```
devinivy, line 64: If sqlite throws an error when the db doesn't exist, perhaps we could use that rather than the eager check.  Would just save some ticks to open a db.

packages/pds/src/actor-store/index.ts
```diff
@@ -83,44 +53,46 @@ export class ActorStore {
...
   async read<T>(did: string, fn: ActorStoreReadFn<T>) {
-    const reader = await this.reader(did)
-    return fn(reader)
+    const keypair = await this.keypair(did)
+    const db = await this.openDb(did)
+    try {
+      const reader = createActorReader(did, db, keypair, this.resources)
+      return await fn(reader)
+    } finally {
+      await db.close()
```
devinivy, line 78: I'm not sure that we need to `await` the close here, it may be fine to let node just manage those in the background. 
 Though in that case especially, we'd want to `catch()` any log errors that may result during the close to avoid unhandled rejections.

packages/pds/src/actor-store/index.ts
```diff
@@ -83,44 +53,46 @@ export class ActorStore {
...
-  async reader(did: string) {
-    const [db, keypair] = await Promise.all([this.db(did), this.keypair(did)])
-    return createActorReader(did, db, keypair, this.resources)
+    return getDb(dbLocation, this.cfg.disableWalAutoCheckpoint)
   }
 
   async read<T>(did: string, fn: ActorStoreReadFn<T>) {
-    const reader = await this.reader(did)
-    return fn(reader)
+    const keypair = await this.keypair(did)
```
devinivy, line 72: I believe this keypair is used with `ActorStoreReader.transact`, which may not be used anywhere.  So might be able to nix it?

packages/pds/src/api/com/atproto/sync/getRecord.ts
```diff
@@ -19,17 +20,27 @@ export default function (server: Server, ctx: AppContext) {
...
+        const carIter = repo.getRecords(storage, commit, [{ collection, rkey }])
+        carStream = byteIterableToStream(carIter)
+      } catch (err) {
+        await actorDb.close()
+        throw err
       }
-      const carIter = repo.getRecords(storage, commit, [{ collection, rkey }])
-      const carStream = byteIterableToStream(carIter)
+      carStream.on('error', actorDb.close)
+      carStream.on('close', actorDb.close)
```
devinivy, line 42: I think that `close()` may not be bound to `actorDb`, we might need something like this (but shouldn't be too bad to test):
```suggestion
      const closeDb = () => actorDb.close()
      carStream.on('error', closeDb)
      carStream.on('close', closeDb)
```

packages/pds/src/api/com/atproto/sync/getRepo.ts
```diff
@@ -36,16 +35,19 @@ export const getCarStream = async (
...
   } catch (err) {
+    await actorDb.close()
     if (err instanceof RepoRootNotFoundError) {
       throw new InvalidRequestError(`Could not find repo for DID: ${did}`)
     }
     throw err
   }
-  return byteIterableToStream(carIter)
+  carStream.on('error', actorDb.close)
+  carStream.on('close', actorDb.close)
```
devinivy, line 51: Same here, worth checking!

packages/pds/src/actor-store/index.ts
dholms: I don't think it does unfortunately :/

And actually, calling `getDb` would cause the database to be _created_ i beliee

packages/pds/src/actor-store/index.ts
dholms: yup good call 👍

packages/pds/src/api/com/atproto/sync/getRecord.ts
dholms: yup yup good catch 👍

packages/pds/src/actor-store/index.ts
dholms: Ended up making `close` on the database class synchronous & moving the `catch()` logic to that method

Thinking is that it should protect against more footguns in catching errors while closing dbs, such as in the callbacks on streams (elsewhere in this PR)

PR #1844, opened 2023-11-10
Randomize suggestions

dholms:
Return randomized suggestions instead of always in order.

However, always return the first two in order - bsky.app & atproto.com in our case

Since this breaks our previous pagination mechanism which relied on order, we prevent duplicate results by building a cursor out of all previously included ids concatenated together.

### DIFF
--- a/packages/bsky/src/api/app/bsky/actor/getSuggestions.ts
+++ b/packages/bsky/src/api/app/bsky/actor/getSuggestions.ts
@@ -71,10 +71,13 @@ const skeleton = async (
   const rest = suggestions.filter((row) => row.order !== 1 && row.order !== 2)
   const limited = firstTwo.concat(shuffle(rest)).slice(0, params.limit)
 
-  const cursor = limited
-    .map((row) => row.order.toString())
-    .concat(alreadyIncluded.map((id) => id.toString()))
-    .join(':')
+  const cursor =
+    limited.length > 0
+      ? limited
+          .map((row) => row.order.toString())
+          .concat(alreadyIncluded.map((id) => id.toString()))
+          .join(':')
+      : undefined
 
   return { params, suggestions: limited, cursor }
 }
--- a/packages/bsky/src/api/app/bsky/actor/getSuggestions.ts
+++ b/packages/bsky/src/api/app/bsky/actor/getSuggestions.ts
@@ -71,6 +71,7 @@ const skeleton = async (
   const rest = suggestions.filter((row) => row.order !== 1 && row.order !== 2)
   const limited = firstTwo.concat(shuffle(rest)).slice(0, params.limit)
 
+  // if the result set ends up getting larger, consider using a seed included in the cursor for for the randomized shuffle
   const cursor =
     limited.length > 0
       ? limited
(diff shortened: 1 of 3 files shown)


packages/bsky/tests/views/suggestions.test.ts
```diff
@@ -63,18 +65,26 @@ describe('pds user search views', () => {
...
+    const result3 = await agent.api.app.bsky.actor.getSuggestions(
+      { limit: 1, cursor: result2.data.cursor },
+      { headers: await network.serviceHeaders(sc.dids.carol) },
+    )
+    expect(result3.data.actors.length).toBe(0)
+    expect(result3.data.cursor).toBeUndefined()
   })
 
+  return
+
```
devinivy, line 87:
```suggestion
```

packages/bsky/src/api/app/bsky/actor/getSuggestions.ts
```diff
@@ -57,27 +57,29 @@ const skeleton = async (
...
+  const rest = suggestions.filter((row) => row.order !== 1 && row.order !== 2)
+  const limited = firstTwo.concat(shuffle(rest)).slice(0, params.limit)
+
+  const cursor =
+    limited.length > 0
+      ? limited
+          .map((row) => row.order.toString())
+          .concat(alreadyIncluded.map((id) => id.toString()))
+          .join(':')
+      : undefined
```
devinivy, line 80: Another way we could approach this is for the cursor to contain a seed for the random order, then we shuffle and paginate in memory (as we currently do) based on the seed.

packages/bsky/src/api/app/bsky/actor/getSuggestions.ts
```diff
@@ -110,6 +112,24 @@ const presentation = (state: HydrationState) => {
   return { actors: suggestedActors, cursor }
 }
 
+const parseCursor = (cursor?: string): number[] => {
+  if (!cursor) {
+    return []
+  }
+  try {
+    return cursor.split(':').map((id) => parseInt(id))
```
devinivy, line 120: If someone is mucking around with the API manually, it's likely you'll get some `NaN`s out of this.

devinivy, line 120: Not super important, just avoiding support of weird number formats like hex.
```suggestion
    return cursor.split(':').map((id) => parseInt(id, 10))
```

devinivy (approved):
Nice!

packages/bsky/tests/views/suggestions.test.ts
dholms: 🙏

packages/bsky/src/api/app/bsky/actor/getSuggestions.ts
dholms: good call - i'll throw a filter in 👍

packages/bsky/src/api/app/bsky/actor/getSuggestions.ts
dholms: I actually like that quite a bit

The current approach should work fine right now, but if we end up serving a larger result here, we may want to change the approach.

I'll ship what we have now so we can get it out, but make a note in the code of the alternate approach - maybe we can hit it in v2 👌

PR #1858, opened 2023-11-15
Add temp.fetchLabels method

dholms:
This is a temporary method for fetching all labels from a labeler based on time.

This is being introduced in lieu of [`com.atproto.label.subscribeLabels`](https://github.com/bluesky-social/atproto/blob/main/lexicons/com/atproto/label/subscribeLabels.json) to avoid implemented a new streaming interface on the current appview. However, the expectation is that it will be used in a similar manner - to stay up to date with a given labeler's entire label set.

### DIFF
--- a/lexicons/com/atproto/temp/fetchLabels.json
+++ b/lexicons/com/atproto/temp/fetchLabels.json
@@ -1,0 +1,35 @@
+{
+  "lexicon": 1,
+  "id": "com.atproto.temp.fetchLabels",
+  "defs": {
+    "main": {
+      "type": "query",
+      "description": "Find labels relevant to the provided URI patterns.",
+      "parameters": {
+        "type": "params",
+        "properties": {
+          "since": { "type": "integer" },
+          "limit": {
+            "type": "integer",
+            "minimum": 1,
+            "maximum": 250,
+            "default": 50
+          }
+        }
+      },
+      "output": {
+        "encoding": "application/json",
+        "schema": {
+          "type": "object",
+          "required": ["labels"],
+          "properties": {
+            "labels": {
+              "type": "array",
+              "items": { "type": "ref", "ref": "com.atproto.label.defs#label" }
+            }
+          }
+        }
+      }
+    }
+  }
+}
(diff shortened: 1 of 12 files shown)


bnewbold:
mostly out of curiosity: why `.temp.` and not `.unspecced.` like we have used for others? or could even put under `app.bsky.unspecced` to keep `com.atproto` clean.

not a big deal either way, just curious

lexicons/com/atproto/temp/fetchLabels.json
```diff
@@ -0,0 +1,35 @@
...
+  "lexicon": 1,
+  "id": "com.atproto.temp.fetchLabels",
+  "defs": {
+    "main": {
+      "type": "query",
+      "description": "Fetch all labels from a labeler created after a certain date.",
+      "parameters": {
+        "type": "params",
+        "properties": {
+          "since": { "type": "integer" },
```
bnewbold, line 11: The response labels will have `cts` as a datetime string, so caller will need to parse those and turn them in to UNIX timestamp for next request.

I know @whyrusleeping prefers UNIX timestamps, but wanted to double-check that whether is actually more ergonomic in this case.

I did a quick grep and didn't see any use of a UNIX timestamps in the current lexicons. On the other hand, don't want to hold things up: if why is :+1: on it this way than no problem from me on a temp endpoint.

dholms:
> mostly out of curiosity: why .temp. and not .unspecced. like we have used for others? or could even put under app.bsky.unspecced to keep com.atproto clean.

I chose `com.atproto` since all the label routes are there so it just seemed most obvious to put them there. Not opposed to `app.bsky` for unspecced/temp routes like this tho

The honest answer for temp is: we have a codegen issue that prevents us from having overlapping namespaces (`com.atproto.unspecced` & `app.bsky.unspecced`). Known issue & we need to fix it, but don't have the bandwidth right now.

Post hoc rationalization - most unspecced routes gesture at something that we probably will support at some point but haven't settled on the right shape for. This route is just a stopgap measure and probably won't ever get support once we implement the actual apis - subscribeLabels & queryLabels

devinivy (approved):

lexicons/com/atproto/temp/fetchLabels.json
dholms: Checked in with why & he said UNIX timestamps are better for him.

I'll just stick with that but we can reconsider when it comes around to the official api

bnewbold:
thanks for the explanations and double-checks!

PR #1862, opened 2023-11-16
Bugfix: fetchLabels

dholms:
Two issues: 
- we weren't correctly formatting undefined cids on method output
- for large integer query params were getting messed up by an xrpc util - we were using the bitwise OR operator (`|`) instead of the logical OR operator (`||`) to recover from unparseable numbers

### DIFF
--- a/packages/xrpc-server/src/util.ts
+++ b/packages/xrpc-server/src/util.ts
@@ -60,7 +60,7 @@ export function decodeQueryParam(
   if (type === 'float') {
     return Number(String(value))
   } else if (type === 'integer') {
-    return Number(String(value)) | 0
+    return Number(String(value)) || 0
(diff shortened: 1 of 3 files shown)


packages/xrpc-server/src/util.ts
```diff
@@ -60,7 +60,7 @@ export function decodeQueryParam(
   if (type === 'float') {
     return Number(String(value))
   } else if (type === 'integer') {
-    return Number(String(value)) | 0
+    return Number(String(value)) || 0
```
devinivy, line 63: I actually think this was a cast to integer!

dholms: ah you're totally right

okay updated this to a `parseInt` instead which i think is best of both worlds

PR #1869, opened 2023-11-19
Fix role auth for access-or-role verifier, getBlob check on actor takedowns

devinivy:
Two fixes here on pds v2:
 - Ensure the access-or-role auth verifier works with non-admin roles.
 - Ensure taken down actors can fetch their own blobs, admins can fetch their blobs, and others cannot fetch their blobs.


dholms (approved):
awesome thx for hitting that 🙏

PR #1871, opened 2023-11-20
App-level retries for sqlite on pds

devinivy:
The issue with sqlite's busy timeout is that it may cause unnecessary blocking in the node process.  This moves sqlite's busy timeout handling up into the application, using the same backoff/retry approach that sqlite itself employs.

 - Run transactions with busy handling.
 - Run writes outside of transactions with busy handling.
 - Move WAL mode outside of db opening, into the db creation/migration process.  This avoids a "long" (sub-millisecond) block that occurs every time a db is opened.
 - Refactor retry helper to allow different backoff strategies including max timeouts, add tests.

Some portions of this will be easier to review [ignoring whitespace changes](https://github.com/bluesky-social/atproto/pull/1871/files?w=1).

### DIFF
--- a/packages/pds/src/actor-store/record/transactor.ts
+++ b/packages/pds/src/actor-store/record/transactor.ts
@@ -74,12 +74,6 @@ export class RecordTransactor extends RecordReader {
     log.info({ uri }, 'deleted indexed record')
   }
 
-  async deleteForActor(_did: string) {
-    // Not done in transaction because it would be too long, prone to contention.
-    // Also, this can safely be run multiple times if it fails.
-    await this.db.db.deleteFrom('record').execute()
-  }
-
   async removeBacklinksByUri(uri: AtUri) {
     await this.db.db
       .deleteFrom('backlink')
--- a/packages/pds/src/db/db.ts
+++ b/packages/pds/src/db/db.ts
@@ -1,5 +1,6 @@
 import assert from 'assert'
 import {
+  sql,
   Kysely,
   SqliteDialect,
   KyselyPlugin,
@@ -14,8 +15,7 @@ import { dbLogger } from '../logger'
 import { retrySqlite } from './util'
 
 const DEFAULT_PRAGMAS = {
-  journal_mode: 'WAL',
-  strict: 'ON',
+  strict: 'ON', // @TODO strictness should live on table defs instead
 }
 
 export class Database<Schema> {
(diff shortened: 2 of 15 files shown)


packages/pds/src/actor-store/record/transactor.ts
```diff
@@ -74,12 +74,6 @@ export class RecordTransactor extends RecordReader {
     log.info({ uri }, 'deleted indexed record')
   }
 
-  async deleteForActor(_did: string) {
-    // Not done in transaction because it would be too long, prone to contention.
-    // Also, this can safely be run multiple times if it fails.
-    await this.db.db.deleteFrom('record').execute()
-  }
-
```
devinivy, line 82: This appeared unused—I'm pretty sure we just delete the whole db now when getting rid of an actor.

packages/pds/src/db/db.ts
```diff
@@ -73,13 +76,15 @@ export class Database<Schema> {
...
-      retryable: (err) =>
-        typeof err?.['code'] === 'string' && RETRY_ERRORS.has(err['code']),
-      maxRetries: 5,
-      backoffMultiplier: 50,
-      backoffMax: 2000,
-    })
+    return retrySqlite(() => this.transactionNoRetry(fn))
+  }
+
+  async executeWithRetry<T>(query: { execute: () => Promise<T> }) {
```
dholms, line 82: any reason for this syntax instead of a call back?

not really opposed, but it may be more inline with the rest of the db code as a call back 🤔 

if we wanted to get fancy, could even pass the underlying kysely instance into the callback and avoid the `db.db` 😛

dholms: oh actually... since it doesn't include the `execute`, i can see that this pattern encourages you to not do more than one write at a time in the retry 🤔

dholms (approved):
this looks great 👍 

dig the approach & thoroughness of the retry logic

small q about the `executeWithRetry` api but I think i came around to it so no worries there 👍

packages/pds/src/db/db.ts
devinivy: I know what you mean, went back and forth on it a little.  But yeah two things here, neither of which is crucial:
 - We only want to retry individual statements, and this allows us to enforce that.
 - Small thing, but we only need to compile the query one time no matter the number of retries. Also, when the statement is already in a transaction the no-op here is extremely cheap.

dholms: nice i dig 👍

PR #1882, opened 2023-11-22
Put canReply state on post viewer state instead of thread viewer state

dholms:
This information is needed in most places that a post is encountered (namely in feeds).

Suprisingly, we already had nearly all the information hydrated that we needed to determine if an interaction was allowed. The only additional piece of information was if the user was included in a list

### DIFF
--- a/packages/bsky/src/api/app/bsky/feed/getPosts.ts
+++ b/packages/bsky/src/api/app/bsky/feed/getPosts.ts
@@ -65,7 +65,7 @@ const presentation = (state: HydrationState, ctx: Context) => {
   const actors = actorService.views.profileBasicPresentation(
     Object.keys(profiles),
     state,
-    { viewer: params.viewer },
+    params.viewer,
   )
   const postViews = postUris.flatMap((uri) => {
     const postView = feedService.views.formatPostView(
@@ -76,6 +76,7 @@ const presentation = (state: HydrationState, ctx: Context) => {
       state.embeds,
       state.labels,
       state.lists,
+      params.viewer,
     )
     return postView ?? SKIP
   })
--- a/packages/bsky/src/api/app/bsky/feed/getPosts.ts
+++ b/packages/bsky/src/api/app/bsky/feed/getPosts.ts
@@ -1,10 +1,13 @@
 import { dedupeStrs } from '@atproto/common'
-import { AtUri } from '@atproto/syntax'
 import { Server } from '../../../../lexicon'
 import { QueryParams } from '../../../../lexicon/types/app/bsky/feed/getPosts'
 import AppContext from '../../../../context'
 import { Database } from '../../../../db'
-import { FeedHydrationState, FeedService } from '../../../../services/feed'
+import {
+  FeedHydrationState,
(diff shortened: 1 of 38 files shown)


packages/bsky/src/api/app/bsky/feed/getPosts.ts
```diff
@@ -31,51 +34,51 @@ export default function (server: Server, ctx: AppContext) {
   })
 }
 
-const skeleton = async (params: Params) => {
-  return { params, postUris: dedupeStrs(params.uris) }
+const skeleton = async (params: Params, ctx: Context) => {
+  const deduped = dedupeStrs(params.uris)
+  const feedItems = await ctx.feedService.postUrisToFeedItems(deduped)
```
dholms, line 39: this does require an extra step on `getPosts` and `searchPosts` where we turn post uris into feed items first so that we can hydration information about the replyRoot along the way

packages/bsky/tests/views/threadgating.test.ts
```diff
@@ -29,6 +29,19 @@ describe('views with thread gating', () => {
     await network.close()
   })
 
+  // check that canReply state is applied correctly in a simple method like getPosts
+  const checkCanReplyForUser = async (
```
dholms, line 33: I felt that this was an economical option to ensure that `canReply` state is being applied to all views correctly

Other option is to either augment the seed client to include some threadgates or write a bunch of thorough tests on every route that hydrates posts.

`getPosts` is the simplest hydration method for posts & the same hydration logic is used for every route, so the thinking is that if it is applied correctly here, that it should be applied correctly elsewhere as well

packages/bsky/src/services/feed/views.ts
```diff
@@ -217,6 +239,65 @@ export class FeedViews {
     }
   }
 
+  userCanReply(
```
devinivy, line 242: We still have `feedutils.violatesThreadGate()` hanging around too.  I realize it might be kind of a pain to DRY these up due to the way one is tied-in with indexing logic.  Maybe for the time being we can just leave a comment so that anyone working on one will be aware of the other.

devinivy (approved):
Nice, made it look easy!

packages/bsky/tests/views/__snapshots__/author-feed.test.ts.snap
```diff
@@ -54,7 +54,9 @@ Array [
       "replyCount": 0,
       "repostCount": 1,
       "uri": "record(0)",
-      "viewer": Object {},
+      "viewer": Object {
+        "canReply": true,
+      },
```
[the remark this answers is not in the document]
devinivy: Ah I see what you mean!  A couple thoughts:
 - I think I prefer the weirdness of the default to `true` over using something phrased in the negative e.g. `cannotReply`.  In other words `canReply !== false` is a little nicer to me than `!cannotReply`, but this is probably just a matter of personal preference.
 - Only potential issue I see with `replyGated` is that we may want to leave the door open to there being other reasons the user isn't able to reply, outside of thread gates.

PR #1884, opened 2023-11-23
Support unauthed usage of feeds

devinivy:
By making auth optional on `app.bsky.feed.getFeed`, we open the door to public usage of feeds via the appview if the feed chooses to support it.


estrattonbailey (approved):
LGTM! Tested locally, works great 👍

PR #1893, opened 2023-11-27
Cleanup outdated notifications in appview, add daemon for similar tasks

devinivy:
This adds a daemon to the appview that cleans-up outdated notifications.  The daemon has its own service entrypoint.  We might consider moving the materialized view maintainer into this service as well.

The rules for removing old notifications are based on three thresholds: "days-before-last-seen", "days-before-latest-unread", and an "unread-kept-count".  Read notifications are removed when they are older than "days-before-last-seen" as compared to the last time the user saw their notifications.  Unread notifications are only removed when the user has more notifications than than the "unread-kept-count".  Once that threshold is hit, notifications are removed if they are older than "days-before-latest-unread" as compared to their most recent notification.

### DIFF
--- a/packages/bsky/src/daemon/notifications.ts
+++ b/packages/bsky/src/daemon/notifications.ts
@@ -5,21 +5,32 @@ import logger from './logger'
 export class NotificationsDaemon {
   ac = new AbortController()
   running: Promise<void> | undefined
+  count = 0
+  lastDid: string | null = null
+
   constructor(private ctx: DaemonContext) {}
 
-  run() {
+  run(opts?: RunOptions) {
     if (this.running) return
-    this.running = this.tidyNotifications()
+    this.count = 0
+    this.lastDid = null
+    this.ac = new AbortController()
+    this.running = this.tidyNotifications({
+      ...opts,
+      forever: opts?.forever !== false, // run forever by default
+    })
       .catch((err) => logger.error({ err }, 'notifications daemon failed'))
       .finally(() => (this.running = undefined))
   }
 
-  private async tidyNotifications() {
+  private async tidyNotifications(opts: RunOptions) {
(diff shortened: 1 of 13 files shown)


packages/bsky/src/daemon/notifications.ts
```diff
@@ -0,0 +1,46 @@
...
+  run(opts?: RunOptions) {
+    if (this.running) return
+    this.count = 0
+    this.lastDid = null
+    this.ac = new AbortController()
+    this.running = this.tidyNotifications({
+      ...opts,
+      forever: opts?.forever !== false, // run forever by default
+    })
+      .catch((err) => logger.error({ err }, 'notifications daemon failed'))
```
dholms, line 22: should this restart if it fails? alternately, should it crash if it fails?

seems like we could get stuck with a zombie daemon like this

dholms (approved):
sweet this all looks great to me 👍

packages/bsky/src/services/util/notification.ts
```diff
@@ -0,0 +1,71 @@
+import { sql } from 'kysely'
+import { countAll } from '../../db/util'
+import { PrimaryDatabase } from '../../db'
+
+// @TODO decide thresholds
+// i.e. 30 days before the last time the user checked their notifs
+export const BEFORE_LAST_SEEN_DAYS = 30
+// i.e. 180 days before the latest unread notification
+export const BEFORE_LATEST_UNREAD_DAYS = 30
```
devinivy, line 9:
```suggestion
// i.e. 30 days before the last time the user checked their notifs
export const BEFORE_LAST_SEEN_DAYS = 30
// i.e. 180 days before the latest unread notification
export const BEFORE_LATEST_UNREAD_DAYS = 180
```

dholms (approved):

PR #1901, opened 2023-11-28
Do not generate notifs when post violates threadgate

devinivy:
Ensures that replies that violate a threadgate do not generate notifications within the thread.

### DIFF
--- a/packages/bsky/src/services/indexing/plugins/post.ts
+++ b/packages/bsky/src/services/indexing/plugins/post.ts
@@ -113,6 +113,7 @@ const insertFn = async (
       obj.reply,
     )
     if (invalidReplyRoot || violatesThreadGate) {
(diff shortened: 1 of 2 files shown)


packages/bsky/src/services/indexing/plugins/post.ts
```diff
@@ -243,6 +244,7 @@ const notifsForInsert = (obj: IndexedPost) => {
   }
 
   for (const ancestor of obj.ancestors ?? []) {
+    if (obj.post.violatesThreadGate) continue // don't notify thread when post violates threadgate
```
dholms, line 247: this will always do a `continue` for every ancestor - not a big deal, but we could put the check before both for loops & just return earlier

dholms (approved):
yup looks good 👍 

one other thing - we may want to update the aggregate queries in this file as well so that invalid replies are not counted

dholms (approved):

PR #1905, opened 2023-11-30
:sparkles: Expose labels attached with legacy actions when events are queried

foysalit:


devinivy (approved):

packages/bsky/src/services/moderation/views.ts
```diff
@@ -155,7 +163,7 @@ export class ModerationViews {
       if (res.action === 'com.atproto.admin.defs#modEventEmail') {
         eventView.event = {
           ...eventView.event,
-          subject: res.meta?.subject ?? undefined,
+          subjectLine: res.meta?.subjectLine ?? undefined,
```
devinivy, line 166: This field is required, so perhaps we should default to `''` rather than `undefined`:
```suggestion
          subjectLine: res.meta?.subjectLine ?? '',
```

packages/bsky/src/services/moderation/views.ts
```diff
@@ -145,6 +145,33 @@ export class ModerationViews {
...
+              : [],
+          }
+        }
+
+        if (res.negateLabelVals?.length) {
+          eventView.event = {
+            ...eventView.event,
+            negateLabelVals: res.negateLabelVals?.length
+              ? res.negateLabelVals.split(' ')
+              : [],
```
devinivy, line 170:
```suggestion
            negateLabelVals: res.negateLabelVals.split(' '),
```

devinivy (approved):

PR #1906, opened 2023-11-30
Handle missing creator on lists and feed generators

dholms:
When the creator was missing (for instance in the case of a takedown), we were returning improperly formatted list & feed gen views. Now if the creator is missing, we just do not return the relevant view


devinivy (approved):
💎

PR #1913, opened 2023-12-01
Add flag for running db migrations on appview

dholms:
In prod/staging, we run migrations manually, but in other environments (for instance sandbox) we run them automatically

### DIFF
--- a/services/bsky/api.js
+++ b/services/bsky/api.js
@@ -33,7 +33,7 @@ const main = async () => {
   const env = getEnv()
   assert(env.dbPrimaryPostgresUrl, 'missing configuration for db')
 
-  if(env.enableMigrations) {
+  if (env.enableMigrations) {
     // separate db needed for more permissions
     const migrateDb = new PrimaryDatabase({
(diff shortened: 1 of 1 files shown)


services/bsky/api.js
```diff
@@ -32,6 +32,18 @@ const {
 const main = async () => {
   const env = getEnv()
   assert(env.dbPrimaryPostgresUrl, 'missing configuration for db')
+
+  if(env.enableMigrations) {
```
devinivy, line 36:
```suggestion
  if (env.enableMigrations) {
```

dholms: 💩

services/bsky/api.js
```diff
@@ -102,12 +114,12 @@ const main = async () => {
     algos,
   })
   // separate db needed for more permissions
-  const migrateDb = new PrimaryDatabase({
+  const viewMaintainerDb = new PrimaryDatabase({
```
devinivy, line 117: There's a reference to `migrateDb` down in the teardown.

devinivy (approved):

PR #1920, opened 2023-12-05
Fix dev-env images

dholms:
Scopes were screwy for the dev-env image server so it wasn't returning images


estrattonbailey (approved):

dholms:
Wait nvm it was fine how it was 😅

PR #1927, opened 2023-12-05
:sparkles: Create a mod event when automod applies labels to a record

foysalit:


packages/bsky/tests/auto-moderator/labeler.test.ts
```diff
@@ -94,6 +100,28 @@ describe('labeler', () => {
...
+      action: 'com.atproto.admin.defs#modEventLabel',
+      subjectUri: uri.toString(),
+      createLabelVals: 'test-label',
+      negateLabelVals: null,
+      comment: `[AutoModerator]: Applying labels`,
+      createdBy: labelerDid,
+    })
+
+    // Cleanup the temporary assignment, knowing that by default, moderation service is not available
+    autoMod.services.moderation = undefined
```
foysalit, line 124: feels a tiny bit hacky, maybe doing this through before and after hook would look nicer? lmk if anyone feels strongly about this.

devinivy (approved):

bnewbold (approved):
nice!

PR #1942, opened 2023-12-08
refactor(crypto): remove circular dependency

matthieusieben:
in `crypto`, there is a circular dependency between `did.ts` > `plugins.ts` > `p256/plugin.ts` > `p256/operations.ts` > `did.ts`. This PR fixes that by extracting common code into a `utils.ts` file, and uses that module instead of `did.ts` from plugin operations.


packages/crypto/src/secp256k1/keypair.ts
```diff
@@ -43,7 +42,7 @@ export class Secp256k1Keypair implements Keypair {
   }
 
   publicKeyStr(encoding: SupportedEncodings = 'base64pad'): string {
-    return uint8arrays.toString(this.publicKey, encoding)
+    return toString(this.publicKey, encoding)
```
dholms, line 45: is there a reason to import this as just `toString`? if so I think it'd be nice to import it as something like `ui8ToString` or similar. otherwise `toString` is pretty generic & gotta check imports to see what it's coming from

dholms (approved):
just the one nit, otherwise this looks good 👍

matthieusieben:
Done. Also imported these changes in https://github.com/bluesky-social/atproto/pull/2169.

matthieusieben:
Merged through https://github.com/bluesky-social/atproto/pull/2169

PR #1943, opened 2023-12-08
Move handlebar templates in own directory

matthieusieben:


matthieusieben:
#1941 is very hard to fix because of the circular dependencies. This PR was an attempt to extract parts of this rework into smaller PR but I realize now that there is no build performed on the CI, which make this change pointless.

PR #1958, opened 2023-12-12
🚧 OAuth2 - Authorization Server

matthieusieben:
This branch contains the implementation of [OAuth for Atproto](https://github.com/bluesky-social/proposals/tree/main/0004-oauth).

Is is based on the [build-rework](https://github.com/matthieusieben/atproto/tree/build-rework) branch (though I can't reflect that in this PR since the `build-rework` branch does not exist in this repo). First relevant commit is [feat(oauth-provider): initial implementation](https://github.com/bluesky-social/atproto/pull/1958/commits/6870ca8ec95b63057f15e83f39bec4360e8f40a4).


packages/syntax/src/did.ts
```diff
@@ -2,7 +2,7 @@
 //   - valid W3C DID (https://www.w3.org/TR/did-core/#did-syntax)
 //      - entire URI is ASCII: [a-zA-Z0-9._:%-]
 //      - always starts "did:" (lower-case)
-//      - method name is one or more lower-case letters, followed by ":"
+//      - method name is one or more lower-case letter or digit, followed by ":"
```
bnewbold, line 5: Good catch!

Allowing digits does seem to be correct upstream:
https://www.w3.org/TR/did-core/#did-syntax

and there are methods like `did:3` and `did:web7` which use them:
https://w3c.github.io/did-spec-registries/#did-methods

But I don't think we should relax this without updating the atproto spec; coordinating with the other major implementations (in particular, golang); updating the shared interop test vectors to cover this syntax; and probably some small dev comms. At a minimum we should do this change here in a separate PR. Current spec:
https://atproto.com/specs/did#at-protocol-did-identifier-syntax

Feels like a bit of a bummer that they allow this as it visually blurs with `host:port` style syntax, but I guess the required second colon disambiguates.

bnewbold: created a tracking issue: https://github.com/bluesky-social/atproto-website/issues/292

matthieusieben: I'm reworking this branch. I'll drop this particular change from this PR !

matthieusieben:
Move to a branch in this repo. PR here https://github.com/bluesky-social/atproto/pull/2407

PR #1959, opened 2023-12-12
Add seenAt time to listNotifications

dholms:
Adds the optional field `seenAt` to `listNotifications` which allows a service to inform the client of their currently tracked last seen time.


devinivy (approved):
💎

PR #1960, opened 2023-12-12
Report separate error on getProfile for account suspension

dholms:
When accessing a takendown profile, check if the takedown is a temporary suspension or permanent

### DIFF
--- a/packages/bsky/src/services/moderation/index.ts
+++ b/packages/bsky/src/services/moderation/index.ts
@@ -327,6 +327,7 @@ export class ModerationService {
     const res = await this.db.db
       .selectFrom('moderation_subject_status')
       .where('did', '=', did)
+      .where('recordPath', '=', '')
       .where('suspendUntil', '>', new Date().toISOString())
       .select('did')
       .limit(1)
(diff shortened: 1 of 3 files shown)


packages/bsky/src/services/moderation/index.ts
```diff
@@ -323,6 +323,17 @@ export class ModerationService {
     return subjectsDueForReversal
   }
 
+  async isSubjectSuspended(did: string): Promise<boolean> {
+    const res = await this.db.db
+      .selectFrom('moderation_subject_status')
+      .where('did', '=', did)
+      .where('suspendUntil', '>', new Date().toISOString())
```
devinivy, line 330: Could be nice to ensure we're talking about a suspension on an account takedown and not a record (despite that not being something that typically happens).  Would also allow you to drop the `limit(1)` since `(did, recordPath)` is unique.
```suggestion
      .where('did', '=', did)
      .where('recordPath', '=', '')
      .where('suspendUntil', '>', new Date().toISOString())
```

devinivy (approved):

packages/bsky/src/services/moderation/index.ts
dholms: ah cool good call - wasn't totally familiar with the new table schema 👌

PR #1969, opened 2023-12-14
:sparkles: Allow appealing a moderator decision through special report type

foysalit:
This PR implements a very basic mechanism for allowing users to appeal a moderator decision, leveraging existing lexicons and infrastructure. The expected user flow, through this feature is

1. User creates a report of special kind (reasonType: 'reasonAppeal') on their own content (or a moderator creates that report using ozone, on behalf of the author) which sets a boolean flag `appealed` to `true` on the moderation_subject_status row
2. The appealed column will only be set to false when a moderator emits `modEventResolveAppeal`
3. Within ozone, moderators can filter the queue to easily find subjects that have been appealed


lexicons/com/atproto/admin/defs.json
```diff
@@ -167,9 +168,18 @@
           "type": "string",
           "format": "datetime"
         },
+        "lastAppealedAt": {
+          "type": "string",
+          "format": "datetime",
+          "description": "Timestamp referencing when the owner of the subject appealed a moderation action"
```
bnewbold, line 174: It is a small nit (could ignore this comment), but "owner" feels subtly weird to me here. We use "author" just below, that feels better. Or "creator".

packages/bsky/src/services/moderation/status.ts
```diff
@@ -118,6 +127,11 @@ export const adjustModerationSubjectStatus = async (
     return null
   }
 
+  // Exit early if someone is trying to game the system by appealing someone else's content
+  if (isAppealEvent && createdBy !== subjectDid) {
+    return null
+  }
+
```
bnewbold, line 134: This is the part i'm still least sure of myself, even after settling on this behavior after earlier discussion. It feels a little weird to special case like this (and re-write the createdBy elsewhere).

I think it's probably fine though, and changing the behavior probably won't be a change to Lexicons so we aren't particularly locked-in to these behaviors ("who can create appeal reports").

bnewbold (commented):
(didn't code-review, just lexicon and behavior review)

Having some last-minute indecision about the "who can create an appeal for who" behavior, but I think how this PR implements it is fine, I don't have a specific alternative to propose, and we can probably change our minds later.

lexicons/com/atproto/admin/defs.json
```diff
@@ -167,9 +168,18 @@
...
+        },
         "takendown": {
           "type": "boolean"
         },
+        "appealed": {
+          "type": "boolean",
+          "description": "True Indicates that the a previously taken moderator action was appealed against, by the author of the content. False indicates last appeal was resolved by moderators."
```
devinivy, line 181:
```suggestion
          "description": "True indicates that the a previously taken moderator action was appealed against, by the author of the content. False indicates last appeal was resolved by moderators."
```

lexicons/com/atproto/admin/queryModerationStatuses.json
```diff
@@ -64,6 +64,10 @@
             "type": "boolean",
             "description": "Get subjects that were taken down"
           },
+          "appealed": {
```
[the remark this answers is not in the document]
devinivy: This does mirror the name of the new field on `subjectStatusView`.  Do you think it should be updated there too?

packages/bsky/src/services/moderation/status.ts
devinivy: Agree, I think we can iterate on it.  We might be able to push it up into authorization layer then forget about it here.

lexicons/com/atproto/admin/queryModerationStatuses.json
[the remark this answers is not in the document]
foysalit: yh we also use terms like `takendown` for boolean flag fields so at least it's consistent. happy to rethink naming in a future pass but I think I'd want to move forward with this as is.

PR #1970, opened 2023-12-15
Split out moderation backend

dholms:
This pulls the moderation backend out of the appview & into it's own service.

Mod service:
- Basically a copy of the moderation service from within the appview
- a few basic implementation changes:
  - refactored methods to use `Map` instead of `Record` (which we did recently on the appview & is quite nice)
  - Added some new classes for wrangling the all of the "subjectType" typeswitches we were doing
  - run event reversal in a separate daemon
- larger architectural changes:
    - Goes out to the appview to hydrate in records (including profiles) and handles
    - We treat takedowns as "push events". We transactionally store them in push event tables & then the daemon routinely polls & attempts to push those out to downstream services (appview/PDS). It retries up to 10 times
    - For push events, the mod service uses a service jwt signed by it's own key (which is announced in it's did doc)

Appview
- removed all existing admin routes
- added simple routes for getting account info & getting/updating subject status (such as takedowns)
- added a new service jwt auth method for an appointed admin service. This functions like user auth to the appview but it validates that the issuer is a trusted DID
- auto-mod pushes all events out to ozone backend instead of tracking locally
  - cdn invalidation is no longer done by auto-mod, but is handled by the appview app instances when the takedown request comes through from ozone

PDS
- We previously got in some of the admin service auth stuff into the PDS, but there was a strange semantic where the audience was set to the DID of the account that the action was about. This is updated so that it uses the PDS's DID
- Proxies createReport to mod service now (instead of appview) & gives the PDS a config var for a trusted mod service & URL

Random other things that got wrapped up into this: 
- auth verifier in appview (refactored the api to look like the PDS's & added in an "admin service" verifier
  - refactored some of the auth checks for viewing takedowns in application routes (like `getProfile`) to use a new `parseCreds` util
- moved appview test seeds to dev-env (for re-use in ozone)
  - it would be great to reuse the seeds in PDS as well, but there are some subtle differences between them that I didn't want to sort through righ tnow


packages/bsky/src/api/app/bsky/actor/getProfile.ts
```diff
@@ -52,36 +50,29 @@ const skeleton = async (
...
-        'AccountTakedown',
-      )
-    } else {
-      throw new InvalidRequestError(
-        'Account has been taken down',
-        'AccountTakedown',
-      )
-    }
+  if (!canViewTakedowns && softDeleted(actor)) {
+    // @TODO throw a different error if the accoutn is suspended
```
dholms, line 60: we don't have the info in the appview for this with the current architecture anymore - would be easier to do with labels

lexicons/com/atproto/admin/getAccountInfos.json
```diff
@@ -0,0 +1,36 @@
+{
+  "lexicon": 1,
+  "id": "com.atproto.admin.getAccountInfos",
```
dholms, line 3: this is a batch version of `getAccountInfo` (implemented on PDS) - think there's a decent argument we just deprecate the non-batch version

The basic idea here is to go to all relevant services (pds & appview) for all the info they have on a user & then collate in the ozone backend

packages/bsky/src/db/migrations/20231220T225126090Z-blob-takedowns.ts
```diff
@@ -0,0 +1,61 @@
...
+    .dropConstraint('actor_takedown_id_fkey')
+    .execute()
+  await db.schema
+    .alterTable('record')
+    .dropConstraint('record_takedown_id_fkey')
+    .execute()
+  await db.schema
+    .alterTable('actor')
+    .alterColumn('takedownId')
+    .setDataType('varchar')
```
dholms, line 22: may require a bit of care doing this in prod - but i'm pretty sure we can do the int -> string cast pretty straightforwardly

packages/bsky/src/db/tables/blob-takedown.ts
```diff
@@ -0,0 +1,9 @@
+export interface BlobTakedown {
```
dholms, line 1: I don't think we're tracking blobs anywhere else in the appview, so needed a new table for takedown state

packages/bsky/tests/auto-moderator/takedowns.test.ts
```diff
@@ -36,35 +37,36 @@ describe('takedowner', () => {
...
+    autoMod = network.bsky.indexer.ctx.autoMod
     autoMod.imageFlagger = new TestFlagger()
     pdsAgent = new AtpAgent({ service: network.pds.url })
     sc = network.getSeedClient()
     await usersSeed(sc)
     await network.processAll()
     alice = sc.dids.alice
     const fileBytes1 = await fs.readFile(
-      'tests/sample-img/key-portrait-small.jpg',
+      '../dev-env/src/seed/img/key-portrait-small.jpg',
```
dholms, line 50: these are pretty ugly but i'm not sure a better way to do it...

packages/ozone/src/api/admin/searchRepos.ts
```diff
@@ -0,0 +1,42 @@
...
+        const found = repos.get(query)
+        return {
+          encoding: 'application/json',
+          body: {
+            repos: found ? [found] : [],
+          },
+        }
+      }
+
+      const res = await ctx.appviewAgent.api.app.bsky.actor.searchActors(params)
```
dholms, line 26: since we don't maintain an index of all actors in ozone, i just fell back to the search actors route on the appview (but preserved the special case for dids)

packages/bsky/src/ingester/label-subscription.ts
```diff
@@ -0,0 +1,73 @@
+import AtpAgent from '@atproto/api'
+import { PrimaryDatabase } from '../db'
+import { sql } from 'kysely'
+import { dbLogger } from '../logger'
+import { SECOND } from '@atproto/common'
+
+export class LabelSubscription {
```
dholms, line 7: Added this as part of the appview ingester. You can imagine it consuming the `subscribeLabels` stream, but since we don't have that yet, it "subscribes" to labels by repeatedly calling `temp.fetchLabels`

packages/ozone/src/daemon/event-pusher.ts
```diff
@@ -0,0 +1,295 @@
...
+
+  async pushRepoEvents() {
+    return await this.db.transaction(async (dbTxn) => {
+      const toPush = await dbTxn.db
+        .selectFrom('repo_push_event')
+        .selectAll()
+        .forUpdate()
+        .skipLocked()
+        .where('confirmedAt', 'is', null)
+        .where('attempts', '<', 10)
```
dholms, line 137: The event pusher will retry each push 10 times and then give up

We may want to add smarter logic here to handle the case of downtime on the service it's pushing too

packages/ozone/src/daemon/index.ts
```diff
@@ -0,0 +1,71 @@
...
+import { EventPusher } from './event-pusher'
+import { EventReverser } from './event-reverser'
+import { ModerationService } from '../mod-service'
+import { Keypair } from '@atproto/crypto'
+import { createServiceAuthHeaders } from '@atproto/xrpc-server'
+
+export { EventPusher } from './event-pusher'
+export { EventReverser } from './event-reverser'
+
+export class OzoneDaemon {
```
dholms, line 14: Daemon can run in-process with the application server (if you only have one app server) or as a separate daemon

packages/ozone/src/mod-service/subject.ts
```diff
@@ -0,0 +1,139 @@
...
+  did: string
+  recordPath: string | undefined
+  blobCids?: string[]
+  isRepo(): this is RepoSubject
+  isRecord(): this is RecordSubject
+  info(): SubjectInfo
+  lex(): RepoRef | StrongRef
+}
+
+export class RepoSubject implements ModSubject {
```
dholms, line 76: Added these new subject classes to wrangle all the typeswitching around subjects that we were doing on string types

packages/ozone/src/util.ts
```diff
@@ -0,0 +1,26 @@
+import { AxiosError } from 'axios'
```
dholms, line 1: untanglining these dependencies is a big enough pain that i just copied this over

I'd really like a generic place for "service utils"

packages/ozone/src/db/schema/repo_push_event.ts
```diff
@@ -0,0 +1,18 @@
+import { Generated } from 'kysely'
+
+export const eventTableName = 'repo_push_event'
+
+export type RepoPushEventType = 'takedown'
+
+export interface RepoPushEvent {
```
dholms, line 7: I opted to use separate tables here so that we can enforce constraints about the data we need (like uri in the case of a record)

I've gone back and forth on where I should capture the fact that a takedown gets fanned out to multiple services. ie:
- fanout on write to push_event table with different event types: `pds_takedown` & `appview_takedown`
- or fanout on read from push_event table

I implemented the former mostly for simplicities sake, but am now leaning towards the latter for better error tracking - any thoughts?

packages/ozone/src/mod-service/index.ts
```diff
@@ -0,0 +1,637 @@
...
+      } else if (subject.isRecord()) {
+        await this.reverseTakedownRecord(subject)
+      }
+    }
+
+    return result
+  }
+
+  async takedownRepo(subject: RepoSubject, takedownId: number) {
+    await this.db.db
```
dholms, line 334: we transactionally add the repo to the push_event table and then actually push it out & mark it as completed in the daemon

I'm considering attempting a push in a commit hook here, but not sure if it's worth the hassle. I kinda like the simplicity of the actual push living in just one place 🤔

packages/bsky/src/api/app/bsky/actor/getProfile.ts
devinivy: One idea is to smuggle it into the takedown reference: e.g. something like `5917` vs. `5917-SUSPEND`.

packages/bsky/src/db/migrations/20231220T225126090Z-blob-takedowns.ts
devinivy: Do we prefer it over adding a new column `takedownRef`?  That was pretty smooth last time and has the side benefit of being consistent with the pds.

packages/pds/src/config/config.ts
```diff
@@ -175,10 +175,20 @@ export const envToCfg = (env: ServerEnvironment): ServerConfig => {
...
     did: env.bskyAppViewDid,
-    proxyModeration: env.bskyAppViewModeration ?? false,
     cdnUrlPattern: env.bskyAppViewCdnUrlPattern,
   }
 
+  if (!env.modServiceUrl) {
+    throw new Error('Must configure PDS_MOD_SERVICE_URL')
+  } else if (!env.modServiceDid) {
+    throw new Error('Must configure PDS_MOD_SERVICE_DID')
+  }
```
devinivy, line 185: Really doesn't matter a ton, but I find that `assert()` is convenient for this kind of thing.

packages/ozone/build.js
```diff
@@ -0,0 +1,19 @@
...
+  logLevel: 'info',
+  entryPoints: ['src/index.ts', 'src/db/index.ts'],
+  bundle: true,
+  sourcemap: true,
+  outdir: 'dist',
+  platform: 'node',
+  external: [
+    // Referenced in pg driver, but optional and we don't use it
+    'pg-native',
+    'sharp',
```
devinivy, line 16: I expect we may be able to nix sharp here.

packages/bsky/src/api/app/bsky/actor/getProfile.ts
dholms: yeah i thought of that... kinda hacky but i suppose it works 🤔

packages/bsky/src/db/migrations/20231220T225126090Z-blob-takedowns.ts
dholms: ah right i forgot about that approach... i'm game 👍

packages/xrpc/src/client.ts
```diff
@@ -115,6 +115,9 @@ export class ServiceClient {
         this.baseClient.lex.assertValidXrpcOutput(methodNsid, res.body)
       } catch (e: any) {
         if (e instanceof ValidationError) {
+          console.log(methodNsid)
+          console.log(res.body)
+          console.log('E: ', e)
```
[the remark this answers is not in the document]
dholms: 🙏

packages/dev-env/src/pds.ts
```diff
@@ -47,6 +47,8 @@ export class TestPds {
       bskyAppViewUrl: 'https://appview.invalid',
       bskyAppViewDid: 'did:example:invalid',
       bskyAppViewCdnUrlPattern: 'http://cdn.appview.com/%s/%s/%s',
+      modServiceUrl: 'https:/moderator.invalid',
```
devinivy, line 50:
```suggestion
      modServiceUrl: 'https://moderator.invalid',
```

packages/bsky/tests/views/profile.test.ts
```diff
@@ -207,61 +207,15 @@ describe('pds profile views', () => {
...
-        reason: 'Y',
-      },
-      {
-        encoding: 'application/json',
-        headers: network.pds.adminAuthHeaders(),
-      },
-    )
-  })
-
-  it('blocked by actor suspension', async () => {
```
devinivy, line 227: If we fix the suspension functionality using the ref convention, this is just a placeholder to bring this test back.

packages/bsky/src/api/app/bsky/actor/getProfile.ts
devinivy: If it's not all that important to be able to provide this error message then maybe we drop it.  But if it's important, might be worth the hack?

packages/bsky/src/api/app/bsky/actor/getProfiles.ts
```diff
@@ -13,11 +13,11 @@ import { createPipeline, noRules } from '../../../../pipeline'
 export default function (server: Server, ctx: AppContext) {
   const getProfile = createPipeline(skeleton, hydration, noRules, presentation)
   server.app.bsky.actor.getProfiles({
-    auth: ctx.authOptionalVerifier,
+    auth: ctx.authVerifier.standardOptional,
     handler: async ({ auth, params, res }) => {
       const db = ctx.db.getReplica()
       const actorService = ctx.services.actor(db)
-      const viewer = auth.credentials.did
+      const viewer = auth.credentials.iss
```
devinivy, line 20: Wondering in the back of my head if we should be looking at `sub` 🤔 I suppose it's known/implied that for this kind of credential `iss` and `sub` are pretty much the same, but I could imagine an implementation looking to set `iss` to the source pds.

devinivy: In any case, it's not super specific to this work.  We can reconsider later, just didn't want to lose the thought.

packages/bsky/src/api/blob-resolver.ts
```diff
@@ -85,16 +85,22 @@ export async function resolveBlob(
...
       .select('id')
       .where('blobCids', '@>', sql`CAST(${JSON.stringify([cidStr])} AS JSONB)`)
       .where('takendown', 'is', true)
       .executeTakeFirst(),
+    db.db
+      .selectFrom('blob_takedown')
+      .select('takedownId')
+      .where('did', '=', did)
+      .where('cid', '=', cid.toString())
+      .executeTakeFirst(),
```
devinivy, line 101: I imagine we'll eventually get rid of one of these checks.

packages/bsky/src/api/com/atproto/moderation/createReport.ts
```diff
@@ -1,42 +0,0 @@
-import { AuthRequiredError } from '@atproto/xrpc-server'
```
devinivy, line 1: If it's useful to the rollout, we could pass requests to here through to automod for a short time period.

packages/bsky/src/db/tables/actor.ts
```diff
@@ -2,7 +2,7 @@ export interface Actor {
   did: string
   handle: string | null
   indexedAt: string
-  takedownId: number | null // @TODO(bsky)
+  takedownId: string | null // @TODO(bsky)
```
devinivy, line 5: This TODO was way back from the appview split, I don't think it's relevant anymore.
```suggestion
  takedownId: string | null
```

packages/ozone/src/api/admin/emitModerationEvent.ts
```diff
@@ -0,0 +1,129 @@
...
+          }
+        }
+
+        if (subject.isRecord()) {
+          if (isTakedownEvent) {
+            await moderationTxn.takedownRecord(subject, result.id)
+          } else if (isReverseTakedownEvent) {
+            await moderationTxn.reverseTakedownRecord(subject)
+          }
+        }
```
devinivy, line 90: _Clean_

packages/pds/src/config/config.ts
```diff
@@ -175,10 +175,20 @@ export const envToCfg = (env: ServerEnvironment): ServerConfig => {
...
+
+  const modServiceCfg: ServerConfig['modService'] = {
+    url: env.modServiceUrl,
+    did: env.modServiceDid,
+  }
```
devinivy, line 190: Ahhh, I bet we'll have to get this going on the entryway.  Maybe time to properly land that work...

packages/ozone/src/index.ts
```diff
@@ -0,0 +1,160 @@
...
+  }
+
+  static create(opts: {
+    db: Database
+    config: ServerConfig
+    signingKey: Keypair
+  }): OzoneService {
+    const { db, config, signingKey } = opts
+    const app = express()
+    app.set('trust proxy', true)
```
devinivy, line 49: ✅

packages/ozone/src/index.ts
```diff
@@ -0,0 +1,160 @@
...
+    await events.once(server, 'listening')
+    const { port } = server.address() as AddressInfo
+    this.ctx.cfg.assignPort(port)
+    return server
+  }
+
+  async destroy(): Promise<void> {
+    await this.terminator?.terminate()
+    await this.ctx.backgroundQueue.destroy()
+    await this.ctx.daemon?.destroy()
```
devinivy, line 154: All for separating the daemon out as you mentioned 👍

packages/ozone/src/daemon/event-pusher.ts
```diff
@@ -0,0 +1,295 @@
...
+
+  async pushBlobEvents() {
+    return await this.db.transaction(async (dbTxn) => {
+      const toPush = await dbTxn.db
+        .selectFrom('blob_push_event')
+        .selectAll()
+        .forUpdate()
+        .skipLocked()
+        .where('confirmedAt', 'is', null)
+        .execute()
```
devinivy, line 170: Two questions here:
 - there isn't an index supporting the `confirmedAt` checks, I don't think.  Maybe we need one as this table grows in size?
 - Is there intended to be a check on `attempts` here?

packages/ozone/src/config.ts
```diff
@@ -0,0 +1,157 @@
...
+    const pdsDid = process.env.PDS_DID
+    const dbPostgresUrl =
+      overrides?.dbPostgresUrl || process.env.DB_POSTGRES_URL
+    assert(dbPostgresUrl)
+    const dbPostgresSchema = process.env.DB_POSTGRES_SCHEMA
+    const didPlcUrl = process.env.DID_PLC_URL || 'http://localhost:2582'
+    const adminPassword = process.env.ADMIN_PASSWORD || 'admin'
+    const moderatorPassword = process.env.MODERATOR_PASSWORD || undefined
+    const triagePassword = process.env.TRIAGE_PASSWORD || undefined
+    const labelerDid = process.env.LABELER_DID || 'did:example:labeler'
```
devinivy, line 46: I wonder if we should use the convention of prefixing the env vars with the service name `OZONE_` as we've done on the pds (and recently on the appview).

packages/ozone/src/api/moderation/createReport.ts
```diff
@@ -0,0 +1,34 @@
+import { Server } from '../../lexicon'
+import AppContext from '../../context'
+import { getReasonType } from './util'
+import { subjectFromInput } from '../../mod-service/subject'
+
+export default function (server: Server, ctx: AppContext) {
+  server.com.atproto.moderation.createReport({
+    // @TODO anonymous reports w/ optional auth are a temporary measure
+    auth: ctx.authOptionalAccessOrRoleVerifier,
```
devinivy, line 9: I don't think we _necessarily_ need to support role-based auth here since admins could go straight to `emitModerationEvent()`.  In #1969 this is actually utilized a bit.

packages/ozone/src/daemon/config.ts
```diff
@@ -0,0 +1,88 @@
...
+    const dbPostgresSchema =
+      overrides?.dbPostgresSchema || process.env.DB_POSTGRES_SCHEMA
+    assert(dbPostgresUrl)
+    const serverDid = overrides?.serverDid || process.env.SERVER_DID
+    assert(serverDid)
+    const appviewUrl = overrides?.appviewUrl || process.env.APPVIEW_URL
+    assert(appviewUrl)
+    const appviewDid = process.env.APPVIEW_DID
+    const pdsUrl = process.env.PDS_URL
+    const pdsDid = process.env.PDS_DID
```
devinivy, line 30: Similar here, we might consider prefixing these with `OZONE_` for consistency with pds and recent appview work.

packages/ozone/src/daemon/event-pusher.ts
devinivy: I suppose as an escape hatch we could always reset the attempts on events that failed during an outage.

packages/pds/src/config/config.ts
dholms: nice yup 👍

packages/ozone/src/db/schema/repo_push_event.ts
devinivy: It looks like in the current state it fans out on read.  To be honest I lean a little bit towards fanout on write.  I have two reasons for that:
 - pushes represent a single push to a single service.  That strikes me as being simpler and allows them to fail independently, which gives us strictly more information.
 - it's more flexible, will be more natural to adapt if we need to start taking different actions on pds versus appview.  Or maybe more likely, take an action on one but not the other.

packages/ozone/build.js
dholms: good catch 🙏

packages/bsky/src/api/blob-resolver.ts
dholms: uhh yes good call - actually i should just nix it now 👍 

I think we'll want to keep both checks in the intermediate branch

packages/bsky/src/api/com/atproto/moderation/createReport.ts
dholms: hmmm yup yup good point 👍

packages/pds/src/config/config.ts
dholms: yeahh we should check in on how exactly that'll function

packages/ozone/src/config.ts
dholms: yup agreed - i'll edit all of those 👍

dholms: ended up refactoring all the config to look more similar to the PDS

packages/ozone/src/daemon/event-pusher.ts
dholms: yeah that was basically my plan in the interim 👍

packages/ozone/src/db/schema/repo_push_event.ts
dholms: kk that was my feeling as well - i switched it over 👍

packages/bsky/src/api/app/bsky/actor/getProfile.ts
dholms: alrighty did the hack - I ended up adding a bit more info to the refs as well so they're not just opaque numbers. now they look like `BSKY-TAKEDOWN-5917` or `BSKY_SUSPEND-5917`. makes it look less hacky 😛

packages/ozone/src/daemon/event-pusher.ts
dholms: good call on the index 👍 

and yup got that attempts check on the repo event poller but slipped by on the other too

packages/bsky/tests/views/profile.test.ts
dholms: done 👍 & added test back in

PR #1971, opened 2023-12-15
Optimize timeline for limit=1

devinivy:
On `getTimeline`, the `limit=1` case is used commonly to check if there are new items at the top of the timeline. Since it's so common, this is a shot at optimizing it.  The most common strategy that postgres takes to build a timeline is to grab all recent content from each of the user's follow, then paginate it. The downside here is that it requires grabbing all recent content from all follows, even if you only want a single result.  The approach here instead takes the single most recent post from each of the user's follows, then sorts only those and takes the top item.


dholms (approved):
excellent 👌

PR #1974, opened 2023-12-16
Fix app.bsky.graph.getSuggestedFollowsByActor lexicon output

MarshalX:
Hi! Python SDK fails on the profile view model because the server returns the profile view **detailed** model instead. including banner, followers count, etc. More info: https://github.com/MarshalX/atproto/issues/205

so either: 
1. lexicon is wrong
2. the server returns the wrong model

thanks

### DIFF
--- a/lexicons/app/bsky/graph/getSuggestedFollowsByActor.json
+++ b/lexicons/app/bsky/graph/getSuggestedFollowsByActor.json
@@ -22,7 +22,7 @@
               "type": "array",
               "items": {
                 "type": "ref",
-                "ref": "app.bsky.actor.defs#profileView"
+                "ref": "app.bsky.actor.defs#profileViewDetailed"
               }
             }
           }


devinivy:
Thanks for surfacing!  I am checking with the crew to determine which one was intended here.  Part of the reason this wasn't flagged earlier is because the `#profileViewDetailed` is currently lexicon-compatible with `#profileView`, so they are considered valid responses.  The reason for that is that they have the same required fields, and lexicon validation provides that unknown fields should be ignored.  Either way, should be able to resolve this soon!

MarshalX:
thank you! yeah, my SDK now also thinks that this is valid. moved to your flow of forward and backward compatibility of the lexicon. so, it means i will not report something simular anymore xd

devinivy:
@MarshalX turns out this was indeed intended to be a `#profileView`.  So we'll be removing the unspecified fields from this response that exist only on the `#profileViewDetailed`.  For that reason I'm going to close the PR 👍 . Thanks again for surfacing!

PR #1992, opened 2023-12-26
Support http2 client for appview dataplane

devinivy:
Support using http2 to talk to the dataplane.  Also supports an option to ignore tls for testing and local usage.


dholms (approved):
looks good 👍

PR #1993, opened 2023-12-27
Support multiple dataplane urls for appview

devinivy:
This allows this appview to support multiple dataplane urls.  When a response indicates that one dataplane server is unavailable, it immediately retries on another one.


dholms (approved):
sweet looks great 👍

PR #1994, opened 2023-12-28
lexicons: more string limits

bnewbold:
The most sensitive thing here is putting a (very long!) length limit on alt text. I believe this is long enough to not cause accessibility issues, but can update if needed.

The motivation is that an issue with very long embed image descriptions came up that has been causing records to be discarded by golang code. We'll separately work around that by bumping limits in the CBOR library (cborgen), but I think communicating some reasonable limit on these fields is helpful. This isn't intended to be antagonistic; the current problematic records seem to have huge values *by mistake* not intentionally.

While I was at it I did an informal review of all the "string" fields in records (I might have missed a couple, this was an informal scan).

These are technically violations of Lexicon stability, so need a bunch of review and buy-in on these. I'm not strident about these changes, but I think they are worth re-visiting, and this is probably our last chance to push them through.

- I think the URL fields being format=uri would be uncontroversial
- The embed description limits will invalidate a small handful of real-world records
- I don't know how many real-world records the alt-text limit might cause. I expect very few but we might want to verify before merging that.

The specific values are somewhat arbitrary. I chose "large" ones to minimize the change in behavior (given that there are no limits right now).

Haven't run codegen, would do that just before merging.

cc: @whyrusleeping w/r/t embed description length

### DIFF
--- a/.changeset/odd-kids-warn.md
+++ b/.changeset/odd-kids-warn.md
@@ -1,0 +1,5 @@
+---
+'@atproto/api': patch
+---
+
+additional app.bsky.feed.post Lexicon string format limits
--- a/lexicons/app/bsky/embed/external.json
+++ b/lexicons/app/bsky/embed/external.json
@@ -18,8 +18,16 @@
       "required": ["uri", "title", "description"],
       "properties": {
         "uri": { "type": "string", "format": "uri" },
-        "title": { "type": "string" },
-        "description": { "type": "string" },
+        "title": {
+          "type": "string",
+          "maxGraphemes": 300,
+          "maxLength": 3000
+        },
+        "description": {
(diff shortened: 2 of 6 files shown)


dholms (approved):
These all make sense to me 👍 

I can't imagine any of these limits being hit in normal use

Let's get @pfrazee's opinion before merging tho

pfrazee:
Runs the risk of invalidating some existing records but I dont know how many without, you know, scanning the whole network. I'd lean towards yoloing it.

pfrazee:
The limits look fine to me btw

bnewbold:
Updated PR with codegen and changeset. My current disposition is to merge, and some small dev comms.

PR #2010, opened 2024-01-02
Improve handling of `resumeSession` failures

estrattonbailey:
**Motivation:** flaky networks currently result in the app nuking JWTs, effectively logging the user out. We want to log out the user if the session is invalid, but for other unknowns, we can keep it around on-device until we get a legit response from our server that says otherwise.

This PR adds an additional `AtpSessionEvent` event, `network-error`, and emits that event from `resumeSession` if we catch an unexpected exception.

### DIFF
--- a/packages/api/src/agent.ts
+++ b/packages/api/src/agent.ts
@@ -175,7 +175,7 @@ export class AtpAgent {
     } catch (e) {
       this.session = undefined
 
-      if (e instanceof XRPCError && e.error) {
+      if (e instanceof XRPCError) {
         /*
          * `ExpiredToken` and `InvalidToken` are handled in
          * `this_refreshSession`, and emit an `expired` event there.
--- a/packages/api/src/agent.ts
+++ b/packages/api/src/agent.ts
@@ -176,11 +176,20 @@ export class AtpAgent {
       this.session = undefined
 
       if (e instanceof XRPCError && e.error) {
-        // `ExpiredToken` and `InvalidToken` are handled by the `this._refreshSession`
-        if (['AuthMissing', 'InvalidDID'].includes(e.error)) {
+        /*
+         * `ExpiredToken` and `InvalidToken` are handled in
+         * `this_refreshSession`, and emit an `expired` event there.
+         *
+         * `AuthMissing` is a respose from `getSession`, and happens AFTER the
+         * initial `expired` event is sent. If we handle that here, we'll
+         * double-emit `expired`.
+         *
+         * `InvalidDID` is emit above, and should be handled here.
+         *
+         * Everything else is unexpected.
+         */
+        if (['InvalidDID'].includes(e.error)) {
           this._persistSession?.('expired', undefined)
-        } else {
-          this._persistSession?.('network-error', undefined)
         }
       } else {
         this._persistSession?.('network-error', undefined)
--- a/packages/api/src/agent.ts
+++ b/packages/api/src/agent.ts
@@ -170,7 +170,7 @@ export class AtpAgent {
       this.session.handle = res.data.handle
       this.session.emailConfirmed = res.data.emailConfirmed
       this._updateApiEndpoint(res.data.didDoc)
-      this._persistSession?.('create', this.session)
+      this._persistSession?.('update', this.session)
       return res
     } catch (e) {
       this.session = undefined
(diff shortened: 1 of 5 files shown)


packages/api/src/agent.ts
```diff
@@ -159,23 +159,45 @@ export class AtpAgent {
...
+         *
+         * `AuthMissing` is a respose from `getSession`, and happens AFTER the
+         * initial `expired` event is sent. If we handle that here, we'll
+         * double-emit `expired`.
+         *
+         * `InvalidDID` is emit above, and should be handled here.
+         *
+         * Everything else is unexpected.
+         */
+        if (['InvalidDID'].includes(e.error)) {
```
dholms, line 191: I'm a bit confused what circumstance this happens in?

packages/api/src/agent.ts
```diff
@@ -159,23 +159,45 @@ export class AtpAgent {
...
     } catch (e) {
       this.session = undefined
-      throw e
-    } finally {
-      if (this.session) {
-        this._persistSession?.('create', this.session)
+
+      if (e instanceof XRPCError && e.error) {
+        /*
+         * `ExpiredToken` and `InvalidToken` are handled in
```
dholms, line 180: is `this._refreshSession` implicitly called by the retry logic from calling `getSession`?

packages/api/src/agent.ts
```diff
@@ -159,23 +159,45 @@ export class AtpAgent {
...
+         *
+         * `InvalidDID` is emit above, and should be handled here.
+         *
+         * Everything else is unexpected.
+         */
+        if (['InvalidDID'].includes(e.error)) {
+          this._persistSession?.('expired', undefined)
+        } else if (['InternalServerError'].includes(e.error)) {
+          this._persistSession?.('network-error', undefined)
+        }
```
devinivy, line 195: I was under the impression that network errors (e.g. user does not have connectivity to the service) would have status `ResponseType.Unknown`, and may not have an `error` property.

packages/api/src/agent.ts
estrattonbailey: It's a little confusing, but we override the static `AtpAgent.fetch` here so that if we get an `ExpiredToken` we can try to refresh the session. So calling `getSession` actually calls `this._fetch` here, which calls `this._refreshSession`.

But yeah, basically `expired` will be emit from that handling + afaict we don't get a `ExpiredToken` error from `getSession`.

packages/api/src/agent.ts
estrattonbailey: This is just handling the logic that was already here, where a did doesn't match the requested session did. I honestly don't know why this would happen.

packages/api/src/agent.ts
devinivy: Alright, I think I see!  These 500s are treated akin to a network error.  That makes sense since it's another kind of temporary unavailability, and we don't necessarily want to log the user out.  Other status codes for temporary unavailability such as 502 and 503 might be handled similarly.

Tiny P.S.: I think that checking the `status` property of an `XRPCError` for this kind of thing might be more reliable.

estrattonbailey: `network-error` is just a name Paul suggested for this. This particular branch is tested in `agent.test.ts`, where we do get an `XRPCError` called `InternalServerError`.

Basically we're trying to handle any not-already-handled knowns with `expired`, and leave unknowns (like a network timeout that doesn't even hit our server) as `network-error`.

packages/api/src/agent.ts
dholms: ah cool cool 👍

dholms: and i guess if that happens, then somethings gone wonky & the appropriate action would be to just log the user out - hence the `expired`

dholms (approved):
thx for talking us through this - makes sense to me 👍 

I agree with Devin that status code checks are probably more reliable than error message checks. But it seems like any unknown error gets turned into `network-error` regardless, so it probably works either way

packages/api/src/agent.ts
```diff
@@ -159,23 +159,41 @@ export class AtpAgent {
...
+      this._persistSession?.('update', this.session)
       return res
     } catch (e) {
       this.session = undefined
-      throw e
-    } finally {
-      if (this.session) {
-        this._persistSession?.('create', this.session)
+
+      if (e instanceof XRPCError && e.error) {
```
devinivy, line 178: Now that we're doing the status code checks I think this can change to:
```suggestion
      if (e instanceof XRPCError) {
```

estrattonbailey: Bless

PR #2015, opened 2024-01-04
Include dids in bsky logs

devinivy:
This aligns appview logs with the pds logs, including the requester's did.


dholms (approved):
yup looks great 🙌

PR #2016, opened 2024-01-04
Lexicons for 3rd-party labelers

pfrazee:
Beginning the draft PR

### DIFF
--- a/lexicons/app/bsky/label/defs.json
+++ b/lexicons/app/bsky/label/defs.json
@@ -1,0 +1,103 @@
+{
+  "lexicon": 1,
+  "id": "app.bsky.label.defs",
+  "defs": {
+    "labelerView": {
+      "type": "object",
+      "required": ["uri", "cid", "did", "creator", "displayName", "indexedAt"],
+      "properties": {
+        "uri": { "type": "string", "format": "at-uri" },
+        "cid": { "type": "string", "format": "cid" },
+        "did": { "type": "string", "format": "did" },
+        "creator": { "type": "ref", "ref": "app.bsky.actor.defs#profileView" },
+        "displayName": { "type": "string" },
+        "description": {
+          "type": "string",
+          "maxGraphemes": 300,
+          "maxLength": 3000
+        },
+        "descriptionFacets": {
+          "type": "array",
+          "items": { "type": "ref", "ref": "app.bsky.richtext.facet" }
+        },
+        "avatar": { "type": "string" },
+        "policies": {"type": "ref", "ref": "app.bsky.label.defs#labelerPolicies"},
+        "likeCount": { "type": "integer", "minimum": 0 },
+        "viewer": { "type": "ref", "ref": "#labelerViewerState" },
+        "indexedAt": { "type": "string", "format": "datetime" }
+      }
+    },
+    "labelerViewerState": {
+      "type": "object",
+      "properties": {
+        "like": { "type": "string", "format": "at-uri" }
+      }
+    },
+    "labelerPolicies": {
+      "type": "object",
+      "required": ["reportReasons", "labelValues"],
+      "properties": {
+        "description": {
+          "type": "string",
+          "maxGraphemes": 10000,
+          "maxLength": 100000
+        },
+        "descriptionFacets": {
+          "type": "array",
+          "items": { "type": "ref", "ref": "app.bsky.richtext.facet" }
+        },
+        "reportReasons": {
+          "type": "array",
+          "items": {
+            "type": "string",
+            "knownValues": [
+              "com.atproto.moderation.defs#reasonSpam",
+              "com.atproto.moderation.defs#reasonViolation",
+              "com.atproto.moderation.defs#reasonMisleading",
+              "com.atproto.moderation.defs#reasonSexual",
+              "com.atproto.moderation.defs#reasonRude",
+              "com.atproto.moderation.defs#reasonOther",
+              "com.atproto.moderation.defs#reasonAppeal"
+            ]
+          }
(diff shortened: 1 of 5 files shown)


pfrazee:
In the preferences, I have the user setting visibility for label groups. In the labeler declaration, however, I have the labeler declaring the supported label values. My thinking is that our label definitions will be updated to include the label <=> label-group mapping, and that we'll assemble the label groups from the list of declared labels.

pfrazee:
Unsure if we want to support `!` imperative labels from 3P labelers

lexicons/app/bsky/label/labeler.json
```diff
@@ -0,0 +1,43 @@
+{
+  "lexicon": 1,
+  "id": "app.bsky.label.laneler",
```
dholms, line 3: typo

lexicons/app/bsky/label/defs.json
```diff
@@ -0,0 +1,103 @@
...
+        },
+        "descriptionFacets": {
+          "type": "array",
+          "items": { "type": "ref", "ref": "app.bsky.richtext.facet" }
+        },
+        "reportReasons": {
+          "type": "array",
+          "items": {
+            "type": "string",
+            "knownValues": [
```
dholms, line 53: this is defined in `com.atproto.moderation.defs` & we may just wish to re-use it

lexicons/app/bsky/label/defs.json
```diff
@@ -0,0 +1,103 @@
...
+      "type": "object",
+      "properties": {
+        "like": { "type": "string", "format": "at-uri" }
+      }
+    },
+    "labelerPolicies": {
+      "type": "object",
+      "required": ["reportReasons", "labelValues"],
+      "properties": {
+        "description": {
```
dholms, line 40: just so i understand: a labeler description will give a user-friendly high-level description of the labeler
and the policy description is the in the weeds more formal version right?

lexicons/app/bsky/label/defs.json
```diff
@@ -0,0 +1,103 @@
...
+              "com.atproto.moderation.defs#reasonViolation",
+              "com.atproto.moderation.defs#reasonMisleading",
+              "com.atproto.moderation.defs#reasonSexual",
+              "com.atproto.moderation.defs#reasonRude",
+              "com.atproto.moderation.defs#reasonOther",
+              "com.atproto.moderation.defs#reasonAppeal"
+            ]
+          }
+        },
+        "labelValues": {
```
dholms, line 64: may wanna pull this out into it's own def as well - could even re-use the def in the `label` "record" type

lexicons/app/bsky/label/defs.json
```diff
@@ -0,0 +1,103 @@
...
+          "type": "string",
+          "maxGraphemes": 300,
+          "maxLength": 3000
+        },
+        "descriptionFacets": {
+          "type": "array",
+          "items": { "type": "ref", "ref": "app.bsky.richtext.facet" }
+        },
+        "avatar": { "type": "string" },
+        "policies": {"type": "ref", "ref": "app.bsky.label.defs#labelerPolicies"},
```
dholms, line 24: may be worth having a `#labelerView` & `#labelerViewBasic` where the basic view doesn't include policies. Seems a little heavy duty to be sending down in every embed for instance

dholms:
> Unsure if we want to support ! imperative labels from 3P labelers

I would advocate yes

dholms:
> My thinking is that our label definitions will be updated to include the label <=> label-group mapping, and that we'll assemble the label groups from the list of declared labels.

Not quite sure I follow here... By "label definitions" do you mean the signed label objects that get passed around? Or do you mean the definitions that live over [here](https://github.com/bluesky-social/atproto/blob/main/packages/api/docs/labels.md)?

lexicons/app/bsky/label/defs.json
pfrazee: Are we able to use a ref for the items value? I honestly can't remember

lexicons/app/bsky/label/defs.json
pfrazee: Yep that's the logic

lexicons/app/bsky/label/defs.json
pfrazee: +1 if we can do refs for these kinds of values, my memory was that refs only worked for objects and arrays

lexicons/app/bsky/label/defs.json
pfrazee: Ah yes, I forgot about embeds in general. Will add that

pfrazee:
>> Unsure if we want to support ! imperative labels from 3P labelers

> I would advocate yes

Just bear in mind, it's not configurable. Essentially this would give any labeler you add the power to quietly `!filter`, for instance

pfrazee:
> > My thinking is that our label definitions will be updated to include the label <=> label-group mapping, and that we'll assemble the label groups from the list of declared labels.
> 
> Not quite sure I follow here... By "label definitions" do you mean the signed label objects that get passed around? Or do you mean the definitions that live over [here](https://github.com/bluesky-social/atproto/blob/main/packages/api/docs/labels.md)?

Yeah the latter, the label _value_ definitions

lexicons/app/bsky/label/defs.json
dholms: we've been using them for strings like this 👍 

you can see an example of this in `com.atproto.admin.defs#modEventReport`

dholms (approved):
looks good to me 👍

PR #2019, opened 2024-01-04
Alter entryway for separate mod backend

devinivy:
Buds with #1970, preparing the entryway service to run with a backing moderation service separate from the appview.


dholms (approved):
bless you 🙏

PR #2020, opened 2024-01-04
Ozone service partial bav

dholms:
Updated appview but it retains the routes from the previous appview (with moderation info architecture)


dholms:
Closing now - only used for cut-over during deploys

PR #2023, opened 2024-01-06
Appview v2 search, post threads fix, todos

devinivy:
A small smattering of changes:
 - using the search service for now to handle search skeletons, rather than the dataplane.
 - support receiving not-found error from dataplane thread skeleton.
 - make some endpoints fail that are not implemented on v2.


dholms (approved):
schweet 👍

PR #2031, opened 2024-01-09
Bsky appview sync service

devinivy:
Bsync is a private service that helps coordinate state between otherwise independent appviews.  In particular, this offers an interface to sync private mute state.  The interface is non-xrpc as this is not a public service with network-wide compatibility concerns—instead it is grpc/connect based on protobufs.  State is stored in postgres.

The service methods offered here are `addMuteOperation` and `scanMuteOperations`.  The former allows tracking an addition, removal, or full clearing of mute state for a given actor.  The operations are serialized, and you can scan over them using `scanMuteOperations`.  This endpoint offers long-polling to provide live updates.  These methods are authed using bearer API keys.

The protos can be found here: https://github.com/bluesky-social/atproto/blob/bsky-sync/packages/bsync/proto/bsync.proto


dholms (approved):
this all looks great 💎 

dig the long poll approach with NOTIFY/LISTEN & the use of grpc

DavidBuchanan314:
> between otherwise independent appviews.

Does this mean "independent" as in hosted by different orgs, or just multiple server instances controlled by the same entity?

Would it be fair to say this is an appview implementation detail, rather than a "real" protocol-level feature?

devinivy:
@DavidBuchanan314 good q!

> Does this mean "independent" as in hosted by different orgs, or just multiple server instances controlled by the same entity?

The latter 👍 We're going to make use of this to maintain multiple of our own shared-nothing appviews.  There's a small amount of state (i.e. mutes) that each appview can't maintain on its own from public data, and this service helps us manage that.

> Would it be fair to say this is an appview implementation detail, rather than a "real" protocol-level feature?

Accurate!

PR #2033, opened 2024-01-09
Support label issuer tied to appview v2

devinivy:
When talking to the dataplane, the appview may want to specify one or more labelers from which it always receives labels.  This adds a configuration `BSKY_LABELS_FROM_ISSUER_DIDS` which may be a comma-separated list of labeler dids.


dholms (approved):
ah cool i dig it 👍 

this will actually intersect nicely with my work over on https://github.com/bluesky-social/atproto/pull/2035

PR #2035, opened 2024-01-09
3p label hydration

dholms:


packages/bsky/src/context.ts
```diff
@@ -69,6 +71,15 @@ export class AppContext {
     })
   }
 
+  reqLabelers(req: express.Request): string[] {
+    const parsed =
+      req
+        .header('atproto-labelers')
+        ?.split(',')
```
devinivy, line 78: If the header is present but empty, we might end-up with `['']`.

PR #2037, opened 2024-01-09
Tidy feed generator language headers

dholms:
Building on https://github.com/bluesky-social/atproto/pull/2030

Tidy code a bit to bring it in line with our style & forward `content-language` header on responses to make the negotiation symmetrical

### DIFF
--- a/packages/pds/src/api/app/bsky/feed/getFeed.ts
+++ b/packages/pds/src/api/app/bsky/feed/getFeed.ts
@@ -1,11 +1,10 @@
 import { Server } from '../../../../lexicon'
 import AppContext from '../../../../context'
-import { forwardResHeader } from '@atproto/xrpc-server/src/util'
 
 export default function (server: Server, ctx: AppContext) {
   server.app.bsky.feed.getFeed({
     auth: ctx.authVerifier.access,
-    handler: async ({ req, res, params, auth }) => {
+    handler: async ({ req, params, auth }) => {
       const requester = auth.credentials.did
 
       const { data: feed } =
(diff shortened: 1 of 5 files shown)


packages/pds/src/api/app/bsky/feed/getFeed.ts
```diff
@@ -1,10 +1,11 @@
 import { Server } from '../../../../lexicon'
 import AppContext from '../../../../context'
+import { forwardResHeader } from '@atproto/xrpc-server/src/util'
```
devinivy, line 3: I wonder if this could create issues at runtime in the build, similar to what we've seen when trying to dip into the `@atproto/api/...`.

packages/pds/src/api/app/bsky/feed/getFeed.ts
```diff
@@ -16,24 +17,18 @@ export default function (server: Server, ctx: AppContext) {
...
-      const res = await ctx.appViewAgent.api.app.bsky.feed.getFeed(
+      // forward accept-language header to upstream services
+      serviceAuthHeaders.headers['accept-language'] =
+        req.headers['accept-language']
+      const feedRes = await ctx.appViewAgent.api.app.bsky.feed.getFeed(
         params,
-        feedOpts,
+        serviceAuthHeaders,
       )
+      forwardResHeader(res, feedRes.headers, 'content-language')
```
devinivy, line 27: I don't have any qualms with doing it imperatively like this, but on xrpc errors we support a `headers` field that xrpc-server will set on the response.  Eventually we should probably get the same thing going on non-error responses.

devinivy (approved):
Had a small logistical question around builds and referencing package internals.  That question aside, looks great 👍

packages/pds/src/api/app/bsky/feed/getFeed.ts
dholms: ah good call - i missed this import

ended up just passing headers through the route return statement

packages/pds/src/api/app/bsky/feed/getFeed.ts
dholms: actually we do have that! I just forgot about it - it should work if we just return headers in the route, next to `body`

PR #2041, opened 2024-01-11
Fix unnecessary decompress-compress of Multikey

MarshalX:
@dholms i didn't add new tests because imo `packages/identity/tests/did-document.test.ts` should cover it


MarshalX:
@dholms I thought about it again and probably want to close this... because DidDocument could be returned from untrusted sources like web resolver. Such keys could be poorly formatted or malformed. The decompress-compress roundtrip verifies that this Multikey is valid. Otherwise, raise an exception with a description. Removing this round trip will bypass such checks. This is not what we want. Probably adding a comment to the code about why we are doing it (like I described) is enough.

PR #2043, opened 2024-01-12
Ozone label endpoints

dholms:
Adds endpoints for
- subscribeLabels
- queryLabels

Subscribe labels uses the sequencer/outbox model that we also used on the PDS, backed by Postgres & just routine polling.

Note on deploying this:
I added a new serial `id` column to the label table. Not exactly sure how we want to manage this one but we'll have to figure out something sneaky on it 🤔


packages/ozone/src/config/env.ts
```diff
@@ -13,6 +13,9 @@ export const readEnv = (): OzoneEnvironment => {
     pdsDid: envStr('OZONE_PDS_DID'),
     dbPostgresUrl: envStr('OZONE_DB_POSTGRES_URL'),
     dbPostgresSchema: envStr('OZONE_DB_POSTGRES_SCHEMA'),
+    dbPoolSize: envInt('OZONE_DB_POOL_SIZE'),
```
dholms, line 16: these crept in cause I thought i was running into some connection pool issues in the test, but it turns out that wasn't the case

kept them around since it's just nice to have 👍

packages/ozone/src/api/label/queryLabels.ts
```diff
@@ -0,0 +1,55 @@
...
+        for (const pattern of uriPatterns) {
+          // if no '*', then we're looking for an exact match
+          if (!pattern.includes('*')) {
+            qb = qb.orWhere('uri', '=', pattern)
+          } else {
+            if (pattern.indexOf('*') < pattern.length - 1) {
+              throw new InvalidRequestError(`invalid pattern: ${pattern}`)
+            }
+            const searchPattern = pattern.slice(0, -1)
+            qb = qb.orWhere('uri', 'like', `${searchPattern}%`)
```
devinivy, line 25: May be nice to sanitize these search patterns so that they can't use weird arbitrary `like` pattern syntax.

Also just a note—I am pretty sure postgres will turn `like '{prefix}%'` into a simple range scan if `uri` is indexed 👍

packages/ozone/src/api/label/queryLabels.ts
```diff
@@ -0,0 +1,55 @@
...
+          }
+        }
+        return qb
+      })
+    }
+    if (sources && sources.length > 0) {
+      builder = builder.where('src', 'in', sources)
+    }
+    if (cursor) {
+      const cursorId = parseInt(cursor)
```
devinivy, line 35: Just avoids weirdness like incidentally supporting hex 😆 
```suggestion
      const cursorId = parseInt(cursor, 10)
```

packages/ozone/src/api/label/subscribeLabels.ts
```diff
@@ -0,0 +1,25 @@
+import { Server } from '../../lexicon'
+import AppContext from '../../context'
+import Outbox from '../../sequencer/outbox'
+import { InvalidRequestError } from '@atproto/xrpc-server'
+
+export default function (server: Server, ctx: AppContext) {
```
devinivy, line 6: Tidy ✅

packages/ozone/src/api/label/subscribeLabels.ts
```diff
@@ -0,0 +1,25 @@
...
+  server.com.atproto.label.subscribeLabels(async function* ({
```
devinivy, line 7: Tidy ✅

packages/ozone/src/mod-service/index.ts
```diff
@@ -858,12 +857,14 @@ export class ModerationService {
...
     }))
     const { ref } = this.db.db.dynamic
+    await sql`notify ${ref(LabelChannel)}`.execute(this.db.db) // emitted transactionally
     const excluded = (col: string) => ref(`excluded.${col}`)
     await this.db.db
       .insertInto('label')
       .values(dbVals)
       .onConflict((oc) =>
         oc.columns(['src', 'uri', 'cid', 'val']).doUpdateSet({
+          id: sql`${excluded('id')}`,
```
devinivy, line 867: Interesting!  Slightly surprised this just works, cool.

packages/ozone/src/mod-service/index.ts
```diff
@@ -858,12 +857,14 @@ export class ModerationService {
       neg: !!l.neg,
     }))
     const { ref } = this.db.db.dynamic
+    await sql`notify ${ref(LabelChannel)}`.execute(this.db.db) // emitted transactionally
```
devinivy, line 860: Re: the comment—does this need to be emitted transactionally?  It seems like it probably doesn't need to be transactional, and I have a feeling it often runs outside of a transaction.

devinivy (approved):
Looks solid—nice to be able to pinch the sequencer!

packages/ozone/src/api/label/queryLabels.ts
```diff
@@ -0,0 +1,55 @@
...
+            const searchPattern = pattern.slice(0, -1)
+            qb = qb.orWhere('uri', 'like', `${searchPattern}%`)
+          }
+        }
+        return qb
+      })
+    }
+    if (sources && sources.length > 0) {
+      builder = builder.where('src', 'in', sources)
+    }
```
devinivy, line 33: Do we need to support both multiple sources and multiple patterns w/ pagination?  I think multiple patterns w/ pagination or multiple sources w/ pagination will probably be simple enough for postgres, but all three at once could put it in a tricky spot for a large result set.

packages/ozone/src/mod-service/index.ts
dholms: Yeah i was unsure if it was going to. But seems to work nicely & thinking it through, it make sense since the ID is taken from the sequence before the write actually occurs

devinivy (approved):

PR #2044, opened 2024-01-12
Add from:me search syntax

mary-ext:
Alternative to https://github.com/bluesky-social/social-app/pull/2504 which does it in the client itself

Adds `from:me` convenient search syntax to search your own posts, done by replacing instances of `from:me` with the account's DID, with special care taken to avoid string replacements inside quotes as they are exact string matches.

I thought of adding it to Palomar directly, but since Palomar is unauthenticated I decided it'd be wise to implement it in the AppView layer instead.


bnewbold:
@mary-ext I think the AppView might actually be the best place to do this near-term? Not sure we want to pass DID on to palomar if we don't need to (TBD). Putting it in the appview allows mixing in account context, while not having things be client-specific.

PR #2050, opened 2024-01-15
Appview v2: utilize sorted-at field in bsky protos

devinivy:
We can defer to the bsky dataplane for sorted-at times on records.  This is generally a composite of the record's creation and indexed times.


dholms (approved):
nice 👌

PR #2055, opened 2024-01-17
Phone verification on entryway

dholms:
Adds optional phone verification to the entryway using Twilio's verify API

Building off of the schemas in https://github.com/bluesky-social/atproto/pull/2056


packages/pds/src/api/com/atproto/server/createAccount.ts
```diff
@@ -40,6 +40,30 @@ export default function (server: Server, ctx: AppContext) {
       const now = new Date().toISOString()
       const passwordScrypt = await scrypt.genSaltAndHash(password)
 
+      if (ctx.cfg.phoneVerification.required && ctx.twilio) {
+        if (!input.body.verificationPhone) {
+          throw new InvalidRequestError(
+            'Phone number verification is required on this server and none was provided.',
+            'InvalidPhoneVerification',
```
devinivy, line 47: We might consider adding this error type to the lexicon.

packages/pds/src/api/com/atproto/server/createAccount.ts
```diff
@@ -97,6 +121,19 @@ export default function (server: Server, ctx: AppContext) {
...
 
+        if (
+          ctx.cfg.phoneVerification.required &&
+          input.body.verificationPhone
+        ) {
+          await dbTxn.db
+            .insertInto('phone_verification')
+            .values({
+              did,
+              phoneNumber: input.body.verificationPhone,
```
devinivy, line 132: Not sure if twilio spits out a normalized version of phone numbers—but if they do it might be worth tracking it here.  It's a real pain to validate and normalize phone numbers, and we'll end-up storing a wide variety of different formats in here.

packages/pds/src/db/migrations/20240117T001106576Z-phone-verification.ts
```diff
@@ -0,0 +1,18 @@
...
+export async function up(db: Kysely<unknown>): Promise<void> {
+  await db.schema
+    .createTable('phone_verification')
+    .addColumn('did', 'varchar', (col) => col.primaryKey())
+    .addColumn('phoneNumber', 'varchar', (col) => col.notNull())
+    .execute()
+  await db.schema
+    .createIndex('phone_verification_number_idx')
+    .on('phone_verification')
+    .column('phoneNumber')
```
devinivy, line 12: We'll want to ensure the numbers stored here are normalized.

packages/pds/tests/phone-verification.test.ts
```diff
@@ -0,0 +1,156 @@
...
+      const code = verificationCodes[number]
+      sentCodes.push({ code, number })
+    }
+    ctx.twilio.verifyCode = async (number: string, code: string) => {
+      if (verificationCodes[number] === code) {
+        delete verificationCodes[number]
+        return true
+      }
+      return false
+    }
```
devinivy, line 43: I believe there's some comparable mocking code in dev-env—could we make use of that here?

devinivy (approved):
Lookin good!  Main thing to maybe give a think is storing/handling normalized phone numbers.

packages/pds/src/api/com/atproto/server/createAccount.ts
dholms: yeah i played kinda fast and loose cuz i know this is gonna be temporary. but yup probably a good idea 👍

packages/pds/src/api/com/atproto/server/createAccount.ts
dholms: yup looks like it does 👍

packages/pds/src/db/migrations/20240117T001106576Z-phone-verification.ts
dholms: ah right that's a good point 🤔

dholms: okay actually took a different track on this & decided to just enforce the format that twilio wants (https://www.twilio.com/docs/glossary/what-e164#regex-matching-for-e164)

the api forces it pretty close, but it still allows some slight variations (like dashes between numbers)

PR #2058, opened 2024-01-18
Ozone ACLs - old

dholms:
Switches out basic auth for ACLs in Ozone backend.
- list of moderators & triagers in ozone env
- appview & pds/entryway are configured with a `modServiceDid` for all moderation routes
  - retain (narrowly distributed) admin tokens on appview & pds as an escape hatch
- all requests that go to Ozone travel through the user's PDS
  - PDS signs a service auth token for the user & sends it to ozone backend (similar to appview auth)
  - Ozone will then handle the request or sign another request out to the relevant service using it's own service DID


dholms:
closing in favor of https://github.com/bluesky-social/atproto/pull/2252

PR #2059, opened 2024-01-18
lexicons: new app.bsky.graph.getRelationships endpoint

bnewbold:
This proposes the endpoint schema for a public (not auth'd) endpoint for fetching the public bsky social graph relationships (follows and blocks) between one actor and a list of "other" actors.

The initial motivation is to use this metadata from automod, for example in the context of a reply (wanting to know the relationships with the "root" and "parent" post authors.

If this looks good, would pass off to one of the backend folks to wire up and implement. v2 implementation should be simple, re-using an existing dataplane helper.

### DIFF
--- a/lexicons/app/bsky/graph/getRelationships.json
+++ b/lexicons/app/bsky/graph/getRelationships.json
@@ -12,6 +12,7 @@
           "actor": { "type": "string", "format": "at-identifier" },
           "others": {
             "type": "array",
+            "maxLength": 30,
             "items": {
               "type": "string",
               "format": "at-identifier"
--- a/lexicons/app/bsky/graph/getRelationships.json
+++ b/lexicons/app/bsky/graph/getRelationships.json
@@ -49,17 +49,6 @@
       "required": ["did"],
       "properties": {
         "did": { "type": "string", "format": "did" },
-        "handle": { "type": "string", "format": "handle" },
-        "blocking": {
-          "type": "string",
-          "format": "at-uri",
-          "description": "if the actor blocks this DID, this is the AT-URI of the block record or mod list record"
-        },
-        "blockedBy": {
-          "type": "string",
-          "format": "at-uri",
-          "description": "if the actor is blocked by this DID, this is the AT-URI of the block record or mod list record"
-        },
         "following": {
           "type": "string",
           "format": "at-uri",
--- a/lexicons/app/bsky/graph/getRelationships.json
+++ b/lexicons/app/bsky/graph/getRelationships.json
@@ -31,7 +31,10 @@
               "type": "array",
               "items": {
                 "type": "union",
-                "refs": ["#relationship", "app.bsky.graph.defs#notFoundActor"]
+                "refs": [
+                  "app.bsky.graph.def#relationship",
+                  "app.bsky.graph.defs#notFoundActor"
+                ]
               }
             }
           }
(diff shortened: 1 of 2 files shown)


lexicons/app/bsky/graph/getRelationships.json
```diff
@@ -0,0 +1,85 @@
...
+      "required": ["did"],
+      "properties": {
+        "did": { "type": "string", "format": "did" },
+        "handle": { "type": "string", "format": "handle" },
+        "blocks": {
+          "type": "string",
+          "format": "at-uri",
+          "description": "if the actor blocks this DID, this is the AT-URI of the block record"
+        },
+        "blocks_via_list": {
```
devinivy, line 57: In viewer state it's "blocked" and "blocking".  I wouldn't claim it's perfect, but maybe we should be consistent with it?

lexicons/app/bsky/graph/getRelationships.json
```diff
@@ -0,0 +1,85 @@
...
+          "type": "string",
+          "format": "at-uri",
+          "description": "if the actor blocks this DID, because of a mod list, this is the AT-URI of the mod list record"
+        },
+        "blocked_by": {
```
devinivy, line 62: Can we switch these fields to camel-case for consistency?

lexicons/app/bsky/graph/getRelationships.json
bnewbold: mmm, yeah, always get confused by those terms, but consistency is best; updated

lexicons/app/bsky/graph/getRelationships.json
bnewbold: haha, don't even notice when i'm sssslighering off

bnewbold:
Updated with camelCase and more consistent naming, and also realized we can put either record type (block record or mod list record) in the at-uri field, so removed the "ViaList" variants.

To capture some earlier discussion: this does make blocks a bit more visible to third parties. This is already public info but there is some friction. This endpoint is intentionally not helpful for bulk-enumeration of unknown relationships, but it does make specific pairs easy to check. This could actually be helpful for some devs, but there might also be abuse cases we aren't thinking of (beyond the existing abuse and drama around blocks being public at all). I'd be ok making this endpoint admin-only for now, but it would make automod less accessible to third parties, so i'd lean towards making it open.

bnewbold:
Thinking about it again, let's not tickle the dragon right now, and not make block relationships more accessible like this.

This can be an admin-auth'd rought, or actually I don't even need the block info right now for automod, so this could just return follow-graphic info and still be helpful.

lexicons/app/bsky/graph/getRelationships.json
```diff
@@ -0,0 +1,65 @@
...
+        }
+      },
+      "errors": [
+        {
+          "name": "ActorNotFound",
+          "description": "the primary actor at-identifier could not be resolved"
+        }
+      ]
+    },
+    "relationship": {
```
dholms, line 47: I think we should move this to `app.bsky.graph.defs` as it's a general description of a relationship & I suspect it may be reused across other lexicons. Thoughts?

bnewbold: yup, that makes sense!

bnewbold: pushed a change.

having it be in defs does make the "which of the two actors are we talking about" thing a bit more complex; I tried to clarify that with a description, but i'd be open if you want to include both DIDs in each relationship object (bloats the wire size a bit), or rename the `did` field to make it clearer.

PR #2064, opened 2024-01-19
Communication template admin endpoints on entryway

devinivy:
In support of #2045, this proxies the new communication template admin endpoints through the entryway.


dholms (approved):
looks good 👍

PR #2067, opened 2024-01-21
Appview v1 generating and ingesting mute ops w/ bsync

devinivy:
Two main changes here:
 - allow appview frontends to push mute ops to bsync.  Depending on the configuration, the appview frontends might write only to the db locally, only to bsync, or both.
 - allow appview ingester to consume and index mute ops from bsync.

Along the way I got bsync setup in the dev-env package, and made a small change to the convention for where built protos are written (now built into the package's `src/protos/` rather than the more ambiguous `src/gen/`).


packages/bsky/src/api/app/bsky/graph/muteActor.ts
```diff
@@ -18,10 +20,26 @@ export default function (server: Server, ctx: AppContext) {
...
+
+      if (ctx.bsyncClient) {
+        try {
+          await ctx.bsyncClient.addMuteOperation({
+            type: MuteOperation_Type.ADD,
+            actorDid: requester,
+            subject: subjectDid,
+          })
+        } catch (err) {
+          req.log.warn(err, 'failed to sync mute op to bsync')
```
dholms, line 40: i think we wanna throw here if `bsyncOnlyMutes` is true because then the operation failed

if it's false then it was partially applied but presumably the local db is considered the canonical state?

devinivy: Yep 💯 thanks for nabbing that.

PR #2068, opened 2024-01-21
:bug: Allow all mods to read comm templates

foysalit:


foysalit:
We decided not to allow this for now.

PR #2069, opened 2024-01-22
Ozone !takedown labels

dholms:
Adds two new labels to Ozone: `!takedown` and `!takedown-blobs`. These are imperative labels that instruct downstream services to remove the labels content from their indices entirely. `!takedown-blobs` has a subject of a record, and the semantic is to purge all blobs referenced by that record.

These labels are automatically applied when performing a takedown on either a repo or record.

### DIFF
--- a/packages/ozone/tests/moderation-appeals.test.ts
+++ b/packages/ozone/tests/moderation-appeals.test.ts
@@ -39,7 +39,7 @@ describe('moderation-appeals', () => {
 
   beforeAll(async () => {
     network = await TestNetwork.create({
-      dbPostgresSchema: 'ozone_moderation_statuses',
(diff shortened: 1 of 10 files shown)


packages/ozone/tests/moderation-appeals.test.ts
```diff
@@ -39,7 +39,7 @@ describe('moderation-appeals', () => {
 
   beforeAll(async () => {
     network = await TestNetwork.create({
-      dbPostgresSchema: 'ozone_moderation_statuses',
+      dbPostgresSchema: 'ozone_moderation_appeals',
```
dholms, line 42: this was just mislabeled from a previous PR

devinivy (approved):

PR #2073, opened 2024-01-23
Appview v1 maintaining device tokens and pushing notifications w/ courier

devinivy:
Two main changes here:

 - allow appview frontends to maintain device tokens with courier. Depending on the configuration, the appview frontends might write only to the db locally, only to courier, or both.
 - allow appview indexer to push notifications using courier.

Did a mini cleanup pass on the `NotificationServer` along the way.


dholms (approved):
looks great 🙌

PR #2083, opened 2024-01-24
Enable alternate appview proxy on pds

dholms:
Adds two new env variables for configuring an alternate appview url & a subset of users to use that appview

### DIFF
--- a/packages/pds/src/context.ts
+++ b/packages/pds/src/context.ts
@@ -44,7 +44,7 @@ export type AppContextOptions = {
   crawlers: Crawlers
   appViewAgent: AtpAgent
   alternateAppviewAgent?: AtpAgent
-  alternateAppviewUsers: Map<string, boolean>
+  alternateAppviewUsers: Set<string>
   moderationAgent: AtpAgent
   entrywayAgent?: AtpAgent
   authVerifier: AuthVerifier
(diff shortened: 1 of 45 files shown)


packages/pds/src/context.ts
```diff
@@ -162,6 +168,16 @@ export class AppContext {
...
     const appViewAgent = new AtpAgent({ service: cfg.bskyAppView.url })
+
+    const alternateAppviewAgent = cfg.bskyAppView.alternate
+      ? new AtpAgent({ service: cfg.bskyAppView.alternate.url })
+      : undefined
+    const alternateAppviewUsers = (
+      cfg.bskyAppView.alternate?.users ?? []
+    ).reduce((acc, cur) => {
+      return acc.set(cur, true)
+    }, new Map<string, boolean>())
```
devinivy, line 179: Really doesn't make a practical difference but a set would be quite tidy here!
```ts
const alternateAppviewUsers = new Set(cfg.bskyAppView.alternate?.users)
```

devinivy (approved):
Awesome, thanks!

packages/pds/src/context.ts
dholms: oh doh

dholms:
Closing - no longer needed

PR #2085, opened 2024-01-24
Add getRelationships route to appview

dholms:
Follow on implementation to the new lexicons in https://github.com/bluesky-social/atproto/pull/2059

Note: I did not add proxying for this route to pds/entryway as this is meant as a lower-level developer-oriented route.


devinivy (approved):
💎

bnewbold (approved):
cool, looks good!

automod does already talk directly to https://api.bsky.app for this sort of thing, so no gateway proxying needed for my current use case.

PR #2086, opened 2024-01-24
Add interest tags to preferences

estrattonbailey:
Adds a `interests` prop to preferences with the shape `{ onboardingTags: string[] }`. I decided to use `onboarding*` prefix in case we improve this later and want to use the more generic `tags` key then.

Update: talked to Paul and we decided to just use `tags` instead of `onboardingTags` 👍


dholms (approved):
looks good 👍

PR #2087, opened 2024-01-24
Ozone cdn invalidation

dholms:
Move CDN invalidation to Ozone. Invalidation is pushed out after the takedown event is pushed to appview.


packages/ozone/src/mod-service/index.ts
```diff
@@ -463,14 +471,38 @@ export class ModerationService {
...
+                ),
+            ),
           )
+
+          if (this.imgInvalidator) {
+            await Promise.allSettled(
+              (subject.blobCids ?? []).map((cid) => {
+                const paths = (this.cdnPaths ?? []).map((path) =>
+                  path.replace('%s', subject.did),
+                )
```
devinivy, line 494: I'm not noticing a place where the `cid` is also inserted into the path.  Pretty sure the invalidators themselves don't do this, so I think we'd want it around here.

services/ozone/api.js
```diff
@@ -12,6 +12,11 @@ require('dd-trace') // Only works with commonjs
 
 // Tracer code above must come before anything else
 const path = require('path')
+const {
+  BunnyInvalidator,
+  CloudfrontInvalidator,
+  MultiImageInvalidator,
+} = require('@atproto/aws')
```
devinivy, line 19: We may need to adjust the service's package.json and dockerfile to include `@atproto/aws`.

devinivy: Easiest thing to check/catch this is to make a build and run it, even if it's unconfigured.

devinivy (commented):
Looks solid, just a couple small things to peek at.

services/ozone/api.js
dholms: yup yup good point 👍

packages/ozone/src/mod-service/index.ts
dholms: hmm yup you're totally right i thought the img invalidator took care of that tbh 😅

packages/ozone/src/mod-service/index.ts
```diff
@@ -486,14 +494,38 @@ export class ModerationService {
...
+                  log.error({ err, ...evt }, 'failed to push blob event'),
+                ),
+            ),
           )
+
+          if (this.imgInvalidator) {
+            await Promise.allSettled(
+              (subject.blobCids ?? []).map((cid) => {
+                const paths = (this.cdnPaths ?? []).map((path) =>
+                  path.replace('%s', subject.did).replace('%s', cid),
```
devinivy, line 516: No issue here, but node's `util.format()` does this exact printf-style replacement!

devinivy (approved):

PR #2088, opened 2024-01-24
Appview v1 handling clearly bad cursors

devinivy:
This adjusts the appview so that clearly mismatching cursors return an empty result set rather than failing with an error.  Cursors that are close to the format but "not quite" there may still produce an error since it indicates some ambiguity.  This should facilitate a potential change in the cursor format without significantly affecting consumers.


dholms (approved):
beaut ✨

PR #2092, opened 2024-01-25
Appview v2 handling clearly bad cursors

devinivy:
This adjusts the appview so that clearly mismatching cursors return an empty result set rather than failing with an error. Cursors that are close to the format but "not quite" there may still produce an error since it indicates some ambiguity. This should facilitate a potential change in the cursor format without significantly affecting consumers.  In this case we target cursors from v1 with the hallmark `::`-separator.


dholms (approved):
amazing 👌

PR #2101, opened 2024-01-26
Temporary unspecced labels relating takedowns in ozone

devinivy:
While the names/semantics of these labels get nailed down, move to a clearly unspecced label for relating takedowns.  Based on top of #2069.


dholms (approved):
nice looks good 👍

PR #2106, opened 2024-01-29
Remove appview feedgen

dholms:
This removes the in-appview feed generator. 

All of our feeds are now moved over to a proper feed generator and this simplifies the logic for `getFeed`.

Also took the oppurtunity to deprecate the unspecced `getTimelineSkeleton` rout (verified that it has not been used at all in the last week).


devinivy (approved):
🤤 so nice to see this tidied up.

PR #2108, opened 2024-01-29
Simplify appview automod

dholms:
Since we're running a dedicated auto-mod service, we can pull the abyss & fuzzy matcher tests out of the in-appview automod.

Once we verify that img labeling is working correctly in the new auto-mod, we can fully deprecate the indexer automod


bnewbold:
This looks directionally correct to me:

- only in AppView, PDS still does some keyword stuff (eg, when creating account, changing handle, and some record updates like display name)
- removing abyss from appview indexing: check, automod is doing this in prod
- removing appview-side keyword checks resulting in mod reports: check, automod is doing this in prod

bnewbold (approved):

devinivy (approved):

PR #2109, opened 2024-01-29
Entryway: non-transactional createAccount

dholms:
Do network requests outside of the createAccount transaction on entrway. If the requests errors, then attempt to delete/cleanup the partially created entryway account.

### DIFF
--- a/packages/pds/src/api/com/atproto/server/createAccount.ts
+++ b/packages/pds/src/api/com/atproto/server/createAccount.ts
@@ -20,6 +20,7 @@ import Database from '../../../../db'
 import { didDocForSession } from './util'
 import { getPdsEndpoint } from '../../../../pds-agents'
 import { isThisPds } from '../../../proxy'
+import { dbLogger as log } from '../../../../logger'
 
 export default function (server: Server, ctx: AppContext) {
   server.com.atproto.server.createAccount({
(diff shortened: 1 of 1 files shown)


packages/pds/src/api/com/atproto/server/createAccount.ts
```diff
@@ -149,17 +150,21 @@ export default function (server: Server, ctx: AppContext) {
...
+            req.log.error(
+              { didKey: ctx.plcRotationKey.did(), handle },
+              'failed to create did:plc',
+            )
+            throw err
+          }
         }
-      })
+      } catch (err) {
+        await cleanupUncreatedAccount(ctx, did)
```
devinivy, line 165: I don't think there's an issue here, but we should just ensure there's no way to trigger a pre-existing did from getting "cleaned-up."

packages/pds/src/api/com/atproto/server/createAccount.ts
```diff
@@ -132,6 +120,10 @@ export default function (server: Server, ctx: AppContext) {
             .execute()
         }
 
+        if (!entrywayAssignedPds) {
+          await repoTxn.createRepo(did, [], now)
+        }
```
devinivy, line 125: I'm not sure how important this case actually is for the entryway, which shouldn't end-up hosting any repositories.  But if we create a repository and sequence them onto the stream prior to submitting their plc op, it's possible their first event wont get processed by a consumer.

devinivy (approved):
A couple notes to maybe give a think, but it looks nice to me 👍

packages/pds/src/api/com/atproto/server/createAccount.ts
dholms: yup i thought about that

but i also think we just want to remove all this repo logic out of the entryway anyway & it's not currently in use so we don't really run a risk with it

packages/pds/src/api/com/atproto/server/createAccount.ts
dholms: good call & we def want to be safe about that. Double checked the logic & the main transaction will throw an error if the did is already registered

PR #2112, opened 2024-01-30
Appview v2 don't apply 3p self blocks

devinivy:
Users aren't able to block themselves: now enforcing that in the logic for 3p blocks.


dholms (approved):
👍

PR #2113, opened 2024-01-30
Fix timing on appview v2 repo rev header

dholms:
Repo rev is processed last on record ingestion but is being grabbed first on view construction.

Change the order of fetching repoRev to the last thing we do before sending the response out. 

This will cause read-after-write behavior to err on the side of not showing the post rather than showing duplicates


bnewbold:
I like this kind of bug

devinivy (approved):

PR #2115, opened 2024-01-30
Signup queue notifications

dholms:
Building on top of https://github.com/bluesky-social/atproto/pull/2076

Adds notifications that go out when a user's account is activated on entryway:
- push notifications - sent to courier
- email - with a configurable rate limit per day


devinivy (approved):
Awesome

PR #2117, opened 2024-01-31
Appview v2 config tidy

devinivy:
Tidying appview v2 configuration:
 - appview v2 no longer behaves as a feedgen, related config and endpoints are removed.
 - removed deprecated timeline skeleton endpoint.
 - removed moderator and triage-specific roles/creds.
 - allow multiple admin creds, supporting better rotation.
 - standardized config names for urls to external services.
 - misc fixes in tests, e.g. removing unneeded `bsky` references in ozone tests.


dholms (approved):
amazing 🙌

PR #2119, opened 2024-01-31
Phone verification bypass number

dholms:
Add a specially configured phone number that allows you to bypass required phone verification on `createAccount`

### DIFF
--- a/packages/pds/src/api/com/atproto/server/createAccount.ts
+++ b/packages/pds/src/api/com/atproto/server/createAccount.ts
@@ -46,7 +46,7 @@ export default function (server: Server, ctx: AppContext) {
       const verificationPhone = await ensurePhoneVerification(
         ctx,
         input.body.verificationPhone,
-        input.body.verificationCode,
+        input.body.verificationCode?.trim(),
       )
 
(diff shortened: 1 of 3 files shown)


packages/pds/src/api/com/atproto/server/createAccount.ts
```diff
@@ -486,6 +464,42 @@ const ensureUnusedHandleAndEmail = async (
...
+    throw new InvalidRequestError(
+      `Text verification is now required on this server. Please make sure you're using the latest version of the Bluesky app.`,
+      'InvalidPhoneVerification',
+    )
+  }
+  if (ctx.cfg.phoneVerification.bypassPhoneNumber === phone) {
+    return undefined
+  }
+
+  if (!code) {
```
devinivy, line 486: We might consider trimming the code before this check.

devinivy (approved):

packages/pds/src/api/com/atproto/server/createAccount.ts
dholms: nice yup 👍

PR #2120, opened 2024-01-31
Appview v2 enforce post thread root boundary

devinivy:
The idea here is to ensure all posts in a given thread view have the same root as the anchor post.  During hydration we were looking-up each post in the thread's root, and as a result even bad roots would ended-up presented inside the thread if they were referenced somewhere.  When posts are well-behaved and set the parent/root as intended it's not an issue, so this is a bit of an edge case.

### DIFF
--- a/packages/bsky/src/views/index.ts
+++ b/packages/bsky/src/views/index.ts
@@ -46,7 +46,7 @@ import {
   isRecordWithMedia,
 } from './types'
 import { Label } from '../hydration/label'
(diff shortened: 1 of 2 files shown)


dholms (approved):
looks great 👍 

a test on this behavior would be great cause i could see it getting tweaked at some point

devinivy:
@dholms good call, meant to do that!  Added one, seems to have the intended effect (test fails w/o these changes).

PR #2122, opened 2024-02-02
Appview v2: don't apply missing modlists

dholms:
When hydrating a profile basic, check to ensure that relevant mod lists both exist & are have a purpose of being modlists. If the list does not exist as a modlist, then remove it from profile viewer state so that it does not affect filtering & views

### DIFF
--- a/packages/bsky/src/hydration/hydrator.ts
+++ b/packages/bsky/src/hydration/hydrator.ts
@@ -567,6 +567,7 @@ const listUrisFromProfileViewer = (item: ProfileViewerState | null) => {
   if (item?.blockingByList) {
     listUris.push(item.blockingByList)
   }
+  // blocked-by list does not appear in views, but will be used to evaluate the existence of a block between users.
   if (item?.blockedByList) {
     listUris.push(item.blockedByList)
   }
--- a/packages/bsky/src/hydration/hydrator.ts
+++ b/packages/bsky/src/hydration/hydrator.ts
@@ -87,18 +87,17 @@ export class Hydrator {
   }
 
   // app.bsky.actor.defs#profileView
-  // - profile
+  // - profile viewer
   //   - list basic
-  async hydrateProfiles(
+  // Note: builds on the naive profile viewer hydrator and removes references to lists that have been deleted
+  async hydrateProfileViewers(
     dids: string[],
-    viewer: string | null,
-    includeTakedowns = false,
+    viewer: string,
   ): Promise<HydrationState> {
-    const [actors, labels, profileViewers] = await Promise.all([
-      this.actor.getActors(dids, includeTakedowns),
-      this.label.getLabelsForSubjects(labelSubjectsForDid(dids)),
-      viewer ? this.actor.getProfileViewerStates(dids, viewer) : undefined,
-    ])
+    const profileViewers = await this.actor.getProfileViewerStatesNaive(
+      dids,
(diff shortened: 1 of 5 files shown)


packages/bsky/src/hydration/hydrator.ts
```diff
@@ -563,9 +567,36 @@ const listUrisFromProfileViewer = (item: ProfileViewerState | null) => {
   if (item?.blockingByList) {
     listUris.push(item.blockingByList)
   }
+  if (item?.blockedByList) {
+    listUris.push(item.blockedByList)
+  }
```
devinivy, line 572: It's relatively clear how `mutedByList` and `blockingByList` play into profile hydration (since they appear in the profile view), but this one is a bit less apparent—maybe we add a comment about it.
```suggestion
  // blocked-by list does not appear in views, but will be used to evaluate the existence of a block between users.
  if (item?.blockedByList) {
    listUris.push(item.blockedByList)
  }
```

packages/bsky/src/hydration/hydrator.ts
```diff
@@ -104,6 +104,10 @@ export class Hydrator {
       listUris.push(...listUrisFromProfileViewer(item))
     })
     const listState = await this.hydrateListsBasic(listUris, viewer)
+    // if a list no longer exists or is not a mod list, then remove from viewer state
+    profileViewers?.forEach((item) => {
+      removeNonModListsFromProfileViewer(item, listState)
+    })
```
devinivy, line 110: I'm a little torn on modifying the `ProfileViewer`s in this way.  There's an implicit dependency between `getProfileViewerStates()` and this logic.  We could also handle it on the flip side, when blocks are being evaluated, in `viewerBlockExists()` and `viewerMuteExists()`, and in some ways I think that fits better into how hydration vs. view logic currently works.  Either case introduces some dependencies, so I'd be for leaving another comment or two to document those.  That would be either:
 - documenting in `getProfileViewerStates()` the dependence on a later call to `removeNonModListsFromProfileViewer()` for evaluation of blocks/mutes or
 - documenting in `viewerMute/BlockExists()` that referenced lists must be hydrated.

packages/bsky/src/hydration/hydrator.ts
dholms: good call 👍

packages/bsky/src/hydration/hydrator.ts
dholms: yeah i thought about doing it in the view layer

but I think there's a decent argument that this should be handled in hydration. in v1, we were doing this in hydration & the lists on ProfileViewers don't actually "exist". Imo this isn't "view logic", it's an additional piece of logic on hydration that would ideally be handled in the hydration query but isn't because everything is denormalized & this is expensive to materialize 🤔 

Def down to rework logic/document it better

dholms: I agree it would be more ideal for this to be in `getProfileViewerStates` but then it mixes it up with list logic which is kinda pretty strange because we have to jam a list hydration call in there...

maybe we could add a higher level `getProfilerViewerStates` to the hydrator that handles this logic

and have a lower level `getUncheckedProfileViewerStates` on the profile hydrator that calls out the response is incomplete

devinivy (approved):

devinivy (approved):

PR #2124, opened 2024-02-02
:sparkles: Add date range and comment filter to queryModerationEvents endpoint

foysalit:
This PR adds a few more handy params/filters to the queryModerationEvents lexicon in order to allow mods to look at events 
1. within a certain date-time range
2. with a certain comment/note keyword
3. with comments only
4. emitted by a specific user/mod
5. with one or more report types
6. with one or more labels added/removed


packages/ozone/src/mod-service/index.ts
```diff
@@ -139,6 +147,20 @@ export class ModerationService {
...
+      builder = builder.where('createdAt', '<=', createdBefore)
+    }
+    if (hasComment) {
+      builder = builder.where((qb) => {
+        if (commentKeyword) {
+          return qb.where('comment', 'ilike', `%${commentKeyword}%`)
+        }
+        return qb.where('comment', 'is not', null)
+      })
+    }
```
devinivy, line 163: Would there be any issue making `hasComment` and `commentKeyword` independent from each other?

devinivy:
@foysalit would you mind adding a changeset for the change to `@atproto/api`?  Adding a changeset will facilitate the next release of the api package.  The command is `pnpm changeset`.  We'll want to mark the `@atproto/api` package for a patch release.  It will generate a file, which just needs to be checked-in to this PR.  Feel free to reach out if you want a hand with it, I realize you may not have had to add a changeset here before!

packages/ozone/src/mod-service/index.ts
```diff
@@ -162,6 +168,25 @@ export class ModerationService {
...
     }
 
+    // If multiple labels are passed, then only retrieve events where all those labels exist
+    if (addedLabels.length) {
+      builder = builder.where(
+        'createLabelVals',
+        'ilike',
+        `%${addedLabels.join('%')}%`,
+      )
+    }
```
devinivy, line 178: I believe this condition would require that the labels appear in a particular order in `createLabelVals`.

packages/ozone/src/mod-service/index.ts
```diff
@@ -162,6 +168,25 @@ export class ModerationService {
       })
     }
 
+    // If multiple labels are passed, then only retrieve events where all those labels exist
```
devinivy, line 171: The description on this param says "only events where one of these labels were added are returned".  I could imagine either behavior working fine for us, but may need to tweak either the implementation or description so they're consistent.

packages/ozone/src/mod-service/index.ts
foysalit: great catch!

packages/ozone/src/mod-service/index.ts
foysalit: good catch, the mods told me that majority of cases, only 1 label is applied so I thought when multiple labels are passed here, we would be specifically looking for events where all labels exist. the implementation followed that so I needed to adjust the comment here.

lexicons/com/atproto/admin/queryModerationEvents.json
```diff
@@ -35,6 +45,14 @@
...
             "default": 50
           },
+          "hasComment": {
+            "type": "boolean",
+            "description": "If true, only events with comments are returned"
+          },
+          "commentKeyword": {
+            "type": "string",
+            "description": "If specified, only events with comments containing the keyword are returned"
+          },
```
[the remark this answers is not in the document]
foysalit: yeah makes sense. I'll rename to comment.

packages/ozone/src/mod-service/index.ts
foysalit: nope, no issues at all.

PR #2125, opened 2024-02-02
:bug: Forward sendEmail event to moderationAgent instead of appview

foysalit:


devinivy (approved):

devinivy:
Just a headsup that this fix does already seems to be in place on the entryway (`bsky.social`).

PR #2131, opened 2024-02-03
Clear email tokens on email update

dholms:
https://github.com/bluesky-social/atproto/pull/2130

### DIFF
--- a/packages/pds/src/account-manager/index.ts
+++ b/packages/pds/src/account-manager/index.ts
@@ -327,7 +327,7 @@ export class AccountManager {
       await this.db.transaction((dbTxn) =>
         Promise.all([
           account.updateEmail(dbTxn, did, email),
-          emailToken.deleteEmailToken(dbTxn, did, 'update_email'),
+          emailToken.deleteAllEmailTokens(dbTxn, did),
         ]),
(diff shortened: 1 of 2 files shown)


packages/pds/src/account-manager/index.ts
```diff
@@ -327,7 +327,7 @@ export class AccountManager {
       await this.db.transaction((dbTxn) =>
         Promise.all([
           account.updateEmail(dbTxn, did, email),
-          emailToken.deleteEmailToken(dbTxn, did, 'update_email'),
+          emailToken.deleteAllEmailTokens(dbTxn, did),
```
devinivy, line 330: Do we want to run this even when there is no `token`?

devinivy (approved):

packages/pds/src/account-manager/index.ts
dholms: Yup I believe so

becuase we want to clear _all_ tokens regardless since they were associated with the old email

devinivy: Ah, I actually meant that it wasn't being run in that case and I figured we might want to!

PR #2135, opened 2024-02-05
Track blob cids on ozone mod events

devinivy:
By tracking blob cids on ozone moderation events, we have better book-keeping and are able to properly serve them on the `queryModerationEvents` endpoint.  One notable change here is that on a reverse-takedown event, we reverse the takedown on all blobs listed alongside the subject if that subject is takendown.  This is more true to how blob takedowns are modeled in ozone, where they are always attached to a record takedown.


dholms (approved):
looks good 👍 

have you put any thought into the migration?

PR #2155, opened 2024-02-08
Entryway: plivo integration

dholms:
Add the ability to configure the entrway to use plivo instead of twilio for sms verification


packages/pds/src/phone-verification/plivo.ts
```diff
@@ -0,0 +1,124 @@
...
+      ui8.toString(
+        ui8.fromString(`${opts.authId}:${opts.authToken}`, 'utf8'),
+        'base64pad',
+      )
+  }
+
+  async sendCode(phoneNumber: string) {
+    try {
+      const res = await this.makeReq(
+        `https://api.plivo.com/v1/Account/${this.authId}/Verify/Session/`,
```
devinivy, line 32: The trailing slash quirk could be worth a note.

devinivy, line 32: Any chance `authId` could need to be URI-encoded?

devinivy (approved):
💎

packages/pds/src/phone-verification/plivo.ts
dholms: nope it's all alphanumeric 👍

dholms: good call 👍

PR #2168, opened 2024-02-12
Update `blueskyweb.xyz` links in README.md to `bsky.social`

surfdude29:
This PR updates the two links in README.md that point to `blueskyweb.xyz` to the equivalent `bsky.social` URLs.

On a related note: does this email address need updating or is it still valid?

https://github.com/bluesky-social/atproto/blob/9579bec720d30e40c995d09772040212c261d6fb/package.json#L5

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,12 +61,12 @@ make help
 
 ## About AT Protocol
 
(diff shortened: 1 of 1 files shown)


bnewbold:
thanks!

the email is still valid

PR #2170, opened 2024-02-12
Account migration schemas

dholms:


lexicons/com/atproto/identity/signPlcOp.json
```diff
@@ -0,0 +1,41 @@
...
+          }
+        }
+      },
+      "output": {
+        "encoding": "application/json",
+        "schema": {
+          "type": "object",
+          "required": ["plcOp"],
+          "properties": {
+            "plcOp": {
```
devinivy, line 32: Maybe we should sync this property name up with the input of `identity.sendPlcOp` (i.e. both "op" or both "plcOp").

lexicons/com/atproto/repo/checkMissingBlobs.json
```diff
@@ -0,0 +1,36 @@
+{
+  "lexicon": 1,
+  "id": "com.atproto.repo.listMissingBlobs",
```
devinivy, line 3: After seeing this, I could imagine moving migration-specific endpoints like this into their own segment, e.g. `com.atproto.migrate.listMissingBlobs`.

lexicons/com/atproto/server/checkAccountStatus.json
```diff
@@ -0,0 +1,34 @@
...
+            "importedPrivateState",
+            "expectedBlobs",
+            "importedBlobs"
+          ],
+          "properties": {
+            "activated": { "type": "boolean" },
+            "didMigrated": { "type": "boolean" },
+            "uploadedRepo": { "type": "boolean" },
+            "importedRepo": { "type": "boolean" },
+            "importRepoError": { "type": "string" },
```
devinivy, line 25: It could be nice to make room for an error message and a code/name here.

dholms: I actually had a change of heart and decided against processing the repo async, that way we can just report the error directly in the request

lexicons/com/atproto/repo/checkMissingBlobs.json
dholms: I actually think this is a useful route in it's own right fwiw. There are some repos now that are missing blobs and this could be useful to figure out what

lexicons/com/atproto/identity/signPlcOp.json
dholms: yup agreed 👍

lexicons/com/atproto/identity/getRecommendedDidCredentials.json
```diff
@@ -0,0 +1,28 @@
...
+  "defs": {
+    "main": {
+      "type": "query",
+      "description": "Describe the credentials that should be included in the DID doc of an account that is migrating to this service.",
+      "output": {
+        "encoding": "application/json",
+        "schema": {
+          "type": "object",
+          "properties": {
+            "rotationKeys": {
```
bnewbold, line 13: it is a tiny bit weird having `rotationKeys` here. it matches the PLC operation syntax/fields, but the other fields here match DID document fields but this one doesn't. I guess it's fine? `plcRotationKeys` is tempting but probably not worth breaking consistency.

a `description` on this field (in the Lexicon schema) would be good.

lexicons/com/atproto/identity/sendPlcOp.json
```diff
@@ -0,0 +1,20 @@
+{
+  "lexicon": 1,
+  "id": "com.atproto.identity.sendPlcOp",
```
bnewbold, line 3: maybe `submitPlcOp`? that is the verb used in the description and feels a bit better, but not a strong feeling.

lexicons/com/atproto/identity/signPlcOp.json
```diff
@@ -0,0 +1,41 @@
+{
+  "lexicon": 1,
+  "id": "com.atproto.identity.signPlcOp",
+  "defs": {
+    "main": {
+      "type": "procedure",
+      "description": "Signs a PLC operation to update some value(s) in the requesting DID's document.",
```
bnewbold, line 7: does this endpoint also the same safety checks as `sendPlcOp`, or nah?

lexicons/com/atproto/identity/sendPlcOp.json
```diff
@@ -0,0 +1,20 @@
...
+  "defs": {
+    "main": {
+      "type": "procedure",
+      "description": "Validates a PLC operation to ensure that it doesn't violate a service's constraints or get the identity into a bad state, then submits it to the PLC registry",
+      "input": {
+        "encoding": "application/json",
+        "schema": {
+          "type": "object",
+          "required": ["op"],
```
bnewbold, line 12: I would probably expand all these "op" to "operation". eg the plc directory APIs have full "operation".

but not a strong feeling.

lexicons/com/atproto/identity/sendPlcOp.json
```diff
@@ -0,0 +1,20 @@
+{
+  "lexicon": 1,
+  "id": "com.atproto.identity.sendPlcOp",
+  "defs": {
+    "main": {
+      "type": "procedure",
+      "description": "Validates a PLC operation to ensure that it doesn't violate a service's constraints or get the identity into a bad state, then submits it to the PLC registry",
```
bnewbold, line 7: should we allow a param to skip validation, like we do for lexicons?

I'd be fine not making the validation optional to start, but maybe add it in the future.

lexicons/com/atproto/repo/importRepo.json
```diff
@@ -0,0 +1,13 @@
+{
+  "lexicon": 1,
+  "id": "com.atproto.repo.importRepo",
+  "defs": {
+    "main": {
+      "type": "procedure",
+      "description": "Import a repo in the form of a CAR file. Returns a 202 after successfully reading incoming CAR then asynchronously indexes the repo and contained records",
```
bnewbold, line 7: I think we discussed requiring `Content-Length` HTTP header for this one; if so should mention in description.

Should probably also clarify when this endpoint can be called: only during migration? at any time, overwriting existing repo state? only when account is deactivated?

lexicons/com/atproto/server/checkAccountStatus.json
```diff
@@ -0,0 +1,38 @@
+{
+  "lexicon": 1,
+  "id": "com.atproto.server.checkAccountStatus",
+  "defs": {
+    "main": {
+      "type": "query",
+      "description": "Returns the status of an account, especially as pertaining to import or recovery. Can be called many times over the course of an account migration.",
```
bnewbold, line 7: I assume this is auth-required and only the account can call on themself; should probably add that to description.

should we allow admins to call for any account? might help with debugging. in that case would want an optional accountDid param, or something like that. could come back to this later.

lexicons/com/atproto/server/deactivateAccount.json
```diff
@@ -0,0 +1,23 @@
+{
+  "lexicon": 1,
+  "id": "com.atproto.server.deactivateAccount",
+  "defs": {
+    "main": {
+      "type": "procedure",
+      "description": "Deactivates a currently active account. Used to finalize account migration with the old host after the account has been activated on the new host.",
```
bnewbold, line 7: will we open this up to be the way for folks to temporarily deactivate/reactivate accounts as an account status thing? not sure if described the way it is to set near-term or long-term expectations.

lexicons/com/atproto/server/getServiceAuth.json
```diff
@@ -0,0 +1,33 @@
...
+  "lexicon": 1,
+  "id": "com.atproto.server.getServiceAuth",
+  "defs": {
+    "main": {
+      "type": "query",
+      "description": "Get a signed token on behalf of the requesting DID for the requested service.",
+      "parameters": {
+        "type": "params",
+        "required": ["aud"],
+        "properties": {
```
bnewbold, line 11: what about expiration (`exp`)? should client be allowed to request longer/shorter tokens within some range?

lexicons/com/atproto/server/getServiceAuth.json
```diff
@@ -0,0 +1,33 @@
...
+          "aud": {
+            "type": "string",
+            "format": "did",
```
bnewbold, line 14: if we make this `format: uri` it would work with DIDs and leave the door open to other audience URIs/values... but probably best to keep it DID-only.

lexicons/com/atproto/repo/checkMissingBlobs.json
bnewbold: does this let external folks discover which blobs have been purged for an account? or do purged blobs not show up in this output?

lexicons/com/atproto/repo/checkMissingBlobs.json
```diff
@@ -0,0 +1,36 @@
...
+      "output": {
+        "encoding": "application/json",
+        "schema": {
+          "type": "object",
+          "required": ["cids"],
+          "properties": {
+            "cursor": { "type": "string" },
+            "cids": {
+              "type": "array",
+              "items": { "type": "string", "format": "cid" }
```
bnewbold, line 29: I don't know if it is easy or worth it, but if this returned blob objects (eg, with size, mimetype, etc, in addition to CID), might be more ergonomic and helpful for humans. and machines could do things like compute total size of missing objects.

bnewbold (commented):
left a whole lot of comments. mostly just sharing thoughts.

overall pretty excited, this is looking great!

lexicons/com/atproto/identity/getRecommendedDidCredentials.json
dholms: Yeah I agree :/ But I think it makes most sense to return them here & keep it consistent with the other routes.

One thing we can do that may make it go down a bit easier is not return `rotationKeys` when this route is requested by a `did:web`. If we add some additional method that requires another field, we can add another property to the response this would be undefined for plc & web dids

lexicons/com/atproto/identity/sendPlcOp.json
dholms: yeah I like that 👍

lexicons/com/atproto/identity/signPlcOp.json
dholms: No it doesn't, because it's used to sign an operation for a possibly different service. Though that reminds me, I do intend to put it behind email verification 👍

lexicons/com/atproto/identity/sendPlcOp.json
dholms: eh yeah probably a good idea. generally we don't shoot for brevity in lexicon 😛

lexicons/com/atproto/identity/sendPlcOp.json
dholms: In my view, if you want to do that, you should just send the operation to PLC yourself. But yeah we can always add it later if we decide we want it 👍

lexicons/com/atproto/repo/importRepo.json
dholms: Good call on the content-length header.

Hmm yeah I'm not quite sure on when it can be called 🤔 

Part of me thinks it'd be fine to call it at any time?

lexicons/com/atproto/server/checkAccountStatus.json
dholms: yup added to description 👍 

i think we if we open this up for admins, then we can add the accountDid param 👍

lexicons/com/atproto/server/deactivateAccount.json
dholms: I think yes, but we shouldn't advertise it as a feature just yet

lexicons/com/atproto/server/getServiceAuth.json
dholms: Makes me a bit uneasy honestly - I don't think we want to offer long versions of these tokens. If you request one, you should really only be using it once

lexicons/com/atproto/server/getServiceAuth.json
dholms: I like just leaving it DID for now until we see the need for that. I think we can upgrade into `format: uri` without much trouble

lexicons/com/atproto/repo/checkMissingBlobs.json
dholms: hmmm that's not a bad idea 🤔 down to take a stab at it

And to answer the prev question (which GH isn't letting me respond to for some reason), this is not accessible to external accounts, but purged blobs would not show up in the output

dholms: Updated it to the `blob` type (similar to output of `uploadBlob`) which contains all that info 👍

lexicons/com/atproto/server/getServiceAuth.json
bnewbold: I definitely think the returned token should have some `exp`; I guess we can just set it to 5-10min in the future or something like that?

lexicons/com/atproto/repo/importRepo.json
bnewbold: It is a somewhat destructive operation (wipes existing content, I assume). It would be bad if somebody thought it was additive, imported a small CAR, and it blasted all their stuff away. I think it would blast all downstream copies fast also (eg, Relay would ~immediately have the new small copy). This concern could be mitigated by having a param like `clobberExisting` which makes it really clear you are doing something dangerous. It still feels like this is almost in the territory of needing to verify with email/token (like deleting account).

There are also some resource consumption issues: if you kept importing large but different repos, it would consume PDS, Relay, AppView, search, automod, etc resources every time. This can be mitigated by just tightly rate-limiting this endpoint.

New question: what kind of validation of the commit object is done, if any? I assume we should ignore signature (might be an old CAR file, with out-of-date signing key). What if the DID doesn't match? Probably fine in most cases.

How does the new PDS chose a `rev` for the new commit after import? If we lean in to `rev` as TID, it is easy: just use the current timestamp TID and everything should work fine. If the imported commit object has a DID match, and a `rev` number, could also start incrementing from that previous `rev`, though I think this is weaker (maybe the CAR export didn't include some more recent commits, and `rev` would not be current). (I of course lean towards using current TID for rev, but laying out the options)

lexicons/com/atproto/identity/sendPlcOp.json
bnewbold: yes! good call

lexicons/com/atproto/identity/getRecommendedDidCredentials.json
bnewbold: nice! that is a good way to handle it.

lexicons/com/atproto/repo/checkMissingBlobs.json
bnewbold: (dholms replied below; I think the reason the github reply box didn't show in my review is that this comment was on the same line that divy commented on previously?)

lexicons/com/atproto/server/checkAccountStatus.json
bnewbold: hmm! that is a big change (not being async). it does feel like async would be more general and reliable than keeping a long HTTP request open, and enable a nice dynamic progress bar UI.

but we could add that as an alternative endpoint later.

bnewbold:
all looking good! I left follow-up comments on importRepo and authtoken sections.

lexicons/com/atproto/server/getServiceAuth.json
dholms: Oh to clarify - the returned token does have an `exp` and it's set to 60s 👍 

We just don't provide optionality around requesting a longer one

lexicons/com/atproto/repo/importRepo.json
dholms: Good points on the destructiveness & possible abuse vector 🤔 I'll lock this down a bit so it can really only be used for account migration right now (we can make that more lenient in the future if we see the need)

For new commits, we just use the current TID as you pointed out 👍 

In terms of validation on commit object: We don't do any. we just import the "data" part of the repo. We ignore signature for the reason you mention. I don't bother checking did/rev because there's not really any attack you're preventing if we're not checking the signature 🤷‍♂️ 

In a way this is kind've a "massive bulk `applyWrites`" where the writes are just whatever is including in the imported MST

lexicons/com/atproto/server/describeServer.json
```diff
@@ -9,7 +9,7 @@
         "encoding": "application/json",
         "schema": {
           "type": "object",
-          "required": ["availableUserDomains"],
+          "required": ["did", "availableUserDomains"],
```
devinivy, line 12: Ooh this is a sneaky little breaking change we need to be a little careful with for rollout.

dholms: yup yup  - just need to ensure we update all our PDSs before publishing this new agent library

thought it was important to get in before there are other PDSs out in the wild

snarfed: funny, this bit me too, only just now realized I needed to add it. bsky.app was unhappy with my PDS just now until I did. 😆

PR #2173, opened 2024-02-13
repo: commit prev as nullable, but non-optional

bnewbold:
To match spec change: https://github.com/bluesky-social/atproto-website/pull/291

Decided not to bite of subscription event schema changes in this PR. The fields are already marked as deprecated there.


dholms (approved):
yup this looks good 👍

PR #2177, opened 2024-02-13
Support admin account deletion on entryway

devinivy:


packages/pds/src/api/com/atproto/admin/deleteAccount.ts
```diff
@@ -0,0 +1,82 @@
...
+          await retryHttp(() =>
+            agent.com.atproto.admin.deleteAccount(
+              { did },
+              {
+                encoding: 'application/json',
+                headers: ctx.authVerifier.createAdminRoleHeaders(),
+              },
+            ),
+          )
+        } catch (err) {
```
dholms, line 58: this seems like it might be a good place to return an error to the caller rather than silently logging & continuing with deletion on entryway 🤔 

otherwise i think this can get into a state where it's deleted on the entryway but not on the pds

PR #2179, opened 2024-02-14
Account migration

dholms:
Adds account migration flow based on schemas in https://github.com/bluesky-social/atproto/pull/2170
Entryway impl: https://github.com/bluesky-social/atproto/pull/2185

Supporting features for account migration:
- ability to "import" a repo and index the diff (configurable to be not allowed)
- ability to request a signed short-lived service auth token for account
- ability to create an account with a PDS with an existing DID based on a proof of ownership of that DID (service auth token as an authorization header)
- ability to request an arbitrary signed PLC operation from the PDS. This route requires confirmation through an email token
- a route for a PDS to submit a PLC op on the user's behalf. The PDS will check the PLC to ensure that it correctly configures the DID. If it would break the account (bad pds endpoint/signing key/rotation keys/handle), it does not allow submission
- a route to get a recommendation of DID credentials for a user's account
- a route to introspect the status of an account - repo state, indexed record count, whether it is active or not, whether the DID is active or not, how many blobs are missing, etc
- a route to an account to list missing blobs in their repository, along with the record uri that the blob is associated with

Smuggled into this is the ability to activate/deactivate an atproto account. When deactivated, the account cannot make any writes to its repository, and reads for the repository are not served to any viewers (other than the account in question).


packages/pds/src/api/com/atproto/identity/resolveHandle.ts
```diff
@@ -18,7 +18,7 @@ export default function (server: Server, ctx: AppContext) {
     }
 
     let did: string | undefined
-    const user = await ctx.accountManager.getAccount(handle, true)
```
dholms, line 21: is there a reason we were resolving handles for takendown accounts still?

packages/pds/src/account-manager/db/migrations/002-account-deactivation.ts
```diff
@@ -0,0 +1,17 @@
+import { Kysely, sql } from 'kysely'
```
devinivy, line 1:
```suggestion
import { Kysely } from 'kysely'
```

packages/pds/src/account-manager/helpers/account.ts
```diff
@@ -62,16 +72,21 @@ export const registerActor = async (
...
     db.db
       .insertInto('actor')
       .values({
         did,
         handle,
         createdAt: new Date().toISOString(),
+        deactivatedAt: deactivated ? new Date().toISOString() : null,
+        deleteAfter: deactivated
+          ? new Date(Date.now() + 3 * DAY).toISOString()
+          : null,
```
devinivy, line 89: It's total nit but it would be nice for both of these values to be based on the same date.  E.g.
```ts
const now = new Date()
// ...
  deactivatedAt: deactivated ? now.toISOString() : null,
  deleteAfter: deactivated
    ? new Date(now.getTime() + 3 * DAY).toISOString()
    : null,
```

packages/pds/src/actor-store/blob/reader.ts
```diff
@@ -73,4 +73,62 @@ export class BlobReader {
...
+      .selectFrom(
+        this.db.db
+          .selectFrom('record_blob')
+          .select('blobCid')
+          .distinct()
+          .as('unique_cids'),
+      )
+      .select(countAll.as('count'))
+      .executeTakeFirst()
+    return res?.count ?? 0
```
devinivy, line 105: This could be simplified down to `select count(distinct "blobCid") from record_blob`.

packages/pds/src/actor-store/repo/sql-repo-reader.ts
```diff
@@ -136,6 +137,14 @@ export class SqlRepoReader extends ReadableBlockstore {
     return builder.execute()
   }
 
+  async countBlocks(): Promise<number> {
```
devinivy, line 140: Just a note that this could cause some blocking.

packages/pds/src/api/com/atproto/identity/updateHandle.ts
```diff
@@ -42,7 +42,9 @@ export default function (server: Server, ctx: AppContext) {
       })
 
       // Pessimistic check to handle spam: also enforced by updateHandle() and the db.
-      const account = await ctx.accountManager.getAccount(handle)
+      const account = await ctx.accountManager.getAccount(handle, {
+        includeDeactivated: true,
+      })
```
devinivy, line 47: So we reserve handles of deactivated users but don't resolve them from `resolveHandle`—sound right?

packages/pds/src/api/com/atproto/repo/uploadBlob.ts
```diff
@@ -12,9 +12,29 @@ export default function (server: Server, ctx: AppContext) {
     handler: async ({ auth, input }) => {
       const requester = auth.credentials.did
 
-      const blob = await ctx.actorStore.transact(requester, (actorTxn) => {
-        return actorTxn.repo.blob.addUntetheredBlob(input.encoding, input.body)
-      })
+      const blob = await ctx.actorStore.transact(
```
devinivy, line 15: The changes in here underline for me how it might be beneficial to split-up the db work from the upload work.  As I understand it, as long as this transaction remains open other writes to the same actor store will block/timeout.  Adding a read into the transaction makes it a little higher stakes now, as I think that kicks sqlite into doing some additional locking.

packages/pds/src/api/com/atproto/server/activateAccount.ts
```diff
@@ -0,0 +1,31 @@
...
+
+      const root = await ctx.actorStore.read(requester, (store) =>
+        store.repo.storage.getRootDetailed(),
+      )
+      const commitData = {
+        cid: root.cid,
+        rev: root.rev,
+        since: null,
+        prev: null,
+        newBlocks: new BlockMap(),
```
devinivy, line 24: Should the signed commit block end-up in here?

packages/pds/tests/plc-operations.test.ts
```diff
@@ -0,0 +1,226 @@
...
+      { operation },
+      {
+        encoding: 'application/json',
+        headers: sc.getHeaders(alice),
+      },
+    )
+    await expect(attempt).rejects.toThrow(expectedErr)
+  }
+
+  it('prevents submitting an operation that removes the servers rotation key', async () => {
```
devinivy, line 92:
```suggestion
  it('prevents submitting an operation that removes the server\'s rotation key', async () => {
```

packages/pds/src/api/com/atproto/identity/submitPlcOperation.ts
```diff
@@ -0,0 +1,48 @@
...
+      const requester = auth.credentials.did
+      const op = input.body.operation
+
+      if (!check.is(op, plc.def.operation)) {
+        throw new InvalidRequestError('Invalid operation')
+      }
+
+      if (!op.rotationKeys.includes(ctx.plcRotationKey.did())) {
+        throw new InvalidRequestError(
+          'Rotation keys do not include servers rotation key',
```
devinivy, line 20:
```suggestion
          'Rotation keys do not include server\'s rotation key',
```

packages/pds/src/api/com/atproto/repo/importRepo.ts
```diff
@@ -0,0 +1,128 @@
...
+  server.com.atproto.repo.importRepo({
+    auth: ctx.authVerifier.accessNotAppPassword,
+    handler: async ({ input, auth }) => {
+      const did = auth.credentials.did
+      if (!ctx.cfg.service.acceptingImports) {
+        throw new InvalidRequestError('Service is not accepting repo imports')
+      }
+      await ctx.actorStore.transact(did, (store) =>
+        importRepo(store, input.body),
+      )
```
devinivy, line 28: I do think that in general—i.e. even if not given our implementation/policy—we can expect this to last long enough that it might time out the connection.  And some implementations might want to plop it on a job queue and deal with it totally asynchronously.

Even we might consider just tossing the repo car on disk or into s3 then working with it streaming or in parts to make it robust to repos that don't happily fit into memory.

devinivy: (To be clear I think this is fine for our current purposes, mostly want to leave the door open to these possibilities.)

packages/pds/src/api/com/atproto/repo/importRepo.ts
```diff
@@ -0,0 +1,128 @@
...
+  actorStore: ActorStoreTransactor,
+  incomingCar: AsyncIterable<Uint8Array>,
+) => {
+  const now = new Date().toISOString()
+  const rev = TID.nextStr()
+  const did = actorStore.repo.did
+
+  const { roots, blocks } = await readCarStream(incomingCar)
+  if (roots.length !== 1) {
+    throw new InvalidRequestError('expected one root')
```
devinivy, line 43: In the future we might look into failing earlier for this case which could feasibly be detected up front, just a slight vector for abuse.

packages/pds/src/api/com/atproto/repo/importRepo.ts
```diff
@@ -0,0 +1,128 @@
...
+        await actorStore.record.deleteRecord(uri)
+      } else {
+        let parsedRecord: RepoRecord
+        try {
+          const parsed = await getAndParseRecord(blocks, write.cid)
+          parsedRecord = parsed.record
+        } catch {
+          throw new InvalidRequestError(
+            `Could not parse record at '${write.collection}/${write.rkey}'`,
+          )
```
devinivy, line 76: I think any errors in here will end-up unhandled and cause a crash.  What we probably want to do is setup an `AbortController` and user `p-queue``'s support for signals to thread it through and bail on work once there's a failure.  One the `diff.writes` loop ends and the queue is determined idle we can call `signal.throwIfAborted()` to raise the original error if one occurred.

packages/pds/src/api/com/atproto/repo/importRepo.ts
```diff
@@ -0,0 +1,128 @@
...
+            : Promise.resolve()
+        await Promise.all([indexRecord, indexRecordBlobs])
+      }
+    })
+  }
+  await recordQueue.onIdle()
+}
+
+export const findBlobRefs = (val: LexValue, layer = 0): BlobRef[] => {
+  if (layer > 10) {
```
devinivy, line 107: It's possible that we upped the max layer count here in #2185.

packages/pds/src/api/com/atproto/server/confirmEmail.ts
```diff
@@ -9,7 +9,10 @@ export default function (server: Server, ctx: AppContext) {
     handler: async ({ auth, input, req }) => {
       const did = auth.credentials.did
 
-      const user = await ctx.accountManager.getAccount(did)
+      const user = await ctx.accountManager.getAccount(did, {
+        includeDeactivated: true,
+        includeTakenDown: true,
```
devinivy, line 14: I don't think we let takendown accounts to this point thanks to `accessCheckTakedown`.

packages/pds/src/api/com/atproto/server/requestPasswordReset.ts
```diff
@@ -19,7 +19,10 @@ export default function (server: Server, ctx: AppContext) {
     handler: async ({ input, req }) => {
       const email = input.body.email.toLowerCase()
 
-      const account = await ctx.accountManager.getAccountByEmail(email)
+      const account = await ctx.accountManager.getAccountByEmail(email, {
+        includeDeactivated: true,
+        includeTakenDown: true,
```
devinivy, line 24: Adding support for taken-down accounts to reset their password, update their email.  Just double-checking we like that 👍

packages/pds/src/api/com/atproto/server/util.ts
```diff
@@ -40,3 +42,71 @@ export const didDocForSession = async (
...
+
+  const plcRotationKey =
+    ctx.cfg.entryway?.plcRotationKey ?? ctx.plcRotationKey.did()
+  if (rotationKeys !== undefined && !rotationKeys.includes(plcRotationKey)) {
+    throw new InvalidRequestError(
+      'Server rotation key not included in PLC DID data',
+    )
+  }
+
+  if (!pdsEndpoint || pdsEndpoint !== ctx.cfg.service.publicUrl) {
```
devinivy, line 100: If trailing slashes aren't significant, we might consider permitting the pds endpoint both with and without the tailing slash.

packages/pds/src/auth-verifier.ts
```diff
@@ -219,24 +229,39 @@ export class AuthVerifier {
...
+
+  userDidAuthOptional = async (
+    reqCtx: ReqCtx,
+  ): Promise<UserDidOutput | NullOutput> => {
+    if (isBearerToken(reqCtx.req)) {
+      try {
+        return await this.userDidAuth(reqCtx)
+      } catch {
+        return { credentials: null }
+      }
```
devinivy, line 254: I'd recommend we require the auth be valid if it's present.  It's conservative on the server side, but also usually friendly to the caller too, since if they included auth they probably intended to take action in the context of those credentials.  They would also be aware that they were holding onto some bad credentials.

packages/pds/src/auth-verifier.ts
```diff
@@ -219,24 +229,39 @@ export class AuthVerifier {
...
+    } else {
+      return { credentials: null }
+    }
+  }
+
+  adminService = async (reqCtx: ReqCtx): Promise<AdminServiceOutput> => {
+    const payload = await this.verifyServiceJwt(reqCtx, {
+      aud: this.dids.entryway ?? this.dids.pds,
+      iss: [this.dids.admin],
```
devinivy, line 263: In a world where pdses have dids, I wonder if we'll eventually want service auth to include the pds as the issuer and the user as the subject.  Not relevant to this PR, just wanted to preserve the thought before it slips away!

packages/pds/src/config/env.ts
```diff
@@ -9,6 +9,7 @@ export const readEnv = (): ServerEnvironment => {
     version: envStr('PDS_VERSION'),
     privacyPolicyUrl: envStr('PDS_PRIVACY_POLICY_URL'),
     termsOfServiceUrl: envStr('PDS_TERMS_OF_SERVICE_URL'),
+    acceptingImports: envBool('PDS_ACCEPTING_IMPROTS'),
```
devinivy, line 12:
```suggestion
    acceptingImports: envBool('PDS_ACCEPTING_IMPORTS'),
```

devinivy: If we think there could be different kinds of imports in the future, could preemptively make it something like `PDS_ACCEPTING_REPO_IMPORTS`.

packages/pds/tests/account-migration.test.ts
```diff
@@ -0,0 +1,215 @@
...
+      password: sc.accounts[alice].password,
+    })
+  })
+
+  afterAll(async () => {
+    await newPds.close()
+    await network.close()
+  })
+
+  it('migrates an account', async () => {
```
devinivy, line 78: 🔥

packages/pds/tests/account-migration.test.ts
```diff
@@ -0,0 +1,215 @@
...
+      validDid: true,
+      indexedRecords: 103,
+      privateStateValues: 0,
+      expectedBlobs: 3,
+      importedBlobs: 3,
+    })
+
+    await oldAgent.com.atproto.server.deactivateAccount({
+      deleteAfter: new Date(Date.now() + HOUR).toISOString(),
+    })
```
devinivy, line 197: What does this part test—that you can write even after marked for deactivation?

packages/pds/src/api/com/atproto/server/util.ts
devinivy: (If we pursue this, similar logic also appears in `identity.submitPlcOp`.)

devinivy (commented):
What a dream to see this working, it came out real nice 🤤

packages/pds/src/actor-store/repo/sql-repo-reader.ts
dholms: Since it's only used for reads, I don't think it would block, but it might fail with a `SQLITE_BUSY` and have to retry

packages/pds/src/api/com/atproto/repo/uploadBlob.ts
dholms: That's a good point, I'll come back to this so that it doesn't hold the txn open while uploading 👍

packages/pds/src/api/com/atproto/server/activateAccount.ts
dholms: umm yes good catch 👍

packages/pds/src/api/com/atproto/repo/importRepo.ts
dholms: yeah that's a good point 👍 

we need a better CAR parsing lib first 😅

packages/pds/src/api/com/atproto/repo/importRepo.ts
dholms: yup yup 👍

packages/pds/src/api/com/atproto/server/confirmEmail.ts
dholms: ah right 👍

packages/pds/src/api/com/atproto/server/requestPasswordReset.ts
dholms: Removed it from update email, but I do think it's import for resetting your password 👍 

You may want to reset your password in order to export some of your private data or get a signed PLC operation for instance

packages/pds/src/auth-verifier.ts
dholms: Yup that's fair 👍

packages/pds/src/auth-verifier.ts
dholms: yup I could totally see that 👍

packages/pds/src/config/env.ts
dholms: yup i'm down 👍

packages/pds/tests/account-migration.test.ts
dholms: I really just added it in order to finish documentation of the flow

I'll add an additional test to just check the status on the old pds 👌

packages/pds/src/api/com/atproto/repo/importRepo.ts
dholms: oof whatdya know 🙃 

i'll fix this up

dholms: probably worth a quick peek: https://github.com/bluesky-social/atproto/pull/2179/commits/153c90b145072db1a8853ff87ace414ce4f61adf

packages/pds/src/actor-store/repo/sql-repo-reader.ts
devinivy: Ah sorry didn't explain that well!  I was picturing blocking the main app thread.  E.g. if there's a second of work in sqlite to do here, the app thread doing the work will block for a second.

packages/pds/src/api/com/atproto/repo/importRepo.ts
devinivy: Looks good 👍

packages/pds/src/api/com/atproto/server/util.ts
dholms: I'm inclined to just leave it more conservative for now, and we can make it more lax over time if we decide to. Easier to go that way than the other 🤔

packages/pds/src/api/com/atproto/repo/importRepo.ts
dholms: yeah hmmm. is there a way that we can do both? like process it synchronously in most cases, but drop into async if it's large enough? Not sure how that would surface in schemas 🤔

packages/pds/src/api/com/atproto/repo/importRepo.ts
dholms: yup i gave it a test & it worked how you'd expect 👍

PR #2182, opened 2024-02-14
atproto data model interop test files

bnewbold:
These are copied over from the golang implementation (indigo).

I can see where to hook in the generic IPLD JSON/CBOR conversion code for testing, but not where to test validation of records when lexicon is unknown.

I also thought I had some blob-extraction tests in this format, but guess not.


bnewbold:
plan is to move these over to the new interop test repo: https://github.com/bluesky-social/atproto-interop-tests

PR #2185, opened 2024-02-15
Entryway account migration

dholms:


packages/pds/src/api/com/atproto/identity/requestPlcOperationSignature.ts
```diff
@@ -0,0 +1,26 @@
...
+export default function (server: Server, ctx: AppContext) {
+  server.com.atproto.identity.requestPlcOperationSignature({
+    auth: ctx.authVerifier.accessNotAppPassword,
+    handler: async ({ auth }) => {
+      const did = auth.credentials.did
+
+      const account = await ctx.services.account(ctx.db).getAccount(did, true)
+      if (!account) {
+        throw new InvalidRequestError('account not found')
+      }
```
devinivy, line 14:
```suggestion
```

packages/pds/src/api/com/atproto/identity/signPlcOperation.ts
```diff
@@ -0,0 +1,46 @@
...
+      const lastOp = await ctx.plcClient.getLastOp(did)
+      if (check.is(lastOp, plc.def.tombstone)) {
+        throw new InvalidRequestError('Did is tombstoned')
+      }
+      const operation = await plc.createUpdateOp(
+        lastOp,
+        ctx.plcRotationKey,
+        (lastOp) => ({
+          ...lastOp,
+          rotationKeys: input.body.rotationKeys ?? lastOp.rotationKeys,
```
devinivy, line 31: Just making sure it doesn't make sense to do the `lastOp` tombstone check inside the callback here.

packages/pds/src/api/com/atproto/identity/signPlcOperation.ts
```diff
@@ -0,0 +1,46 @@
...
+          verificationMethods:
+            input.body.verificationMethods ?? lastOp.verificationMethods,
+          services: input.body.services ?? lastOp.services,
+        }),
+      )
+      return {
+        encoding: 'application/json',
+        body: {
+          operation,
+        },
```
devinivy, line 42: Just an idea—we could add a field to this response that indicates whether submitting the op would result in the service no longer being able to submit ops on your behalf.

packages/pds/src/api/com/atproto/server/getServiceAuth.ts
```diff
@@ -0,0 +1,28 @@
...
+  server.com.atproto.server.getServiceAuth({
+    auth: ctx.authVerifier.accessCheckTakedown,
+    handler: async ({ auth, params, req }) => {
+      const proxied = await proxy(
+        ctx,
+        auth.credentials.audience,
+        async (agent) => {
+          const result = await agent.api.com.atproto.server.getServiceAuth(
+            params,
+            authPassthru(req, true),
```
devinivy, line 16: Can't really hurt, but I don't think this param needs to be passed.
```suggestion
            authPassthru(req),
```

devinivy (approved):
Excellent

packages/pds/src/api/com/atproto/identity/signPlcOperation.ts
dholms: yup checks in the right spot, but i updated the code to use `prev` in the callback to avoid confusion

packages/pds/src/api/com/atproto/identity/signPlcOperation.ts
dholms: I'm not opposed to it but i don't think it's necessary. If you're interested in having those checks, you can always use `submitPlcOperation`

PR #2189, opened 2024-02-16
Setup alternative self-verification in account creation

devinivy:
Additional support for out-of-band verification code in place of phone verification.


dholms (approved):
this all looks great to me 👌

PR #2195, opened 2024-02-19
Muted items prefs

estrattonbailey:
Adds prefs for:
- muted words & tags
- hidden posts

New methods:
- `upsertMutedWords(words: MutedWord[])`
- `updateMutedWord(word: MutedWord)`
- `removeMutedWord(word: MutedWord)`
- `hidePost(postUri)`
- `unhidePost(postUri)`

### DIFF
--- a/lexicons/app/bsky/actor/defs.json
+++ b/lexicons/app/bsky/actor/defs.json
@@ -230,7 +230,7 @@
         "target": {
           "type": "string",
           "description": "The intended target of the muted word.",
-          "knownValues": ["content", "tag"],
+          "knownValues": ["content", "tag", "any"],
           "maxLength": 640,
           "maxGraphemes": 64
         }
--- a/lexicons/app/bsky/actor/defs.json
+++ b/lexicons/app/bsky/actor/defs.json
@@ -223,8 +223,8 @@
       "properties": {
         "value": {
           "type": "string",
-          "maxLength": 640,
-          "maxGraphemes": 64
+          "maxLength": 10000,
+          "maxGraphemes": 1000
         }
       }
     },
@@ -234,7 +234,6 @@
       "properties": {
         "items": {
           "type": "array",
-          "maxLength": 100,
           "items": {
             "type": "ref",
             "ref": "app.bsky.actor.defs#mutedWordOrTag"
@@ -249,7 +248,6 @@
       "properties": {
         "items": {
           "type": "array",
-          "maxLength": 100,
           "items": {
             "type": "ref",
             "ref": "app.bsky.actor.defs#mutedWordOrTag"
--- a/lexicons/app/bsky/actor/defs.json
+++ b/lexicons/app/bsky/actor/defs.json
@@ -215,6 +215,59 @@
           "description": "A list of tags which describe the account owner's interests gathered during onboarding."
         }
       }
+    },
+    "mutedWordOrTag": {
+      "type": "object",
+      "description": "A word or tag that the account owner has muted.",
+      "required": ["value"],
+      "properties": {
+        "value": {
+          "type": "string",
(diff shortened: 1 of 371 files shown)


lexicons/app/bsky/actor/defs.json
```diff
@@ -215,6 +215,70 @@
           "description": "A list of tags which describe the account owner's interests gathered during onboarding."
         }
       }
+    },
+    "mutedWordOrTag": {
```
dholms, line 219: is this defined as a separate ref (as opposed to just making `mutedWordsPref` an array of strings) so that it's open to adding more metadata/properties on to?

estrattonbailey: Yep exactly e.g. timed mutes.

pfrazee (approved):
LGTM

packages/api/src/bsky-agent.ts
```diff
@@ -609,3 +645,103 @@ async function updateFeedPreferences(
...
+        outer: for (const newItem of mutedWords) {
+          for (const existingItem of mutedWordsPref.items) {
+            if (existingItem.value === newItem.value) {
+              existingItem.targets =
+                action === 'upsert'
+                  ? Array.from(
+                      new Set([...existingItem.targets, ...newItem.targets]),
+                    )
+                  : newItem.targets
+              continue outer
```
dholms, line 675: i'll be honest, i never use this syntax. does this accomplish the same thing as a `break` in the inner loop? and if so could we use that?

packages/api/src/bsky-agent.ts
```diff
@@ -609,3 +645,103 @@ async function updateFeedPreferences(
...
+      } else if (action === 'remove') {
+        let i = mutedWordsPref.items.length
+
+        while (i--) {
+          const item = mutedWordsPref.items[i]
+
+          for (const word of mutedWords) {
+            if (item.value === word.value) {
+              if (word.targets.length) {
+                item.targets = word.targets
```
dholms, line 693: this seems a little bit strange 🤔 

so if you're current prefs are is something like `{ word: 'blah', targets: ['tag', 'content'] }`, then if you do a `remove {word: 'blah', targets: ['tag']}`, it'll update the value to `{word: 'blah', targets: ['tag']}`?

feels like a strange syntax to me 🤔

packages/api/src/bsky-agent.ts
```diff
@@ -609,3 +645,103 @@ async function updateFeedPreferences(
...
+
+          // never hit `continue`, add new item
+          if (action === 'upsert') {
+            mutedWordsPref.items.push(newItem)
+          }
+        }
+      } else if (action === 'remove') {
+        let i = mutedWordsPref.items.length
+
+        while (i--) {
```
dholms, line 687: this is a nit, but i don't love this syntax. we often use `mapDefined` for stuff like this & if the thing is going to be deleted then the callback fn just returns `undefined` 🤔

packages/api/src/bsky-agent.ts
estrattonbailey: Oh sure sure

packages/api/src/bsky-agent.ts
estrattonbailey: Actually just removed this, good catch, was a remnant of removing words with no targets. I decided to allow this bc it's simpler and allows to temp disable a word, basically.

packages/api/src/bsky-agent.ts
estrattonbailey: Oh yeah I hate this too, good to know about `mapDefined` 👍

dholms (approved):
looks good 👍

PR #2196, opened 2024-02-19
fix(xrpc-server): properly parse & process content-encoding

matthieusieben:
[content-encoding](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding) can contain multiple values (comma separated).

- The current implementation ignores any unknown value it received (including comma separated values). This PR generates an error when unknown algorithms are provided.
- The current implementation uses `createDeflate` instead of `createInflate` to decompress bytes encoded using the `deflate` algo. This PR changes that.
- The current implementation will allow `content-length` headers that are not a number. This PR ensures that the parsing of the number did not result in `NaN`.

### DIFF
--- a/packages/common/src/streams.ts
+++ b/packages/common/src/streams.ts
@@ -4,7 +4,21 @@ import {
   PassThrough,
   Transform,
   TransformCallback,
+  Duplex,
 } from 'stream'
+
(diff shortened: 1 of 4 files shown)


matthieusieben:
https://github.com/bluesky-social/atproto/pull/2464

PR #2197, opened 2024-02-20
:sparkles: Show author feed of takendown author to admins only

foysalit:
Porting over already implemented behavior from main.


devinivy (approved):
Sweet!

PR #2198, opened 2024-02-20
:sparkles: Move appealed subject's review state to escalated

foysalit:
Since we want appealed subjects to only show up on escalation queue, this PR sets the `reviewState` to `reviewEscalated`.


bnewbold (approved):
looks right to me!

devinivy (approved):

PR #2201, opened 2024-02-20
Appview v2: fix not-found and blocked uris in threads

devinivy:
Just a little mixup!  Would result in not-found and blocked posts looking like they have the uri of their parent in thread responses.


dholms (approved):
good catch 🙌

PR #2207, opened 2024-02-21
Emit handle updates on account state changes

dholms:
We're planning to spec & implement new `#identity` and `#account` events on the firehose, but for now, `#handle` is the only event that is treated as an identity refresh & passed through the relay, so we emit it on account state changes


devinivy (approved):

dholms:
Closing in favor of https://github.com/bluesky-social/atproto/pull/2208

PR #2208, opened 2024-02-21
Firehose Identity event

dholms:
Adds a new `#identity` to the firehose that prompts all downstream services to refresh their identity cache for the given DID.

This is implemented in lieu of the `#migrate` event and will eventually replace the `#handle` event as well.

In a future iteration, more information may be attached to the event such as the DID document.

### DIFF
--- a/packages/pds/src/sequencer/events.ts
+++ b/packages/pds/src/sequencer/events.ts
@@ -89,7 +89,7 @@ export const formatSeqIdentityEvt = async (
   }
   return {
     did,
-    eventType: 'handle',
+    eventType: 'identity',
     event: cborEncode(evt),
     sequencedAt: new Date().toISOString(),
   }
--- a/packages/pds/src/sequencer/events.ts
+++ b/packages/pds/src/sequencer/events.ts
@@ -81,6 +81,20 @@ export const formatSeqHandleUpdate = async (
   }
 }
 
+export const formatSeqIdentityEvt = async (
+  did: string,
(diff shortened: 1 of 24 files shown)


bnewbold:
I think the `#tombstone` event can also be deprecated, right? Or are we using that in an PDS-specific way right now? Would be good to clarify in the `#tombstone` description either way.

dholms:
Clarified in the `#tombstone` description 👍 

I don't think we can deprecate it just yet since it communicates some extra info ("i want you to delete this account") that isn't immediately discoverable from the DID/PDS

packages/pds/src/sequencer/events.ts
```diff
@@ -81,6 +81,20 @@ export const formatSeqHandleUpdate = async (
...
 
+export const formatSeqIdentityEvt = async (
+  did: string,
+): Promise<RepoSeqInsert> => {
+  const evt: IdentityEvt = {
+    did,
+  }
+  return {
+    did,
+    eventType: 'handle',
```
devinivy, line 92:
```suggestion
    eventType: 'identity',
```

dholms: yup that one was biting me 😅

devinivy (approved):

mackuba:
Hmm, looks like it doesn't include the handle data like the #handle event does, so I'm guessing I'm gonna have to start fetching handle info from somewhere when this new event replaces #handle events?

PR #2211, opened 2024-02-22
Optional service config in pds distribution

devinivy:
This loosens the config requirements of the pds distribution by allowing functionality provided by other services to be disabled:
 - configuring a bsky appview service is optional.
 - configuring a moderation service is optional.
 - a service specifically for receiving reports may be configured separately from the privileged moderation service.


dholms (approved):
sweet looks good 👍

packages/pds/src/auth-verifier.ts
```diff
@@ -1,3 +1,4 @@
+import assert from 'node:assert'
```
devinivy, line 1:
```suggestion
```

PR #2218, opened 2024-02-22
Fix sanitization, improve test to catch

estrattonbailey:
Missed a sanitization and the test didn't catch due to the order the calls were made. Sanitized, updated test: should only insert one entry for `hashtag`.


pfrazee (approved):
LGTM

PR #2227, opened 2024-02-25
pds: allow root domains to be handles

iameli:
I'm running a PDS at `https://iame.li` and I want my handle to be `@iame.li`. There's presently a check in `com.atproto.identity.resolveHandle` that disallows any domain in our `serviceHandleDomains` to be resolved as a handle. I'm not sure there's any virtue to this check, and I assume that PDS operators running PDS on the same domain as their handles will be something that comes up.

Naturally, you still need to prove that you own the domain through the regular TXT record/.well-known entry, so there shouldn't be any harm here.

This PR removes this check and adds tests to validate the other cases. I've tested it on my own PDS and my handle is once again `@iame.li`. Hooray!

### DIFF
--- a/packages/pds/src/api/com/atproto/identity/resolveHandle.ts
+++ b/packages/pds/src/api/com/atproto/identity/resolveHandle.ts
@@ -24,7 +24,7 @@ export default function (server: Server, ctx: AppContext) {
       did = user.did
     } else {
       const supportedHandle = ctx.cfg.identity.serviceHandleDomains.some(
-        (host) => handle.endsWith(host) || handle === host.slice(1),
+        (host) => handle.endsWith(host),
       )
       // this should be in our DB & we couldn't find it, so fail
       if (supportedHandle) {
--- a/packages/pds/tests/handles.test.ts
+++ b/packages/pds/tests/handles.test.ts
@@ -7,6 +7,8 @@ import { AppContext } from '../src'
 // outside of suite so they can be used in mock
 let alice: string
 let bob: string
+let carol: string
+let dan: string
 
 jest.mock('dns/promises', () => {
   return {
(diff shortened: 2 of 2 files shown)


ledlamp:
We should also allow initial creation of the handle to be consistent: https://github.com/bluesky-social/atproto/issues/2267#issuecomment-1975019672

iameli:
@ledlamp That makes sense - do you know if that uses a different code path or whether that's already handled by my change?

ledlamp:
@iameli It's a different error message, I find it here: https://github.com/bluesky-social/atproto/blob/dc7564be01e93bf12fcaf7ecd3668b705dc06636/packages/pds/src/handle/index.ts#L30-L43
https://github.com/bluesky-social/atproto/blob/dc7564be01e93bf12fcaf7ecd3668b705dc06636/packages/pds/src/handle/index.ts#L65-L70

We probably need something to allow only admin to create this handle

iameli:
@ledlamp That's interesting, could dig into that. But if my change lands, you're at least allowed to change to a root handle after the fact.

> We probably need something to allow only admin to create this handle

I'm not sure I understand why. If a user _isn't_ an admin, why are they allowed to create a `.well-known` record or DNS TXT entry to prove ownership of this domain? In fact, after they do so, they'll be able to successfully claim the handle on any _other_ PDS, just not the one that lives on the root.

ledlamp:
because the pds domain is pointed at the pds, it will verify automatically, /.well-known/atproto-did will serve did for whoever has the handle with req.hostname.

so if it's enabled as a service domain then any user could claim it if it's not already

iameli:
> so if it's enabled as a service domain then any user could claim it if it's not already

Hmm, let me set up a test PDS and check whether this happens

PR #2229, opened 2024-02-26
Add `parentAuthor` to `feedViewPost`

haileyok:
Prefacing this with this is the first time I've really dug at this repo, definitely might have made a mistake but hopefully it's a small enough change where there wasn't much room to. Also, feel free to hate `parentAuthor`, normally I'd ask and figure that out beforehand but since this was kind of born out of "let me play with this thing..." figured I'd just push up and we can discuss.

Also, no idea if this accounts for v2 changes. If it doesn't, hopefully it at least serves as an example of what we'd like to see here.

## Why

Right now the `feedViewPost` returns a post and it's parent `postView` in the `feedViewPost`'s reply ref. However, there is no information given for the parent's parent (parent 2, we'll call it). There's very little information that we need for parent 2, but we do need to have the handle/display name of parent 2's author. This lets us display the "Reply to..." label on the post.

As a result, right now in the app (and most third party apps), this requires us making a separate query for `getProfile`, as seen below. Especially in the case of third party clients, this seems to have been a headscratcher for multiple people. And, while our client is doing some caching to limit the number of times we query `getProfile`, it's likely that not every client is doing so.

Screenshot 2024-02-26 at 1 39 45 AM

The `getProfile` calls add up quickly, and easily outnumber the requests we are making for the timeline itself. Especially on initial loads, the number of `getProfile` queries can add up to around 15-20. Each subsequent page that we load from `getTimeline` or `getFeed` usually has ~5 queries to `getProfile` that we have to make.

While we could alternatively batch the `getProfile` requests into a single `getProfiles` request, this doesn't eliminate the UX downside to having to query for this data separately. There's a brief period of time (brief on a good connection, a weaker one would have a fairly significant delay in displaying this additional data) where we can't even display the text for the "Reply to..." label.

## How

It seems the simplest way that we can add this information is done in three steps.

1. Create a separate set of post URIs of `replyParent`s when we create the `feedItemRefs`. We'll call this `parentUris`.
2. After getting all of the post infos in `feedHydration`, looping through this `parentUris`  and matching them to their respective post info. There, we can get the parent's reply ref (if there is one), and get the additional DID we need to look up. We add that DID to the `dids` set that is used after getting post infos.
3. Update the `feedViewPost` lexicon and add a `parentAuthor` field, which is a `profileViewBasic` of parent 2's author.

Screenshot 2024-02-26 at 1 31 50 AM

### DIFF
--- a/lexicons/app/bsky/feed/defs.json
+++ b/lexicons/app/bsky/feed/defs.json
@@ -49,6 +49,10 @@
       "properties": {
         "post": { "type": "ref", "ref": "#postView" },
         "reply": { "type": "ref", "ref": "#replyRef" },
+        "replyParentAuthor": {
+          "type": "ref",
+          "ref": "app.bsky.actor.defs#profileViewBasic"
+        },
         "reason": { "type": "union", "refs": ["#reasonRepost"] }
       }
     },
--- a/packages/api/src/client/lexicons.ts
+++ b/packages/api/src/client/lexicons.ts
@@ -6001,6 +6001,10 @@ export const schemaDict = {
             type: 'ref',
             ref: 'lex:app.bsky.feed.defs#replyRef',
           },
+          replyParentAuthor: {
+            type: 'ref',
+            ref: 'lex:app.bsky.actor.defs#profileViewBasic',
+          },
           reason: {
             type: 'union',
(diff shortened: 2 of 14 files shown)


devinivy:
Replaced by #2461.  But this PR serves as a great explanation of the issue being addressed.

PR #2234, opened 2024-02-27
Entryway: support jwe self-verification tokens

devinivy:
Adds support for encrypted jwts as self-verification tokens during account creation.


packages/pds/tests/phone-verification.test.ts
```diff
@@ -181,114 +185,151 @@ describe('phone verification', () => {
   })
 
   describe('self-verification', () => {
```
dholms, line 187: maybe a test in here for a poorly signed JWE? like encrypting/signing it with the wrong key?

dholms (approved):
sweet this all looks good to me 👍

packages/pds/tests/phone-verification.test.ts
devinivy: Done deal!

PR #2235, opened 2024-02-27
Introduce #reviewOptional as reviewState for non-impactful events on a subject

foysalit:
Up until now, emitting events like `modEventComment` or `modEventTag` would set the `reviewState` of the subject to `reviewOpen` which is a bit disruptive to the moderation queue since these events are meant for setting optional metadata on the subject and should not require a moderator's review. This PR introduces a new `reviewOptional` state which will be set when such events are emitted on a subject with no prior (impactful) events.

**TODO:**
This means for events with type `modEventLabel`, we will no longer set the subject's status to `reviewOpen` so if tools like automod wants a labeled subject to be reviewd by a moderator, it will have to emit a separate `modEventReport` event.


bnewbold (approved):
I don't love the name `#reviewOptional`. Maybe `#reviewNone` or "empty" or something like that?

Automod already does label + create review in cases where that is needed or appropriate; it will also be good to be able to apply labels *without* requiring review. This is basically just parity with how image auto-labeling works (IIRC).

foysalit:
I don't have strong opinion on the naming so happy to rename to `reviewNone`.
yeah, labels events will now only set to `reviewNone` if no previous state is set.

packages/ozone/tests/__snapshots__/moderation-events.test.ts.snap
```diff
@@ -22,7 +22,7 @@ Object {
         "lastReportedAt": "1970-01-01T00:00:00.000Z",
         "lastReviewedAt": "1970-01-01T00:00:00.000Z",
         "lastReviewedBy": "user(1)",
-        "reviewState": "com.atproto.admin.defs#reviewEscalated",
+        "reviewState": "com.atproto.admin.defs#reviewOpen",
```
devinivy, line 25: This change stood out to me a little—just making sure we can account for the change in behavior.

devinivy (approved):
Had one question, but looks good!

packages/ozone/tests/__snapshots__/moderation-events.test.ts.snap
foysalit: great catch! as discussed in slack, this was caused by a race condition and I missed the update in snapshot.
pushed a fix by adding `forUpdate()` to the current status selector query which should be safe since we shouldn't have that many concurrent events on the same subject.

PR #2236, opened 2024-02-28
Ozone: Add sigs to all labels

dholms:
Adds optional signature to label schema.

Ozone signs all labels with signing key as they are created and stored.

When returning labels, it does a quick check during formatting to ensure they're signed with the proper key. If they are not signed correctly, then we resign, send the resigned label, and update the signature in the database.


bnewbold:
It would be good to clarify the base64 variant in the description. I assume it is the same as for `$bytes`, so "standard" (not URL-safe), and without padding (or padding optional).

packages/ozone/src/db/migrations/20240228T003647759Z-add-label-sigs.ts
```diff
@@ -0,0 +1,17 @@
+import { Kysely, sql } from 'kysely'
+
+export async function up(db: Kysely<unknown>): Promise<void> {
+  await db.schema
+    .alterTable('label')
+    .addColumn('sig', sql`bytea`)
+    .execute()
+  await db.schema
+    .alterTable('label')
+    .addColumn('signingKey', 'varchar')
```
devinivy, line 10: It's not a big issue, but if it's easy to avoid storing the full signing key with each label that would be pretty nice.

One way it could work: during startup the public signing key could be upserted into a table and exchanged for an id.  Then the labels table stores the id.  You never really need to join over to that table, I don't think, just used once at startup.  Definitely not necessary, but figured I'd toss it out there!

packages/ozone/src/mod-service/views.ts
```diff
@@ -408,7 +417,28 @@ export class ModerationViews {
...
+      try {
+        await db.db
+          .updateTable('label')
+          .set({ sig: Buffer.from(signed.sig), signingKey })
+          .where('id', '=', row.id)
+          .execute()
+      } catch (err) {
+        dbLogger.error({ err, label: row }, 'failed to update resigned label')
+      }
+    })
```
devinivy, line 440: If someone resyncs all labels after changing your signing key, things could get a bit interesting here.  If that's a concern, I think the most at-hand thing to do would be to introduce some batching.

packages/ozone/tests/_util.ts
```diff
@@ -61,6 +62,10 @@ export const forSnapshot = (obj: unknown) => {
       const [, did, cid] = match
       return str.replace(did, take(users, did)).replace(cid, take(cids, cid))
     }
+    // decent check for 64-byte base64 encoded signatures
+    if (str.length === 86 && !str.includes(' ')) {
+      return take(sigs, str)
+    }
```
devinivy, line 68: Nice 😎

packages/ozone/src/mod-service/views.ts
dholms: actually yeah i like the idea of just doing it inline for back pressure - we'll still catch/log the error & send out the signed label if this fails for some reason

packages/ozone/src/db/migrations/20240228T003647759Z-add-label-sigs.ts
dholms: yeah that's a good point - this is unnecessarily bulky

i'll take a stab at it 👌

PR #2237, opened 2024-02-28
Add custom label definitions and tune various behaviors

pfrazee:
This PR updates the `3p-labelers` with custom label definitions and some simplifications to labeler declarations and label management. This PR leaves multiple TODOs but it's at a good stopping-point to review the approach being taken.

- Adds custom label definitions
  - Added `com.atproto.label.defs#labelValueDefinition`
  - Added `labelerDid?: string` to the label preference
  - Added tooling for custom label definitions to be passed into the moderate methods (unfortunately required another change to the ModerationOpts signature)
- Drops moderation.getService and updates the batch method to support a `detailed` flag (still needs backend implementation)
- Drops the concept of label-groups entirely (we now configure label preferences directly)
- Dramatically reduces the set of globally-defined labels


lexicons/app/bsky/moderation/defs.json
```diff
@@ -64,29 +46,22 @@
     },
     "modServicePolicies": {
       "type": "object",
-      "required": ["reportReasons", "labelValues"],
```
pfrazee, line 67: I'm dropping the custom report reasons for now, we'll FF with it but I think we gotta eject for now

lexicons/app/bsky/moderation/defs.json
```diff
@@ -64,29 +46,22 @@
...
-          "maxLength": 100000
-        },
-        "descriptionFacets": {
-          "type": "array",
-          "items": { "type": "ref", "ref": "app.bsky.richtext.facet" }
-        },
-        "reportReasons": {
+        "labelValues": {
           "type": "array",
+          "description": "The label values which this labeler publishes. Be sure to prefix custom labels with 'x-'.",
```
dholms, line 53: just to check - we're doing away with both descriptions?

PR #2239, opened 2024-02-29
:bug: Increment attempt count after each attempt to push ozone event

foysalit:


dholms (approved):
that looks right 👍 

thanks for snagging that

devinivy (approved):
Great catch!

PR #2240, opened 2024-02-29
Fix muted word sanitization

estrattonbailey:
`##hashtag` is a valid tag based on the parsing we shipped ~5 months ago. Today, [users noticed](https://bsky.app/profile/josiah.lol/post/3kmjejr4fux26) that muting `##` added a muted word to their prefs (`#`), great, but they then could not remove.

This is because `##` was being truncated down to `#` by this sanitization. Then, when trying to remove, it was further being truncated down to an empty string, hence no match, hence no removal.

So this PR fixes support for tags like `##hashtag` AND allows removal of existing muted words only containing `#` **by only sanitizing on insert.** On update/remove, we expect an exact match.

I re-split these CRUD actions out and updated some naming for clarity.

Update: because it was closely related, I added `#️⃣` and `##️⃣` handling as well.

### DIFF
--- a/.changeset/rotten-actors-dance.md
+++ b/.changeset/rotten-actors-dance.md
@@ -1,0 +1,5 @@
+---
+'@atproto/api': patch
+---
+
+Fix double sanitization bug when editing muted words.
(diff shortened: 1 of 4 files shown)


pfrazee (approved):

estrattonbailey:
More changes to come, but will continue to do in separate PRs

PR #2242, opened 2024-02-29
Do not parse hashtag emoji as tag

mary-ext:
Closes #2243

Fixes the issue where #️⃣ is being parsed as an emoji because it starts with the same char code as regular hashtag, as the emoji is composed of 0x23 (#) 0xfe0f (variation selector) 0x20e3 (keycap)

Fixes length calculation as well

### DIFF
--- a/.changeset/dull-hotels-beam.md
+++ b/.changeset/dull-hotels-beam.md
@@ -1,0 +1,5 @@
+---
+'@atproto/api': patch
+---
+
+Prevent hashtag emoji from being parsed as a tag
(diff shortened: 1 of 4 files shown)


mary-ext:
cc @estrattonbailey

mary-ext:
also you can use 0x200b (zero width space) to allow for tags with seemingly no content, curious if that should be fixed as well

mary-ext:
if we do agree on having hashtag emoji-based tags, we should at least normalize it so that #tag and :hash:tag points to the same tag

mary-ext:
also despite the 64 char limit, the parser and the tests checks if it's over 65 chars instead, which seems wrong

estrattonbailey (approved):
Thank youuuuuu 🙏 looks great, gonna do a lil testing

PR #2243, opened 2024-02-29
Hashtag fixes

mary-ext:
Closes #2242

Alternate fix to #2242, where hashtag emojis are normalized, along with some other change like having it check if it's more than 64 characters rather than 65


estrattonbailey:
Ah I broke the auto close logic bc of how I did this, my bad. Thanks for pushing these up!

PR #2251, opened 2024-02-29
Add client-set proxy headers to PDS

dholms:
Add the ability to set an `atproto-proxy` header to a request to inform a PDS where it should forward that message to (with service auth). Currently only works for specced methods.

The syntax is `atproto-proxy: ${did}#${serviceId}`. For instance: `atproto-proxy: did:web:api.bsky.app#bsky_appview`.

When unset, the request falls back to the current PDS configuration


ShreyanJain9:
Nice, so this will allow clients to easily hook into custom AppViews for non-Bluesky use-cases? 👀

dholms:
Yup yup exactly! This is actually needed for the composable mod services feature we're working on, but it will also allow a client to configuring a different appview to communicate with. Unfortunately it still only works for `app.bsky` methods, but soon we plan to make it work generically for any method 👌

packages/pds/src/pipethrough.ts
```diff
@@ -1,22 +1,131 @@
...
 import { httpLogger } from './logger'
-import { noUndefinedVals } from '@atproto/common'
+import { getServiceEndpoint, noUndefinedVals } from '@atproto/common'
+import AppContext from './context'
+
+const defaultService = (
+  ctx: AppContext,
+  path: string,
+): { url: string; did: string } | null => {
+  const nsid = path.replace('/xrpc/', '')
```
devinivy, line 16: A few misc things to maybe just peek at here:
 - Do we need to check that the path starts with `/xrpc` too?
 - May want to handle trailing slash in the URL, I believe the pds routing permits it.
 - Express's `req.url` is relative to the nearest router or something like that, I forget the details but may be relevant since I know it gets passed here.  There is also `req.originalUrl` which might be more reliable for us.

PR #2252, opened 2024-02-29
Ozone ACLs

dholms:
Switches out basic auth for ACLs in Ozone backend.

- list of moderators & triagers in ozone env
- appview & pds/entryway are configured with a modServiceDid for all moderation routes
  - retain (narrowly distributed) admin tokens on appview & pds as an escape hatch
- all requests that go to Ozone travel through the user's PDS
  - PDS signs a service auth token for the user & sends it to ozone backend (similar to appview auth)
  - Ozone will then handle the request or sign another request out to the relevant service using it's own service DID

Supersedes https://github.com/bluesky-social/atproto/pull/2058 but broken up such that it does not introduce breaking changes


packages/ozone/src/auth-verifier.ts
```diff
@@ -0,0 +1,218 @@
...
+    iss: string
+    isAdmin: boolean
+    isModerator: boolean
+    isTriage: boolean
+  }
+}
+
+type NullOutput = {
+  credentials: {
+    type: 'null'
```
devinivy, line 43: I struggle a little bit with the string `'null'` here, mostly because I wonder whether it's a typo when I read checks like `credentials.type === 'null'`.

I like keeping a string here so there's a value to discriminate credentials on.  This covers the case where the user did not bring any credentials: would `'none'` work?

packages/bsky/src/auth-verifier.ts
```diff
@@ -159,19 +167,19 @@ export class AuthVerifier {
     }
   }
 
-  adminService = async (reqCtx: ReqCtx): Promise<AdminServiceOutput> => {
+  modService = async (reqCtx: ReqCtx): Promise<ModServiceOutput> => {
     const { iss, aud } = await this.verifyServiceJwt(reqCtx, {
       aud: this.ownDid,
-      iss: [this.adminDid],
+      iss: [this.modServiceDid, `${this.modServiceDid}#atproto_mod`],
```
devinivy, line 173: I think `#atproto_mod` is the id of the key.  Do we want to use the id of the service here, map it to the id of the key when verifying?

packages/ozone/src/auth-verifier.ts
```diff
@@ -0,0 +1,218 @@
...
+    iss: null
+  }
+}
+
+export type AuthVerifierOpts = {
+  serviceDid: string
+  admins: string[]
+  moderators: string[]
+  triage: string[]
```
devinivy, line 52: I have a hunch we'll end-up wanting to toss these in the database before long, but seems like this is the right place to start 👍

packages/pds/src/auth-verifier.ts
```diff
@@ -253,28 +254,28 @@ export class AuthVerifier {
...
-  adminService = async (reqCtx: ReqCtx): Promise<AdminServiceOutput> => {
-    if (!this.dids.admin) {
+  modService = async (reqCtx: ReqCtx): Promise<ModServiceOutput> => {
+    if (!this.dids.modService) {
       throw new AuthRequiredError('Untrusted issuer', 'UntrustedIss')
     }
     const payload = await this.verifyServiceJwt(reqCtx, {
       aud: this.dids.entryway ?? this.dids.pds,
-      iss: [this.dids.admin],
+      iss: [this.dids.modService, `${this.dids.modService}#atproto_mod`],
```
devinivy, line 263: We might consider using the service id rather than the key id for the issuer here.

packages/pds/tests/moderator-auth.test.ts
```diff
@@ -1,43 +1,62 @@
...
+    // kinda goofy but we need to know the dids before creating the testnet for the PDS's config
+    modServiceKey = await Secp256k1Keypair.create()
+    const modServiceInfo = await opAndDid('mod.test', modServiceKey)
+    const altModInfo = await opAndDid('alt-mod.test', modServiceKey)
+    modServiceDid = modServiceInfo.did
+    altModDid = altModInfo.did
+
     network = await TestNetworkNoAppView.create({
-      dbPostgresSchema: 'pds_admin_auth',
+      dbPostgresSchema: 'pds_moderator_auth',
```
devinivy, line 48: I appreciate the rebrand from "admin" to "moderator" 👍

devinivy (approved):
I think the main thing to nail down here is whether the credential issuer should be the service id rather than the key id.  I am pro-service-id but we can chat it through!  Pumped to have the auth story in a much better state!

packages/bsky/src/auth-verifier.ts
```diff
@@ -236,16 +251,19 @@ export class AuthVerifier {
...
+    creds: StandardOutput | RoleOutput | ModServiceOutput | NullOutput,
   ) {
     const viewer =
       creds.credentials.type === 'standard' ? creds.credentials.iss : null
     const canViewTakedowns =
       (creds.credentials.type === 'role' && creds.credentials.admin) ||
-      creds.credentials.type === 'admin_service'
+      creds.credentials.type === 'mod_service' ||
+      (creds.credentials.type === 'standard' &&
+        this.isModService(creds.credentials.iss))
```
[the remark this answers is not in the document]
dholms: Yup that does sound right - maybe it could be communicated better.

But the latter comes into play on some of the view routes where we still want to show taken down content - `getProfile` for instance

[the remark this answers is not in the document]
dholms: We could use a separate authVerifier there like `optionalStandardOrModService` but I thought it was a bit clearer to just use `optionalStandard` and then do a follow on check to see if it came from a trusted did

packages/ozone/src/auth-verifier.ts
dholms: i like that 👍

packages/bsky/src/auth-verifier.ts
dholms: Yup we chatted about this - I changed this to be the serviceId to keep it inline with possible future updates 👍

packages/ozone/src/auth-verifier.ts
dholms: Yup yup agreed

packages/pds/src/auth-verifier.ts
dholms: yup done 👍

packages/bsky/src/auth-verifier.ts
```diff
@@ -195,12 +203,15 @@ export class AuthVerifier {
...
+      iss: string,
       _forceRefresh: boolean, // @TODO consider propagating to dataplane
     ): Promise<string> => {
-      if (opts.iss !== null && !opts.iss.includes(did)) {
+      if (opts.iss !== null && !opts.iss.includes(iss)) {
         throw new AuthRequiredError('Untrusted issuer', 'UntrustedIss')
       }
+      const [did, serviceId] = iss.split('#')
+      const keyId =
+        serviceId === 'atproto-labeler' ? 'atproto-label' : 'atproto'
```
devinivy, line 214:
```suggestion
        serviceId === 'atproto_labeler' ? 'atproto-label' : 'atproto'
```

dholms: good snag 👍

PR #2255, opened 2024-03-01
Clean up role-based auth

dholms:
Follow up to https://github.com/bluesky-social/atproto/pull/2252 that deprecates the no longer need role-based basic auth


packages/pds/tests/proxied/admin.test.ts
```diff
@@ -9,11 +9,13 @@ import { forSnapshot } from '../_util'
 import { NotFoundError } from '@atproto/api/src/client/types/app/bsky/feed/getPostThread'
 
 // @TODO skipping during appview v2 buildout, as appview frontends no longer contains moderation endpoints
```
devinivy, line 11:
```suggestion
```

PR #2260, opened 2024-03-01
Improve tag detection

estrattonbailey:
This adds a few things that weren't here:
- support for leading numbers
- ~support for only numbers~
- disallows zero-width space characters and other hidden characters


packages/api/tests/rich-text-detection.test.ts
```diff
@@ -319,9 +325,12 @@ describe('detectFacets', () => {
           },
         ],
       ],
+      ['this #​ should not be a tag', [], []],
+      ['this #​a should not be a tag', [], []],
```
pfrazee, line 329: Why isnt' this a tag? is there a zerowidth character there?

packages/api/tests/rich-text-detection.test.ts
```diff
@@ -319,9 +325,12 @@ describe('detectFacets', () => {
...
+      ['this #a​b should be a tag', ['a'], [{ byteStart: 5, byteEnd: 7 }]],
```
pfrazee, line 330: Same question, I'm guessing there's a zero-width char in there?

packages/api/tests/rich-text-detection.test.ts
```diff
@@ -233,12 +233,12 @@ describe('detectFacets', () => {
...
         ],
       ],
-      ['#1', [], []],
+      ['#1', ['1'], [{ byteStart: 0, byteEnd: 2 }]],
       ['#tag', ['tag'], [{ byteStart: 0, byteEnd: 4 }]],
       ['body #tag', ['tag'], [{ byteStart: 5, byteEnd: 9 }]],
       ['#tag body', ['tag'], [{ byteStart: 0, byteEnd: 4 }]],
       ['body #tag body', ['tag'], [{ byteStart: 5, byteEnd: 9 }]],
-      ['body #1', [], []],
+      ['body #1', ['1'], [{ byteStart: 5, byteEnd: 7 }]],
```
pfrazee, line 241: This actually shouldnt be a tag

packages/api/tests/rich-text-detection.test.ts
```diff
@@ -233,12 +233,12 @@ describe('detectFacets', () => {
           { byteStart: 3, byteEnd: 5 },
         ],
       ],
-      ['#1', [], []],
+      ['#1', ['1'], [{ byteStart: 0, byteEnd: 2 }]],
```
pfrazee, line 236: This shouldnt be a tag

pfrazee (changes requested):
Added comments, also need a test case for tags that start with numbers (`#123abc`)

packages/api/tests/rich-text-detection.test.ts
estrattonbailey: Yep yep GitHub doesn't render them

packages/api/tests/rich-text-detection.test.ts
estrattonbailey: Same same

packages/api/src/rich-text/util.ts
```diff
@@ -0,0 +1,6 @@
+export const MENTION_REGEX = /(^|\s|\()(@)([a-zA-Z0-9.-]+)(\b)/g
+export const URL_REGEX =
+  /(^|\s|\()((https?:\/\/[\S]+)|((?<domain>[a-z][a-z0-9]*(\.[a-z0-9]+)+)[\S]*))/gim
+export const TAG_REGEX =
+  /(^|\s)#((?!\ufe0f)[^\s\u00AD\u2060\u200A\u200B\u200C\u200D]*)?/g
```
mary-ext, line 5:
```suggestion
  /(^|\s)#((?!\ufe0f|\u20e2)[^\s\u00AD\u2060\u200A\u200B\u200C\u200D]*)?/g
```

This prevents matching `#⃢test` as a hashtag

packages/api/tests/rich-text-detection.test.ts
mary-ext: Perhaps using escape codes rather than the actual char itself would be beneficial here?

packages/api/src/rich-text/util.ts
estrattonbailey: 🙏 you da best

packages/api/tests/rich-text-detection.test.ts
```diff
@@ -319,9 +327,14 @@ describe('detectFacets', () => {
           },
         ],
       ],
+      ['this #​ should not be a tag', [], []],
+      ['this #​a should not be a tag', [], []],
+      ['this #a​b should be a tag', ['a'], [{ byteStart: 5, byteEnd: 7 }]],
+      ['no match \\u20e2: #⃢tag', [], []],
+      ['no match a\\u20e2b: #a⃢b', ['a'], [{ byteStart: 19, byteEnd: 21 }]],
```
estrattonbailey, line 334: These actually contain a [combining enclosed screen](https://symbl.cc/en/20E2/#:~:text=Symbol%20Meaning,-Combining%20Enclosing%20Screen&text=The%20symbol%20%E2%80%9CCombining%20Enclosing%20Screen,U%2B20E2) as Mary pointed out

packages/api/src/rich-text/util.ts
```diff
@@ -0,0 +1,6 @@
+export const MENTION_REGEX = /(^|\s|\()(@)([a-zA-Z0-9.-]+)(\b)/g
+export const URL_REGEX =
+  /(^|\s|\()((https?:\/\/[\S]+)|((?<domain>[a-z][a-z0-9]*(\.[a-z0-9]+)+)[\S]*))/gim
+export const TAG_REGEX =
+  /(^|\s)[#＃]((?!\ufe0f|\u20e2)[^\s\u00AD\u2060\u200A\u200B\u200C\u200D\u20e2]*[^\p{P}\d\s\u00AD\u2060\u200A\u200B\u200C\u200D\u20e2]+[^\s\u00AD\u2060\u200A\u200B\u200C\u200D\u20e2]*)?/gu
```
mary-ext, line 5: the punctuation character set includes zwj zws and the other zero width characters 

now that it's like this, the core matching could be simplified to `(?!\ufe0f|\u20e3)[\d]*[^ \n\d\p{P}]+[\d]*` i suppose?

estrattonbailey: Testing, but appears that `/\p{P}/u.test(<zws>)` is `false` 🤔

I'd love to try to capture them all, or a range, but this regex will likely continue to grow either way

estrattonbailey: Looks like `Cf` works, wonder if that breaks anything...

estrattonbailey: `\p{Cf}\p{Me}` 🤔

estrattonbailey: Matches some characters we might as well support though, like arabic number sign

PR #2263, opened 2024-03-02
sync lex: #account firehose event; account status errors; and getAccountStatus endpoint

bnewbold:
As discussed earlier today. Did not include any changes to the `#identity` event as we don't have consensus on that yet.

Notes/details:

- I wrote this somewhat quickly and should probably self-review
- I'm not super opinionated about "Account" / "Repo" terminology
- bummer that there are required fields on, eg, `getLatestCommit`. I wedged `rev` in to `getAccountStatus` and we could put the `cid` in there too, and just deprecate `getLatestCommit`. In that case maybe `getRepoStatus` is a better name, and clearer in the case of Relay (which doesn't really have an "account", just the repo)
- I included one inconsistency between the event stream and error status codes: there is no distinction between "AccountDeleted" and "AccountNotFound". I think this sort of makes sense; if the account was deleted, it should look to the outside world the same as if it never existed (on a particular host), but there does need to be a "deleted" account status that goes out on the firehose. I'm not super opinionated about this and we could include an indefinite "account deleted" status if we want them to match one-to-one though


lexicons/com/atproto/sync/subscribeRepos.json
```diff
@@ -121,6 +122,30 @@
...
+      "properties": {
+        "seq": { "type": "integer" },
+        "did": { "type": "string", "format": "did" },
+        "time": { "type": "string", "format": "datetime" },
+        "active": {
+          "type": "boolean",
+          "description": "Indicates that the account has a repository which can be fetched from the host that emitted this event."
+        },
+        "status": {
+          "type": "string",
```
dholms, line 138: made a tweak here

with token values we normally just use a `string` type & list the token values in `knownValues`.

see `com.atproto.moderation.defs` for an example with report types

dholms (approved):
besides the change I made, this all looks good 👍

dholms:
Actually on second pass i think i like `getRepoStatus` better. `com.atproto.sync` routes generally speak in terms of "repos"  and I agree that it probably makes sense for something like `getRepoStatus` to eventually supersede `getLatestCommit`

lexicons/com/atproto/sync/subscribeRepos.json
devinivy: To be honest I am not even sure if we need tokens here— its possible it could be non-namespaced strings.

In my mind the value in tokens is that it allows extensibility by other namespaces, especially in cases where it's friendly for applications to be able to use values that will never make it into the lexicon's `knownValues`.  I don't think that's necessarily needed or even desirable here in the sync endpoints.  One reason may not want that, is that it's nice to keep the values one-to-one with the error types on these sync endpoints, which are not namespaced.  That allows you to understand the repo's status fully through the response of e.g. `sync.getRepo`.

dholms:
Also - we should codegen in this PR before merging 👍

lexicons/com/atproto/sync/subscribeRepos.json
dholms: That's a good point - and I do tend to agree that this isn't a place that we're asking for extensibility on. Any changes should be part of the formal sync protocol 🤔 

So maybe we just do hardcoded values?

devinivy: That would be my vote!  I still support keeping it an open union as a little safety net.

lexicons/com/atproto/sync/defs.json
```diff
@@ -0,0 +1,22 @@
...
+      "type": "token",
+      "description": "Repo hosting status indicating that an administrator has taken down the repo, for a limited (but possibly indefinite) time period."
+    },
+    "deleted": {
+      "type": "token",
+      "description": "Repo hosting status indicating that the repository has been removed. The repo may be re-opened or migrated back to this host in the future, but the contents have been deleted for now. Does not clarify if the account self-deleted or an administrator or operator intervened."
+    },
+    "deactivated": {
+      "type": "token",
+      "description": "Repo hosting status indicating that the repository has been pause and should not be re-distributed, usually on request of the account holder. This may be temporary or indefinite."
```
mackuba, line 19: s/pause/paused/ (though I'm not sure if "repository has been paused" is the best word here...)

lexicons/com/atproto/sync/subscribeRepos.json
bnewbold: it feels like a place "hardcoded" would make sense. agree on keeping it open-union semantics.

lexicons/com/atproto/sync/getRepoStatus.json
```diff
@@ -0,0 +1,42 @@
...
+            "status": {
+              "type": "string",
+              "description": "If active=false, this optional field indicates a possible reason for why the account is not active. If active=false and no status is supplied, then the host makes no claim for why the repository is no longer being hosted.",
+              "knownValues": ["takendown", "suspended", "deactivated"]
+            },
+            "rev": {
+              "type": "string",
+              "description": "Optional field, the current rev of the repo, if active=true"
+            }
+          }
```
devinivy, line 36: Should results in `sync.listRepos` also include the hosting status, and perhaps generally have the same shape as this?

lexicons/com/atproto/sync/subscribeRepos.json
```diff
@@ -118,7 +119,31 @@
...
+      "required": ["seq", "did", "time", "active"],
+      "properties": {
+        "seq": { "type": "integer" },
+        "did": { "type": "string", "format": "did" },
+        "time": { "type": "string", "format": "datetime" },
+        "active": {
+          "type": "boolean",
+          "description": "Indicates that the account has a repository which can be fetched from the host that emitted this event."
+        },
+        "status": {
```
devinivy, line 142: I'm alright with keeping `active` around, though I am not sure it is necessary.  It allows a PDS to express some conflicting information, like setting a `status` with `active=true`.

bnewbold: just to expand on the reasoning there a bit, I wanted to future-proof a bit against us adding additional status strings in the future. If simple infra implementations match on the status string to either make the account visible or not make the account visible (who knows what kind of weird match/switch control flow they might have), then adding a status code could result in the incorrect behavior. Making it a boolean flag makes the "is account visible or not" behavior very clear, with the status being more of an end-user-visible context and less of a high-stakes "is content redistributed at all".

still agree with your point, just wanted to document that bit of thinking

PR #2266, opened 2024-03-02
Update README.md

PSingletary:
There was a line for Developers filling out a form to get an invite code. Since BlueSky is open to everyone, suggested an edit for that line.


dholms (approved):
Good catch 👍 

Thanks for fixing

dholms:
Not quite sure the issue the linter found, but do you mind running a quick `pnpm format`?

PR #2273, opened 2024-03-05
Appview: support rate limit bypass key for blob resolution

devinivy:
When configured, a rate limit bypass key can be included in blob resolution requests to specific hostnames.


dholms (approved):
yup looks good 🙌

PR #2279, opened 2024-03-06
Show only replies from people you follow in Following by default

gaearon:
It has been a pretty common complaint that the Following feed becomes unusable when someone you follow goes on a reply spree because we're showing every single reply by default. This changes so that *by default* (if you haven't expressed a preference), we're going to show a reply if you follow *both* people in the conversation.

People who *have* expressed a particular preference will keep it.

## Test Plan

Create a new user or take an existing one who hasn’t changed this setting. Observe that their default preferences have the "Reply Filters" toggle on whereas previously they wouldn’t.  

Screenshot 2024-03-06 at 14 56 45

### DIFF
--- a/.changeset/ten-tools-exercise.md
+++ b/.changeset/ten-tools-exercise.md
@@ -1,0 +1,8 @@
+---
+'@atproto/ozone': minor
+'@atproto/bsky': minor
+'@atproto/api': minor
+'@atproto/pds': minor
+---
+
+Change Following feed prefs to only show replies from people you follow by default
--- a/.changeset/ten-tools-exercise.md
+++ b/.changeset/ten-tools-exercise.md
@@ -1,8 +1,8 @@
 ---
-'@atproto/ozone': minor
-'@atproto/bsky': minor
-'@atproto/api': minor
-'@atproto/pds': minor
+'@atproto/ozone': patch
+'@atproto/bsky': patch
+'@atproto/api': patch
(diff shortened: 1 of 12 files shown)


packages/api/src/bsky-agent.ts
```diff
@@ -17,7 +17,7 @@ import { sanitizeMutedWordValue } from './util'
 
 const FEED_VIEW_PREF_DEFAULTS = {
```
estrattonbailey, line 18: Optional: could be a good time to export these and integrate them into the app instead of the duplicative code we have rn

estrattonbailey (approved):
LGTM but wanna double check with @dholms on what version we wanna push out

.changeset/ten-tools-exercise.md
```diff
@@ -0,0 +1,8 @@
+---
+'@atproto/ozone': minor
+'@atproto/bsky': minor
+'@atproto/api': minor
+'@atproto/pds': minor
```
dholms, line 5: we definitely what the pds at least to be a `patch` bump since this intersects with our distribution verisioning & we don't want to require a hands on migration

packages/api/src/bsky-agent.ts
gaearon: Don't have strong preference on this and don't want to think about it yet so I'll pass

PR #2284, opened 2024-03-07
Ozone: configure to fetch but not pull from appview

devinivy:
Allows ozone to be run in such a way that it doesn't actually push takedowns out to the configured appview.  Note that this is the new default functionality.  Also allows configuring ozone without a pds.


dholms (approved):
👍

PR #2292, opened 2024-03-11
Updated semantics for atproto labelers header

dholms:
Updated semantics for atproto labelers headers

```
# on requests
atproto-accept-labelers: did:web:mod.bsky.app;redact, did:plc:abc123, did:plc:xyz789

# on responses
atproto-content-labelers: did:web:mod.bsky.app;redact, did:plc:abc123, did:plc:xyz789
```

### DIFF
--- a/packages/bsky/src/util.ts
+++ b/packages/bsky/src/util.ts
@@ -1,21 +1,25 @@
+import { parseList } from 'structured-headers'
+
 export type ParsedLabelers = {
   dids: string[]
   redact: Set<string>
 }
 
-export const parseLabelerHeader = (header: string): ParsedLabelers => {
-  const labelers = header.split(',').map((part) => part.trim())
+export const parseLabelerHeader = (
+  header: string | undefined,
+): ParsedLabelers | null => {
+  if (!header) return null
   const labelerDids = new Set<string>()
   const redactDids = new Set<string>()
-  for (const labeler of labelers) {
-    if (labeler.length === 0) {
-      continue
+  const parsed = parseList(header)
+  for (const item of parsed) {
+    const did = item[0].toString()
+    if (!did) {
+      return null
     }
-    const parts = labeler.split(';')
-    const did = parts[0].trim()
     labelerDids.add(did)
-    const rest = parts.slice(1).map((part) => part.trim())
-    if (rest.includes('redact')) {
+    const redact = item[1].get('redact')?.valueOf()
+    if (redact === true) {
       redactDids.add(did)
     }
   }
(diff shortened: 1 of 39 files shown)


packages/pds/src/pipethrough.ts
```diff
@@ -152,15 +152,24 @@ export const doProxy = async (url: URL, reqInit: RequestInit) => {
...
 
+const RES_HEADERS_TO_FORWARD = [
+  'atproto-repo-rev',
+  'content-language',
+  'atproto-content-labelers',
+]
+
+const makeResHeaders = (res: Response): Record<string, string> => {
+  const headers = RES_HEADERS_TO_FORWARD.reduce((acc, cur) => {
+    acc[cur] = res.headers.get(cur) ?? undefined
```
devinivy, line 167: Is it possible for `headers.get(cur)` to be `null`?  That is the only case I can imagine the `?? undefined` having any effect.  If we want/need to deal with empty strings we could make this `|| undefined`.

packages/bsky/src/util.ts
```diff
@@ -0,0 +1,40 @@
...
+}
+
+export const parseLabelerHeader = (header: string): ParsedLabelers => {
+  const labelers = header.split(',').map((part) => part.trim())
+  const labelerDids = new Set<string>()
+  const redactDids = new Set<string>()
+  for (const labeler of labelers) {
+    if (labeler.length === 0) {
+      continue
+    }
```
devinivy, line 13: This would technically make the header invalid.  If we want to pull in a parser, `parseList()` from [structured-headers](https://www.npmjs.com/package/structured-headers) does what we need here.

devinivy (approved):
Looks good!  We could handle malformed headers by erroring or treating it as though it's empty and assuming the default.

devinivy:
It may be a bit much for now, but it's allowed to combined header values for lists if the header appears multiple times (see [here](https://www.rfc-editor.org/rfc/rfc7230#section-3.2.2)).  For example:
```
atproto-accept-labelers: did:plc:dan, did:plc:bryan
atproto-accept-labelers: did:plc:foysal

# may be equivalent to

atproto-accept-labelers: did:plc:dan, did:plc:bryan, did:plc:foysal
```

packages/pds/src/pipethrough.ts
dholms: yeah the return type for `headers.get()` is `string | null` 👍

packages/bsky/src/util.ts
dholms: ah nice, that sounds like it's probably worth it 👍 

My thinking here was specificlaly for the case of an empty string val ie `atproto-accept-labelers: `

Issue #2265, opened 2024-03-02
`com.atproto.identity.submitPlcOperation` returns "Rotation keys do not include server's rotation key" for unclear reasons


imax9000:
`com.atproto.identity.submitPlcOperation` returns "XRPC ERROR 400: InvalidRequest: Rotation keys do not include server's rotation key", even though none of the keys from the last valid PLC operation are getting removed.

**To Reproduce**

Example code: https://github.com/imax9000/bluesky-repo-backup/blob/59b4b1a4ec909838bd505ceaf1d9c82e48e2cb49/cmd/add-key-to-plc/main.go#L57

Prettified log output:

```
2024/03/02 20:17:27 Recommended credentials: map[
  alsoKnownAs:[at://imax.in.ua]
  rotationKeys:[did:key:zQ3shpKnbdPx3g3CmPf5cRVTPe1HtSwVn5ish3wSnDPQCbLJK]
  services:map[
    atproto_pds:map[
      endpoint:https://enoki.us-east.host.bsky.network
      type:AtprotoPersonalDataServer]]
  verificationMethods:map[
    atproto:did:key:zQ3shme71CNwgBggBYhTmn8VYNZr8wCbnKgiWPRqvL3UUi2qk]]
2024/03/02 20:17:27 Sending the following operation to https://enoki.us-east.host.bsky.network:
map[
  operation:map[
    alsoKnownAs:[at://imax.in.ua]
    prev:bafyreidc46uthkret2gxxnv7y266qrn2x5yjnzolmltct356tkudkwrrhi
    rotationKeys:[
      did:key:zDnaeSDTwrWMnCzhkfJpu4VDETWJHm4AfVv8DEJUXy94ErYnP
      did:key:zQ3shhCGUqDKjStzuDxPkTxN6ujddP4RkEKJJouJGRRkaLGbg
      did:key:zQ3shpKnbdPx3g3CmPf5cRVTPe1HtSwVn5ish3wSnDPQCbLJK]
    services:map[
      atproto_pds:map[
        endpoint:https://enoki.us-east.host.bsky.network
        type:AtprotoPersonalDataServer]]
    sig:XGE_LZbpNladyuYQLWlF8wf6OOPoHeL9tJiYHE7qG7wI6GXqrdU6-qZm0_X4m364gCAtPTquymZIzpCP_gDUgg
    type:plc_operation
    verificationMethods:map[
[...]

PR #2296, opened 2024-03-12
Clean up PLC op submission

dholms:
Set of fixes for PLC op submission:
- check against entryway plc rotation key when running PDS in entryway mode
- emit an `#identity` event on firehose after submitting operation
- bust DID cache for requester after submitting operation

Closes https://github.com/bluesky-social/atproto/issues/2265


devinivy (approved):
Perf!

PR #2297, opened 2024-03-12
Allow 3p labelers to query appview account infos

dholms:
This is a route for bulk resolving handles & profile records. We currently have it behind moderation auth, but instead we put it behind standard auth with a special moderation check for viewing takedowns (just like `getProfile`)

### DIFF
--- a/packages/bsky/src/api/com/atproto/admin/getAccountInfos.ts
+++ b/packages/bsky/src/api/com/atproto/admin/getAccountInfos.ts
@@ -5,18 +5,25 @@ import { INVALID_HANDLE } from '@atproto/syntax'
 
 export default function (server: Server, ctx: AppContext) {
   server.com.atproto.admin.getAccountInfos({
-    auth: ctx.authVerifier.roleOrModService,
-    handler: async ({ params }) => {
+    auth: ctx.authVerifier.optionalStandardOrRole,
+    handler: async ({ params, auth }) => {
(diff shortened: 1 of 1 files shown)


packages/bsky/src/api/com/atproto/admin/getAccountInfos.ts
```diff
@@ -5,18 +5,25 @@ import { INVALID_HANDLE } from '@atproto/syntax'
...
+    handler: async ({ params, auth }) => {
       const { dids } = params
+      const { canViewTakedowns } = ctx.authVerifier.parseCreds(auth)
+
       const actors = await ctx.hydrator.actor.getActors(dids, true)
 
       const infos = mapDefined(dids, (did) => {
         const info = actors.get(did)
         if (!info) return
+        if (info.takedownRef && !canViewTakedowns) return
```
devinivy, line 18: Do we need to apply takedown labels here as well?

devinivy (approved):

packages/bsky/src/api/com/atproto/admin/getAccountInfos.ts
dholms: I lean towards no since it's not an application route, & I would expect most ozone instances to want to get as much info as possible. But not opposed to adding it, in some ways it's simpler 🤷‍♂️

PR #2298, opened 2024-03-12
Appview: labels on feed gens and list views

devinivy:
Adds labels to feed generator and list views lexicons, and satisfies them in the appview.


dholms (approved):
Looks great 👍 

this will stack nicely with https://github.com/bluesky-social/atproto/pull/2270 where we were already doing some label hydration 👌

PR #2301, opened 2024-03-13
:sparkles: Detect language from record content if lang property is not set

foysalit:
For records like list, feedgen etc. in order to put subjects in the right queue, we detect language from the record's description/name.


devinivy:
I believe that there's some related language-detection logic in the Ozone UI.  Is there a world in which we locate all of that logic on the backend, rather than splitting it across here and the UI?

foysalit:
> I believe that there's some related language-detection logic in the Ozone UI. Is there a world in which we locate all of that logic on the backend, rather than splitting it across here and the UI?

the language detection stuff on ozone, just like the one here, is originally from the app's (social-app repo) code. it was introduced to the ozone client app to detect "non-english" language content and show the "translate" option just like the app does. which has nothing to do with tagging records. once this is in, the client code can go away and it will just depend on the tag to determine if it should show the translate button or not.

packages/ozone/src/mod-service/lang.ts
```diff
@@ -40,6 +45,27 @@ export class ModerationLangService {
...
+  ): string | undefined {
+    const atUri = new AtUri(uri)
+    let text: string | undefined
+    if (
+      [ids.AppBskyGraphList, ids.AppBskyFeedGenerator].includes(
+        atUri.collection,
+      )
+    ) {
+      text = (recordValue?.description || recordValue?.name) as string
+    }
```
devinivy, line 60: Another way to slice this is to use `AppBskyGraphList.isRecord(recordValue)` to check the record's `$type` (i.e. its collection). A nice upside is that it ends-up attributing `recordValue` the proper typescript type (assuming `recordValue` is a valid record), which could help avoid the need to cast.  E.g. might look something like:
```ts
if (AppBskyGraphList.isRecord(recordValue)) {
  text = recordValue.description || recordValue.name
} else if (AppBskyFeedGenerator.isRecord(recordValue)) {
  text = recordValue.description || recordValue.name
} else if (AppBskyFeedPost.isRecord(recordValue)) {
  text = recordValue.text
}
```

packages/ozone/tests/lang.test.ts
```diff
@@ -0,0 +1,110 @@
...
+    })
+
+    return subjectStatuses[0]
+  }
+
+  it('Adds language tag to post from text', async () => {
+    const createPostAndReport = async (text: string) => {
+      const post = await sc.post(sc.dids.carol, text)
+      // Without this delay, it seems like from time to time, the appview `getRecord` does not find the record
+      await new Promise((res) => setTimeout(res, 1000))
```
devinivy, line 47: We probably want to use `await network.processAll()` here in order to ensure all writes get indexed by the appview.

packages/ozone/tests/__snapshots__/moderation-statuses.test.ts.snap
```diff
@@ -94,7 +94,7 @@ Array [
     "subjectBlobCids": Array [],
     "subjectRepoHandle": "alice.test",
     "tags": Array [
-      "lang:und",
+      "lang:ha",
```
devinivy, line 97: I wonder how this got picked-up as Hausa, huh!

devinivy (commented):
This is going to be awesome.  Just a few small notes to look at, then we can get this rolled out!

packages/ozone/src/mod-service/lang.ts
```diff
@@ -40,6 +45,27 @@ export class ModerationLangService {
...
+
+    if (!text && recordValue?.text) {
+      text = recordValue.text as string
+    }
+
+    return text?.trim()
```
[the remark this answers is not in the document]
foysalit: hmm... I took a few smaples from prod and they are always strings so thought this was safe. what would you recommend here. we could always wrap this in a try catch and return falsy value so that the entire evaluation doesn't fail. keep in mind this is meant to be a fallback and eventually, we should get the right lang tags from the record itself instead of having to determine from the text.

packages/ozone/tests/lang.test.ts
foysalit: hmm... I vaguely remember trying that but still seeing the intermittent issues but retried it just now and it works fun on re-runs! thanks.

packages/ozone/src/mod-service/lang.ts
foysalit: much cleaner! thanks

packages/ozone/tests/__snapshots__/moderation-statuses.test.ts.snap
foysalit: yh I have no idea why lande thinks this is not english but FWIW it's a 1 word text and here's the context

```
    {
      recordLang: undefined,
      recordText: 'again',
      detectedLanguages: [
        [ 'hau', 0.9999185800552368 ],
        [ 'ita', 0.00004457800605450757 ],
        [ 'fra', 0.000017356069292873144 ],
        [ 'eus', 0.000010054123777081259 ],
        [ 'lit', 0.000003914965873264009 ],
        [ 'cat', 0.000002379479610681301 ],
        [ 'ron', 9.423987989976013e-7 ],
        [ 'nld', 7.186478683252062e-7 ],
        [ 'fin', 7.134470365599554e-7 ],
        [ 'spa', 2.7768982135967235e-7 ],
        [ 'tgl', 2.735682471666223e-7 ],
        [ 'deu', 6.705305821697038e-8 ],
        [ 'ind', 6.230607141333167e-8 ],
        [ 'eng', 2.998981329938033e-8 ],
        [ 'dan', 2.6060373414793503e-8 ],
        [ 'swe', 1.1692682377884012e-8 ],
        [ 'nob', 5.23202203783768e-10 ],
        [ 'est', 1.266992483150986e-10 ],
        [ 'por', 5.452261211957854e-11 ],
        [ 'hrv', 5.3259594651189346e-11 ],
        [ 'srp', 3.363964662383978e-11 ],
        [ 'vie', 2.8038439625621692e-11 ],
        [ 'afr', 2.2980886291157354e-11 ],
        [ 'aze', 1.99868680633708e-11 ],
        [ 'tur', 1.8872434864869447e-11 ],
        [ 'hin', 4.599553810741286e-12 ],
        [ 'run', 9.66062733281925e-13 ],
        [ 'isl', 4.791121384177532e-13 ],
        [ 'mar', 4.763421644973787e-13 ],
        [ 'kor', 3.793560788851319e-13 ],
        [ 'pol', 2.0960263960676656e-13 ],
        [ 'hun', 1.7962217271298014e-13 ],
        [ 'ces', 1.2116070408248192e-13 ],
        [ 'rus', 1.0275874389679279e-13 ],
        [ 'kaz', 1.518541694299491e-14 ],
        [ 'ukr', 6.888189664651639e-15 ],
        [ 'cmn', 2.8416115201303866e-15 ],
        [ 'bel', 5.244914830275721e-16 ],
        [ 'bul', 1.613237510819063e-16 ],
        [ 'mkd', 1.2462328255014537e-16 ],
        [ 'ell', 7.413036478000584e-17 ],
        [ 'jpn', 4.699662384422489e-17 ],
        [ 'slk', 3.8189330768629835e-17 ],
        [ 'ben', 2.2677584972698556e-17 ],
        [ 'ckb', 1.0249701676443585e-17 ],
        [ 'pes', 6.417232037010355e-18 ],
        [ 'kat', 1.9474931478843814e-18 ],
        [ 'hye', 1.8691173712546757e-18 ],
        [ 'heb', 5.477421478276342e-19 ],
        [ 'ara', 2.237874785598848e-20 ]
      ]
    }
```

so, there was no lang set for the record and 1 word text `again` was evaluated to be of `hau` language code by lande.

packages/ozone/src/mod-service/lang-data.ts
```diff
@@ -0,0 +1,560 @@
+interface Language {
+  code3: string
+  code2: string
+  name: string
+}
+
+export const LANGUAGES: Language[] = [
```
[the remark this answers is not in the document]
foysalit: we use this in the social-app here https://github.com/bluesky-social/social-app/blob/main/src/locale/languages.ts#L7 which doesn't mention a source so I'm just gonna reference to that file for now.

packages/ozone/src/mod-service/lang-data.ts
```diff
@@ -0,0 +1,560 @@
...
+  { code3: 'oji', code2: 'oj', name: 'Ojibwa' },
+  { code3: 'ori', code2: 'or', name: 'Oriya' },
+  { code3: 'orm', code2: 'om', name: 'Oromo' },
+  { code3: 'osa', code2: '', name: 'Osage' },
+  { code3: 'oss', code2: 'os', name: 'Ossetian; Ossetic' },
+  { code3: 'ota', code2: '', name: 'Turkish, Ottoman (1500-1928)' },
+  { code3: 'oto', code2: '', name: 'Otomian languages' },
+  { code3: 'paa', code2: '', name: 'Papuan languages' },
+  { code3: 'pag', code2: '', name: 'Pangasinan' },
+  { code3: 'pal', code2: ' ', name: 'Pahlavi' },
```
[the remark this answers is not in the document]
foysalit: yeah I'm not sure why that is but these mostly look like obscure languages and lande probably doesn't support most of these anyways so we're probably fine for our current use case.

devinivy (approved):
Looking good!

PR #2304, opened 2024-03-13
Mod sdk fixes

estrattonbailey:
Ensures our default labeler is supplied to `moderationPrefs` and thereby `ModerationOpts`


pfrazee (approved):
Yep thanks eric

PR #2307, opened 2024-03-13
:bug: Append path to base url when diverting blob

foysalit:


packages/ozone/src/daemon/blob-diverter.ts
```diff
@@ -87,7 +87,9 @@ export class BlobDiverter {
       subjectUri,
     }: { subjectDid: string; subjectUri: string | null },
   ) {
-    const url = new URL(this.serviceConfig.url)
+    const url = new URL(
+      `${this.serviceConfig.url}/xrpc/com.atproto.unspecced.reportBlob`,
+    )
```
devinivy, line 92: A nice way to handle this that is agnostic to a trailing slash on `this.serviceConfig.url` is to use the second arg of `URL`:
```ts
const url = new URL('xrpc/com.atproto.unspecced.reportBlob', this.serviceConfig.url)
```

devinivy (approved):

PR #2310, opened 2024-03-13
Remove `false` `neg` on the labels that ozone emits

dholms:
When `neg` is set to `false`, we omit the field entirely when signing & emitting labels.

Note: this should be paired with a database migration that removes any cached label sigs


devinivy (approved):
Looks great

PR #2311, opened 2024-03-13
Pass proxy header along with auth passthrough

devinivy:
Ensuring that the entryway respects atproto-proxy and atproto-accept-labeler headers.


dholms (approved):
yup yup!

PR #2317, opened 2024-03-14
Dedupe labels during appview hydration

devinivy:
We were seeing some duplicate labels come through, this should ensure they are deduped during the course of the appview's hydration steps.

### DIFF
--- a/packages/bsky/src/data-plane/server/routes/labels.ts
+++ b/packages/bsky/src/data-plane/server/routes/labels.ts
@@ -3,28 +3,44 @@ import { noUndefinedVals } from '@atproto/common'
 import { ServiceImpl } from '@connectrpc/connect'
 import { Service } from '../../../proto/bsky_connect'
 import { Database } from '../db'
+import { Selectable } from 'kysely'
+import { Label } from '../db/tables/label'
+
+type LabelRow = Selectable<Label>
 
 export default (db: Database): Partial<ServiceImpl<typeof Service>> => ({
   async getLabels(req) {
     const { subjects, issuers } = req
     if (subjects.length === 0 || issuers.length === 0) {
-      return { records: [] }
+      return { labels: [] }
     }
-    const res = await db.db
+    const res: LabelRow[] = await db.db
       .selectFrom('label')
       .where('uri', 'in', subjects)
       .where('src', 'in', issuers)
       .selectAll()
       .execute()
 
-    const labels = res.map((l) => {
-      const formatted = noUndefinedVals({
-        ...l,
-        cid: l.cid === '' ? undefined : l.cid,
-        neg: l.neg === true ? true : undefined,
-      })
-      return ui8.fromString(JSON.stringify(formatted), 'utf8')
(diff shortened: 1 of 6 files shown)


packages/bsky/src/hydration/label.ts
```diff
@@ -16,10 +17,36 @@ export type { Label } from '../lexicon/types/com/atproto/label/defs'
...
+      } else {
+        this.set(key, theirs)
+      }
+    })
+    return this
+  }
+  getBySubject(sub: string): Label[] {
+    const it = this.get(sub)?.labels.values()
+    if (!it) return []
+    const labels: Label[] = []
```
dholms, line 43: could be a nice fit for `mapDefined`?

dholms: oh i guess it's an iterator. maybe not 🤔

packages/bsky/src/hydration/label.ts
```diff
@@ -16,10 +17,36 @@ export type { Label } from '../lexicon/types/com/atproto/label/defs'
 
 export type SubjectLabels = {
   isTakendown: boolean
-  labels: Label[]
+  labels: HydrationMap<Label> // src + val -> label
```
dholms, line 20: uri (subject) + val right?

packages/bsky/src/data-plane/server/routes/labels.ts
```diff
@@ -3,28 +3,44 @@ import { noUndefinedVals } from '@atproto/common'
...
-        cid: l.cid === '' ? undefined : l.cid,
-        neg: l.neg === true ? true : undefined,
+    const labelsBySubject = new Map<string, LabelRow[]>()
+    res.forEach((l) => {
+      const labels = labelsBySubject.get(l.uri) ?? []
+      labels.push(l)
+      labelsBySubject.set(l.uri, labels)
+    })
+
+    // intentionally duplicate label results, appview frontend should be defensive to this
```
dholms, line 31: probs fine for now, but might be nice to pull this out into a mock in tests or something - would be nice if the dataplane impl was a "proper" dataplane

dholms (approved):
couple small things but looks good 👍

packages/bsky/src/hydration/label.ts
devinivy: All the labels in `labels` should be for the same subject.  We want to ensure that if we encounter the same label twice, we don't include it both times on the subject.  We want to allow the same label value from two different labelers, though.

In short, we're already narrowed by `uri`, and in this step we narrow on `src + val` to make sure labels in the responses are unique by `(uri, src, val)`.

packages/bsky/src/data-plane/server/routes/labels.ts
devinivy: I agree, but for now I think it's useful for it to behave as close to the one in production as possible.

packages/bsky/src/hydration/label.ts
devinivy: Ooh maybe `mapDefined()` should work on iterators...

PR #2319, opened 2024-03-14
avoid un-necessarily loading the keypair when creating an ActorStoreReader

matthieusieben:
Currently, creating a `localViewer` requires a sync keypair. That keypair is however only actually used in the following function:

https://github.com/bluesky-social/atproto/blob/99b6aee18ec85c4f18ac36e502987304ed7e26ff/packages/pds/src/read-after-write/viewer.ts#L88-L97

Because of this, the keypair might get loaded without being needed, such as here:

https://github.com/bluesky-social/atproto/blob/99b6aee18ec85c4f18ac36e502987304ed7e26ff/packages/pds/src/api/app/bsky/feed/getPostThread.ts#L49-L51

and here:

https://github.com/bluesky-social/atproto/blob/99b6aee18ec85c4f18ac36e502987304ed7e26ff/packages/pds/src/read-after-write/util.ts#L55-L56

This PR changes the `sync` keypair for an async keypair getter, allowing the keypair to be read from disk (and parsed) only when actually needed (in the `serviceAuthHeaders` method) instead of everytime a `localViewer` is needed.


matthieusieben:
These change are motivated by the fact that the oAuth provider interface needs to be able to instantiate a localViewer but does not need to sign anything. I wanted to avoid un-necessarily having to provide the key.

devinivy (approved):
Looks good! @dholms mind taking a quick peek?

dholms (commented):
looks like there's a linter warning for an unused variable in 
`import { Keypair } from '@atproto/crypto'`

dholms (approved):
This looks great! Thanks for getting this going 👍

PR #2316, opened 2024-03-14
Labeler takedowns

dholms:
- check takedown of labelers on hydration of labeler service views
- check takedown of labelers and existence of labeler on hydration of labels

PR #2322, opened 2024-03-14
Labeler takedowns (alternate)

devinivy:
Another option alongside #2316.  I wanted to see what this would look like if we took into account labeler takedowns while building the hydration context rather than during label hydration.  The upsides I'm going for here are a. only perform the takedown checks once, b. improve accuracy of `atproto-content-labelers` response header.

### DIFF
--- a/packages/bsky/tests/label-hydration.test.ts
+++ b/packages/bsky/tests/label-hydration.test.ts
@@ -111,16 +111,7 @@ describe('label hydration', () => {
   it('does not hydrate labels from takendown labeler', async () => {
     AtpAgent.configure({ appLabelers: [alice, sc.dids.dan] })
     pdsAgent.configureLabelersHeader([])
-    await agent.api.com.atproto.admin.updateSubjectStatus(
-      {
-        subject: { $type: 'com.atproto.admin.defs#repoRef', did: alice },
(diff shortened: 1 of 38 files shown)


packages/bsky/tests/label-hydration.test.ts
```diff
@@ -106,6 +108,40 @@ describe('label hydration', () => {
...
+        encoding: 'application/json',
+        headers: network.bsky.adminAuthHeaders(),
+      },
+    )
+    const res = await pdsAgent.api.app.bsky.actor.getProfile(
+      { actor: carol },
+      { headers: sc.getHeaders(bob) },
+    )
+    const { labels = [] } = res.data
+    expect(labels.map((l) => ({ val: l.val, src: l.src }))).toEqual([])
```
devinivy, line 129:
```suggestion
    expect(labels).toEqual([])
```

dholms (approved):
Nice yup 👍 

I like giving this a discrete step & think more stuff will likely make it in there over time.

One thing: this missed the part of https://github.com/bluesky-social/atproto/pull/2316 where we action takedown labels on labelers. So we oughta get that and the test for it in & then i think this is good 👍

PR #2324, opened 2024-03-14
Appview: fix lookup for labeler via repo.getRecord

devinivy:
Just fixing a little issue here affecting the ability to lookup a labeler service record from the appview via `com.atproto.repo.getRecord`.


dholms (approved):
good catch 👍

PR #2325, opened 2024-03-15
Change `reset code` to `deletion code` in 2nd usage in account deletion email

surfdude29:
In the Account Deletion Requested email that a user receives when they request the deletion of their account, there are currently two terms used for the code provided. It is first referred to as a `Deletion code` and subsequently as a `reset code`.

IMG_6190

I think it would be clearer if the second usage is changed to `deletion code`, which is what this PR proposes.

(I don't know what the second change at line 392 at the end of the file is, I only changed line 340 and GitHub seems to have added the other change.)


surfdude29:
After opening this PR I realised that the code is referred to as a `confirmation code` in the app, so perhaps it's better to change both instances in the email to that?

IMG_6191

bnewbold:
Yeah, I think "confirmation code" is probably best.

surfdude29:
> Yeah, I think "confirmation code" is probably best.

I'll go ahead and close this PR then as I'm guessing it's easier for you just to change a couple of words directly in the code :)

PR #2327, opened 2024-03-15
Remove unused dependencies

futurGH:
Since atproto packages are CJS*, consumers can't easily tree shake them, so this'll save quite a few people some bundle size.

*would love to submit a separate PR to change that — only question is whether it'd be preferred to go hybrid CJS/ESM (larger bundle for CJS users, should still be smaller overall for ESM users) or ESM only (same size for CJS, ESM gets the benefit of tree shaking, but CJS has to `await import`)


devinivy (approved):
Appreciate it!

futurGH:
Sweet! This feels too small to open a PR for but `"files": ["dist"]` in package.jsons would also be appreciated — nearly a third of @atproto/api's 2.78 MB bundle size is src and tests at the moment

PR #2344, opened 2024-03-19
Handle apostrophes and other punctuation when muting words

estrattonbailey:


packages/api/tests/moderation-mutewords.test.ts
```diff
@@ -417,7 +495,7 @@ describe(`hasMutedWord`, () => {
           outlineTags: [],
         })
 
-        expect(match).toBe(false)
+        expect(match).toBe(true)
```
estrattonbailey, line 498: This update allows `superbad` to match `super-bad`, basically collapses the punctuation. Previously `super bad` would match `super-bad` but not the collapsed version.

This was done so that `blueskys` match `Bluesky's`. I'm trying to think of cases where this would be TOO eager, but haven't come up with anything yet.

packages/api/tests/moderation-mutewords.test.ts
```diff
@@ -474,7 +552,7 @@ describe(`hasMutedWord`, () => {
...
```
estrattonbailey, line 555: Same thing here, collapses `_` so `idkwhatthiswouldbe` matches `idk_what_this_would_be`

pfrazee (approved):
seems solid to me

PR #2347, opened 2024-03-19
Exclude posts from author feed when they don't fit into a self-thread

devinivy:
When viewing `getAuthorFeed` with the `posts_and_author_threads` filter, previously the appview would serve any reply that the author posted in their own thread in reply to one of their own posts.  This could end-up including posts that were not part of a self-thread, even if they appeared to be "locally" by inspecting their root and parent.

With this change, when we encounter a reply we ensure that the full reply chain up to the root is included in the author feed.  Otherwise we filter it out—this should ensure that only proper self-threads are served when using the `posts_and_author_threads` filter.


dholms (approved):
sweet looks good!

PR #2354, opened 2024-03-20
Exports lexicon from @atproto/bsky. Fixes #2353.

### LINKED ISSUE #2353: Export lexicons from @atproto/bsky package
I'm trying to do a PR for bluesky-social/feed-generator#41 and the biggest issue I keep running into is that `@atproto/bsky` doesn't export a lexicon at all. Specifically I'm needing:

- The `Server` class exported by `@atproto/bsky/src/lexicon`
- `ids` and `lexicons` (specifically `assertValidXrpcMessage()`) from `@atproto/bsky/src/lexicon/lexicons`
- `createServer()` from `@atproto/bsky/src/lexicon`

It feels like `@atproto` packages that use the lexicon auto-generator tool should export those in their own namespace where possible to prevent the necessity of needing to use the lexicon auto-generator tool in downstream packages (e.g. individual feed generators created by the `feed-generator` template.)

aendra-rininsland:
I haven't been able to run tests locally but I don't imagine this would break anything; the build succeeds if nothing else and I suspect if the export was clobbering an existing namespace (the only thing I can think of that would actually break anything with this), `tsc` would complain.

### DIFF
--- a/packages/bsky/src/index.ts
+++ b/packages/bsky/src/index.ts
@@ -24,6 +24,7 @@ import { authWithApiKey as bsyncAuth, createBsyncClient } from './bsync'
 import { authWithApiKey as courierAuth, createCourierClient } from './courier'
 
 export * from './data-plane'
+export * from './lexicon';
 export type { ServerConfigValues } from './config'
 export { ServerConfig } from './config'
 export { Database } from './data-plane/server/db'


aendra-rininsland:
I'm really not sure why linting failed; seems to be in files I didn't touch.

PR #2355, opened 2024-03-20
Fix inclusion for reposts on author feed w/ posts_and_author_threads

devinivy:
In #2347 I introduced a bug, where reposted content was being excluded from author feeds under the `posts_and_author_threads` filter.  This fixes it!


dholms (approved):
sweet looks great 👌

PR #2357, opened 2024-03-20
Always include counts on post and profile views

devinivy:
Ensure counts are set in all post and profile views, even when aggregations are missing.  Apply the same to read-after-write logic in the PDS, and tidy some `$type` typos found along the way.

### DIFF
--- a/packages/pds/src/read-after-write/viewer.ts
+++ b/packages/pds/src/read-after-write/viewer.ts
@@ -168,6 +168,9 @@ export class LocalViewer {
     return {
       uri: uri.toString(),
       cid: cid.toString(),
+      likeCount: 0, // counts presumed to be 0 directly after post creation
+      replyCount: 0,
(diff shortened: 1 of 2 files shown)


packages/pds/src/read-after-write/viewer.ts
```diff
@@ -259,24 +260,20 @@ export class LocalViewer {
...
     } else if (collection === ids.AppBskyFeedGenerator) {
       const res = await this.appViewAgent.api.app.bsky.feed.getFeedGenerator(
-        {
-          feed: embed.record.uri,
-        },
+        { feed: embed.record.uri },
         await this.serviceAuthHeaders(this.did),
       )
       return {
-        $type: 'app.bsaky.feed.defs#generatorView',
```
dholms, line 268: lolrip

dholms (approved):
looks great!

PR #2359, opened 2024-03-21
Use blocked and not-found unions in post reply refs

devinivy:
Feed item views served from the appview include a `post` and a `reply` field.  The `reply` field contains views for the post's reply root and reply parent.  Historically if the reply root/parent were missing or were blocked posts, we would simply omit the `reply` field entirely.  With this change we will start to include it, and use the `app.bsky.feed.defs#notFoundPost` or `app.bsky.feed.defs#blockedPost` if the root/parent are missing or blocked respectively.  This aligns the implementation with how the feed view lexicons were intended to be used.

### DIFF
--- a/packages/bsky/src/views/index.ts
+++ b/packages/bsky/src/views/index.ts
@@ -465,48 +465,33 @@ export class Views {
       post,
       reason,
       reply: !postInfo?.violatesThreadGate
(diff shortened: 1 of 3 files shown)


dholms (approved):
big yup 👍

PR #2363, opened 2024-03-23
Configure labeler service account in dev-env

foysalit:
This PR re-wires the dev-env setup a little bit to create labeler service account which makes it easier to develop and plug ozone client locally into the dev-env.

The implementation is a bit "hacky" so want to lay out the flow and the reasoning. 

The previous implementation created a DID for ozone without actually creating an account on any pds so with that DID, we couldn't create a labeler service record. However, the DID is needed when instantiating the appview and the pds so it's not as straight forward as starting the pds first and creating the labeler account.

So, this PR instantiates a temporary pds, creates an account and uses the DID for instantiating all the services (appview, pds and ozone) then migrates the account from temporary pds to the dev-env pds and shuts down the temporary pds. After migration, the labeler service record is created too.


packages/dev-env/src/network.ts
```diff
@@ -73,7 +82,7 @@ export class TestNetwork extends TestNetworkNoAppView {
       plcUrl: plc.url,
       signingKey: ozoneKey,
       serverDid: ozoneDid,
-      dbPostgresSchema: `ozone_${dbPostgresSchema}`,
+      dbPostgresSchema: `ozone_${dbPostgresSchema || 'db'}`,
```
foysalit, line 85: unrelated change but this prevents the local db from being named `ozone_undefined`

packages/dev-env/src/ozone-service-profile.ts
```diff
@@ -0,0 +1,412 @@
...
+    })
+
+    this.did = modUser.data.did
+    this.key = await Secp256k1Keypair.create({ exportable: true })
+    return { did: this.did, key: this.key }
+  }
+
+  async createServiceDetails(pds: TestPds, ozoneUrl: string) {
+    const pdsClient = pds.getClient()
+    const describeRes = await pdsClient.api.com.atproto.server.describeServer()
```
foysalit, line 37: This mostly mimics the way migration is done in [account-migrations.test.ts](https://github.com/bluesky-social/atproto/blob/main/packages/pds/tests/account-migration.test.ts#L78)

devinivy (approved):

PR #2369, opened 2024-03-28
PDS: Enable to specify name of email sender.

YOCKOW:
While nodemailer can parse address with formatted-name such as "John john@example.com", the name part must be enclosed in double quotes if special symbols are contained.
That may be complicated because we pass the address via an environment variables to PDS.
This PR enables to specify the name of the sender by adding configurable points with backward compatibility.


YOCKOW:
@devinivy 
May I request review from you or someone you recommend?

PR #2377, opened 2024-03-31
lexicon: add support for TID and record-key string formats

bnewbold:
Pretty simple? Doesn't seem like there is any support needed in `lex-cli`.

Will add the actual formats to lexicons in a separate PR.

See also specs PR: https://github.com/bluesky-social/atproto-website/pull/248


dholms (approved):
sweet looks good 👍

PR #2378, opened 2024-03-31
lexicons: use 'tid' and 'record-key' formats in sync+repo lexicons

bnewbold:
Depends on: https://github.com/bluesky-social/atproto/pull/2377

~We'd probably want to hold off on actually merging this until we've done some dev comms and let the specs and library support PRs sink in.~

Needs codegen.

### DIFF
--- a/lexicons/com/atproto/repo/applyWrites.json
+++ b/lexicons/com/atproto/repo/applyWrites.json
@@ -18,7 +18,8 @@
             },
             "validate": {
               "type": "boolean",
-              "description": "Can be set to 'false' to skip Lexicon schema validation of record data across all operations, 'true' to require it, or leave unset to validate only for known Lexicons."
+              "default": true,
+              "description": "Can be set to 'false' to skip Lexicon schema validation of record data, for all operations."
             },
             "writes": {
               "type": "array",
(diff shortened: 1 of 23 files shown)


sugyan:
[My Rust library is ready](https://github.com/sugyan/atrium/pull/155) to handle these new formats correctly and code generating :)

bnewbold:
I think this is now a good candidate to merge and deploy.

All of these use-points are on query parameters, not records, so there are no old/invalid records to worry about (an issue with some other Lexicon changes). These don't change what is an isn't valid: presumably all of these parameters should already validate against these identifier types. This mostly just improves documentation, and will help with codegen in other languages which have specific types for these string formats.

lexicons/com/atproto/repo/applyWrites.json
```diff
@@ -18,7 +18,8 @@
             },
             "validate": {
               "type": "boolean",
-              "description": "Can be set to 'false' to skip Lexicon schema validation of record data across all operations, 'true' to require it, or leave unset to validate only for known Lexicons."
+              "default": true,
```
dholms, line 21: Don't we still want the semantic of "leave unset to validate only for known Lexicons"? If we add a `default`, we lose that

dholms (approved):
TID/rkey changes look good to me 👍 

I think an accidental change may have slipped in around `validate` in `applyWrites`

PR #2383, opened 2024-04-03
Feed interactions schema

dholms:
Adds a lexicon for a client to send interactions back to a feed generator.

This is an optional (but helpful) API for clients to implement to give feedgens feedback on what the user is seeing and whether the user is giving any positive or negative signals about the content, including an "I don't like this" / "Less of this content" signal.


lexicons/app/bsky/feed/generator.json
```diff
@@ -30,6 +30,10 @@
             "accept": ["image/png", "image/jpeg"],
             "maxSize": 1000000
           },
+          "acceptsFeedback": { 
+            "type": "boolean",
+            "description": "Declaration that a feed accepts feedback from a client through app.bsky.feed.sendInteractions"
+          },
```
devinivy, line 36: I wonder if there's a way to unify the terminology around either "feedback" or "interactions."

lexicons/app/bsky/feed/defs.json
```diff
@@ -177,6 +186,81 @@
           "items": { "type": "ref", "ref": "app.bsky.graph.defs#listViewBasic" }
         }
       }
+    },
+    "interaction": {
+      "type": "object",
+      "properties": {
+        "item": { "type": "string", "format": "at-uri" },
```
devinivy, line 193: Will this generally be either a post or repost URI?

lexicons/app/bsky/feed/defs.json
```diff
@@ -177,6 +186,81 @@
...
+        "event": {
+          "type": "string",
+          "knownValues": [
```
devinivy, line 196: If you have any thoughts on `{ item: at-uri, event: token, feedContext: string }` versus open-union-style `{ $type, item: at-uri, feedContext: string, ... }` I'd be interested to hear!

lexicons/app/bsky/feed/defs.json
```diff
@@ -49,7 +49,11 @@
       "properties": {
         "post": { "type": "ref", "ref": "#postView" },
         "reply": { "type": "ref", "ref": "#replyRef" },
-        "reason": { "type": "union", "refs": ["#reasonRepost"] }
+        "reason": { "type": "union", "refs": ["#reasonRepost"] },
+        "feedContext": {
+          "type": "string",
+          "description": "Context provided by feed generator that may be passed back alongside interactions."
+        }
```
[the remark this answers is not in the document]
dholms: Yeah `unknown` just really blows it open and we can't add any constraints to it. Figured keeping it a string made it a bit easier to wrangle. Paul & Why both seemed cozy with an opaque string

Which speaking of, I like throwing a utf8 limit on it. 2000 sounds good 👍

lexicons/app/bsky/feed/generator.json
dholms: Yeah I was wondering that myself. I guess `acceptsInteractions` works just seemed a bit off 🤔

lexicons/app/bsky/feed/defs.json
dholms: Right now I was thinking it would always be the post uri & any repost that got it in there would be associated through the `feedContext` that gets sent around 🤔

I left it somewhat vague because I think it's likely we'll start passing down non-post objects in feeds at some point

lexicons/app/bsky/feed/defs.json
dholms: Yup that was my first pass on it but then ended up redoing it this way. 

Mainly reconsidered because an array of open unions is tough in languages like Go. But also nearly all these interaction types had similar values on them so they were pretty friendly to squash down into a flat object like this so didn't seem like any resolution was lost

lexicons/app/bsky/feed/generator.json
dholms: thought about it & don't mind it so I changed it to `acceptsInteractions` 👍

henoya:
Looking at this definition of "interaction", let me raise a few questions and requests.

Question:
- Is "interactionShare" sent by selecting share from the post menu?
- Is the "clickthroughEmbed" trigger only for embedded content like link cards or Youtube?

henoya:
Requests:
- I would also like to see an item like "clickthroughLink." If possible, it would be nice to be able to identify which link was clicked at this time, if there are multiple links attached.
I once used the feed generator to create a maze exploration game. After that, it stopped working properly due to a change in the display cache specifications of the official custom feed client.
https://henoya.com/posts/20231210/20231210_feed_screen.png
It would be nice to be able to get some kind of interaction from the post in the feed, such as choices (yes, just like the choices in Ozone with the labeler's warnings, etc...! I don't need something that rich), I think it would expand the use of custom feeds.

dholms:
> Is "interactionShare" sent by selecting share from the post menu?

Yup that's right 👍 

> Is the "clickthroughEmbed" trigger only for embedded content like link cards or Youtube? 

As well as record embeds (posts, feed generators, lists, etc). You can tell what type of embed it is by looking at the post

> I would also like to see an item like "clickthroughLink." If possible, it would be nice to be able to identify which link was clicked at this time, if there are multiple links attached.

I dig the creativity there, but it's a bit out of scope for what we're shooting for right now. We'll likely add more clickthrough events and metadata in the future but wanted to keep it fairly simple for now. The schemas open for extension :)

PR #2390, opened 2024-04-08
:sparkles: Allow muting reporter

foysalit:
This PR introduces 2 new moderation events `#modEventMuteReporter` and `modEventUnmuteReporter` which allows moderators to mute incoming reports from a user.

Incoming reports from muted users will still be recorded but it won't change the moderation review state of the subject.


lexicons/tools/ozone/moderation/defs.json
```diff
@@ -284,6 +292,10 @@
       "required": ["durationInHours"],
       "properties": {
         "comment": { "type": "string" },
+        "reportingOnly": {
+          "type": "boolean",
+          "description": "Mute incoming reports from this subject. Only enabled when the subject is a DID"
+        },
```
devinivy, line 298: The way I read this, "reporting only" sort of implies to me that:
 - when `false` reports of this user and reports by this user are muted.
 - when `true` _only_ reports by this user are muted.

Admittedly, this is a pretty one awkward one to name!  What `muteReporting` or `muteIncoming`?  @bnewbold I feel you may have a good thought on this.

packages/ozone/src/mod-service/index.ts
```diff
@@ -341,6 +357,15 @@ export class ModerationService {
...
       .executeTakeFirstOrThrow()
 
+    // If reporting is muted for this reporter, we don't want to create a subject status
+    if (meta.isReporterMuted) {
+      const subjectStatus = await this.getStatus(subject)
+      return {
+        event: modEvent,
+        subjectStatus,
+      }
+    }
```
devinivy, line 367: Could we handle this logic in `adjustModerationSubjectStatus()`?  It seems to have all the necessary information, and might just be a little more uniform to do it there.

packages/ozone/src/mod-service/index.ts
```diff
@@ -757,6 +784,14 @@ export class ModerationService {
       )
     }
 
+    if (onlyMuted) {
+      builder = builder.where((qb) =>
+        qb
+          .where('muteUntil', '>', new Date().toISOString())
+          .orWhere('muteReportingUntil', '>', new Date().toISOString()),
+      )
+    }
```
devinivy, line 793: I expected to see `muteReportingUntil` in here, but I don't totally understand what `muteUntil` is doing in here too.  What's the right way to think about it?

devinivy (commented):
Good stuff!  Just had a few small thoughts/qs.

packages/ozone/src/mod-service/index.ts
foysalit: In my mind, `Mute` is just one state applied in 2 different fronts. Either the subject is muted from showing up in queue regardless of state change or the subject is muted from changing state of other subjects through reports. With that explained, `onlyMuted` filter is meant to show *all subjects* that are `muted` which is why both `muteUntil` and `muteReportingUntil` are checked here.

lexicons/tools/ozone/moderation/defs.json
foysalit: hmm.. yeah I get what you mean but the first point is not correct. `false` there means it will mute the subject itself from showing up in mod queue (the original mute). the rest is correct.

I like the property name itself since it feels accurate. However, I agree that in the context of our current "mute" and what setting "reportingOnly" to true does is kinda confusing. Happy to consider a different name but also, we can consider an entirely new event for this?

devinivy: > hmm.. yeah I get what you mean but the first point is not correct. false there means it will mute the subject itself from showing up in mod queue (the original mute). the rest is correct.

Ah yeah, I didn't communicate that very well!  I realize that the first point is not correct, and was trying to say that the fact that it's incorrect tells me that the name might be confusing.  The term "only" implies to me that it's limiting down some broader behavior, e.g. muting both reporter and reportee limited down to _only_ muting reportee.  But what we want are just two independent behaviors: muting reporter or muting reportee, and not both at the same time.

Getting quite semantic here 🤪 @bnewbold mind breaking the tie on this?

bnewbold: - the description of the overall event should be updated to clarify it isn't always "Mute incoming reports on a subject"
- it does kind of feel like an entirely different event would make sense, though this would make the PR more complex. it would clarify the "only for accounts, not records" part. I would lean towards doing a separate event to keep the API "correct". maybe "BlockReporter"? we already use that word elsewhere, but it might be clear enough what is meant. or "MuteReporter".
- if we want to stick with just a flag, I would be call it `reporter` or (verbosely) `reportsFromAccount`. and update all the relevant descriptions and docs (eg, there is a dev doc in the Ozone repo) to clarify what is going on

What would actually be best is probably renaming the existing event "MuteSubject". But I feel like renaming an existing event would be really disruptive.

devinivy (approved):
Good call with those final changes 👍 this looks good!

PR #2395, opened 2024-04-09
Appview: include like/repost/reply counts on post embed

devinivy:
This adds the `likeCount`, `repostCount`, `replyCount` fields to the `#viewRecord` embed type.  We may continue to cater the `#viewRecord` lexicon to post embeds, bringing it more in line with list, feedgen, and labeler embed views.


dholms (approved):
beautiful 💎

haileyok:
💎

PR #2399, opened 2024-04-10
getSuggestionsSkeleton lexicon

dholms:
Adds a schema for `getSuggestionsSkeleton`

Return values are of type `app.bsky.actor.defs#skeletonActor` which is just a wrapper object around `did` but leaves us room to add some annotations in the future if we need to


devinivy (approved):
Looking good!  My only question is whether `app.bsky.actor.getSuggestionsSkeleton` should be `unspecced` similar to other non-public backing endpoints such as `app.bsky.unspecced.searchActorsSkeleton`.

PR #2411, opened 2024-04-15
Add `primaryAlgoPrefs` lex and `setPrimaryAlgorithm` method

estrattonbailey:
Adds `primaryAlgoPref` to preferences, and adds a method `setPrimaryAlgorithm({ enabled: boolean, uri: string })` to `BskyAgent`. See tests for usage.

Notes:
- if `enabled` is `undefined` we know the user has no interacted with this pref, which is an ability I want to retain, lmk if any issues with the optionality of this prop — I guarded against this value being set back to `undefined` once set to a boolean
- when a feed is set as the primary algo, it's removed from saved/pinned feeds to avoid confusion

### DIFF
--- a/.changeset/bright-actors-shout.md
+++ b/.changeset/bright-actors-shout.md
@@ -2,4 +2,4 @@
 "@atproto/api": patch
 ---
 
-Adds `homeAlgoPref` lexicons and related method to `BskyAgent`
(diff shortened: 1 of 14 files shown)


dholms (approved):
looks good 👍

PR #2416, opened 2024-04-17
Lexicons: email auth factor

devinivy:
This contains lexicons for supporting an email auth factor within the current authentication system.  The PDS will eventually support other factors as a more long-term solution via OAuth (see #1958 #2408).  These lexicons will be used as a stand-in for now, at minimum for use with the entryway service (bsky.social) which will support them.


pfrazee (approved):
LGTM

PR #2423, opened 2024-04-18
Appview: ensure hydration context tracks viewer did and not full service ref

devinivy:
Service refs (e.g. `did:plc:example#service_id`) were getting preserved into the hydration and dataplane layer, where we really just want to work with the relevant DIDs (e.g. `did:plc:example`).


dholms (approved):
👌

PR #2424, opened 2024-04-18
Appview: maintain language info when going out to a suggestions service

devinivy:
This ensures that when a service is used to provide suggestions on `actor.getSuggestions`, language information associated with the request is preserved.


dholms (approved):
yup 👍

PR #2425, opened 2024-04-18
Generic PDS proxy

dholms:
Allow the PDS to proxy generic `/xrpc/` endpoints based on `atproto-proxy`.

Auth with the PDS is validated and a service auth token for the did is signed and attached to the outgoing request.


packages/pds/src/pipethrough.ts
```diff
@@ -104,11 +64,11 @@ const REQ_HEADERS_TO_FORWARD = [
 export const createUrlAndHeaders = async (
   ctx: AppContext,
   req: express.Request,
-  requester?: string,
+  requester: string | null,
   audOverride?: string,
 ): Promise<{ url: URL; headers: { authorization?: string } }> => {
   const proxyTo = await parseProxyHeader(ctx, req)
-  const defaultProxy = defaultService(ctx, req.path)
+  const defaultProxy = defaultService(ctx, req.baseUrl)
```
dholms, line 71: The mount path doesn't get included in `req.path` (https://expressjs.com/en/4x/api.html#req.path) so had to change to `baseUrl`

bnewbold (commented):
A couple thoughts (I didn't look carefully enough to see if these are implemented):

- only allow exactly `/xrpc/<nsid>`: no additional path components, no `/xrpc/../blah`, etc
- only authenticated users (looks like this is enforced?)
- `atproto-accept-labelers`, `accept-language`, `content-type`, probably a few other HTTP headers passed through (some request, some response)
- bodies: JSON probably the priority, but might want to support other mimetypes? presumably things like size limits are enforced by load-balancers already
- `502 Bad Gateway` and `504 Gateway Timeout` when needed? any 5xx or 4xx status codes can probably be passed back as-is?
- if the upstream redirects (3xx), PDS could follow the redirects a couple hops, or return an error to user, but probably shouldn't return the redirect to the user
- to start, just the usual rate-limits?
- probably shouldn't support streams (websocket) in this early impl (!)
- will probably need to figure out some way to hint the level of auth ("scope") at some point: app-password, OAuth scope, etc

packages/pds/src/pipethrough.ts
devinivy: I think this works due to an implementation detail of xrpc-server, that it happens to write its catchall route this way:
```ts
    this.router.use(
      '/xrpc/:methodId', // <- baseUrl comes from here
      opts?.catchall ?? this.catchall.bind(this),
    )
```

If we're able to base this on `req.originalUrl` (need to account for it containing query params) I think we may be able to avoid a potential footgun.  Alternately can probably carefully concat `req.baseUrl` and `req.path` together.

dholms: yup yup totally 👍 

ended up rewriting to use `req.originalUrl`

also fwiw, this logic is only used to find the default service so it's relatively low stakes 👌

packages/pds/src/pipethrough.ts
```diff
@@ -1,125 +1,75 @@
...
-      return ctx.cfg.modService
-    case ids.ComAtprotoModerationCreateReport:
-      return ctx.cfg.reportService
-    default:
-      return ctx.cfg.bskyAppView
+export const proxyHandler =
+  (ctx: AppContext): CatchallHandler =>
+  async (req, res, next) => {
+    try {
+      const auth = await ctx.authVerifier.access({ req })
```
devinivy, line 22: It would be kind of nice to bail early and not bother with the work of auth'ing if the request is non-lexrpc.

packages/pds/src/pipethrough.ts
```diff
@@ -145,18 +92,70 @@ export const createUrlAndHeaders = async (
...
+// -------------------
+
+export const makeRequest = async (
+  url: URL,
+  reqInit: RequestInit,
+): Promise<Response> => {
   let res: Response
-  let buffer: ArrayBuffer
   try {
     res = await fetch(url, reqInit)
```
devinivy, line 152: Something to look out for in general is ensuring we always read the request body.  By not reading the body, I think we'd end-up leaking connections and other resources.

devinivy: I think we're all set on this, but just wanted to put it out there!

devinivy (approved):
Looks good.  This is gonna rule.

packages/pds/src/pipethrough.ts
dholms: yup good looks 👍 

think we always do read it - but taking a look back at it, i'm gonna adjust the error handling just a smidge 👌

packages/pds/src/pipethrough.ts
dholms: nice - i split up the `createUrlAndHeaders` function and now do url + audience before checking auth and then create the signed headers after checking auth 👌

devinivy (approved):

PR #2426, opened 2024-04-18
:sparkles: Search accounts by normalized email

foysalit:
This PR adds new lexicon that allows moderators to search for accounts. This first iteration will only allow searching by email address. The search query allows 2 types of keywords

1. keyword that looks like `@domain.tld` will be treated specially and all accounts that signed up with an email address  from domain.tld will be returned
2. keywords that does not start with `@` will be matched against normalized email addresses and all matches will be returned. to facilitate this, along with email, a new column `normalizedEmail` has been added to the `account` table which stores a normalized version of the email address and removes characters like + or . from the address.


bnewbold:
I kind of suspect we will need to have a normalized column and query on that?

Or maybe as a hack could parse the email and split off the username; split at any `+`; and do a query: `LIKE 'prefix%@domain.tld'`

But that won't work with `.` and other ways that folks fiddle with domains.

We don't strictly require email verification at signup, so tons of folks use services like gmail; that makes just domain queries somewhat ineffective.

lexicons/com/atproto/admin/searchAccounts.json
```diff
@@ -0,0 +1,45 @@
...
+    "accountSearchResult": {
+      "type": "object",
+      "required": ["did"],
+      "properties": {
+        "did": { "type": "string" },
+        "email": { "type": "string" },
+        "normalizedEmail": { "type": "string" },
+        "handle": { "type": "string" }
+      }
+    }
```
devinivy, line 43: What would you think of serving a `com.atproto.admin.defs#accountView`, keeping it consistent with similar endpoints like `admin.getAccountInfos`?

lexicons/com/atproto/admin/searchAccounts.json
```diff
@@ -0,0 +1,45 @@
...
+        }
+      }
+    },
+    "accountSearchResult": {
+      "type": "object",
+      "required": ["did"],
+      "properties": {
+        "did": { "type": "string" },
+        "email": { "type": "string" },
+        "normalizedEmail": { "type": "string" },
```
devinivy, line 40: I think of the `normalizedEmail` as an implementation detail of how our email search functionality works.  If we don't have a strong need for it in the client, or if the client can normalize the email itself for the user, I wonder if that may be preferable.

lexicons/com/atproto/admin/searchAccounts.json
foysalit: yep, makes sense to serve accountView if we get rid of normalizedEmail

PR #2429, opened 2024-04-19
PDS: handle S3 upload timeout more gracefully

devinivy:
This makes upload timeouts to S3 handled a bit more consistently, and surfaces them to the user in a more legible way.
 - Support a config on `S3BlobStore` for upload timeouts.
 - Apply  `S3BlobStore` timeouts to  both `putTemp()` and `putPermanent()`.
 - Clear upload timeouts more reliably (e.g. in cases where upload fails for some reason other that the timeout).
 - Add PDS config `PDS_BLOBSTORE_S3_UPLOAD_TIMEOUT_MS`, defaulting to `20000`.  The previous timeout was 10sec which was too short in enough cases that it was worth increasing to 20sec.
 - Surface better error to user when `repo.uploadBlob` endpoint fails due to a timeout.

### DIFF
--- a/packages/aws/src/s3.ts
+++ b/packages/aws/src/s3.ts
@@ -55,12 +55,13 @@ export class S3BlobStore implements BlobStore {
 
   async putTemp(bytes: Uint8Array | stream.Readable): Promise<string> {
     const key = this.genKey()
+    // @NOTE abort results in error from aws-sdk "Upload aborted." with name "AbortError"
     const abortController = new AbortController()
     const timeout = setTimeout(
-      () => abortController.abort('upload timed out'),
+      () => abortController.abort(),
       this.uploadTimeoutMs,
     )
-    await new Upload({
+    const upload = new Upload({
       client: this.client,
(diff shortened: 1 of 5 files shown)


packages/aws/src/s3.ts
```diff
@@ -53,12 +55,13 @@ export class S3BlobStore implements BlobStore {
 
   async putTemp(bytes: Uint8Array | stream.Readable): Promise<string> {
     const key = this.genKey()
+    // @NOTE abort results in error from aws-sdk "Upload aborted." with name "AbortError"
     const abortController = new AbortController()
     const timeout = setTimeout(
-      () => abortController.abort('upload timed out'),
```
devinivy, line 58: The aws-sdk was throwing the `reason` passed to `abort()` away, which was actually a little confusing, so I nixed the message/reason here.

pfrazee (commented):
Cool this lgtm

dholms (approved):
looks great 👍

PR #2430, opened 2024-04-19
Appview: remove replies to blocked posts from feeds

devinivy:
Remove any replies appearing in a feed or timeline that are in response to a blocked post.  Before we would serve the reply and signal that the parent is blocked using a `#blockedPost`, which remains the behavior on author feeds.


dholms (approved):
yup! 💎

mackuba:
Ok, what if I *want* to see posts which are replies to blocked posts in my feeds? There are sometimes some interesting discussions e.g. about ATProto stuff that show up in the ATProto feeds, which I don't want to miss just because they happen somewhere below a post from someone who has blocked me.

PR #2431, opened 2024-04-20
Persist PDS URL in session object

mary-ext:
from https://github.com/bluesky-social/social-app/pull/3620#discussion_r1572786499
> We should probs make an API on the agent for this

See https://github.com/bluesky-social/social-app/pull/3620#issue-2253150163 for rationale behind this change.

cc @pfrazee

### DIFF
--- a/packages/api/src/agent.ts
+++ b/packages/api/src/agent.ts
@@ -36,10 +36,17 @@ export class AtpAgent {
   labelersHeader: string[] = []
   proxyHeader: string | undefined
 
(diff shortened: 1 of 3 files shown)


mary-ext:
Let me look into the tests...

pfrazee:
We're going to turn attention to the OAuth updates so I'm probably going to fold this kind of thing into those updates, but this is helpful and I want to keep it open until then as a reference

PR #2440, opened 2024-04-23
more syntax tests: better TID syntax coverage, and add CID, language, URI examples

bnewbold:
Would be good to wire up the CID, language, and URI examples with syntax checking code in the SDK, but these are already used in the golang implementation and might be useful to 3rd parties, so just adding the files for now.


bnewbold:
plan is to move these to the newish interop repo: https://github.com/bluesky-social/atproto-interop-tests

PR #2442, opened 2024-04-23
:sparkles: Allow dynamic labeler header through ozone

foysalit:
This PR aims to allow moderators of an ozone instance plug in an external ozone service/moderation authority and view labels they have added to a given subject.

The current implementation only adds this to `getRecord` lexicon but lays out the ground work to easily enable this for other lexicons.


bnewbold:
Cool!

Any thoughts on adding `labels` on `repoView`, to be symmetric with the addition to `recordView`? Both of the "ViewDetail" variants already have labels.

packages/api/src/agent.ts
```diff
@@ -263,14 +263,26 @@ export class AtpAgent {
         'atproto-proxy': this.proxyHeader,
       }
     }
+
+    const labelerHeaderName = 'atproto-accept-labelers'
```
foysalit, line 267: @pfrazee wanted to run these changes by you

packages/bsky/src/api/com/atproto/repo/getRecord.ts
```diff
@@ -18,12 +18,22 @@ export default function (server: Server, ctx: AppContext) {
       throw new InvalidRequestError(`Could not locate record: ${uri}`)
     }
 
+    const labelers = ctx.reqLabelers(req)
+    const labelMap = await ctx.hydrator.label.getLabelsForSubjects(
+      [uri],
+      labelers,
+    )
+    const labels = Array.from(labelMap.get(uri)?.labels?.values() || [])
```
devinivy, line 26: Hydrating labels here in `repo.getRecord` is a little bit funky since it's quietly off-contract (and I don't think we'll want to add it to the contract).  I believe we had plans for the appview to implement `com.atproto.label.queryLabels` to provide the labels on arbitrary subjects, which I think would speak to ozone's use case.  Perhaps we could implement that now, I don't think it's going to be super big lift.

@bnewbold what do you think?

packages/bsky/src/api/com/atproto/label/queryLabels.ts
```diff
@@ -0,0 +1,33 @@
...
+    const { uriPatterns, sources } = params
+
+    if (uriPatterns.find((uri) => uri.includes('*'))) {
+      throw new InvalidRequestError('wildcards not supported')
+    }
+
+    const labelMap = await ctx.hydrator.label.getLabelsForSubjects(
+      uriPatterns,
+      // If sources are provided, use them. Otherwise, use the labelers from the request header
+      sources
```
foysalit, line 16: Not sure if this creates some kind of confusion because someone might pass different labelers through header and then pass different labelers through the source param and get results only for the source param.

packages/ozone/src/util.ts
```diff
@@ -45,3 +46,52 @@ export function retryableHttp(err: unknown) {
 const retryableHttpStatusCodes = new Set([
   408, 425, 429, 500, 502, 503, 504, 522, 524,
 ])
+
+export type ParsedLabelers = {
```
foysalit, line 50: This is copied from `bsky/src/utils`

packages/bsky/src/api/com/atproto/label/queryLabels.ts
devinivy: We can always ignore the header values here.

foysalit: I guess my question is, should we? to me, it feels a bit nicer to be able to just send it via header and ignore source param entirely.

packages/ozone/tests/3p-labeler.test.ts
```diff
@@ -0,0 +1,271 @@
...
+  beforeAll(async () => {
+    network = await TestNetwork.create({
+      dbPostgresSchema: 'ozone_admin_third_party_labeler',
+    })
+    ozone = network.ozone
+
+    const ozoneKey = await Secp256k1Keypair.create({ exportable: true })
+    const ozoneDid = await createOzoneDid(network.plc.url, ozoneKey)
+    thirdPartyLabeler = await TestOzone.create({
+      port: ozone.port + Math.floor(Math.random() * 10) + 1,
```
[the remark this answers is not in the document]
foysalit: I think it was causing tests to fail randomly but it may have been a separate issue, let me try reverting this to see what happens.

packages/bsky/src/api/com/atproto/label/queryLabels.ts
devinivy: In my view it's not really the purpose of the header, which is about looking at the content of the response and applying labels to it.  This endpoint doesn't really have serve any contents that can be labeled, so it's a little bit outside the purview of the header.  As you pointed out, it may also cause confusion because there's a parameter dedicated to querying specific labelers, and it's not well-defined how that parameter and the header should interact with each other.  By relying only on the dedicated parameter here, I think it makes the behavior pretty clear/predictable.  @bnewbold @dholms any thoughts to add?

foysalit: alright we're no longer defaulting to labelers in header and throwing error if sources param is missing.

packages/bsky/src/api/com/atproto/admin/getAccountInfos.ts
```diff
@@ -6,7 +6,7 @@ import { INVALID_HANDLE } from '@atproto/syntax'
 export default function (server: Server, ctx: AppContext) {
   server.com.atproto.admin.getAccountInfos({
     auth: ctx.authVerifier.optionalStandardOrRole,
-    handler: async ({ params, auth }) => {
+    handler: async ({ params, auth, req }) => {
```
devinivy, line 9: The lingering changes here can probably be removed.

packages/bsky/src/api/com/atproto/label/queryLabels.ts
```diff
@@ -0,0 +1,35 @@
...
+
+    const labelMap = await ctx.hydrator.label.getLabelsForSubjects(
+      uriPatterns,
+      // If sources are provided, use them. Otherwise, use the labelers from the request header
+      {
+        dids: sources,
+        redact: new Set(),
+      },
+    )
+    const labels = uriPatterns.map((uri) => labelMap.getBySubject(uri)).flat()
```
devinivy, line 25: Or even 👀 
```suggestion
    const labels = uriPatterns.flatMap((uri) => labelMap.getBySubject(uri))
```

packages/ozone/src/mod-service/views.ts
```diff
@@ -303,27 +310,46 @@ export class ModerationViews {
...
+      labels: [...labels, ...selfLabels, ...externalLabels],
     }
   }
 
+  async getExternalLabels(
+    subjects: string[],
+    labelers?: ParsedLabelers,
+  ): Promise<Label[]> {
+    if (!labelers?.dids.length && !labelers?.redact.size) return []
+    const auth = await this.appviewAuth(labelers)
```
devinivy, line 341: I think we can omit auth from here now.

packages/pds/src/api/com/atproto/label/queryLabels.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export default function (server: Server, ctx: AppContext) {
+  server.com.atproto.label.queryLabels({
+    auth: ctx.authVerifier.accessCheckTakedown,
+    handler: async ({ req, auth }) => {
+      const requester = auth.credentials.did
+      return pipethrough(ctx, req, requester)
+    },
+  })
+}
```
devinivy, line 13: Hey good news, we actually don't need to add every new endpoint to the pds anymore 😆 Daniel covered this in #2425, so now the PDS will proxy any lexrpc endpoint.

packages/bsky/src/api/com/atproto/admin/getAccountInfos.ts
foysalit: good eye!

packages/ozone/src/mod-service/views.ts
foysalit: yep thanks 👍

devinivy (approved):
Solid!  Before merge mind just adding a `pnpm changeset` for a patch version bump of the @atproto/api package?

PR #2448, opened 2024-04-25
:construction: Setting up groundwork for communicating to reporters on mod action

foysalit:
PLEASE DON'T MERGE!


foysalit:
We have decided to table this approach for now and will package this feature in a better way down the road.

PR #2450, opened 2024-04-26
:sparkles: Add describeServer lexicon for ozone

foysalit:
This PR introduces a new `tools.ozone.server.describeServer` route to fetch info about ozone instance. As of now, it exposes the moderators with their role/access level.

### DIFF
--- a/lexicons/tools/ozone/server/describeServer.json
+++ b/lexicons/tools/ozone/server/describeServer.json
@@ -1,0 +1,36 @@
+{
+  "lexicon": 1,
+  "id": "tools.ozone.server.describeServer",
+  "defs": {
+    "main": {
+      "type": "query",
+      "description": "Describes the server's configuration.",
+      "output": {
+        "encoding": "application/json",
+        "schema": {
+          "type": "object",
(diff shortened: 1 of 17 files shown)


bnewbold:
Just to clarify, we won't make the set of moderators public and enumerable. If we do have an endpoint to enumerate them, requiring auth, it should have a different name to avoid confusion with `com.atproto.server.describeServer`, which is public.

foysalit:
yep, this enpoint is authed so not exposing anything publicly.
makes sense to change it from `describeServer` but I was hoping this will be a generic endpoint that will fetch all the server level config and not just the moderator list so it kinda made sense from that angle. 
until we have proper user management, a dedicated endpoint just for exposing the moderator list feels a bit "extra" especially because the current list is configured through env var.

devinivy:
I think the user only needs to know their own access level, so perhaps the full list could just be boiled down to the requesting user's access level.

foysalit:
Closing this in favor of https://github.com/bluesky-social/atproto/pull/2460 which handles moderator ACL management and will create a separate PR for exposing server configs only.

PR #2452, opened 2024-04-27
Add fast path for validating non-required undefined fields

gaearon:
Creating a thousand error objects (as observed [in this post](https://bsky.app/profile/bencollins.bsky.social/post/3kr2kcufdxf2t)) is not good. Let's not do that.

The costs are shown with 6x throttling but can be higher on lower-end Android.

## Before

Screenshot 2024-04-27 at 07 02 55

Screenshot 2024-04-27 at 07 04 58

## After

Screenshot 2024-04-27 at 07 06 16

### DIFF
--- a/packages/lexicon/src/validators/complex.ts
+++ b/packages/lexicon/src/validators/complex.ts
@@ -122,6 +122,10 @@ export function object(
       if (value[key] === null && nullableProps.has(key)) {
         continue
       }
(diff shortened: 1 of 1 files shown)


gaearon:
Hmm. This fails tests due to defaults. I'll need to look at how that works. Really hope we can have a fast path for this..

pfrazee (approved):
At a glance this seems fine to me

PR #2456, opened 2024-04-30
feat: add descriptionFacets to profiles

mary-ext:
There doesn't seem to be any other blocker aside from the ability to serialize facets back into something that can be edited in the text input, which had been implemented a while back via https://github.com/bluesky-social/social-app/pull/2485.

Adding this field removes [the need for the app to parse and resolve handles](https://github.com/bluesky-social/social-app/blob/4de78fb69e095779e652c20c60b6f84a92881161/src/view/screens/Profile.tsx#L472-L500) just to display someone's profile bio.

### DIFF
--- a/lexicons/app/bsky/actor/profile.json
+++ b/lexicons/app/bsky/actor/profile.json
@@ -20,6 +20,11 @@
             "maxGraphemes": 256,
             "maxLength": 2560
           },
+          "descriptionFacets": {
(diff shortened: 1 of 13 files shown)


mary-ext:
I think there's something wrong with the lex-cli, is it reliant on Node.js returning a sorted directory listing? because the diffs is mostly just reordering files that were already there.

PR #2458, opened 2024-04-30
Appview: support alternate audiences on standard auth

devinivy:
This will allow trusted supporting services to make service requests to the appview on behalf of users.  I expect in the future we'll have a more explicit form of delegation/attenuation which would support the same use case.


dholms (approved):
perfect 👌

PR #2460, opened 2024-04-30
:sparkles: Ozone team member manager

foysalit:
This PR adds lexicons for a full set of CRUD actions to manage team members with 3 predefined roles (that bluesky currently uses).

Right now, team members are maintained via environment variables. This PR will move the storage to a table in the database. In order to maintain backward compatibility

In order to keep members configured via env vars compatible with this implementation, on boot, we iterate over those and upsert into the table as needed.


lexicons/tools/ozone/moderator/addUser.json
```diff
@@ -0,0 +1,41 @@
...
+        "encoding": "application/json",
+        "schema": {
+          "type": "ref",
+          "ref": "tools.ozone.moderator.defs#user"
+        }
+      },
+      "errors": [
+        {
+          "name": "ModeratorAlreadyExists",
+          "description": "The user is already a moderator"
```
bnewbold, line 36: probably `UserAlreadyExists` and "already a moderator" if we have a specific role called "Moderator"?

lexicons/tools/ozone/moderator/deleteUser.json
```diff
@@ -0,0 +1,30 @@
...
+          }
+        }
+      },
+      "errors": [
+        {
+          "name": "ModeratorNotFound",
+          "description": "The user being deleted is not a moderator"
+        },
+        {
+          "name": "LastAdmin",
```
bnewbold, line 24: maybe `OnlyRemainingAdmin`? "LastAdmin" would be confusing to me out of context

lexicons/tools/ozone/moderator/addUser.json
```diff
@@ -0,0 +1,41 @@
+{
+  "lexicon": 1,
+  "id": "tools.ozone.moderator.addUser",
+  "defs": {
+    "main": {
+      "type": "procedure",
+      "description": "Add a user as moderator in the ozone service.",
```
bnewbold, line 7: Would be good docs to mention in each of these descriptions what privileges are required to use the endpoint. Eg, all require auth, but maybe `listUsers` only requires an ozone user of any role?

bnewbold (commented):
general direction looks good!

packages/ozone/src/api/moderator/listUsers.ts
```diff
@@ -0,0 +1,57 @@
+import { Server } from '../../lexicon'
+import AppContext from '../../context'
+import { ProfileViewDetailed } from '../../lexicon/types/app/bsky/actor/defs'
+import { chunkArray } from '@atproto/common'
+
+// In memory cache for profiles. Current assumption is that the list shouldn't be big enough for this to be an issue in the near term
+const profileCache = new Map<string, ProfileViewDetailed>()
```
devinivy, line 7: It would be great to get this cache local to a specific `ctx` rather than free-floating, and perhaps ensure that cache items do naturally get updated over time.  Ideally we would also include these profiles on all `#user` views, not just in this list view.  So maybe we should pull the profile hydration and view logic into `moderatorService`, e.g. adjacent to `view()`.

It also may just not be necessary to cache these at all!  I'd be all for that if it simplifies things.

packages/ozone/src/moderator/index.ts
```diff
@@ -0,0 +1,188 @@
...
+
+  view(moderator: Selectable<Moderator>): User {
+    return {
+      id: `${moderator.id}`,
+      did: moderator.did,
+      role: moderator.role,
+      disabled: moderator.disabled,
+      createdAt: moderator.createdAt,
+      updatedAt: moderator.updatedAt,
+      lastUpdatedBy: moderator.lastUpdatedBy,
```
devinivy, line 185: These aren't part of the `#user` lexicon, but we may want to include `handle` and `profile`.

packages/ozone/tests/moderators.test.ts
```diff
@@ -0,0 +1,140 @@
...
+      identifier: sc.accounts[sc.dids.carol].handle,
+      password: sc.accounts[sc.dids.carol].password,
+    })
+  })
+
+  afterAll(async () => {
+    await network.close()
+  })
+
+  describe.only('listUsers', () => {
```
devinivy, line 38:
```suggestion
  describe('listUsers', () => {
```

packages/ozone/src/api/moderator/updateUser.ts
```diff
@@ -0,0 +1,42 @@
+import { AuthRequiredError, InvalidRequestError } from '@atproto/xrpc-server'
+import { Server } from '../../lexicon'
+import AppContext from '../../context'
+import { getUserRole } from '../util'
+
+export default function (server: Server, ctx: AppContext) {
+  server.tools.ozone.moderator.updateUser({
+    auth: ctx.authVerifier.moderator,
```
devinivy, line 8: I think this is right, basing on the others:
```suggestion
    auth: ctx.authVerifier.modOrAdminToken,
```

packages/ozone/src/api/moderator/deleteUser.ts
```diff
@@ -0,0 +1,27 @@
...
+          'Must be an admin to delete a moderator user',
+        )
+      }
+      const moderatorService = ctx.moderatorService(db)
+
+      const canDelete = await moderatorService.canDelete(did)
+
+      if (canDelete) {
+        await moderatorService.delete(did)
+      }
```
devinivy, line 24: Since `canDelete()` can only return `true` or fail, I'd recommend renaming it to something like `assertCanDelete()` and not returning anything from it.

Another trick for these kinds of things, though, is to create a transaction, perform the deletion, check a post condition, and roll the postcondition back if it fails.  It can protect from races more effectively than checking a precondition.  Roughly:

```ts
await db.transaction(async (txn) => {
  const moderatorTxn = ctx.moderatorService(txn)
  await moderatorTxn.delete(did)
  await moderatorTxn.assertAdminExists() // throws if there are no admins, which causes transaction to rollback
})
```

packages/ozone/src/api/moderator/addUser.ts
```diff
@@ -0,0 +1,41 @@
+import { AuthRequiredError, InvalidRequestError } from '@atproto/xrpc-server'
+import { Server } from '../../lexicon'
+import AppContext from '../../context'
+import { getUserRole } from '../util'
+
+export default function (server: Server, ctx: AppContext) {
+  server.tools.ozone.moderator.addUser({
+    auth: ctx.authVerifier.moderator,
```
devinivy, line 8:
```suggestion
    auth: ctx.authVerifier.modOrAdminToken,
```

packages/ozone/src/api/moderator/updateUser.ts
```diff
@@ -0,0 +1,42 @@
...
+
+      const newUser = await moderatorService.update(did, {
+        disabled,
+        role: getUserRole(role),
+        lastUpdatedBy: access.iss,
+      })
+
+      return {
+        encoding: 'application/json',
+        body: moderatorService.view(newUser),
```
devinivy, line 38:
```suggestion
      const updatedUser = await moderatorService.update(did, {
        disabled,
        role: getUserRole(role),
        lastUpdatedBy: access.iss,
      })

      return {
        encoding: 'application/json',
        body: moderatorService.view(updatedUser),
```

devinivy (commented):
This is quite the chunk of work!  Good stuff, going to be a big upgrade 👍

lexicons/tools/ozone/moderator/addUser.json
```diff
@@ -0,0 +1,41 @@
...
+          }
+        }
+      },
+      "output": {
+        "encoding": "application/json",
+        "schema": {
+          "type": "ref",
+          "ref": "tools.ozone.moderator.defs#user"
+        }
+      },
```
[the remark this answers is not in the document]
foysalit: we discussed most of these outside of github but leaving some traces of that here.
I kinda like the entity specific namespace since it isolates the ACL management in one place. `tools.ozone.admin` will probably end up with a lot more than ACL management, eventually.
also, in my mind, all users in ozone are moderators so, the fact that we have a role named `moderator` is the only confusing part. could we rename that to something like `escalation`? I like the term moderator instead of user because it simplifies the definition of ozone users in the sense that any `user` in the network can be a `moderator`. 

with that, my counter proposal to clarify and simplify the terminologies:
- the namespace remains `tools.ozone.moderator` and we ditch the term user everywhere so, `tools.ozone.moderator.addUser` becomes `tools.ozone.moderator.add`
- rename the `modRoleModerator` to `modRoleEscalation` (probably needs a better name)

lexicons/tools/ozone/moderator/deleteUser.json
```diff
@@ -0,0 +1,30 @@
...
+      },
+      "errors": [
+        {
+          "name": "ModeratorNotFound",
+          "description": "The user being deleted is not a moderator"
+        },
+        {
+          "name": "OnlyRemainingAdmin",
+          "description": "There must always be at least one moderator with admin access"
+        }
```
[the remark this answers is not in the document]
foysalit: oh right, yeah then we just don't need this error at all.

lexicons/tools/ozone/moderator/listUsers.json
```diff
@@ -0,0 +1,39 @@
+{
+  "lexicon": 1,
+  "id": "tools.ozone.moderator.listUsers",
+  "defs": {
+    "main": {
+      "type": "query",
+      "description": "List all users with access to the ozone service. Any ozone user can access this endpoint.",
```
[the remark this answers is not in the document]
foysalit: hmm... probably no need for that to be guaranteed.

packages/ozone/src/db/migrations/20240521T211332580Z-moderators.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export async function up(db: Kysely<unknown>): Promise<void> {
+  await db.schema
+    .createTable('moderator')
+    .addColumn('id', 'serial', (col) => col.primaryKey())
+    .addColumn('did', 'varchar', (col) => col.notNull().unique())
+    .addColumn('role', 'varchar', (col) => col.notNull())
+    .addColumn('disabled', 'boolean', (col) => col.defaultTo(false).notNull())
+    .addColumn('createdAt', 'timestamptz')
+    .addColumn('updatedAt', 'timestamptz')
```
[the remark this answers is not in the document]
foysalit: we don't usually seem to define these as non-nullables in other tables though?

packages/ozone/src/db/migrations/20240521T211332580Z-moderators.ts
```diff
@@ -0,0 +1,18 @@
+import { Kysely } from 'kysely'
+
+export async function up(db: Kysely<unknown>): Promise<void> {
+  await db.schema
+    .createTable('moderator')
+    .addColumn('id', 'serial', (col) => col.primaryKey())
+    .addColumn('did', 'varchar', (col) => col.notNull().unique())
```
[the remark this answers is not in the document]
foysalit: don't think that's exactly necessary but not exactly seeing an issue with keeping it either.

packages/ozone/src/api/moderator/addUser.ts
foysalit: we try to keep track of requester's DID in `lastUpdatedBy` but with admin tokens, we don't get a DID. wondering if we should just store something like `admin_token` in `lastUpdatedBy` when these requests are made with admin token?

lexicons/tools/ozone/moderator/addUser.json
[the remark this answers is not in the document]
bnewbold: We use `actor` in a couple other Lexicons, seems like it would be a good fit here? `tools.ozone.actor.*`. I also like "Member" as in "team member"... that also points to `tools.ozone.team.*`.

In all these terms we need to differentiate between a general atproto "user" / "account" from somebody who can log in to the Ozone instance specifically.

To propose something specific:

* use the namespace `tools.ozone.team.*`
* use the team "member" for any account which is part of the team
* keep the existing triage / moderator / admin terminology for now.

packages/ozone/src/db/migrations/20240521T211332580Z-moderators.ts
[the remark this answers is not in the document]
devinivy: I think we do on `moderation_subject_status` and `moderation_event`, but don't on `communication_template`.

packages/ozone/src/api/moderator/addUser.ts
devinivy: Ah interesting!  Yeah, that could work.

packages/ozone/src/moderator/index.ts
```diff
@@ -0,0 +1,188 @@
...
+    return {
+      isModerator,
+      isAdmin,
+      isTriage,
+    }
+  }
+
+  view(moderator: Selectable<Moderator>): User {
+    return {
+      id: `${moderator.id}`,
```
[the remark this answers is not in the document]
foysalit: yep, decided to just get rid of the id.

packages/ozone/src/moderator/index.ts
foysalit: Decided to get rid of the handle since that's available in profile.

foysalit:
alright we should be close now. a few major things to review
1. the auth-verifier now gives the serviceDid to always access as admin
2. update and delete actions are no longer allowed on serviceDid
3. `moderator` terminology has been updated everywhere with `member` instead
4. profile hydration has been moved into the team service and applied on add, update and list endpoints
5. profile views are no longer cached and always fetched from appview on demand

packages/ozone/src/team/index.ts
```diff
@@ -0,0 +1,208 @@
...
+    let builder = this.db.db.selectFrom('member').selectAll()
+    if (cursor) {
+      builder = builder.where('createdAt', '>', new Date(cursor))
+    }
+    const members = await builder
+      .limit(limit)
+      .orderBy('createdAt', 'asc')
+      .execute()
+
+    return { members, cursor: members.slice(-1)[0]?.createdAt.toISOString() }
```
devinivy, line 36:
```suggestion
    return { members, cursor: members.at(-1)?.createdAt.toISOString() }
```

packages/ozone/src/team/index.ts
```diff
@@ -0,0 +1,208 @@
...
+      .insertInto('member')
+      .values({
+        role,
+        did,
+        lastUpdatedBy,
+        disabled: false,
+        updatedAt: now,
+        createdAt: now,
+      })
+      .onConflict((oc) => oc.column('did').doUpdateSet({ role }))
```
devinivy, line 86: Should we maintain `updatedAt` and `lastUpdatedBy` here too?

packages/ozone/src/team/index.ts
```diff
@@ -0,0 +1,208 @@
...
+      >
+    >,
+  ): Promise<Selectable<Member>> {
+    const updatedMember = await this.db.db
+      .updateTable('member')
+      .where('did', '=', did)
+      .set({
+        ...updates,
+        updatedAt: updates.updatedAt || new Date(),
+      })
```
devinivy, line 105: There is a bit of a security issue here, since `updates` may contain fields not explicitly `Pick`ed out of `Selectable<Member>`, and they end up passed directly through to the query.  For example, imagine if `did` were a field on `updates`.

Looking at the `team.updateMember` endpoint, I think the user could end-up successfully changing the user's `createdAt` time through the API.  On its own it that isn't a critical issue, but as we add new columns to the `members` table we might run into more sensitive issues.

packages/ozone/src/auth-verifier.ts
```diff
@@ -113,9 +108,30 @@ export class AuthVerifier {
...
+        credentials: {
+          type: 'standard',
+          iss,
+          aud: payload.aud,
+          isAdmin: true,
+          isModerator: true,
+          isTriage: true,
+        },
+      }
+    }
```
devinivy, line 124: Perhaps we could avoid this exception if we folded the account into `seedInitialMembers()`.

packages/ozone/src/auth-verifier.ts
```diff
@@ -113,9 +108,30 @@ export class AuthVerifier {
...
+
+    const member = await this.teamService.getMember(iss)
+
+    if (member?.disabled) {
+      throw new AuthRequiredError('member is disabled', 'MemberDisabled')
+    }
+
+    const { isAdmin, isModerator, isTriage } =
+      await this.teamService.getMemberRole(member)
```
devinivy, line 133:
```suggestion
    const { isAdmin, isModerator, isTriage } =
      this.teamService.getMemberRole(member)
```

packages/ozone/src/auth-verifier.ts
devinivy: You know, the more I think about it we may want to keep the deletion guard but not add any other exceptions for the service DID.  I think there could be legitimate reasons for a deployment to not want to make the service DID a team member.

For the Ozone distribution we configure the service DID to be an admin anyway, so most users will have this by default.  Perhaps the deletion rule should be to ensure you don't delete or demote yourself.  That should be enough to ensure there's no way to lock yourself out.

bnewbold:
Looking forward to this! Let me know when I can give it a spin in staging.

PR #2461, opened 2024-05-02
Appview: add grandparent author to reply ref, ensure no replies to blocked grandparent in feeds

devinivy:
Two changes here:
 - introduces a new field on feed items `item.reply.grandparentAuthor` which contains a basic profile for the grandparent (i.e. parent's parent) reply of a post.
 - ensures that in feed settings (e.g. timeline), posts whose reply grandparent author is blocked do not appear.  This essentially applies the block rule to `item.reply.parent` that #2430 applied to `item.post`, if that makes sense.

Replaces #2229.

### DIFF
--- a/lexicons/app/bsky/feed/defs.json
+++ b/lexicons/app/bsky/feed/defs.json
@@ -68,6 +68,11 @@
         "parent": {
           "type": "union",
           "refs": ["#postView", "#notFoundPost", "#blockedPost"]
+        },
+        "parentReplyAuthor": {
+          "type": "ref",
+          "ref": "app.bsky.actor.defs#profileViewBasic",
+          "description": "When parent is a reply to another post, this is the author of that post."
         }
       }
     },
--- a/lexicons/app/bsky/feed/defs.json
+++ b/lexicons/app/bsky/feed/defs.json
@@ -69,7 +69,7 @@
           "type": "union",
           "refs": ["#postView", "#notFoundPost", "#blockedPost"]
         },
-        "parentReplyAuthor": {
+        "grandparentAuthor": {
           "type": "ref",
           "ref": "app.bsky.actor.defs#profileViewBasic",
           "description": "When parent is a reply to another post, this is the author of that post."
--- a/packages/api/src/client/lexicons.ts
+++ b/packages/api/src/client/lexicons.ts
@@ -4833,7 +4833,7 @@ export const schemaDict = {
               'lex:app.bsky.feed.defs#blockedPost',
             ],
           },
-          parentReplyAuthor: {
+          grandparentAuthor: {
             type: 'ref',
             ref: 'lex:app.bsky.actor.defs#profileViewBasic',
             description:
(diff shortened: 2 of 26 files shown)


dholms (approved):
looks great 🙌

mackuba:
Please don't do this ☹️ If I understand it correctly, this means that if some user blocks me, I will not only not see their posts (which is expected), and not only someone's comments to them, but also someone's comments to those comments. This goes way further than is expected IMHO.

So imagine a case like this: we have a user, let's call them Karen, who has blocked me. Karen asks a question about the AT Protocol. Shreyan (who I follow) answers that question using the word "ATProto". Daniel (who I also follow) chimes in and clarifies the answer, also mentioning "ATProto". Now, because Karen has blocked me, that comment by Shreyan and the reply by Daniel, who are both my mutuals, will not appear either in my Following timeline or in the ATProto feed which I follow. These are two comments by people I both follow that are both relevant to the ATProto-themed feed I'm interested in, but because I'm blocked by Karen, these posts won't be visible to me in any of the main feeds. What's more, because this is now being implemented at the AppView level, I won't even be able to make a customized client app that chooses to show these comments. Which means some interesting discussions on topics that interest me between people I follow will be hidden from me…

Bluesky is all about choice and letting you customize your experience and what you want and don't want to see. Why should the decision to hide these posts from my view be made by Karen in this case and not by me? She has a right to hide her own posts from me, but not my friends' replies to one another. My opinion is that these posts should not be filtered out at the AppView level, they should be returned in the API response and there should be client-side configuration that should let me choose to see all posts that aren't directly blocked in my feeds if that is my preference (even if that's not the default setting).

MasterJ93:
This is a fair point that mackuba is saying. I understand that my example could potentially be extreme, but what mackuba says caused me some concern that I've been worried could happen:

> Why should the decision to hide these posts from my view be made by Karen in this case and not by me?

I fear that this can possibly lead to abusive scenarios where someone could simply block you, then decide to follow you around and begin saying whatever they want to say, thereby disallowing you to even see many of the replies (as a show of hatred). I've always been wary of having things like this be done, so I don't think it would be a great idea to implement it in this sort of way at this time. Perhaps there can be a boolean setting that bypasses that. Then, you have the choice of allowing this sort of thing to happen.

I think this should probably be reverted for now and a rethink of a better solution could be possible.

mackuba:
I realize that changes like this and the earlier change to hide direct replies to a blocked parent are in response to many people saying "I keep seeing my friends replying to someone I blocked, I don't want to be seeing those, please fix this". And this is a perfectly good (and probably common) position to have! Like I said, Bluesky is all about letting you control what you want to see and not see. And a lot of people want to avoid seeing something that might be potentially annoying, even at the cost of missing some relevant content. But other people like me want to choose a different tradeoff, and this kind of implementation forces this other choice on me - because you can hide something unwanted, but you can't show something that isn't there at all.

devinivy:
Totally heard!  We'll take a closer look at this behavior 👍

mackuba:
Thanks! 🙌

PR #2467, opened 2024-05-05
api: Always apply self labels even if post is your own, downgrade `hide` moderation preference to `blur` when the post is your own

haileyok:
Fixes: https://github.com/bluesky-social/social-app/issues/3871

## Why

This is one part of adjusting the difficulty in seeing whether your post is actually labeled or not whenever you apply a self label to the post. Take this scenario:

- You post an image and add the self-label `nudity`
- The post gets upgraded to the `porn` label immediately. To the user, this Bluesky upgraded label appears to be _their_ self label
- If the user realizes that the label was upgraded, they appeal the upgrade
- Bluesky moderation accepts the appeal, and downgrades it back to `nudity`
- To the poster, the UI makes it look like _all_ labels were removed from their post, even though they were not.

Another, simpler scenario where this is problematic:

- You post an image and add the self-label `nudity`
- Your moderation preferences are set to blur nudity
- The posts that you labeled as `nudity` on your timeline are _not_ blurred, while other people's are. This in some ways defeats the purpose of being able to "blur" NSFW content.

This also allows for showing the "X labels have been placed on this content" even for self-labels so that there can be less confusion about what is going on for the user. https://github.com/bluesky-social/social-app/pull/3874

## How

Instead of returning a default `ModerationUI` whenever the post is yours, we should always check for labels when the c context is `contentView` or `contentMedia`. In those cases, we should also downgrade the `hide` setting to `warn` so the user still sees their post.

## Test Plan

Make a post and self label it as something like nudity. Toggle the various settings and check your following feed for the post.

https://github.com/bluesky-social/atproto/assets/153161762/5fc9f111-85af-4527-bed0-161b7760ede3

### DIFF
--- a/.changeset/clever-trainers-own.md
+++ b/.changeset/clever-trainers-own.md
@@ -1,0 +1,5 @@
+---
+"@atproto/api": patch
+---
+
+Modify label-handling on user's own content to still apply blurring
--- a/packages/api/src/moderation/decision.ts
+++ b/packages/api/src/moderation/decision.ts
@@ -78,33 +78,15 @@ export class ModerationDecision {
 
   ui(context: keyof ModerationBehavior): ModerationUI {
     const ui = new ModerationUI()
-    // Always apply labels on posts even if the user is the author. Also, we
-    // want to downgrade the `hide` setting to `warn` in these cases so the
-    // user is always aware of the label.
-    if (this.isMe) {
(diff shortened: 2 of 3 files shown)


pfrazee (approved):
Tweaked the approach to be a little bit cleaner, going to merge & publish

PR #2470, opened 2024-05-08
Dev-env introspection server

dholms:
Adds an optional introspection server to the dev-env so that other packages have an entrypoint to join the network. 

For now, it's just a simple server with only one route that supplies the url & did of relevant services in the network

### DIFF
--- a/packages/dev-env/src/introspect.ts
+++ b/packages/dev-env/src/introspect.ts
@@ -50,4 +50,5 @@ export class IntrospectServer {
   async close() {
     this.server.close()
     await events.once(this.server, 'close')
+  }
 }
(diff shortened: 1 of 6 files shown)


packages/dev-env/src/introspect.ts
```diff
@@ -0,0 +1,61 @@
...
+    return new Promise<void>((resolve, reject) => {
+      this.server.close((err) => {
+        if (err) {
+          reject(err)
+        } else {
+          resolve()
+        }
+      })
+    })
+  }
```
devinivy, line 60: Since we have the `await events.once(server, 'listening')` line above—another way to slice this is:
```suggestion
    this.server.close()
    await events.once(this.server, 'close')
```

devinivy (approved):

packages/dev-env/src/introspect.ts
dholms: ah nice I dig 👍

PR #2471, opened 2024-05-08
Appview: add chat details to profile views, chronological likes

devinivy:
Updating the appview dataplane protobufs with support for chat details and time-sorted likers.
 - Add `profile.associated.chat.allowIncoming` to profile views.
 - Update `getLikes` to be ordered reverse-chronologically.

### DIFF
--- a/packages/bsky/src/hydration/hydrator.ts
+++ b/packages/bsky/src/hydration/hydrator.ts
@@ -669,6 +669,7 @@ export class Hydrator {
         (await this.label.getLabelers([did], includeTakedowns)).get(did) ??
         undefined
       )
+      // @TODO use lexicon constant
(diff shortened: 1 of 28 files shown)


dholms (commented):
sweet looks good to me 👌

packages/bsky/src/hydration/hydrator.ts
```diff
@@ -669,6 +669,13 @@ export class Hydrator {
         (await this.label.getLabelers([did], includeTakedowns)).get(did) ??
         undefined
       )
+    } else if (collection === 'chat.bsky.actor.declaration') {
```
dholms, line 672: TODO here as well for switching it out for an `id` 👌

PR #2483, opened 2024-05-13
🚧 OAuth - Client SDK

matthieusieben:
This is a new implementation of the `@aproto/api` package that makes it ready for being used with OAuth based credentials.


packages/api/README.md
```diff
@@ -132,6 +180,9 @@ if (AppBskyFeedPost.isRecord(post)) {
 }
 ```
 
+Note that the agent class will validate records for you, so you don't need to do
+this before calling the API.
+
```
devinivy, line 185: I'm not sure that this is this case.  The `@atproto/xrpc` package doesn't validate params/inputs to my knowledge, and if it did then it wouldn't necessarily validate specific record types passed to e.g. `com.atproto.repo.createRecord`.

packages/api/src/types.ts
```diff
@@ -4,7 +4,7 @@ import { ModerationPrefs } from './moderation/types'
 /**
  * Supported proxy targets
  */
-export type AtprotoServiceType = 'atproto_labeler'
+export type AtprotoServiceType = string // 'atproto_labeler'
```
devinivy, line 7: Was this change intended to stick around?  Maybe we can split the difference and go with something like `'atproto_labeler' | (string & {})`.

.changeset/funny-points-cover.md
```diff
@@ -0,0 +1,5 @@
+---
+"@atproto/xrpc-server": minor
+---
+
+Allow upload of payload of zero bytes (e.g. empty txt file)
```
devinivy, line 5: I think this has been supported since https://github.com/bluesky-social/atproto/pull/2599

I don't see any material changes to xrpc-server here, so I think we can nix this changeset.

.changeset/honest-hornets-approve.md
```diff
@@ -0,0 +1,5 @@
+---
+"@atproto/lexicon": patch
+---
+
+Add the ability to instantiate a Lexicon from an iterable, and to use a Lexicon as iterable.
```
devinivy, line 5: This landed in https://github.com/bluesky-social/atproto/pull/2482 but didn't get a changeset/publish, so going to preserve this.

.changeset/purple-pans-shout.md
```diff
@@ -0,0 +1,35 @@
+---
+"@atproto/lex-cli": minor
+"@atproto/xrpc": minor
+"@atproto/api": minor
+"@atproto/xrpc-server": patch
```
devinivy, line 5: There are no changes to xrpc-server— it will get a patch by virtue of its dependencies being updated.
```suggestion
```

.changeset/honest-hornets-approve.md
matthieusieben: This was left over from a rework of the git branches.  Can be removed indeed!

packages/api/src/types.ts
matthieusieben: I don't remember exactly why I did this. I am 95% sure that if I did this, it must have been because I got an error somewhere (ozone repo?).

I'm in favor of reverting it if we can't find the source.

packages/api/README.md
matthieusieben: Hum. Not sure why I wrote that. I guess I was daydreaming about my own view of the ideal behavior of the xrpc client's call implementation 😜

packages/xrpc/src/util.ts
```diff
@@ -85,100 +111,343 @@ export function encodeQueryParam(
...
+    },
+    async cancel() {
+      await generator.return()
+    },
+  })
 }
 
+// Browsers don't have Buffer. This syntax is to avoid bundlers from including
+// a Buffer polyfill in the bundle if it's not used elsewhere.
+const globalName = `${{ toString: () => 'Buf' }}fer` as 'Buffer'
```
devinivy, line 392: This is giving some build tools some issues.  Next.js projects choke on it during build, for example, here: https://github.com/vercel/nft/blob/a930bc78911aa1af6326382038730461ae828a8b/src/utils/static-eval.ts#L525
```
TypeError: Cannot convert object to primitive value
```


Is this common bundler functionality that we indeed do need to circumvent?  If so, is there another way we could do it that doesn't cause issues for Next.js?

matthieusieben: This purpose of this is to support using an iterable as XRPC data on most environments, by providing a polyfill for `ReadableStream.from`, when `ReadableStream` is available.

I've remove this code and replaced it with a comment recommending the user of the lib to provide their own polyfill for `ReadableStream.from` if they need to support using iterables as XRPC request body.

https://github.com/bluesky-social/atproto/pull/2483/commits/3681c907b29762e2c4ab965894d02a67484b85e5

matthieusieben: https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream/from_static

Note that chrome does not support `ReadableStream.from`:
https://issues.chromium.org/issues/40267032

devinivy: Coolcool!  Not sure if it makes a difference, but it was specifically this line where I bumped into the issue:
```ts
const globalName = `${{ toString: () => 'Buf' }}fer` as 'Buffer'
```

The build tool was trying to do some static evaluation and didn't know what to do with `{ toString: () => 'Buf' }` within the tagged template literal.

matthieusieben: Which was my exact intent when writing this XD

I expected that bundlers would just ignore unexpected patterns instead of throwing. My assumption being wrong, I just removed this.

matthieusieben:
I was able to run tests successfully locally

.changeset/pre.json
```diff
@@ -0,0 +1,48 @@
...
+    "@atproto/xrpc": "0.5.0",
+    "@atproto/xrpc-server": "0.5.3"
+  },
+  "changesets": [
+    "brown-dryers-fail",
+    "honest-hornets-approve",
+    "purple-pans-shout",
+    "rare-feet-happen"
+  ]
+}
```
matthieusieben, line 48: Should this file be remove since I also removed every other mention to rc* versions (in both CHANGELOG.md and package.json) ?

devinivy: Yeah, that should be fine—I am maintaining a branch now for future RCs.

.changeset/purple-pans-shout.md
```diff
@@ -0,0 +1,537 @@
...
+  during the instance creation. Attempting to set/get the `fetch` property will
+  now result in an error.
+
+- The `fetch()` method, as well as WhatWG compliant `Request` and `Headers`
+  constructors, must be globally available in your environment. Use a polyfill
+  if necessary.
+
+- The `AtpBaseClient` has been removed. The `AtpServiceClient` has been renamed
+  `AtpBaseClient`. Any code using either of these classes will need to be
+  updated.
```
bnewbold, line 62: This reads a little confusing: there is still a `AtpBaseClient`, but the old implementation was removed and replaced with what was formerly called `AtpServiceClient`?

matthieusieben: Yes! I guess I could revert the renaming ...

bnewbold: I don't mean change the code, just re-phrase this sentence.

devinivy (approved):

PR #2488, opened 2024-05-13
Update legacy blob refs in profile records

dholms:
This is a relatively crude/ham-fisted fix intended to help with the fact that some accounts still have legacy blob refs in their profile records which is preventing them from updating their profile record.

Ended up landing this in `putRecord` instead of the `prepareWrite` functions because we need access to the actor store to find the `size` of the blob. We only implement it in `putRecord` when updating specifically a profile record in order to both reduce the number of checks we're required to do, but also because this issue is mostly isolated to historical profile records.

This is a stopgap until we have time to run a proper migration that updates all legacy blob refs.

### DIFF
--- a/packages/pds/src/api/com/atproto/repo/putRecord.ts
+++ b/packages/pds/src/api/com/atproto/repo/putRecord.ts
@@ -2,6 +2,7 @@ import { CID } from 'multiformats/cid'
 import { AtUri } from '@atproto/syntax'
 import { AuthRequiredError, InvalidRequestError } from '@atproto/xrpc-server'
 import { CommitData } from '@atproto/repo'
+import { BlobRef } from '@atproto/lexicon'
 import { Server } from '../../../../lexicon'
 import { prepareUpdate, prepareCreate } from '../../../../repo'
 import AppContext from '../../../../context'
@@ -12,6 +13,9 @@ import {
   PreparedCreate,
   PreparedUpdate,
 } from '../../../../repo'
+import { ids } from '../../../../lexicon/lexicons'
+import { Record as ProfileRecord } from '../../../../lexicon/types/app/bsky/actor/profile'
+import { ActorStoreTransactor } from '../../../../actor-store'
 
 export default function (server: Server, ctx: AppContext) {
   server.com.atproto.repo.putRecord({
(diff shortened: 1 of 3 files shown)


packages/pds/src/api/com/atproto/repo/putRecord.ts
```diff
@@ -61,6 +65,14 @@ export default function (server: Server, ctx: AppContext) {
         did,
         async (actorTxn) => {
           const current = await actorTxn.record.getRecord(uri, null, true)
+          const isUpdate = current !== null
+
+          // @TODO temporaray hack for legacy blob refs in profiles - remove after migrating legacy blobs
+          if (isUpdate && collection === ids.AppBskyActorProfile) {
+            console.log('HERE')
```
devinivy, line 72:
```suggestion
```

packages/pds/src/api/com/atproto/repo/putRecord.ts
```diff
@@ -61,6 +65,14 @@ export default function (server: Server, ctx: AppContext) {
...
+            await updateProfileLegacyBlobRef(actorTxn, record)
+            // console.log(record.avatar?.original)
```
devinivy, line 74:
```suggestion
```

packages/pds/src/api/com/atproto/repo/putRecord.ts
```diff
@@ -122,3 +134,26 @@ export default function (server: Server, ctx: AppContext) {
...
   })
 }
+
+// WARNING: mutates object
+const updateProfileLegacyBlobRef = async (
+  actorStore: ActorStoreTransactor,
+  record: ProfileRecord,
+) => {
+  if (record.avatar && !record.avatar.original['$type']) {
+    const blob = await actorStore.repo.blob.getBlobMetadata(record.avatar.ref)
```
devinivy, line 144: Just double-checking that we like the behavior of throwing here if the blob is not found.

devinivy (approved):
Looking good 👍

packages/pds/src/api/com/atproto/repo/putRecord.ts
dholms: Yup think so 👍 

I believe that's the behavior we'd do you tried to update a record and were missing a blob anyways 👍

devinivy (approved):

PR #2490, opened 2024-05-14
Ozone chat integration

dholms:
Adds functionality to Ozone to handle integration with a `chat.bsky` service.

Specifically:
- handling of reports/views on message subjects
- some extra authorization (ensure moderation access) around special tags such as `chat-disabled`
- proxying of `chat.bsky.moderation.getActorMetadata` and `chat.bsky.moderation.getMessageContext`
- additional checks on `chat.bsky.moderation.getMessageContext` to limit the size of the context window that mods can see based on authorization level & ensure that the message being fetched actually corresponds to a user report


packages/ozone/src/api/moderation/emitEvent.ts
```diff
@@ -214,6 +220,32 @@ export default function (server: Server, ctx: AppContext) {
...
+        throw new Error(`Must be an admin to add tag: ${tag}`)
+      } else if (
+        TAG_AUTH[tag] === 'moderator' &&
+        !auth.credentials.isModerator
+      ) {
+        throw new Error(`Must be a full moderator to add tag: ${tag}`)
+      }
+    }
+  }
+  return true
```
devinivy, line 246: No real issue, just think this can go.
```suggestion
```

packages/ozone/src/db/migrations/20240506T225055595Z-message-subject.ts
```diff
@@ -0,0 +1,22 @@
...
+    .column('subjectMessageId')
+    .execute()
+}
+
+export async function down(db: Kysely<unknown>): Promise<void> {
+  await db.schema
+    .alterTable('moderation_event')
+    .dropColumn('subjectMessageId')
+    .execute()
+  await db.schema.dropIndex('moderation_event_message_id_index').execute()
```
devinivy, line 21: I think dropping the column causes the index to be dropped too.

devinivy (approved):
Looks great 👍

packages/ozone/src/api/moderation/emitEvent.ts
dholms: ah good looks - used to be an `is` function but converted it to `assert`

packages/ozone/src/db/migrations/20240506T225055595Z-message-subject.ts
dholms: sweet yup 👍

PR #2492, opened 2024-05-14
Add new preference and api for bsky app state; also put preference updates within transactional lock regions

pfrazee:
- Adds `#bskyAppStatePref` preference
- Adds `bskyAppQueueNudges(nudges: string|string[])`
- Adds `bskyAppDismissNudges(nudges: string|string[])`
- Adds `bskyAppSetActiveProgressGuide(guide: {guide: string, ...})`
- Wraps `updatePreferences` in a lock region to avoid transaction clobbering in the future

To be used by the bsky app to track nudges that the user can dismiss (things like "Introducing DMs!") and to run progress guides. Not meant to be used by other clients, so explicitly namespaced with `bksyApp` within the API.

### DIFF
--- a/.changeset/strange-camels-float.md
+++ b/.changeset/strange-camels-float.md
@@ -1,0 +1,8 @@
+---
+"@atproto/ozone": patch
+"@atproto/bsky": patch
+"@atproto/api": patch
+"@atproto/pds": patch
+---
+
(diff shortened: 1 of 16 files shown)


gaearon (approved):
lock makes sense

PR #2498, opened 2024-05-17
Appview: support handles in post threads, remove deleted reply placeholders

devinivy:
Two changes here to`app.bsky.feed.getPostThread` in the appview:
 - support handles in the `uri` parameter (no longer just DIDs), e.g. `at://handle.bsky.social/app.bsky.feed.post/22222222222`.
 - Do not serve `#notFoundPost` placeholders in thread replies.  There were appearing from posts by takendown/suspended accounts.

### DIFF
--- a/packages/bsky/src/api/app/bsky/feed/getPostThread.ts
+++ b/packages/bsky/src/api/app/bsky/feed/getPostThread.ts
@@ -1,4 +1,5 @@
 import { InvalidRequestError } from '@atproto/xrpc-server'
+import { AtUri } from '@atproto/syntax'
 import { Server } from '../../../../lexicon'
 import { isNotFoundPost } from '../../../../lexicon/types/app/bsky/feed/defs'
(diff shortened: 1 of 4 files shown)


packages/bsky/src/api/app/bsky/feed/getPostThread.ts
```diff
@@ -118,3 +120,16 @@ type Skeleton = {
   anchor: string
   uris: string[]
 }
+
+const resolveUri = async (ctx: Context, uriStr: string) => {
+  const uri = new AtUri(uriStr)
+  if (uri.host.startsWith('did:')) {
```
dholms, line 126: `getDids` actually already does this check 👍 

this may be a nice function to stick on _something_ so we can reuse it around other routes, but we can always do that once we start scattering it around a bit more

dholms (approved):
nice! 👌

PR #2499, opened 2024-05-17
Lexicon: misc fixes to chat lexicons

devinivy:
A few small `chat.bsky` lexicon tweaks in here:
 - Add temporarily optional `convoId` to `getMessageContext`.  Looking to be consistent about always qualifying `messageId` within a given `convoId`.
 - Rename `#message` to `#messageInput` and remove its optional `id` field.
 - Add a limit to the number of messages in `sendMessageBatch`.
 - Add required `convoId` to `#messageRef` used by ozone, make requisite ozone changes.

### DIFF
--- a/.changeset/heavy-pears-sip.md
+++ b/.changeset/heavy-pears-sip.md
@@ -1,0 +1,5 @@
+---
+"@atproto/api": patch
+---
+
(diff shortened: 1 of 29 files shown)


dholms (approved):
looks great 🙌

PR #2506, opened 2024-05-21
remove deprecated sync fields

bnewbold:
Part of splitting up https://github.com/bluesky-social/atproto/pull/2376 in to smaller chunks for easier merging.

These have been marked deprecated for a long time.

Removes from client and API-side at the same time. These are fairly low-level and not used by, eg, the app.


dholms (approved):
looks good 👍

PR #2510, opened 2024-05-23
Correctly proxy procedures to chat

dholms:


lexicons/com/atproto/server/createAppPassword.json
```diff
@@ -14,6 +14,10 @@
             "name": {
               "type": "string",
               "description": "A short name for the App Password, to help distinguish them."
+            },
+            "scopes": {
+              "type": "array",
+              "items": { "type": "string" }
```
devinivy, line 20: Might recommend eventually putting some basic max length limit here, even if it's high like `100`.

dholms: oh oops this wasn't supposed to make it in there 🤦

devinivy (approved):

PR #2515, opened 2024-05-23
Privileged app passwords

dholms:
Adds in a notion of a "privileged" app password which has greater access than a normal app password - for instance to sensitive account state such as chats - but still less access than a full access token.

### DIFF
--- a/packages/pds/src/account-manager/db/schema/app-password.ts
+++ b/packages/pds/src/account-manager/db/schema/app-password.ts
@@ -3,6 +3,7 @@ export interface AppPassword {
   name: string
   passwordScrypt: string
   createdAt: string
+  privileged: boolean | null
 }
 
 export const tableName = 'app_password'
--- a/packages/pds/src/account-manager/db/schema/app-password.ts
+++ b/packages/pds/src/account-manager/db/schema/app-password.ts
@@ -3,7 +3,7 @@ export interface AppPassword {
   name: string
   passwordScrypt: string
   createdAt: string
-  privileged: 0 | 1 | null
+  privileged: 0 | 1
 }
 
 export const tableName = 'app_password'
--- a/packages/pds/src/account-manager/db/schema/app-password.ts
+++ b/packages/pds/src/account-manager/db/schema/app-password.ts
@@ -3,7 +3,7 @@ export interface AppPassword {
   name: string
   passwordScrypt: string
   createdAt: string
-  privileged: boolean | null
+  privileged: 0 | 1 | null
 }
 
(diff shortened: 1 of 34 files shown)


packages/pds/src/account-manager/db/schema/app-password.ts
```diff
@@ -3,6 +3,7 @@ export interface AppPassword {
   name: string
   passwordScrypt: string
   createdAt: string
+  privileged: 0 | 1 | null
```
devinivy, line 6: What does the `null` do for us?

dholms: I don't think you can make new "non-nullable" columns in sqlite unfortunately so we're just stuck with the `null` as a result of the migration :/

dholms: (may be missing some sqlite feature that allows this, but that was the case the last time i checked)

packages/pds/tests/app-passwords.test.ts
```diff
@@ -90,7 +138,51 @@ describe('app_passwords', () => {
       },
     )
 
-    const attempt = appAgent.api.com.atproto.server.createAppPassword(
+    const priviAttempt = appAgent.api.com.atproto.server.getServiceAuth({
+      aud: 'did:example:test',
+    })
+    await expect(priviAttempt).rejects.toThrow('Bad token scope')
```
devinivy, line 144: I wonder if these lines were intended to land in a different test.

packages/pds/src/account-manager/db/schema/app-password.ts
devinivy: Huh, I really thought you could!  If you can, you would def need to set a default at the same time.

devinivy: e.g.
```sql
ALTER TABLE app_password
  ADD COLUMN privileged NOT NULL DEFAULT 0;
```

devinivy (approved):

packages/pds/src/account-manager/db/schema/app-password.ts
dholms: oh whatdya know, you can as long as you supply a default. good catch 👍

packages/pds/tests/app-passwords.test.ts
dholms: ah nope - that's just an example route that allows only privileged app password auth

i added some comments to clarify 👌

PR #2519, opened 2024-05-24
Sequencer recovery

dholms:
Adds script to pds for performing recovery based off of the sequencer database:
- reapply sequenced repo updates to actor stores
- delete actor stores for accounts that have been deleted (based on account evts)
- recreate accounts that were created during the recovery window. This involves creating a new keypair and update the user's DID


packages/common-web/src/arrays.ts
```diff
@@ -21,3 +21,7 @@ export const mapDefined = <T, S>(
   }
   return output
 }
+
+export const filterDefined = <T>(arr: (T | undefined)[]): T[] => {
+  return arr.filter((item) => item !== undefined) as T[]
+}
```
devinivy, line 27: I think we need to take a little care now to ensure we include changesets for changes to deps, so that the PDS distro will rely on a proper published version of its deps.

packages/pds/src/scripts/common.ts
```diff
@@ -0,0 +1,27 @@
...
+  const currSigningKey = await ctx.idResolver.did.resolveAtprotoKey(did, true)
+  if (updateTo.did() === currSigningKey) {
+    // already up to date
+    return
+  }
+  if (ctx.entrywayAdminAgent) {
+    await ctx.entrywayAdminAgent.api.com.atproto.admin.updateAccountSigningKey({
+      did,
+      signingKey: updateTo.did(),
+    })
```
devinivy, line 19: I think we'll always land in this case since `entrywayAdminAgent` exists when there's an entryway admin token, which fallsback to `env.adminPassword`, which should always be set.

packages/pds/src/scripts/rotate-keys.ts
```diff
@@ -0,0 +1,29 @@
...
+import PQueue from 'p-queue'
+import AppContext from '../context'
+import { updatePlcSigningKey } from './common'
+
+export const rotateKeys = async (ctx: AppContext, args: string[]) => {
+  const filepath = args[0]
+  if (!filepath) {
+    throw new Error('Expected filepath as argument')
+  }
+  const concurrency = args[1] ? parseInt(args[1]) : 25
```
devinivy, line 11:
```suggestion
  const concurrency = args[1] ? parseInt(args[1], 10) : 25
```

packages/pds/src/scripts/rotate-keys.ts
```diff
@@ -0,0 +1,29 @@
...
+    .filter((did) => did.startsWith('did:plc'))
+
+  const queue = new PQueue({ concurrency })
+  for (const did of dids) {
+    queue.add(async () => {
+      try {
+        await updatePlcSigningKey(ctx, did)
+      } catch (err) {
+        console.log(`failed to update key for ${did}: ${err}`)
+      }
```
devinivy, line 26: May be useful to log which ones completed, to help with reruns.

packages/pds/src/scripts/sequencer-recovery/index.ts
```diff
@@ -0,0 +1,14 @@
+import AppContext from '../../context'
+import { Recoverer } from './recoverer'
+
+export const sequencerRecovery = async (ctx: AppContext, args: string[]) => {
+  const cursor = args[0] ? parseInt(args[0]) : 0
+  const concurrency = args[1] ? parseInt(args[1]) : 10
```
devinivy, line 6: I know, super annoying 🙈 
```suggestion
  const cursor = args[0] ? parseInt(args[0], 10) : 0
  const concurrency = args[1] ? parseInt(args[1], 10) : 10
```

Even more annoying, might be nice to do a `!isNaN()` assertion on the inputs.

packages/pds/src/scripts/sequencer-recovery/user-queues.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  async processAll() {
+    await this.main.onIdle()
+  }
+
+  async destroy() {
+    this.main.pause()
+    this.main.clear()
+    this.queues.forEach((q) => q.clear())
+    await this.processAll
```
devinivy, line 39:
```suggestion
    await this.processAll()
```

packages/pds/src/scripts/sequencer-recovery/recoverer.ts
```diff
@@ -0,0 +1,169 @@
...
+    }
+    await this.queues.processAll()
+  }
+
+  async processAll() {
+    await this.queues.processAll()
+  }
+
+  async destroy() {
+    await this.queues.destroy
```
devinivy, line 43:
```suggestion
    await this.queues.destroy()
```

packages/pds/src/scripts/sequencer-recovery/recoverer.ts
```diff
@@ -0,0 +1,169 @@
...
+    }
+    const keypair = await Secp256k1Keypair.create({ exportable: true })
+    await this.ctx.actorStore.create(did, keypair)
+    await this.ctx.actorStore.transact(did, (store) =>
+      store.repo.createRepo([], rev),
+    )
+    if (this.rotateKeys) {
+      await this.updateAccountSigningKey(did, keypair.did())
+    } else {
+      console.log(`skipping key rotation for ${did}`)
```
devinivy, line 102: If we do add some canonical output logs (e.g. list of completed DIDs), we may be happy to have other logs separated on stderr.
```suggestion
      console.warn(`skipping key rotation for ${did}`)
```

devinivy (approved):
Ended up being more straightforward than I expected, nice!  Made it look easy.

One idea that comes to mind is that it's a little bit of a shame the whole accounts table is processed in a single process, and it would be nice if we could break that up.  I think all we would need is for the scripts to support a partitioning scheme, e.g. an option to say "run this for partition #1 of 10".  Then we'd run the script in 10 processes and get a little more juice.

packages/common-web/src/arrays.ts
dholms: Ended up just scrapping this util func and using a type annotation instead 👌 

(most recent TS version actually fixes the need for this 👀)

packages/pds/src/scripts/common.ts
dholms: it's also contingent on having an entryway cfg as well which is the main thing

packages/pds/src/scripts/sequencer-recovery/index.ts
dholms: nah i dig 👍

packages/pds/src/scripts/sequencer-recovery/user-queues.ts
dholms: oh geez that's embarassing 🤦

packages/pds/src/scripts/sequencer-recovery/recoverer.ts
dholms: how did i screw these up 😅

lexicons/com/atproto/admin/updateAccountSigningKey.json
```diff
@@ -0,0 +1,25 @@
+{
+  "lexicon": 1,
+  "id": "com.atproto.admin.updateAccountSigningKey",
```
dholms, line 3: note: this is already implemented/deployed on entryway

packages/pds/src/context.ts
```diff
@@ -235,6 +239,14 @@ export class AppContext {
...
       ? new AtpAgent({ service: cfg.entryway.url })
       : undefined
+    let entrywayAdminAgent: AtpAgent | undefined
+    if (cfg.entryway && secrets.entrywayAdminToken) {
+      entrywayAdminAgent = new AtpAgent({ service: cfg.entryway.url })
+      entrywayAdminAgent.api.setHeader(
+        'authorization',
+        basicAuthHeader('admin', secrets.entrywayAdminToken),
+      )
+    }
```
matthieusieben, line 249: FYI this can now be written using an `Agent` instead of an `AtpAgent` (since we don't need the session management stuff)

```suggestion
    const entrywayAdminAgent =
      cfg.entryway && secrets.entrywayAdminToken
        ? new Agent({
            service: cfg.entryway.url,
            headers: {
              authorization: basicAuthHeader('admin', secrets.entrywayAdminToken),
            },
          })
        : undefined
```

packages/pds/src/scripts/publish-identity.ts
```diff
@@ -0,0 +1,55 @@
...
+  if (!filepath) {
+    throw new Error('Expected filepath as argument')
+  }
+  const timeBetween = args[1] ? parseIntArg(args[1]) : 5
+  const file = await fs.readFile(filepath)
+  const dids = file
+    .toString()
+    .split('\n')
+    .map((did) => did.trim())
+    .filter((did) => did.startsWith('did:plc'))
```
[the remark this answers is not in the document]
dholms: actually looking at this, I reliazed we don't need to be filtering for did:plc since we're just publishing identity events

packages/pds/src/scripts/publish-identity.ts
```diff
@@ -0,0 +1,55 @@
...
+export const publishIdentity = async (
+  ctx: PublishIdentityContext,
+  args: string[],
+) => {
+  const dids = args
+  await publishIdentityEvtForDids(ctx, dids)
+  console.log('DONE')
+}
+
+export const publishIdentityMany = async (
```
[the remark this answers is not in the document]
dholms: yeah i dig 👍

packages/pds/src/context.ts
dholms: doesn't look like the agent constructor takes headers so this ended up looking basically exactly the same except `Agent` instead of `AtpAgent`. I think i'll leave as is for now until we switch all the other agents on the the context over to the `Agent` class

PR #2522, opened 2024-05-24
Set CORS max-age header

devinivy:
This sets the CORS max-age header to the maximum practical value on pds, ozone, and bsky.  Should significantly reduce the number of preflights web users encounter.  Kudos to @matthieusieben for calling this out.


dholms (approved):
yup yup!

PR #2528, opened 2024-05-27
Rebuild repo script

dholms:
An admin script fro rebuilding a repo from the bottom up. For use if a repo gets into a broken state

### DIFF
--- a/packages/pds/bin/rebuild-repo.ts
+++ b/packages/pds/bin/rebuild-repo.ts
@@ -2,6 +2,7 @@ import { CID } from 'multiformats/cid'
 import { ActorStoreTransactor } from '../src/actor-store'
 import AppContext from '../src/context'
 import {
+  BlockMap,
   CidSet,
   MST,
   MemoryBlockstore,
@@ -14,7 +15,7 @@ import { TID } from '@atproto/common'
 export const rebuildRepo = async (ctx: AppContext, did: string) => {
   const memoryStore = new MemoryBlockstore()
   const rev = TID.nextStr()
-  await ctx.actorStore.transact(did, async (store) => {
+  const commit = await ctx.actorStore.transact(did, async (store) => {
     const [records, existingCids] = await Promise.all([
       listAllRecords(store),
       listExistingBlocks(store),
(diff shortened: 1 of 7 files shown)


packages/pds/bin/rebuild-repo.ts
```diff
@@ -0,0 +1,98 @@
...
+  }
+  return cids
+}
+
+const listAllRecords = async (
+  store: ActorStoreTransactor,
+): Promise<RecordDescript[]> => {
+  const records: RecordDescript[] = []
+  let cursor: string | undefined = ''
+  while (cursor) {
```
devinivy, line 73:
```suggestion
  let cursor: string | undefined = ''
  while (cursor !== undefined) {
```

packages/pds/bin/rebuild-repo.ts
```diff
@@ -0,0 +1,98 @@
...
+    await store.repo.storage.updateRoot(commitCid, rev)
+  })
+}
+
+const listExistingBlocks = async (
+  store: ActorStoreTransactor,
+): Promise<CidSet> => {
+  const cids = new CidSet()
+  let cursor: string | undefined = ''
+  while (cursor) {
```
devinivy, line 52:
```suggestion
  let cursor: string | undefined = ''
  while (cursor !== undefined) {
```

packages/pds/src/scripts/rebuild-repo.ts
```diff
@@ -0,0 +1,147 @@
...
+  const rl = readline.createInterface({
+    input: process.stdin,
+    output: process.stdout,
+  })
+
+  return new Promise<boolean>((resolve) => {
+    rl.question('Continue? y/n', (answer) => {
+      resolve(answer === 'y')
+    })
+  })
```
devinivy, line 140: This looks all good, though may be able to use `node:readline/promises` to avoid some of the fun here with promises.

devinivy (approved):

packages/pds/src/scripts/rebuild-repo.ts
dholms: ah nice 👌

PR #2531, opened 2024-05-28
Account deactivation

dholms:
Adds facilities for account deactivation to the PDS, appview & Ozone
- PDS
  - return `status` on session routes
  - take account deactivation into account in auth verifier
  - as part of that^^ revamp standard auth-verifier to take some params for more bespoke auth configurations
- appview
  - strip deactivated accounts from routes. treat similar to takedowns
  - throw on `getProfile` on a deactivated account
- ozone
  - hydrate account deactivation state onto repo views


packages/bsky/src/views/index.ts
```diff
@@ -63,8 +63,21 @@ export class Views {
   // Actor
   // ------------
 
+  actorIsNoHosted(did: string, state: HydrationState): boolean {
+    return (
+      this.actorIsDeactivated(did, state) || this.actorIsTakendown(did, state)
+    )
+  }
```
devinivy, line 70: I saw it defined elsewhere to take into account suspensions.  Do we need to do that here?  From `packages/bsky/src/hydration/actor.ts`:
```ts
      const isNoHosted =
        actor.takenDown ||
        ['takendown', 'suspended', 'deactivated'].includes(
          actor.upstreamStatus ?? '',
        )
```

devinivy: Ohp nevermind, I see it!

lexicons/com/atproto/server/createSession.json
```diff
@@ -33,7 +33,12 @@
             "didDoc": { "type": "unknown" },
             "email": { "type": "string" },
             "emailConfirmed": { "type": "boolean" },
-            "emailAuthFactor": { "type": "boolean" }
+            "emailAuthFactor": { "type": "boolean" },
+            "status": {
+              "type": "string",
+              "description": "Hosting status of the account. If not specified, then assume 'active'.",
+              "knownValues": ["active", "takendown", "suspended", "deactivated"]
```
devinivy, line 40: How do we feel about the fact that `active` may appear here but is not a hosting status under `com.atproto.sync`?

I think I like the convention that a status indicates a reason for a no-host, so if you see a value you don't understand you can at least assume it still indicates no-host.  One result of that is that the logic around this field could be `const noHost = !!status && status !== 'active'`, which is pretty nice.  Would it be even nicer if it were just `const noHost = !!status`?  I notice in a fair amount of logic in this PR we check explicitly for the known values to determine a no-host.

packages/bsky/src/hydration/actor.ts
```diff
@@ -116,6 +122,7 @@ export class ActorHydrator {
         takedownRef: safeTakedownRef(actor),
         isLabeler: actor.labeler ?? false,
         allowIncomingChatsFrom: actor.allowIncomingChatsFrom || undefined,
+        upstreamStatus: actor.upstreamStatus ?? undefined,
```
devinivy, line 125: I think thanks to grpc that `actor.upstreamStatus` could be `''` when missing, does that seem right?  If that's the case we might want:
```suggestion
        upstreamStatus: actor.upstreamStatus || undefined,
```

packages/bsky/src/hydration/actor.ts
```diff
@@ -95,9 +96,14 @@ export class ActorHydrator {
     const res = await this.dataplane.getActors({ dids })
     return dids.reduce((acc, did, i) => {
       const actor = res.actors[i]
+      const isNoHosted =
+        actor.takenDown ||
+        ['takendown', 'suspended', 'deactivated'].includes(
+          actor.upstreamStatus ?? '',
```
devinivy, line 102: May be able to future-proof this using something like `actor.upstreamStatus && actor.upstreamStatus !== 'active'`.  Or if this is an area where it should never be set to `active`, perhaps just `!!actor.upstreamStatus`.

packages/pds/src/api/com/atproto/repo/uploadBlob.ts
```diff
@@ -6,7 +6,9 @@ import { BlobMetadata } from '../../../../actor-store/blob/transactor'
 
 export default function (server: Server, ctx: AppContext) {
   server.com.atproto.repo.uploadBlob({
-    auth: ctx.authVerifier.accessCheckTakedown,
+    auth: ctx.authVerifier.accessStandard({
+      checkTakedown: true,
+    }),
```
devinivy, line 11: No biggie either way, but I imagine we could do a deactivated step too since this is part of a write flow.

dholms: Yeah I originally did that, but uploadBlob is used in the account migration flow (during which the account is deactivated). Since `uploadBlob` doesn't actually produce firehose events, I figured it was okay to let it through

May be worth revisiting if we want an account that is in the process of importing their repo to actually be "deactivated"

devinivy (approved):
Ace!

packages/pds/src/api/com/atproto/server/requestAccountDelete.ts
```diff
@@ -18,7 +18,7 @@ export default function (server: Server, ctx: AppContext) {
         calcKey: ({ auth }) => auth.credentials.did,
       },
     ],
-    auth: ctx.authVerifier.accessCheckTakedown,
+    auth: ctx.authVerifier.accessFull({ checkTakedown: true }),
```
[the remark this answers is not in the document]
dholms: yeah i'm not really sure this matters because ultimately you need a password to delete an account. But this request should really only be coming from a fully-permissioned session. I think we didn't the full check earlier because we didn't have an authVerifier helper to make it possible

lexicons/com/atproto/server/createSession.json
dholms: ended up updating this to match the sematnics of `getRepoStatus` with the `active` boolean

PR #2541, opened 2024-06-02
Remove unused lexicon import

tom-sherman:
This confused me a bit as I was reading the code 😅


tom-sherman:
This is an export, more caffeine needed 🤦

PR #2548, opened 2024-06-04
Social proof

dirkmc:
When you view a user’s profile we’d like to provide some details about followers of the given account that you follow.  This establishes social connections between the two users, and we hope that it will facilitate users discovering their broader clusters and ultimately following each other.  Here’s an example:
image

This PR adds a `socialProof` to the `viewer` section of the profile, eg:
```
{
    ...
    socialProof: {
      count: 1,
      follows: [
        {
          did: 'did:plc:x3bne6hoa7xokmoccsqzgxvm',
          handle: 'bob.test',
          displayName: 'bobby',
          avatar: 'https://bsky.public.url/img/avatar/plain/did:plc:x3bne6hoa7xokmoccsqzgxvm/bafkreiaivizp4xldojmmpuzmiu75cmea7nq56dnntnuhzhsjcb63aou5ei@jpeg',
          viewer: [Object],
          labels: []
        }
      ]
    }
}
```

The PR implements
- a `getFollowsFollowing({ actorDid, targetDid })` method at the data-plane layer, that returns a list of dids
- hydration:
  - the `socialProof` element of the `ProfileViewerState` with a `count` and `follows` for the first 5 dids returned from the data-plane layer
  - each did is hydrated into a profile on the `HydrationState`
- a view layer  that maps each of the follows into a basic profile

### DIFF
--- a/.changeset/thirty-dolls-try.md
+++ b/.changeset/thirty-dolls-try.md
@@ -1,7 +1,0 @@
----
-"@atproto/api": patch
-"@atproto/pds": patch
-"@atproto/dev-env": patch
----
-
-Allow updating deactivation state through admin.updateSubjectStatus
--- a/.changeset/thirty-dolls-try.md
+++ b/.changeset/thirty-dolls-try.md
@@ -1,0 +1,7 @@
+---
+"@atproto/api": patch
+"@atproto/pds": patch
+"@atproto/dev-env": patch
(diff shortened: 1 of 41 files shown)


estrattonbailey:
Thanks @dirkmc! Gonna merge into the base and take it from here 👍

PR #2553, opened 2024-06-06
Starter packs

devinivy:


lexicons/app/bsky/graph/starterpack.json
```diff
@@ -0,0 +1,50 @@
...
+            "maxLength": 3000
+          },
+          "descriptionFacets": {
+            "type": "array",
+            "items": { "type": "ref", "ref": "app.bsky.richtext.facet" }
+          },
+          "list": {
+            "type": "string",
+            "format": "at-uri",
+            "description": "Reference (AT-URI) to the list record."
```
dholms, line 31: this is a list of recommended users?

lexicons/app/bsky/actor/defs.json
```diff
@@ -47,6 +48,7 @@
           "ref": "#profileAssociated"
         },
         "indexedAt": { "type": "string", "format": "datetime" },
+        "createdAt": { "type": "string", "format": "datetime" },
```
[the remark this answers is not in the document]
devinivy: This relates to some things we want to do specifically for newskies.  It relates to starter packs because we may want to feature the starter pack someone used to join on their profile, but only for a period of time.

We'll probably give `createdAt` the same amount of credence we do for posts or the many other records that have it, and lean more on the server to determine when the user started using Bluesky.  But for completeness I figured it should probably be reflected on repo in some way (e.g. for the next appview's benefit!).  Most records have the `createdAt` field, so it felt natural enough to me to include it.

lexicons/app/bsky/graph/starterpack.json
devinivy: Yep, that's right!

lexicons/app/bsky/graph/starterpack.json
```diff
@@ -0,0 +1,50 @@
...
+      "type": "record",
+      "description": "Record defining a starter pack of actors and feeds for new users.",
+      "key": "any",
+      "record": {
+        "type": "object",
+        "required": ["name", "createdAt"],
+        "properties": {
+          "name": {
+            "type": "string",
+            "maxLength": 64,
```
haileyok, line 15: Cool if we drop this down to 50?

devinivy: The main consideration is that it would be an exception from other display names, e.g.
 - lists
 - profiles
 - labels

If that seems fine to yall then sounds good!

lexicons/app/bsky/graph/starterpack.json
```diff
@@ -0,0 +1,50 @@
...
+            "items": { "type": "ref", "ref": "app.bsky.richtext.facet" }
+          },
+          "list": {
+            "type": "string",
+            "format": "at-uri",
+            "description": "Reference (AT-URI) to the list record."
+          },
+          "feeds": {
+            "type": "array",
+            "maxLength": 50,
```
haileyok, line 35: Note to self that we want to figure out the right max for this. I'm leaning toward 3 after some more though (so 5 feeds pinned when you sign up including Discover and Following) but will try to get a consensus. Definitely going to be smaller than 50 though 😅

lexicons/app/bsky/graph/starterpack.json
haileyok: I know we can't put constraints on list size but are we able to at least require this field to be set?

PR #2557, opened 2024-06-06
Social proof

estrattonbailey:
Adds `knownFollowers` to `ViewerState` on `ProfileDetailed` views. This includes:
- `getProfiles`
- `getProfile`
- `getSuggestedFollowsByActor`
- `getSuggestions`

Also adds `getKnownFollowers` method that returns the full list of known followers for a given `actor`. This is set up for pagination, but is not actually paginated atm due to dataplane constraints.


lexicons/app/bsky/actor/defs.json
```diff
@@ -122,7 +122,28 @@
...
+    },
+    "knownFollowers": {
+      "type": "object",
+      "description": "The subject's followers whom you also follow",
+      "required": ["count", "followers"],
+      "properties": {
+        "count": { "type": "integer" },
+        "followers": {
+          "type": "array",
+          "minLength": 1,
```
dholms, line 140: I kinda want this to be `0`. To me an undefined response means something different from an empty array.
"I didn't try to load any knownFollowers on here" vs "I tried, but didn't find any"

thoughts?

packages/bsky/src/api/app/bsky/graph/getKnownFollowers.ts
```diff
@@ -0,0 +1,122 @@
...
+    handler: async ({ params, auth, req }) => {
+      const viewer = auth.credentials.iss
+      const labelers = ctx.reqLabelers(req)
+      const hydrateCtx = await ctx.hydrator.createContext({
+        labelers,
+        viewer,
+      })
+
+      const result = await getKnownFollowers(
+        { ...params, hydrateCtx: hydrateCtx.copy({ viewer }) },
```
dholms, line 35: what's the point of the `copy` here?

packages/bsky/src/api/app/bsky/graph/getKnownFollowers.ts
```diff
@@ -0,0 +1,122 @@
...
+  })
+  return skeleton
+}
+
+const presentation = (
+  input: PresentationFnInput<Context, Params, SkeletonState>,
+) => {
+  const { ctx, hydration, skeleton } = input
+  const { knownFollowers } = skeleton
+  const isNoHosted = (did: string) => ctx.views.actorIsNoHosted(did, hydration)
```
dholms, line 96: since we never include takedowns in hydration (`includeTakedowns` on `HydrateCtx`), it's actually not required to do this check, they get stripped out ahead of time 👍

packages/bsky/src/hydration/actor.ts
```diff
@@ -146,18 +150,36 @@ export class ActorHydrator {
...
+    hydrateKnownFollowers = false,
   ): Promise<ProfileViewerStates> {
     if (!dids.length) return new HydrationMap<ProfileViewerState>()
-    const res = await this.dataplane.getRelationships({
-      actorDid: viewer,
-      targetDids: dids,
-    })
+
+    const subjectDidToKnownFollowersMap = {}
+    const knownFollowersRequests = hydrateKnownFollowers
```
dholms, line 158: Could we do this without mutating an outside map? And instead resolve an array of values, then use the index in the reduce function afterwards to grab the relevant known follows?

packages/bsky/src/hydration/actor.ts
```diff
@@ -167,6 +189,13 @@ export class ActorHydrator {
         blockingByList: parseString(rels.blockingByList),
         following: parseString(rels.following),
         followedBy: parseString(rels.followedBy),
+        knownFollowers:
+          knownFollowersDids?.length > 0
```
dholms, line 193: maybe to just make this a little more clear, how about `knownFollowersDids && knownFollowersDids?.length > 0`?

that way we're not comparing `undefined > 0`

packages/bsky/src/hydration/actor.ts
dholms: hmm yknow could also be nice to just move all this logic out into a separate helper function on this class 🤔 

idk not required, but you could make a nice little hydration map that cleans up the reducer logic later on as well

packages/bsky/src/hydration/hydrator.ts
```diff
@@ -127,13 +127,27 @@ export class Hydrator {
   async hydrateProfileViewers(
     dids: string[],
     ctx: HydrateCtx,
+    hydrateKnownFollowers = false,
```
dholms, line 130: mer feel like this param is a little out of place in this class... we don't have any other methods with parameter switches like this. but also not really seeing how to get around it

dholms: well actually, combining this with the feedback from actor hydrator, one option would be just to treat `profileKnownFollowers` as a separate hydration map entirely on `HydrationState` (alongside `profileViewers` instead of inside it). Then we wouldn't need to parameterize this method, it could clean up the actor hydrator a bit, and we would just call that function from within `hydrateProfilesDetailed` 👀

dholms: does that make any sense? sorry it's a bit bigger of a change, but should just be moving logic around

packages/bsky/src/views/index.ts
```diff
@@ -132,10 +134,14 @@ export class Views {
     }
   }
 
-  profile(did: string, state: HydrationState): ProfileView | undefined {
+  profile(
+    did: string,
+    state: HydrationState,
+    includeKnownFollowers = false,
```
dholms, line 140: yeah similarly to the hydrator, we could remove the parameter on this, and only call `knownFollowers(did, state) from within the `profileDetailed()` view, then just munge that into profile viewer state

packages/bsky/src/views/index.ts
```diff
@@ -213,9 +221,23 @@ export class Views {
...
       followedBy: viewer.followedBy && !block ? viewer.followedBy : undefined,
+      knownFollowers:
+        includeKnownFollowers && viewer.knownFollowers
+          ? this.knownFollowersDidsToProfileViews(viewer.knownFollowers, state)
+          : undefined,
     }
   }
 
+  knownFollowersDidsToProfileViews(
+    knownFollowers: Required<HydratorProfileViewerState>['knownFollowers'],
```
dholms, line 232: would clean this param up too 😛

dholms: and maybe shorten this method to just `knownFollowers`

dholms:
Overall looking really solid!

Implementation seems sound & tests looks solid.

After thinking it through, I think we should reconsider `knownFollowers` as a new top level thing on the Hydrator and reflect that in the hydrator, actor hydrator & views class instead of parameterizing the current methods. Lmk what you think/if you have any questions

packages/bsky/src/api/app/bsky/graph/getKnownFollowers.ts
estrattonbailey: oh lawl just copied this bit from getBlocks, will remove

estrattonbailey: oh actually forgot, it make TS happy — I want `viewer` to be defined here — that OK or is there a better way?

packages/bsky/src/api/app/bsky/graph/getKnownFollowers.ts
estrattonbailey: oh excellent 👍

packages/bsky/src/api/app/bsky/graph/getKnownFollowers.ts
dholms: ah yeah that's fine 👍

packages/bsky/proto/bsky.proto
```diff
@@ -969,13 +969,21 @@ message GetRecordTakedownResponse {
 
 // Polo-backed Graph Endpoints
 
-// GetFollowsFollowing gets the list of DIDs that the actor follows that also follow the target
+
```
dholms, line 972: I think we ended up accumulating some more protos here than we needed

dholms: also I think we need a `pnpm format` to clean up some of the buf codegen as well

packages/bsky/src/hydration/actor.ts
```diff
@@ -148,10 +154,14 @@ export class ActorHydrator {
     viewer: string,
   ): Promise<ProfileViewerStates> {
     if (!dids.length) return new HydrationMap<ProfileViewerState>()
-    const res = await this.dataplane.getRelationships({
-      actorDid: viewer,
-      targetDids: dids,
-    })
+
+    const [res] = await Promise.all([
```
dholms, line 158: can ditch the `Promise.all` after the refactor 👌

packages/bsky/src/hydration/graph.ts
```diff
@@ -195,4 +195,24 @@ export class GraphHydrator {
     })
     return { followers: res.followers, cursor: res.cursor }
   }
+
+  async getKnownFollowers(input: {
```
dholms, line 199: This is a bit confusing because there's a method with the same name on the actor hydrator & it's not really clear the difference

Since this is only used for building the skeleton for the `getKnownFollowers` route & not actually for hydration, I might move this logic over to the route handler itself

dholms (approved):
lookin good! 💪 

left a couple notes - big one is to tidy up the protos file

packages/bsky/src/api/app/bsky/graph/getKnownFollowers.ts
```diff
@@ -0,0 +1,119 @@
...
+  }
+  if (clearlyBadCursor(params.cursor)) {
+    return { subjectDid, knownFollowers: [], cursor: undefined }
+  }
+
+  const res = await ctx.hydrator.dataplane.getFollowsFollowing({
+    actorDid: params.hydrateCtx.viewer,
+    targetDids: [subjectDid],
+  })
+  const result = res.results[0]
```
devinivy, line 62: This helps the type system acknowledge the case that there's no result, which I see you're handling below:
```suggestion
  const result = res.results.at(0)
```

packages/bsky/src/api/app/bsky/graph/getKnownFollowers.ts
```diff
@@ -0,0 +1,119 @@
...
+  const knownFollowers = result ? result.dids.slice(0, params.limit || 50) : []
```
devinivy, line 63: I am pretty sure `params.limit` is guaranteed to be there and `>0` based on lexicon validation.

packages/bsky/src/data-plane/server/routes/follows.ts
```diff
@@ -92,7 +93,47 @@ export default (db: Database): Partial<ServiceImpl<typeof Service>> => ({
...
+            .selectFrom('follow')
+            .where('follow.subjectDid', '=', subjectDid)
+            .select(['creator']),
+        )
+        .select(['subjectDid'])
+      const rows = await followsReq.execute()
+      results.push({
+        targetDid: subjectDid,
+        dids: rows.map((r) => r.subjectDid),
+      } as FollowsFollowing)
```
devinivy, line 134: If this doesn't require the type cast it would be awesome if we could remove it.

packages/bsky/src/hydration/hydrator.ts
```diff
@@ -93,6 +94,7 @@ export type HydrationState = {
   labelers?: Labelers
   labelerViewers?: LabelerViewerStates
   labelerAggs?: LabelerAggs
+  knownFollowers?: KnownFollowers
```
devinivy, line 97: There should be a line down in `mergeStates()` to merge this new entry of hydration state.

packages/bsky/src/views/index.ts
```diff
@@ -216,6 +227,16 @@ export class Views {
...
   }
 
+  knownFollowers(
+    knownFollowers: Required<HydratorProfileViewerState>['knownFollowers'],
+    state: HydrationState,
+  ) {
+    const followers = mapDefined(knownFollowers.followers, (did) => {
+      return this.profileBasic(did, state)
+    })
+    return { count: knownFollowers.count, followers }
```
devinivy, line 237: I believe that there is a minimum of one item for `followers`.  It would be nice to have some logic enforcing that colocated around here, or perhaps drop the requirement from the lexicon if you think the frontend wont mind.

We may also want to consider enforcing blocks, and possibly mutes, in this list.

packages/bsky/src/api/app/bsky/graph/getKnownFollowers.ts
estrattonbailey: Oh good call, v nice

packages/bsky/src/api/app/bsky/graph/getKnownFollowers.ts
estrattonbailey: Oh ok, so if there's a `default` then we know we'll have a value here, got it 👍

packages/bsky/src/data-plane/server/routes/follows.ts
estrattonbailey: Yeah true, I can create a new instance for each 👍

packages/bsky/src/hydration/hydrator.ts
estrattonbailey: Ah geez, thanks 🤦

packages/bsky/src/views/index.ts
estrattonbailey: [We updated](https://github.com/bluesky-social/atproto/pull/2557#discussion_r1631762637) this the other day to `0`. Yeah I'll make sure frontend can handle it, no issue there.

Yeah should def filter out blocks, I think mutes are OK to leave in though.

estrattonbailey: Ya know, hadn't codegenned though, glad you brought this up.

Blocks are now filtered, added a test

packages/bsky/src/api/app/bsky/graph/getKnownFollowers.ts
devinivy: Yes!  But the types should also reflect that: `number` (has a default / is always set) vs. `number | undefined` (does not have a default / is not always set).

packages/bsky/tests/views/blocks.test.ts
```diff
@@ -595,4 +597,15 @@ describe('pds views with blocking', () => {
...
+
+  it('returns knownFollowers with blocks filtered', async () => {
+    const carolForAlice = await agent.api.app.bsky.actor.getProfile(
+      { actor: bob },
+      { headers: await network.serviceHeaders(alice) },
+    )
+
+    const knownFollowers = carolForAlice.data.viewer?.knownFollowers
+    expect(knownFollowers?.count).toBe(1)
+    expect(knownFollowers?.followers).toHaveLength(0)
```
estrattonbailey, line 609: Just noting here `count` is 1, but the blocked user has been filtered out. We return 5 to the client, and the client uses only 3, so we're already overfetching. Discussed with Devin and decided this was a rare enough case to leave in for now. Frontend will handle the difference.

PR #2558, opened 2024-06-06
Thread mutes

dholms:
Store and action thread mutes in the appview in order to:
- properly prevent push notifications on muted threads
- relieve the client of the burden of filtering notifications based on locally stored thread mutes

Thread mutes are synced through bsync similarly to actor mutes. However, unlike actor mute, thread mutes are actioned at indexing time.


devinivy (approved):
Ace!

PR #2560, opened 2024-06-07
:construction: Proposal for ozone config lexicons

foysalit:
This proposal introduces a new entity `config` which is essentially a list of key value pairs where each key may have one or many values.

A few open questions
1. Updates are tricky. The current proposal does not include updating a key or update values by wildcard.
2. Create and update allows array input but delete only allows key and wildcard value input, do we want array input for deletion too?

One of the primary reasons why I'm in favor of calling this `config` and generalizing this is to facilitate storing shareable config for entire instance, that are currently browser based and as a result, user specific or stored in env var. For instance, our queue config is only stored in the env var of the client build and we could move that to the db under configs where each row with key `queue` would be the config for an individual queue. We also have external labeler config currently [in review](https://github.com/bluesky-social/ozone/pull/93) but ideally, we would want to store these on the server so that all mods using the instance see labels from same external labelers.

### DIFF
--- a/lexicons/tools/ozone/config/createConfig.json
+++ b/lexicons/tools/ozone/config/createConfig.json
@@ -1,0 +1,58 @@
+{
+  "lexicon": 1,
+  "id": "com.atproto.config.createConfig",
+  "defs": {
+    "main": {
+      "type": "procedure",
+      "description": "Create new configuration key-value pairs.",
+      "input": {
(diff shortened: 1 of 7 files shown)


foysalit:
We're moving forward with sets api instead of config https://github.com/bluesky-social/atproto/pull/2636

Vincentroland (approved):

PR #2561, opened 2024-06-07
Fix flaky tests surrounding ozone/pds

devinivy:
Found a couple small issues that are probably contributing to test flakiness:
 - `network.processAll()` should ensure to process ozone before appview in order for hosting status changes to propagate.
 - minor issue with two tests using same backing pg schema.

### DIFF
--- a/packages/ozone/tests/lang.test.ts
+++ b/packages/ozone/tests/lang.test.ts
@@ -15,7 +15,7 @@ describe('moderation status language tagging', () => {
 
   beforeAll(async () => {
     network = await TestNetwork.create({
-      dbPostgresSchema: 'ozone_blob_divert_test',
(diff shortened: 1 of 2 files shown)


dholms (approved):
good looks 👌

packages/ozone/tests/lang.test.ts
```diff
@@ -15,7 +15,7 @@ describe('moderation status language tagging', () => {
 
   beforeAll(async () => {
     network = await TestNetwork.create({
-      dbPostgresSchema: 'ozone_blob_divert_test',
```
foysalit, line 18: dang I think I had this fixed in one of my inprogress prs, prolly the mod-manager one? anyways, thanks for the fix!

PR #2563, opened 2024-06-07
replace graphemer by unicode-segmenter

cometkim:
I made a Unicode library that is much smaller and faster than graphemer. Check it out: https://github.com/cometkim/unicode-segmenter?tab=readme-ov-file#unicode-segmentergrapheme-vs

It ensures compliance with the latest Unicode data by performing tests and fuzzing against the `Intl.Segmenter` API.


cometkim:
ok, I fixed the lockfile diff

PR #2570, opened 2024-06-10
fix(api): `hasMutedWord` for facets with multiple features

sugyan:
The `hasMutedWord` function used in moderation uses the value of the tag in the facets passed as an argument as `tags` for judging. The current implementation was to “add the 0th of the features to the `tags` if a feature corresponding to `isTag` is found”.

This is incorrect, `features` is an array that can contain multiple elements, and those of type `#tag` can be included after the first. The 0th element can be hidden by a tag that is not the target word of the mute, or it can be a completely unrelated type.

Actually, it is possible to create such a post (e.g. https://bsky.app/profile/sugyan.pds.shigepon.net/post/3kukmykhshk2l), and an error will occur if a user with `mutedWords` set tries to view it.

Fixed it so all `#tag` types in `features` are subject to the decision, and added a test.

### DIFF
--- a/.changeset/six-teachers-tie.md
+++ b/.changeset/six-teachers-tie.md
@@ -1,0 +1,5 @@
+---
+"@atproto/api": patch
+---
+
+Fix `hasMutedWord` for facets with multiple features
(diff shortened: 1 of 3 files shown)


pfrazee (approved):
Nice yep LGTM

Issue #2555, opened 2024-06-06
Birthdate is erased when clients authenticated with app passwords update preferences


mikeyclarke:
**Describe the bug**

The `app.bsky.actor.putPreferences` endpoint appears to delete the `app.bsky.actor.defs.personalDetailsPref` preference when the client is authenticated with an app password.

**To Reproduce**

1. Using the Bluesky app/website set a birthdate in your moderation settings
1. Use a client authenticated with an app password to change a preference

**Expected behavior**

Your birthdate remains.

**Actual behavior**

Your birthdate is removed and must be re-entered again.

**Details**

I noticed this whilst adding content safety settings to my own app but I’ve also tried changing moderation settings in the Graysky app and it has the same effect.

**Additional context**

It looks like #1624 deliberately excluded sending through the birthdate preference in the `app.bsky.actor.getPreferences` response for privacy reasons but the put endpoint erases preferences that aren’t supplied to it, hence the bug.

pfrazee:
Ah lovely. @dholms I figure we either need to make the backend preserve the `app.bsky.actor.defs.personalDetailsPref` item if it's not included in the put, or we need to drop the app password filter

PR #2573, opened 2024-06-11
Scope put prefs by AuthScope

dholms:
We don't let app passwords read `personalDetailsPref`. This prevents app passwords from writing or deleting the preferences as well

Closes https://github.com/bluesky-social/atproto/issues/2555

### DIFF
--- a/packages/pds/src/api/app/bsky/actor/putPreferences.ts
+++ b/packages/pds/src/api/app/bsky/actor/putPreferences.ts
@@ -19,7 +19,11 @@ export default function (server: Server, ctx: AppContext) {
         }
       }
       await ctx.actorStore.transact(requester, async (actorTxn) => {
-        await actorTxn.pref.putPreferences(checkedPreferences, 'app.bsky')
+        await actorTxn.pref.putPreferences(
+          checkedPreferences,
+          'app.bsky',
+          auth.credentials.scope,
+        )
       })
     },
   })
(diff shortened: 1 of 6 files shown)


pfrazee:
Is this right? Isn't the problem that the apps are dropping `personalDetailsPref` because it's written as a single blob, and they don't have access to it? I think the fix needs to be that the server preserves the `personalDetailsPref` in this case, not that it throws

dholms:
Yeah there's two parts to the putPreferences - it throws if an app password is trying to write a new value to `personalDetailsPref` but just strips it out of the values to delete if an app password doesn't provide a value

pfrazee:
Right -- my bad, I had the wrong mental model for the backend, so the diff seemed wrong.

This LGTM

packages/pds/src/api/app/bsky/actor/putPreferences.ts
```diff
@@ -19,7 +19,11 @@ export default function (server: Server, ctx: AppContext) {
         }
       }
       await ctx.actorStore.transact(requester, async (actorTxn) => {
-        await actorTxn.pref.putPreferences(checkedPreferences, 'app.bsky')
+        await actorTxn.pref.putPreferences(
+          checkedPreferences,
+          'app.bsky',
+          auth.credentials.scope,
+        )
```
devinivy, line 26: Part of me likes "terminating" scopes and access control in the handlers.  That would involve not passing a scope into `putPreferences()`, instead passing a function or perhaps a deny-list based on the scope.

That said, it's super minor and definitely arguable either way, so mostly just making a note for future thought if we run into more cases like this.

dholms: I actually started out like that, but then switched it to passing the scope through so we could share the filtering logic & didn't have to pass around magic lists of hard coded strings in the handlers 🤔 

I get what you're pointing at, but think I'll just keep it like this for now - we'll get to clean all this up when we get around to the private state api

PR #2593, opened 2024-06-21
Fix getActorLikes documentation to reflect auth required

LilahTovMoon:
You can only view your own likes and this fixes the documentation to note that auth is required and that the actor must be the requesting account

https://github.com/bluesky-social/atproto/pull/1440#issuecomment-1700193127

It looks like `getActorLikes` got marked as "Does not require auth" when it does require auth.


devinivy (approved):
Thanks!

PR #2594, opened 2024-06-21
:sparkles: Proxy starterpack requests through ozone and include takedowns

foysalit:


packages/dev-env/src/seed/basic.ts
```diff
@@ -155,6 +155,8 @@ export default async (
     })
   }
 
+  await sc.network.processAll()
+
```
devinivy, line 159: All the other seeds let the caller perform the final `processAll()`— does it cause any issues if we stick with that for now?

devinivy: I see in the new test suite it works the same way:
```ts
    await basicSeed(sc)
    await network.processAll()
```

devinivy (approved):
Looks great, thanks!

packages/dev-env/src/seed/basic.ts
foysalit: nope, don't think there's any issue with that. just cleaned up!

PR #2596, opened 2024-06-22
Use detailed profile view for `getSuggestions`

mozzius:
Allows us to show social proof in the suggested follows section. Includes a minor lexicon change from `ProfileView` to `ProfileViewDetailed`, but I don't think the AppView is doing much more work here - seems like it's fetching the detailed profile views already but just not sending them

### DIFF
--- a/lexicons/app/bsky/actor/getSuggestions.json
+++ b/lexicons/app/bsky/actor/getSuggestions.json
@@ -28,7 +28,7 @@
               "type": "array",
               "items": {
                 "type": "ref",
-                "ref": "app.bsky.actor.defs#profileView"
+                "ref": "app.bsky.actor.defs#profileViewDetailed"
               }
             }
           }
--- a/lexicons/app/bsky/actor/getSuggestions.json
+++ b/lexicons/app/bsky/actor/getSuggestions.json
@@ -28,7 +28,7 @@
               "type": "array",
               "items": {
                 "type": "ref",
-                "ref": "app.bsky.actor.defs#profileViewDetailed"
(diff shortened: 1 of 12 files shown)


estrattonbailey (approved):
Nice one! Get Devin's final sign off, but lgtm

lexicons/app/bsky/actor/getSuggestions.json
```diff
@@ -28,7 +28,7 @@
               "type": "array",
               "items": {
                 "type": "ref",
-                "ref": "app.bsky.actor.defs#profileView"
+                "ref": "app.bsky.actor.defs#profileViewDetailed"
```
devinivy, line 31: One intentional feature of the way `knownFollowers` was added to `ViewerState` is that it's decoupled from whether we serve a `ProfileView` vs. `ProfileViewDetailed`.  This is nice because it means we can sprinkle `knownFollowers` wherever we want it without converting all those areas to detailed views.

How would we feel about instead leaving the lexicons unchanged and just adding `knownFollowers` into the `ViewerState`s served from this endpoint?

devinivy (approved):
Good stuff!!  Adding a new view like this is a simple improvement, hard to go wrong 👍 

I think we'll still want to align what we're hydrating with what we're presenting, which this change doesn't quite get us.  But I can hit that in some followup work.  In the spirit of keeping all the view/hydration wires exposed in the route handlers, the way I'd do this would probably be to add an option to the profile hydration and profile view methods that determines whether `knownFollowers` should be included.

PR #2599, opened 2024-06-24
Add exception for HTTP clients that set a content-type when there is no body

matthieusieben:


packages/xrpc-server/tests/bodies.test.ts
```diff
@@ -229,6 +229,17 @@ describe('Bodies', () => {
     expect(compressed.cid).toEqual(expectedCid.toString())
   })
 
+  it('supports empty payload', async () => {
+    const expectedCid = await cidForCbor(new Uint8Array(0))
+
+    // Using "undefined" as body to avoid encoding as JsonLD { $bytes: "<base64>" }
```
devinivy, line 235:
```suggestion
    // Using "undefined" as body to avoid encoding as lexicon { $bytes: "<base64>" }
```

devinivy (approved):
Awesome, thanks!

PR #2603, opened 2024-06-25
Social proof blocks

estrattonbailey:
This PR ensures that if I follow Daniel, and Daniel follows Devin, but Devin blocks Daniel, when I view Devin's profile I do not see Daniel as part of `knownFollowers`.

I also extracted some utils and added a few tests + added a new seed specifically for `knownFollowers` testing to handle these 2nd-party blocks more clearly/easily.

### DIFF
--- a/packages/bsky/src/hydration/hydrator.ts
+++ b/packages/bsky/src/hydration/hydrator.ts
@@ -36,11 +36,13 @@ import {
 } from './label'
 import {
   HydrationMap,
-  Merges,
   RecordInfo,
   ItemRef,
   didFromUri,
   urisByCollection,
+  mergeMaps,
+  mergeNestedMaps,
+  mergeManyMaps,
 } from './util'
 import {
   FeedGenAggs,
@@ -1064,25 +1066,15 @@ export const mergeStates = (
     labelerAggs: mergeMaps(stateA.labelerAggs, stateB.labelerAggs),
     labelerViewers: mergeMaps(stateA.labelerViewers, stateB.labelerViewers),
     knownFollowers: mergeMaps(stateA.knownFollowers, stateB.knownFollowers),
-    bidirectionalBlocks: mergeMaps(
+    bidirectionalBlocks: mergeNestedMaps(
       stateA.bidirectionalBlocks,
       stateB.bidirectionalBlocks,
     ),
   }
 }
 
-const mergeMaps = <M extends Merges>(mapA?: M, mapB?: M): M | undefined => {
-  if (!mapA) return mapB
-  if (!mapB) return mapA
-  return mapA.merge(mapB)
-}
-
 const mergeManyStates = (...states: HydrationState[]) => {
   return states.reduce(mergeStates, {} as HydrationState)
-}
-
-const mergeManyMaps = <T>(...maps: HydrationMap<T>[]) => {
-  return maps.reduce(mergeMaps, undefined as HydrationMap<T> | undefined)
 }
 
 const actionTakedownLabels = <T>(
--- a/packages/bsky/src/hydration/hydrator.ts
+++ b/packages/bsky/src/hydration/hydrator.ts
@@ -102,6 +102,7 @@ export type HydrationState = {
   labelerViewers?: LabelerViewerStates
   labelerAggs?: LabelerAggs
   knownFollowers?: KnownFollowers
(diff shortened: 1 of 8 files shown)


packages/bsky/src/hydration/hydrator.ts
```diff
@@ -737,6 +752,33 @@ export class Hydrator {
...
+  async hydrationBidirectionalBlocks(
+    didMap: Map<string, string[]>,
+  ): Promise<BidirectionalBlocks> {
+    const pairs: RelationshipPair[] = []
+
+    for (const [uri, followers] of didMap) {
+      for (const follower of followers) {
+        pairs.push([didFromUri(uri), follower])
+      }
+    }
```
devinivy, line 764: It looks like `didMap` is `DID -> DID[]`, in which case we don't have a URI to `didFromUri()` with.

packages/bsky/src/hydration/hydrator.ts
```diff
@@ -737,6 +752,33 @@ export class Hydrator {
     return { follows, followBlocks }
   }
 
+  async hydrationBidirectionalBlocks(
+    didMap: Map<string, string[]>,
+  ): Promise<BidirectionalBlocks> {
+    const pairs: RelationshipPair[] = []
+
+    for (const [uri, followers] of didMap) {
```
devinivy, line 760: How would you feel about loosening the `followers` verbiage in here?  I wonder if it may be a little confusing if you're reading this method in isolation, as it's quietly sort of alluding to the known followers feature.

packages/bsky/src/hydration/hydrator.ts
```diff
@@ -737,6 +752,33 @@ export class Hydrator {
...
+      }
+    }
+
+    const blocks = await this.graph.getBidirectionalBlocks(pairs)
+    const result = new HydrationMap<HydrationMap<boolean>>()
+
+    for (const [did] of didMap) {
+      const followBlocks = new HydrationMap<boolean>()
+
+      for (const followerDid of didMap.get(did)!) {
```
devinivy, line 772: I think there's a nice way to not need the non-null assertion and lean on the types:
```suggestion
    for (const [did, targetDids] of didMap) {
      const followBlocks = new HydrationMap<boolean>()

      for (const followerDid of targetDids) {
```

packages/bsky/src/hydration/hydrator.ts
```diff
@@ -1022,6 +1064,10 @@ export const mergeStates = (
     labelerAggs: mergeMaps(stateA.labelerAggs, stateB.labelerAggs),
     labelerViewers: mergeMaps(stateA.labelerViewers, stateB.labelerViewers),
     knownFollowers: mergeMaps(stateA.knownFollowers, stateB.knownFollowers),
+    bidirectionalBlocks: mergeMaps(
+      stateA.bidirectionalBlocks,
+      stateB.bidirectionalBlocks,
+    ),
```
devinivy, line 1070: The way `bidirectionalBlocks` works, I think this would allow some clobbering to happen.  For example, how should these two merge?

```
{ did:example:a -> { did:example:b -> true } }
{ did:example:a -> { did:example:c -> true } }
```

I think we'd want to find something like this:
```
{ did:example:a -> { did:example:b -> true, did:example:c -> true } }
```

packages/bsky/src/hydration/hydrator.ts
estrattonbailey: Oy 🤦

packages/bsky/src/hydration/util.ts
```diff
@@ -25,6 +25,34 @@ export type RecordInfo<T> = {
   takedownRef: string | undefined
 }
 
+export const mergeMaps = <V, M extends HydrationMap<V>>(
+  mapA?: M,
+  mapB?: M,
+): M | undefined => {
```
estrattonbailey, line 31: @devinivy special note: I axed the `Merges` interface here so that I could use this more easily within `mergeNestedMaps`. No change to types or functionality that I can tell, but unsure if I knocked over Chesteron's Fence anywhere else.

estrattonbailey: Also general note that I extracted these and made a `util.test.ts` test

devinivy (approved):
Good stuff!

PR #2628, opened 2024-07-07
Change max length of names to 64 graphemes

haileyok:
## Why

Something that got surfaced by starter packs is that lists currently allow only 64 _utf-8 bytes_, not 64 graphemes. This is extremely restrictive, especially for foreign languages which use non-roman alphabets. For example, just adding the world `hello` to a list name (i.e. `こんにちは`) takes up 15 bytes.

I think bumping this to what I suspect is the intended max limit should be okay, since any previously created list should already be compliant with this.

### DIFF
--- a/lexicons/app/bsky/graph/defs.json
+++ b/lexicons/app/bsky/graph/defs.json
@@ -8,7 +8,12 @@
       "properties": {
         "uri": { "type": "string", "format": "at-uri" },
         "cid": { "type": "string", "format": "cid" },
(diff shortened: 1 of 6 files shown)


bnewbold (approved):
Feels good and comfy to me. I would guess these values were mostly copypasta between Lexicons and should have been 64 graphemes from the start.

In terms of roll-out, I think doing everything but PDS for a week, then rolling out to PDS, is what would go easiest. Not sure how hard that would be to coordinate.

Vincentroland (approved):

PR #2629, opened 2024-07-08
Filter out reference lists from `getLists`

haileyok:
These lists are not used for anything other than little tools for starter packs. They shouldn't show up in responses for `getLists`. Our client is already filtering them out, but this will ensure other clients don't need to implement that logic themselves.


devinivy (approved):
Just made a small change to use our "rules" step to pull out the reference lists.   Looks great, thanks!

PR #2630, opened 2024-07-08
Update data source for `getSuggestedFollowsByActor`

estrattonbailey:
We've added a new parameter to our otherwise generic `app.bsky.unspecced.getSuggestionsSkeleton` called `relativeToDid`, which requests suggestions relative to that DID instead of that of the `viewer`. The intent here is to unify the algorithm we use to suggest follows to the `viewer` by replacing the algo that powers `getSuggestedFollowsByActor` with the one we use for our other suggestions in the app.

I've integrated this unspecced endpoint and new param into the existing endpoint and put it behind a feature flag so we can selectively turn it on to test.

To do so, this PR also integrates StatSig. I've added a mgmt class `FeatureGates` that lives on `AppContext` and integrates with the `start` and `destroy` lifecycle of the app's context.

It can be used within route handlers like this:
```typescript
const gates = ctx.featureGates

if (gates.check(
  await gates.user({ did: '...' }),
  gates.ids.MyGateId
)) {
  // gate enabled, do something
}
```

In testing, StatSig is put into "local mode", which allows you to override flags for testing. See [here for more info.](https://docs.statsig.com/server/nodejsServerSDK#local-overrides)


packages/bsky/src/api/app/bsky/graph/getSuggestedFollowsByActor.ts
```diff
@@ -27,14 +30,27 @@ export default function (server: Server, ctx: AppContext) {
...
-      const result = await getSuggestedFollowsByActor(
-        { ...params, hydrateCtx: hydrateCtx.copy({ viewer }) },
-        ctx,
-      )
+      const headers = noUndefinedVals({
+        'accept-language': req.headers['accept-language'],
+        'x-bsky-topics': Array.isArray(req.headers['x-bsky-topics'])
+          ? req.headers['x-bsky-topics'].join(',')
+          : req.headers['x-bsky-topics'],
+      })
```
estrattonbailey, line 38: This handling is based on that of `getSuggestions`, now that we're aiming to unify these

packages/bsky/src/api/app/bsky/graph/getSuggestedFollowsByActor.ts
```diff
@@ -27,14 +30,27 @@ export default function (server: Server, ctx: AppContext) {
...
+        'content-language': resultHeaders?.['content-language'],
+      })
       return {
         encoding: 'application/json',
         body: result,
-        headers: resHeaders({ labelers: hydrateCtx.labelers }),
+        headers: {
+          ...responseHeaders,
+          ...resHeaders({ labelers: hydrateCtx.labelers }),
+        },
```
estrattonbailey, line 53: This handling is based on that of `getSuggestions`, now that we're aiming to unify these

packages/bsky/src/index.ts
```diff
@@ -116,6 +117,12 @@ export class BskyAppView {
       adminPasses: config.adminPasswords,
     })
 
+    if (config.statsigKey) {
+      Statsig.initialize(config.statsigKey, {
+        environment: { tier: config.statsigEnv },
+      })
```
estrattonbailey, line 123: [This is async](https://docs.statsig.com/server/nodejsServerSDK#initialize-the-sdk), but this happens on server start and then periodically in the background to get updated gates, so I'm thinking that we probably don't need to await here.

estrattonbailey: Confirmed with StatSig team: calls to `checkGateSync` (checking a feature flag) that happen synchronously after this call will be fine and return `false` by default. However, gate check calls should NOT happen before `initialize()` is called, since they will throw.

Since this is called on server startup, any gate check calls should happen after `initialize()`, so I think we don't need to worry about this. Could use a second opinion though.

lexicons/app/bsky/unspecced/getSuggestionsSkeleton.json
```diff
@@ -19,7 +19,12 @@
             "maximum": 100,
             "default": 50
           },
-          "cursor": { "type": "string" }
+          "cursor": { "type": "string" },
+          "relative_to_did": {
```
devinivy, line 23: I think this is the only place `snake_case` is used as a lexicon field—could we bring this in line with the `camelCase` convention?

packages/bsky/src/api/app/bsky/graph/getSuggestedFollowsByActor.ts
```diff
@@ -46,13 +62,32 @@ const skeleton = async (input: SkeletonFnInput<Context, Params>) => {
...
-    relativeToDid,
-  })
-  return {
-    suggestedDids: dids,
-    cursor: cursor || undefined,
+
+  const statsigUser = await didToStatsigUser(params.hydrateCtx.viewer)
+  if (
+    ctx.suggestionsAgent &&
+    Statsig.checkGateSync(statsigUser, gates.newSuggestedFollowsByActor)
```
devinivy, line 69: My understanding is that this would throw if statsig isn't configured, which is allowed by the appview configuration.  In prod, for example, there's currently a `suggestionsAgent` but no statsig config, so I believe this endpoint would be throwing 500s.  If we can fail open here instead that would be awesome.

packages/bsky/src/index.ts
```diff
@@ -116,6 +117,12 @@ export class BskyAppView {
       adminPasses: config.adminPasswords,
     })
 
+    if (config.statsigKey) {
+      Statsig.initialize(config.statsigKey, {
+        environment: { tier: config.statsigEnv },
+      }).catch((_) => {})
```
devinivy, line 123: My understanding of the statsig sdk is that it needs to make it into a "ready" state before you can start checking gates, and awaiting `initialize()` is how you ensure you're in that state.  By silently ignoring the error here, we could end-up not in a ready state and then I'm pretty sure checking gates would always fail.

We should be able to move the await into the context's lifecycle `start()`.  By a similar token we also want to shut statsig down cleanly so that the process doesn't hang on shutdown due to open handles.  And we can perform that in the context's lifecycle `destroy()`.

packages/crypto/src/sha.ts
```diff
@@ -2,6 +2,7 @@ import * as noble from '@noble/hashes/sha256'
 import * as uint8arrays from 'uint8arrays'
 
 // takes either bytes of utf8 input
+// TODO this can be sync
```
devinivy, line 5:
```suggestion
// @TODO this can be sync
```

packages/crypto/src/sha.ts
```diff
@@ -10,6 +11,7 @@ export const sha256 = async (
   return noble.sha256(bytes)
 }
 
+// TODO this can be sync
```
devinivy, line 14:
```suggestion
// @TODO this can be sync
```

devinivy (commented):
Lookin good, just left a few notes!  The main ones are a. lifecycle of statsig, b. failing open when statsig is failing or not configured at all.  I think they both could be addressed and brought into a common pattern used in the node services by making a new class for managing feature gates, which will live on the `AppContext` and used as part of the `AppContext`'s lifecycle.  Maybe could look roughly like this:
```ts
interface FeatureGates {
  private statsig: StatsigInstance | undefined
  constructor(config: StatsigConfig | undefined)
  // when statsig is being used, initializes statsig. called in AppContext's start().
  start(): Promise<void>
  // when statsig is being used, shuts-down statsig. called in AppContext's destroy().
  destroy(): Promise<void>
  // when statsig is being used returns statsig's gate, otherwise fails open with a log and returning the default.
  // when statsig is not being used returns a default for the gate (probably false).
  check(gate: Gate, userId: string): boolean
  getUserId(did: string): Promise<string>
}
```

packages/bsky/src/util/statsig.ts
```diff
@@ -0,0 +1,12 @@
...
+export async function didToStatsigUser(did: string) {
+  const userID = await sha256Hex(did)
+  return {
+    userID,
+  }
+}
+
+export const gates = {
+  newSuggestedFollowsByActor: 'new_sugg_foll_by_actor',
+}
```
devinivy, line 12: No biggie either way, but TS enums do kinda rule for this:
```ts
export enum Gate {
  NewSuggestedFollowsByActor = 'new_sugg_foll_by_actor',
}
```

lexicons/app/bsky/unspecced/getSuggestionsSkeleton.json
estrattonbailey: Oh sure

packages/bsky/src/api/app/bsky/graph/getSuggestedFollowsByActor.ts
estrattonbailey: Ah true, should have a fallback 👍

packages/bsky/src/index.ts
estrattonbailey: My understanding based on the docs is that even if this fails to load the fresh gates, setup has already occurred and gate checks will return `false` by default, not throw.

But ah yes, good call, I didn't realize how the lifecycle worked here! I'll move that in. Actually started this branch by adding to app context but reverted for the simpler option 🙃

packages/bsky/src/index.ts
```diff
@@ -161,10 +168,12 @@ export class BskyAppView {
     await events.once(server, 'listening')
     const { port } = server.address() as AddressInfo
     this.ctx.cfg.assignPort(port)
+    await this.ctx.featureGates.start()
     return server
   }
 
   async destroy(): Promise<void> {
+    this.ctx.featureGates.destroy()
     await this.terminator?.terminate()
```
devinivy, line 177: Swapping these allows features gates to continue to be applied while requests drain from the server.
```suggestion
    await this.terminator?.terminate()
    this.ctx.featureGates.destroy()
```

packages/bsky/src/index.ts
```diff
@@ -161,10 +168,12 @@ export class BskyAppView {
     await events.once(server, 'listening')
     const { port } = server.address() as AddressInfo
     this.ctx.cfg.assignPort(port)
+    await this.ctx.featureGates.start()
```
devinivy, line 171: I think we'll want to move this up to before the server is listening / fielding requests.

packages/bsky/src/feature-gates.ts
```diff
@@ -0,0 +1,67 @@
...
+        await this.statsig.initialize('secret-key', {
+          localMode: true,
+          environment: {
+            tier: 'test',
+          },
+        })
+        this.ready = true
+      }
+    } catch (_) {
+      this.ready = false
```
devinivy, line 45: We should at minimum log the error so we can see there's an issue—you'll find some loggers in `./logger` and we can add a new one if we want!

packages/bsky/src/feature-gates.ts
```diff
@@ -0,0 +1,67 @@
...
+  async start() {
+    try {
+      if (this.config.apiKey) {
+        await this.statsig.initialize(this.config.apiKey, {
+          environment: {
+            tier: this.config.env || 'development',
+          },
+        })
+        this.ready = true
+      } else if (process.env.NODE_ENV === 'test') {
```
devinivy, line 32: Would it be possible to hoist this behavior up either to `config.ts` or even the appview's statsig config in the tests?  This isn't a major deal / can totally punt for now and just leave a note if it's a pain!  The reason we've been trying to do things that way is just to avoid behaviors that don't flow from the app entrypoint/config, and perhaps a sprinkle of brain being a little wired to certain [12 factor](https://arc.net/l/quote/yrqrzool) thinking 🤣 .

devinivy (commented):
Clean!!  Just a few tiiiny tweaks and I think we'll be rockin.

packages/bsky/src/index.ts
estrattonbailey: Oh awesome, makes sense

devinivy (approved):

devinivy (approved):
Ace!

PR #2633, opened 2024-07-10
OAuth: 2FA

matthieusieben:
This PR contains the non-SDK changes of [`feat-oauth-client`](https://github.com/bluesky-social/atproto/tree/feat-oauth-client) (#2483)


packages/oauth/oauth-provider/src/assets/app/views/error-view.tsx
```diff
@@ -1,25 +1,33 @@
...
   const desc = errorData?.error_description
-  switch (desc) {
-    case 'Unknown request_uri': // Request was removed from database
-    case 'This request has expired':
+  switch (true) {
+    case desc?.startsWith('Unknown request_uri'): // Request was removed from database
+    case desc === 'This request has expired':
       return 'This sign-in session has expired'
     default:
       return desc || 'An unknown error occurred'
```
devinivy, line 33: Just a small style nit, but I find it quite a bit more common to see a regular old `if` statement for this kind of thing.
```ts
if (
  desc?.startsWith('Unknown request_uri') ||
  desc === 'This request has expired'
) {
  // ...
} else {
  // ...
}
```

packages/oauth/oauth-provider/src/output/customization.ts
```diff
@@ -75,13 +62,15 @@ function parseColor(color: string): undefined | ParsedColor {
     return undefined
   }
 
-  const rgbMatch = color.match(/rgb\((\d+),\s*(\d+),\s*(\d+)\)/)
+  const rgbMatch = color.match(/rgb\(\s*(\d+)\s*,\s*(\d+)\s*,\s*(\d+)\s*\)/)
```
devinivy, line 65: Do we want some `^`/`$` in here?  (And same question below for rgba.)

packages/oauth/oauth-client/src/oauth-client.ts
```diff
@@ -297,6 +347,13 @@ export class OAuthClient {
     )
   }
 
+  async abortRequest(authorizeUrl: URL) {
+    const requestUri = authorizeUrl.searchParams.get('request_uri')
+    if (!requestUri) return
+
+    // @TODO find a way to actually do this
```
devinivy, line 354: Just flagging to ensure this TODO is still intended to be here.

packages/oauth/oauth-client/src/oauth-client.ts
```diff
@@ -91,7 +90,44 @@ export type OAuthClientOptions = {
...
+    if (response.status !== 200) {
+      throw new TypeError(`Failed to fetch client metadata: ${response.status}`)
+    }
+
+    const mime = response.headers.get('content-type')?.split(';')[0].trim()
+    if (mime !== 'application/json') {
+      throw new TypeError(`Invalid client metadata content type: ${mime}`)
+    }
+
+    const json: unknown = await response.json()
```
devinivy, line 124: In the case of an error it looks like the response body would not be read.  If I'm not mistaken, in node environments the body should always be read to avoid a leak.

devinivy (approved):
Had a few small notes, but this looks great 👍

packages/oauth/oauth-client-browser/src/browser-runtime-implementation.ts
```diff
@@ -34,32 +39,14 @@ export class BrowserRuntimeImplementation implements RuntimeImplementation {
...
-  name,
-}: DigestAlgorithm): AlgorithmIdentifier {
-  switch (name) {
-    case 'sha256':
-    case 'sha384':
-    case 'sha512':
-      return `SHA-${name.slice(-3)}`
-    default:
-      throw new TypeError(`Unknown hash algorithm ${name}`)
+    const buf = await crypto.subtle.digest(`SHA-${name.slice(3)}`, data)
```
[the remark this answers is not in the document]
matthieusieben: not sure why I didn't keep the check... i'll add it back

packages/oauth/oauth-client/src/oauth-client.ts
matthieusieben: This is fine for now. Improved my comments there.

packages/oauth/oauth-client/src/oauth-client.ts
matthieusieben: nice catch!

PR #2636, opened 2024-07-11
:sparkles: Ozone sets

foysalit:
This PR implements sets api in ozone.

The initial use case for sets api is let moderators use ozone to manage some of the configurations for automod. In general, this module is helpful for storing long lists of string values associated with a key and description as metadata.


lexicons/tools/ozone/set/defs.json
```diff
@@ -0,0 +1,49 @@
...
+        "name": {
+          "type": "string",
+          "minLength": 3,
+          "maxLength": 128
+        },
+        "description": {
+          "type": "string",
+          "minLength": 1,
+          "maxLength": 1024
+        }
```
bnewbold, line 18: It's a nit, but for internationalization we are often doing "descriptions" as a grapheme limit, then 10x that as maxLength. so:

```
maxGraphemes: 1024
maxLength: 10240
```

and then update in the views/references to this field

lexicons/tools/ozone/set/remove.json
```diff
@@ -0,0 +1,30 @@
+{
+  "lexicon": 1,
+  "id": "tools.ozone.sets.remove",
+  "defs": {
+    "main": {
+      "type": "procedure",
+      "description": "Remove values from a specific set",
```
bnewbold, line 7: should describe semantics here in the description: if you try to remove a string which isn't in the set, is that an error? (I think not, but double-checking, and should document it here)

lexicons/tools/ozone/set/removeSet.json
```diff
@@ -0,0 +1,23 @@
+{
+  "lexicon": 1,
+  "id": "tools.ozone.sets.removeSet",
```
bnewbold, line 3: I'd go with `deleteSet` to reduce terminology conflict with `remove`.

not a blocking request, but I think it would be better.

lexicons/tools/ozone/set/add.json
```diff
@@ -0,0 +1,31 @@
...
+          "type": "object",
+          "required": ["name", "values"],
+          "properties": {
+            "name": {
+              "type": "string",
+              "description": "Name of the set to add values to"
+            },
+            "values": {
+              "type": "array",
+              "minLength": 1,
```
bnewbold, line 20: `maxLength` as 250 or 1000 or so?

lexicons/tools/ozone/set/querySets.json
```diff
@@ -0,0 +1,51 @@
...
+            "maximum": 100,
+            "default": 50
+          },
+          "cursor": {
+            "type": "string"
+          },
+          "namePrefix": {
+            "type": "string"
+          },
+          "sortBy": {
```
bnewbold, line 23: maybe add a "reverse sort order" flag called `reverseSort` (?) while we are at it? not a blocking need (we could add it later if needed), just a thought.

lexicons/tools/ozone/set/removeSet.json
```diff
@@ -0,0 +1,23 @@
...
+  "defs": {
+    "main": {
+      "type": "procedure",
+      "description": "Remove an entire set",
```
bnewbold, line 7: I would also say in the description whether there is an error if there is an attempt to delete a set which doesn't exist.

@devinivy should we have a house style on this behavior (error or not)? IIRC deleteRecord doesn't error. maybe there is an HTTP success response code variant we could use to indicate this situation?

this isn't blocking feedback, we could move ahead without addressing this one.

lexicons/tools/ozone/set/get.json
```diff
@@ -0,0 +1,50 @@
+{
+  "lexicon": 1,
+  "id": "tools.ozone.sets.get",
+  "defs": {
+    "main": {
+      "type": "query",
+      "description": "Get a specific set and its values",
```
bnewbold, line 7: for automod, it would be good to distinguish between HTTP 404 (set doesn't exist) from other errors (server error, something wrong with query). IIRC we had some history of doing 400 instead of 404.

this isn't feedback on this PR or the lexicon, just thinking outloud for the implementation half of this in ozone repo.

bnewbold (approved):
Finally getting around to looking at this, sorry for the super slow review.

Looking good!

left a couple small comments but nothing here is really blocking and I think we can move ahead.

lexicons/tools/ozone/set/add.json
```diff
@@ -0,0 +1,31 @@
+{
+  "lexicon": 1,
+  "id": "tools.ozone.sets.add",
```
devinivy, line 3: To date there's been a bit of a convention going that lexicon methods are "verb-subject", which serves to distinguish lexicon methods from e.g. records or defs (which are always lowercase).  There aren't any lexicons methods that are all lowercase that I know of.  Should we keep to that convention?  I think it would become something like `sets.addElement` or `sets.addItem`.

@bnewbold I think you may have proposed these lexicons, so tagging you in!

devinivy: Another small one— to date there aren't any pluralized namespaces either, we could use `tools.ozone.set` if we wanted to continue with that.

lexicons/tools/ozone/set/add.json
```diff
@@ -0,0 +1,31 @@
...
+            "name": {
+              "type": "string",
+              "description": "Name of the set to add values to"
+            },
+            "values": {
+              "type": "array",
+              "minLength": 1,
+              "items": {
+                "type": "string"
+              },
```
devinivy, line 23: Do we want any limits on set item sizes?

lexicons/tools/ozone/set/add.json
```diff
@@ -0,0 +1,31 @@
...
+      "input": {
+        "encoding": "application/json",
+        "schema": {
+          "type": "object",
+          "required": ["name", "values"],
+          "properties": {
+            "name": {
+              "type": "string",
+              "description": "Name of the set to add values to"
+            },
```
devinivy, line 17: Should we reflect the set name restrictions (length between `3` and `128`) here?

lexicons/tools/ozone/set/defs.json
```diff
@@ -0,0 +1,49 @@
...
+  "id": "tools.ozone.sets.defs",
+  "defs": {
+    "set": {
+      "type": "object",
+      "required": ["name"],
+      "properties": {
+        "name": {
+          "type": "string",
+          "minLength": 3,
+          "maxLength": 128
```
devinivy, line 12: Just vibes, but this number sounds a little low.

lexicons/tools/ozone/set/get.json
devinivy: There still isn't a 404 in the standard xrpc error types we have on hand, and using it is a little confusing because the client converts it into an `XRPCNotSupported` error.  As you know, I am a fan of getting this going!  But that is the source of the awkwardness at this point.

lexicons/tools/ozone/set/remove.json
```diff
@@ -0,0 +1,30 @@
+{
+  "lexicon": 1,
+  "id": "tools.ozone.sets.remove",
```
devinivy, line 3: On the lexicon naming conventions again— in this PR we have `sets.remove` and `sets.removeSet`.  I think this also sort of points towards moving to `sets.removeItems` and `sets.removeSet` (or even `sets.deleteSet`).

lexicons/tools/ozone/set/upsertSet.json
```diff
@@ -0,0 +1,24 @@
+{
+  "lexicon": 1,
+  "id": "tools.ozone.sets.upsertSet",
```
devinivy, line 3: We totally don't have to use it, but figure it's worth mentioning that we use the name `repo.putRecord` for upserting records into a repo.

lexicons/tools/ozone/set/upsertSet.json
```diff
@@ -0,0 +1,24 @@
...
+        "schema": {
+          "type": "ref",
+          "ref": "tools.ozone.sets.defs#set"
+        }
+      },
+      "output": {
+        "encoding": "application/json",
+        "schema": {
+          "type": "ref",
+          "ref": "tools.ozone.sets.defs#set"
```
devinivy, line 19: Should these set outputs be `#setView` rather than `#set`?  I am sort of assuming `#set` is designed for writing and `#setView` for reading.

lexicons/tools/ozone/set/removeSet.json
devinivy: There's no success status code that I know of— I'd recommend adding a field to the response if you both want idempotent deletions (which I support) and knowledge of whether anything got deleted.

lexicons/tools/ozone/set/removeSet.json
devinivy, line 3: I'd recommending having an `output` from this lexicon, even if it's empty for now.  It helps with future-proofing, since it's sort of a breaking change to transition from no response to some response.

lexicons/tools/ozone/set/remove.json
devinivy: If within `tools.ozone.sets` we work with both "items" and "sets", I would assume that the methods such as `sets.remove` apply to sets.  I.e. I'd assume "sets" are what get elided from the name.

lexicons/tools/ozone/set/defs.json
```diff
@@ -0,0 +1,49 @@
+{
+  "lexicon": 1,
+  "id": "tools.ozone.sets.defs",
```
devinivy, line 3: The directory structure differs from the name here: `tools/ozone/set/defs.json` vs `tools.ozone.sets.defs`, i.e. `set` vs. `sets`.  I have slight preference for the latter, as it seems to fit more neatly with conventions (singular preferred over plural).

packages/ozone/src/db/migrations/20240710T051104136Z-sets.ts
```diff
@@ -0,0 +1,63 @@
+import { Kysely, sql } from 'kysely'
+
+export async function up(db: Kysely<unknown>): Promise<void> {
+  // Create the sets table
+  await db.schema
+    .createTable('ozone_set')
```
devinivy, line 6: I assume `ozone_` is prefixed here to avoid naming the table the reserved word "set".  There could be other ways that feel natural to name the tables and still avoid the reserved word, e.g. `set_detail` and `set_item`.

packages/ozone/src/db/migrations/20240710T051104136Z-sets.ts
```diff
@@ -0,0 +1,63 @@
...
+    .addColumn('id', 'serial', (col) => col.primaryKey())
+    .addColumn('name', 'varchar(128)', (col) => col.notNull().unique())
+    .addColumn('description', 'varchar(1024)')
```
devinivy, line 9: IMO it's probably fine to just use `varchar` if you like.  In postgres there's no difference in the storage between `varchar(128)`, `varchar`, and `text` under the hood.  It just adds a validation check on the length.

packages/ozone/src/db/migrations/20240710T051104136Z-sets.ts
```diff
@@ -0,0 +1,63 @@
...
+    .addColumn('setId', 'integer', (col) =>
+      col.notNull().references('ozone_set.id').onDelete('cascade'),
+    )
+    .addColumn('value', 'varchar(255)', (col) => col.notNull())
+    .addColumn('createdAt', 'timestamptz', (col) =>
+      col.defaultTo(sql`now()`).notNull(),
+    )
+    .execute()
+
+  // Add indexes for better performance
```
devinivy, line 31: These indexes don't especially cater to some of the pagination that we do, and some small tweaks may make a big difference.  For example we have two separate indexes for `ozone_set_value` on `setId` and `value`.  The access pattern we have for paginating over values within a set is to filter by `setId` and order those by `value`.  To best support this pagination we probably want a single index on `(setId, value)` rather than two independent indexes on each column.

devinivy: Ah I see that we do have a unique constraint on `(setId, value)` which would do the trick there.  So perhaps we can instead get rid of some of these indexes! 🎉

packages/ozone/src/api/set/add.ts
```diff
@@ -0,0 +1,39 @@
...
+
+      const setService = ctx.setService(db)
+
+      await db.transaction(async (txn) => {
+        const set = await setService.getByName(name, txn)
+        if (!set) {
+          throw new InvalidRequestError(
+            `Set with name "${name}" does not exist`,
+          )
+        }
```
devinivy, line 33: I'd recommend moving this fetch out of the transaction—it leaves the transaction open longer but doesn't give us any additional guarantees.  For example, `set` could still be deleted before the call to `addValues()` below.

packages/ozone/src/api/set/get.ts
```diff
@@ -0,0 +1,42 @@
...
+import { Server } from '../../lexicon'
+import AppContext from '../../context'
+
+export default function (server: Server, ctx: AppContext) {
+  server.tools.ozone.sets.get({
+    auth: ctx.authVerifier.modOrAdminToken,
+    handler: async ({ params, auth }) => {
+      const access = auth.credentials
+      const db = ctx.db
+      const { name, limit = 100, cursor } = params
```
devinivy, line 11: Lexicon oughtta be applying the default for us here.
```suggestion
      const { name, limit, cursor } = params
```

packages/ozone/src/api/set/querySets.ts
```diff
@@ -0,0 +1,35 @@
...
+import { Server } from '../../lexicon'
+import AppContext from '../../context'
+
+export default function (server: Server, ctx: AppContext) {
+  server.tools.ozone.sets.querySets({
+    auth: ctx.authVerifier.modOrAdminToken,
+    handler: async ({ params, auth }) => {
+      const access = auth.credentials
+      const db = ctx.db
+      const { limit = 50, cursor, namePrefix, sortBy = 'name' } = params
```
devinivy, line 11: Lexicon should have our back here with the default.
```suggestion
      const { limit, cursor, namePrefix, sortBy = 'name' } = params
```

packages/ozone/src/api/set/remove.ts
```diff
@@ -0,0 +1,39 @@
...
+
+      const setService = ctx.setService(db)
+
+      await db.transaction(async (txn) => {
+        const set = await setService.getByName(name, txn)
+        if (!set) {
+          throw new InvalidRequestError(
+            `Set with name "${name}" does not exist`,
+          )
+        }
```
devinivy, line 33: Should be able to safely pull this fetch and check outside of the transaction.

packages/ozone/src/api/set/removeSet.ts
```diff
@@ -0,0 +1,35 @@
...
+
+      const setService = ctx.setService(db)
+
+      await db.transaction(async (txn) => {
+        const set = await setService.getByName(name, txn)
+        if (!set) {
+          throw new InvalidRequestError(
+            `Set with name "${name}" does not exist`,
+          )
+        }
```
devinivy, line 29: Should be able to safely pull this up out of the transaction.

packages/ozone/src/api/set/upsertSet.ts
```diff
@@ -0,0 +1,44 @@
...
+          txn,
+        )
+      })
+
+      return {
+        encoding: 'application/json',
+        body: {
+          name: upsertedSet.name,
+          description: upsertedSet.description ?? undefined,
+        },
```
devinivy, line 40: Since `description` defaults to `''` above, I wonder if this could end-up being an empty string, which isn't allowed by the lexicon.

packages/ozone/src/set/service.ts
```diff
@@ -0,0 +1,203 @@
...
+        (join) => join.onRef('sv.setId', '=', 's.id'),
+      )
+      .select([
+        's.id',
+        's.name',
+        's.description',
+        's.createdAt',
+        's.updatedAt',
+        sql<number>`coalesce(sv.count, 0)`.as('setSize'),
+      ])
```
devinivy, line 48: You may be able to nix the complex join with something like:
```suggestion
      .select([
        's.id',
        's.name',
        's.description',
        's.createdAt',
        's.updatedAt',
        (eb) =>
          eb
            .selectFrom('ozone_set_value')
            .select((e) => e.fn.count('setId').as('count'))
            .whereRef('setId', '=', 's.id')
            .as('setSize'),
      ])
```

packages/ozone/src/set/service.ts
```diff
@@ -0,0 +1,203 @@
...
+      .selectFrom('ozone_set_value')
+      .select(['value'])
+      .where('setId', '=', set.id)
+      .limit(limit)
+
+    if (cursor) {
+      qb = qb.where('value', '>', cursor)
+    }
+
+    qb = qb.orderBy('value', 'asc')
```
devinivy, line 114: We can paginate over `value`, though it does have the property that if you're trying to sync the whole set you can miss values that are added while you're paginating.  It might be natural to page over creation time and break ties with the item `id`, so if items are added while you're paginating you'll only encounter them on the final pages.

packages/ozone/src/set/service.ts
```diff
@@ -0,0 +1,203 @@
...
+
+    return await query.executeTakeFirstOrThrow()
+  }
+
+  async addValues(
+    setId: number,
+    values: string[],
+    txn: Database,
+  ): Promise<void> {
+    const query = txn.db
```
devinivy, line 151: If `txn` this must be a transaction for correctness / data integrity it can be enforced here:
```suggestion
    txn: Database,
  ): Promise<void> {
    txn.assertTransaction()
    const query = txn.db
```

packages/ozone/src/set/service.ts
```diff
@@ -0,0 +1,203 @@
...
+
+  async removeValues(
+    setId: number,
+    values: string[],
+    txn: Database,
+  ): Promise<void> {
+    const query = txn.db
+      .deleteFrom('ozone_set_value')
+      .where('setId', '=', setId)
+      .where('value', 'in', values)
```
devinivy, line 174: This will fail with a sql syntax error if `values` is empty.

packages/ozone/src/set/service.ts
```diff
@@ -0,0 +1,203 @@
...
+      .where('id', '=', setId)
+      .execute()
+  }
+
+  async deleteSet(setId: number, txn: Database): Promise<void> {
+    await txn.db.deleteFrom('ozone_set').where('id', '=', setId).execute()
+    await txn.db
+      .deleteFrom('ozone_set_value')
+      .where('setId', '=', setId)
+      .execute()
```
devinivy, line 191: I think you have the `ozone_set_value` deletions set to cascade.  I'd recommend either a. removing the cascade from the schema or b. leaving a comment here indicating that the values will be deleted from the cascade.

packages/ozone/tests/sets.test.ts
```diff
@@ -0,0 +1,198 @@
...
+
+  const sampleSet2 = {
+    name: 'test-set-2',
+    description: 'Test set 2',
+  }
+
+  const sampleSet3 = {
+    name: 'another-set',
+    description: 'Another test set',
+  }
```
devinivy, line 24: May be worth making one of these samples be missing a `description`.

devinivy (commented):
I dig how this is coming along!

lexicons/tools/ozone/set/querySets.json
foysalit: yep, can't recall why I didn't do this in the first place. decided to go with `sortDirection: 'asc' | 'desc'` because we already have this pattern for `queryEvents`

packages/ozone/src/set/service.ts
foysalit: we have a check in `/set/remove.ts` to ensure that values is not empty before calling this. added a check in this method too.

packages/ozone/src/set/service.ts
foysalit: ended up moving the transactions inside the methods since we moved all the additional checks out of the transactions.

packages/ozone/src/db/migrations/20240710T051104136Z-sets.ts
foysalit: yep yep. leaving the index on `name`

packages/ozone/src/set/service.ts
foysalit: good catch! i'll get rid of the cascade from schema.

lexicons/tools/ozone/set/removeSet.json
foysalit: hmm... I kinda like removeSet since it's suffixed already, don't feel like it'll be confusing.

packages/ozone/tests/sets.test.ts
foysalit: great catch, moved around some lexicon and found some issues with how description is handled. everything should be better now. a quick summary
1. `description` is truly optional now. in both input (set) and output (setView)
2. test code was not comparing real value of returned data from upsertSetm, this is fixed now.

lexicons/tools/ozone/set/upsertSet.json
foysalit: yep. I was thinking these wouldn't need the value count but I guess it doesn't hurt to unify these all to setView.

lexicons/tools/ozone/set/removeSet.json
foysalit: Added the error case to the description and also added a specific error type for SetNotFound. 
Leading it with an empty response body, should be easy to add a success/failure flag later down the road.

lexicons/tools/ozone/set/add.json
foysalit: yeah I was going off the proposed lexicons by bryan but yeah, agreed that it would be more consistent to name it something like `addItems` or `addValues`

lexicons/tools/ozone/set/add.json
foysalit: this is only for specifying which set the values will be added to so the caller is expected to either know the name of the set or fetch the name in a previous call to `querySets` so the restriction on name doesn't make a lot of sense to me. wdyt?

lexicons/tools/ozone/set/defs.json
foysalit: based on @bnewbold 's proposal so I'll wait for him to chime in.

foysalit:
@devinivy I think I've addressed everything you pointed out except for some of the naming suggestions that contradicts what @bnewbold proposed so I'll wait for him to chime in on those but the rest are RFR again.

lexicons/tools/ozone/set/add.json
bnewbold: ah, you are right about `verbSubject`. I agree we should reserve "single word lower-case" for records ("nouns") not endpoints ("verbs"). so probably `addValues` here. that makes the `remove` situation clearer as well.

for `tools.ozone.set`, "set" as a word is a bit weird (both a noun and verb in English) and "sets" clarifies that a bit. I would probably lean devin's way and prefer singular though.

lexicons/tools/ozone/set/add.json
bnewbold: that is a good point. I think something like 2048 bytes? or 8kbyte to align with the `uri` string format? don't need a grapheme limit for this one.

lexicons/tools/ozone/set/defs.json
bnewbold: I'm ok with it being low. this is basically a "slug" variable name, or similar to a record-key (though I wouldn't use the record key string format specifically). probably the backend implementation should enforce that it has no whitespace when doing an upsert? and we could describe that here in a description in the lexicon.

lexicons/tools/ozone/set/remove.json
bnewbold: "values" is a bit more natural to me than "items", but i'd be ok with either.

agree on "nounVerb" for all these; there was a parallel review thread about "delete" vs "remove" and I'd still lean a bit towards "delete" for the overall set (vs "removeValues" for values"), but not a blocking thing for me.

lexicons/tools/ozone/set/removeSet.json
bnewbold: this is a great point for a general lexicon style guide, I would not have thought of this.

bnewbold:
replied on naming points

Vincentroland (approved):

packages/ozone/src/api/set/deleteValues.ts
```diff
@@ -0,0 +1,31 @@
...
+    handler: async ({ input, auth }) => {
+      const access = auth.credentials
+      const db = ctx.db
+      const { name, values } = input.body
+
+      if (!access.isAdmin) {
+        throw new AuthRequiredError(
+          'Must be a moderator to remove values from a set',
+        )
+      }
```
devinivy, line 17: Perhaps this is intended to be `access.isModerator`— would match perms on other set endpoints and this error message.
```suggestion
      if (!access.isModerator) {
        throw new AuthRequiredError(
          'Must be a moderator to remove values from a set',
        )
      }
```

packages/ozone/src/set/service.ts
```diff
@@ -0,0 +1,227 @@
...
+    const set = await this.getByNameWithSize(name)
+    if (!set) return undefined
+
+    const { ref } = this.db.db.dynamic
+    let qb = this.db.db
+      .selectFrom('set_value')
+      .selectAll()
+      .where('setId', '=', set.id)
+
+    const keyset = new TimeIdKeyset(ref(`createdAt`), ref('id'))
```
devinivy, line 123: I didn't catch an index supporting this pagination scheme.

packages/ozone/src/set/service.ts
```diff
@@ -0,0 +1,227 @@
...
+  async upsert({
+    name,
+    description,
+  }: Pick<SetDetail, 'name' | 'description'>): Promise<void> {
+    await this.db.db
+      .insertInto('set_detail')
+      .values({
+        name,
+        description,
+        updatedAt: new Date(),
```
devinivy, line 148: I think you have postgres setup to set this field automatically on insert (same as `createdAt`).

packages/ozone/src/db/schema/ozone_set.ts
```diff
@@ -0,0 +1,24 @@
...
+  description: string | null
+  createdAt: Generated<Date>
+  updatedAt: Generated<Date>
+}
+
+export interface SetValue {
+  id: GeneratedAlways<number>
+  setId: number
+  value: string
+  createdAt: string
```
devinivy, line 18: It looks like this column is also `Generated` based on the migration.  However, it looks like in practice we mostly aren't relying on postgres to set these default values—I think it's only used for `SetDetail.createdAt`.  One option is to leave those defaults in postgres as a backstop, but remove them from here in the kysely model and always set these values explicitly.

bnewbold:
I think i'm up to date on naming review/feedback; let me know if there are any last things to look at.

packages/ozone/src/db/schema/ozone_set.ts
foysalit: yeah i was seeing some type errors and didn't wanna f around so just moved it to string but Generated<Date> also works and went with that.

foysalit: to clarify, still inserting the date values from JS and not really depending on postgres.

packages/ozone/src/db/migrations/20240909T205730722Z-sets.ts
```diff
@@ -0,0 +1,53 @@
...
+  // Create a unique constraint on setId and value
+  await db.schema
+    .alterTable('set_value')
+    .addUniqueConstraint('set_value_setid_value_unique', ['setId', 'value'])
+    .execute()
+  await db.schema
+    .createIndex('set_value_created_at_idx')
+    .on('set_value')
+    .column('createdAt')
+    .execute()
```
devinivy, line 47: This would index `createdAt` across sets.  I believe that the kind of pagination we do is across all values within a particular set, so to take advantage of the index I think it would need to be something like `("setId", "createdAt")`.

devinivy (approved):

lexicons/tools/ozone/set/upsertSet.json
```diff
@@ -0,0 +1,24 @@
...
+  "id": "tools.ozone.set.upsertSet",
+  "defs": {
+    "main": {
+      "type": "procedure",
+      "description": "Create or update set metadata",
+      "input": {
+        "encoding": "application/json",
+        "schema": {
+          "type": "ref",
+          "ref": "tools.ozone.set.defs#set"
```
MarshalX, line 12: Hi there! Do you have any particular reason to make it a reference? This is a historical moment where the ref is used in the input of the procedure. Typically it is an object. I do not see any reuse of `tools.ozone.set.defs#set`. Only as input... 

I noticed because my Python code generator started raising errors because I did not implement ref resolving in the input schema. Now I've implemented it, but does it worth it

cc @bnewbold, @devinivy

bnewbold: good question! I'm not sure if we support it other codegen tooling yet, guess we will find out.

my initial thought is that it is fine? the ref semantics are that this is basically the same as copying the `tools.ozone.set.defs#set` in to this location, and the type of that schema is `object`.

having it be a ref is nice because then the schema only needs to be updated in one place. even if it is only used in this one place in this set of Lexicions, having it be in a defs means other external Lexicons can reuse the definition (which they would not be able to do if it was just the "schema" of a "procedure").

I'm not sure if I would feel the same way about trying to use a ref in, for example, query params. things are basically symmetrical between `input` and `output`, and we allow refs in `output`, so seems like they should be allowed here.

PR #2646, opened 2024-07-17
Small OAuth client adaptations to proposal

matthieusieben:


packages/oauth/oauth-types/src/util.ts
```diff
@@ -18,3 +18,11 @@ export function isLoopbackUrl(input: URL | string): boolean {
...
   return isLoopbackHost(url.hostname)
 }
+
+export function safeUrl(input: URL | string): URL | null {
+  try {
+    return new URL(input)
+  } catch {
+    return null
+  }
+}
```
ThisIsMissEm, line 28: You could also do `URL.canParse` now, which might be helpful here (though it may still need polyfills)

PR #2651, opened 2024-07-19
Appview: enable insight into full thread context

devinivy:
This allows mods to view threads intact when there are 3p blocks amongst post authors.


foysalit (approved):
nice! would it make sense to use the same criteria to bypass blocks in `getAuthorFeed` too? not sure if the `include3pBlocks` naming is clear enough for this but i'm primarily interested in letting mods see author feed of users that block our moderation account.

PR #2653, opened 2024-07-19
exact matches always show profile regardless of block status

haileyok:
## Why

In an effort to make it easier to find accounts that either you have blocked or that you have been blocked by, exact matches may continue to return an item in the typeahead endpoint. This isn't a differnet level of information that we already show in a user's blocked accounts list.

## Test Plan

Block `carla.test`, and see that the typeahead doesn't return that result until the exact phrase `carla.test` is provided to the typeahead endpoint.

I made some test updates but see that the actor search test is getting skipped, so didn't push that up.


packages/bsky/src/api/app/bsky/actor/searchActorsTypeahead.ts
```diff
@@ -82,10 +82,14 @@ const hydration = async (
...
-  const { ctx, skeleton, hydration } = inputs
-  skeleton.dids = skeleton.dids.filter(
-    (did) => !ctx.views.viewerBlockExists(did, hydration),
-  )
+  const { ctx, skeleton, hydration, params } = inputs
+  skeleton.dids = skeleton.dids.filter((did) => {
+    // Always display exact matches so that users can find profiles that they have blocked
+    const isExactMatch =
+      hydration.actors?.get(did)?.handle?.toLowerCase() ===
+      params.q?.toLowerCase()
```
gaearon, line 90: Due to `?.` usage this check will be true if both sides yield undefined/null for some reason, is that intentional?

haileyok: @devinivy not sure what can possibly get returned. i assume as long as we get an actor (added a check for that) it should be okay?

packages/bsky/src/api/app/bsky/actor/searchActorsTypeahead.ts
```diff
@@ -82,10 +82,14 @@ const hydration = async (
...
-  const { ctx, skeleton, hydration } = inputs
-  skeleton.dids = skeleton.dids.filter(
-    (did) => !ctx.views.viewerBlockExists(did, hydration),
-  )
+  const { ctx, skeleton, hydration, params } = inputs
+  skeleton.dids = skeleton.dids.filter((did) => {
+    // Always display exact matches so that users can find profiles that they have blocked
+    const actor = hydration.actors?.get(did)
+    if (!actor) return false
+    const isExactMatch = actor.handle?.toLowerCase() === params.q?.toLowerCase()
```
devinivy, line 90: Two minor things to poke at here:
 - we're searching on `params.q ?? params.term` rather than `params.q`.
 - relatedly, if the actor has an invalid handle (missing `actor.handle`) and `params.q` is not set, blocked users with invalid handles will make it through.  That could happen in empty searches (though I think these are incidentally disallowed within the search service), or any search that uses `params.term` rather than `params.q`.

haileyok: Just curious, is #1 a backwards compat thing?

devinivy: I assume so— I think at some point @bnewbold standardized all the search query endpoints on `q` rather than `term`.  Either way, I figure we should either drop support for `term` or fully support it, and it looks like we're still supporting it today.

haileyok: yep sg! missed that, will push something up in a moment.

devinivy (approved):

PR #2657, opened 2024-07-23
Fixed length check for hashtags

nnabeyang:
This PR fixes length check for hastags.

### DIFF
--- a/.changeset/hashtag-grapheme-length-fastpath.md
+++ b/.changeset/hashtag-grapheme-length-fastpath.md
@@ -1,0 +1,5 @@
+---
+'@atproto/api': patch
+---
+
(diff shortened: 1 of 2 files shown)


Vincentroland (approved):

nnabeyang:
@estrattonbailey 
I’m following up on PR #2242, where I’ve modified the hashtag length check to use graphemeLen instead of UTF-16 code points. I understand that this change may introduce additional computational cost due to counting graphemes.

According to the change history, it looks like you reviewed the original code changes, so I’d appreciate it if you could review this PR as well.

matthieusieben:
Hey, thank you for this. Could you please:

1) Find an optimization strategy (e.g. a regex?) to avoid the computation overhead in ASCII only text
2) Add a changeset file
3) Rebase/Merge main

Thank you.

matthieusieben (approved):

nnabeyang:
Thanks for the review! I've addressed the feedback and updated the branch with the latest `main`.

PR #2658, opened 2024-07-24
quote aggs, list quotes

haileyok:


packages/bsky/src/data-plane/server/db/migrations/20240723T220703655Z-quotes.ts
```diff
@@ -0,0 +1,23 @@
+import { Kysely, sql } from 'kysely'
+
+export async function up(db: Kysely<unknown>): Promise<void> {
+  await db.schema
+    .createTable('quote')
```
devinivy, line 5: We probably want to add an index on this table based on how we're going to query it (I anticipate `(subject, sortAt)`?).

packages/bsky/src/data-plane/server/indexing/plugins/post.ts
```diff
@@ -175,6 +175,37 @@ const insertFn = async (
...
+        subjectCid: record.cid,
+        createdAt: normalizeDatetimeAlways(obj.createdAt),
+        indexedAt: timestamp,
+      }
+      await db
+        .insertInto('quote')
+        .values(quote)
+        .onConflict((oc) => oc.doNothing())
+        .returningAll()
+        .executeTakeFirst()
```
devinivy, line 192: Perhaps we only want to index these quotes when `record` is a post.

packages/bsky/src/data-plane/server/indexing/plugins/post.ts
```diff
@@ -228,6 +259,8 @@ const notifsForInsert = (obj: IndexedPost) => {
       })
     }
   }
+
+  // TODO remove this, flagging for reference
```
devinivy, line 263: Perhaps this is still here intentionally— flagging just for good measure.

packages/bsky/src/data-plane/server/indexing/plugins/post.ts
```diff
@@ -334,6 +367,23 @@ const deleteFn = async (
...
+          .selectFrom('quote')
+          .where('quote.subjectCid', '=', deletedPosts.embedCid)
+          .select(countAll.as('count')),
+      })
+      .onConflict((oc) =>
+        oc
+          .column('uri')
+          .doUpdateSet({ quoteCount: excluded(db, 'quoteCount') }),
+      )
+      .execute()
```
devinivy, line 386: Flip side of the insertion path— maybe we only need to do this when the embed is a post.

packages/bsky/src/data-plane/server/routes/quotes.ts
```diff
@@ -0,0 +1,32 @@
...
+      .selectFrom('quote')
+      .where('quote.subject', '=', subject.uri)
+      .select(['quote.uri', 'quote.cid', 'quote.sortAt'])
+
+    const keyset = new TimeCidKeyset(ref('quote.sortAt'), ref('quote.cid'))
+    builder = paginate(builder, {
+      limit,
+      cursor,
+      keyset,
+    })
```
devinivy, line 23: We'll want an index on `quote` based on this access pattern 👍

packages/bsky/src/api/app/bsky/feed/getQuotes.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  const quotesRes = await ctx.hydrator.dataplane.getQuotesBySubject({
+    subject: { uri: params.uri, cid: params.cid },
+    cursor: params.cursor,
+    limit: params.limit,
+  })
+
+  return {
+    refs: quotesRes.refs,
+  }
```
devinivy, line 54: We'll want a cursor to be passed along here as well.

packages/bsky/src/data-plane/server/indexing/plugins/post.ts
haileyok: Might need a refresher haha, but isn't that what `isEmbedRecord()` will ensure?

packages/bsky/src/data-plane/server/indexing/plugins/post.ts
haileyok: Same question here, maybe I'm wrong but I assumed this would only be posts.

```tsx
    db
      .deleteFrom('post_embed_record')
      .where('postUri', '=', uriStr)
      .returningAll()
      .executeTakeFirst(),
```

packages/bsky/src/data-plane/server/routes/quotes.ts
haileyok: Added but do please check, not quite certain if this is what we want.

packages/bsky/src/data-plane/server/indexing/plugins/post.ts
devinivy: I believe this deletes all embeds for the post identified by `uriStr `.  But the embedded record isn't necessarily a post (e.g. could be a list, feedgen, or starter pack).  And in those cases we don't consider them QPs.

packages/bsky/src/data-plane/server/indexing/plugins/post.ts
devinivy: I believe `isEmbedRecord()` ensures that a record is being embedded, but that record isn't necessarily a post— could also be e.g. a list, feedgen, or starter pack.  And I assume we only want embedded posts to be considered QPs.

packages/bsky/src/data-plane/server/db/migrations/20240723T220703655Z-quotes.ts
```diff
@@ -0,0 +1,33 @@
...
+  await db.schema
+    .createIndex('quote_subject_idx')
+    .on('quote')
+    .column('subject')
+    .execute()
+  await db.schema
+    .createIndex('quote_order_by_idx')
+    .on('quote')
+    .columns(['sortAt', 'cid'])
+    .execute()
```
devinivy, line 28: I think this is the vibe!
```suggestion
  await db.schema
    .createIndex('quote_subject_cursor_idx')
    .on('quote')
    .columns(['subject', 'sortAt', 'cid'])
    .execute()
```

packages/bsky/src/data-plane/server/indexing/plugins/post.ts
haileyok: Oh, interesting! That makes sense. What is the way to check this then? Checking the `uri`?

haileyok: Think I see how we did it elsewhere!

packages/bsky/src/data-plane/server/indexing/plugins/post.ts
```diff
@@ -304,6 +339,7 @@ const deleteFn = async (
       .executeTakeFirst(),
     db.deleteFrom('feed_item').where('postUri', '=', uriStr).executeTakeFirst(),
   ])
+  await db.deleteFrom('quote').where('subject', '=', uriStr).execute()
```
haileyok, line 342: I think this is the right place to stick this, deleting all the quotes for the post once we delete it.

haileyok:
Having problems getting `indexing.test.ts` to pass, maybe someone has a hint as to why :(

seboslaw:
@haileyok  I'm wondering (and I'm probably not seeing it in the code) but shouldn't there also be an optional quoteID in the viewer? That way we can display that the user has already quoted a post.

haileyok:
@seboslaw I'm thinking that we shouldn't care about the user has already quoted a post or not? If the user makes more than one quote each should show up both in listing and in the quote count. From what I could tell this is also how Twitter works (though maybe there's more to it?)

seboslaw:
@haileyok I just checked. You're right. Twitter also doesn't show it (which I thought it did) and since you can quote multiple times - like you said - it makes no sense to indicate a 'made quote'. 👍

packages/bsky/src/data-plane/server/indexing/plugins/post.ts
devinivy: There's another one of these below.  I think we can nix this one.
```suggestion
```

packages/bsky/tests/views/quotes.test.ts
```diff
@@ -0,0 +1,105 @@
...
+      },
+    )
+
+    expect(alicePostQuotes.data.posts.length).toBe(1)
+    expect(forSnapshot(alicePostQuotes.data)).toMatchSnapshot()
+  })
+
+  it('does not return any quotes when the quoted post is deleted', async () => {
+    await sc.deletePost(alice, sc.posts[alice][0].ref.uri)
+    await network.processAll(1000)
```
devinivy, line 81: Is the timeout required, or leftover from debugging?  If there's a known reason it's required may just be nice to leave a comment.

packages/dev-env/src/seed/quotes.ts
```diff
@@ -0,0 +1,50 @@
...
+      sc.post(
+        sc.dids.eve,
+        `MASSIVE QUOTE SPAM ${i + 1}`,
+        undefined,
+        undefined,
+        sc.posts[sc.dids.alice][1].ref,
+      ),
+    )
+  }
+  await Promise.all(spamPosts)
```
devinivy, line 47: Is the "quote spam" just used to test pagination?  If so you may be able to use smaller pages, as we do some other places.  One upside is that it makes the test suites more reliable and quicker.  Doing all the writes in parallel also _could_ cause one of them to fail, as that is how the PDS bails if it struggles to serialize the writes to a given repo.

packages/bsky/tests/views/quotes.test.ts
```diff
@@ -0,0 +1,105 @@
...
+        headers: await network.serviceHeaders(alice, ids.AppBskyFeedGetQuotes),
+      },
+    )
+
+    expect(alicePostQuotesAfter.data.posts.length).toBe(0)
+  })
+
+  it('decrements quote count when a quote is deleted', async () => {
+    await sc.deletePost(eve, sc.posts[eve][2].ref.uri)
+    await network.processAll(1000)
```
devinivy, line 95: Another one here to peek at.

devinivy: Oops just notes there are some others!  Wont comment on them all.

packages/bsky/src/data-plane/server/indexing/plugins/post.ts
haileyok: okay looks like w/o this the test fails. let me actually pull down the code and look while i fix up that test.

packages/bsky/tests/views/quotes.test.ts
haileyok: probs copy/paste 👍

packages/bsky/src/data-plane/server/indexing/plugins/post.ts
devinivy: Ohh you know, I think I had it wrong.  The quote gets deleted if either the post or the quoted post get deleted.

haileyok: ah yep. added this line back now.

packages/dev-env/src/seed/quotes.ts
haileyok: ha yea didnt realize the writes were made one by one till yesterday 😅

devinivy (approved):

PR #2660, opened 2024-07-24
Use default Statsig export

estrattonbailey:
We were getting errors like `Cannot read properties of undefined (reading 'initialize')`. I was able to repro this locally using `localMode` settings. When using the default export, the error went away.


dholms (approved):
looks good 👍

PR #2662, opened 2024-07-25
Filter out relativeToDid target from suggestions results

estrattonbailey:


estrattonbailey:
We're gonna do this deeper in the stack

PR #2664, opened 2024-07-25
[Detach] Lexicons

estrattonbailey:
Adds a `app.bsky.feed.detach` lexicon and codegens. `targets` limit set to 1k for now.

### DIFF
--- a/lexicons/app/bsky/feed/detach.json
+++ b/lexicons/app/bsky/feed/detach.json
@@ -1,0 +1,26 @@
+{
+  "lexicon": 1,
+  "id": "app.bsky.feed.detach",
+  "defs": {
+    "main": {
+      "type": "record",
+      "key": "tid",
+      "description": "Record defining post URIs detached from a root post. The record key (rkey) of the detach record must match the record key of the root post in question, and that record must be in the same repository.",
+      "record": {
+        "type": "object",
+        "required": ["targets", "updatedAt"],
+        "properties": {
+          "targets": {
+            "type": "array",
+            "items": {
+              "type": "string",
+              "format": "at-uri"
+            },
+            "description": "List of detached post URIs."
+          },
+          "updatedAt": { "type": "string", "format": "datetime" }
+        }
+      }
+    }
+  }
+}
--- a/lexicons/app/bsky/feed/detach.json
+++ b/lexicons/app/bsky/feed/detach.json
@@ -12,6 +12,7 @@
         "properties": {
           "targets": {
             "type": "array",
+            "maxLength": 1000,
             "items": {
               "type": "string",
               "format": "at-uri"
--- a/lexicons/app/bsky/feed/detach.json
+++ b/lexicons/app/bsky/feed/detach.json
@@ -8,11 +8,16 @@
       "description": "Record defining post URIs detached from a root post. The record key (rkey) of the detach record must match the record key of the root post in question, and that record must be in the same repository.",
       "record": {
         "type": "object",
-        "required": ["targets", "updatedAt"],
+        "required": ["post", "targets", "updatedAt"],
         "properties": {
+          "post": {
+            "type": "string",
+            "format": "at-uri",
+            "description": "Reference (AT-URI) to the post record."
+          },
           "targets": {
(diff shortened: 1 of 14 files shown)


lexicons/app/bsky/feed/detach.json
```diff
@@ -0,0 +1,27 @@
...
+      "type": "record",
+      "key": "tid",
+      "description": "Record defining post URIs detached from a root post. The record key (rkey) of the detach record must match the record key of the root post in question, and that record must be in the same repository.",
+      "record": {
+        "type": "object",
+        "required": ["targets", "updatedAt"],
+        "properties": {
+          "targets": {
+            "type": "array",
+            "maxLength": 1000,
```
dholms, line 15: I'm inclined to start off with a lower maxLength. I think Paul threw out 50, but maybe 100 if we want some extra leeway?

lexicons/app/bsky/feed/detach.json
```diff
@@ -0,0 +1,27 @@
...
+  "id": "app.bsky.feed.detach",
+  "defs": {
+    "main": {
+      "type": "record",
+      "key": "tid",
+      "description": "Record defining post URIs detached from a root post. The record key (rkey) of the detach record must match the record key of the root post in question, and that record must be in the same repository.",
+      "record": {
+        "type": "object",
+        "required": ["targets", "updatedAt"],
+        "properties": {
```
dholms, line 12: Realized that the `threadgate` record has a reference to the post it's for.

It's deterministic and not really needed right now, but I think we probably want to add it here for:
- symmetry with threadgate records
- in case detach records can ever apply to another type of record

dholms:
Snuck in some protos here too 👍

lexicons/app/bsky/feed/detach.json
```diff
@@ -0,0 +1,32 @@
+{
+  "lexicon": 1,
+  "id": "app.bsky.feed.detach",
+  "defs": {
+    "main": {
+      "type": "record",
+      "key": "tid",
+      "description": "Record defining post URIs detached from a root post. The record key (rkey) of the detach record must match the record key of the root post in question, and that record must be in the same repository.",
```
bnewbold, line 8: "detaching" works against both replies and quote-posts, right? either way we should probably clarify here in the description was "detatch" actually means.

lexicons/app/bsky/feed/detach.json
```diff
@@ -0,0 +1,32 @@
...
+      "record": {
+        "type": "object",
+        "required": ["post", "targets", "updatedAt"],
+        "properties": {
+          "post": {
+            "type": "string",
+            "format": "at-uri",
+            "description": "Reference (AT-URI) to the post record."
```
bnewbold, line 16: "Reference (AT-URI) to the original post record"

lexicons/app/bsky/feed/detach.json
```diff
@@ -0,0 +1,32 @@
...
+          "targets": {
+            "type": "array",
+            "maxLength": 50,
+            "items": {
+              "type": "string",
+              "format": "at-uri"
+            },
+            "description": "List of detached post URIs."
+          },
+          "updatedAt": { "type": "string", "format": "datetime" }
```
bnewbold, line 27: I assume this should get updated every time posts are added or removed from the `targets` array? should clarify in description.

lexicons/app/bsky/feed/detach.json
```diff
@@ -0,0 +1,32 @@
...
+        "required": ["post", "targets", "updatedAt"],
+        "properties": {
+          "post": {
+            "type": "string",
+            "format": "at-uri",
+            "description": "Reference (AT-URI) to the post record."
+          },
+          "targets": {
+            "type": "array",
+            "maxLength": 50,
```
bnewbold, line 20: you said "1k" in the PR description, but only 50 here; was that intentional? either is probably ok/plausible to me, if I was painting the bike shed i'd probably go with 200

bnewbold (commented):
added some comments after the fact! can do a follow-up PR based on responses

PR #2668, opened 2024-07-29
Service auth method binding - PDS

dholms:
Just the PDS implementation of https://github.com/bluesky-social/atproto/pull/2663

This allows us to deploy our PDSs & publish a PDS distribution so that PDSs start emitting method-bound service auth tokens. Once the changes have had time to roll out, we can then deploy https://github.com/bluesky-social/atproto/pull/2663 more broadly to start checking for methods on service auth tokens.

This PR does not lift the restriction that an app password is unable to call `server.getServiceAuth`. This prevents a capability escalation attack where an app-password account could request a scoped service auth token that a sensitive service (such as chat) may not properly check.


.changeset/selfish-emus-wink.md
```diff
@@ -0,0 +1,6 @@
+---
+"@atproto/xrpc-server": minor
```
devinivy, line 2: Just confirming— this is a breaking change in xrpc-server?

packages/pds/src/api/com/atproto/server/getServiceAuth.ts
```diff
@@ -1,16 +1,34 @@
...
+      if (exp) {
+        const diff = exp - Date.now()
+        if (diff < 0) {
+          throw new InvalidRequestError(
+            'expiration is in past',
+            'BadExpiration',
+          )
+        } else if (diff > MINUTE) {
+          throw new InvalidRequestError(
+            'cannot request a token with an expiration more than an hour in the future',
```
devinivy, line 22: I expect the error text here is right and the check needs to be corrected:
```suggestion
        } else if (diff > HOUR) {
          throw new InvalidRequestError(
            'cannot request a token with an expiration more than an hour in the future',
```

packages/pds/src/api/com/atproto/server/getServiceAuth.ts
```diff
@@ -1,16 +1,34 @@
-import { createServiceJwt } from '@atproto/xrpc-server'
+import { InvalidRequestError, createServiceJwt } from '@atproto/xrpc-server'
 import AppContext from '../../../../context'
 import { Server } from '../../../../lexicon'
+import { MINUTE } from '@atproto/common'
```
devinivy, line 4:
```suggestion
import { HOUR } from '@atproto/common'
```

packages/pds/src/api/com/atproto/server/getServiceAuth.ts
```diff
@@ -1,16 +1,34 @@
...
+          )
+        }
+      }
+
       const token = await createServiceJwt({
         iss: did,
         aud: params.aud,
+        lxm: params.lxm ?? null,
         keypair,
       })
```
devinivy, line 33: Does the expiration need to get passed through here?

packages/pds/src/api/com/atproto/server/getServiceAuth.ts
```diff
@@ -1,16 +1,34 @@
...
       const did = auth.credentials.did
       const keypair = await ctx.actorStore.keypair(did)
+      const exp = params.exp ? params.exp * 1000 : undefined
+      if (exp) {
+        const diff = exp - Date.now()
+        if (diff < 0) {
+          throw new InvalidRequestError(
+            'expiration is in past',
+            'BadExpiration',
+          )
```
devinivy, line 19: One trick about using `exp` versus something like a duration is that it gets another clock in the mix (our server plus the requester), plus whatever latency between the servers.  I do like that it's a passthrough to the jwt claim, though.  Not strong feels/suggestions here, just an observation.

packages/xrpc-server/src/auth.ts
```diff
@@ -1,33 +1,45 @@
...
 }
 
-type ServiceJwtParams = ServiceJwtPayload & {
-  keypair: crypto.Keypair
+type ServiceJwtPayload = {
+  iss: string
+  aud: string
+  exp: number
+  lxm?: string
+  nonce?: string
```
devinivy, line 22: We talked about potentially replacing this with the standard `jti` claim.

packages/xrpc-server/src/util.ts
```diff
@@ -306,3 +306,7 @@ export interface ServerTiming {
   duration?: number
   description?: string
 }
+
+export const parseReqNsid = (req: express.Request): string => {
+  return req.originalUrl.split('?')[0].replace('/xrpc/', '')
```
devinivy, line 311: I wonder if a trailing slash might throw this off.

packages/pds/src/pipethrough.ts
```diff
@@ -14,14 +14,24 @@ import { ids, lexicons } from './lexicon/lexicons'
...
 
 export const proxyHandler = (ctx: AppContext): CatchallHandler => {
   const accessStandard = ctx.authVerifier.accessStandard()
   return async (req, res, next) => {
     try {
-      const { url, aud } = await formatUrlAndAud(ctx, req)
+      const { url, aud, nsid } = await formatUrlAndAud(ctx, req)
       const auth = await accessStandard({ req })
-      const headers = await formatHeaders(ctx, req, aud, auth.credentials.did)
+      if (!auth.credentials.isPrivileged && PRIVILEGED_METHODS.includes(nsid)) {
```
devinivy, line 25: Could turn `PRIVILEGED_METHODS` into a set for the micro-optimization 😆

packages/pds/src/pipethrough.ts
```diff
@@ -89,17 +102,18 @@ export const formatUrlAndAud = async (
...
-  return { url, aud }
+  return { url, aud, nsid }
 }
 
 export const formatHeaders = async (
   ctx: AppContext,
   req: express.Request,
   aud: string,
+  lxm: string,
   requester: string | null,
```
devinivy, line 113: Not necessarily a change we need to make now, but some of these functions have enough positional `string` args next to each other that it may be worth moving them to an object to avoid mixups.

.changeset/selfish-emus-wink.md
devinivy: Looks like it is 👍

devinivy (approved):
Tossed in a few ideas, but this is looking great!

PR #2674, opened 2024-08-01
Run test env with `updateSnapshot`

haileyok:
Small thing that bothers me every time I need to do this, would be nice for the `updateSnapshot` script to run with the test env, unless there's a reason for us not to.


dholms (approved):
yup looks good 👍

PR #2675, opened 2024-08-02
Detached QPs and hidden replies

estrattonbailey:
- [ ] merge main before merge, ensure uri utils are all replaced


lexicons/app/bsky/feed/postgate.json
```diff
@@ -0,0 +1,32 @@
...
+          },
+          "detachedQuotes": {
+            "type": "array",
+            "maxLength": 50,
+            "items": {
+              "type": "string",
+              "format": "at-uri"
+            },
+            "description": "List of detached quote post URIs."
+          }
```
brianolson, line 27:
```suggestion
          },
          "rules": {
            "type": "array",
            "maxLength": 5,
            "items": {
              "type": "union",
              "refs": ["#noQP", "#followingRule"]
            }
          }
```

lexicons/app/bsky/feed/postgate.json
```diff
@@ -0,0 +1,32 @@
...
+        }
+      }
+    }
```
brianolson, line 30:
```suggestion
    },
    "noQP": {
      "type": "object",
      "description": "No new quote-posts allowed",
      "properties": {}
    },
    "followingQP": {
      "type": "object",
      "description": "Allow quote-post from actors you follow",
      "properties": {}
    }
```

packages/bsky/src/api/app/bsky/feed/getTimeline.ts
```diff
@@ -102,9 +106,14 @@ const presentation = (inputs: {
   hydration: HydrationState
 }) => {
   const { ctx, skeleton, hydration } = inputs
-  const feed = mapDefined(skeleton.items, (item) =>
-    ctx.views.feedViewPost(item, hydration),
-  )
+  const feed = mapDefined(skeleton.items, (item) => {
+    const post = ctx.views.feedViewPost(item, hydration)
+    if (!post) return
+    if (isFeedPostHiddenReply({ post, hydration })) {
```
dholms, line 112: i think this would be comfier in the "rules" part of the pipeline

(arguably the rules & presentation parts could be collapsed into a single step, but since they're separate for now I think we move it there)

packages/bsky/src/api/util.ts
```diff
@@ -25,3 +29,25 @@ export const clearlyBadCursor = (cursor?: string) => {
   // hallmark of v1 cursor, highly unlikely in v2 cursors based on time or rkeys
   return !!cursor?.includes('::')
 }
+
+export function isFeedPostHiddenReply({
```
dholms, line 33: kinda wanna move this onto the `Views` class next to other checks like `viewerBlockExists` and `actorIsTakendown`

packages/bsky/src/hydration/feed.ts
```diff
@@ -203,6 +206,57 @@ export class FeedHydrator {
...
+    if (!postUris.length) return new HydrationMap<PostGate>()
+    const uris = postUris.map((uri) => {
+      const parsed = new AtUri(uri)
+      return AtUri.make(
+        parsed.hostname,
+        ids.AppBskyFeedPostgate,
+        parsed.rkey,
+      ).toString()
+    })
+    const res = await this.dataplane.getPostGateRecords({ uris })
```
dholms, line 236: looks like this is the same logic as `getPostGateRecords` maybe we just call that method?

dholms: same for `getThreadGateRecords` above 👍

packages/bsky/src/hydration/hydrator.ts
```diff
@@ -409,6 +412,8 @@ export class Hydrator {
       this.hydrateFeedGens(nestedFeedGenUris, ctx),
       this.hydrateLabelers(nestedLabelerDids, ctx),
       this.hydrateStarterPacksBasic(nestedStarterPackUris, ctx),
+      // only gates for quoted posts
+      this.feed.getPostGatesForPosts(postUrisLayer1),
```
dholms, line 416: this var could use a better name, had to really look through the code to figure out what it was. it only includes quoted posts right? maybe something like `embedPostUrisLayer1`?

packages/bsky/src/views/index.ts
```diff
@@ -608,13 +609,14 @@ export class Views {
     }
     const post = this.post(item.post.uri, state)
     if (!post) return
-    return {
+    const view = {
```
dholms, line 612: any reason for making this a var before returning? (guessing this is just cruft from rewriting something)

packages/bsky/tests/feed-generation.test.ts
```diff
@@ -582,6 +583,116 @@ describe('feed generation', () => {
       )
     })
 
+    describe('postgates', () => {
```
dholms, line 586: thoughts on moving this into it's own test suite next to `threadgating`?

this test suite is really about appview integration with feed generators & while that does intersect to some degree with postgates, I think we might want to have a separate test suite for it as a concept

dholms: oh wait i shoulda read further. i thought there were more tests here. think this is cozy here 👍

packages/bsky/src/api/util.ts
estrattonbailey: Ah yes that's much better!

packages/bsky/tests/views/thread.test.ts
```diff
@@ -368,4 +371,86 @@ describe('pds thread views', () => {
       )
     })
   })
+
+  describe('postgates', () => {
```
dholms, line 375: think this might be fit better in `posts.test.ts` instead 🤔 

tho idk I don't hate having the test in the context of `getPostThread` as well, but I am kinda itching for at least one test on the simplest posts route which is `getPosts`

packages/bsky/src/api/app/bsky/feed/getTimeline.ts
estrattonbailey: Moved to View

packages/bsky/src/hydration/feed.ts
estrattonbailey: Ah yeah good call

packages/bsky/src/hydration/hydrator.ts
```diff
@@ -91,6 +92,7 @@ export type HydrationState = {
   follows?: Follows
   followBlocks?: FollowBlocks
   threadgates?: Threadgates
+  postGates?: PostGates
```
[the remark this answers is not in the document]
estrattonbailey: Yeah threadgates is inconsistently named so I opted for camel for postGates throughout. But you're right that the generated types are not camel + now that I'm implementing I realize that our client uses lowercase. I'll update!

packages/bsky/tests/feed-generation.test.ts
estrattonbailey: I need to add some more so I think it's a good idea to split this out before this PR is said and done 👍

lexicons/app/bsky/feed/postgate.json
estrattonbailey: Thanks for this. I think we probably want to keep parity with threadgates in terms of naming and features. I'll push something up and we can discuss in Slack 👍

mackuba:
Hmm, so does this still return all entries from the AppView as before, like in the description, or has something changed?…

devinivy (commented):
Phewwie, what a solid read!  This was a ton of work, I think we're just about in good shape 🙏

packages/bsky/src/data-plane/server/db/migrations/20240808T224251220Z-post-violates-quotegate.ts
```diff
@@ -0,0 +1,12 @@
+import { Kysely } from 'kysely'
+
+export async function up(db: Kysely<unknown>): Promise<void> {
+  await db.schema
+    .alterTable('post')
+    .addColumn('violatesQuoteGate', 'boolean')
```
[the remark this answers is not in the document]
estrattonbailey: Ah yeah noticed yesterday and fixed. Same for other locations. This was a remnant that I had missed on the last rename go-around.

packages/bsky/src/views/index.ts
```diff
@@ -1101,6 +1202,14 @@ const postToGateUri = (uri: string) => {
   return aturi.toString()
 }
 
+const postToPostgateUri = (uri: string) => {
+  const aturi = new AtUri(uri)
+  if (aturi.collection === ids.AppBskyFeedPost) {
+    aturi.collection = ids.AppBskyFeedPostgate
+  }
+  return aturi.toString()
+}
```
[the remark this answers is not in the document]
estrattonbailey: Assert is probably a good idea here, but yeah if the check fails for some reason, the postgates will come back `null` and features will fail silently. Not a bad fallback, but also difficult to debug.

I think we can probably add the assert, lmk if you feel differently 👍

packages/bsky/src/hydration/hydrator.ts
```diff
@@ -517,18 +528,21 @@ export class Hydrator {
...
+      await Promise.all([
+        this.hydratePosts(postAndReplyRefs, ctx, {
+          posts: posts.merge(replies), // avoids refetches of posts
+        }),
+        this.hydrateProfiles(
+          [...repostUris.map(didFromUri), ...replyParentAuthors],
+          ctx,
+        ),
+        this.feed.getReposts(repostUris, ctx.includeTakedowns),
+        this.feed.getThreadgatesForPosts(rootUris),
```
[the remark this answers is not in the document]
estrattonbailey: Ah you know what, looking at this made me realize I was making redundant calls here. We already were hydrating all threadgates for root posts within `hydratePosts` since it's needed to power `viewer.replyDisabled`. So we can remove this line entirely.

packages/bsky/src/views/index.ts
```diff
@@ -1037,6 +1102,33 @@ export class Views {
...
+    const viewerDid = state.ctx?.viewer
+    if (!gate || !viewerDid) {
+      return false
+    }
+    const {
+      embeddingRules: { canQuotepost },
+    } = parsePostgate({
+      gate,
+      viewerDid,
+      authorDid: new AtUri(uri).hostname,
```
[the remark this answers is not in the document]
estrattonbailey: https://github.com/bluesky-social/atproto/pull/2675/commits/a1ba63e96d10d70a5059210c4faa75cfddc135b9

packages/bsky/src/hydration/hydrator.ts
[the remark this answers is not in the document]
estrattonbailey: https://github.com/bluesky-social/atproto/pull/2675/commits/2447fa5ba9bd827c5097d7c73bdea94d7ba34a0e

packages/bsky/src/hydration/hydrator.ts
```diff
@@ -713,19 +727,27 @@ export class Hydrator {
     ctx: HydrateCtx,
   ): Promise<HydrationState> {
     const uris = notifs.map((notif) => notif.uri)
+    const replyRootPostUris = notifs
+      .filter((notif) => {
+        return notif.reason === 'reply'
+      })
+      .map((notif) => {
+        return notif.reasonSubject
+      })
```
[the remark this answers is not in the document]
estrattonbailey: Whatcha think of this? https://github.com/bluesky-social/atproto/pull/2675/commits/4740a9de067e5e6b9945d07f46e4b6803c510d08

packages/bsky/src/views/index.ts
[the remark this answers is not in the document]
estrattonbailey: https://github.com/bluesky-social/atproto/pull/2675/commits/829bb2983835f6b3b4593ffb2e12378c01206179

packages/bsky/src/hydration/feed.ts
```diff
@@ -203,6 +220,36 @@ export class FeedHydrator {
...
+    includeTakedowns = false,
+  ): Promise<Postgates> {
+    if (!postUris.length) return new HydrationMap<Postgate>()
+    const uris = postUris.map((uri) => {
+      const parsed = new AtUri(uri)
+      return AtUri.make(
+        parsed.hostname,
+        ids.AppBskyFeedPostgate,
+        parsed.rkey,
+      ).toString()
```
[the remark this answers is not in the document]
estrattonbailey: Made a new util file for this and `uriToDid` https://github.com/bluesky-social/atproto/pull/2675/commits/678bd53c0f2304bbc19dbf094e70e8e38be37656

packages/bsky/src/views/index.ts
```diff
@@ -1037,6 +1102,33 @@ export class Views {
...
+    const post = state.posts?.get(uri)
+    if (!post) {
+      return true
+    }
+    const postgateRecordUri = postToPostgateUri(uri)
+    const gate = state.postgates?.get(postgateRecordUri)?.record
+    const viewerDid = state.ctx?.viewer
+    if (!gate || !viewerDid) {
+      return false
+    }
```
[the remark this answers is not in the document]
estrattonbailey: Good call, that's nice

packages/bsky/src/data-plane/server/indexing/plugins/post-gate.ts
```diff
@@ -0,0 +1,92 @@
...
+    .deleteFrom('post_gate')
+    .where('uri', '=', uri.toString())
+    .returningAll()
+    .executeTakeFirst()
+  return deleted || null
+}
+
+const notifsForDelete = () => {
+  return { notifs: [], toDelete: [] }
+}
```
estrattonbailey, line 74: Note: hidden replies should yank out notifications

packages/bsky/src/views/index.ts
```diff
@@ -608,12 +621,35 @@ export class Views {
...
+          return undefined
+        } else if (isPostView(reply.parent)) {
+          const parentPostIsHiddenReply = this.replyIsHidden(
+            reply.parent.uri,
+            reply.root.uri,
+            state,
+          )
+          if (parentPostIsHiddenReply) {
+            reply.parent = this.notFoundPost(reply.parent.uri)
+          }
```
estrattonbailey, line 645: Check this after blocks get in, may want to overwrite root too

packages/bsky/src/data-plane/server/indexing/plugins/post.ts
```diff
@@ -265,6 +295,8 @@ const notifsForInsert = (obj: IndexedPost) => {
         recordCid: obj.post.cid,
         sortAt: obj.post.sortAt,
       })
+      // found hidden reply, don't notify any higher ancestors
+      if (threadgateHiddenReplies.includes(ancestorUri.toString())) break
```
estrattonbailey, line 299: A little uncertain of how this plays with dev env

packages/bsky/src/views/index.ts
```diff
@@ -1050,6 +1130,16 @@ export class Views {
...
+      // filter hidden replies
+      if (recordInfo && notif.reason === 'reply') {
+        const rootPostUri = isPostRecord(recordInfo.record)
+          ? recordInfo.record.reply?.root.uri
+          : undefined
+        const isHiddenReply = rootPostUri
+          ? this.replyIsHidden(notif.uri, rootPostUri, state)
+          : false
+        if (isHiddenReply) return
+      }
```
estrattonbailey, line 1142: Filters out hidden replies for _everyone's_ notifications. Replies to a hidden reply are not filtered out.

estrattonbailey: Removed this, will not filter from notifications feed for now, and if we do, we'll do it at the dataplane

packages/bsky/src/views/index.ts
```diff
@@ -584,18 +590,22 @@ export class Views {
...
         : undefined,
     }
   }
 
   feedViewPost(
     item: FeedItem,
     state: HydrationState,
+    opts?: {
+      includeHiddenReplies?: boolean
+    },
```
estrattonbailey, line 608: This OK? I want to maybe drill this into `replyRef` too, since it'd be nice to group the reply logic

.changeset/perfect-parrots-appear.md
```diff
@@ -0,0 +1,8 @@
+---
+"@atproto/ozone": patch
+"@atproto/bsky": patch
+"@atproto/api": patch
+"@atproto/pds": patch
+---
+
+Adds `postgate` records to power quote gating and detached quote posts, plus `hiddenReplies` to the `threadgate` record.
```
devinivy, line 8: I think we can remove ozone and pds here since the change is incidental to them (e.g. don't need this line in their changelog).  They'll still get bumped by virtue of the update to their `@atproto/api` version.
```suggestion
---
"@atproto/bsky": patch
"@atproto/api": patch
---

Adds `postgate` records to power quote gating and detached quote posts, plus `hiddenReplies` to the `threadgate` record.
```

devinivy, line 8: I think we can remove ozone from here since the change is incidental to ozone (e.g. doesn't need this line in its changelog).  It'll still get bumped by virtue of the update to its `@atproto/api` version.
```suggestion
---
"@atproto/bsky": patch
"@atproto/api": patch
"@atproto/pds": patch
---

Adds `postgate` records to power quote gating and detached quote posts, plus `hiddenReplies` to the `threadgate` record.
```

packages/bsky/tests/seed/postgates.ts
```diff
@@ -0,0 +1,56 @@
...
+) {
+  const u = Object.assign({}, users)
+
+  await sc.createAccount('quotee', u.quotee)
+  await sc.createAccount('quoter', u.quoter)
+  await sc.createAccount('viewer', u.viewer)
+
+  Object.values(u).forEach((user) => {
+    u[user.id].did = sc.dids[user.id]
+  })
```
devinivy, line 48: I'm sorry 😭 this is still mutating `users`.  I do want to address this though, because it's going to be very confusing if it ever bites.
```suggestion
  const u = structuredClone(users)

  await sc.createAccount('quotee', u.quotee)
  await sc.createAccount('quoter', u.quoter)
  await sc.createAccount('viewer', u.viewer)

  Object.values(u).forEach((user) => {
    u[user.id].did = sc.dids[user.id]
  })
```

devinivy: Alternately above we can do something like:
```ts
const createUsers = () => ({
  viewer: createUser('viewer'),
  quotee: createUser('quotee'),
  quoter: createUser('quoter'),
})

Comment (estrattonbailey) [reply packages/bsky/tests/seed/postgates.ts:48]: 🤦

Comment (estrattonbailey) [reply packages/bsky/tests/seed/postgates.ts]: structuredClone is clean

File: packages/bsky/tests/seed/feed-hidden-replies.ts [hunk 25]
```diff
@@ -0,0 +1,62 @@
...
+  replier: createUser('replier'),
+  reposter: createUser('reposter'),
+}
+
+export type Users = typeof users
+
+export async function feedHiddenRepliesSeed(
+  sc: SeedClient<TestNetwork | TestNetworkNoAppView>,
+) {
+  const u = Object.assign({}, users)
```
devinivy: Same here, re: `users` getting mutated below.

packages/bsky/src/views/util.ts
```diff
@@ -27,8 +26,8 @@ export const parseThreadGate = (
     return { canReply: true }
   }
 
-  const allowMentions = !!gate.allow.find(isMentionRule)
-  const allowFollowing = !!gate.allow.find(isFollowingRule)
+  const allowMentions = gate.allow.some(isMentionRule)
+  const allowFollowing = gate.allow.some(isFollowingRule)
```
devinivy, line 30: It's the little things 💎

packages/bsky/src/hydration/hydrator.ts
```diff
@@ -341,27 +343,40 @@ export class Hydrator {
...
     for (const [uri, post] of postsLayer0) {
       if (post) {
         threadRootUris.add(rootUriFromPost(post) ?? uri)
       }
     }
+    const postUrisWithThreadgates = new Set<string>()
+    for (const uri of threadRootUris) {
+      const post = postsLayer0.get(uri)
+      // post could be deleted, still want to check for existing threadgate
+      if (!post || (post && post.hasThreadGate)) {
```
devinivy, line 370:
```suggestion
      if (!post || post.hasThreadGate) {
```

packages/bsky/src/hydration/hydrator.ts
```diff
@@ -341,27 +343,40 @@ export class Hydrator {
...
     const threadRootUris = new Set<string>()
     for (const [uri, post] of postsLayer0) {
       if (post) {
         threadRootUris.add(rootUriFromPost(post) ?? uri)
       }
     }
+    const postUrisWithThreadgates = new Set<string>()
+    for (const uri of threadRootUris) {
+      const post = postsLayer0.get(uri)
+      // post could be deleted, still want to check for existing threadgate
```
devinivy, line 369: The post wouldn't have to be deleted to be missing here—I think the main reason this method needs to fetch threadgates in this case is because there's no guarantee the caller also asked to hydrate each post's root.

The way I think we can message around this (and I appreciate the comment, definitely needed here), is to talk about checking `post.hasThreadGate` as an optimization for a common case of hydrating feed items.  When we're hydrating feed items the roots _should_ be included.  This method `hydratePosts()` doesn't rely on it, but it is allows the optimization to take place when possible.

devinivy: Oh forgive me if I have this wrong!  It looks like we _do_ fetch roots even when we're just hydrating a single post.  Curious!

packages/bsky/src/hydration/hydrator.ts
```diff
@@ -383,13 +398,19 @@ export class Hydrator {
...
     const threadRefs = allRefs.map((ref) => ({
       ...ref,
       threadRoot: posts.get(ref.uri)?.record.reply?.root.uri ?? ref.uri,
     }))
+    const postUrisWithPostgates = new Set<string>()
+    for (const [uri, post] of posts) {
+      if (post && post.hasPostGate) {
+        postUrisWithPostgates.add(uri)
+      }
+    }
```
devinivy, line 413: If this is too complicated feel free to ignore, but I _think_ we only need postgates on nested posts, not necessarily for e.g. `postsLayer0`.  That's going to be a lot fewer to check since most posts don't have a post embed.  I believe that would be `[...postUrisLayer1, ...postUrisLayer2]`.

packages/bsky/src/hydration/hydrator.ts
estrattonbailey: Heh yep yep, always a thing now

packages/bsky/src/hydration/hydrator.ts
estrattonbailey: Postgates also power `ViewerState.embeddingDisabled`, so I think we do unfortunately need it for all posts, even top level :/

devinivy: Makes sense!

packages/bsky/src/hydration/hydrator.ts
estrattonbailey: How's this btw https://github.com/bluesky-social/atproto/pull/2675/commits/3aa36efc345c657aabc26502c1de6ee93993e612

packages/bsky/src/data-plane/server/indexing/plugins/post.ts
```diff
@@ -444,10 +479,59 @@ async function validateReply(
...
+  embedUri: string,
+  parentUri: string,
+) {
+  const postgateRecordUri = postUriToPostgateUri(embedUri)
+  const results = await db
+    .selectFrom('record')
+    .where('record.uri', '=', postgateRecordUri)
+    .selectAll()
+    .execute()
+  const postgateRecord = results.find((ref) => ref.uri === postgateRecordUri)
```
devinivy, line 508: Does this work just as well?
```suggestion
  const postgateRecord = await db
    .selectFrom('record')
    .where('record.uri', '=', postgateRecordUri)
    .selectAll()
    .executeTakeFirst()
```

estrattonbailey: oh duh, yeah lgtm

packages/bsky/src/hydration/hydrator.ts
devinivy: Perf 👍

packages/bsky/src/data-plane/server/indexing/plugins/post.ts
estrattonbailey: tested locally quick, committing

estrattonbailey:
Thanks for sticking with me on this one @devinivy been a trip!

packages/bsky/src/data-plane/server/indexing/plugins/post.ts
```diff
@@ -218,7 +218,7 @@ const insertFn = async (
 
         const { violatesEmbeddingRules } = await validatePostEmbed(
           db,
-          record.uri,
+          embedUri.toString(),
```
estrattonbailey, line 221: Using `embedUri` since we already checked if it was a post above.

packages/bsky/src/views/index.ts
```diff
@@ -979,8 +982,10 @@ export class Views {
     if (parsedUri.collection === ids.AppBskyFeedPost) {
       const view = this.embedPostView(uri, state, depth)
       if (!view) return this.embedNotFound(uri)
-      const postgateRecordUri = postUriToPostgateUri(uri)
-      const postgate = state.postgates?.get(postgateRecordUri)
+      const postgateRecordUri = postUriToPostgateUri(parsedUri.toString())
```
estrattonbailey, line 985: Using `parseUri` since we checked if it was a post above.

devinivy (approved):

PR #2676, opened 2024-08-02
Remove detach

estrattonbailey:
Removes the new record added in #2664. We revised the approach mid-implementation, which is now taking place in #2675.


dholms (approved):
yup yup 👍

PR #2680, opened 2024-08-02
Use corepack to manage the pnpm version

mozzius:
Two very small changes:

1. Adds an `.nvmrc` file. This allows you to use `nvm use` without specifying a version
2. Specifies the pnpm version to use via `package.json`'s `packageManager` field. This lets corepack know which version of pnpm to use. Also changes the makefile to enable corepack rather than install pnpm via npm

Reason for change - they changes the lockfile version in v9 so every time I run `pnpm i` it tries to modify the lockfile.


devinivy (approved):
Nice, thanks!

.github/workflows/repo.yaml
```diff
@@ -63,7 +61,6 @@ jobs:
       - uses: actions/checkout@v4
       - uses: pnpm/action-setup@v4
         with:
-          version: 8
```
devinivy, line 66: Is the idea that this isn't needed any longer since the action pays attention to `packageManager` within the package.json?

mozzius: Yep, it complains if it's present in both places

PR #2682, opened 2024-08-02
Fix run-dev-env-logged command

estrattonbailey:
Noticed logging wasn't actually working.

Another suggestion: add `| tee logs.txt` so they're easier to browse. Idk about y'all but my tmux setup only persists like 2k lines lol.


dholms (approved):
looks good 👍 

I'd leave it as is where it prints to the console, but if you set `LOG_DESTINATION` before running the command you can send logs to a file

PR #2683, opened 2024-08-02
Add ref to `threadgate` record from `threadgateView`

estrattonbailey:


devinivy (commented):
This is far from the only place we use an `unknown` in this way— I'd propose we talk it through then address it more generally throughout the lexicons rather than make a one off.

Vincentroland (approved):

PR #2693, opened 2024-08-06
Add quote gate props to postgate

estrattonbailey:
Adds `quotePostRules[]` array to the `postgate` record. It re-uses the rulesets and naming conventions from `threadgate`. Absence of the `quotePostRules` array denotes the default state where everyone is allowed QP the post.

It also adds an additional rule called `disableRule` so that the disabled state is explicit, as opposed to inferred via an empty `allow` array on `threadgate` records. This makes it easier to maintain this record, now that they contain multiple properties managed separately.

### DIFF
--- a/lexicons/app/bsky/feed/postgate.json
+++ b/lexicons/app/bsky/feed/postgate.json
@@ -25,12 +25,17 @@
             },
             "description": "List of detached quote post URIs."
           },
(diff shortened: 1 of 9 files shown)


pfrazee (approved):
Ya lgtm

brianolson (approved):

PR #2696, opened 2024-08-07
Add autoplay to preferences

mozzius:
With video coming, we should migrate the "autoplay" local pref to the server

TO DECIDE: current behaviour defers the initial value to `prefers-reduced-motion` or equivalent. How should this factor into this?


mozzius:
Let's keep it local

PR #2697, opened 2024-08-08
Add quote gating to `postgate`

estrattonbailey:
### Stacked on #2675 

Adds a `quotepostRules` array to `postgate` record, which behaves essentially the same as `threadgate.allow[]`. The only rule we support right now is `disableRule`, which disables quote posts for the target post.

This PR largely follows threadgates example for implementation with the mock data plane and view handling.

I split out the test suite and created a separate seed, otherwise these gates will wreak havoc on other tests. I plan to do the same with the reply hiding, but I'll do that as a separate PR from the base branch.


packages/bsky/src/data-plane/server/db/tables/post.ts
```diff
@@ -15,6 +15,7 @@ export interface Post {
   tags: string[] | null
   invalidReplyRoot: boolean | null
   violatesThreadGate: boolean | null
+  violatesQuoteGate: boolean | null
```
estrattonbailey, line 18: Coming from the dataplane this will be `violatesQuoteGate`. Everywhere else I refer to it as `violatesQuotegate`.

packages/bsky/src/data-plane/server/util.ts
```diff
@@ -150,3 +150,9 @@ export const postToThreadgateUri = (postUri: string) => {
   gateUri.collection = ids.AppBskyFeedThreadgate
   return gateUri.toString()
 }
+
+export const postToPostgateUri = (postUri: string) => {
+  const gateUri = new AtUri(postUri)
+  gateUri.collection = ids.AppBskyFeedPostgate
+  return gateUri.toString()
+}
```
estrattonbailey, line 158: There was already some duplication here with views utils, might be able to clean up

packages/bsky/src/hydration/feed.ts
```diff
@@ -86,7 +89,11 @@ export class FeedHydrator {
     return need.reduce((acc, uri, i) => {
       const record = parseRecord<PostRecord>(res.records[i], includeTakedowns)
       const violatesThreadGate = res.meta[i].violatesThreadGate
-      return acc.set(uri, record ? { ...record, violatesThreadGate } : null)
+      const violatesQuotegate = res.meta[i].violatesQuoteGate
```
estrattonbailey, line 92: Coming from the dataplane this will be violatesQuoteGate, I think bc the buf gen converts snake to camel. We could probably also rename the proto to `violates_quotegate`, but figured might as well keep it consistent in the dataplane.

Everywhere else I refer to it as violatesQuotegate.

lexicons/app/bsky/feed/defs.json
```diff
@@ -41,7 +41,8 @@
         "repost": { "type": "string", "format": "at-uri" },
         "like": { "type": "string", "format": "at-uri" },
         "threadMuted": { "type": "boolean" },
-        "replyDisabled": { "type": "boolean" }
+        "replyDisabled": { "type": "boolean" },
+        "quotepostDisabled": { "type": "boolean" }
```
devinivy, line 45: In lexicon quote posts are expressed as "post record embeds."  So I'm wondering if `embedDisabled` would be more consistent here.

estrattonbailey:
Gonna merge into base and continue from there, easier to review and reason about

PR #2700, opened 2024-08-09
relax ozone label naming rule/convention, allow numbers.

### LINKED ISSUE #2699: relax ozone label naming rule/convention, allow numbers.

as described in #2698, it needs  easier-to-read label names for human.
number(0-9) is not allowed in current rule so it is too hard to read/distinguish for human.

for example, regulation uses numbers.
series(generation, season), model also often uses numbers or YEAR in YYYY

```
restricted-NN-years-old is not allowed then 
restricted-eighteen-years-old
restricted-thirteen-years-old

how about ISO with 5 digits?
twothousands-twentyfour  instead of 2024?
```

relax label naming rule in atproto/api.

**Describe alternatives you've considered**

no alternatives easier than the above.

**Additional context**

itaru2622:
fix  #2699

### DIFF
--- a/packages/api/src/moderation/types.ts
+++ b/packages/api/src/moderation/types.ts
@@ -10,7 +10,7 @@ import { KnownLabelValue } from './const/labels'
 // syntax
 // =
 
-export const CUSTOM_LABEL_VALUE_RE = /^[a-z-]+$/
+export const CUSTOM_LABEL_VALUE_RE = /^[a-z0-9-]+$/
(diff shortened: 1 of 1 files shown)


itaru2622:
@bnewbold could you review this PR, please.

bnewbold:
@itaru2622 I think we are not going to make this change in the near future

Vincentroland (approved):

PR #2701, opened 2024-08-09
Updating naming conventions for postgate-related attributes

estrattonbailey:


lexicons/app/bsky/feed/postgate.json
```diff
@@ -16,16 +16,16 @@
...
-          "detachedQuotes": {
+          "detachedEmbeds": {
             "type": "array",
             "maxLength": 50,
             "items": {
               "type": "string",
               "format": "at-uri"
             },
-            "description": "List of detached quote post URIs."
+            "description": "List of detached embed URIs."
```
brianolson, line 26: 'embedding' right? These are the URIs of posts that embed the post modified by this postgate

estrattonbailey: +1

brianolson (approved):

lexicons/app/bsky/feed/postgate.json
```diff
@@ -16,16 +16,16 @@
             "format": "at-uri",
             "description": "Reference (AT-URI) to the post record."
           },
-          "detachedQuotes": {
+          "detachedEmbeddingUris": {
```
devinivy, line 19: While we're in here I wonder if we may want to sync-up the terminology on embed views.  Currently there's a `#viewRemoved` lexicon that gets used within the posts listed on this field.  For consistency perhaps that should become `#viewDetached`.

devinivy (approved):
Left a final thought or two, but this is looking good! 👍

lexicons/app/bsky/feed/postgate.json
```diff
@@ -16,16 +16,16 @@
...
             "maxLength": 50,
             "items": {
               "type": "string",
               "format": "at-uri"
             },
-            "description": "List of detached quote post URIs."
+            "description": "List of AT-URIs embedding this post that the author has detached from."
           },
-          "quotepostRules": {
+          "embeddingRules": {
```
devinivy, line 28: On threadgate records the corresponding property is called `allow`, which contains the rules determining whether a given reply is allowed.

I am not opposed to `embeddingRules`, though if we want to be really consistent with threadgates it might become something like `allowEmbedding`.  I'll leave it up to you and @brianolson whether you want to adopt that change.

brianolson: a few people noted that an empty list of allow rules being implicitly deny is less clear than an explicit deny rule. it's also a little weird because nil-record is different than empty-record. so I'm in favor of the explicit deny rule. I _think_ this implicitly makes the logic that you can have deny rules _and_ allow rules, and I'd interpret an allow rule as superseding the deny rule. `{deny *, allow foo, allow bar}` _does_ allow foo and bar

brianolson (approved):

PR #2703, opened 2024-08-09
✨  Apply embed specific tag on subjects for video, image and external

foysalit:
This PR refactors the existing subject language tagger to build a more generic subject tagger that is capable or running multiple taggers that evaulates the record content from the subject and applies necessary tags.
The current implementation adds tag for embed content type along with existing language tags.


packages/ozone/src/tag-service/index.ts
```diff
@@ -0,0 +1,61 @@
...
+    try {
+      const tags: string[] = []
+
+      await Promise.all(
+        this.taggers.map(async (tagger) => {
+          try {
+            if (tagger.isApplicable()) {
+              const newTags = await tagger.getTags()
+              if (newTags.length) tags.push(...newTags)
+            }
```
devinivy, line 35: What is the intended behavior if you call `tagger.getTags()` without checking `tagger.isApplicable()`?  It seems like `isApplicable()` may not need to be part of the public interface— the implementation of `getTags()` could alternatively check this itself and no-op by returning `[]`.

packages/ozone/src/tag-service/embed-content-type-tagger.ts
```diff
@@ -0,0 +1,45 @@
...
+import { langLogger as log } from '../logger'
+import { ContentTagger } from './content-tagger'
+
+export class EmbedContentTypeTagger extends ContentTagger {
+  tagPrefix = 'embed-content-type:'
+  isApplicable(): boolean {
+    return Boolean(
+      this.subjectStatus &&
+        this.subject.isRecord() &&
+        this.subject.uri.includes('app.bsky.feed.post') &&
```
devinivy, line 11: It would be cozier if we either checked `includes('/app.bsky.feed.post/')`, or went ahead and parsed the URI and checked its collection.

```ts
const getCollection(uri: string) {
  return new AtUri(uri).collection
}
```

packages/ozone/src/tag-service/embed-content-type-tagger.ts
```diff
@@ -0,0 +1,45 @@
...
+        !this.subjectStatus.tags?.find((tag) => tag.includes(this.tagPrefix)),
```
devinivy, line 12: Just a little more specific!
```suggestion
        !this.subjectStatus.tags?.some((tag) => tag.startsWith(this.tagPrefix)),
```

packages/ozone/src/tag-service/embed-content-type-tagger.ts
```diff
@@ -0,0 +1,45 @@
...
+  async getTags(): Promise<string[]> {
+    try {
+      const recordValue = await this.getRecordValue()
+      if (!recordValue) {
+        return []
+      }
+      const tags: string[] = []
+      if (AppBskyEmbedImages.isView(recordValue)) {
+        tags.push(`${this.tagPrefix}image`)
+      }
```
devinivy, line 25: If `recordValue` is a record then we don't have our hands on a view (`isView()`).  I'd expect something roughly along these lines:
```suggestion
      if (
        AppBskyFeedPost.isRecord(recordValue) &&
        (AppBskyEmbedImages.isMain(recordValue.embed) ||
          AppBskyEmbedRecordWithMedia.isMain(recordValue.embed))
      ) {
        tags.push(`${this.tagPrefix}image`)
      }
```

packages/ozone/src/tag-service/embed-content-type-tagger.ts
```diff
@@ -0,0 +1,45 @@
...
+      //     @TODO: check for video embed here
+      return tags
+    } catch (err) {
+      log.error({ subject: this.subject, err }, 'Error getting record langs')
```
devinivy, line 29: Good call failing open 👍

packages/ozone/src/tag-service/language-tagger.ts
```diff
@@ -5,46 +5,29 @@ import {
...
-          subject,
-          createdBy,
-        })
-      } catch (err) {
-        log.error({ subject, err }, 'Error getting record langs')
-      }
+  isApplicable(): boolean {
+    return Boolean(
+      this.subjectStatus &&
+        !this.subjectStatus.tags?.find((tag) => tag.includes(this.tagPrefix)),
```
devinivy, line 18: Just a little more specific!
```suggestion
        !this.subjectStatus.tags?.some((tag) => tag.startsWith(this.tagPrefix)),
```

devinivy (commented):
I like the new tagging setup!

packages/ozone/src/tag-service/content-tagger.ts
```diff
@@ -0,0 +1,14 @@
+import { ModerationService } from '../mod-service'
+import { ModSubject } from '../mod-service/subject'
+import { ModerationSubjectStatusRow } from '../mod-service/types'
+
+export abstract class ContentTagger {
```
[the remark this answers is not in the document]
foysalit: I wanted to include the constructor signature and AFAIK `interface` does not allow defining constructor signature.

packages/ozone/src/tag-service/index.ts
foysalit: I normally try to split these and let the consumer of the class have the option to perform necessary checks before invoking the getter. for getters like this, i often find use-cases down the line where we would wanna know "what would be the tags IF they were applicable?". 
however, this isn't the case here... yet so I'm happy to make that check implicit.

packages/ozone/src/tag-service/language-tagger.ts
foysalit: thanks, didn't like having to change this in 2 different places so refactored the class abstract class a bit more to move this logic in one place.

packages/ozone/src/tag-service/embed-content-type-tagger.ts
foysalit: thanks. was gonna ask for the correct check here once we have the video stuff available on main.

packages/ozone/src/tag-service/embed-content-type-tagger.ts
```diff
@@ -0,0 +1,54 @@
...
+      if (!this.isApplicable()) {
+        return []
+      }
+      const recordValue = await this.getRecordValue()
+      if (!recordValue) {
+        return []
+      }
+      const tags: string[] = []
+      if (
+        AppBskyEmbedImages.isView(recordValue) &&
```
devinivy, line 29: I think we want this:
```suggestion
        AppBskyFeedPost.isRecord(recordValue) &&
```

This functionality should be fairly testable, I think—maybe should pass a post record through it.

packages/ozone/src/tag-service/index.ts
devinivy: That makes sense—it just seemed like the class could be prone to misuse, as it seemed implementations of `getTags()` were assuming it was gated by `isApplicable()`.

I think there's a way to get the best of both worlds using the abstract class, where there's a default implementation for `getTags()` that performs the `isApplicable()` check.  And then there would be a separate method for implementers `getTagsImple()` that is used within the default implementation of `getTags()`.

devinivy (approved):
Looking good!  I think just need a final pass on that one check for image tags 👍

packages/ozone/src/tag-service/embed-content-type-tagger.ts
foysalit: yeah good idea, was easier to get the right checks through a test. LMK if the current checks look right. my reasoning was, we wanna check if the record is a post with media embed first. and then, apply tags based on the type of media.

packages/ozone/src/tag-service/index.ts
foysalit: ohh I see what you mean. refactored it a bit to get that going.

packages/ozone/src/tag-service/content-tagger.ts
```diff
@@ -12,7 +12,16 @@ export abstract class ContentTagger {
   protected abstract tagPrefix: string
 
   protected abstract isApplicable(): boolean
-  abstract getTags(): Promise<string[]>
+  protected abstract buildTags(): Promise<string[]>
+
+  async getTags(): Promise<string[]> {
+    console.log(this.isApplicable(), 'isApplicable')
```
devinivy, line 18:
```suggestion
```

packages/ozone/src/tag-service/embed-tagger.ts
```diff
@@ -15,24 +21,29 @@ export class EmbedContentTypeTagger extends ContentTagger {
...
       if (!recordValue) {
         return []
       }
       const tags: string[] = []
       if (
-        AppBskyEmbedImages.isView(recordValue) &&
-        (AppBskyEmbedImages.isMain(recordValue.embed) ||
-          AppBskyEmbedRecordWithMedia.isMain(recordValue.embed))
+        AppBskyFeedPost.isRecord(recordValue) &&
+        AppBskyEmbedRecordWithMedia.isMain(recordValue.embed)
```
devinivy, line 33: There are four cases for `recordValue.embed` we should be looking for:
 - image embed
 - video embed
 - "record with media" containing image media
 - "record with media" containing video media

packages/ozone/tests/content-tagger.test.ts
```diff
@@ -0,0 +1,125 @@
...
+      ])
+
+      expect(japaneseListStatus.tags).toContain('lang:ja')
+      expect(chineseListStatus.tags).toContain('lang:ja')
+    })
+  })
+
+  describe('embed tagger', () => {
+    it('Adds image tag to post with image', async () => {
+      const post = sc.posts[sc.dids.carol][0]
```
devinivy, line 110: This is a record-with-media embed.  if we add a test for `sc.replies[sc.dids.bob][0]` I think we'll hit the direct image embed case.

devinivy (commented):
Suuuuper duper close, think we just need to add cases for those additional embed types.

devinivy (approved):

PR #2707, opened 2024-08-12
Update changeset to better reflect changes

matthieusieben:


.changeset/silver-buckets-hope.md
```diff
@@ -0,0 +1,6 @@
+---
+"@atproto/lex-cli": minor
+"@atproto/xrpc-server": patch
+---
+
+Updated to use refactored version of `@atproto/xrpc`.
```
devinivy, line 6: If this is truly just a dependency update then we don't need a changeset for it, as changesets will detect this and add a line to the changelog.  E.g. like this: https://github.com/bluesky-social/atproto/blob/main/packages/dev-env/CHANGELOG.md#0338

devinivy: Ah that's right, the `lex-cli` changes are more in-depth and are an actual version bump.

.changeset/rude-drinks-relate.md
```diff
@@ -0,0 +1,8 @@
+---
+"@atproto/dev-env": patch
+"@atproto/ozone": patch
+"@atproto/bsky": patch
+"@atproto/pds": patch
+---
+
+Updated to use refactored version of `@atproto/api`.
```
devinivy, line 8: Same here, if it's truly just for a version update then changesets will automatically detect that and add a note to the changelog automatically.

devinivy: Certainly wont do any harm though, so not terribly worried about it!

devinivy (approved):

devinivy (approved):

PR #2708, opened 2024-08-12
Export `AtpAgentOptions` type from @atproto/api

devinivy:
This gives us proper support for extending `AtpAgent`, as described in the changelog for v0.13.0.


mozzius (approved):
futur will be happy

PR #2711, opened 2024-08-13
Ensure presence of DPoP related response headers

matthieusieben:
In order to work with DPoP, the `AuthVerifier` class needs to be able to set response headers ([here](https://github.com/bluesky-social/atproto/blob/6da2fa0fa76fefed959fad02859da9c0453e1781/packages/pds/src/auth-verifier.ts#L468-L474) and [here](https://github.com/bluesky-social/atproto/blob/6da2fa0fa76fefed959fad02859da9c0453e1781/packages/pds/src/auth-verifier.ts#L501-L506)).

The current catch all handler does not provide the response object to the verifier context, preventing it from setting the headers.

This PRs ensures that the `AuthVerifier`'s `accessStandard` authentication strategy indeed receives the `res` object.

### DIFF
--- a/.changeset/giant-starfishes-fry.md
+++ b/.changeset/giant-starfishes-fry.md
@@ -1,0 +1,5 @@
+---
+"@atproto/xrpc-server": patch
+---
+
+Expose the request context for AuthVerifier and StreamAuthVerifier as distinct types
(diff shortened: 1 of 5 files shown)


devinivy (approved):
Yup!

PR #2713, opened 2024-08-14
Don't return profiles in reference lists/starter packs that have a block relationship with the owner 

haileyok:


packages/bsky/src/api/app/bsky/graph/getList.ts
```diff
@@ -68,9 +69,41 @@ const hydration = async (
...
+    // We show all users regardless of blocks if the viewer is the owner of the list, so no need to hydrate them
+    params.hydrateCtx.viewer !== ownerDid
+  ) {
+    const pairs = new Map()
+    for (const { did } of listitems) {
+      pairs.set(did, [did, ownerDid])
+    }
+    const bidirectionalBlocks =
+      await ctx.hydrator.hydrateBidirectionalBlocks(pairs)
+    return mergeStates(listState, { ...profileState, bidirectionalBlocks })
```
devinivy, line 86: If `profileState` contains some `bidirectionalBlocks` they would get clobbered here, so we'd be safer with something like `mergeManyStates(listState, profileState, { bidirectionalBlocks })`.

packages/bsky/src/api/app/bsky/graph/getList.ts
```diff
@@ -68,9 +69,41 @@ const hydration = async (
...
+  const ownerDid = new AtUri(listUri).hostname
+  if (
+    list?.record.purpose === 'app.bsky.graph.defs#referencelist' &&
+    // We show all users regardless of blocks if the viewer is the owner of the list, so no need to hydrate them
+    params.hydrateCtx.viewer !== ownerDid
+  ) {
+    const pairs = new Map()
+    for (const { did } of listitems) {
+      pairs.set(did, [did, ownerDid])
+    }
```
devinivy, line 83: If I'm not mistaken we'd be good with a single item `ownerDid -> listitems.map((li) => li.did)`.

packages/bsky/src/api/app/bsky/graph/getList.ts
```diff
@@ -68,9 +69,41 @@ const hydration = async (
...
+    const bidirectionalBlocks =
+      await ctx.hydrator.hydrateBidirectionalBlocks(pairs)
+    return mergeStates(listState, { ...profileState, bidirectionalBlocks })
+  }
+
   return mergeStates(listState, profileState)
```
devinivy, line 89: Would be kind of slick if we could remove the branch in here and end-up with something like:
```ts
const bidirectionalBlocks = maybeGetBlocksForReferenceList(/* args */)
return mergeManyStates(listState, profileState, { bidirectionalBlocks })
```

packages/bsky/src/api/app/bsky/graph/getList.ts
```diff
@@ -68,9 +70,41 @@ const hydration = async (
...
 }
 
+const noBlocks = (input: RulesFnInput<Context, Params, SkeletonState>) => {
+  const { skeleton, hydration } = input
+  if (!hydration.bidirectionalBlocks) {
+    return skeleton
+  }
+  skeleton.listitems = skeleton.listitems.filter(({ did }) => {
+    const blocks = hydration.bidirectionalBlocks?.get(did)?.values() ?? []
+    for (const block of blocks || []) {
```
devinivy, line 100: Already got the default rockin above.
```suggestion
    for (const block of blocks) {
```

packages/bsky/src/api/app/bsky/graph/getList.ts
```diff
@@ -68,7 +70,31 @@ const hydration = async (
...
+  if (!hydration.bidirectionalBlocks) {
+    return skeleton
+  }
+  skeleton.listitems = skeleton.listitems.filter(({ did }) => {
+    const blocks = hydration.bidirectionalBlocks?.get(did)?.values() || []
+    for (const block of blocks) {
+      if (block) return false
+    }
+    return true
+  })
```
devinivy, line 96: I very well may have this wrong, but I think the way hydration is working, we want something like this:
```suggestion
  const blocks = hydration.bidirectionalBlocks?.get(ownerDid)
  skeleton.listitems = skeleton.listitems.filter(({ did }) => {
    return blocks?.get(did)
  })
```

devinivy: (I believe `bidirectionalBlocks` is a map of maps `DID -> DID -> boolean`.)

packages/bsky/src/api/app/bsky/graph/getList.ts
```diff
@@ -88,6 +114,31 @@ const presentation = (
   return { list, items, cursor }
 }
 
+const maybeGetBlocksForReferenceList = async (input: {
```
devinivy, line 117: We may need similar logic in `app.bsky.graph.getStarterPack` for `starterPack.listItemsSample`.

packages/bsky/src/api/app/bsky/graph/getList.ts
haileyok: hmm let me try that again. i think i actually tried that but now curious again.

haileyok: argh, i wasn't building before test lol, sick.

packages/bsky/src/api/app/bsky/graph/getList.ts
haileyok: Added something for that, tell me if it makes sense or not - seemed to me to make the most sense to handle it inside of `hydrateStarterPacks()`.

packages/bsky/src/api/app/bsky/graph/getList.ts
```diff
@@ -68,7 +70,29 @@ const hydration = async (
...
+  const bidirectionalBlocks = await maybeGetBlocksForReferenceList({
+    ctx,
+    params,
+    skeleton,
+    listState,
+  })
+  return [listState, profileState, { bidirectionalBlocks }].reduce(
+    mergeStates,
+    {} as HydrationState,
+  )
```
devinivy, line 82: The `mergeManyStates()` helper will reduce for you!  May need to export it, but it lives right next to `mergeStates()`.

packages/bsky/src/api/app/bsky/graph/getList.ts
```diff
@@ -68,7 +70,29 @@ const hydration = async (
...
+}
+
+const noBlocks = (input: RulesFnInput<Context, Params, SkeletonState>) => {
+  const { skeleton, hydration } = input
+  if (!hydration.bidirectionalBlocks) {
+    return skeleton
+  }
```
devinivy, line 89: This branch is fine, but IMO also cool to nix it.  If we want to keep it, can get rid of the defensive check below `hydration.bidirectionalBlocks?.get(creator)`.

packages/bsky/src/api/app/bsky/graph/getList.ts
```diff
@@ -68,7 +70,29 @@ const hydration = async (
...
+  const creator = new AtUri(skeleton.listUri).hostname
```
devinivy, line 90: There's a helper `didFromUri()` that could be used here.

packages/bsky/src/hydration/hydrator.ts
```diff
@@ -650,6 +651,19 @@ export class Hydrator {
     const listMemberDids = listsMembers.flatMap((lm) =>
       lm.listitems.map((li) => li.did),
     )
+    const listBlocks: Map<string, Blocks> = new Map()
+    await Promise.all(
+      listUris.map(async (uri) => {
+        const creator = new AtUri(uri).hostname
```
devinivy, line 657: Suggest using `didFromUri()` here.

devinivy: (There are a couple more of these lying around but wont leave comments on all of them.)

packages/bsky/src/hydration/hydrator.ts
```diff
@@ -650,6 +651,19 @@ export class Hydrator {
     const listMemberDids = listsMembers.flatMap((lm) =>
       lm.listitems.map((li) => li.did),
     )
+    const listBlocks: Map<string, Blocks> = new Map()
+    await Promise.all(
+      listUris.map(async (uri) => {
```
devinivy, line 656: I'm not sure that breaking the blocks up by list helps us out a ton, can probably get rid of that idea.
```ts
const listCreatorMemberPairs = [...listMembersByList.entries()]
  .flatMap(([listUri, members]) => {
    const creator = didFromUri(listUri)
    return members.map((member) => [creator, member.did])
  })

const blocks = await this.graph.getBidirectionalBlocks(listCreatorMemberPairs)

// ...

const creator = didFromUri(sp.record.list)
// update aggregation with list items for top 12 most followed members
agg.listItemSampleUris = [...members.listitems]
  .filter((li) => !blocks.isBlocked(creator, li.did))
  // ...
```

devinivy (approved):
Good stuff 👍

PR #2714, opened 2024-08-14
Instantiate XrpcClient from an OAuthAgent

matthieusieben:
- `@atproto/api`'s `Agent` class can now be instantiated using a `SessionManager` interface (`new Agent(sessMgr)`)
- `@atproto/client`'s `OAuthSession` implements the `SessionManager` interface meaning that we can do `new Agent(oauthSession)`
- `@atproto/api`'s `AtpAgent` works exactly as before. A comment was added to say that is will soon become deprecated as it adds no value anymore.
- `@atproto/api` now exposes an `CredentialSession` class that can be used to instantiate both `new Agent(credSession)` as-well-as `new XrpcClient(credSession, myLex)`

Potentially breaking changes:

- `Agent` is no longer abstract, and its constructor signature has changed. This means that any implementation extending that class will be broken. Since this abstract class was release very recently, the impact should be minimal.

- `Agent` no longer extends `AtpBaseClient` (this is an implementation detail so should no really impact anything)

- `@atproto/oauth-client*` no longer expose an `agent`, but a `session` instead. Adapting to this is very straight forward (`new Agent(session)`).

- `OAuthAgent` was renamed into `OAuthSession` and some method signatures where changed


devinivy (approved):
Looks good!  I support this change to the oauth client 👍 I think it will be much friendlier to folks building their own apps, and at little cost of convenience.

It does stand out to me that the dependency of `@atproto/oauth-client` on `@atproto/api` is now _very_ shallow.  I can't help but wonder if there's a way to maintain the convenience of obtaining an `OAuthAtpClient` without the hard dependency on `@atproto/api`.

matthieusieben:
This whole PR introduces a breaking change in the `oauth-client*` packages but I think it's fine since those were not released yet (will impact ozone but no big deal).

cc @foysalit

packages/oauth/oauth-client/README.md
```diff
@@ -147,20 +147,142 @@ const result = await client.callback(params)
...
+const oauthSession = result.session
+```
+
+The sign-in process results in an `OAuthSession` instance that can be used to make
+authenticated requests to the resource server. This instance will automatically
+refresh the credentials when needed.
+
+### Making authenticated requests
+
+The `OAuthSession` instance obtained after singing in can be used to make
```
surfdude29, line 159: singing --> signing
```suggestion
The `OAuthSession` instance obtained after signing in can be used to make
```

packages/api/OAUTH.md
```diff
@@ -234,14 +244,16 @@ if (!agent) {
 // TO BE CONTINUED
 ```
 
-At this point in the script, the user **will** be authenticated. API calls can
-be made using the `agent`. The `agent` is an instance of a sub-class of the
-`Agent` from `@atproto/api`. Let's make a simple call to the API to retrieve the
-user's profile. Replace the `// TO BE CONTINUED` comment with the following
-code:
+At this point in the script, the user **will** be authenticated. Authenticated
+API calls can be made using the `session`. The `session` can be used to the
```
matthieusieben, line 248:
```suggestion
API calls can be made using the `session`. The `session` can be used to instantiate the
```

matthieusieben:
Thanks @surfdude29 for your careful review <3

packages/api/README.md
```diff
@@ -88,24 +88,32 @@ are available:
...
 ```
 
 ### API calls
 
 The agent includes methods for many common operations, including:
 
 ```typescript
+// The DID of the user currently authenticated (or undefined)
+agent.did
+agent.accountDid // Throw if the user is not authenticated
```
matthieusieben, line 115:
```suggestion
agent.accountDid // Throws if the user is not authenticated
```

packages/api/README.md
```diff
@@ -151,11 +159,12 @@ await agent.updateSeenNotifications()
 await agent.resolveHandle(params, opts)
 await agent.updateHandle(params, opts)
 
-// Session management (OAuth based agent instances have a different set of methods)
+// Legacy: Session management directly on the agent instance
```
matthieusieben, line 162:
```suggestion
// Legacy: Session management should be performed through the SessionManager
// rather than the Agent instance.
```

packages/oauth/oauth-client-node/README.md
```diff
@@ -415,9 +419,15 @@ app.get('/atproto-oauth-callback', async (req, res) => {
...
+      const { session, state } = await client.callback(params)
+
+      // Process successful authentication here. For example:
+
+      const agent = new Agent(session)
+
+      const profile = await agent.getProfile({ actor: agent.did })
 
-      // Process successful authentication here
+      console.log('Bsky profile:', profile.data)
```
devinivy, line 430: Was this change intentional or part of debugging?  All good either way, just checking!

.changeset/polite-toys-happen.md
```diff
@@ -0,0 +1,7 @@
+---
+"@atproto/api": minor
+---
+
+`Agent` is no longer an abstract class. Instead it can be instantiated using object implementing a new `SessionManager` interface.
+
+
```
devinivy, line 7: We flagged this as a possible but unlikely breaking change, and if it is breaking for a project it would fail early at build/typecheck time rather than runtime.  So perhaps we can knock this down to a patch and leave a short note about what to consider if you're affected.

devinivy (approved):

devinivy (approved):
A couple tiny questions but this looks good— very excited by these changes.  Breaking the dependency between `@atproto/api` and `@atproto/oauth-client` is also 💎.

packages/oauth/oauth-client-node/README.md
matthieusieben: Intentional (this is just example code)

packages/api/README.md
```diff
@@ -88,24 +88,32 @@ are available:
...
 ```
 
 ### API calls
 
 The agent includes methods for many common operations, including:
 
 ```typescript
+// The DID of the user currently authenticated (or undefined)
+agent.did
+agent.accountDid // Throws if the user is not authenticated
```
pfrazee, line 115: Interesting! I just noticed this part of the api for the first time.

Small thing: I wonder if this should be `assertDid` instead of `accountDid` to make this clearer?

pfrazee (approved):
Really great. I'm excited to be making progress on cleaning up all these interfaces -- they've been clunky for a long time before oauth.

.changeset/polite-toys-happen.md
devinivy:
```suggestion
---
"@atproto/api": patch
---

`Agent` is no longer an abstract class. Instead it can be instantiated using object implementing a new `SessionManager` interface. If your project extends `Agent` and overrides the constructor or any method implementations, consider that you may want to call them from `super`.
```

packages/api/README.md
matthieusieben: yep that would probably have been a better name -_-

devinivy: I'd recommend we add `assertDid` and deprecate `accountDid` at some point soon.

matthieusieben: https://github.com/bluesky-social/atproto/commit/bbca17bc5388e0b2af26fb107347c8ab507ee42f

PR #2715, opened 2024-08-14
:sparkles: Add report based automation for label and escalation

foysalit:


Vincentroland (approved):

foysalit:
Discussion around this revealed that we will need to scope this down to certain automation flows that will be allowed in ozone while most will remain in automod. Closing this for now but will revisit later.

PR #2717, opened 2024-08-15
Allow for optional charset in PAR content-type header

tom-sherman:
Note: I haven't actually tested this fixes things with my setup, but I'm pretty sure it will!

I'm using [oauth4webapi](https://github.com/panva/oauth4webapi) to implement atproto oauth, you can see my code for the PAR request here: https://github.com/likeandscribe/unravel/blob/b32f709d6221da03b716c19c14ce661445b9b91c/packages/frontpage/lib/auth.ts#L92

When I run this code, i receive an Error 400:

```
{"error":"invalid_request","error_description":"Unsupported content-type"}
```

oauth4webapi's `pushedAuthorizationRequest` sets a Content-Type header that can't be overwritten of `application/x-www-form-urlencoded;charset=UTF-8` (see [here](https://github.com/panva/oauth4webapi/blob/29410d8c1d420588823d72abeeeead1fe23604a6/src/index.ts#L2515)). I believe this fails the parser test in oauth-provider.

I've added a regex to allow for an optional charset on the end of the content-type

## Test plan

Not sure! I'd love some help here as setting up and deploying an entire PDS seemed quite a chore for this tiny change 😅. I couldn't see any tests for this code but happy to add them with guidance.

### DIFF
--- a/packages/oauth/oauth-provider/src/lib/http/parser.ts
+++ b/packages/oauth/oauth-provider/src/lib/http/parser.ts
@@ -38,7 +38,9 @@ export const parsers = [
   {
     name: 'urlencoded',
     test: (type: string): type is 'application/x-www-form-urlencoded' => {
-      return type === 'application/x-www-form-urlencoded'
+      return /^application\/x-www-form-urlencoded(?:;\s*charset=utf-8)?$/i.test(
+        type,
+      )
     },
     parse: (buffer: Buffer): Partial<Record<string, string>> => {
       try {


tom-sherman:
I've verified this failure by changing the content-type header in the demo at https://github.com/bluesky-social/cookbook/pull/11 to include charset, it fails in the same way.

bnewbold:
The diagnosis and solution seem correct to me! We can and should be a bit flexible here.

In terms of testing, as a unit test it would be nice to see the regex specifically match and not-match against a couple example strings. It would be great to have integration tests in general, and we do have them for most of the code in this repo, but that is probably a heavy lift for this one specific change.

I'm not a TS dev so won't review/merge, but we'll try to get to that early next week.

For bsky reviewer: we should check for other content-type checks and make them flexible to charset as well. maybe a utility for this? maybe there already is a helper in our web framework? it would be good to have at least a start of some integration tests (the dev-env stuff could help with this?). It will be some work, but it would be great to be capturing real-world issues like this as regression tests.

matthieusieben:
Replaced by: https://github.com/bluesky-social/atproto/pull/2728

PR #2721, opened 2024-08-16
Compute 3p blocks for `ReplyRef.root`

estrattonbailey:
With our new frontend feed slice construction logic, it wasn't possible to detect 3p blocks from the data we sent down in feed responses. This PR ensures that `ReplyRef.root` takes 3p blocks into account and returns a `BlockedPost` (or `NotFoundPost`) view for the frontend to compute from.

Frontend component: https://github.com/bluesky-social/social-app/pull/4944


packages/bsky/src/hydration/hydrator.ts
```diff
@@ -514,6 +514,7 @@ export class Hydrator {
       const parent = replies.get(uri)
       if (!parent?.record.reply) return
       replyParentAuthors.push(didFromUri(parent.record.reply.parent.uri))
+      postAndReplyRefs.push(parent.record.reply.parent)
```
devinivy, line 517: Is this still needed?

packages/bsky/src/views/index.ts
```diff
@@ -633,6 +633,13 @@ export class Views {
         root = parent
       }
     }
+    if (
+      !state.ctx?.include3pBlocks &&
+      state.postBlocks?.get(parent.uri)?.reply &&
```
devinivy, line 638: I think we need to alter `hydratePostBlocks()` to ensure that this the data behind this block is getting pulled.

packages/bsky/src/hydration/hydrator.ts
devinivy: I see now that this is probably what guarantees that post block data between a post's parent and root gets hydrated.  It's a pretty sneaky way to get that though, and it's going to kick off a fair amount of other work along the way.  I think getting more explicit with it in `hydratePostBlocks` is probably the way to go.

estrattonbailey: Nope was redundant

packages/bsky/src/views/index.ts
estrattonbailey: I believe it is! Within `hydratePosts`

packages/bsky/src/hydration/hydrator.ts
estrattonbailey: I think the work is already being done here 🤔 https://github.com/bluesky-social/atproto/blob/reply-grandparent/packages/bsky/src/hydration/hydrator.ts#L504

devinivy: I think that helps us cover the case `A -> B -> C`, but in the case `A -> ... -> C -> D`, there's nothing that checks the 3p blocks between `C` (`D`'s parent) and `A` (`D`'s root).

devinivy: In the case `A -> B -> C` it works because you hydrate `B` and therefore check its 3p blocks with its parent `A`, which just so happens to be `C`'s root.

estrattonbailey: I think actually that in the case of `A -> ... -> C -> D`, we do hydrate post blocks for `A <-> C`, since `C` is hydrated due to its presence on `D.reply.parent`. Added a test for this, where `D.reply.root` is a blocked post because `A` blocks `C` https://github.com/bluesky-social/atproto/pull/2721/commits/7d0deb0a08bb1be5c030e1cb9b1d502d6a6cc3a9#diff-7f86c57c90267a6c6c49cb89dd5233108fd81f59c8de18680fa7b3c1f3893205R141

estrattonbailey: Given `A -> B -> C -> D`, we can even check if `A` blocks `B` based on the data we already hydrate https://github.com/bluesky-social/atproto/pull/2721/files#diff-7f86c57c90267a6c6c49cb89dd5233108fd81f59c8de18680fa7b3c1f3893205R184

estrattonbailey: Actually, just kidding, that's only possible if the grandparent happens to be included in the first page of results. I've restricted the tests to fetching a single `FeedViewPost` now, which therefore matches the frontend, since it uses only a single `FeedViewPost` to construct a `FeedSlice`. We moved to that approach for consistency, instead of _maybe_ having the data we need in the first page of results to do further sorting/munging.

estrattonbailey: >I think actually that in the case of A -> ... -> C -> D, we do hydrate post blocks for A <-> C

Yeah and no on that too, that test was viewing as C so of course A was blocked.

Clarified the block relationships here and fixed up the tests. Triple checking again now to ensure I'm not still confused how this works :P

packages/bsky/tests/views/feed-view-post.test.ts
```diff
@@ -0,0 +1,336 @@
...
+    carol = sc.dids.carol
+    dan = sc.dids.dan
+
+    await sc.follow(carol, alice)
+    await sc.follow(carol, bob)
+    await sc.follow(carol, dan)
+    await sc.follow(dan, alice)
+    await sc.follow(dan, bob)
+    await sc.follow(dan, carol)
+  })
```
devinivy, line 46:
```suggestion
    await sc.follow(dan, carol)
    await network.processAll()
  })
```

packages/bsky/tests/views/feed-view-post.test.ts
```diff
@@ -0,0 +1,336 @@
...
+
+  // account dids, for convenience
+  let alice: string
+  let bob: string
+  let carol: string
+  let dan: string
+
+  beforeAll(async () => {
+    network = await TestNetwork.create({
+      dbPostgresSchema: 'bsky_views_feed_thread_slices',
```
devinivy, line 29: Looks like the name of the test file, the `describe()` for the test suite, and the db postgres schema may need to get aligned with each other.

packages/bsky/tests/views/feed-view-post.test.ts
```diff
@@ -0,0 +1,336 @@
...
+    expect(sliceD?.reply).toBeDefined()
+
+    if (!sliceD || !sliceD.reply) {
+      throw new Error('sliceD is undefined')
+    }
+
+    expect(sliceD.reply.parent.uri).toEqual(C.ref.uriStr)
+    expect(sliceD.reply.root.uri).toEqual(A.ref.uriStr)
+    expect(AppBskyFeedDefs.isPostView(sliceD.reply.parent)).toBe(true)
+    expect(AppBskyFeedDefs.isPostView(sliceD.reply.root)).toBe(true)
```
devinivy, line 237: This is the case `[A] -> [B] -> [C] -> [D], A blocks C, viewed as D`.

Don't we want the root (`A`) to appear blocked, so that it doesn't appear in context of the parent (`C`)?

packages/bsky/src/hydration/hydrator.ts
devinivy: When we hydrate `D` within a feed perhaps we need to ask for three blocking relationships: `(author(D), author(C))`, `(author(A), author(C))`, and `(author(A), author(D))`.  That would ensure that no two of these posts with a 3p block appear in context of each other as a part of the same thread.

packages/bsky/src/hydration/hydrator.ts
```diff
@@ -448,7 +452,14 @@ export class Hydrator {
...
         const pair: RelationshipPair = [creator, parentDid]
         relationships.push(pair)
-        postBlockPairs.reply = pair
+        postBlockPairs.parent = pair
+      }
+      const rootUri = post.reply?.root.uri
+      const rootDid = rootUri && didFromUri(rootUri)
+      if (rootDid) {
+        const pair: RelationshipPair = [creator, rootDid]
+        relationships.push(pair)
```
devinivy, line 461: Only adding `rootDid` if it's different from `parentDid` may be a nice way to mitigate unnecessary extra fetching.  We actually may even want to add a check to ensure we don't try to fetch `[creator, creator]`, which I can imagine being pretty common.

devinivy (approved):
Nailed it!  I do think we may benefit from some of the low-hanging fruit around deduplication of fetching blocks for some of these common cases (first-level replies, self-threads).  But that's a small detail and generally this looks great.

packages/bsky/src/hydration/hydrator.ts
devinivy: To be more explicit about the cases I'm thinking of—
```ts
if (rootDid !== creator) {}
if (parentDid !== creator && parentDid !== rootDid) {}
```

estrattonbailey: The logic in `replyRef` view is to check for `childBlocks?.root || parentBlocks?.root`, so not setting `root` if it's the same as `parent` would mean we'd need to check for `childBlocks.parent && parent.uri === root.uri`. We had a check like that in `replyRef` but I think the logic is a little more clear now.

`getBidirectionalBlocks` dedupes these pairs, so if `parentDid === rootDid`, I think it should be OK so long as neither `=== creator`, since in the end it will fetch the relationship once and we'll set it on both `parent` and `root`.

So I think just a check on both conditions here for `did !== creator` should serve well?

devinivy (approved):

PR #2728, opened 2024-08-19
Allow charset in content-type header of incoming requests

matthieusieben:
Replaces https://github.com/bluesky-social/atproto/pull/2717
Thanks @tom-sherman


devinivy (approved):
💎

PR #2729, opened 2024-08-20
Remove non-standard *_endpoint_auth_method

matthieusieben:
The use of `<name>_endpoint_auth_method` was inspired by https://github.com/panva/node-oidc-provider. This is however non-standard behavior that was removed in `node-oidc-provider`.

This PR removes that code from our oauth implementation


devinivy (approved):
Great!

PR #2731, opened 2024-08-21
OAuth: Removal of OpenID compatibility

matthieusieben:
This change removes the compatibility with OpenID.

The reason is that although the implementation was technically OpenID compatible, ATProto identifiers are distributed identifiers. When a client relies on OpenID to authenticate users, it will use the auth provider (issuer) in combination with the identifier to uniquely identify the user. Since ATProto identifiers are meant to be able to move from one provider to the other, OpenID compatibility could break authentication after a user was migrated to a different provider. Future authentication attempts would be considered as distinct identities from the one previsouly used by the client.

The way OpenID compliant clients would adapt to this particularity would typically be to remove the issuer + identifier combination and use the identifier alone. While this is indeed the right way to handle ATProto identifiers, it requires more work to avoid impersonation. In particular, when obtaining a user identifier, the client **must** verify that the issuer of the identity token is indeed the server responsible for that user. This mechanism being not enforced by the OpenID standard, OpenID compatibility could lead to security issues. For this reason, we decided to remove OpenID compatibility from the OAuth provider.

Note that a trusted central authority could still offer OpenID compatibility by relying on ATProto's regular OAuth flow under the hood. This capability is out of the scope of this library.


packages/oauth/oauth-provider/src/client/client-manager.ts
```diff
@@ -383,11 +373,33 @@ export class ClientManager {
...
     }
 
+    if (metadata.authorization_details_types?.length) {
+      const dupAuthDetailsType =
+        metadata.authorization_details_types.find(isDuplicate)
+      if (dupAuthDetailsType) {
+        throw new InvalidClientMetadataError(
+          `Duplicate authorization_details_type "${dupAuthDetailsType}"`,
+        )
+      }
```
devinivy, line 383: I have no issue with it, but I'm curious what spurred all the new "duplicate" checks?

packages/oauth/oauth-provider/src/token/token-manager.ts
```diff
@@ -272,12 +261,6 @@ export class TokenManager {
       sub: account.sub,
     }
 
-    await this.hooks.onTokenResponse?.call(null, tokenResponse, {
-      client,
-      parameters,
-      account,
-    })
```
devinivy, line 279: Any special reason this hook was removed?

packages/oauth/oauth-provider/src/metadata/build-metadata.ts
```diff
@@ -62,15 +35,15 @@ export function buildMetadata(
...
-      'code id_token',
-      'code token',
-      'id_token token',
-      'id_token',
+      // 'none',
+      // 'code id_token token',
+      // 'code id_token',
+      // 'code token',
+      // 'id_token token',
+      // 'id_token',
```
devinivy, line 46: Can we remove these entirely, or prefer to keep them around commented for some purpose?

devinivy (approved):
Looking good!  One question I have is how we're dealing with the default scopes for loopback clients, given that the `atproto` scope is required.

packages/oauth/oauth-provider/src/client/client-manager.ts
matthieusieben: I got the code in for scopes, then realized it was really easy to add for some other checks. Just trying to enforce better quality for clients, while providing good feelback to devs.

packages/oauth/oauth-provider/src/token/token-manager.ts
matthieusieben: Initially I wanted the `oauth-provider` to be as generic as possible and added hooks wherever it made sense to add atproto specific logic. Then we agreed that this was not a necessity. So now, I try to remove unused code where applicable.

packages/oauth/oauth-provider/src/metadata/build-metadata.ts
matthieusieben: I kept them around as a way to state that we know about those other values but we explicitly do not support them.

matthieusieben:
Basically, if the metadata document does not reference any scope, it means that there is no restriction on the scopes that can be requester during oauth flow:

https://github.com/bluesky-social/atproto/blob/24e0f45ab7c957861072e639b6ecb7b64a4d7444/packages/oauth/oauth-provider/src/request/request-manager.ts#L144-L145

Other options:

1) Remove the concept loopback clients (and require devs to publish a document somewhere)
2) Add a `scope` query string param in the loopback client id (same as for `redirect_uris`)
3) Hard code the allowed scopes. This is not very future proof.
4) Some other custom logic.

IMO, clients should ideally always declare their scopes in their metadata. For loopback clients, this would mean that they need to encode the set of scopes twice when performing an oauth flow (in the client_id and in the request parameters). This feels redundant.

Since those client are subject to a bunch of restrictions, I figured this was good enough.

devinivy (approved):

matthieusieben:
Closed because this branch is actually a sub-set of https://github.com/bluesky-social/atproto/pull/2734

PR #2734, opened 2024-08-22
OAuth: Add authorization scopes & remove OpenID compatibility

matthieusieben:
This PR adds support for scopes during the authorization flow. Three scopes are introduced

- `atproto`: acts as a negotiation scope between the client and the server so that both know that this is an ATProto OAuth flow
- `transition:generic`: This temporary scope allows to give access tokens issued through OAuth the same permissions as current "app passwords".
- `transition:chat.bsky`: Grants access to chat functionalities, as was the case for "privileged app passwords"

This change also removes the compatibility with OpenID.

The reason is that although the implementation was technically OpenID compatible, ATProto identifiers are distributed identifiers. When a client relies on OpenID to authenticate users, it will use the auth provider (issuer) in combination with the identifier to uniquely identify the user. Since ATProto identifiers are meant to be able to move from one provider to the other, OpenID compatibility could break authentication after a user was migrated to a different provider. Future authentication attempts would be considered as distinct identities from the one previsouly used by the client.

The way OpenID compliant clients would adapt to this particularity would typically be to remove the issuer + identifier combination and use the identifier alone. While this is indeed the right way to handle ATProto identifiers, it requires more work to avoid impersonation. In particular, when obtaining a user identifier, the client must verify that the issuer of the identity token is indeed the server responsible for that user. This mechanism being not enforced by the OpenID standard, OpenID compatibility could lead to security issues. For this reason, we decided to remove OpenID compatibility from the OAuth provider.

Note that a trusted central authority could still offer OpenID compatibility by relying on ATProto's regular OAuth flow under the hood. This capability is out of the scope of this library.


packages/pds/src/auth-verifier.ts
```diff
@@ -372,10 +380,6 @@ export class AuthVerifier {
     ) {
       throw new InvalidRequestError('Malformed token', 'InvalidToken')
     }
-    if ((payload.cnf as any)?.jkt) {
-      // DPoP bound tokens must not be usable as regular Bearer tokens
-      throw new InvalidRequestError('Malformed token', 'InvalidToken')
-    }
```
matthieusieben, line 378: This is covered by the `typ` test above.

packages/oauth/oauth-client-browser/example/src/auth/auth-form.tsx
```diff
@@ -10,10 +10,7 @@ export function AuthForm({
   atpSignIn?: AtpSignIn
   oauthSignIn?: OAuthSignIn
 }) {
-  const defaultMethod = useCallback(
-    () => (oauthSignIn ? 'oauth' : atpSignIn ? 'atp' : undefined),
-    [],
-  )
+  const defaultMethod = oauthSignIn ? 'oauth' : atpSignIn ? 'atp' : undefined
```
devinivy, line 13: Small thing, but we could align this with the terminology we use for session managers: `oauth` and `credentials`.

packages/oauth/oauth-client-browser/example/src/main.tsx
```diff
@@ -10,11 +10,12 @@ ReactDOM.createRoot(document.getElementById('root')!).render(
...
       // dev-env
-      // plcDirectoryUrl="http://localhost:2582"
-      // handleResolver="http://localhost:2584"
+      plcDirectoryUrl="http://localhost:2582"
+      handleResolver="http://localhost:2584"
       // production
-      plcDirectoryUrl={undefined}
-      handleResolver="https://bsky.app"
+      // plcDirectoryUrl={undefined}
+      // handleResolver="https://bsky.social"
```
devinivy, line 17: Just double checking these changes are intentional, not from development/debugging.

packages/oauth/oauth-provider/src/oauth-provider.ts
```diff
@@ -1215,20 +1254,14 @@ export class OAuthProvider extends OAuthVerifier {
...
-        const { account, info } = await server.signIn(
+        return server.signIn(
           deviceId,
+          input.request_uri,
+          input.client_id,
           input.credentials,
         )
-
-        // Prevent fixation attacks
-        await deviceManager.rotate(req, res, deviceId)
```
devinivy, line 1226: Is this mitigation no longer necessary here?

packages/pds/src/auth-verifier.ts
```diff
@@ -522,7 +539,7 @@ export class AuthVerifier {
     const { did, scope, token, audience } = await this.validateBearerToken(
       ctx,
       scopes,
-      { audience: this.dids.pds },
+      { audience: this.dids.pds, typ: 'at+jwt' },
```
devinivy, line 542: Will these new `typ` checks break existing sessions?

devinivy: Ah, I see you handle this case 👍

packages/pds/src/oauth/provider.ts
```diff
@@ -44,6 +44,8 @@ export class PdsOAuthProvider extends OAuthProvider {
         // & resource server, in which case the issuer origin is also the
         // resource server uri.
         protected_resources: [new URL(issuer).origin],
+
+        scopes_supported: ['transition:generic', 'transition:chat.bsky'],
```
devinivy, line 48: Should `atproto` be in here too?

packages/xrpc-server/src/auth.ts
```diff
@@ -65,6 +72,12 @@ export const verifyJwt = async (
   if (parts.length !== 3) {
     throw new AuthRequiredError('poorly formatted jwt', 'BadJwt')
   }
+
+  const header = parseHeader(parts[0])
+  if (header['typ'] !== 'JWT') {
+    throw new AuthRequiredError('jwt is not of type "JWT"', 'BadJwtType')
+  }
```
devinivy, line 79: I don't think this should cause any issues with our implementation since we've set this for some time, but also not sure that it's necessary.  Is there a particular reason it was introduced here?  It's not currently explicitly referenced in the spec: https://atproto.com/specs/xrpc#inter-service-authentication-jwt

devinivy: If it's serving some useful purpose to disambiguate from access/refresh tokens, perhaps a middle-ground is to loosen this a little bit and assume it is a JWT if the claim is missing.

packages/pds/src/auth-verifier.ts
```diff
@@ -484,18 +481,38 @@ export class AuthVerifier {
...
       )
 
-      const { sub } = result.claims
+      const { sub, scope } = result.claims
       if (typeof sub !== 'string' || !sub.startsWith('did:')) {
         throw new InvalidRequestError('Malformed token', 'InvalidToken')
       }
 
+      if (!scope?.includes('transition:generic')) {
+        throw new AuthRequiredError('Missing required scope', 'InvalidToken')
```
devinivy, line 490: Folks will run into this on release, it would be nice if we can help them out a bit:
```suggestion
      if (!scope?.includes('transition:generic')) {
        throw new AuthRequiredError('Missing required scope: transition:generic', 'InvalidToken')
```

packages/pds/src/auth-verifier.ts
```diff
@@ -484,18 +481,38 @@ export class AuthVerifier {
...
+      }
+
+      const scopeEquivalent = scope.includes('transition:chat.bsky')
```
devinivy, line 493: This means of checking the scopes probably works fine today, but we could run into some issues with it.  E.g. if you can get the AS to hand out a scope it doesn't know about e.g. `unknown:transition:chat.bsky` then this check would be satisfied.  Or if down the road you can further qualify/attenuate legitimate scopes as prefixes/suffixes e.g. `transition:chat.bsky.convo.getLog`.

devinivy: Would support parsing it into an array or set to makes these checks more precise.

devinivy (commented):
A few questions worth working through, but this looks great— pretty much read to roll.

packages/oauth/oauth-client-browser/example/src/main.tsx
matthieusieben: I use this a lot in dev, which means that I often have to change those params. I figured since this is not the right place for an example app anyways (docs site probably is), I could repurpose this for dev.

packages/oauth/oauth-provider/src/oauth-provider.ts
matthieusieben: This is covered by the `true` argument on the `deviceManager.load()` call.

https://github.com/bluesky-social/atproto/pull/2734/files/24e0f45ab7c957861072e639b6ecb7b64a4d7444#diff-c8e565c82e08b3dec17e4b2c34b802572ec1f20822160e72891d3c60178f3371R1257

packages/xrpc-server/src/auth.ts
matthieusieben: This was done with the same philosophy as this:
https://datatracker.ietf.org/doc/html/rfc9449#section-11.5

I guess that now that service auth tokens contain an `lxm`, they are less likely to be confused for something else but I think that when it comes to security, the more constrained the rules, the better.

packages/pds/src/auth-verifier.ts
matthieusieben: Yep, missing `typ` are allowed (for now)

packages/pds/src/auth-verifier.ts
matthieusieben: I usually parse those into an array/Set. Missed this one.

packages/xrpc-server/src/auth.ts
devinivy: My only concern is that we're making some subtle changes that affect the protocol specs, and we may want to go through a bit more process/discussion before making changes here.  For example, what if we decide we want to use a different value than `JWT`?  If nothing else in this work strictly relies on this change I wonder if we should table it temporarily.

matthieusieben: Well, a missing `typ` could arguably be interpreted as meaning `"JWT"`:
https://datatracker.ietf.org/doc/html/rfc7519#section-5.1

itaru2622:
@matthieusieben @devinivy 
please change the title of this PR,  for easy tracking the lost of openID compatibility.

I understand that this scoping feature needs to break openID compatibility through discussions, 
but current title is hard to track when openID compatibility is lost.

devinivy (approved):

PR #2735, opened 2024-08-22
Add `quoteCount` to `embed` view

haileyok:
Missed adding the count to the `embed` view, which results in a bit of layout jank in-app.


packages/bsky/src/views/index.ts
```diff
@@ -64,6 +64,7 @@ import {
 } from '../lexicon/types/app/bsky/labeler/defs'
 import { Notification } from '../proto/bsky_pb'
 import { postUriToThreadgateUri, postUriToPostgateUri } from '../util/uris'
+import post from '../data-plane/server/indexing/plugins/post'
```
devinivy, line 67:
```suggestion
```

devinivy (approved):
Once tests are passing (snapshot updates) this should be good to go!

devinivy (approved):

PR #2737, opened 2024-08-22
Return `ThreadgateView` on response from `getPostThread`

estrattonbailey:
Adds `threadgate: ThreadgateView` on the response from `getPostThread`.

The reason for this is: we need the threadgate record on thread views in the client in order to sort replies. We can't reliably get this from the root `PostView` bc that post could have been deleted, or detached via a block. We may also not have it due to the default `height` of 80 posts.

However, when constructing a feed response, we always load the threadgate, and we have a reference to the root post of the thread on the `ReplyRef`. So this PR checks if there's a threadgate hydrated for this root post, and returns it if so.

### DIFF
--- a/packages/bsky/src/api/app/bsky/feed/getPostThread.ts
+++ b/packages/bsky/src/api/app/bsky/feed/getPostThread.ts
@@ -115,14 +115,12 @@ const presentation = (
     // @TODO technically this could be returned as a NotFoundPost based on lexicon
     throw new InvalidRequestError(`Post not found: ${params.uri}`, 'NotFound')
   }
-  let threadgate: OutputSchema['threadgate']
-  if (isThreadViewPost(thread) && isPostRecord(thread.post.record)) {
-    const rootUri = thread.post.record.reply?.root?.uri || thread.post.uri
-    threadgate = ctx.views.threadgate(
-      postUriToThreadgateUri(rootUri),
-      hydration,
-    )
-  }
+  const rootUri =
+    hydration.posts?.get(params.uri)?.record.reply?.root.uri ?? params.uri
+  const threadgate = ctx.views.threadgate(
+    postUriToThreadgateUri(rootUri),
+    hydration,
+  )
   return { thread, threadgate }
 }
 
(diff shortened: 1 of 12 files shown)


estrattonbailey:
Gonna need snapshots, just getting initial thoughts right now.

packages/bsky/src/api/app/bsky/feed/getPostThread.ts
```diff
@@ -106,11 +111,20 @@ const presentation = (
     height: params.parentHeight,
     depth: params.depth,
   })
+
+  let rootUri: string | undefined = undefined
+  if (isThreadViewPost(thread) && isPostRecord(thread.post.record)) {
```
dholms, line 116: it may be a bit simpler to just get the rootUri off of the hydration state with the help of params rather than off the thread view. Something like:
`const rootUri = hydration.posts?.get(params.uri)?.record.reply?.root ?? params.uri`

estrattonbailey: Ah yeah, that was kinda my initial thought too, but `params.uri` could be a child within a thread too, and we then return its parents. So basically need to find any post within a thread and get its reply root. And if it's the root post in the thread, then there's no `ReplyRef`.

dholms: hmmm I think there's 4 cases:
- `params.uri` is the root & the record is found, in which case the record won't have a reply root set and we'll end up using `params.uri`
- `params.uri` is the root & the record is not found, in which case we'll just end up using the `params.uri`
- `params.uri` is a child & the record is found, in which case, it should have reply root set (and the dataplane will hinge it's thread construction off that reply root) so we'll end up using reply root
- `params.uri` is a child & the record is not found, in which case we'll end up using `params.uri`

I believe the first 3 are correct. But you're right that the 4th is incorrect. Maybe we could check if the post is found and if not then rootUri is `undefined`? That's essentially the same logic you have just driving it off of the hydration map instead of off the view.

estrattonbailey: Oy, yes, misunderstood this. This is much cleaner!

dholms (approved):
nice looks great 👌

PR #2741, opened 2024-08-23
Appview: fix uri used to place threadgate in getPostThread response

devinivy:
The issue here was that `params.uri` is the "unresolved" thread anchor URI, which may contain a handle as the authority.  By switching to `skeleton.anchor` we use the fully resolved URI instead.


haileyok:
🙏

haileyok (approved):

PR #2745, opened 2024-08-26
Repo write commit metadata

dholms:
Return some optional commit metadata on repo write methods, namely commit CID & commit rev.

This is helpful metadata generally, but can be used in particular for something like managing read-after-write semantics by writing applications


lexicons/com/atproto/repo/applyWrites.json
```diff
@@ -37,6 +37,16 @@
           }
         }
       },
+      "output": {
+        "encoding": "application/json",
+        "schema": {
+          "type": "object",
+          "properties": {
+            "commitCid": { "type": "string", "format": "cid" },
+            "commitRev": { "type": "string" }
```
devinivy, line 46: If we want these to be required when either is present we could reshape it as `commit: { cid, rev }`.

packages/pds/src/api/com/atproto/repo/applyWrites.ts
```diff
@@ -125,6 +125,14 @@ export default function (server: Server, ctx: AppContext) {
 
       await ctx.sequencer.sequenceCommit(did, commit, writes)
       await ctx.accountManager.updateRepoRoot(did, commit.cid, commit.rev)
+
+      return {
+        encoding: 'application/json',
+        body: {
+          commitCid: commit?.cid.toString(),
+          commitRev: commit?.rev,
```
devinivy, line 133:
```suggestion
          commitCid: commit.cid.toString(),
          commitRev: commit.rev,
```

devinivy (approved):
Lookin good!  Tossed out one idea but down with whatever you land on @dholms.

lexicons/com/atproto/repo/applyWrites.json
dholms: yeah i was back and forth on this, but i dig 👍

PR #2746, opened 2024-08-26
Handle empty bodies in lexicon validation

dholms:
There's a weird edge case in lexicon validation around a schema moving from having _no_ output schema to having an output schema with all optional properties. This is allowed in Lexicon. And it actually works with the current code but not for the reasons we expect.

The existing code will parse the body as an empty array and then treat it as an object and do lexicon checks against it as if it is the parsed JSON of the response.

Instead, we parse an empty body with no content type as `undefined`, and then in the higher level lexicon assertions (for validating output/input/params), pass instead pass the `undefined` through as an empty JSON object.

### DIFF
--- a/.changeset/afraid-mirrors-smoke.md
+++ b/.changeset/afraid-mirrors-smoke.md
@@ -1,0 +1,6 @@
+---
+"@atproto/lexicon": patch
+"@atproto/xrpc": patch
+---
+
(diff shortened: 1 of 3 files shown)


Vincentroland (approved):

dholms:
there's more thinking that needs to go into this. closing for now

PR #2749, opened 2024-08-27
OAuth: fix client-side crash on authorize page

devinivy:
Fixes a client-side issue when attempting to display the client id during the authorize flow.


bnewbold (approved):
Not a React-ionary, but LGTM

PR #2751, opened 2024-08-27
Video lexicons and appview views

devinivy:
This contains three pieces of work:
1. Lexicons for embedding video w/ captions within bsky post records.
2. Lexicons for speaking to the video upload service, under `app.bsky.video`.  Note that will not necessarily live at the same domain as the bsky appview API.
3. Implementation in the appview for presenting views of video embeds.

From a high level view, here's how video interplays between bsky and pds hosting:
 - the pds hosts originals of videos.
 - the appview runs a service (specified by `app.bsky.video` lexicons) which processes originals and serves them from a cdn.
 - you may upload a video to your pds via this service, it will ensure it's processed before your post is viewed.


packages/bsky/src/index.ts
```diff
@@ -63,6 +64,10 @@ export class BskyAppView {
     const imgUriBuilder = new ImageUriBuilder(
       config.cdnUrl || `${config.publicUrl}/img`,
     )
+    const videoUriBuilder = new VideoUriBuilder({
+      playlistUrlPattern: `${config.publicUrl}/vid/%s/%s/playlist.m3u8`,
```
dholms, line 68: should this be the config value instead (ie `config.videoPlaylistUrlPattern`) and only if that isn't set then it uses the `publicUrl` based pattern? a la the imgUriBuilder

dholms (approved):
looks great 🙌 just the one comment

packages/bsky/src/index.ts
devinivy: Ay, good catch, thanks!

PR #2752, opened 2024-08-28
Atproto sync package

dholms:
Introduces the `@atproto/sync` package for syncing data from the atproto network. Currently this library only supports firehose(relay) subscriptions.

The package hinges on two main classes:
- `Firehose`: spins up a firehose connection to the requested service, parses events, authenticates commits and handle updates, and then passes the parsed events off to a supplied handler
- `SyncQueue`: an in-memory partitioned queue that allows the firehose to process events concurrently (while still processing them serially per-repo). Tracks the latest cursor to ensure no dropped events


packages/sync/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "@atproto/sync",
+  "version": "0.1.0",
```
devinivy, line 3: We can probably set this to `0.0.0` while adding a changeset for a minor version bump.

packages/sync/src/events.ts
```diff
@@ -0,0 +1,61 @@
...
+export type CommitMeta = {
+  seq: number
+  time: string
+  commit: CID
+  blocks: BlockMap
+  rev: string
+  uri: AtUri
+  did: string
+  collection: string
+  rkey: string
```
devinivy, line 18: What are these in the context of `CommitMeta`?  I will read on to find out!

packages/sync/src/events.ts
```diff
@@ -0,0 +1,61 @@
...
+  event: 'delete'
+}
+
+export type IdentityEvt = {
+  seq: number
+  time: string
+  event: 'identity'
+  did: string
+  handle?: string
+  didDocument: DidDocument | null
```
devinivy, line 45: Is there room to standardize around optional fields or `null` fields?  Or perhaps the semantics between the two are intentionally different here.

packages/sync/src/firehose/index.ts
```diff
@@ -0,0 +1,342 @@
...
+import { didAndSeqForEvt } from '../util'
+
+export type FirehoseOptions = {
+  idResolver: IdResolver
+
+  handleEvt: (evt: Event) => Promise<void>
+  onError: (err: Error) => void
+  getCursor?: () => Promise<number | undefined>
+
+  syncQueue?: SyncQueue // sync queue getCursor takes precedence over `getCursor`
```
devinivy, line 47: Perhaps we oughtta fail early if both are specified.

packages/sync/src/firehose/index.ts
```diff
@@ -0,0 +1,342 @@
...
+          const cursor = await opts.getCursor()
+          return { cursor }
+        }
+        return undefined
+      },
+      validate: (value: unknown) => {
+        try {
+          return isValidRepoEvent(value)
+        } catch (err) {
+          this.opts?.onError(new FirehoseValidationError(err, value))
```
devinivy, line 86: I think this is always going to exist.
```suggestion
          this.opts.onError(new FirehoseValidationError(err, value))
```

packages/sync/src/firehose/index.ts
```diff
@@ -0,0 +1,342 @@
...
+        } else {
+          await this.processEvt(evt)
+        }
+      }
+    } catch (err) {
+      if (err && err['name'] === 'AbortError') {
+        this.destoryDefer.resolve()
+        return
+      }
+      this.opts.onError?.(new FirehoseSubscriptionError(err))
```
devinivy, line 119:
```suggestion
      this.opts.onError(new FirehoseSubscriptionError(err))
```

packages/sync/src/firehose/index.ts
```diff
@@ -0,0 +1,342 @@
...
+  IdentityEvt,
+} from '../events'
+import { CID } from 'multiformats/cid'
+import { SyncQueue } from '../queue'
+import { didAndSeqForEvt } from '../util'
+
+export type FirehoseOptions = {
+  idResolver: IdResolver
+
+  handleEvt: (evt: Event) => Promise<void>
```
devinivy, line 43: Just a suggestion, but I'd lean towards spelling this out in the public API.
```suggestion
  handleEvent: (evt: Event) => Promise<void>
```

devinivy, line 43: This allows the method to not be defined as `async`.
```suggestion
  handleEvt: (evt: Event) => Awaited<void>
```

packages/sync/src/firehose/index.ts
```diff
@@ -0,0 +1,342 @@
...
+  onError: (err: Error) => void
+  getCursor?: () => Promise<number | undefined>
```
devinivy, line 45:
```suggestion
  getCursor?: () => Awaited<number | undefined>
```

packages/sync/src/firehose/index.ts
```diff
@@ -0,0 +1,342 @@
...
+          await this.processEvt(evt)
+        }
+      }
+    } catch (err) {
+      if (err && err['name'] === 'AbortError') {
+        this.destoryDefer.resolve()
+        return
+      }
+      this.opts.onError?.(new FirehoseSubscriptionError(err))
+      await wait(3000)
```
devinivy, line 120: What is this wait up to?

packages/sync/src/firehose/index.ts
```diff
@@ -0,0 +1,342 @@
...
+          this.opts.idResolver,
+          evt,
+          this.opts.unauthenticatedHandles,
+        )
+        return parsed ? [parsed] : []
+      } else {
+        return []
+      }
+    } catch (err) {
+      this.opts.onError?.(new FirehoseParseError(err, evt))
```
devinivy, line 151:
```suggestion
      this.opts.onError(new FirehoseParseError(err, evt))
```

packages/sync/src/firehose/index.ts
```diff
@@ -0,0 +1,342 @@
...
+    const results = await verifyProofs(evt.blocks, claims, did, key)
+    results.verified.forEach((op) => {
+      const path = formatDataKey(op.collection, op.rkey)
+      verifiedCids[path] = op.cid
+    })
+  } catch (err) {
+    if (err instanceof RepoVerificationError) {
+      return parseCommitAuthenticated(idResolver, evt, true)
+    }
+    throw err
```
devinivy, line 199: Probably want to ensure this retries just once, wont loop.
```suggestion
    if (err instanceof RepoVerificationError && !forceKeyRefresh) {
      return parseCommitAuthenticated(idResolver, evt, true)
    }
    throw err
```

packages/sync/src/firehose/index.ts
```diff
@@ -0,0 +1,342 @@
...
+  return formatCommitOps(evt, evt.ops)
+}
+
+const formatCommitOps = async (evt: Commit, ops: RepoOp[]) => {
+  const car = await readCar(evt.blocks)
+
+  const evts: CommitEvt[] = []
+
+  for (const op of ops) {
+    const uri = new AtUri(`at://${evt.repo}/${op.path}`)
```
devinivy, line 223:
```suggestion
    const uri = AtUri.make(evt.repo, op.path)
```

packages/sync/src/firehose/index.ts
```diff
@@ -0,0 +1,342 @@
...
+      uri,
+      did: uri.host,
+      collection: uri.collection,
+      rkey: uri.rkey,
+    }
+
+    if (op.action === 'create' || op.action === 'update') {
+      if (!op.cid) continue
+      const recordBytes = car.blocks.get(op.cid)
+      if (!recordBytes) continue
```
devinivy, line 240: I wonder if there's a case that these should cause an error in the `parseCommitAuthenticated()` path.

packages/sync/src/firehose/index.ts
```diff
@@ -0,0 +1,342 @@
...
+
+export const parseAccount = (evt: Account): AccountEvt | undefined => {
+  if (evt.status && !isValidStatus(evt.status)) return
+  return {
+    event: 'account',
+    seq: evt.seq,
+    time: evt.time,
+    did: evt.did,
+    active: evt.active,
+    status: evt.status as AccountStatus,
```
devinivy, line 303: Based on the check above, could `evt.status` also be missing (in addition to `AccountStatus`)?

packages/sync/README.md
```diff
@@ -0,0 +1,93 @@
...
+})
+firehose.start()
+
+// on service shutdown
+await firehose.destroy()
+```
+
+For more robust indexing pipelines, it's recommended to use the supplied `SyncQueue` class. This provides an in-memory partitioned queue. As events from a given repo must be processed in order, this allows events to be processed concurrently while still processing events from any given repo serially.
+
+The `SyncQueue` also tracks an internal cursor based on the last finished consecutive work. This ensures that no events are dropped, altho it does mean that some events may occassionally be replayed (if the websocket drops and reconnects) and therefore it's recommended that any indexing logic is idempotent. An optional `setCursor` parameter may be supplied to the `SyncQueue` which can be used to persistently store the most recently processed cursor.
```
devinivy, line 56:
```suggestion
The `SyncQueue` also tracks an internal cursor based on the last finished consecutive work. This ensures that no events are dropped, although it does mean that some events may occassionally be replayed (if the websocket drops and reconnects) and therefore it's recommended that any indexing logic is idempotent. An optional `setCursor` parameter may be supplied to the `SyncQueue` which can be used to persistently store the most recently processed cursor.
```

packages/sync/README.md
```diff
@@ -0,0 +1,93 @@
...
+    } else if (evt.event === 'delete') {
+      // ...
+    }
+  },
+  onError: (err) => {
+    console.error(err)
+  },
+  filterCollections: ['com.myexample.app'],
+})
+firehose.start()
```
devinivy, line 48: It might not be the time to nail this, but it would be kind of nice if `start()` could be awaited and crash if the connection could not be initially established.

packages/sync/src/queue/index.ts
```diff
@@ -0,0 +1,64 @@
...
+// Partitions are created lazily and taken out of memory when they go idle.
+export class SyncQueue {
+  consecutive = new ConsecutiveList<number>()
+  mainQueue: PQueue
+  partitions = new Map<string, PQueue>()
+  cursor: number
+
+  constructor(public opts: SyncQueueOptions = {}) {
+    this.mainQueue = new PQueue({ concurrency: opts.concurrency ?? Infinity })
+    this.cursor = opts.startCursor ?? 0
```
devinivy, line 22: Is it more natural to backfill or to hop right into the stream by default?  I lean towards the latter because it's less of a commitment, less load on the relay, and both are fairly arbitrary times to pick-up from.

packages/sync/src/queue/index.ts
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  async addTask(partitionId: string, task: () => Promise<void>) {
+    if (this.mainQueue.isPaused) return
```
devinivy, line 26: I wonder if we want to hoist this up into `trackEvt` so that the consecutive/cursor logic, etc. doesn't get triggered during destroy.

devinivy: Oh I see, you want `addTask` to be public API so it isn't necessarily guarded.  Maybe both places then?

packages/sync/src/queue/index.ts
```diff
@@ -0,0 +1,64 @@
...
+    let partition = this.partitions.get(partitionId)
+    if (!partition) {
+      partition = new PQueue({ concurrency: 1 })
+      partition.once('idle', () => this.partitions.delete(partitionId))
+      this.partitions.set(partitionId, partition)
+    }
+    return partition
+  }
+
+  async trackEvt(did: string, seq: number, handler: () => Promise<void>) {
```
devinivy, line 42: Just a suggestion, but for the same reason as `handleEvt`, sort of lean towards `trackEvent` given it's a public API.

packages/sync/src/util.ts
```diff
@@ -0,0 +1,10 @@
+import { isAccount, isCommit, isIdentity, RepoEvent } from './firehose/lexicons'
+
+export const didAndSeqForEvt = (
+  evt: RepoEvent,
+): { did: string; seq: number } | undefined => {
+  if (isCommit(evt)) return { seq: evt.seq, did: evt.repo }
+  else if (isAccount(evt) || isIdentity(evt))
+    return { seq: evt.seq, did: evt.did }
```
devinivy, line 8: RIP

packages/sync/tsconfig.build.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "../../tsconfig/isomorphic.json",
```
devinivy, line 2: It would be nice if this package could be isomorphic, but I don't think its deps really permit it.
```suggestion
  "extends": "../../tsconfig/node.json",
```

packages/sync/README.md
```diff
@@ -0,0 +1,93 @@
...
+  },
+  onError: (err) => {
+    console.error(err)
+  },
+})
+firehose.start()
+
+// on service shutdown
+await firehose.destroy()
+await syncQueue.destroy
```
devinivy, line 83:
```suggestion
await syncQueue.destroy()
```

packages/sync/package.json
dholms: yeah i'm game

packages/sync/src/events.ts
dholms: _kinda_ different semantics, but not enough.  i'll make them both `undefined` 👌

packages/sync/src/firehose/index.ts
dholms: yea you right

dholms: wow how did i not know this. ran into this so many times lol

packages/sync/src/firehose/index.ts
dholms: ah is supposed to use the `subscriptionReconnectDelay` with a default value of `3000`

packages/sync/src/firehose/index.ts
dholms: yeah good call

packages/sync/src/firehose/index.ts
dholms: not sure i follow?

packages/sync/README.md
dholms: lmao my slack spelling making it into my readme writing

packages/sync/src/queue/index.ts
dholms: yeah you're right 👍

packages/sync/src/queue/index.ts
dholms: yeah good call

packages/sync/src/util.ts
dholms: yup 🙃

packages/sync/tsconfig.build.json
dholms: good catch

packages/sync/src/firehose/index.ts
devinivy: `evt.status` is being casted to `AccountStatus`, but from looking at some of the lines of code above I believe it also may be missing.

devinivy: We have passed the check:
```ts
if (evt.status && !isValidStatus(evt.status)) return
```

So that means `evt.status` is either falsy or it's a valid `AccountStatus`.

dholms: ahhh got it got it 👍

Vincentroland (approved):

Vincentroland (approved):

PR #2753, opened 2024-08-28
OAuth: Enable to configure scope via query parameter for loopback client metadata

sugyan:
The draft OAuth Spec says that for localhost clients, the `scope` "can be configured via query parameter in the `client_id` URL", and "Default is `atproto`”, but this was not the case.
https://github.com/bluesky-social/atproto-website/blob/bnewbold/spec-oauth/content/specs/auth.md#localhost-client-development
So I implemented it to work as per the spec.

### DIFF
--- a/packages/oauth/oauth-provider/src/request/request-manager.ts
+++ b/packages/oauth/oauth-provider/src/request/request-manager.ts
@@ -119,7 +119,7 @@ export class RequestManager {
     // > server MUST include the scope response parameter in the token response
     // > (Section 3.2.3) to inform the client of the actual scope granted.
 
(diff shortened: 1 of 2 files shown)


matthieusieben:
Hey ! Thank you for you suggestion. This issue was actually reported to us yesterday and I already started integrating that fix in [another branch](https://github.com/bluesky-social/atproto/pull/2755).

Since both are changes redundant, and because this branch did not account for "multiple allowed" part of the spec, I'll go ahead and close this PR in favor of the other one.

Feel free to comment / open another issue.

Vincentroland (approved):

PR #2755, opened 2024-08-28
OAuth spec alignment

matthieusieben:
This PR fixes several details in order to make atproto's reference implementation match the atproto Oauth spec.


.changeset/late-wolves-itch.md
```diff
@@ -0,0 +1,5 @@
+---
+"@atproto/oauth-provider": patch
+---
+
+Enfore code_challenge_method=S256 request parameter
```
devinivy, line 5:
```suggestion
Enforce code_challenge_method=S256 request parameter
```

packages/oauth/oauth-provider/src/errors/invalid-client-metadata-error.ts
```diff
@@ -12,8 +12,16 @@ export class InvalidClientMetadataError extends OAuthError {
...
+  static from(
+    cause: unknown,
+    fallbackMessage = 'Invalid client metadata document',
+  ): OAuthError {
+    if (cause instanceof OAuthError) {
+      return cause
+    }
+    if (cause instanceof TypeError) {
+      return new InvalidClientMetadataError(cause.message, cause)
+    }
```
devinivy, line 24: I don't think we need to change this, but it is a bit interesting that a system `TypeError` (e.g. "null has no properties") could make it into a user-facing error, if I'm understanding correctly.

matthieusieben: This is due to the fact that this error is meant to be used in the context of parsing a client_id, meaning that any TypeError (e.g. InvalidURL) is actually not a code error but rather a user data error. I could move that logic elsewhere where `TypeError` are actually thrown ?

matthieusieben: I added a comment to explain the reasonning.

Vincentroland (approved):

Vincentroland (approved):

matthieusieben:
Likely to fix  #2814

.changeset/great-news-shout.md
```diff
@@ -0,0 +1,5 @@
+---
+"@atproto/oauth-types": patch
+---
+
+Enfoce ClientID URL path to be normalized
```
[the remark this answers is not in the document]
devinivy: I took a closer look, and I see now that the path normalization check is fairly in line with draft-parecki-oauth-client-id-metadata-document-01 👍

packages/oauth/oauth-types/src/oauth-client-id-loopback.ts
```diff
@@ -15,44 +18,89 @@ export function isOAuthClientIdLoopback<C extends OAuthClientId>(
...
+
+  if (clientId.length === queryStringIdx) {
+    return {} // no query string to parse
   }
 
-  if (url.hash) {
-    throw new TypeError('Loopback ClientID must not contain a fragment')
+  if (clientId[queryStringIdx] !== '?') {
+    throw new TypeError('Loopback ClientID must not contain a path component')
   }
```
devinivy, line 53: Do we handle the case that there's no path component and also no query params?

devinivy (approved):

packages/oauth/oauth-types/src/oauth-client-id-loopback.ts
matthieusieben: Yep, in the previous check (`if (clientId.length === queryStringIdx)`)

PR #2757, opened 2024-08-28
Fix blocked posts showing up in `getQuotes`

mozzius:
3rd party blocks are showing up in `getQuotes` due to passing a `uri` rather than a `did` to `viewerBlockExists`. This PR should fix that


estrattonbailey:
Fixed https://github.com/bluesky-social/atproto/pull/2758

estrattonbailey:
Tests and util in https://github.com/bluesky-social/atproto/pull/2757

PR #2767, opened 2024-08-30
Update email templates

estrattonbailey:
These emails are generated with https://react.email, and a small custom script to output `.hbs` templates. Preview variables are replaced with `{{ var }}` syntax.

Only functional change is that `reset-password` how accepts a `handle` instead of `identifier`, so that the templates match our entryway.

Before/after

# Confirm email
CleanShot 2024-08-30 at 15 05 13@2x
CleanShot 2024-08-30 at 15 05 23@2x

# Delete account
CleanShot 2024-08-30 at 15 06 00@2x
CleanShot 2024-08-30 at 15 05 55@2x

# PLC update operation
Was a little unsure on wording here, unsure when this is used.
CleanShot 2024-08-30 at 15 06 26@2x
CleanShot 2024-08-30 at 16 39 54@2x

# Reset password
CleanShot 2024-08-30 at 15 06 59@2x
CleanShot 2024-08-30 at 15 06 52@2x

# Update email
CleanShot 2024-08-30 at 15 07 32@2x
CleanShot 2024-08-30 at 15 07 26@2x


pfrazee (approved):
LFG. I think the only tweak is the plc operation needs to specify that you're seeing the confirmation code

estrattonbailey:
Updated, updated screenshot:

`We received a request to update your PLC. Your confirmation code is:`

heikadog:
@estrattonbailey at least on my end i've seen the PLC email when i set up my labeler and i have 2FA on 
Screenshot_20240831-125136

Vincentroland (approved):

dholms (approved):
incredible. beautiful. worthy of praise.

PR #2770, opened 2024-08-31
PDS pipethrough optimizations

matthieusieben:


Vincentroland (approved):

matthieusieben:
Still need to add some tests

packages/internal/fetch-node/src/ssrf.ts
```diff
@@ -43,15 +43,6 @@ export function ssrfFetchWrap<C = FetchContext>({
     }
 
     if (url.protocol === 'http:' || url.protocol === 'https:') {
-      // @ts-expect-error non-standard option
-      if (request.dispatcher) {
```
matthieusieben, line 47: `dispatcher` was never exposed on `Request`

packages/xrpc-server/src/types.ts
```diff
@@ -238,20 +277,52 @@ export function isHandlerError(v: unknown): v is HandlerError {
...
-    super(ResponseType.InvalidRequest, errorMessage, customErrorName)
+  constructor(
+    errorMessage?: string,
+    customErrorName?: string,
+    options?: ErrorOptions,
+  ) {
+    super(ResponseType.InvalidRequest, errorMessage, customErrorName, options)
+  }
+
+  [Symbol.hasInstance](instance: unknown): boolean {
```
matthieusieben, line 288: Allows `err instanceof InvalidRequestError` to return `true` when error was build using `new XRPCError(ResponseType.InvalidRequest, ...)`

packages/common/src/streams.ts
```diff
@@ -29,12 +32,21 @@ export const streamSize = async (stream: Readable): Promise<number> => {
...
+    if (chunk instanceof Uint8Array) {
+      chunks.push(chunk)
+      totalLength += Buffer.byteLength(chunk)
+    } else {
+      throw new TypeError('expected Uint8Array')
+    }
   }
-  return new Uint8Array(Buffer.concat(bufs))
+  // Note Buffer is a subclass of Uint8Array
+  return Buffer.concat(chunks, totalLength)
```
dholms, line 49: any reason for changing the return type here?

i've been bit by subtle differences between Buffers & Uint8Arrays before & kinda prefer using just one or the other in most of our stuff is possible. or at least differentiating between the two and saying "bytes" for Uint8Arrays and "buffer" for Buffers

packages/xrpc-server/src/server.ts
```diff
@@ -286,29 +290,28 @@ export class Server {
...
 
-        if (outputUnvalidated && isHandlerPipeThrough(outputUnvalidated)) {
-          // set headers
-          if (outputUnvalidated?.headers) {
-            Object.entries(outputUnvalidated.headers).forEach(([name, val]) => {
-              res.header(name, val)
-            })
-          }
+        if (isHandlerPipeThroughBuffer(outputUnvalidated)) {
+          setHeaders(res, outputUnvalidated)
```
dholms, line 294: clean

packages/xrpc-server/src/server.ts
```diff
@@ -481,30 +484,27 @@ export class Server {
...
-  if (!isString(v['encoding']) || !(v['buffer'] instanceof ArrayBuffer)) {
-    return false
-  }
-  if (v['headers'] !== undefined) {
-    if (v['headers'] === null || typeof v['headers'] !== 'object') {
-      return false
-    }
-    if (!Object.values(v['headers']).every(isString)) {
-      return false
+const createValidator =
```
dholms, line 487: I think we avoided zod here because it can be kinda heavy duty for validation of stuff that is often invalid. I know it used to handle this by throwing and catching errors which can be rough on Node so we tried to keep it out of the happy/critical path of our server stuff. 

But I also have some vague memory that they fixed this & made it faster? Any idea? Or any idea how these compare?

packages/xrpc-server/src/server.ts
```diff
@@ -481,30 +484,27 @@ export class Server {
...
+  <T>(schema: ZodSchema<T>) =>
+  (v: unknown): v is T =>
+    schema.safeParse(v).success
```
dholms, line 490: we have a helper function for this that we use in most places in `common-web/check`

`check.is(v, schema)`

packages/xrpc-server/src/util.ts
```diff
@@ -360,11 +321,38 @@ export interface ServerTiming {
...
 export const parseReqNsid = (
   req: express.Request | IncomingMessage,
 ): string => {
-  const originalUrl =
-    ('originalUrl' in req && req.originalUrl) || req.url || '/'
-  const nsid = originalUrl.split('?')[0].replace('/xrpc/', '')
-  return nsid.endsWith('/') ? nsid.slice(0, -1) : nsid // trim trailing slash
+  // /!\ Hot path
+
+  const originalUrl = ('originalUrl' in req && req.originalUrl) || req.url
```
dholms, line 331: could we rewrite as ternary? may be a bit easier to read on first glance

packages/pds/src/context.ts
```diff
@@ -256,12 +263,36 @@ export class AppContext {
       appviewCdnUrlPattern: cfg.bskyAppView?.cdnUrlPattern,
     })
 
+    const safeAgent = new undici.Agent({
+      allowH2: cfg.fetch.allowHTTP2, // This is experimental
+      headersTimeout: cfg.fetch.headersTimeout,
+      maxResponseSize: cfg.fetch.maxResponseSize,
```
dholms, line 269: are we cool re-using these config values for both the safeFetch & safeAgent? 

only issue I could think of is if we have different tradeoffs on `maxResponseSize`

packages/pds/src/context.ts
```diff
@@ -256,12 +263,36 @@ export class AppContext {
       appviewCdnUrlPattern: cfg.bskyAppView?.cdnUrlPattern,
     })
 
+    const safeAgent = new undici.Agent({
```
dholms, line 266: I almost want to call this `proxyAgent` or `safeProxyAgent` or something similar to give the sense of what it's for since it's only use there

packages/pds/src/pipethrough.ts
```diff
@@ -179,32 +334,111 @@ export const parseProxyHeader = async (
...
+
+function formatAccepted(accept: readonly Accept[]): string {
+  return accept.map(formatEncodingDev).join(', ')
+}
+
+function formatEncodingDev([enc, { q }]: Accept): string {
+  return q != null ? `${enc};q=${q}` : enc
+}
+
+function nameIncludedIn(this: readonly string[], accept: Accept): boolean {
```
dholms, line 376: is there a big performance gain from the `this`/`thisArg` syntax? I find it harder to reason about personally

packages/pds/src/pipethrough.ts
```diff
@@ -179,32 +334,111 @@ export const parseProxyHeader = async (
...
-// Sending request
+// Request parsing/forwarding
 // -------------------
 
-export const makeRequest = async (
-  url: URL,
-  reqInit: RequestInit,
-): Promise<Response> => {
-  let res: Response
+type Accept = [name: string, flags: { q: string }]
```
dholms, line 340: thoughts on 

```
type Accept = {
  name: string
  q: string
}
```

instead of the array syntax? seems more clear & type-safe

dholms (approved):
left some small tweaks/questions, but broadstrokes this looks awesome 💪 super stoked to get this in

packages/common/src/streams.ts
matthieusieben: `Buffer` is a sub-class of `Uint8Array`. So it kinda makes no sense to wrap a Buffer within an `Uint8Array` since we loose functionalities & we create two objects instead of one.
Since this function is only used in the pipethrough code, I will rename the fn.

packages/common-web/src/did-doc.ts
```diff
@@ -83,40 +83,54 @@ export const getServiceEndpoint = (
...
-  } else if (!url.hostname) {
-    return undefined
-  } else {
+
+  try {
+    const url = new URL(urlStr)
+
+    if (!url.hostname) {
+      return undefined
+    }
```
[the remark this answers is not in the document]
matthieusieben: Absolutely

packages/oauth/oauth-provider/src/lib/http/stream.ts
```diff
@@ -73,9 +47,9 @@ export async function parseStream(
   )
 
   if (!parser) {
-    throw createHttpError(400, 'Unsupported content-type')
+    throw new TypeError(`Unsupported content-type: ${type.mime}`)
```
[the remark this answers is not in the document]
matthieusieben: Handled here:
https://github.com/bluesky-social/atproto/blob/3368726a77835d0a7197ba31239c17ea63ce2e38/packages/oauth/oauth-provider/src/lib/http/request.ts#L27

The idea is that `parseStream` is generic (does not know if its parsing a request, a response, a file on disk, etc.) so the responsibility to convert the error falls onto the caller.

packages/xrpc-server/src/server.ts
matthieusieben: Added a `check.create(schema)` as the semantics I needed are a bit different

packages/xrpc-server/src/server.ts
matthieusieben: You are right, validating here is way overkill. Since we know the input is of type `HandlerOutput`, all we need is to discriminate based on that. We can even make some of the checks we currently do disappear !

packages/common/src/streams.ts
matthieusieben: Note that `common-web` already has a `streamToBuffer` (that actually performs a `AsyncIterable<Uint8Array>` to `Uint8Array`). I think it's best if we don't change that function's name so I named this one to `streamToNodeBuffer` ...

packages/pds/src/pipethrough.ts
matthieusieben: Well, creating an inline function is one additional object to initialize + garbage collect on every incoming HTTP call. JS is arguably good at creating inline function but I really tried to get the best perfs where I could.

I don't have a bench to back me though...

packages/pds/src/pipethrough.ts
matthieusieben: The idea of this code is to be generic of the `accept-*` header being negotiated.

Consider for example:

```
Accept-Patch: text/example;charset=utf-8, text/example
```

In that case, the `Accept` type could be generalized as such:
```ts
type Accept = [name: string, flags: Record<string, string>]
```

If you think this is too much YAGNI I can make this more straight forward.

What type safety difference are you thinking of ?

packages/oauth/oauth-provider/src/lib/http/stream.ts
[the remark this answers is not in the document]
matthieusieben: As to distinguishing null pointer error, my hypothesis is that when using TypeScript, `TypeError` are more likely invalid input errors (e.g. `new URL('ff')`) rather than actual null pointer errors.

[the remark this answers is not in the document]
matthieusieben: I could use a custom Error class for this particular purpose (because I don't want to throw something that is explicitely either a 415 or a 500). I just find `TypeError` to be handy and consistent with the way some native function work.

PR #2771, opened 2024-08-31
Pinned posts

mozzius:
Behavior on author feed:
- if pinned post is in the first page of results, filter it from its normal position on that page and set it as the first post along with `#reasonPin`
- if pinned post is on subsequent pages, do not filter it, it will appear duplicated
- `viewer.pinned` remains the same in either case, and should be `true`

Behavior in other feeds:
- `viewer.pinned` should reflect `profile.pinnedPost`

Screenshot 2024-08-31 at 23 13 41


gbramorim:
Please  squash tem commits...

Vincentroland (approved):

estrattonbailey:
Ah one thing: if you pin a reply, it only shows up in your posts & replies 🤔

packages/bsky/src/api/app/bsky/feed/getAuthorFeed.ts
```diff
@@ -79,21 +79,47 @@ export const skeleton = async (inputs: {
...
   })
+
+  let items = res.items.map((item) => ({
+    post: { uri: item.uri, cid: item.cid || undefined },
+    repost: item.repost
+      ? { uri: item.repost, cid: item.repostCid || undefined }
+      : undefined,
+  }))
+
+  if (hasPinnedPost && actor.profile?.pinnedPost) {
```
estrattonbailey, line 101: Nit: we're checking `pinnedPost` twice here. Maybe `s/hasPinnedPost/shouldInsertPinnedPost` is more clear? The comment below here could also move up to clarify what we're doing: if we have a pinned post, we only insert and filter it for the first page.

mozzius: yeah agreed, renamed 👍

packages/bsky/src/api/app/bsky/feed/getAuthorFeed.ts
```diff
@@ -79,21 +79,47 @@ export const skeleton = async (inputs: {
...
   })
+
+  let items = res.items.map((item) => ({
+    post: { uri: item.uri, cid: item.cid || undefined },
+    repost: item.repost
+      ? { uri: item.repost, cid: item.repostCid || undefined }
+      : undefined,
+  }))
+
+  if (shouldInsertPinnedPost && actor.profile?.pinnedPost) {
```
estrattonbailey, line 101: Nice! You can lose the additional `actor.profile?.pinnedPost` too 👍

mozzius: nope I cannot - TypeScript gets mad :(

packages/bsky/src/hydration/feed.ts
```diff
@@ -71,7 +71,15 @@ export type ThreadRef = ItemRef & { threadRoot: string }
...
-export type FeedItem = { post: ItemRef; repost?: ItemRef }
+export type FeedItem = {
+  post: ItemRef
+  repost?: ItemRef
+  /**
+   * If true, overrides the `reason` with `app.bsky.feed.defs#reasonPin`. Used
+   * only in author feeds.
+   */
+  authorPinned?: boolean
+}
```
mozzius, line 82: This is the part I'm least sure about - is it ok to add this extra field here?

packages/bsky/src/api/app/bsky/feed/getAuthorFeed.ts
```diff
@@ -79,21 +79,47 @@ export const skeleton = async (inputs: {
...
+    const pinnedItem = {
+      post: {
+        uri: actor.profile.pinnedPost.uri,
+        cid: actor.profile.pinnedPost.cid,
+      },
+      repost: undefined,
+      authorPinned: true,
+    }
```
mozzius, line 109: I set `authorPinned` here, I'm not expecting the value to come from the dataplane or anything

lexicons/app/bsky/feed/defs.json
```diff
@@ -186,6 +194,10 @@
         "repost": { "type": "string", "format": "at-uri" }
       }
     },
+    "skeletonReasonPin": {
```
dholms, line 197: I actually like that this could be used in the future for a feed gen to "pin" a post to the top of a feed

lexicons/app/bsky/feed/defs.json
```diff
@@ -44,7 +44,8 @@
         "like": { "type": "string", "format": "at-uri" },
         "threadMuted": { "type": "boolean" },
         "replyDisabled": { "type": "boolean" },
-        "embeddingDisabled": { "type": "boolean" }
+        "embeddingDisabled": { "type": "boolean" },
+        "pinned": { "type": "boolean" }
```
dholms, line 48: Just curious, does this affect the presentation of a post elsewhere in the app (other than a user's own author feed) when viewing their own pinned post?

packages/bsky/src/api/app/bsky/feed/getAuthorFeed.ts
```diff
@@ -79,21 +79,47 @@ export const skeleton = async (inputs: {
...
+      : undefined,
+  }))
+
+  if (shouldInsertPinnedPost && actor.profile?.pinnedPost) {
+    const pinnedItem = {
+      post: {
+        uri: actor.profile.pinnedPost.uri,
+        cid: actor.profile.pinnedPost.cid,
+      },
+      repost: undefined,
```
dholms, line 107: (kinda a nit but) if you annotate line 94 as `let items: FeedItem[]` then you can delete this line, which i think clarifies & cleans it up a bit

packages/bsky/src/hydration/feed.ts
dholms: Yeah this type of a bit funny and could honestly probably find a cozier spot. This type is only used as a _parameter_ for both hydration & views. So you should be good to add this extra property here. The lil note is appreciated 👍

lexicons/app/bsky/feed/defs.json
mozzius: We use it to show "Pin post" or "Unpin post" in the context menu

packages/bsky/src/views/index.ts
```diff
@@ -1128,6 +1139,16 @@ export class Views {
     return true
   }
 
+  viewerPinned(uri: string, state: HydrationState) {
+    const authorDid = creatorFromUri(uri)
```
dholms, line 1143: This function fully parses the at-uri under the hood before returning the authorDid. It's a bit unfortunate to have to do a second parse for every single post uri that we hydrate from now on. 

it's probably not a huge deal, but I do wonder if there's a way around it.

Ideas that come to mind are: 
- doing a simple string comparison instead (`uri.includes(state.ctx.viewer)`) for instance
- piping through the post creator from the `post()` view function where it's already been parsed out so we don't have to do it again down here

packages/bsky/tests/views/author-feed.test.ts
```diff
@@ -386,6 +388,142 @@ describe('pds author feed views', () => {
...
+
+      const newProfile: AppBskyActorProfile.Record = {
+        ...profile,
+        pinnedPost: {
+          uri: post.ref.uriStr,
+          cid: post.ref.cid.toString(),
+        },
+      }
+
+      await pdsAgent.com.atproto.repo.putRecord(
```
dholms, line 415: there's a nice seed client function for helping with this as well:
```ts
sc.updateProfile(alice, newProfile)
``

File: packages/bsky/tests/views/author-feed.test.ts [hunk 10]
```diff
@@ -386,6 +388,142 @@ describe('pds author feed views', () => {
...
+        (item) => item.post.uri === post.ref.uriStr,
+      )
+      expect(laterPinnedPost?.post?.uri).toEqual(post.ref.uriStr)
+      expect(laterPinnedPost?.post?.viewer?.pinned).toBeTruthy()
+      expect(AppBskyFeedDefs.isReasonPin(laterPinnedPost?.reason)).toBeFalsy()
+      expect(forSnapshot(page2.feed)).toMatchSnapshot()
+    })
+
+    it('params.includePins = false', async () => {
+      const post = await createAndPinPost()
```
dholms, line 508: would be nice to create one other post here as well just to ensure that it doesn't still stick it at the top of the page

packages/bsky/src/views/index.ts
mozzius: Yeah this makes sense, I think passing it in from `post()` makes the most sense

dholms (approved):
looks great!

i'll toss a changeset in here and get it merged 👌

packages/bsky/src/data-plane/server/db/migrations/20240831T134810923Z-pinned-posts.ts
```diff
@@ -0,0 +1,17 @@
+import { Kysely } from 'kysely'
```
DanielHe4rt, line 1: The backend isn't using Scylla? What is this kysely about? Just curious

mozzius: this is the Postgres-based reference implementation

PR #2773, opened 2024-09-01
PDS proxy to appview performance

dholms:
Introduces a temporary measure where a PDS can proxy directly through to the appview without signing a service auth header. This is mean to reduce load on Bluesky PDS instances, and is not intended to be implemented by other Appviews in the network. While this is a bit awkward, it is directionally aligned with possible plans to have the client talk directly to the Appview.


packages/bsky/src/auth-verifier.ts
```diff
@@ -103,6 +115,17 @@ export class AuthVerifier {
           credentials: { type: 'standard', iss, aud },
         }
       } else if (isBearerToken(ctx.req)) {
+        // @NOTE temporarily accept entryway session tokens to shed load from PDS instances
+        const token = bearerTokenFromReq(ctx.req)
+        const header = token ? jose.decodeProtectedHeader(token) : undefined
+        if (header?.typ === 'at+jwt') {
```
matthieusieben, line 121: Love it that the `typ` header is coming in handy :-)

packages/bsky/tests/entryway-auth.test.ts
```diff
@@ -0,0 +1,153 @@
...
+  const privKeyRaw = await keypair.export()
+  const privKeyEncoded = keyEncoder.encodePrivate(
+    ui8.toString(privKeyRaw, 'hex'),
+    'raw',
+    'pem',
+  )
+  return nodeCrypto.createPrivateKey(privKeyEncoded)
+}
+
+describe('entryway auth', () => {
```
devinivy, line 24: Testing expiration could be nice too!

devinivy (approved):

Vincentroland (approved):

Vincentroland (approved):

packages/bsky/src/auth-verifier.ts
Vincentroland: Expilct all

packages/bsky/src/auth-verifier.ts
```diff
@@ -182,6 +205,51 @@ export class AuthVerifier {
...
+          'InvalidToken',
+        )
+      })
+
+    const { sub, aud, scope } = res.payload
+    if (typeof sub !== 'string' || !sub.startsWith('did:')) {
+      throw new AuthRequiredError('Malformed token', 'InvalidToken')
+    } else if (
+      typeof aud !== 'string' ||
+      !aud.startsWith('did:web') ||
```
matthieusieben, line 236: nit (?) the method name should probably not be alterable

```suggestion
      !aud.startsWith('did:web:') ||
```

PR #2783, opened 2024-09-04
Appview: retry dataplane requests on abort

devinivy:
In the appview we properly retry requests when the dataplane is partially offline, but have been failing to retry in-flight requests that have been aborted.


dholms (approved):
excellent 👌

PR #2784, opened 2024-09-04
Cache repo revs in order to avoid read-after-write for up-to-date responses

matthieusieben:


Vincentroland (approved):

Vincentroland (approved):

Vincentroland (approved):

Vincentroland (approved):

packages/dev-env/src/seed/client.ts
```diff
@@ -548,6 +548,15 @@ export class SeedClient<
     return result.data
   }
 
+  pdsAgent(did: string) {
```
dholms, line 551: slick addition 👌

packages/pds/src/account-manager/repo-rev-cache-redis.ts
```diff
@@ -0,0 +1,48 @@
...
+import Redis from 'ioredis'
+
+import { redisLogger } from '../logger'
+
+const key = (did: string) => `latestRev:${did}`
+
+export class RepoRevCacheRedis implements SimpleStore<string, string> {
+  /**
+   * @param redis - Redis client
+   * @param maxAge - Maximum age of a cached revision in milliseconds
```
dholms, line 11: just noting that this is called `maxAge` here but `px` in the constructor

dholms:
Overall looking good!

Importantly, we need to update the cache value when we do a write (createRecord, putRecord, deleteRecord, applyWrites). Otherwise we will likely be doing comparisons on stale values on read

packages/pds/src/account-manager/index.ts
```diff
@@ -187,7 +189,8 @@ export class AccountManager
   }
 
   async updateRepoRoot(did: string, cid: CID, rev: string) {
-    return repo.updateRoot(this.db, did, cid, rev)
+    await repo.updateRoot(this.db, did, cid, rev)
+    await this.repoRevCache?.set(did, rev)
```
dholms, line 193: may be good to fail open on this rather than returning an error to the user

packages/pds/src/read-after-write/util.ts
```diff
@@ -45,9 +45,23 @@ export const readAfterWriteInternal = async <T>(
   res: HandlerPipeThrough,
   munge: MungeFn<T>,
 ): Promise<HandlerResponse<T> | HandlerPipeThrough> => {
-  const rev = getRepoRev(res.headers ?? {})
+  const rev = getRepoRev(res.headers)
   if (!rev) return res
+
+  // If the response's "atproto-repo-rev" header matches the current repo rev,
+  // we can skip the munge step and return the response as-is.
+  const repoRev = await ctx.repoRevCache?.get(requester)
```
dholms, line 53: similarly, it'd be great to fail open here, maybe with a quick timeout on the redis request

packages/pds/src/account-manager/index.ts
dholms: oh rip saw you handle these errors in the simple store 😛

packages/pds/src/config/env.ts
```diff
@@ -119,6 +119,9 @@ export const readEnv = (): ServerEnvironment => {
 
     // fetch
     fetchDisableSsrfProtection: envBool('PDS_DISABLE_SSRF_PROTECTION'),
+
+    // Repo revision cache
+    repoRevCacheMaxAge: envInt('PDS_REPO_REV_CACHE_MAX_AGE'), // Seconds (use 0 to disable)
```
devinivy, line 124: To make the PDS config more uniform I recommend we switch this to be `_MAX_TTL` and to accept a value in milliseconds.

dholms (approved):
yup 👍

matthieusieben:
Dropping since it didn't produce the expected perf improvements.

PR #2792, opened 2024-09-05
Preferences caching

dholms:


packages/pds/src/config/env.ts
```diff
@@ -120,8 +120,9 @@ export const readEnv = (): ServerEnvironment => {
     // fetch
     fetchDisableSsrfProtection: envBool('PDS_DISABLE_SSRF_PROTECTION'),
 
-    // Repo revision cache
+    // caching
     repoRevCacheMaxTTL: envInt('PDS_REPO_REV_CACHE_MAX_TTL'), // milliseconds (use 0 to disable)
+    preferencesCacheMaxTTL: envInt('PDS_REPO_REV_CACHE_MAX_TTL'), // milliseconds (use 0 to disable)
```
devinivy, line 125:
```suggestion
    preferencesCacheMaxTTL: envInt('PDS_PREFS_CACHE_MAX_TTL'), // milliseconds (use 0 to disable)
```

packages/pds/src/api/app/bsky/actor/getPreferences.ts
```diff
@@ -7,6 +7,16 @@ export default function (server: Server, ctx: AppContext) {
...
       const requester = auth.credentials.did
+
+      const cachedPrefs = await ctx.preferencesCache?.get(requester)
+      if (cachedPrefs) {
+        const preferences = JSON.parse(cachedPrefs)
+        return {
+          encoding: 'application/json',
+          body: { preferences },
+        }
+      }
```
devinivy, line 18: We still want to apply the scope here.

packages/pds/src/api/app/bsky/actor/putPreferences.ts
```diff
@@ -18,13 +18,19 @@ export default function (server: Server, ctx: AppContext) {
...
           checkedPreferences,
           'app.bsky',
           auth.credentials.scope,
         )
       })
+
+      await ctx.preferencesCache?.set(
+        requester,
+        JSON.stringify(checkedPreferences),
+      )
```
devinivy, line 33: I believe this could cache different contents depending on the user's scope.  Maybe it would be more straightforward to control the cache contents entirely on the read path as a read-through cache, and turn this into an invalidation.

PR #2793, opened 2024-09-05
:sparkles: Optionally ack all open subjects from the author with takedown event

foysalit:
This PR adds a `acknowledgeAllSubjectsOfAccount` flag to takedown event which allows moderators to acknowledge all subjects in reviewOpen/reviewEscalated/Appealed state from the account being takendown.

This helps clean up moderation queue of unnecessary subjects that won't require individual reviews if the author's account is already taken down permanently.

An additional but unrelated feature in this PR is a `forAccount` param in `tools.ozone.moderation.queryStatuses` which allows moderators to view all subject statuses for a specific account.

### DIFF
--- a/lexicons/tools/ozone/moderation/defs.json
+++ b/lexicons/tools/ozone/moderation/defs.json
@@ -210,6 +210,10 @@
         "durationInHours": {
           "type": "integer",
           "description": "Indicates how long the takedown should be in effect before automatically expiring."
+        },
+        "acknowledgeAllSubjectsOfAccount": {
+          "type": "boolean",
+          "description": "Set to true if all subjects authored by the account should be resolved."
         }
       }
     },
--- a/lexicons/tools/ozone/moderation/defs.json
+++ b/lexicons/tools/ozone/moderation/defs.json
@@ -211,7 +211,7 @@
           "type": "integer",
           "description": "Indicates how long the takedown should be in effect before automatically expiring."
         },
-        "acknowledgeAllSubjectsOfAccount": {
+        "acknowledgeAccountSubjects": {
           "type": "boolean",
           "description": "If true, all other reports on content authored by this account will be resolved (acknowledged)."
         }
--- a/lexicons/tools/ozone/moderation/defs.json
+++ b/lexicons/tools/ozone/moderation/defs.json
@@ -213,7 +213,7 @@
         },
         "acknowledgeAllSubjectsOfAccount": {
           "type": "boolean",
-          "description": "Set to true if all subjects authored by the account should be resolved."
+          "description": "If true, all other reports on content authored by this account will be resolved (acknowledged)."
         }
       }
     },
--- a/lexicons/tools/ozone/moderation/queryStatuses.json
+++ b/lexicons/tools/ozone/moderation/queryStatuses.json
@@ -8,15 +8,14 @@
       "parameters": {
         "type": "params",
         "properties": {
-          "forAccount": {
-            "type": "string",
-            "format": "did",
-            "description": "All subjects belonging to this account will be returned. If provided, this will take priority over 'subject' parameter."
+          "includeAllUserRecords": {
+            "type": "boolean",
+            "description": "All subjects belonging to the account specified in the 'subject' param will be returned."
           },
           "subject": {
             "type": "string",
             "format": "uri",
-            "description": "The subject to get the status for. If provided along with 'forAccount', this will be ignored."
+            "description": "The subject to get the status for."
           },
           "comment": {
             "type": "string",
@@ -69,18 +68,12 @@
           "sortField": {
             "type": "string",
             "default": "lastReportedAt",
-            "enum": [
-              "lastReviewedAt",
-              "lastReportedAt"
-            ]
+            "enum": ["lastReviewedAt", "lastReportedAt"]
           },
           "sortDirection": {
             "type": "string",
             "default": "desc",
-            "enum": [
-              "asc",
-              "desc"
-            ]
+            "enum": ["asc", "desc"]
           },
           "takendown": {
             "type": "boolean",
@@ -117,9 +110,7 @@
         "encoding": "application/json",
         "schema": {
           "type": "object",
-          "required": [
-            "subjectStatuses"
-          ],
+          "required": ["subjectStatuses"],
           "properties": {
             "cursor": {
               "type": "string"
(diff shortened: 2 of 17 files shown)


lexicons/tools/ozone/moderation/defs.json
```diff
@@ -210,6 +210,10 @@
         "durationInHours": {
           "type": "integer",
           "description": "Indicates how long the takedown should be in effect before automatically expiring."
+        },
+        "acknowledgeAllSubjectsOfAccount": {
+          "type": "boolean",
+          "description": "Set to true if all subjects authored by the account should be resolved."
```
bnewbold, line 216: "If true, all other reports on content authored by this account will be resolved (acknowledged)"

lexicons/tools/ozone/moderation/queryStatuses.json
```diff
@@ -8,7 +8,16 @@
       "parameters": {
         "type": "params",
         "properties": {
-          "subject": { "type": "string", "format": "uri" },
+          "forAccount": {
```
bnewbold, line 11: kind of feels like this should be a boolean? named `allAccountSubjects` or something. you could give either a DID as `subject`, or an AT-URI. either way all subjects for the account would be returned if this was true.

lexicons/tools/ozone/moderation/defs.json
```diff
@@ -210,6 +210,10 @@
         "durationInHours": {
           "type": "integer",
           "description": "Indicates how long the takedown should be in effect before automatically expiring."
+        },
+        "acknowledgeAllSubjectsOfAccount": {
```
bnewbold, line 214: "acknowledgeReportedContent"? hrm

foysalit: acknowledgeAccountSubjects` or `closeUnreviewedAccountSubjects`?

I don't like the verbosity either but `reportedContent` doesn't feel quite right and doesn't line up with the language used in other places. `

lexicons/tools/ozone/moderation/queryStatuses.json
foysalit: makes sense. went with the term `includeAllUserRecords` since we're using that for queryEvents.

packages/ozone/tests/ack-all-subjects-of-account.test.ts
```diff
@@ -0,0 +1,117 @@
...
+    statusesAfter.forEach((item) => {
+      if (item.subject.uri === postOne.uriStr) {
+        expect(item.reviewState).toBe(REVIEWCLOSED)
+      } else if (item.subject.uri === postTwo.uriStr) {
+        expect(item.reviewState).toBe(REVIEWCLOSED)
+        expect(item.appealed).toBe(false)
+      } else if (!item.subject.uri && item.subject.did === sc.dids.bob) {
+        expect(item.reviewState).toBe(REVIEWCLOSED)
+      }
+    })
```
devinivy, line 115: If `includeAllUserRecords` isn't working properly, these lists could be empty and cause the test to trivially pass.  Potentially could select each item out of the array with something like:
```ts
const getItemBySubject = (statuses, match: { uri?: string, did?: string }) => {
  return statuses.find((item) => {
    if (match.did) {
      return !item.subject.uri && item.subject.did === match.did
    }
    return item.subject.uri === match.uri
  })
}
```

foysalit: ahh I see it. refactored to use maps

lexicons/tools/ozone/moderation/defs.json
bnewbold: I'm fine with both of those. I'd stick with `acknowledge` because that is the term we are using elsewhere and consistency is important. So `acknowledgeAccountSubjects`.

packages/ozone/src/api/moderation/emitEvent.ts
```diff
@@ -162,6 +162,10 @@ const handleModerationEvent = async ({
       }
     }
 
+    if (isTakedownEvent && result.event.meta?.acknowledgeAccountSubjects) {
+      await moderationTxn.resolveSubjectsForAccount(subject.did, createdBy)
+    }
```
devinivy, line 167: My biggest question about this work is: should the flag be tied specifically to the takedown event?  It becomes a bit of a complicated idea, feels like the API reflects a fairly specific workflow, and we could probably peel these ideas apart without sacrificing much.  As a bulk action, the failure modes of the acknowledgement are also pretty different from the takedown.

Since this deals with acknowledgement, I'd propose to instead add a flag to `#modEventAcknowledge`, e.g. something like `event.applyToAllRepoSubjects`.  And then the UI would perform this bulk acknowledgement after successfully performing the takedown.

Any thoughts/feels?

foysalit: I can totally see how adding the flag to ack event is useful and helps us decouple this from a particular event and re-use in other cases. However, I am not sure if we have such cases where we'd want to resolve all subjects without review outside of an account level takedown?

there's also this concern around surfacing these in "mod history" and while we should be able to get away with unreviewed "ack" on records if the author account is takendown but doing this in other cases might raise questions to the viewer around "why reports were acked without action on the subject".

I don't feel too strongly about either direction tbh but I like the restriction that only takedown allows us to do this seemingly aggressive ack without review.

packages/ozone/src/mod-service/index.ts
```diff
@@ -720,13 +774,19 @@ export class ModerationService {
...
+        subjectInfo.did,
+      )
+
+      if (!includeAllUserRecords) {
+        builder = builder.where((qb) =>
           subjectInfo.recordPath
             ? qb.where('recordPath', '=', subjectInfo.recordPath)
             : qb.where('recordPath', '=', ''),
         )
+      }
```
devinivy, line 789: When `includeAllUserRecords` is `true` perhaps we want `qb.where('recordPath', '!=', '')` so that we only get records back.  (If it's an issue it's small, wouldn't block on this.)

devinivy (approved):

packages/ozone/src/mod-service/index.ts
foysalit: hmmm... I do see what you mean but the name of the param doesn't really imply that ONLY records will be returned so I think I would actually expect this to include both user's account entry and record entries to be returned by this.

PR #2797, opened 2024-09-06
Avoid performing a dummy query when performing read queries

matthieusieben:
Does this make sense ?

### DIFF
--- a/.changeset/six-geckos-travel.md
+++ b/.changeset/six-geckos-travel.md
@@ -1,0 +1,5 @@
+---
+"@atproto/pds": patch
+---
+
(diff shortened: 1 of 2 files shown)


devinivy (commented):
I'd want to do a quick survey of how/where `read()` is used, but in general I think this seems sound!  If the process is responsible for [WAL recovery](https://www.sqlite.org/walformat.html#recovery) then the first read query within `fn()` would fail with a locking error then `fn()` would get retried.

Since it's a bit sensitive would recommend @dholms take a peek too.

matthieusieben:
I'll close for now. Let's re-consider if we ever need more perf out of the PDSs

PR #2804, opened 2024-09-09
✨ Track record/account delete and update data in subject status

foysalit:
This PR updates moderation subject status with 3 new properties `recordUpdatedAt` `recordDeletedAt` and `recordStatus`. as account/record events are received, these help moderators query subject statuses with delete/update time range and record status.


packages/ozone/src/lexicon/lexicons.ts
```diff
@@ -10831,6 +10831,24 @@ export const schemaDict = {
             description:
               'Timestamp referencing the first moderation status impacting event was emitted on the subject',
           },
+          recordUpdatedAt: {
+            type: 'string',
+            format: 'datetime',
+            description:
+              'Last update timestamp of the record the subject is associated with',
```
foysalit, line 10838: I feel like the language here is quite convoluted. would love suggestions to make it clearer.

packages/ozone/src/lexicon/lexicons.ts
```diff
@@ -11127,7 +11145,7 @@ export const schemaDict = {
         type: 'object',
         description:
           'Logs account status related events on a repo subject. Normally captured by automod from the firehose and emitted to ozone for historical tracking.',
-        required: ['timestamp', 'status', 'active'],
+        required: ['timestamp', 'active'],
```
foysalit, line 11148: TODO: if this PR doesn't make it into the main branch, we'd wanna port this change over to the parent PR.

packages/ozone/src/mod-service/status.ts
```diff
@@ -126,14 +126,79 @@ const getSubjectStatusForModerationEvent = ({
...
+    }
+
+    if (event.meta?.status === 'deleted') {
+      status.recordDeletedAt = timestamp
+    } else {
+      status.recordUpdatedAt = timestamp
+      // When deactivated accounts are re-activated, we receive the event with just the active flag set to true
+      // so we want to make sure that the recordStatus is not set to an outdated value
+      if (currentStatus?.recordStatus !== 'active' && event.meta?.active) {
+        status.recordStatus = 'active'
```
foysalit, line 171: default for `recordStatus` is null, may be we should fallback to that? although, doing this is maybe helpful if we want to see which accounts `reactivated`.

lexicons/tools/ozone/moderation/defs.json
```diff
@@ -131,6 +131,21 @@
...
+        "recordDeletedAt": {
+          "type": "string",
+          "format": "datetime",
+          "description": "Timestamp referencing when the record the subject is associated with was deleted"
+        },
+        "recordStatus": {
+          "type": "string",
+          "description": "Status of the record the subject is associated with. Statuses are different when the subject references an account vs. a record",
+          "knownValues": ["takendown", "suspended", "deleted", "deactivated"]
+        },
```
devinivy, line 148: This is quite a complex field, I think there are a few potential issues with it:

1. It's called  "record status" but it may contain the hosting status of an account.
2. The hosting statuses of records and accounts are mixed together, which could make it seem like a record can be marked as takendown here, which isn't the case.
3. Potential confusion between the ozone status and the hosting status: the terms "takendown" and "suspended" are also terms used by ozone overlapping with the `takendown` and `suspendedUntil` fields.

Here's one idea: we could combine the three new fields into something more like this (using typescript just to illustrate the concept):
```ts
type SubjectStatusView = {
  // ...
  hosting?: AccountHosting | RecordHosting
}

type AccountHosting = {
  $type: 'tools.ozone.moderation.defs#accountHosting',
  status: 'active' | 'takendown' | 'suspended' | 'deleted' | 'deactivated'
  createdAt: Date
  updatedAt: Date
  deletedAt?: Date
  suspendedAt?: Date
  deactivatedAt?: Date
  takendownAt?: Date
  reactivatedAt?: Date
}

type RecordHosting = {
  $type: 'tools.ozone.moderation.defs#recordHosting'
  status: 'active' | 'deleted'
  createdAt: Date
  updatedAt: Date
  deletedAt?: Date
}
```

This also gives us a way to make certain fields required together if we want, e.g. if `hosting` is present then `hosting.createdAt` can always be present too.

lexicons/tools/ozone/moderation/queryStatuses.json
```diff
@@ -28,6 +28,31 @@
             "format": "datetime",
             "description": "Search subjects reviewed after a given timestamp"
           },
+          "recordDeletedAfter": {
```
devinivy, line 31: My qualm with these fields is just that they are named "record" when they really may apply to any subject, e.g. an account rather than a record.

packages/ozone/src/db/migrations/20240909T205730722Z-add-record-status-to-subject-status.ts
```diff
@@ -0,0 +1,31 @@
...
+    .addColumn('recordStatus', 'varchar')
+    .execute()
+  await db.schema
+    .alterTable('moderation_subject_status')
+    .addColumn('recordDeletedAt', 'varchar')
+    .execute()
+  await db.schema
+    .alterTable('moderation_subject_status')
+    .addColumn('recordUpdatedAt', 'varchar')
+    .execute()
```
devinivy, line 15: I'd recommend distinguishing these as e.g. `hostingStatus`, `hostingDeletedAt`, etc.

packages/ozone/src/lexicon/lexicons.ts
devinivy: Perhaps something along the lines of: "Timestamp at which the record or account was last updated."

packages/ozone/src/mod-service/status.ts
```diff
@@ -126,14 +126,80 @@ const getSubjectStatusForModerationEvent = ({
...
+const getSubjectStatusForRecordEvent = ({
+  event,
+  currentStatus,
+}: {
+  event: ModerationEventRow
+  currentStatus?: ModerationSubjectStatusRow
+}): Partial<ModerationSubjectStatusRow> => {
+  const timestamp = event.meta?.timestamp
+    ? `${event.meta?.timestamp}`
+    : event.createdAt
```
devinivy, line 144: The types here are a little funky to wrangle.  To avoid ending up with a string like `'true'` here, could we go with something more explicit like:
```suggestion
  const timestamp = typeof event.meta?.timestamp === 'string'
    ? event.meta.timestamp
    : event.createdAt
```

packages/ozone/src/mod-service/status.ts
devinivy: I'd recommend reflecting the exact hosting status here (where active is either `null` or `'active'`, slight preference for the latter), and then if we want to keep tract of reactivations we do so with a separate field such as `reactivatedAt`.  That allows the status to not be dependent on past statuses, and if we want to know if something is active we don't need to check for multiple different values.

packages/ozone/src/mod-service/status.ts
```diff
@@ -126,14 +126,80 @@ const getSubjectStatusForModerationEvent = ({
...
+    return status
+  }
+
+  if (event.action === 'tools.ozone.moderation.defs#identityEvent') {
+    const status: Partial<ModerationSubjectStatusRow> = {
+      recordUpdatedAt: timestamp,
+    }
+
+    if (event.meta?.tombstone) {
+      status.recordStatus = 'tombstoned'
```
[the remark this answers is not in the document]
devinivy: (P.S. is this separate from `'deleted'`?  Historically we have used the term "tombstoned" in a couple different ways, and I'm not sure which one is meant here.)

lexicons/tools/ozone/moderation/defs.json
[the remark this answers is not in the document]
foysalit: I guess I was using `record` as a generic term here without the context of atproto records. I'm also not sure about `subject` prefix either since we already have `updatedAt` column on subject, we will end up with `subject.updatedAt` and `subject.subjectUpdatedAt` which feels odd. I think I'm onboard with the term `hosting` just from technical standpoint.

foysalit: this is nice!

packages/ozone/src/mod-service/status.ts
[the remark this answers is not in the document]
foysalit: to my understanding, the `tombstoned` event is only emitted on identity and this is "technically" equivalend as "deleted" but not exactly the same thing.

packages/ozone/src/mod-service/status.ts
foysalit: yeah the `null` value is basically "we haven't tracked any hosting status for this" so I think safer option is to check for actual `deactivated` value for status.

lexicons/tools/ozone/moderation/defs.json
```diff
@@ -619,6 +630,58 @@
...
+      }
+    },
+    "recordHosting": {
+      "type": "object",
+      "required": ["status"],
+      "properties": {
+        "status": {
+          "type": "string",
+          "knownValues": ["takendown", "suspended", "deleted", "deactivated"]
+        },
```
devinivy, line 671: From a hosting perspective individual records can't be deactivated, takendown, or suspended— only accounts can.  They can be deleted, though.

lexicons/tools/ozone/moderation/defs.json
```diff
@@ -619,6 +630,58 @@
...
       }
+    },
+    "accountHosting": {
+      "type": "object",
+      "required": ["status"],
+      "properties": {
+        "status": {
+          "type": "string",
+          "knownValues": ["takendown", "suspended", "deleted", "deactivated"]
+        },
```
devinivy, line 641: I see the value `unknown` used in the implementation—may be worth adding here and to the same field in `#recordHosting`.

devinivy (approved):
Nice!  Let's just add a changeset here before merge.

lexicons/tools/ozone/moderation/defs.json
devinivy: Ah, I see the values I'd expect reflected in the implementation, so this probably just needs to be updated.

foysalit:
this is going to be merged into https://github.com/bluesky-social/atproto/pull/2661 and I'll add the changeset there since that's the one landing on main.

PR #2805, opened 2024-09-09
Add `isFallback` to `getSuggestedFollowsByActor` method

estrattonbailey:
We want to know on the client if the results from this endpoint are actually specific to the actor we passed in as `relativeToDid` or not. This endpoint is used on Profile screens and nowhere else.

### DIFF
--- a/lexicons/app/bsky/graph/getSuggestedFollowsByActor.json
+++ b/lexicons/app/bsky/graph/getSuggestedFollowsByActor.json
@@ -24,6 +24,11 @@
                 "type": "ref",
                 "ref": "app.bsky.actor.defs#profileView"
               }
+            },
+            "isFallback": {
+              "type": "boolean",
+              "description": "If true, response has fallen-back to generic results, and is not scoped using relativeToDid",
+              "default": false
             }
           }
         }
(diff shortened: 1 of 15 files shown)


lexicons/app/bsky/graph/getSuggestedFollowsByActor.json
```diff
@@ -24,6 +24,11 @@
                 "type": "ref",
                 "ref": "app.bsky.actor.defs#profileView"
               }
+            },
+            "isFallback": {
```
dholms, line 28: I think I'm inclined to use a positive here rather than a negative. Like `isPersonalized` which would be set to `true` if so and `false` if not. 

Couple reasons:
- generally, i like to avoid negative variable names when there's not a good reason not to
- specific to this case, i'm trying to think of the baseline functionality here & make that the default. It seems more natural to me that the baseline would be not-personalized and the "special case" is personalized

dholms: Not overly concerned about this fwiw if you disagree or if there might be compatibility issues.

dholms (approved):
Looking good besides the note about maybe switching `isFallback` to `isPersonalized`

lexicons/app/bsky/graph/getSuggestedFollowsByActor.json
estrattonbailey: Good points here. This endpoint's baseline actually is personalized though, and the special case is to fall back to generic suggestions if Why can't compute cluster-based results.

Given that, how do you feel about the naming?

dholms: (followed up in DMs, let's stick with `isFallback`)

PR #2817, opened 2024-09-14
Add fast path skipping grapheme counting

gaearon:
Not a huge deal but since we visit this codepath for every single validated string, seems like we'd want to avoid any unnecessary computation here, especially on low-end devices.

I don't know how encodings and grapheme work very well so this needs a close look.

## Commits

- https://github.com/bluesky-social/atproto/pull/2817/commits/0d6b54d09a5674bd689d3b7d7df14d5448a541ec
- https://github.com/bluesky-social/atproto/pull/2817/commits/7b973823bad0759970307d95f278e5cb27c02874
- https://github.com/bluesky-social/atproto/pull/2817/commits/8b0bdcb0c1ebc03dc7118f2135ca8d09eacd0611?w=1

## Before

This was showing up in a profile for switching tabs with 6x throttling. (Note this is a DEV profile so in PROD its % would be higher compared to other things around it.)

Screenshot 2024-09-14 at 16 23 43

## After

It doesn't show up.

Screenshot 2024-09-14 at 16 30 34

### DIFF
--- a/packages/lexicon/src/validators/primitives.ts
+++ b/packages/lexicon/src/validators/primitives.ts
@@ -197,75 +197,89 @@ export function string(
     }
   }
 
-  // Lazily calculated and reused between checks.
-  let cachedUtf8Len: number | undefined
-  let cachedGraphemeLen: number | undefined
+  // maxLength and minLength
+  if (typeof def.maxLength === 'number' || typeof def.minLength === 'number') {
+    const len = utf8Len(value)
 
-  // maxLength
-  if (typeof def.maxLength === 'number') {
-    const len = cachedUtf8Len ?? (cachedUtf8Len = utf8Len(value))
-    if (len > def.maxLength) {
-      return {
-        success: false,
-        error: new ValidationError(
-          `${path} must not be longer than ${def.maxLength} characters`,
-        ),
-      }
-    }
-  }
-
-  // minLength
-  if (typeof def.minLength === 'number') {
-    const len = cachedUtf8Len ?? (cachedUtf8Len = utf8Len(value))
-    if (len < def.minLength) {
-      return {
-        success: false,
-        error: new ValidationError(
-          `${path} must not be shorter than ${def.minLength} characters`,
-        ),
-      }
-    }
-  }
-
-  // maxGraphemes
-  if (typeof def.maxGraphemes === 'number') {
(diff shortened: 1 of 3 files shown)


pfrazee (approved):
Yeah good call

packages/lexicon/src/validators/primitives.ts
```diff
@@ -197,9 +197,14 @@ export function string(
     }
   }
 
+  // Lazily calculated and reused between checks.
+  let cachedUtf8Len: number | undefined
```
bnewbold, line 201: The "caching" logic could maybe be more readable like:

```typescript
if (typeof def.maxLength ==='number' || typeof def.minLength ==='number') {
    const len = utf8Len(value)
    if (typeof def.maxLength ==='number') {
        if (len > def.maxLength) {
            ....
        }
    }
    if (typeof def.minLength ==='number') {
        ...
    }
}
```
and similar for the grapheme value.

but not sure what idiomatic typescript style would be here. not a big deal either way.

packages/lexicon/src/validators/primitives.ts
```diff
@@ -223,25 +229,45 @@ export function string(
...
   // maxGraphemes
   if (typeof def.maxGraphemes === 'number') {
-    if (graphemeLen(value) > def.maxGraphemes) {
-      return {
-        success: false,
-        error: new ValidationError(
-          `${path} must not be longer than ${def.maxGraphemes} graphemes`,
-        ),
+    if (value.length <= def.maxGraphemes) {
+      // If the JavaScript string length is within the maximum limit,
```
bnewbold, line 233: might be worth mentioning here (and below) that JavaScript string length is UTF-16, just to be super redundant about that, because it is confusing to many devs.

bnewbold (approved):
behavior and overall approach looks good to me! this particular set of optimizations is on a relatively hot path and improves performance in the common case and feels worthwhile.

we'll probably want to do the same thing in the go implementation, and can copy these test cases (maybe make them "shared interop test vectors") when we do.

packages/lexicon/src/validators/primitives.ts
gaearon: ah yea that's much nicer

gaearon:
Pushed style tweaks per review. Have not verified that it's equivalent.

PR #2819, opened 2024-09-14
Add fast path for UTF8 length counting

gaearon:
## Commits

- https://github.com/bluesky-social/atproto/commit/e3dabaf9bb0960b9b290dfd9f10d68a6c930997e
- https://github.com/bluesky-social/atproto/commit/9337c2b122b0626e7565634e56747fba49f84003?w=1

## What

Similar to https://github.com/bluesky-social/atproto/pull/2817, I'm trying to avoid calling into `TextEncoder().encode(str).byteLength` for every string. After this change, I basically don't hit it in the app at all — the fast path always lets me out early.

The fast pass itself is pretty general. The idea is that `.length` counts UTF-16 code units, and each UTF-16 code unit corresponds to at most 3 bytes in UTF-8 encoding. So we can safely use `value.length * 3` as an upper bound on what `utf8Len(value)` could possibly be. If this upper bound is below the `minLength`, the same is true for `utf8Len`. If this upper bound is within `maxLength`, the same is true for `utf8Len`. This generalizes to the `''` case with `minLength = 1` too.

Why `* 3`?

- Codepoints fit into a single UTF-16 code unit become 1 to 3 bytes in UTF-8. (Worst case is 3x.)
- Codepoints that need two UTF-16 code units become 4 bytes in UTF-8. (Worst case is 2x.)

So `.length * 3` should always give us a valid upper bound. But this needs a look from an expert.

I've added some test cases.

### DIFF
--- a/packages/lexicon/src/validators/primitives.ts
+++ b/packages/lexicon/src/validators/primitives.ts
@@ -197,9 +197,14 @@ export function string(
     }
   }
 
+  // Lazily calculated and reused between checks.
+  let cachedUtf8Len: number | undefined
+  let cachedGraphemeLen: number | undefined
+
   // maxLength
   if (typeof def.maxLength === 'number') {
-    if (utf8Len(value) > def.maxLength) {
+    const len = cachedUtf8Len ?? (cachedUtf8Len = utf8Len(value))
+    if (len > def.maxLength) {
       return {
         success: false,
         error: new ValidationError(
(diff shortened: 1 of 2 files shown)


matthieusieben:
Does that `* 3` hypothesis also hold for surrogate pairs (used, amongst other things, by emojis) ?

bnewbold:
this seems reasonable, though I should probably re-read more carefully and maybe cook up more corner-cases. I kind of suspect that it won't be as much of a win as the earlier grapheme cluster and utf8 caching patch though? I guess UTF-16 to UTF-8 does cost something through, and this probably does help with the happy path, and we do a lot of these, hrm.

devinivy (approved):
Good thinkin!  Re: the factor of 3 in here, I am quite sure that checks out.

gaearon:
Alright, had to rebase to account for stylistic changes in https://github.com/bluesky-social/atproto/pull/2817, but let's land this. Re: perf impact, in React Native calling into encoder/decoder goes between JS and native so it's not guaranteed to be super cheap and it would just be nice to not worry about it for the common case.

PR #2825, opened 2024-09-23
Align OAuth 2FA code handling with social-app

mozzius:
Fixes https://github.com/bluesky-social/social-app/issues/5448

The email OTP code we send is in lowercase for some reason - I'll try and fix that as well, but in social-app it tries some basic corrections on the code in case someone types it by hand. This PR uses the same function that social-app uses, which uppercases it and adds the dash in the middle if it's not present.

> [!NOTE]
> social-app uses the regex `/^[A-Z2-7]{5}-[A-Z2-7]{5}$/`, whereas I found oauth was using the slightly more lax `/^[A-Z0-9]{5}-[A-Z0-9]{5}$/`. I replaced it with social-app's one, I hope that's ok?

### DIFF
--- a/packages/oauth/oauth-provider/src/assets/app/components/sign-in-form.tsx
+++ b/packages/oauth/oauth-provider/src/assets/app/components/sign-in-form.tsx
@@ -58,6 +58,7 @@ export type SignInFormProps = Override<
     secondFactorPattern?: string
     secondFactorFormat?: string
     secondFactorHint?: string
+    secondFactorParseValue?: (value: string) => string | false
 
     rememberVisible?: boolean
     rememberDefault?: boolean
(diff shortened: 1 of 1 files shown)


damiensedgwick:
Just so you know, for most browsers this would usually have been fine but this appears to be a very long standing Webkit bug!

mozzius:
Oh huh, I didn't realise!

matthieusieben (approved):

packages/oauth/oauth-provider/src/assets/app/components/sign-in-form.tsx
```diff
@@ -155,7 +157,15 @@ export function SignInForm({
       if (secondFactor) {
         const element = event.currentTarget.secondFactor
         if (!element) throw new Error('Second factor input not found')
-        credentials[secondFactor.type] = element.value
+        const value = secondFactorParseValue(element.value)
+        if (!value) {
+          setSecondFactor({
+            type: secondFactor.type,
+            hint: `Make sure to matches the format: ${secondFactorFormat}`,
```
surfdude29, line 164:
```suggestion
            hint: `Make sure to match the format: ${secondFactorFormat}`,
```

PR #2828, opened 2024-09-24
bsky: Add caller did in grpc calls to dataplane

matthieusieben:
Adds a `bsky-caller-did` header to the GRPC calls to the dataplane that contains the `did` of the authenticated user making the inbound HTTP request.

Does so by creating a new `HydrateCtx` per HTTP request that contains a wrapped dataplane instance (that will add the header), and that is used from controllers (including pipelines).


devinivy:
This is a much simpler change— could we peel it out of this PR so that we can ship it separately? https://github.com/bluesky-social/atproto/pull/2828/commits/e21ada77e340ce7abaed556b94c137dbf2b96317

packages/bsky/src/context.ts
```diff
@@ -100,6 +132,142 @@ export class AppContext {
...
+  ) {
+    const { authVerifier } = this
+    const {
+      // Store as const to allow code path optimizations by compiler
+      allowInclude3pBlocks = false,
+      allowIncludeTakedowns = false,
+      canPerformTakedownRequired = false,
+      exposeRepoRev = false,
+    } = options
+
```
devinivy, line 200: I have a slight concern around getting rid of all handler function bodies and replacing them with this method, which is that it becomes awkward to add a custom behaviors to a given route.  Instead, all the custom behavior ends-up needing to get expressed within this method, which seems like it will end-up accumulating many special cases over time.

A good example is the `canPerformTakedownRequired` option— this clearly stands out as a behavior of some specific endpoint that performs a takedown, but it ends-up folded into this method next to other unrelated behaviors from all other routes.  I find that `admin.updateSubjectStatus` was easier to reason about and to maintain in its previous state.

Overall it seems more natural to me to place the special behaviors directly where they're need, and instead create utilities if necessary to make the handlers more concise.  I am interested to hear your thoughts on all of this!

matthieusieben: I generally agree with your point of view. Special behaviors should be visible where they are needed. And _ideally_, those special code paths should not impact places where they are not needed.

- The case of `canPerformTakedownRequired` is quite isolated and was easily [reverted back](https://github.com/bluesky-social/atproto/pull/2828/commits/0ed5be4397041ea37502f719c1a056ebbb243381) to the way it was.

- `allowInclude3pBlocks` and `allowIncludeTakedowns` are options needed to create the `HydrateCtx` object, so it is quite natural to put them here IMO.

- `inputHeaders` and `ouputHeaders` were [removed](https://github.com/bluesky-social/atproto/pull/2828/commits/cdb759f2ab6e7936facd4cb260c9879e6929d16b).

- Regarding `exposeRepoRev`, there are lots of alternative solution, some of which are detailed below. In my opinion, it comes down to two things:

  1) This is needed in about 50% of the methods. Having an explicit boolean makes it easier to distinguish omissions from explicit opt-in.
  2) Most options that make this available as a function make the code more verbose & significantly increase repetition.

<details>
<summary>`exposeRepoRev` alternatives</summary>

we could easily re-write it as a `withExposedRepoRev` wrapper function. With this, the following:

```ts
    auth: ctx.authVerifier.optionalStandardOrRole,
    handler: ctx.createPipelineHandler(
      skeleton,
      hydration,
      noBlocksOrMutedReposts,
      presentation,
      {
        allowIncludeTakedowns: true,
        exposeRepoRev: true
      },
    ),
  })
```

would be replaced by:

```ts
  const getAuthorFeed = createPipeline(
    skeleton,
    hydration,
    noBlocksOrMutedReposts,
    presentation,
  )

    auth: ctx.authVerifier.optionalStandardOrRole,
    handler: ctx.createHandler(withExposedRepoRev(getAuthorFeed), {
      allowIncludeTakedowns: true,
    }),
  })
```

I personally think this is not really better in terms of readability. It increases repetition (`ctx.createHandler` + `createPipeline` instead of `ctx.createPipelineHandler`) & makes the code either more "dense" (like above) or more repetitive (if introducing intermediary variables).


A more readable alternative, that would be more aligned with the current implementation, could look like this:

```ts
  const getAuthorFeed = createPipeline(
    skeleton,
    hydration,
    noBlocksOrMutedReposts,
    presentation,
  )

    auth: ctx.authVerifier.optionalStandardOrRole,
    handler: async (reqCtx) => {
      const hCtx = await ctx.createHydrateCtx(reqCtx, {
        allowIncludeTakedowns: true,
      })

      hCtx.exposeLabelers()
      await hCtx.exposeRepoRev()

      return getAuthorFeed(hCtx)
    },
  })
```

I find that this notation has several drawbacks:
- `exposeLabelers()` must basically always be called. This yields a lot of repetition that adds un-necessary noise to the function.
- `exposeRepoRev()` is async and `exposeLabelers()` not. could be error prone (forget to await in some instances)
- As maintainer, it is not clear when a function does not contain `exposeRepoRev()` whether it is intentional or an omission (`{ exposeRepoRev: boolean }` makes this explicit)
- More complex typings are needed to properly carry over the `Param`, `Auth`, `Input` and `Output`

</details>

matthieusieben: Note: I refactored that function to be much more lightweight and easier to use in a raw handler function:

https://github.com/bluesky-social/atproto/blob/7c8774c749b4ffd04b8edf2a09a635a0fe4a7f2d/packages/bsky/src/context.ts#L206-L218

matthieusieben:
Abandoning as too many changes where made on main.

PR #2834, opened 2024-09-27
Default to unencoded responses

matthieusieben:
This enables PDSs to serve un-compressed responses when the request does not contain `accept-encoding`. This is done by specifying `identity` as preferred `content-encoding` when performing the upstream request. If the upstream server did not respect our priority list, the PDS will decode the response before serving it.

If, at some point we would like to reduce the bandwidth (at the cost of increased CPU usage on both the PDS and AppView(s)), we can set `PDS_PROXY_PREFER_COMPRESSED=true`. This will cause the PDS to ask for compressed responses with higher priority than uncompressed (`identity`) payloads.

### DIFF
--- a/packages/common-web/src/util.ts
+++ b/packages/common-web/src/util.ts
@@ -21,7 +21,7 @@ export function omit<
   rejectedKeys: readonly K[],
 ): T extends undefined ? undefined : T extends null ? null : Omit<T, K>
 export function omit(
-  src: Record<string, unknown>,
+  src: undefined | null | Record<string, unknown>,
   rejectedKeys: readonly string[],
 ): undefined | null | Record<string, unknown> {
   // Hot path
--- a/packages/common-web/src/util.ts
+++ b/packages/common-web/src/util.ts
@@ -10,47 +10,33 @@ export const noUndefinedVals = <T>(
 }
 
 /**
- * Creates a copy of the object with the specified keys omitted. If the object
- * is `undefined`, it returns `undefined`. If the object did not have any of the
- * keys to omit, it returns the same object.
+ * Returns a shallow copy of the object without the specified keys. If the input
+ * is nullish, it returns the input.
  */
 export function omit<
-  T extends undefined | Record<string, unknown>,
+  T extends undefined | null | Record<string, unknown>,
   K extends keyof NonNullable<T>,
->(obj: T, keys: readonly K[]): T extends undefined ? undefined : Omit<T, K>
+>(
+  object: T,
+  rejectedKeys: readonly K[],
+): T extends undefined ? undefined : T extends null ? null : Omit<T, K>
 export function omit(
-  obj: Record<string, unknown>,
-  keys: readonly string[],
-): undefined | Record<string, unknown> {
+  src: Record<string, unknown>,
+  rejectedKeys: readonly string[],
+): undefined | null | Record<string, unknown> {
   // Hot path
-  if (obj) {
-    const objKeys = Object.keys(obj)
-    for (let i = 0; i < objKeys.length; i++) {
-      // If the key at position i must be excluded
-      if (keys.includes(keys[i])) {
-        // Create a copy
-        const newObj = {}
 
-        // Copy all the keys already visited (that did not have to be excluded),
-        // without re-checking if they need to be excluded.
-        for (let j = 0; j < i; j++) {
-          const key = objKeys[j]
-          newObj[key] = obj[key]
-        }
+  if (!src) return src
 
-        // Copy all the remaining keys, excluding the ones that need to be
-        // excluded.
-        for (let j = i + 1; j < objKeys.length; j++) {
-          const key = objKeys[j]
-          if (!keys.includes(key)) newObj[key] = obj[key]
-        }
-
-        return newObj
-      }
+  const dst = {}
+  const srcKeys = Object.keys(src)
+  for (let i = 0; i < srcKeys.length; i++) {
+    const key = srcKeys[i]
+    if (!rejectedKeys.includes(key)) {
+      dst[key] = src[key]
     }
   }
-
-  return obj
+  return dst
 }
 
 export const jitter = (maxMs: number) => {
--- a/packages/common-web/src/util.ts
+++ b/packages/common-web/src/util.ts
@@ -9,6 +9,11 @@ export const noUndefinedVals = <T>(
   return obj as Record<string, T>
 }
 
+/**
+ * Creates a copy of the object with the specified keys omitted. If the object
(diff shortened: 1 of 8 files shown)


packages/common-web/src/util.ts
```diff
@@ -17,11 +22,35 @@ export function omit(
   obj: Record<string, unknown>,
   keys: readonly string[],
 ): undefined | Record<string, unknown> {
-  if (!obj) return obj
+  // Hot path
+  if (obj) {
+    const objKeys = Object.keys(obj)
+    for (let i = 0; i < objKeys.length; i++) {
+      // If the key at position i must be excluded
+      if (keys.includes(keys[i])) {
```
devinivy, line 30: This condition looks like it's always true, so long as `keys` isn't empty—I assume we want to check `objKeys[i]` instead.

This change to `omit()` is fairly complicated, partly demonstrated by this little mistake which is easy to make and hard to catch.  Another subtle change here is that previously the caller always received a shallow clone, and now you it may not receive a shallow clone, which changes how a caller might consider handling the result.

In your estimation do you think the perf wins here are going to be measurable for us?  If it's not a win, I wonder if this is a case where we can safely reach for the simpler implementation.  If it is going to be a win, let's go for it and get some tests to document the behavior (e.g. around the shallow copying).

packages/common/src/streams.ts
```diff
@@ -107,9 +107,13 @@ export function decodeStream(
  * Create a series of decoding streams based on the content-encoding header. The
  * resulting streams should be piped together to decode the content.
  */
-export function createDecoders(contentEncoding?: string): Duplex[] {
+export function createDecoders(contentEncoding?: string | string[]): Duplex[] {
   const decoders: Duplex[] = []
 
+  if (Array.isArray(contentEncoding)) {
+    throw new TypeError('content-encoding must be a comma separated string')
+  }
```
devinivy, line 115: Not sure I totally understand this— why do we permit an array in the types if it's unsupported?

packages/pds/src/pipethrough.ts
```diff
@@ -367,80 +418,135 @@ function handleUpstreamRequestError(
...
-  return val != null
+  return [encoding.toLowerCase(), flags]
 }
 
-export function isJsonContentType(contentType?: string): boolean | undefined {
+export function isJsonContentType(
+  contentType?: string | string[],
+): boolean | undefined {
   if (contentType == null) return undefined
+  if (typeof contentType !== 'string') return undefined
```
devinivy, line 549: I don't think `isJsonContentType()` is ever called with a `string[]`— do we need this change?

packages/common/src/streams.ts
devinivy: Oh, is this something about representing duplicate headers?  If so, technically I think the move is supposed to be to combine them per https://datatracker.ietf.org/doc/html/rfc2616#section-4.2.

packages/pds/src/pipethrough.ts
```diff
@@ -93,6 +122,28 @@ export const proxyHandler = (ctx: AppContext): CatchallHandler => {
...
+              pipeline([...dec, res], (_err: Error | null) => {})
+              res.removeHeader('content-encoding')
+              res.removeHeader('content-length')
+              return dec[0]
+            }
+          } catch {
+            // Upstream encoding not supported, return the response as-is. Note
+            // that in this case we could also throw an UpstreamFailure since
+            // the upstream server is misbehaving (it should have returned a 406
+            // if it dod not support any of the default encoding we provided).
```
devinivy, line 143:
```suggestion
            // if it did not support any of the default encoding we provided).
```

packages/pds/src/pipethrough.ts
```diff
@@ -194,13 +233,25 @@ export async function pipethrough(
   }
 
   const upstream = await pipethroughRequest(ctx, dispatchOptions)
+  const encoding =
+    safeString(upstream.headers['content-type']) ?? 'application/json'
+  const headers = Object.fromEntries(responseHeaders(upstream.headers))
+
+  // If no accept-encoded header was provided, we should return a decoded stream
```
devinivy, line 240:
```suggestion
  // If no accept-encoding header was provided, we should return a decoded stream
```

packages/pds/src/pipethrough.ts
```diff
@@ -367,80 +418,135 @@ function handleUpstreamRequestError(
...
-  supported: readonly Accept[],
-): readonly Accept[] {
-  // Optimization: if no accept-encoding header is present, skip negotiation
-  if (!acceptHeader?.length) {
-    return supported
+  preferences: readonly [Accept, ...Accept[]],
+): string {
+  if (!acceptHeader) {
+    // If no accept-encoding header is provided, we will use our own preferences
+    // and decode the response if needed.
```
devinivy, line 433: Would some code paths be simplified if treated a missing accept-encoding by just applying a default of `identity`?  It seems as though we go down different code paths depending on whether `accept-encoding` is missing vs. when it's `identity`, but we end-up trying to generate the same result for the consumer (output is not encoded).

devinivy (commented):
Looking good!  I agree with the change in default behavior to preference no encoding 👍 Just left a few misc thoughts.

packages/common-web/src/util.ts
matthieusieben: I like to optimize by default but in this instance you are right, the efficiency vs complexity is not the best. Simplified accordingly.

packages/common/src/streams.ts
matthieusieben: NodeJS's http server will ensure that this header is never an array:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/e58b2667d22cf8a4ba8a34c2b811b3e911c55583/types/node/http.d.ts#L69C1-L69C49

But Undici doesn't:
https://github.com/nodejs/undici/blob/fbdcdb5fd1ed69a0b8a3fb8444ff0e2ac9755e63/types/header.d.ts#L6

Because of node's implementation I assumed this was not allowed.

I adapted to allow arrays.

packages/pds/src/pipethrough.ts
matthieusieben: Letfover form intermediary changes I think. Removed.

packages/pds/src/pipethrough.ts
matthieusieben: I wanted to allow the PDS to prefer compressed upstream bodies, even if the client requested for identity. Then I realized that we wouldn't want to decode+re-encode so I only supported that when the client did not specify an encoding as this was much simpler to support. You are right though, this is inconsistent and makes the code much harder to read.

I reverted to doing what you suggested.

PR #2836, opened 2024-09-27
✨ Ozone batch repo and record getters

foysalit:
This PR introduces 2 new endpoints on ozone `tools.ozone.moderation.getRecords` and `tools.ozone.moderation.getRepos` in order to allow fetching repo and record details in batches.


packages/ozone/src/api/moderation/getRecord.ts
```diff
@@ -11,18 +11,20 @@ export default function (server: Server, ctx: AppContext) {
       const db = ctx.db
       const labelers = ctx.reqLabelers(req)
 
-      const [record, accountInfo] = await Promise.all([
-        ctx.modService(db).views.recordDetail(params, labelers),
+      const [records, accountInfos] = await Promise.all([
+        ctx.modService(db).views.recordDetail([params], labelers),
```
dholms, line 15: Can we pluralize everything that is taking a list & returning a map?

ie `modService(db).views.recordDetails`

and `getPdsAccountInfos`

packages/ozone/src/api/moderation/getRecords.ts
```diff
@@ -0,0 +1,56 @@
...
+        ),
+        getPdsAccountInfo(
+          ctx,
+          params.uris.map((uri) => new AtUri(uri).hostname),
+        ),
+      ])
+
+      const results: (RecordViewDetail | RecordViewNotFound)[] = []
+
+      params.uris.forEach((uri) => {
```
dholms, line 30: a `map` might be a bit clearer here

`const results = params.uris.map((uri) => ...)`

packages/ozone/src/api/moderation/getRepos.ts
```diff
@@ -0,0 +1,48 @@
...
+      const db = ctx.db
+      const labelers = ctx.reqLabelers(req)
+      const [partialRepos, accountInfo] = await Promise.all([
+        ctx.modService(db).views.repoDetail(dids, labelers),
+        getPdsAccountInfo(ctx, dids),
+      ])
+
+      const repos: (RepoViewDetail | RepoViewNotFound)[] = []
+
+      dids.forEach((did) => {
```
dholms, line 23: same thing here: a `map()` might be a bit clearer

packages/ozone/src/api/util.ts
```diff
@@ -30,17 +30,28 @@ import { ids } from '../lexicon/lexicons'
 
 export const getPdsAccountInfo = async (
   ctx: AppContext,
-  did: string,
-): Promise<AccountView | null> => {
+  did: string | string[],
```
dholms, line 33: I'm inclined to just have this always take a `string[]` if it's always returning a `Map` of account views. Just keeps the function a bit simpler

packages/ozone/src/mod-service/views.ts
```diff
@@ -224,24 +224,33 @@ export class ModerationViews {
   }
 
   async repoDetail(
-    did: string,
+    dids: string | string[],
```
dholms, line 227: similarly, i'm inclined to just have this always take an array

packages/ozone/src/mod-service/views.ts
```diff
@@ -224,24 +224,33 @@ export class ModerationViews {
...
-      ...repo,
-      moderation: {
-        ...repo.moderation,
-      },
-      labels: [...localLabels, ...externalLabels],
-    }
+    repos.forEach((repo, did) => {
+      const labels = [
+        ...(localLabels.get(did) || []),
+        ...externalLabels.filter((label) => label.uri === did),
```
dholms, line 241: I wonder if `getExternalLabels` should return a `Map<string, Label[]>` instead of just a big unfiltered array. That way we don't need to repeatedly scan over the array to get our list of labels for each subject

packages/ozone/src/mod-service/views.ts
```diff
@@ -309,39 +318,46 @@ export class ModerationViews {
...
+      await Promise.all([
+        this.records(subjects),
+        this.getSubjectStatus(subjectUris),
+        this.labels(subjectUris),
+        this.getExternalLabels(subjectUris, labelers),
+      ])
+
+    const results = new Map<string, RecordViewDetail>()
+
+    records.forEach(async (record, uri) => {
```
dholms, line 335: Pretty sure you lose track of the promise like this. I think you either need to use a for loop or `map()` with a `Promise.all`. 

I'd be inclined to do the latter so you can grab the blob infos in parallel

packages/ozone/tests/get-records.test.ts
```diff
@@ -0,0 +1,87 @@
...
+} from '@atproto/dev-env'
+import { AtpAgent } from '@atproto/api'
+import {
+  REASONOTHER,
+  REASONSPAM,
+} from '../src/lexicon/types/com/atproto/moderation/defs'
+import { forSnapshot } from './_util'
+import { ids } from '../src/lexicon/lexicons'
+
+describe('admin get record view', () => {
```
dholms, line 16:
```suggestion
describe('admin get records view', () => {
```

packages/ozone/tests/get-records.test.ts
```diff
@@ -0,0 +1,87 @@
...
+        uri: sc.posts[sc.dids.alice][0].ref.uriStr,
+        cid: sc.posts[sc.dids.alice][0].ref.cidStr,
+      },
+    })
+    await network.bsky.ctx.dataplane.takedownRecord({
+      recordUri: sc.posts[sc.dids.alice][0].ref.uriStr,
+    })
+  })
+
+  it('gets a records by uris', async () => {
```
dholms, line 72:
```suggestion
  it('get some records by uris', async () => {
```

packages/pds/src/account-manager/helpers/account.ts
```diff
@@ -64,6 +64,41 @@ export const getAccount = async (
   return found || null
 }
 
+export const getAccounts = async (
+  db: AccountDb,
+  handlesAndDids: string[],
```
dholms, line 69: since `getAccountInfos` only takes dids, you can probably simplify this by just taking an array of DIDs (no handles)

packages/pds/src/account-manager/helpers/account.ts
```diff
@@ -64,6 +64,41 @@ export const getAccount = async (
...
+    if (handleOrDid.startsWith('did:')) {
+      dids.add(handleOrDid)
+    } else {
+      handles.add(handleOrDid)
+    }
+  }
+
+  const accounts = await selectAccountQB(db, flags)
+    .if(handles.size > 0, (qb) =>
+      qb.where('actor.handle', 'in', Array.from(handles)),
```
dholms, line 85: I'm pretty sure these two `.where`s strung together are going to turn into a `WHERE ... AND ...`

Kysely has an `orWhere` function, but sometimes the grouping around those can be a little bit tricky. Partially why I recommend just ditching the handles 😛

packages/pds/src/account-manager/helpers/account.ts
dholms, line 69: also kysely gets mad if you pass in an empty array, so the first thing I'd do in this function is check if the array length is 0 & if so then just return an empty map

packages/pds/src/account-manager/helpers/invite.ts
```diff
@@ -155,6 +155,30 @@ export const getAccountInviteCodes = async (
   }))
 }
 
+export const getAccountsInviteCodes = async (
+  db: AccountDb,
+  dids: string[],
+): Promise<Map<string, CodeDetail[]>> => {
+  const results = new Map<string, CodeDetail[]>()
+  const res = await selectInviteCodesQb(db)
+    .where('forAccount', 'in', dids)
```
dholms, line 164: like above^^ make sure you don't pass in an empty array

packages/pds/src/account-manager/helpers/invite.ts
```diff
@@ -155,6 +155,30 @@ export const getAccountInviteCodes = async (
   }))
 }
 
+export const getAccountsInviteCodes = async (
```
dholms, line 158: Can dry this up a bit & use this function to implement `getAccountInviteCodes`?

packages/pds/src/api/com/atproto/admin/getAccountInfos.ts
```diff
@@ -0,0 +1,34 @@
...
+          includeTakenDown: true,
+        }),
+        ctx.accountManager.getAccountsInvitesCodes(params.dids),
+        ctx.accountManager.getInvitedByForAccounts(params.dids),
+      ])
+
+      const managesOwnInvites = !ctx.cfg.entryway
+      const infos: AccountView[] = []
+
+      accounts.forEach((account) => {
```
dholms, line 22: `.map()` might be a bit cozier here too

dholms:
Overall looking pretty good! Some style tweaks and what not but nothing too major. 

Main thing that is prone to errors is accidentally passing empty arrays into kysely `where()` clauses

packages/pds/src/account-manager/helpers/invite.ts
foysalit: decided to stick to the pattern of `take an array and return a map` and got rid of the singular `getAccountInviteCodes` in favor of `getAccountsInviteCodes`

dholms (approved):
looking good ✨

PR #2837, opened 2024-09-27
Follow-on cleanup after #2483

lindner:
- Change tests to use AtpClient instead of the deprecated BskyClient

### DIFF
--- a/packages/api/tests/atp-agent.test.ts
+++ b/packages/api/tests/atp-agent.test.ts
@@ -3,7 +3,7 @@ import { TID } from '@atproto/common-web'
 import {
   AppBskyActorDefs,
   AppBskyActorProfile,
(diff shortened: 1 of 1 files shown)


bnewbold:
docker/CI/github have been really struggling recently; I think the CI failures are due to infra issues for now. Will need to loop back after a few hours for the infra issues to settle and retry those runs.

lindner:
Looks like all tests pass.  Looking forward to more PRs :-D

bnewbold:
cc: @devinivy @matthieusieben

matthieusieben (approved):

matthieusieben:
There is so much deprecated stuff that was introduced with that change.. thanks for helping us get rid of it 💙

PR #2840, opened 2024-09-30
Disallow pinning someone else's post

mozzius:
Just a simple check that the post URI includes the user's DID. Includes a new test

### DIFF
--- a/packages/bsky/src/api/app/bsky/feed/getAuthorFeed.ts
+++ b/packages/bsky/src/api/app/bsky/feed/getAuthorFeed.ts
@@ -14,6 +14,7 @@ import {
 import { Views } from '../../../../views'
 import { DataPlaneClient } from '../../../../data-plane'
 import { parseString } from '../../../../hydration/util'
+import { uriToDid } from '../../../../util/uris'
 import { Actor } from '../../../../hydration/actor'
 import { FeedItem, Post } from '../../../../hydration/feed'
 import { FeedType } from '../../../../proto/bsky_pb'
@@ -85,7 +86,7 @@ export const skeleton = async (inputs: {
     isFirstPageRequest &&
     params.includePins &&
     !!actor.profile?.pinnedPost &&
-    actor.profile.pinnedPost.uri.includes(actor.did)
+    uriToDid(actor.profile.pinnedPost.uri) === actor.did
 
   const res = await ctx.dataplane.getAuthorFeed({
     actorDid: did,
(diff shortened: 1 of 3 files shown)


aliceisjustplaying:
This is the kind of thing that makes Bluesky hostile to its users and frankly this makes me feel disgusted. I'm going back to Threads.

futurGH:
This is the kind of thing that makes Bluesky hostile to its users and frankly this makes me feel disgusted. I'm going back to 8chan.

packages/bsky/src/api/app/bsky/feed/getAuthorFeed.ts
```diff
@@ -82,7 +82,10 @@ export const skeleton = async (inputs: {
 
   const isFirstPageRequest = !params.cursor
   const shouldInsertPinnedPost =
-    isFirstPageRequest && params.includePins && !!actor.profile?.pinnedPost
+    isFirstPageRequest &&
+    params.includePins &&
+    !!actor.profile?.pinnedPost &&
+    actor.profile.pinnedPost.uri.includes(actor.did)
```
estrattonbailey, line 88: Any chance of there being a handle in the pinned post URI?

estrattonbailey (approved):
I'm going back to AIM

packages/bsky/src/api/app/bsky/feed/getAuthorFeed.ts
mozzius: I don't believe that if it was a handle in there that it would get resolved during hydration, so it wouldn't be a valid pinned post URI anyway

dholms, line 88: Think this would be better if we either parsed out the actual DID
```ts
new AtUri(actor.profile.pinnedPost.uri).hostname
```

or explicitly checked the start of the uri 

```ts
actor.profile.pinnedPost.uri.startsWith(`at://${actor.did}`)
```

think i prefer the former

estrattonbailey: Nice. We have the `uriToDid` util for this too 👍

wish13yt:
Boo!!! I'm going back to Windows Live Messenger

aliceisjustplaying:
@wish13yt You do have Messenger Plus! installed, right?

wish13yt:
> @wish13yt You do have Messenger Plus! installed, right?

Had. I reinstalled WLM like two days ago and never got around to reinstalling Plus.

elfprince13:
This is the kind of thing that makes Bluesky hostile to its users and frankly this makes me feel disgusted. I'm going back to phpBB.

dholms (approved):
code looks good 👍 

Though I'm a bit worried that this is the kind of thing that makes Bluesky hostile to its users and frankly this makes me feel disgusted. I'm going back to IRC

PR #2841, opened 2024-09-30
Don't replace first post with pin if limit=1

mozzius:
We had weird divergent behaviour if you asked for limit=1 or not where limit=1 would *only* return the pinned post. Since we ask for limit=1 to check to see if there are new posts, we should instead return the pin and the top post here so we can check both


estrattonbailey (approved):

dholms (approved):
yup looks good 👍

PR #2844, opened 2024-09-30
Fix wrong imports in oauth-client-node README

KevSlashNull:
In the `@atproto/oauth-client-node` README code examples, this:

1. adds missing imports,
2. fixes the `NodeOAuthClient` import (because `NodeOAuthClientOptions` is just a type, not a constructor),
3. and adds `async` to some functions in the options, so they actually return `Promise`s.

### DIFF
--- a/packages/oauth/oauth-client-node/README.md
+++ b/packages/oauth/oauth-client-node/README.md
@@ -22,9 +22,10 @@ needs of your application, and must respect the [ATPROTO].
 The `client_metadata` object will typically be built by the backend at startup.
 
 ```ts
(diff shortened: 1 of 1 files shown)


matthieusieben (approved):
Thanks

dholms:
Not exactly sure what the linting error here is, but do you mind running a quick `pnpm format` @KevSlashNull?

surfdude29:
Sorry for jumping in here, I think the linting error might have been in a different README and was fixed in 3c7bf81b60b50653ee2b49f17a1505f0a828cddc ?

KevSlashNull:
@dholms, I think @surfdude29 is right, so I rebased.

PR #2857, opened 2024-10-03
Add check for external media within RecordWithMedia

estrattonbailey:
We had a report a while back that if a link card was attached along with a quoted post, mute words didn't apply to the link card. This fixes that.


haileyok (approved):

haileyok:
didnt test but seems fine, same logic as above checks

PR #2858, opened 2024-10-03
remove interop tests dir

bnewbold:
Moved these over to a separate repo to be more implementation-neutral: https://github.com/bluesky-social/atproto-interop-tests


matthieusieben (commented):
This looks good. I would:

1) Add a comment in every test fixture referencing the reference file from the [atproto-interop-tests](https://github.com/bluesky-social/atproto-interop-tests) repo (for easy tracking)
2) This change make NPM packages bigger. It's probably time we create `.npmignore` files that exclude the `test` directories from being bundled with the package on NPM
3) NIT: Remove the `interop-test-files` reference from `.prettierignore`.

bnewbold:
Removed line from `.prettierignore`, and added reference links to the test repo.

Setting up `.npmignore` is a bit out of my wheelhouse. Eg, do folks definitely never try to run tests on packages themselves? Could you add those in a follow-up PR?

PR #2860, opened 2024-10-04
lex: update bsky schemas to have actual record refs, not 'unknown'

bnewbold:


mary-ext:
I believe the TypeScript lex-cli is currently set up around refs pointing towards `Main` type which wouldn't work as records have their interface named `Record`?

bnewbold:
Talked to @pfrazee some more about this, and to some degree it is about keeping the schemas flexible. We could in theory end up with a "app.bsky.feed.postV2" (or "app.bsky.feed.post.v2" or whatever) iterated record type, and might want to pass that back in the existing view APIs. We could also iterate on the view API overall. My take is that this would be better served with either an open union (instead of "unknown"), and probably add a "view v2" endpoint which supports records of either type. AKA, make the future view APIs backwards compatible, instead of trying to make current view API forwards-compatible. I suspect plenty of folks have implemented clients in ways such that a different record type showing up in a "post view" would result in an error.

PR #2864, opened 2024-10-04
install pnpm through corepack

matthieusieben:


devinivy (approved):
This looks right to me, let's just coordinate landing this so we know when we're doing the next deploys of these updated docker builds.

PR #2869, opened 2024-10-06
Add alg argument

miker2049:
While this shouldn't be necessary in backend/Node environment, jose seems to use its browser package if you are trying to use this in Bun.  The added argument here will be ignored in proper node environments, but also makes it work in Bun.  Clue here: https://github.com/panva/jose/issues/471


matthieusieben:
Covered by: https://github.com/bluesky-social/atproto/pull/2879

Issue #2816, opened 2024-09-14
OAuth token refreshes are brittle


mary-ext:
**Describe the bug**

As in the title, token refreshes are very brittle and is likely to fail and logs the user out if the user happens to be refreshing at the wrong moment.

**To Reproduce**

1. Use the OAuth browser library and sign in
2. Wait for token to expire
3. Make it refresh
4. While refreshing, reload the page.

**Expected behavior**

It shouldn't be very likely for reloads to log the user out.

**Details**

While this is entirely unavoidable, this seems to be mostly happening because the client needs to verify whether the new access tokens are received from the right issuer. 

https://github.com/bluesky-social/atproto/blob/319aa7cf6dd7de0262a40d69f695c9a0eb0b5179/packages/oauth/oauth-client/src/oauth-server-agent.ts#L145

Since this happens in the middle of a request to refresh tokens and actually storing that token, this leaves quite a gap on said tokens getting lost before it makes its way into IndexedDB.

It does not help that the cache expiry time for these identity resolutions are low, by the time a token expiry happen and needs refreshing, it's *almost guaranteed* that a network request has to be made.

mary-ext:
I suppose moving resolveFromIdentity to `refresh` method might be the answer? So long as we're not doing it in-between it seems fine.

PR #2871, opened 2024-10-09
Various OAuth related fixes

matthieusieben:


matthieusieben:
Fixes https://github.com/bluesky-social/atproto/issues/2816

packages/oauth/oauth-client/src/util.ts
```diff
@@ -116,3 +117,52 @@ export class CustomEventTarget<EventDetailMap extends Record<string, unknown>> {
...
+  if (length === 1) return signals.find(Boolean)!
+
+  const controller = new AbortController()
+  const { signal } = controller
+
+  for (const signal of signals) {
+    if (signal) {
+      signal.addEventListener('abort', () => controller.abort(), { signal })
+    }
+  }
```
devinivy, line 165: One of the signals may also already be aborted.

packages/oauth/oauth-client/src/util.ts
```diff
@@ -116,3 +117,52 @@ export class CustomEventTarget<EventDetailMap extends Record<string, unknown>> {
...
+  if (length === 0) return undefined
+  if (length === 1) return signals.find(Boolean)!
+
+  const controller = new AbortController()
+  const { signal } = controller
+
+  for (const signal of signals) {
+    if (signal) {
+      signal.addEventListener('abort', () => controller.abort(), { signal })
+    }
```
devinivy, line 164: There's shadowing going on here for `signal` which I think may cause some issues.  I assume the `{ signal }` options are intended to reference the outer `controller.signal`, to remove all listeners when `controller.abort()` occurs.

packages/oauth/oauth-client/src/util.ts
```diff
@@ -116,3 +117,52 @@ export class CustomEventTarget<EventDetailMap extends Record<string, unknown>> {
...
+  | `${Value} ${string}`
+  | `${string} ${Value}`
+  | `${string} ${Value} ${string}`
+
+export const includesSpaceSeparatedValue = <Value extends string>(
+  input: string,
+  value: Value,
+): input is SpaceSeparatedValue<Value> => {
+  // Optimized version of:
+  // return input.split(' ').includes(value)
```
devinivy, line 132: Seems pretty nice!  Is there an issue with using this implementation?

packages/oauth/oauth-client/src/session-getter.ts
```diff
@@ -205,11 +210,15 @@ export class SessionGetter extends CachedGetter<string, Session> {
   }
 
   async setStored(sub: string, session: Session) {
+    // Prevent tempering with the stored value
```
devinivy, line 213:
```suggestion
    // Prevent tampering with the stored value
```

packages/oauth/oauth-client/src/util.ts
matthieusieben: The `split()` implementation is [about 8 times slower](https://jsbench.me/s2m24q9yz6/1). I'm just tying to save the planet here, one CPU cycle at the time.

PR #2873, opened 2024-10-11
✨ Ozone mod history API

foysalit:
This PR introduces a set of lexicons that exposes moderation history to users. 
A detailed proposal for this feature can be found here https://github.com/bluesky-social/proposals/pull/60

Few of things to consider
1. do we need a migration to populate mod history for all previously handled subjects?
2. we are adding a flag for automated actions which was not part of the original spec. we may want to reconsider this but right now it's not stored in db and computed on the fly so easy to change quickly.
3. we don't have the facility to add separate public facing comments with mod actions and all comments are assumed to be internal so to the public, it may seem like all actions are left without comments.


packages/ozone/src/api/history/getSubjectHistory.ts
```diff
@@ -0,0 +1,60 @@
...
+        }
+      }
+
+      const modHistoryService = ctx.modStatusHistoryService(db)
+      const modService = ctx.modService(db)
+      const results = await modService.getEvents({
+        subject,
+        limit,
+        cursor,
+        types: publishableModEvents,
```
foysalit, line 35: This ensures we don't accidentally publish an event that end users shouldn't see.

lexicons/tools/ozone/history/defs.json
```diff
@@ -0,0 +1,166 @@
...
+        "modAction": {
+          "type": "string",
+          "knownValues": [
+            "#modActionPending",
+            "#modActionLabel",
+            "#modActionResolve",
+            "#modActionSuspend",
+            "#modActionTakedown"
+          ]
+        },
```
devinivy, line 25:
```suggestion
        "modAction": {
          "type": "string",
          "knownValues": [
            "tools.ozone.history.defs#modActionPending",
            "tools.ozone.history.defs#modActionLabel",
            "tools.ozone.history.defs#modActionResolve",
            "tools.ozone.history.defs#modActionSuspend",
            "tools.ozone.history.defs#modActionTakedown"
          ]
        },
```

lexicons/tools/ozone/history/defs.json
```diff
@@ -0,0 +1,166 @@
...
+      "type": "object",
+      "required": ["subject", "status", "modAction"],
+      "properties": {
+        "subject": {
+          "type": "string",
+          "format": "uri"
+        },
+        "status": {
+          "type": "string"
+        },
```
devinivy, line 15: Are there any known values for the subject status?

lexicons/tools/ozone/history/defs.json
```diff
@@ -0,0 +1,166 @@
...
+  "id": "tools.ozone.history.defs",
+  "defs": {
+    "subjectBasicView": {
+      "type": "object",
+      "required": ["subject", "status", "modAction"],
+      "properties": {
+        "subject": {
+          "type": "string",
+          "format": "uri"
+        },
```
devinivy, line 12: I expect that using this field would end-up written `subject.subject`.  Is there any merit in naming it `subject.uri` or `subject.identifier`?

lexicons/tools/ozone/history/defs.json
```diff
@@ -0,0 +1,166 @@
...
+        "createdAt": {
+          "type": "string",
+          "format": "datetime"
+        },
+        "isAutomated": {
+          "type": "boolean",
+          "default": false
```
devinivy, line 53: Since this is required the default wont come into play.
```suggestion
          "type": "boolean"
```

lexicons/tools/ozone/history/getAccountActions.json
```diff
@@ -0,0 +1,45 @@
...
+  "lexicon": 1,
+  "id": "tools.ozone.history.getAccountActions",
+  "defs": {
+    "main": {
+      "type": "query",
+      "description": "Fetch all subjects from a single account with mod subjects.",
+      "parameters": {
+        "type": "params",
+        "properties": {
+          "account": { "type": "string", "format": "did" },
```
devinivy, line 11: Will this be determined by auth?

lexicons/tools/ozone/history/getSubjectHistory.json
```diff
@@ -0,0 +1,40 @@
...
+          "cursor": { "type": "string" }
+        }
+      },
+      "output": {
+        "encoding": "application/json",
+        "schema": {
+          "type": "object",
+          "properties": {
+            "events": {
+              "type": "array",
```
devinivy, line 28: Perhaps `events` will be required (and an empty array when there are no results).

devinivy (commented):
Did a pass of review on the lexicons—nice progress!

lexicons/tools/ozone/history/defs.json
foysalit: I am not a fan of uri since we did this for labels and it has been a bit confusing to see `uri` containing dids too. i like identifier but not sure if we need to introduce a new terminology. nit but this usually goes under an array property named `subjects` so while we won't be forced to write `subject.subject` right now, it may end up that way soon of course.

lexicons/tools/ozone/history/getAccountActions.json
foysalit: yes. this is only to be used as an escape hatch for admins to see what a user may see when viewing mod history. auth check will ensure that this is allowed only for mods with the right roles.

lexicons/tools/ozone/history/defs.json
foysalit: hmm... but omitting the suffix here works too, no?

lexicons/tools/ozone/history/defs.json
foysalit: yep, added those.

lexicons/tools/ozone/history/defs.json
devinivy: Each of these seem to reflect the value of a token, listed near the end of this lexicon document.  Token values are intentionally namespaced and are represented as a the fully-qualified string including the NSID and the fragment.  For example `#modActionPending` is not a string value for a token but `tools.ozone.history.defs#modActionPending` is.  So I figured you wanted to list the token string values in here.  More deets on tokens [here](https://atproto.com/specs/lexicon#field-type-definitions).

lexicons/tools/ozone/history/defs.json
```diff
@@ -0,0 +1,166 @@
...
+        }
+      }
+    },
+    "eventTakedown": {
+      "type": "object",
+      "description": "Permanently or temporarily taken down",
+      "properties": {
+        "comment": {
+          "type": "string"
+        },
```
devinivy, line 73: What field do the public-facing event comments get mapped from?  Mostly just want to double-check whether these are the mods' internal event comments or something else.

foysalit: we are not mapping any comments right now. . this is one of the open questions in the PR, whether or not we should provide a way to add public facing comments as part of this work or is that a follow up thing?

foysalit:
Closing in favor of https://github.com/bluesky-social/atproto/pull/4893

PR #2874, opened 2024-10-11
Add OAuth tests

matthieusieben:


devinivy (approved):
Nice!

packages/oauth/oauth-client/src/constants.ts
```diff
@@ -2,3 +2,20 @@
...
+  // try/catch to support running in a browser, including when process.env is
+  // shimmed (e.g. by webpack)
+  try {
+    return (
+      process.env.NODE_ENV === 'development' || process.env.NODE_ENV === 'test'
+    )
+  } catch {
+    return false
+  }
+})()
```
devinivy, line 21: Is there a reason why we should set default based on `NODE_ENV` rather than allowing deployments to configure this themselves based on their particular situation?

packages/oauth/oauth-client/src/oauth-authorization-server-metadata-resolver.ts
```diff
@@ -31,18 +35,30 @@ export class OAuthAuthorizationServerMetadataResolver extends CachedGetter<
...
   }
 
   async get(
-    issuer: string,
+    input: string,
     options?: GetCachedOptions,
   ): Promise<OAuthAuthorizationServerMetadata> {
-    return super.get(oauthIssuerIdentifierSchema.parse(issuer), options)
+    const issuer = oauthIssuerIdentifierSchema.parse(input)
+    if (!this.allowUnsecure && !issuer.startsWith('https:')) {
```
devinivy, line 56: Small nit, but I think we want to use "insecure" rather than "unsecure."  We could also make it more specific, referencing whether "http" is allowed.

packages/oauth/oauth-client/src/constants.ts
devinivy: Ah, I see now, this convention already existed 👍  I usually lean towards having uniform behavior across `NODE_ENV`, or only treating `NODE_ENV=production` specially if anything.  Then testing or dev environments would manually set `allowUnsecure` themselves.  This makes the differences in environments explicit, and I think tends to produce fewer surprises when transitioning to run code in production.  Even so, we should probably not change how this works in this PR.

devinivy (approved):
Small nit around the "unsecure" terminology, otherwise looking solid!

packages/oauth/oauth-client/src/constants.ts
matthieusieben: You're right. I actually thought about doing this after I submitted for review.

PR #2875, opened 2024-10-11
Appview: temp fix for bad pinned post values

devinivy:
Looks like a client or bug may have been responsible for some bad pinned post values on profiles, causing some responses to be invalid.  This should set things straight while it gets worked out.


dholms (approved):
yup yup looks good 👍

mary-ext:
Are the bad values because someone's setting a pinnedPost with a string rather than a strongRef? It seems like we're clashing with a rather popular unspecced field

bnewbold:
@mary-ext that does seem to be the issue. none of us (at bluesky) were aware that other folks were using the exact same field name in a different way.

an interesting case-study in lexicon evolution and governance! should probably do a write-up / post-mortem and have better guidance on folks doing "extensions" can chose field names to avoid conflict like this going forward. eg, use an NSID as the field name, or add some non-conflicting string prefix to reduce the chance of a collision.

PR #2876, opened 2024-10-11
Appview: pinned post validation fix

devinivy:
The validation function hard-throws if you pass it a non-object, so we need to perform this additional check.


dholms (approved):
nice 👍

PR #2881, opened 2024-10-16
Lexicon: languages on starter pack, feed gen, and labeler records

devinivy:
This adds optional `langs` fields on starter pack, feed gen, and labeler records.  This is in line with the `langs` field on post records, and should provide more signal on relevance for discovery mechanisms in the app.

### DIFF
--- a/lexicons/app/bsky/feed/generator.json
+++ b/lexicons/app/bsky/feed/generator.json
@@ -39,6 +39,12 @@
             "description": "Self-label values",
             "refs": ["com.atproto.label.defs#selfLabels"]
           },
+          "langs": {
+            "type": "array",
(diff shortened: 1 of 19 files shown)


lexicons/app/bsky/feed/generator.json
```diff
@@ -39,6 +39,12 @@
             "description": "Self-label values",
             "refs": ["com.atproto.label.defs#selfLabels"]
           },
+          "langs": {
+            "type": "array",
+            "description": "Indicates the human language for the intended audiences of the feed generator.",
```
estrattonbailey, line 44: Should we mention the format here? Like pry leave it open, but specify at least BCP-47 with or without region codes?

PR #2894, opened 2024-10-19
Fix root block missing in too big seq commit

### LINKED ISSUE #2893: Firehose consumers do not receive root block in case of too big seq commit
**Describe the bug**

Firehose consumers do not receive root block in case of too big seq commit. 

**To Reproduce**

I wrote a corresponding unit test in the linked PR that reproduces the issue

**Actual behavior**

`car.blocks` are always empty 

**Expected behavior**

`car.blocks` must be always one (the root one)

**Additional context**

Line no. 29 is so suspicious:
https://github.com/bluesky-social/atproto/blob/81ae1b12389f1c2b1ca9aa5ed0660a0f569c4c9d/packages/pds/src/sequencer/events.ts#L25-L31

Here is why:
- `.add` of `BlockMap` returns a promise and no one waits for resolve (forgotten `await`)
- input arg of `.add` is typed as`LexValue` which goes through `dataToCborBlock(lexToIpld(value))`. That does not make sense since we just move the block from one `BlockMap` to another one. `newBlocks.get` returns bytes already.

P.S. I dug into this because I got reports from my lovely Python devs that `commit.blocks` received from the firehose message frame could be missed. But it is marked as a required field in the lexicon. From our observation, it is missed always in combination with `tooBig=true`. I will continue my journey, but if you have any ideas I would appreciate it

MarshalX:
fixes #2893

### DIFF
--- a/packages/pds/src/sequencer/events.ts
+++ b/packages/pds/src/sequencer/events.ts
@@ -27,9 +27,9 @@ export const formatSeqCommit = async (
     tooBig = true
     const justRoot = new BlockMap()
     const rootBlock = commitData.newBlocks.get(commitData.cid)
-if (rootBlock) {
-  justRoot.set(commitData.cid, rootBlock)
-}
+    if (rootBlock) {
+      justRoot.set(commitData.cid, rootBlock)
+    }
     carSlice = await blocksToCarFile(commitData.cid, justRoot)
   } else {
     tooBig = false
--- a/packages/pds/src/sequencer/events.ts
+++ b/packages/pds/src/sequencer/events.ts
@@ -27,7 +27,9 @@ export const formatSeqCommit = async (
     tooBig = true
     const justRoot = new BlockMap()
     const rootBlock = commitData.newBlocks.get(commitData.cid)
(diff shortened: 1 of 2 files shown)


packages/pds/src/sequencer/events.ts
```diff
@@ -26,7 +26,8 @@ export const formatSeqCommit = async (
   if (writes.length > 200 || commitData.newBlocks.byteSize > 1000000) {
     tooBig = true
     const justRoot = new BlockMap()
-    justRoot.add(commitData.newBlocks.get(commitData.cid))
+    const rootBlock = commitData.newBlocks.get(commitData.cid)
+    rootBlock && justRoot.set(commitData.cid, rootBlock)
```
dholms, line 30: I know this syntax works, but I prefer the explicit `if` statement 

```suggestion
if(rootBlock) {
  justRoot.set(commitData.cid, rootBlock)
}
```

dholms (approved):
Looks great! Thanks for the test.

Made a small tweak, and will get this merged after CI passes

PR #2905, opened 2024-10-22
:sparkles: Ozone instance-wide and user-specific settings

foysalit:
This PR adds `instance level` and `personal level` settings to ozone.
Through the 3 new endpoints `tools.ozone.setting.upsertOption`, `tools.ozone.setting.listOptions` and `tools.ozone.setting.removeOptions` ozone users can manage configuration that may be used instance-wide or just to customize the user's own preferences.

The current use case for this is will be to simply move some of the env var driven configurations that each client deploy needs.


packages/internal/rollup-plugin-bundle-manifest/src/index.ts
```diff
@@ -36,7 +36,7 @@ export default function bundleManifest({
 } = {}): Plugin {
   return {
     name: 'bundle-manifest',
-    generateBundle(outputOptions, bundle) {
+    generateBundle(outupsertOptions, bundle) {
```
devinivy, line 39: Wondered if this may not have been intentional.

lexicons/tools/ozone/setting/listOptions.json
```diff
@@ -0,0 +1,60 @@
...
+            "description": "Filter keys by prefix"
+          },
+          "keys": {
+            "type": "array",
+            "items": {
+              "type": "string",
+              "format": "nsid"
+            },
+            "description": "Filter for only the specified keys. Ignored if prefix is provided"
+          }
```
devinivy, line 36: Often nice to have a max length on array params, even if just to help keep the caller under url length limits.

lexicons/tools/ozone/setting/removeOptions.json
```diff
@@ -0,0 +1,32 @@
...
+      "description": "Delete settings by key",
+      "input": {
+        "encoding": "application/json",
+        "schema": {
+          "type": "object",
+          "required": ["keys", "scope"],
+          "properties": {
+            "keys": {
+              "type": "array",
+              "minLength": 1,
```
devinivy, line 16: Might recommend a max length here too, even if it's relatively high.

lexicons/tools/ozone/setting/removeOptions.json
```diff
@@ -0,0 +1,32 @@
...
+              "items": {
+                "type": "string",
+                "format": "nsid"
+              }
+            },
+            "scope": {
+              "type": "string",
+              "knownValues": ["instance", "personal"],
+              "default": "personal"
```
devinivy, line 25: The scope is required so the default here will have no effect.

lexicons/tools/ozone/setting/removeOptions.json
```diff
@@ -0,0 +1,32 @@
+{
+  "lexicon": 1,
+  "id": "tools.ozone.setting.removeOptions",
+  "defs": {
+    "main": {
+      "type": "procedure",
+      "description": "Delete settings by key",
+      "input": {
```
devinivy, line 8: I'd recommend also having an `output`, even if it has an empty schema.  We found that we wish we had done that on all lexicons w/ empty output!

lexicons/tools/ozone/setting/upsertOption.json
```diff
@@ -0,0 +1,56 @@
...
+              "maxLength": 2000
+            },
+            "managerRole": {
+              "type": "string",
+              "default": "owner",
+              "knownValues": [
+                "tools.ozone.team.defs#roleModerator",
+                "tools.ozone.team.defs#roleTriage",
+                "tools.ozone.team.defs#roleAdmin"
+              ]
```
devinivy, line 36: I assume the default is intended to be one of the known values.

packages/ozone/src/api/setting/upsertOption.ts
```diff
@@ -0,0 +1,136 @@
...
+          scope: 'personal',
+          managerRole: null,
+        })
+      } else {
+        const manageableRoles = getRolesForInstanceOption(access)
+        const existingSetting = await settingService.query({
+          scope: 'instance',
+          keys: [key],
+          limit: 1,
+        })
```
devinivy, line 56: Could `getExistingSetting()` be used here too?

packages/ozone/src/api/setting/upsertOption.ts
```diff
@@ -0,0 +1,136 @@
...
+          scope: 'instance',
+          managerRole: getManagerRole(managerRole),
+        })
+      }
+
+      const newOption = await getExistingSetting(settingService, key, scope)
+
+      if (!newOption) {
+        throw new Error('Failed to get the updated setting')
+      }
```
devinivy, line 75: I quite like assertions for cases like this, from `node:assert`.

packages/ozone/src/db/schema/index.ts
```diff
@@ -9,6 +9,7 @@ import * as signingKey from './signing_key'
 import * as communicationTemplate from './communication_template'
 import * as set from './ozone_set'
 import * as member from './member'
+import * as Setting from './setting'
```
devinivy, line 12: Just a small consistency thing.
```suggestion
import * as member from './member'
import * as setting from './setting'
```

packages/ozone/src/db/schema/setting.ts
```diff
@@ -0,0 +1,24 @@
...
+import { Member } from './member'
+
+export const settingTableName = 'setting'
+
+export type SettingScope = 'personal' | 'instance'
+
+export interface Setting {
+  id: Generated<number>
+  key: string
+  value: any
```
devinivy, line 11: The value here should at minimum be an object due to the way it's modeled as `"type": "unknown"` in lexicon.  (Even if it weren't an object, the `unknown` type may be preferable to `any` here to help with type correctness.)

packages/ozone/src/db/schema/setting.ts
```diff
@@ -0,0 +1,24 @@
+import { Generated } from 'kysely'
+import { Member } from './member'
+
+export const settingTableName = 'setting'
+
+export type SettingScope = 'personal' | 'instance'
+
+export interface Setting {
+  id: Generated<number>
```
devinivy, line 9: May consider this if we never want it set from app code.
```suggestion
  id: GeneratedAlways<number>
```

packages/ozone/tests/settings.test.ts
```diff
@@ -0,0 +1,309 @@
...
+        ],
+        'instance',
+      )
+    })
+
+    it('returns all personal settings', async () => {
+      const result = await listOptions({ prefix: 'tools.ozone.setting.client' })
+      expect(result.options.length).toBe(3)
+      // @TODO: once we're happy with the object definitions, we can snapshot the result
+      // expect(forSnapshot(result.settings)).toMatchSnapshot()
```
devinivy, line 194: This may be good to go now.

packages/ozone/src/setting/service.ts
```diff
@@ -0,0 +1,146 @@
...
+      scope: SettingScope
+      managerRole: Member['role'][]
+    },
+  ): Promise<void> {
+    if (!keys.length) return
+
+    let qb = this.db.db
+      .deleteFrom('setting')
+      .where('key', 'in', keys)
+      .where('scope', '=', filters.scope)
```
devinivy, line 104: It may be worth asserting that `filters.did` is set when `scope` is `personal`.

packages/ozone/src/db/migrations/20241018T205730722Z-setting.ts
```diff
@@ -0,0 +1,34 @@
...
+      col.defaultTo(sql`now()`).notNull(),
+    )
+    .addColumn('updatedAt', 'timestamptz', (col) =>
+      col.defaultTo(sql`now()`).notNull(),
+    )
+    .addColumn('managerRole', 'text')
+    .addColumn('scope', 'text', (col) => col.notNull())
+    .addColumn('createdBy', 'text', (col) => col.notNull())
+    .addColumn('lastUpdatedBy', 'text', (col) => col.notNull())
+    .addUniqueConstraint('setting_scope_key_idx', ['key', 'scope'])
```
devinivy, line 21: I believe `(key, scope)` wont be unique when multiple users have the same `personal` setting with the same `key`.  Maybe qualifying this by `did` does the trick, and then we can nix `setting_did_idx`.

packages/ozone/src/setting/service.ts
```diff
@@ -0,0 +1,146 @@
...
+
+    if (scope) {
+      builder = builder.where('scope', '=', scope)
+    }
+
+    if (did) {
+      builder = builder.where('did', '=', did)
+    }
+
+    const keyset = new TimeIdKeyset(ref(`setting.createdAt`), ref('setting.id'))
```
devinivy, line 52: Is it possible that `id` is enough for pagination?

devinivy (commented):
Left a few notes, but this is looking great.  Let's add a changeset!

packages/internal/rollup-plugin-bundle-manifest/src/index.ts
foysalit: lol I think i did a find and replace which messed this up

lexicons/tools/ozone/setting/listOptions.json
foysalit: yep yep. i had this elsewhere but missed a few spot ig

packages/ozone/src/api/setting/upsertOption.ts
foysalit: that's why i refactored it into the function but forgot to replace in both places.

packages/ozone/src/setting/service.ts
foysalit: yeah that should work I think.

packages/ozone/tests/__snapshots__/settings.test.ts.snap
```diff
@@ -0,0 +1,3 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`ozone-settings listOptions returns all personal settings 1`] = `undefined`;
```
devinivy, line 3: I expect that the snapshot here is probably intended to be an object.

devinivy (approved):

PR #2910, opened 2024-10-25
lex-cli use .js imports

matthieusieben:
In order to use the generated code with `typescript` with `"moduleResolution": "node16"`, imported files must be the full exact path.


devinivy (approved):

matthieusieben:
Added this to https://github.com/bluesky-social/atproto/pull/2999

PR #2911, opened 2024-10-25
lex-cli improvements

matthieusieben:
This change aims to allow performing Typescript type inference based on the `schemas` variable generated through lex-cli. It also makes that inferable `schemas` definition array public in the `@atproto/api` package.


packages/syntax/src/did.ts
```diff
@@ -12,25 +12,31 @@
 //   - hard length limit of 8KBytes
 //   - not going to validate "percent encoding" here
 export const ensureValidDid = (did: string): void => {
+  if (
+    did.length < 7 || // did:x:y
+    did[3] !== ':' ||
+    did[2] !== 'd' ||
+    did[1] !== 'i' ||
+    did[0] !== 'd'
```
devinivy, line 20: It's more idiomatic to write `did.startsWith('did:')`.  Is there a strong case to optimize beyond `startsWith()`, which I would expect to be quite fast?  (It's unlikely, but I wouldn't be _entirely_ surprised if `startsWith()` ends-up being comparable or faster anyway.)

devinivy (approved):

packages/syntax/src/did.ts
matthieusieben: I let myself be influenced by some [code I read](https://github.com/nodejs/undici/blob/c94fda14b0c58cd1083c6efb7a49868bab4d835f/lib/api/util.js#L59-L77) in undici's source

matthieusieben: I guess it's faster because it check for unlikely chars first ?

PR #2922, opened 2024-10-29
Support alternate did methods in update handle

dholms:
If a user's DID is not did:plc, then resolve the DID to ensure that they set their handle correctly & track handle update

### DIFF
--- a/packages/pds/src/api/com/atproto/identity/updateHandle.ts
+++ b/packages/pds/src/api/com/atproto/identity/updateHandle.ts
@@ -54,7 +54,7 @@ export default function (server: Server, ctx: AppContext) {
       })
 
       if (!account) {
-        if (requester.startsWith('did:plc')) {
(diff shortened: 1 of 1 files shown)


packages/pds/src/api/com/atproto/identity/updateHandle.ts
```diff
@@ -53,13 +53,31 @@ export default function (server: Server, ctx: AppContext) {
...
+          if (resolved.handle !== handle) {
+            throw new InvalidRequestError(
+              'DID is not properly configured for handle',
+            )
+          }
+        }
+        await ctx.accountManager.updateHandle(requester, handle)
+      } else {
+        // if we found an account with matching handle, check if it is the same as requester
+        // is so, error. if not, then just emit an identity event
```
devinivy, line 77:
```suggestion
        // if so emit an identity event, otherwise error.
```

devinivy (approved):

PR #2923, opened 2024-10-29
Fix handling of too big events

dholms:
We were using the wrong method on BlockMap here: the async `.add` which serializes & hashes the content instead of the synchronous `.set` since we already know the bytes & cid for the block

Closes https://github.com/bluesky-social/atproto/issues/2893


dholms:
closing in favor of https://github.com/bluesky-social/atproto/pull/2894

PR #2929, opened 2024-10-30
Add threat signatures to admin view

dholms:
This is a relatively freeform field for attaching additional arbitrary threat signatures to a given admin view

### DIFF
--- a/lexicons/com/atproto/admin/defs.json
+++ b/lexicons/com/atproto/admin/defs.json
@@ -33,7 +33,14 @@
         "invitesDisabled": { "type": "boolean" },
         "emailConfirmedAt": { "type": "string", "format": "datetime" },
         "inviteNote": { "type": "string" },
-        "deactivatedAt": { "type": "string", "format": "datetime" }
+        "deactivatedAt": { "type": "string", "format": "datetime" },
+        "threatSignatures": {
+          "type": "array",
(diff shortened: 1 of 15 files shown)


lexicons/com/atproto/admin/defs.json
```diff
@@ -51,6 +58,14 @@
         "cid": { "type": "string", "format": "cid" },
         "recordUri": { "type": "string", "format": "at-uri" }
       }
+    },
+    "threatSignature": {
+      "type": "object",
+      "required": ["property", "value"],
+      "properties": {
+        "property": { "type": "string" },
```
devinivy, line 66: The way I think of it, the threat signature has multiple properties, each of which has a name and a value.  So I think of this as a `name`.  But it's truly a bikeshed, all good either way.

devinivy (approved):

lexicons/com/atproto/admin/defs.json
dholms: hmmm yeah i buy it 👍

dholms: yeah I see where you're coming from, but we already have stuff in entryway codebase around this naming convention and this isn't so far off that I wanna do DB migrations for it 😅

PR #2934, opened 2024-10-31
Add `match: MuteWordMatch` to `muted-word` mod decision `cause`

estrattonbailey:
This PR adds a `match: MuteWordMatch` prop to the `muted-word` `cause` type. Previous logic was checking for truthy/falsy values, so the migration was pretty straightforward. This requires no change to the matching logic itself.

Matching logic now:
- returns `MuteWordMatch | null` instead of a boolean
- migrates some tests to `toBeTruthy` and `toBeNull`
- adds the new prop to `muted-word` cause
- updates a few tests to check for this new prop

### DIFF
--- a/packages/api/src/moderation/decision.ts
+++ b/packages/api/src/moderation/decision.ts
@@ -15,6 +15,7 @@ import {
 } from './types'
 import { ModerationUI } from './ui'
 import { LABELS } from './const/labels'
+import { MuteWordMatch } from './mutewords'
 
 enum ModerationBehaviorSeverity {
   High,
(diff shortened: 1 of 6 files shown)


packages/api/src/moderation/decision.ts
```diff
@@ -195,12 +196,13 @@ export class ModerationDecision {
     }
   }
 
-  addMutedWord(mutedWord: boolean) {
-    if (mutedWord) {
+  addMutedWord(matches: MuteWordMatch[] | undefined) {
+    if (matches) {
```
matthieusieben: Does it make sense to check for the length here ?

```suggestion
    if (matches?.length) {
```

estrattonbailey: Ah yeah good call, should do that just to be safe. Thanks!

PR #2936, opened 2024-10-31
Use node:crypto instead of noble/curves

ghost:
`@atproto/crypto` currently does signature validation with `@noble/curves`, which is a JS implementation.

Our CPU profiling shows that validation takes a good CPU share, so we want to see how that improves by switching to the native `node:crypto`, but not yet changing the current implementation in `@atproto/crypto`.

The solution was to make customizable the signature verification function that happens inside the JWT validation, which happens on `@atproto/xrpc-server`.

So in short:
1. `@atproto/crypto` exports the utils to avoid duplication.
2. `@aproto/xrpc-server` allows customization of the signature verification function.
3. `@atproto/bsky` uses 1 to build a customization for 2, using `node:crypto` instead of `@noble/curves`.

### DIFF
--- a/packages/bsky/src/auth-verifier.ts
+++ b/packages/bsky/src/auth-verifier.ts
@@ -491,13 +491,7 @@ export const verifySignatureWithKey: VerifySignatureWithKeyFn = async (
       throw new Error(`Expected key alg ${alg}, got ${parsed.jwtAlg}`)
     }
 
-    const prefixedBytes = extractPrefixedBytes(extractMultikey(didKey))
-    if (!hasPrefix(prefixedBytes, SECP256K1_DID_PREFIX)) {
-      throw new Error(`Not a secp256k1 did:key: ${didKey}`)
-    }
-    const keyBytes = prefixedBytes.slice(SECP256K1_DID_PREFIX.length)
-
-    return verifySig(keyBytes, msgBytes, sigBytes)
+    return verifySig(parsed.keyBytes, msgBytes, sigBytes)
   }
 
   return cryptoVerifySignatureWithKey(didKey, msgBytes, sigBytes, alg)


packages/xrpc-server/src/auth.ts
```diff
@@ -116,18 +124,13 @@ export const verifyJwt = async (
 
   const msgBytes = ui8.fromString(parts.slice(0, 2).join('.'), 'utf8')
   const sigBytes = ui8.fromString(sig, 'base64url')
-  const verifySignatureWithKey = async (key: string) => {
-    return crypto.verifySignature(key, msgBytes, sigBytes, {
-      jwtAlg: header.alg,
-      allowMalleableSig: true,
-    })
-  }
```
matthieusieben, line 124: Unless there is a strong reason not to do so, I would advocate that we change this function here to be implemented this using `node:crypto` instead of `@atproto/crypto`.

matthieusieben (changes requested):
This looks good. It's kind of a bummer that we are able to benefit from this change everywhere.

Wouldn't be a better solution to add your changes to the following function instead ?

https://github.com/bluesky-social/atproto/blob/709ba301578c1956b8eb0d89bad717615a4fd7ba/packages/xrpc-server/src/auth.ts#L119-L124

matthieusieben (approved):
Gotcha !

packages/bsky/src/auth-verifier.ts
```diff
@@ -440,5 +451,54 @@ export const buildBasicAuth = (username: string, password: string): string => {
...
+  sigBytes: Uint8Array,
+  alg: string,
+) => {
+  if (alg === SECP256K1_JWT_ALG) {
+    const parsed = parseDidKey(didKey)
+    if (alg !== parsed.jwtAlg) {
+      throw new Error(`Expected key alg ${alg}, got ${parsed.jwtAlg}`)
+    }
+
+    const prefixedBytes = extractPrefixedBytes(extractMultikey(didKey))
```
dholms, line 469: I think this chunk of code is redundant with the above `parseDidKey`

packages/bsky/src/auth-verifier.ts
```diff
@@ -440,5 +451,54 @@ export const buildBasicAuth = (username: string, password: string): string => {
...
+      throw new Error(`Not a secp256k1 did:key: ${didKey}`)
+    }
+    const keyBytes = prefixedBytes.slice(SECP256K1_DID_PREFIX.length)
+
+    const verifySig = (
+      publicKey: Uint8Array,
+      data: Uint8Array,
+      sig: Uint8Array,
+    ) => {
+      const keyEncoder = new KeyEncoder('secp256k1')
```
dholms, line 480: maybe we instantiate this outside of the function so we don't have to instantiate this class each time

packages/bsky/src/auth-verifier.ts
```diff
@@ -440,5 +451,54 @@ export const buildBasicAuth = (username: string, password: string): string => {
...
+      throw new Error(`Expected key alg ${alg}, got ${parsed.jwtAlg}`)
+    }
+
+    const prefixedBytes = extractPrefixedBytes(extractMultikey(didKey))
+    if (!hasPrefix(prefixedBytes, SECP256K1_DID_PREFIX)) {
+      throw new Error(`Not a secp256k1 did:key: ${didKey}`)
+    }
+    const keyBytes = prefixedBytes.slice(SECP256K1_DID_PREFIX.length)
+
+    const verifySig = (
```
dholms, line 475: actually same goes for this whole function. I don't think we're using a closure here, so we can probably just pull this out & only define the function once rather than once per call

packages/bsky/tests/views/timeline.test.ts
```diff
@@ -57,6 +58,26 @@ describe('timeline views', () => {
   // @TODO(bsky) blocks posts, reposts, replies by actor takedown via labels
   // @TODO(bsky) blocks posts, reposts, replies by record takedown via labels
 
+  it('throws if the user key is incorrect', async () => {
```
devinivy, line 61: Could you move this test to the auth test suite?

devinivy (approved):

PR #2943, opened 2024-11-04
Appview: apply epoch for indexed-at times in views

devinivy:
This allows the appview to be configured with an epoch after which record indexed-at times are always presented as the time the appview first observed the record.  That is as opposed to usering the appview's hybrid sorted-at time which factors in the untrusted created-at time per https://docs.bsky.app/docs/advanced-guides/timestamps#sortat.

### DIFF
--- a/packages/bsky/src/views/index.ts
+++ b/packages/bsky/src/views/index.ts
@@ -183,7 +183,13 @@ export class Views {
     return {
       ...basicView,
       description: actor.profile?.description || undefined,
-      indexedAt: actor.sortedAt?.toISOString(),
+      indexedAt:
+        actor.indexedAt && actor.sortedAt
+          ? this.indexedAt({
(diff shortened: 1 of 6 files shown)


packages/bsky/src/views/index.ts
```diff
@@ -148,7 +154,7 @@ export class Views {
           }
         : undefined,
       banner: actor.profile?.banner
-        ? this.imgUriBuilder.getPresetUri(
+        ? this.opts.imgUriBuilder.getPresetUri(
```
devinivy, line 157:
```suggestion
        ? this.imgUriBuilder.getPresetUri(
```

dholms (approved):
yup 👍

DavidBuchanan314:
Is this more of an anti-election-misinfo hotfix or is this an intended medium/long-term solution?

PR #2945, opened 2024-11-04
OAuth: Reset password & Sign-up

matthieusieben:


packages/pds/src/account-manager/account-manager.ts
```diff
@@ -502,250 +548,4 @@ export class AccountManager
       ]),
     )
   }
-
```
matthieusieben, line 505: Most of this code didn't change and was just moved in `./oauth-store`.

packages/pds/src/account-manager/oauth-store.ts
```diff
@@ -0,0 +1,485 @@
...
+import * as token from './helpers/token'
+import * as usedRefreshToken from './helpers/used-refresh-token'
+
+/**
+ * This class' purpose is to implement the interface needed by the OAuthProvider
+ * to interact with the account database (through the {@link AccountManager}).
+ *
+ * @note The use of this class assumes that there is no entryway.
+ */
+export class OAuthStore
```
matthieusieben, line 54: Most of this code was just moved from `./account-manager`. What changes is only the stuff related to the `AccountStore`:
- account creation
- binding of device-account
- reset password
- handle availability

See the diff of the `AccountStore` interface to be able to tell which method have changed.

packages/oauth/oauth-client/src/util.ts
```diff
@@ -4,6 +4,8 @@ export type Simplify<T> = { [K in keyof T]: T[K] } & NonNullable<unknown>
 // @ts-expect-error
 Symbol.dispose ??= Symbol('@@dispose')
 
+export const ifString = <V>(v: V) => (typeof v === 'string' ? v : undefined)
```
devinivy, line 7: Does this benefit from being generic vs. typing the argument as `unknown`?

.changeset/tiny-goats-sing.md
```diff
@@ -0,0 +1,5 @@
+---
+"@atproto-labs/fetch": patch
+---
+
+Remove optional dependeny on "zod"
```
devinivy, line 5:
```suggestion
Remove optional dependency on "zod"
```

devinivy (commented):
Started review, just leaving this to flush a couple small review comments left so far.

packages/internal/rollup-plugin-bundle-manifest/src/index.ts
```diff
@@ -14,11 +14,11 @@ type ChunkItem = {
...
-  dynamicImports: string[]
-  isDynamicEntry: boolean
-  isEntry: boolean
-  isImplicitEntry: boolean
-  name: string
+  // dynamicImports: string[]
+  // isDynamicEntry: boolean
+  // isEntry: boolean
+  // isImplicitEntry: boolean
+  // name: string
```
devinivy, line 21: Should we remove these?

matthieusieben: I left them there as an indication to future maintainers of what is available (but not exposed because not used). We can surely get rid of this.

packages/oauth/oauth-client/src/util.ts
matthieusieben: Yes. Consider the following example ([playground](https://www.typescriptlang.org/play/?jsx=0#code/MYewdgzgLgBAlgMwMpQE5zAcxgXhgHgDUA+ACgDcAuGQgSl2JlKgE8AHAUxARnNxzwByaOiyCYAfl4xqAVzAATDggwcFtAFAbQkWIhSjMAVTABrXEyox5psCADuYejkbN2XHnwFCRGTOKk+OUVlVXUtDQB6SJhWTmoAbQAiBBAQJIAaGCSAIwBDVEyYABYAJiyEPIAbCA4AXW1waBgC1AsEjRgu7NT0jM7u3ILMga6y-u6YSpqODTqWiBgdaAjl2BUqqA5UNQtWgDoAWzy2Un00P3ponrSkmAAfbPzCh+sQlTA1LTWpuE3ttQmcx4A7HU7nQxAq4xXxYV7yJQfNRAA))

```ts

const ifString = <V>(v: V) => (typeof v === 'string' ? v : undefined)

const ifStringUnk = (v: unknown) => (typeof v === 'string' ? v : undefined)


// type: ["foo", "bar", 42, false]
const arr = [
    "foo",
    "bar",
    42,
    false
] as const


const filtered = arr.map(ifString) // "foo" | "bar" | undefined

const filteredUnk = arr.map(ifStringUnk) // string | undefined
```

packages/pds/src/account-manager/oauth-store.ts
```diff
@@ -0,0 +1,483 @@
...
+  }
+
+  // AccountStore
+
+  async createAccount({
+    inviteCode,
+    handle,
+    email,
+    password,
+    // @TODO: where is the birthdate saved ?
```
devinivy, line 123: I believe it is usually kept in the special-cased app user preferences.

matthieusieben: I'll remove it from the PDS then, as there is no place to store this data in the PDS until we have private state

surfdude29:
It's exciting to see the OAuth rollout progress 🚀

https://github.com/bluesky-social/atproto/blob/e1528181d1c154331e90e5c86b58db49d18d1490/packages/oauth/oauth-provider/src/assets/app/views/authorize/sign-up/sign-up-disclaimer.tsx#L31-L51

Sorry to jump in here though, if I've understood what's going on in the code (admittedly, a pretty big if 😅) I was wondering, would it be at all possible to have separate strings for each different scenario?

I think that would make it a fair bit easier for us translators, because we could then translate the full strings in context. This could be especially helpful in certain languages where the sentence structure can be different.

It might not be feasible to follow a similar path here, but as an example this is how the equivalent string was handled in `social-app`:
- https://github.com/bluesky-social/social-app/pull/3912

Also, I have a couple of wording tweak suggestions:

https://github.com/bluesky-social/atproto/blob/e1528181d1c154331e90e5c86b58db49d18d1490/packages/oauth/oauth-provider/src/assets/app/views/authorize/accept/accept-form.tsx#L94-L95

- I think "in accordance **with** their" would read more clearly.

https://github.com/bluesky-social/atproto/blob/e1528181d1c154331e90e5c86b58db49d18d1490/packages/oauth/oauth-provider/src/assets/app/views/authorize/reset-password/reset-password-confirm-form.tsx#L53-L54

- I think **Enter** should be capitalized here.

packages/oauth/oauth-provider/src/assets/app/lib/ref.ts
```diff
@@ -0,0 +1,18 @@
...
+    ref.current = value
+  }
+}
+
+export function mergeRefs<T>(refs: readonly (ForwardedRef<T> | undefined)[]) {
+  return (value: T | null) => {
+    for (let i = 0; i < refs.length; i++) {
+      const ref = refs[i]
+      if (ref) updateRef(ref, value)
+    }
```
devinivy, line 16: Any issue with `for..of` here?
```suggestion
    for (const ref of refs) {
      if (ref) updateRef(ref, value)
    }
```

packages/oauth/oauth-provider/src/assets/app/lib/password.ts
```diff
@@ -0,0 +1,79 @@
...
+  regexpsCountToMatch: number = regexps.length,
+): boolean {
+  if (regexpsCountToMatch < 1 || regexpsCountToMatch > regexps.length) {
+    throw new TypeError('Invalid regexpsCountToMatch')
+  }
+  for (let i = 0; i < regexps.length; i++) {
+    if (regexps[i].test(pwd)) {
+      if (--regexpsCountToMatch === 0) return true
+    }
+  }
```
devinivy, line 77: Any issue with `for..of` here?
```suggestion
  for (const regexp of regexps) {
    if (regexp.test(pwd)) {
      if (--regexpsCountToMatch === 0) return true
    }
  }
```

packages/oauth/oauth-provider/src/assets/app/lib/password.ts
```diff
@@ -0,0 +1,79 @@
...
+  pwd: string,
+  regexps: RegExp[],
+  regexpsCountToMatch: number = regexps.length,
+): boolean {
+  if (regexpsCountToMatch < 1 || regexpsCountToMatch > regexps.length) {
+    throw new TypeError('Invalid regexpsCountToMatch')
+  }
+  for (let i = 0; i < regexps.length; i++) {
+    if (regexps[i].test(pwd)) {
+      if (--regexpsCountToMatch === 0) return true
```
devinivy, line 75: Just a stylistic suggestion—since these are a little subtle I tend to lean towards keeping these things dumb / hard to mess-up:
```suggestion
      regexpsCountToMatch--
      if (regexpsCountToMatch === 0) return true
```

packages/oauth/oauth-provider/src/dpop/dpop-nonce.ts
```diff
@@ -1,48 +1,68 @@
 import { createHmac, randomBytes } from 'node:crypto'
```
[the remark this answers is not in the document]
matthieusieben: The important change here is that the validation of the constructor arguments was moved into Zod schemas that are exported, allowing the DpopManager to created a validation object for its own constructor arg based on the `DpopNonce` schemas exported here.

Doing so allows to better define the `DpopManager`'s constructor options type (through the magic of Zod), which ends up being used, in combination of other types, to describe the options of the `OAuthProvider`.

This change should not have any functional effect and was not strictly required as part of this PR.

packages/oauth/oauth-provider/src/oauth-provider.ts
```diff
@@ -1033,54 +1091,64 @@ export class OAuthProvider extends OAuthVerifier {
...
+
+      res.setHeader('Access-Control-Allow-Headers', 'Content-Type,DPoP')
+
+      next()
+    }
+
+    const corsPreflight: Middleware = combineMiddlewares([
+      corsHeaders,
+      (req, res) => {
+        res.writeHead(204).end()
```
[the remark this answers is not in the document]
matthieusieben: Changed to 200 👍

surfdude29:
https://github.com/bluesky-social/atproto/blob/1e78fa00e09b2c3636b5abef0f9ab7bb94765c9c/packages/oauth/oauth-provider/src/lib/locale.ts#L3-L6

Apologies if I've misunderstood what's going on here, I just wanted to check: would this regex work OK with three-letter locales? We currently have one: `ast` (Asturian).

matthieusieben:
@surfdude29 you're right. This would not have been an issue until we actually provided Asturian translations but I fixed it nonetheless.

surfdude29:
> @surfdude29 you're right. This would not have been an issue until we actually provided Asturian translations but I fixed it nonetheless.

Ah got it 👍

Thanks for fixing it 🙏

PR #2946, opened 2024-11-04
Fix optimistic getPostThread for URIs with handle

gaearon:
Previously, `getPostThread` read-after-write handling was not taking `at://<handle>/...` URIs into account. This resulted in the PDS returning 404 for post thread query whenever relay is lagging, which resulted in a confusing UX in the app (the post would show up but clicking into it would show 404). This should fix that.

## Test Plan

See two additions to tests. One of those worked before, the other one failed. Now both work.

### DIFF
--- a/packages/pds/src/api/app/bsky/feed/getPostThread.ts
+++ b/packages/pds/src/api/app/bsky/feed/getPostThread.ts
@@ -41,6 +41,12 @@ export default function (server: Server, ctx: AppContext) {
           if (!rev) throw err
 
           const uri = new AtUri(params.uri)
+          if (!uri.hostname.startsWith('did:')) {
+            const account = await ctx.accountManager.getAccount(uri.hostname)
+            if (account) {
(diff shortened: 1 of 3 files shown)


packages/pds/src/api/app/bsky/feed/getPostThread.ts
```diff
@@ -41,6 +41,12 @@ export default function (server: Server, ctx: AppContext) {
           if (!rev) throw err
 
           const uri = new AtUri(params.uri)
+          if (!uri.hostname.startsWith('did:')) {
+            const account = await ctx.accountManager.getAccount(uri.hostname)
+            if (account) {
+              uri.hostname = account.did
+            }
+          }
           if (uri.hostname !== requester) throw err
```
gaearon, line 50: this check would previously lead to 404

dholms (approved):
yup yup this looks great 👌

PR #2953, opened 2024-11-06
Add `opened` to chat.bsky.convo.defs#convoView

ghost:
Adds `opened` to `chat.bsky.convo.defs#convoView`.


devinivy (commented):
Looks good 👍 Let's add a changeset for the @atproto/api package so that there will be a version with the new types.

devinivy (approved):

Issue #2952, opened 2024-11-06
Generation of `Omit` in lex-cli types causes loss of type safety


tom-sherman:
**Describe the bug**

Methods like `create` that accept an `InputSchema` lose their type safety and accept any object value.

**To Reproduce**

```ts
client.app.bsky.feed.post.create(
    // @ts-expect-error Should be missing repo here
    {},
    post
)

client.app.bsky.feed.post.delete(
    // @ts-expect-error Should be missing rkey and repo here
    {}
)
```

**Expected behavior**

Type errors should happen above. See [playground](https://www.typescriptlang.org/play/?ssl=19&ssc=2&pln=10&pc=1#code/JYWwDg9gTgLgBAbwIIzAIQIYGcCmBhAG2BwDsYAaOJMdLAawE8AxHHAEwAUIsYBfOAGZQIIOACIAAhlTCYEAPQYwwMQChVbHAGMCGKDjhaIJHoaKkYALiqpMuQsTIBuVUZPxIpgLyJVcf4b60uwo1mJi5H4BMDgAHlbiarwu6jqOMAB0SmAZAEb0DBkCrGwZnplaQTEAFFH+8vJwEjBYALRxYNow7VDCUHAAygAWEACuBGxwuQYgwFhYwCQA5nD6kHBDOPp1iLyRAXDlqgCUqeZkWTR5BUUlZdyZmgQ4NTsNTS3tsZ1a3Vt9gxG40m0zgs3mixWUDoOAYcAwJEmawgGy2OB2CF4JyAA).

**Details**

This is caused by the use of `Omit` on a type with an index signature. This is expected behaviour of typescript see https://github.com/microsoft/TypeScript/issues/45367

**Additional context**

tom-sherman:
I think changing the Omit to be an intersection with the property set to unknown could work here eg.

```diff
- Omit<InputSchema, "collection">
+ InputSchema & { collection?: unknown }
```

tom-sherman:
Hmm, the suggestion above doesn't work actually. Not sure what the workaround is then.

matthieusieben:
Basically have no other choice than removing the `[x: string]: unknown` index signature from the `InputSchema` interface.

Doing so causes a bunch of breaking changes though :/ (see related PR)

tom-sherman:
@matthieusieben looks like there's a solution in #2968 🙌

PR #2956, opened 2024-11-07
Add explicit $type property in code gen

matthieusieben:
Part of https://github.com/bluesky-social/atproto/pull/2999
Fixes https://github.com/bluesky-social/atproto/issues/2952

### Breaking changes

This PR changes the interfaces of the generated code in the following ways:

- it removes the `[x: string]: unknonw` index signature in objects & records. Because of this, access to un-speccified fields must be explicit.
- it makes the `$type` property explicit (though optional, except in open unions). This allows to properly discriminate entity types, especially when used in open unions.
- the `isMyType(value)` utilities no longer type cast the `value` being checked. They only type the presence of the `$type` property. Use `isValidMyType` to assert the type through validation.

> [!NOTE]
> Using `main` instead of `msi/lex-gen-optimization` (https://github.com/bluesky-social/atproto/pull/2999) as "base branch" to allows CI to run.

> [!NOTE]
> While reviewing, "codegen" commits can be ignored ;-)


packages/dev-env/src/moderator-client.ts
```diff
@@ -60,19 +60,18 @@ export class ModeratorClient {
...
     const {
       event,
       subject,
       subjectBlobCids,
-      reason = 'X',
       createdBy = 'did:example:admin',
     } = opts
     const result = await this.agent.tools.ozone.moderation.emitEvent(
-      { event, subject, subjectBlobCids, createdBy, reason },
+      { event, subject, subjectBlobCids, createdBy },
```
matthieusieben, line 74: This does not seem to be used in the ozone api. Is this used elsewhere ?

packages/api/src/agent.ts
```diff
@@ -61,15 +61,6 @@ const THREAD_VIEW_PREF_DEFAULTS = {
...
 
-declare global {
-  interface Array<T> {
-    findLast(
-      predicate: (value: T, index: number, obj: T[]) => unknown,
-      thisArg?: any,
-    ): T
-  }
-}
-
```
matthieusieben, line 72: This is not the proper type def for this method. The correct signature is provided by ts's 2023 lib.

Since we are relying on this method, I updated the tsconfig to use 2023 lib instead of 2022.

packages/bsky/src/views/index.ts
```diff
@@ -1049,45 +1113,56 @@ export class Views {
     } else if (parsedUri.collection === ids.AppBskyFeedGenerator) {
       const view = this.feedGenerator(uri, state)
       if (!view) return this.embedNotFound(uri)
-      view.$type = 'app.bsky.feed.defs#generatorView'
-      return this.recordEmbedWrapper(view, withTypeTag)
+      return this.recordEmbedWrapper(
+        { ...view, $type: 'app.bsky.feed.defs#generatorView' },
+        withTypeTag,
+      )
```
matthieusieben, line 1119: This is a good example of the benefit of this change. Before, because `$type` was optional, the type system did not force the `$type` property to be present in unions.

Now, because it is no longer optional in unions, not specifying the `$type` here results in a type error.

tsconfig/browser.json
```diff
@@ -2,7 +2,7 @@
   "$schema": "https://json.schemastore.org/tsconfig",
   "extends": "./base.json",
   "compilerOptions": {
-    "lib": ["ES2022", "DOM", "DOM.Iterable", "ESNext.Disposable"],
+    "lib": ["ES2023", "DOM", "DOM.Iterable", "ESNext.Disposable"],
```
matthieusieben, line 5: NodeJS 18 (our minimal version) uses ES2023.

The `api` package already relies on ES2023 features (`Array.findLast`).

matthieusieben:
Keeping this as a separate PR from #2999 makes these changes hard to maintain. Because of this, I'm making the changes from this PR as part of #2999.

PR #2957, opened 2024-11-07
Resolve facets in parallel

gaearon:
There's no great reason for these to be a sequential client-server waterfall, let's grab them all together.

Ref: https://bsky.app/profile/aaronmcadam.com/post/3laevl2d33y2h

## Before

Screenshot 2024-11-07 at 18 46 00

## After

Screenshot 2024-11-07 at 18 49 45

### DIFF
--- a/packages/api/src/rich-text/rich-text.ts
+++ b/packages/api/src/rich-text/rich-text.ts
@@ -350,17 +350,23 @@ export class RichText {
   async detectFacets(agent: AtpBaseClient) {
     this.facets = detectFacets(this.unicodeText)
     if (this.facets) {
+      const promises: Promise<void>[] = []
       for (const facet of this.facets) {
         for (const feature of facet.features) {
           if (AppBskyRichtextFacet.isMention(feature)) {
-            const did = await agent.com.atproto.identity
-              .resolveHandle({ handle: feature.did })
-              .catch((_) => undefined)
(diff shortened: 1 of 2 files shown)


packages/api/src/rich-text/rich-text.ts
```diff
@@ -350,17 +350,23 @@ export class RichText {
...
-            const did = await agent.com.atproto.identity
-              .resolveHandle({ handle: feature.did })
-              .catch((_) => undefined)
-              .then((res) => res?.data.did)
-            feature.did = did || ''
+            promises.push(
+              agent.com.atproto.identity
+                .resolveHandle({ handle: feature.did })
+                .then((res) => res?.data.did)
+                .catch((_) => undefined)
```
gaearon, line 361: Moved this below for clarity (it's supposed to protect the last step)

packages/api/src/rich-text/rich-text.ts
```diff
@@ -350,17 +350,23 @@ export class RichText {
...
+                .then((did) => {
+                  feature.did = did || ''
+                }),
+            )
           }
         }
       }
+      await Promise.allSettled(promises)
```
gaearon, line 369: these can't fail atm but this better reflects what we want

devinivy (approved):
Good call 👍

PR #2958, opened 2024-11-07
bsky,pds: sharp 0.32.6 -> 0.33.5

pyrox0:
Upgrades sharp from `0.32.6` to `0.33.5`. This brings a lot of improvements, notably [libvips 8.15](https://www.libvips.org/2023/10/10/What's-new-in-8.15.html), which brings quite a lot of performance improvements. It also includes prebuilt binaries of libvips and sharp, which is helpful for cases when there is no support for compiling vips/sharp from source.

I've run the tests in `packages/bsky/tests/image/sharp.test.ts`, and all of them pass as expected.

### DIFF
--- a/packages/bsky/package.json
+++ b/packages/bsky/package.json
@@ -56,7 +56,7 @@
     "pg": "^8.10.0",
     "pino": "^8.21.0",
     "pino-http": "^8.2.1",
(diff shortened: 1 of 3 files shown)


devinivy (approved):
Thanks!

PR #2989, opened 2024-11-13
Return empty cursor if feed echoes input cursor

rafaelbsky:
If a custom feed generator had buggy/malicious code that echoed back to the response cursor the same value as the input cursor, this would cause an infinite loop for a client trying to paginate.

This returns an empty cursor if the returned cursor from the feed generator is the same as the one passed to it. That will effectively stop the pagination.


packages/bsky/src/api/app/bsky/feed/getFeed.ts
```diff
@@ -177,7 +177,7 @@ const skeletonFromFeedGen = async (
 ): Promise<AlgoResponse> => {
   const { feed, headers } = params
   const found = await ctx.hydrator.feed.getFeedGens([feed], true)
-  const feedDid = await found.get(feed)?.record.did
+  const feedDid = found.get(feed)?.record.did
```
devinivy, line 180: Good find!

packages/bsky/src/api/app/bsky/feed/getFeed.ts
```diff
@@ -220,6 +220,16 @@ const skeletonFromFeedGen = async (
...
       },
     )
+
+    const { cursor: resultCursor } = result.data
+    if (params.cursor && resultCursor === params.cursor) {
+      // Prevents loops in custom feeds.
+      throw new UpstreamFailureError(
+        'feed provided an invalid response',
+        'UpstreamFailureError',
+      )
```
devinivy, line 230: How would you feel about failing open instead?  If the cursor is the same, we treat it as empty rather than failing.

devinivy (approved):

PR #2990, opened 2024-11-13
Ensure RAW returns empty quoteCount

estrattonbailey:
Some folks noticed during the outage earlier today that quoteCount wasn't returned on responses for recently made posts, since those responses were served directly by the PDS via RAW and not the App View (bc outage). This was causing some rendering errors.

So this PR ensures that if we fall back to RAW, the response matches what's expected from `getPostThread`. Basically, `quoteCount` should not be `undefined`.

### DIFF
--- a/.changeset/fuzzy-wolves-speak.md
+++ b/.changeset/fuzzy-wolves-speak.md
@@ -1,0 +1,5 @@
+---
+"@atproto/pds": patch
+---
+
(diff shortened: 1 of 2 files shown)


gaearon (approved):
this makes sense to me, i guess would need a changeset? also may need a stamp from @devinivy or @dholms

dholms (approved):
yup yup this LGTM 👍

PR #2992, opened 2024-11-14
Sync: pass through websocket opts

dholms:


devinivy (approved):

packages/sync/src/firehose/index.ts
```diff
@@ -14,6 +14,7 @@ import {
 } from '@atproto/repo'
 import { AtUri } from '@atproto/syntax'
 import { Subscription } from '@atproto/xrpc-server'
+import { ClientOptions } from 'ws'
```
matthieusieben, line 17: You should probably use an `import type` here to avoid runtime side effects

PR #2999, opened 2024-11-14
Improve codegen typings

matthieusieben:
This PR is the first of a 2 steps changes aimed at improving typing in the atproto repo:

- https://github.com/bluesky-social/atproto/pull/2999 Adds a `$type` property to records and user object types. Improves the typing of isX and validateX helpers.
- https://github.com/bluesky-social/atproto/pull/2955 forbids "implicit any" in trypescript

### Breaking changes

This PR changes the interfaces of the generated code in the following ways:

- it removes the `[x: string]: unknown` index signature in objects & records. Because of this, access to un-speccified fields must be explicit.
- it makes the `$type` property explicit (though optional, except in open unions). This allows to properly discriminate entity types, especially when used in open unions.
- it makes the `$type` property mandatory in record types.
- the `isMyType(value)` utilities no longer type cast the `value` being checked. They only type the presence of the `$type` property.
- validating that an unknown value actually matches a lexicon schema can be achieved using the `validateMyType(value)` utilities.
- the `validateMyType` utilities now require that the `$type`, if present, is valid.
- The new "`asPredicate`" helper function allows to convert a `validateMyType` utility into a type predicate like so: `const isValidMyType = asPredicate(validateMyType)`

> [!NOTE]
> When reviewing, "codegen" commits can be ignored as they only contain generated code changes.

Fixes https://github.com/bluesky-social/atproto/issues/2952


packages/api/src/client/util.ts
```diff
@@ -1,13 +1,35 @@
...
+): $type is $Type<Id, Hash> {
+  return (
+    typeof $type === 'string' &&
+    ($type === id
+      ? hash === 'main'
+      : // $type === `${id}#${hash}`
+        $type.length === id.length + 1 + hash.length &&
+        $type[id.length] === '#' &&
+        $type.startsWith(id) &&
+        $type.endsWith(hash))
```
devinivy, line 25: For workloads where `check$type()` is usually `true` it's not clear to me that this is going to have the intended effect of a performance improvement over the simpler version.  Or is the idea specifically to avoid allocating the additional string?

packages/lexicon/src/validators/complex.ts
```diff
@@ -19,25 +19,46 @@ export function validate(
   def: LexUserType,
   value: unknown,
 ): ValidationResult {
+  // Order based on the number of occurrences in the Bluesky official lexicons:
```
devinivy, line 22: Does this have a performance impact?  I wouldn't be surprised if V8 optimizes this into a simple O(1) lookup rather than running through each case.

packages/lexicon/src/types.ts
```diff
@@ -447,23 +447,13 @@ export function isValidLexiconDoc(v: unknown): v is LexiconDoc {
   return lexiconDoc.safeParse(v).success
 }
 
-export function isObj(obj: unknown): obj is Record<string, unknown> {
-  return obj !== null && typeof obj === 'object'
+export function isObj<V>(v: V): v is V & object {
+  return v != null && typeof v === 'object'
```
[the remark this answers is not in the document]
matthieusieben: `==` and `!=` are fine to use with `null` (this is the only case I *ever* use double equals).

`== null` is the same as `=== null OR === undefined`
`!= null` is the same as `!== null AND !== undefined`

We should use the [eqeqeq](https://eslint.org/docs/latest/rules/eqeqeq) eslint rule with the following config `["error", "always", {"null": "ignore"}]` and be done with it :-)

> BTW, are we fine with isObj returning true for arrays? I

In this case, and as the type predicate indicates (`V & object`), this is indeed the expected behavior.

packages/api/src/client/util.ts
matthieusieben: The primary goal is to avoid string allocation.

The fact that the order of operation can be optimized for the `false` case is an added benefit.

packages/lexicon/src/validators/complex.ts
matthieusieben: I don't know that SpiderMonkey and V8 optimize switch statements into lookup tables. I kind of assumed they don't since I read [this article](https://artemis.sh/2022/08/07/emulating-calculators-fast-in-js.html) a while back. But things have probably changed since I read that.

The main point is to change a seemingly random ordering here into something that makes sense, while potentially benefiting from better performances.

matthieusieben: re-reading the article now (in particular its conclusion) I realize that I should always rely on my experience & memory 🙃

Makefile
```diff
@@ -28,10 +28,7 @@ run-dev-env-logged: ## Run a "development environment" shell (with logging)
 
 .PHONY: codegen
 codegen: ## Re-generate packages from lexicon/ files
-	cd packages/api; pnpm run codegen
-	cd packages/pds; pnpm run codegen
-	cd packages/bsky; pnpm run codegen
-	cd packages/ozone; pnpm run codegen
+	pnpm run --parallel codegen
```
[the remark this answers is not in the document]
matthieusieben: https://github.com/bluesky-social/atproto/pull/2999/commits/fb16f405a1d42f639a52f5293512064180c593b0

packages/lexicon/src/validators/complex.ts
matthieusieben: https://github.com/bluesky-social/atproto/pull/2999/commits/617d86ca475c3ce68c27e694f22189c1628a4388#diff-7d6f311165a765285060eb4c8944d5b029c4a4d48ce8dd636d86b37f50644afd

devinivy (approved):

PR #3002, opened 2024-11-14
Lowercase uri before checking for link

JustSNguyen:
The below method will fail for situations like: Https://google.com, cnn.COM even though these strings should be considered as valid urls.

### DIFF
--- a/packages/api/src/rich-text/detection.ts
+++ b/packages/api/src/rich-text/detection.ts
@@ -41,7 +41,7 @@ export function detectFacets(text: UnicodeString): Facet[] | undefined {
     // links
     const re = URL_REGEX
     while ((match = re.exec(text.utf16))) {
-      let uri = match[2]
+      let uri = match[2].toLowerCase()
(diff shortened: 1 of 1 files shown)


packages/api/src/rich-text/detection.ts
```diff
@@ -41,7 +41,7 @@ export function detectFacets(text: UnicodeString): Facet[] | undefined {
     // links
     const re = URL_REGEX
     while ((match = re.exec(text.utf16))) {
-      let uri = match[2]
+      let uri = match[2].toLowerCase()
```
khuddite, line 44: I think it will introduce another problem because URI is now manipulated, and it's a return value of this function.

JustSNguyen: Oh good point ! I think we can create a separate variable to check if the rawUri is a valid url. If yes, we return the rawUri instead.

michaelwschultz:
Run into this all the time on mobile. Hope this makes it in soon. Thanks for the contribution. 💪

PR #3015, opened 2024-11-15
Appview: control max thread depth

devinivy:
This gives the appview some controls over the max thread depth that it supports.


dholms (approved):
excellent

PR #3033, opened 2024-11-18
Fix missing delete blocks

dholms:
Some events that we were sending out over the PDS firehose were missing blocks necessary to validate the included proofs. This seems to have been primarily happening on events that contained a single delete, but hypothetically could happen on any write.

It seems like this was being caused by a reorganization of some blocks in the tree. Such that a block that was in the proof chain for delete in question _existing previously_ in the tree. The differ would not return it as a new block in that case which is strictly true but not sufficient for a non-fully-authorized relay consumer.

This adds in the notion of "relevant blocks" to commit data as well. This include exactly the blocks that make up the proofs for the included operations. 

This is an overlapping set with new blocks (often the same set). But "new blocks" can contain blocks not necessary for the included proof & similarly relevant blocks can include already existing blocks that are relevant to the proof.

When sending an event from the PDS, we send the union of these two block maps.


DavidBuchanan314:
Would be good to get some test cases for this.

I could see it happening when

1. the top node of the tree contains a single key

2. that single key is deleted

(overall, 1 block has been deleted, and the new tree has 0 "new" blocks)

devinivy (approved):
Looks good.  Probably want to add some changesets for both @atproto/repo and @atproto/pds.

packages/pds/src/sequencer/events.ts
```diff
@@ -22,11 +22,15 @@ export const formatSeqCommit = async (
   const blobs = new CidSet()
   let carSlice: Uint8Array
 
+  const blocksToSend = new BlockMap()
+  blocksToSend.addMap(commitData.newBlocks)
+  blocksToSend.addMap(commitData.relevantBlocks)
```
devinivy, line 27: Can we omit `commitData.newBlocks` here?  If I understood correctly, the way things are currently written every "new" block is "relevant".  Or perhaps this is just a way to be defensive around missing any blocks.

dholms: This will de-duplicate fwiw so we won't actually be sending along double blocks.

The way it's currently written "new blocks" are not all necessarily "relevant". Maybe a better word for "relevant blocks" would be "proof blocks". 

As an example take this edge case of a node split from the MST tests: https://github.com/bluesky-social/atproto/blob/main/packages/repo/tests/mst.test.ts#L328

The newly added record (`f`) has very few blocks that are relevant to it's proof (just the root of the tree). However the new MST has many new blocks in it to compensate for the split

PR #3041, opened 2024-11-19
Add back-pressure on subscriptions

matthieusieben:
This PR enables buffering of subscription's `handler` stream.

In the current implementation, the handler iterator will only generate new items when the previous one was actually sent down the wire.

The new implementation allows to control buffering using `highWaterMark` and `lowWaterMark`, which will allow for a smoother data flow.


packages/xrpc-server/src/server.ts
```diff
@@ -369,26 +368,26 @@ export class Server {
...
+
+                const type =
+                  // "<nsid>#<type>" -> "#<type>"
+                  $type.charCodeAt(nsid.length) === 35 /* "#" */ &&
+                  $type.startsWith(nsid) &&
+                  !$type.includes('#', nsid.length + 1) // no more "#" afterwards
+                    ? $type.slice(nsid.length) // strip nsid (keep "#")
+                    : $type
+
+                yield new MessageFrame(body, { type })
```
matthieusieben, line 387: This change:
- avoids using zod
- avoids clone + delete (performs all in one single operation)
- optimizes `type` transformation (avoid un-necessary array creation)

packages/xrpc-server/src/stream/server.ts
```diff
@@ -9,38 +10,53 @@ export class XrpcStreamServer {
   constructor(opts: ServerOptions & { handler: Handler }) {
     const { handler, ...serverOpts } = opts
     this.wss = new WebSocketServer(serverOpts)
-    this.wss.on('connection', async (socket, req) => {
-      socket.on('error', (err) => logger.error(err, 'websocket error'))
+    this.wss.on('connection', async (ws, req) => {
```
matthieusieben, line 13: Renamed to `ws` to avoid confusion with the internal `_socket` property (`socket._socket` feels wierd)

packages/xrpc-server/src/stream/server.ts
```diff
@@ -9,38 +10,53 @@ export class XrpcStreamServer {
...
               if (err) return rej(err)
-              res(undefined)
+
+              // If the sender buffer is full, we need to wait for the drain event
+              if (duplex?.writableNeedDrain) {
+                // @TODO: Should we setup a termination timeout here ?
+                duplex.once('drain', res)
+              } else {
+                res()
+              }
```
matthieusieben, line 43: This is the real change that will avoid filling the PDS's memory when the downstream service is too slow to consume.

PR #3043, opened 2024-11-19
[PDS] Handles: Update rules to allow using root domain

Javinator9889:
This PR is a continuation of https://github.com/bluesky-social/atproto/pull/2227. Here, I intend to pick those changes that may be needed to allow the usage of root domains as custom handles in the BlueSky network.

The changes introduced verify:
* A handle is valid iff it has the same subdomains as the host.
* It ends with the given host.

The problem with the current implementation is the `handle === host.slice(1)` prevents users owning a PDS from effectively using their own domain as a handle - although this is possible throughout the BlueSky network. When verifying the handle through the UI, the request ends up reaching the PDS server instead of only resolving the DNS records.

The PDS server verifies if the handle is supported. As we're using the root domain, the answer is always "Yes", to which the PDS simply responds with a bare `Unable to resolve handle` - it was expecting a known handle to return the `DID` but it is not prepared to try to register a new one.

By changing how the checking is done, it is possible to skip that validation and let the code verify the handle points to a DID by querying the DNS records and the `.well-known` endpoint, as written in https://github.com/bluesky-social/atproto/blob/828f17f53a34fd1194cb37beaa0cc8db42023717/packages/identity/src/handle/index.ts#L17-L34

### DIFF
--- a/packages/pds/tests/handles.test.ts
+++ b/packages/pds/tests/handles.test.ts
@@ -44,7 +44,7 @@ describe('handles', () => {
       dbPostgresSchema: 'handles',
       pds: {
         serviceHandleDomains: ['.test', '.example.com'],
-      }
+      },
     })
     // @ts-expect-error Error due to circular dependency with the dev-env package
     ctx = network.pds.ctx


synace:
Is there a workaround until approved?

Javinator9889:
> Is there a workaround until approved?

From what I've seen @synace it's not possible. I was thinking of migrating data from BlueSky to the PDS with `goat` but even that requires a temporary name and a change afterwards.

So looks like a PDS with the root handle as the account name is not possible...

ledlamp:
> Is there a workaround until approved?

Workaround is to hack the code to bypass the check. I go inside docker container, grep for the error message, edit the `dist` code and restart

nbx3:
Resolves the issue for the short term thanks

PR #3056, opened 2024-11-20
addReservedHandle lexicon

ghost:
Adds lexicon to `com.atproto.temp.addReservedHandle` to `@atproto/api`.

Will be used by https://github.com/bluesky-social/entryway/pull/88.

### DIFF
--- a/packages/api/src/client/lexicons.ts
+++ b/packages/api/src/client/lexicons.ts
@@ -3978,6 +3978,13 @@ export const schemaDict = {
             },
           },
         },
+        output: {
+          encoding: 'application/json',
+          schema: {
+            type: 'object',
+            properties: {},
+          },
+        },
       },
     },
   },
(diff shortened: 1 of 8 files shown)


lexicons/com/atproto/temp/addReservedHandle.json
```diff
@@ -0,0 +1,20 @@
+{
+  "lexicon": 1,
+  "id": "com.atproto.temp.addReservedHandle",
+  "defs": {
+    "main": {
+      "type": "procedure",
+      "description": "Add a handle to the set of reserved handles.",
+      "input": {
+        "encoding": "application/json",
```
devinivy, line 9: I recommend having an `output` as well, even if its schema has no fields in it.  This makes it easier to adapt the endpoint in a backwards-compatible way if it requires some output in the future.  For this reason we've started to establish it as a lexicon best practice.

devinivy (commented):
Just the one small piece of feedback, otherwise looks good!

PR #3066, opened 2024-11-21
OAuth url scheme validation

matthieusieben:


devinivy (approved):

packages/oauth/oauth-types/src/oauth-redirect-uri.ts
```diff
@@ -0,0 +1,55 @@
...
+      // > "http://localhost:{port}/{path}") function similarly to loopback IP
+      // > redirects described in Section 7.3, the use of localhost is NOT
+      // > RECOMMENDED.  Specifying a redirect URI with the loopback IP literal
+      // > rather than localhost avoids inadvertently listening on network
+      // > interfaces other than the loopback interface.  It is also less
+      // > susceptible to client-side firewalls and misconfigured host name
+      // > resolution on the user's device.
+      ctx.addIssue({
+        code: ZodIssueCode.custom,
+        message: 'Use of "localhost" hostname is not allowed (RFC 8252)',
```
devinivy, line 24: If this is a likely common error, it may be worth suggesting that they use a loopback IP literal here in the error message.
```suggestion
        message: 'Use of "localhost" hostname is not allowed (RFC 8252), use a loopback IP such as 127.0.0.1 instead',
```

packages/oauth/oauth-types/src/oauth-protected-resource-metadata.ts
```diff
@@ -10,8 +11,20 @@ export const oauthProtectedResourceMetadataSchema = z.object({
...
    * well-known resources is described in Section 3.
+   *
+   * @note This schema allows non https URLs for testing & development purposes.
+   * Make sure to validate the URL before using it in a production environment.
    */
-  resource: z.string().url(),
+  resource: webUriSchema
+    // .refine((url): url is `https://${string}` => url.startsWith('https://'), {
+    //   message: 'Resource URL must use the https scheme',
+    // })
```
devinivy, line 21:
```suggestion
```

devinivy: This may have been left on purpose with the added note, but it's not obvious that the note is connected to it.  So it ends-up reading as dead code or code that was mistakenly commented-out.  I think the note on its own is sufficient 👍

devinivy (approved):
This looks great, thanks.

PR #3070, opened 2024-11-22
:sparkles: Allow querying events by multiple keywords using OR condition

foysalit:
This PR adds a special character `||` in the comment param for `tools.ozone.moderation.queryEvents` endpoint which allows moderators to query events with certain keywords that are used with the `OR` operator.


matthieusieben (approved):
The changes look good. Might need another lex-gen because the description doesn't match

PR #3091, opened 2024-11-23
Add PDS_MANUAL_CERTS env set

jamesmanes:
**NOTE: Second PR on the PDS side to leverage this can be found here:** https://github.com/bluesky-social/pds/pull/155

## Rationale

PDS administrators may want to leverage caddy for on-demand TLS rather than hack around with nginx or another server. 

## Approach

With the introduction of a PDS_MANUAL_CERTS env property, PDS administrators can add a comma-separated list of subdomains (e.g., j.manes.xyz, portfolio.manes.xyz, work.manes.xyz). 

## Example

I own https://manes.xyz. My handle is james.manes.xyz. I used this domain in the past for a personal landing page at https://j.manes.xyz. I want a cert to exist for j.manes.xyz so that I can properly redirect it or proxy-pass it in the caddy config file without having to replace caddy with nginx or apache.

### DIFF
--- a/packages/pds/src/config/env.ts
+++ b/packages/pds/src/config/env.ts
@@ -17,6 +17,7 @@ export const readEnv = (): ServerEnvironment => {
     acceptingImports: envBool('PDS_ACCEPTING_REPO_IMPORTS'),
     blobUploadLimit: envInt('PDS_BLOB_UPLOAD_LIMIT'),
     devMode: envBool('PDS_DEV_MODE'),
+    manualCertSet: new Set(envStr('PDS_MANUAL_CERTS')?.split(',')),
 
     // branding
     brandColor: envStr('PDS_PRIMARY_COLOR'),


ledlamp:
What? This makes no sense. Just add your sites to the caddy config and it will get certs.

jamesmanes:
> What? This makes no sense. Just add your sites to the caddy config and it will get certs.

Can you offer an example? Adding sites to the caddy config was not enough for me, as there is a check in the index.js file that excludes subdomains that do not have PDS accounts associated with them.

ledlamp:
> > What? This makes no sense. Just add your sites to the caddy config and it will get certs.
> 
> Can you offer an example? Adding sites to the caddy config was not enough for me, as there is a check in the index.js file that excludes subdomains that do not have PDS accounts associated with them.

Put new site block without `tls { on-demand }` inside

jamesmanes:
> > > What? This makes no sense. Just add your sites to the caddy config and it will get certs.
> > 
> > 
> > Can you offer an example? Adding sites to the caddy config was not enough for me, as there is a check in the index.js file that excludes subdomains that do not have PDS accounts associated with them.
> 
> Put new site block without `tls { on-demand }` inside

This does not work. I added a new site block like so:

`b.manes.xyz {
        redir https://james.manes.xyz
}`

I also tried a block with a `reverse_proxy` rather than a `redir`, as I use that in one of my current configs as well. Navigating to this url results in SSL error, because no cert is being offered up for that subdomain. 

This PR would address this issue by allowing for certs to be generated for user-defined subdomains via allow-list.

ledlamp:
It should work. But it has to be unwrapped like this, I don't think caddyfile support one-liners.

```
b.manes.xyz {
    redir https://james.manes.xyz
}
```

Make sure to reload or restart caddy. Check caddy log to see if it is getting cert. It should not be asking PDS if it can get TLS cert for other sites that didn't enable on-demand TLS.

jamesmanes:
> It should work. But it has to be unwrapped like this, I don't think caddyfile support one-liners.
> 
> ```
> b.manes.xyz {
>     redir https://james.manes.xyz
> }
> ```
> 
> Make sure to reload or restart caddy. Check caddy log to see if it is getting cert. It should not be asking PDS if it can get TLS cert for other sites that didn't enable on-demand TLS.

I formatted my comment wrong, I was not using a one-liner. It was multi-line like what you have. This still doesn't work, though. There is a check that happens at `/tls-check` that prevents this:

https://github.com/bluesky-social/pds/pull/155/files#diff-2da25b4e54209ab9824c30fdf2c5f1c1be307c459c4071b31c966763143478afR47-R63

Essentially if this function does not return true, no cert will be generated / leveraged. It first checks if the domain is correct (e.g. manes.xyz). It then checks if the subdomain exists within PDS as an account (e.g. james). Both of these must be true for the tls-check to come back with a success response, which then lets the LTS cert get generated / served. 

With my approach in this PR, it also allows you to gen certs for other domains, too. For example I also host "manes.blog" on this same server along-side "manes.xyz." So I have caddy offering certs and handling proxy for many domains.

ledlamp:
> Yes, caddy should not ask for the cert with that on_demand block removed. I thought you were saying that by removing the on_demand block, I would get certs without the TLS check. My mistake.

Dude, that's the same thing.

What you want to do should have worked without hacking the PDS. Please open a bug report with caddy.

ledlamp:
I just tested it myself, with a PDS installed by their install script. I added a regular site to the end of the caddyfile:

2024-11-30_15-15-37 575_root@server_~

Caddy obtained cert on startup as expected:

2024-11-30_15-15-14 784_root@server_~

It works fine. No need to hack PDS.

2024-11-30_15-17-58 520_root@avps~

So I do not understand why this is not working for you.

ledlamp:
Could it be because the site you add is a subdomain covered by the wildcard of the other site? Explicit site names take precedence over wildcards, but maybe this behavior is affected when on-demand tls is enabled. I will test this later...

jamesmanes:
> Could it be because the site you add is a subdomain covered by the wildcard of the other site? Explicit site names take precedence over wildcards, but maybe this behavior is affected when on-demand tls is enabled. I will test this later...

There seems to be something here... I just tested this out by removing the wildcard cert ->

Old config snippet:
```
*.manes.xyz, manes.xyz {
	tls {
		on_demand
	}
	reverse_proxy http://localhost:3000
}
d.manes.xyz {
	respond "ok"
}```

Old output:
```
root@manes-inst:/pds/caddy/etc/caddy# curl -D - https://d.manes.xyz
curl: (35) error:0A000438:SSL routines::tlsv1 alert internal error
```

New config snippet:
```
manes.xyz {
	tls {
		on_demand
	}
	reverse_proxy http://localhost:3000
}
d.manes.xyz {
	respond "ok"
}```
```
root@manes-inst:/pds/caddy/etc/caddy# curl -D - https://d.manes.xyz
HTTP/2 200 
alt-svc: h3=":443"; ma=2592000
content-type: text/plain; charset=utf-8
server: Caddy
content-length: 2
date: Sun, 01 Dec 2024 00:58:04 GMT
```

Strange quirk. I'm going to dig a bit more soon and see if I can figure out exactly why this is going on.

jamesmanes:
Closing PR. Seems to be issue with Caddy and how it handles wildcard / on_demand.

ledlamp:
Well I tried this

2024-11-30_19-07-17 160_root@server_~

And it still worked as normal.

2024-11-30_19-06-43 990_root@server_~

Maybe you can try clearing the caddy data directory. Hope you figure it out. Not the first time I've encountered odd quirks with caddy that I couldn't reproduce for a bug report.

PR #3107, opened 2024-11-26
doc: fix typos in API docs

ptdewey:
Just fixed a few typos I noticed in the API docs.


ptdewey:
LICENSE.txt also has a similar grammatical mistake ("chose" instead of "choose"), but I didn't change it there because I was unsure if there is any special procedure for modifying license files.

PR #3108, opened 2024-11-26
Appview: fix takendown blocklist application on `actor.getProfile`

devinivy:
Takendown blocklists were taking effect on `actor.getProfile`, but properly not taking effect other places in the app.  This ensures the behavior is consistent on `actor.getProfile`.


dholms (approved):
nice 👌

thealamjahangir (approved):

PR #3113, opened 2024-11-26
Fix comment on atp-agent.ts for AtpAgent class

fsgreco:
Fix a wrong suggestion located inside AtpAgent comment, it had a wrong syntax:  
CredentialSession does not take a `{service: URL}`, it simply takes the URL as it's first argument (according to its source code).


SgtPooki:
I was just about to push up this fix. 

Image

Image

PR #3119, opened 2024-11-27
Add basic subscription listener structure

ghost:
This adds the basic structure to get webhook events from RevenueCat into appview.

Actually doing something with the data (storing it on the dataplane) will be done in a separate PR.

> [!IMPORTANT]
> I'm getting locally some ESM errors related to ozone when running `make run-dev-env`. I'm not 100% sure it is or is not related to this PR. Can you double-check running `make run-dev-env` on `main` and on this branch?

### DIFF
--- a/packages/bsky/src/subscriptions/revenueCat.ts
+++ b/packages/bsky/src/subscriptions/revenueCat.ts
@@ -34,7 +34,8 @@ export class RevenueCatClient {
     path: string,
     method: string = 'GET',
   ): Promise<T> {
-    const res = await fetch(`${this.url}${path}`, {
+    const url = new URL(path, this.url)
+    const res = await fetch(url, {
       method,
       headers: {
         Authorization: `Bearer ${this.apiKey}`,
--- a/packages/bsky/tests/subscriptions.test.ts
+++ b/packages/bsky/tests/subscriptions.test.ts
@@ -4,6 +4,8 @@ import http from 'node:http'
 import { once } from 'node:events'
 import { Entitlement, GetSubscriberResponse } from '../src/subscriptions'
 
+const revenueCatWebhookAuthorization = 'Bearer any-token'
+
 describe('subscriptions views', () => {
   let network: TestNetwork
   let sc: SeedClient
@@ -36,6 +38,7 @@ describe('subscriptions views', () => {
       bsky: {
         revenueCatV1ApiKey: 'any-key',
         revenueCatV1Url: `http://localhost:${revenueCatPort}`,
+        revenueCatWebhookAuthorization,
       },
     })
     bskyUrl = `http://localhost:${network.bsky.port}`
@@ -56,12 +59,37 @@ describe('subscriptions views', () => {
     const createdAt = new Date().toISOString()
     const updatedAt = new Date().toISOString()
 
+    it('returns 403 if authorization is missing', async () => {
+      const response = await fetch(`${bskyUrl}/webhooks/revenuecat`, {
+        method: 'POST',
+        body: JSON.stringify({ event: { app_user_id: alice } }),
+        headers: {
+          'Content-Type': 'application/json',
+        },
+      })
+
+      expect(response.status).toBe(403)
+    })
+
+    it('returns 403 if authorization is invalid', async () => {
+      const response = await fetch(`${bskyUrl}/webhooks/revenuecat`, {
+        method: 'POST',
+        body: JSON.stringify({ event: { app_user_id: alice } }),
+        headers: {
+          Authorization: `not ${revenueCatWebhookAuthorization}`,
+          'Content-Type': 'application/json',
+        },
+      })
+
+      expect(response.status).toBe(403)
+    })
+
     it('sets valid entitlements cache from the API response, excluding expired', async () => {
       await network.bsky.db.db
         .insertInto('subscription_entitlement')
         .values({
           did: alice,
-          entitlements: JSON.stringify(['entitlement0']),
+          entitlements: JSON.stringify(['oldEntitlement']),
           createdAt,
           updatedAt,
         })
@@ -71,7 +99,7 @@ describe('subscriptions views', () => {
         getUserSubscriptionEntitlement(network, alice),
       ).resolves.toStrictEqual({
         did: alice,
-        entitlements: ['entitlement0'],
+        entitlements: ['oldEntitlement'],
         createdAt,
         updatedAt,
       })
@@ -88,12 +116,8 @@ describe('subscriptions views', () => {
         res.json(response)
       })
 
-      await fetch(`${bskyUrl}/webhooks/revenuecat`, {
-        method: 'POST',
-        body: JSON.stringify({ event: { app_user_id: alice } }),
-        headers: {
-          'Content-Type': 'application/json',
-        },
+      await callWebhook(bskyUrl, {
+        event: { app_user_id: alice },
       })
 
       expect(
@@ -116,12 +140,8 @@ describe('subscriptions views', () => {
         res.json(response)
       })
 
-      await fetch(`${bskyUrl}/webhooks/revenuecat`, {
-        method: 'POST',
-        body: JSON.stringify({ event: { app_user_id: alice } }),
-        headers: {
-          'Content-Type': 'application/json',
-        },
+      await callWebhook(bskyUrl, {
+        event: { app_user_id: alice },
       })
 
       expect(
@@ -144,12 +164,8 @@ describe('subscriptions views', () => {
         res.json(response)
       })
 
-      await fetch(`${bskyUrl}/webhooks/revenuecat`, {
-        method: 'POST',
-        body: JSON.stringify({ event: { app_user_id: alice } }),
-        headers: {
-          'Content-Type': 'application/json',
-        },
+      await callWebhook(bskyUrl, {
+        event: { app_user_id: alice },
       })
 
       expect(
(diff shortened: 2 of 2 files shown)


packages/bsky/proto/bsky.proto
```diff
@@ -1393,3 +1396,15 @@ message ClearThreadMutesRequest {
...
 message ClearThreadMutesResponse {}
+
+// Subscription entitlement
+message SubscriptionEntitlement {
+  string did = 1;
+  repeated string entitlements = 2;
+}
+
+message SetSubscriptionEntitlementRequest {
+  SubscriptionEntitlement subscriptionEntitlement = 1;
```
devinivy, line 1407: I believe that we want to use `snake_case` here for consistency.

packages/bsky/proto/bsky.proto
```diff
@@ -1393,3 +1396,15 @@ message ClearThreadMutesRequest {
 }
 
 message ClearThreadMutesResponse {}
+
+// Subscription entitlement
+message SubscriptionEntitlement {
+  string did = 1;
+  repeated string entitlements = 2;
```
estrattonbailey, line 1403: Thoughts on adding a `updated_at` field here? Could be useful in the future for crons that sweep for outdated data.

packages/bsky/src/api/subscription-listener.ts
```diff
@@ -0,0 +1,72 @@
...
+      log.error(error)
+      throw error
+    }
+  }
+
+export const createRouter = (ctx: AppContext): express.Router => {
+  const router = express.Router()
+
+  if (!ctx.revenueCatClient) {
+    return router
```
estrattonbailey, line 63: Will webhooks fail if they hit this? I think we want them to fail, so just double checking.

That way, RC will retry them using their backoff algo. So if we forget this config we can fix it within the retry window and hopefully get back on track.

packages/bsky/src/index.ts
```diff
@@ -96,6 +102,15 @@ export class BskyAppView {
       )
     }
 
+    let revenueCatClient: RevenueCatClient | undefined
+    if (config.revenueCatV1ApiKey) {
```
estrattonbailey, line 106: Should we validate the webhook auth too?

packages/bsky/src/index.ts
```diff
@@ -170,6 +186,9 @@ export class BskyAppView {
     app.use(health.createRouter(ctx))
     app.use(wellKnown.createRouter(ctx))
     app.use(blobResolver.createRouter(ctx))
+    if (revenueCatClient) {
```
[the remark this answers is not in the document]
estrattonbailey: Just noting the question about what happens if config doesn't exist and we don't have an endpoint for RC.

Ideally we'd get some sort of alert if RC webhooks fail 🤔

packages/bsky/src/subscriptions/revenueCat.ts
```diff
@@ -0,0 +1,64 @@
...
+    this.apiKey = apiKey
+    this.url = url
+    this.webhookAuthorization = webhookAuthorization
+  }
+
+  private async fetch<T extends object>(
+    path: string,
+    method: string = 'GET',
+  ): Promise<T> {
+    const res = await fetch(`${this.url}${path}`, {
```
estrattonbailey, line 38: Nit: could use URL constructor for a little more stability

packages/bsky/src/subscriptions/revenueCat.ts
```diff
@@ -0,0 +1,64 @@
...
+  async getSubscriber(did: string): Promise<GetSubscriberResponse> {
+    return this.fetch<GetSubscriberResponse>(
+      `/subscribers/${encodeURIComponent(did)}`,
+    )
+  }
+
+  isWebhookAuthorizationValid(authorization: string | undefined): boolean {
+    // It is either valid if the authorization is:
+    //   1. not configured and undefined in the request.
+    //   2. configured and matches the request.
```
estrattonbailey, line 61: If we configure the token on our server, but the webhook event doesn't have it defined, then it also fails?

packages/bsky/src/subscriptions/util.ts
```diff
@@ -0,0 +1,19 @@
...
+  return Object.entries(subscriber.entitlements)
+    .filter(([_, entitlement]) => {
+      const expiresDate = new Date(entitlement.expires_date).valueOf()
+      const gracePeriodExpiresDate = entitlement.grace_period_expires_date
+        ? new Date(entitlement.grace_period_expires_date).valueOf()
+        : expiresDate
+
+      const expirationMs = Math.max(expiresDate, gracePeriodExpiresDate)
+
+      return now < expirationMs
```
estrattonbailey, line 16: This looks right to me, curious what RC will come back with regarding our questions

estrattonbailey (commented):
Looking good to me! Test makes sense, and so does expiration logic 👍

estrattonbailey (approved):
This looks good to me! Will let Daniel/Devin have final say though 👍

packages/bsky/src/api/subscription-listener.ts
```diff
@@ -0,0 +1,75 @@
...
+const auth =
+  (ctx: AppContextWithRevenueCatClient): RequestHandler =>
+  (req: express.Request, res: express.Response, next: express.NextFunction) =>
+    ctx.revenueCatClient.isWebhookAuthorizationValid(
+      req.header('Authorization'),
+    )
+      ? next()
+      : res
+          .status(403)
+          .send('Forbidden: invalid authentication for RevenuCat webhook')
```
dholms, line 22:
```suggestion
          .send('Forbidden: invalid authentication for RevenueCat webhook')
```

packages/bsky/src/api/subscription-listener.ts
```diff
@@ -0,0 +1,75 @@
...
+
+export const createRouter = (ctx: AppContext): express.Router => {
+  const router = express.Router()
+
+  if (!ctx.revenueCatClient) {
+    throw new Error(
+      'RevenueCat webhook was tried to be set up without providing a RevenueCat client.',
+    )
+  }
+  const ctxWithRevenueCatClient = ctx as AppContextWithRevenueCatClient
```
dholms, line 68: nit, but I might just use a typeguard for this rather than imperatively checking & then throwing or recasting

packages/bsky/src/data-plane/server/db/tables/subscription-entitlement.ts
```diff
@@ -0,0 +1,13 @@
+import { ColumnType } from 'kysely'
+
+export const tableName = 'subscription_entitlement'
+
+export interface SubscriptionEntitlement {
+  did: string
+  // https://github.com/kysely-org/kysely/issues/137
```
dholms, line 7: nice 👌 
thx for comment

packages/bsky/src/index.ts
```diff
@@ -96,6 +102,15 @@ export class BskyAppView {
       )
     }
 
+    let revenueCatClient: RevenueCatClient | undefined
+    if (config.revenueCatV1ApiKey && config.revenueCatWebhookAuthorization) {
```
dholms, line 106: Probably not a huge deal, but I'm a bit more inclined to hinge this off of the existence of just one variable (for instance the api key, or even a new "enabled" string) and then throw if other vairables are missing. Basically make it a bit less likely for silent failures

for instance
```ts
if(config.revenueCatV1ApiKey) {
  assert(config.revenueCatWebhookAuthorization)
  // ...
}
```

packages/bsky/src/subscriptions/util.ts
```diff
@@ -0,0 +1,12 @@
+import { Subscriber } from './revenueCat'
+
+export const entitlementIdentifiersFromSubscriber = (
```
dholms, line 3: I think we can make our revenue cat client a bit smarter and just move this logic into the `getSubscriber` method. Unless there's other info on response that we need elsewhere

packages/bsky/tests/subscriptions.test.ts
```diff
@@ -0,0 +1,224 @@
...
+        did: alice,
+        entitlements: [],
+        createdAt,
+        updatedAt: expect.any(String),
+      })
+    })
+  })
+})
+
+const createMockRevenueCatService = async (
```
dholms, line 183: good stuff 👌

packages/dev-env/src/bsky.ts
```diff
@@ -67,6 +67,15 @@ export class TestBsky {
       bigThreadUris: new Set(),
       ...cfg,
       adminPasswords: [ADMIN_PASSWORD],
+      revenueCatV1Url:
+        cfg.revenueCatV1Url ||
+        process.env.BSKY_REVENUE_CAT_V1_URL ||
```
dholms, line 72: Do we need to get these out of the process env for anything?

dholms: Ohhh is this for running the dev-env outside of tests where it actually talks to rev cat?

packages/bsky/tests/subscriptions.test.ts
```diff
@@ -0,0 +1,224 @@
...
+    network = await TestNetwork.create({
+      dbPostgresSchema: 'bsky_views_subscriptions',
+      bsky: {
+        revenueCatV1ApiKey: 'any-key',
+        revenueCatV1Url: `http://localhost:${revenueCatPort}`,
+        revenueCatWebhookAuthorization,
+      },
+    })
+    bskyUrl = `http://localhost:${network.bsky.port}`
+    network.plc.getClient().updateData
```
dholms, line 45: 🔪

dholms (approved):
looks great!

couple small things, mostly nits

PR #3120, opened 2024-11-28
Change default blob upload limit to 50 MiB

ledlamp:
https://github.com/bluesky-social/pds/pull/91 is not the right way to do it, it should be the default. PDSes installed with older version of install script don't get this change and people wonder why they can't upload videos on the PDS.


ledlamp:
Lexicons were updated to [allow 100MB (not MiB) videos](https://github.com/bluesky-social/atproto/pull/3602) so PDS default blob limit should probably be at least this.

PR #3126, opened 2024-11-28
Cache subscription data

ghost:
This builds on top of https://github.com/bluesky-social/atproto/pull/3119 to cache subscription data.

We listen for a RevenueCat webhook (https://github.com/bluesky-social/atproto/pull/3119) and set/unset the subscription entitlement on the data plane.

I was only able to test the "set" part because I wasn't able to trigger the events where the user doesn't have a subscription (like on expiration). I'll pair up with @estrattonbailey to test that soon.


estrattonbailey (commented):
Looking good! We should work to adapt this to support multiple entitlements though. Right now a user will only receive one (which we'll then map to multiple granular features), but in the future we may introduce additional entitlements tied to separate purchasable products.

PR #3127, opened 2024-11-28
add @atproto/oauth-client-react-native

iameli:
Working OAuth client for React Native. Thanks to @haileyok for the first pass. [I've now published this under `@aquareum/atproto-oauth-client-react-native`](https://github.com/aquareum-tv/atproto-oauth-client-react-native) but will gladly discontinue that if this ends up getting upstreamed.

## The good
* Uses `react-native-quick-crypto`, has pretty fast generation of all of the supported signing algorithms except for ES256K.
* Uses jose for the signing flow, so all of those code paths are reused.
* The `redirect_uri` flow back to the `com.example:/login-return` redirect works great and it's cool to see it in action!
* Transparently falls back to `@atproto/oauth-client-browser` for the react-native-web case; you only need to use this library to support both platforms.

## The bad
* Keys are generated and stored in userspace. Ideally we'd want to lean on on-device security as much as possible for this stuff. A [project called react-native-device-crypto exists](https://github.com/arifaydogmus/react-native-device-crypto) but doesn't seem to see a lot of usage.
* Key storage utilizes `expo-sqlite` which won't be the choice of everyone using this library. You can inject your own storage systems by specifying a `stateStore` and a `sessionStore`, but this library still imports `expo-sqlite` so I think that'll still result in it getting linked into the app by default. I guess I could work around that by making it a separate opt-in export or something. Open to suggestions here.

## The ugly
So... getting jose and @atproto/oauth-client running in React Native was pretty rough. Here's the file full of polyfills that I included:
```typescript
import { Event, EventTarget } from 'event-target-shim'
import { install as installRNQC } from 'react-native-quick-crypto'

// Polyfill for the `throwIfAborted` method of the AbortController
// used in @atproto/oauth-client
import 'abortcontroller-polyfill/dist/polyfill-patch-fetch'

// Polyfill for jose. It tries to detect whether it's been passed a CryptoKey
// instance, and isn't willing to accept RNQC's equivalent. So, this ensures that
// `key instanceof CryptoKey` will always be true.
// @ts-ignore
global.CryptoKey = Object

// This is needed to populate the `crypto` global for jose's export here
// https://github.com/panva/jose/blob/1e8b430b08a18a18883a69e7991832c9c602ca1a/src/runtime/browser/webcrypto.ts#L1
installRNQC()

// These two are needed for @atproto/oauth-client's `CustomEventTarget` to work.
// @ts-ignore
global.EventTarget = EventTarget
// @ts-ignore
global.Event = Event

// And finally, this happens on React Native with every possible input:
// URL.canParse("http://example.com") => false
// I do not know why. Used in @atproto/oauth and @atproto/common-web
if (!URL.canParse('http://example.com')) {
  URL.canParse = (url: string | URL, base?: string) => {
    try {
      new URL(url, base)
      return true
    } catch (e) {
      return false
    }
  }
}
```
Yeesh. And this is with React Native 0.76.2, released two weeks ago! Half of these would go away if you dumped jose, but that would entail a bigger refactor of the other OAuth libraries that I can't do right now. The other necessary polyfills come from a few more modern web platform features like EventTarget and `throwIfAborted()`; I defer to the team on whether it'd be better to polyfill those or refactor them out.


iameli:
Feedback on building on @atproto/oauth-client: this is really well-organized and I love the `runtimeImplementation` bit so I can bring-my-own implementations of stuff. When I made [c2pa-go](https://github.com/aquareum-tv/c2pa-go), it was mede easier by the fact that they exposed the entire signing interface — the `sign()` function itself that ingested a data blob and did the hashing and signing. That made it really easy to implement things like hardware wallet support and PKCS11 for HSMs. It would be really great to see a similar pattern here, and would have made this much cleaner; I could have just used RNQC to sign the data directly instead of shoehorning in the `jose` adapter that I had to make.

iameli:
Closed in favor of https://github.com/haileyok/expo-atproto-auth

PR #3131, opened 2024-11-29
PDS support Deno and Bun by expanding Kysely dialect configuration options

### LINKED ISSUE #2866: Deno support
I tried to create an atproto app in Deno (using the Fresh framework) and it was completely impossible. I couldn't even get the OAuth working before I decided it wasn't worth it. It seems a lot of the atproto npm packages are built specifically for node.

I understand this is not at all a priority but I would be super grateful if the ATProto NPM packages (specifically OAuth and I assume many others) would decouple specifically from node so they could be used in Deno

**Describe alternatives you've considered**
Creating specific packages for Deno separately from the NPM packages.

**Additional context**

acrois:
Problem Summary:

`better-sqlite3` relies on Node internal native APIs and not general-purpose APIs. This means PDS can _only_ run in Node and _only_ use sqlite and _only_ this one dependency _despite_ the underlying library providing DB support (Kysely) being very flexible. PDS just does not allow you to pass that in very easily at the moment.

Proposed Solution:

- Eliminate explicit better-sqlite3 dependency to support non-Node runtimes.
- Support configuring PDS with more DB dialects, even non-sqlite.
  - [Deno](https://jsr.io/@soapbox/kysely-deno-sqlite)
  - [Bun](https://www.npmjs.com/package/kysely-bun-worker)
  - [Postgres](https://www.npmjs.com/package/kysely-postgres-js)
- Implement a standard fetch API example for PDS

Key changes:

1. [x] Create async accessor functions and refactor references to `this.db` to use `const connection = await this.getDb()`
2. [x] Replace direct dependency on `better-sqlite3` with a dynamic import inside of the static constructor.
3. [x] Refactor instances of "this.db.db" to "connection.db" which is easier to understand.
4. [ ] Solution for replacing `opentelemetry-plugin-better-sqlite3` in the tracer instrumentation configuration. **(needs help)**
5. [ ] Quick start usage of PDS which selects the most appropriate dependency based on operating environment and environmental configuration **(needs help)**
6. [ ] Documentation updates **(needs help)**
7. [ ] Testing **(needs help)**
8. [x] Standard fetch API example (separate - in the pds repo)

Summary:

This PR lays the groundwork in the core of PDS for using an alternative Kysely dialect, which will resolve #2866 

In addition, it enables easily exposing PDS via the standard fetch API (such as `deno serve`, Cloudflare Workers, AWS Lambda, etc.) which can reduce errors for first-time adopters. This will have a companion PR in the PDS repo.

Please let me know your feedback 🦃

### DIFF
--- a/packages/pds/package.json
+++ b/packages/pds/package.json
@@ -44,7 +44,6 @@
     "@atproto/xrpc-server": "workspace:^",
     "@did-plc/lib": "^0.0.4",
     "@hapi/address": "^5.1.1",
-    "better-sqlite3": "^10.0.0",
     "bytes": "^3.1.2",
     "compression": "^1.7.4",
     "cors": "^2.8.5",
--- a/packages/pds/src/account-manager/db/index.ts
+++ b/packages/pds/src/account-manager/db/index.ts
@@ -6,14 +6,14 @@ export * from './schema'
 
 export type AccountDb = Database<DatabaseSchema>
 
-export const getDb = (
+export const getDb = async (
   location: string,
   disableWalAutoCheckpoint = false,
-): AccountDb => {
+): Promise<AccountDb> => {
   const pragmas: Record<string, string> = disableWalAutoCheckpoint
     ? { wal_autocheckpoint: '0' }
     : {}
-  return Database.sqlite(location, { pragmas })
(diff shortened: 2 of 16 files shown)


knotbin:
Any progress on this?

PR #3142, opened 2024-11-30
Parse URLs after sentence dot correctly

tpaulshippy:
This is simply a copy of https://github.com/bluesky-social/atproto/pull/913 merged into the current main branch.
All credit should go to the author (@KevSlashNull) and the approver (@dholms).

I noticed that this one is:
A. Still a bug.
B. The oldest non-draft PR (1.5 years old).
C. Already approved.
D. Needing a merge.

Can we get it merged in?

-----
# Original PR description

This fixes the parsing of URLs that are prefixed by a dot (.), i.e. when the user forgot a space between the end of a sentence and a link.

image

The fix is done by preventing the second alternative capture group from having a domain end in .https:// (which would be an invalid domain anyway), since that would capture out.https in the example message Check this out.https://afterdot.com.

### DIFF
--- a/packages/api/src/rich-text/util.ts
+++ b/packages/api/src/rich-text/util.ts
@@ -1,6 +1,6 @@
 export const MENTION_REGEX = /(^|\s|\()(@)([a-zA-Z0-9.-]+)(\b)/g
 export const URL_REGEX =
-  /(^|\s|\()((https?:\/\/[\S]+)|((?<domain>[a-z][a-z0-9]*(\.[a-z0-9]+)+)[\S]*))/gim
+  /(^|\s|\(|\b)((https?:\/\/[\S]+)|((?<domain>[a-z][a-z0-9]*(?!\.https?:\/\/)(\.[a-z0-9]+)+)[\S]*))/gim
 export const TRAILING_PUNCTUATION_REGEX = /\p{P}+$/gu
 
 /**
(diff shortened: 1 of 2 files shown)


KevSlashNull:
thanks @tpaulshippy for trying to push this again!

lamrongol:
Perhaps I should create a new issue but I want to write about similar problem.
In Japanese, URLs after ["fullwidth character"](https://en.wikipedia.org/wiki/Halfwidth_and_fullwidth_forms) like "】" often appears and they are not parsed, either. So many Japanese users' URLs on profiles are not parsed. 
My  idea is:
Change the regexp of https://github.com/bluesky-social/atproto/blob/93bf69df96e1c7e292cf5f5a2eca564b8c4d8e7a/packages/api/src/rich-text/util.ts#L3
to
`/(^|[^a-zA-Z0-9])((https?:\/\/[\S]+)|((?<domain>[a-z][a-z0-9]*(\.[a-z0-9]+)+)[\S]*))/gim`

This also solves about "URLs after sentence dot" and other many pattern, and doesn't raise wrong behavior.
I have two questions:
1. Will this be applied to URLs on profiles?
2. Why is this PR left not merged?

lamrongol:
I've created extended PR https://github.com/bluesky-social/atproto/pull/3942

KevSlashNull:
@pfrazee :wave: could you please review one of the three open PRs trying to fix this bug? :bow: 

you already approved the original PR at https://github.com/bluesky-social/atproto/pull/913 over two years ago, so this one should be fine too :3

PR #3156, opened 2024-12-01
Add subscription routes

ghost:


lexicons/app/bsky/subscription/getActiveSubscriptions.json
```diff
@@ -0,0 +1,49 @@
...
+                "ref": "#subscription"
+              }
+            }
+          }
+        }
+      }
+    },
+    "subscription": {
+      "type": "object",
+      "required": [
```
dholms, line 27: one of the hard things about lexicon backwards compatibility is that you can't make required things optional

we should just double-check that that all of these fields will be on there forever

i'd err on the side of making them optional if we're not sure

lexicons/app/bsky/subscription/getSubscriptionFeatures.json
```diff
@@ -0,0 +1,33 @@
...
+              "type": "ref",
+              "ref": "#features"
+            }
+          }
+        }
+      }
+    },
+    "features": {
+      "type": "object",
+      "required": ["customProfileColor"],
```
dholms, line 24: i would def make this one optional

where `undefined` can just be falsy

packages/bsky/src/api/app/bsky/subscription/getSubscriptionFeatures.ts
```diff
@@ -0,0 +1,61 @@
...
+    customProfileColor: false,
+  }
+
+  if (subscriptionEntitlements?.length === 0) {
+    return gatedFeatures
+  }
+
+  const { entitlements } = subscriptionEntitlements[0]
+
+  return entitlements.reduce((acc, entitlement) => {
```
dholms, line 40: i totally get what you're doing here, and i'm not really opposed to it, but while we only have one subscription tier, i'm a bit more inclined to make it super straightforward and basically say

```ts
if(entitlements.includes('core') {
  return {
    customProfileColor: false
  }
} else {
  return {
    customProfileColor: true
  }
}
```

packages/bsky/src/data-plane/server/routes/subscription-entitlements.ts
```diff
@@ -3,6 +3,18 @@ import { Service } from '../../../proto/bsky_connect'
 import { Database } from '../db'
 
 export default (db: Database): Partial<ServiceImpl<typeof Service>> => ({
+  async getSubscriptionEntitlement(req) {
+    const { dids } = req
+
+    return {
+      subscriptionEntitlements: await db.db
```
dholms, line 10: we want to make sure these are returned in the same order as the params

and that the response as `undefined` as the array item that corresponds to each param that does not have entitlements

you can find similar logic in other dataplane routes such as in the `relationships` file

packages/bsky/proto/bsky.proto
```diff
@@ -1095,7 +1095,19 @@ message GetFollowsFollowingResponse {
 message PingRequest {}
 message PingResponse {}
 
+// Subscription entitlement
+message SubscriptionEntitlement {
+  string did = 1;
```
dholms, line 1100: I think I'd remove the `did` here and just rely on ordering of arrays as we do in some other examples (take `GetDidsByHandles` for example which does not return the handle in the response)

packages/bsky/proto/bsky.proto
```diff
@@ -1095,7 +1095,19 @@ message GetFollowsFollowingResponse {
...
+  repeated string entitlements = 2;
+}
 
+message GetSubscriptionEntitlementRequest {
```
dholms, line 1104: since we're fetching many, i would pluralize this
```suggestion
message GetSubscriptionEntitlementsRequest {
```

packages/bsky/tests/views/subscriptions.test.ts
```diff
@@ -0,0 +1,68 @@
...
+  let pdsAgent: AtpAgent
+  let sc: SeedClient
+
+  // account dids, for convenience
+  let alice: string
+  let bob: string
+
+  beforeAll(async () => {
+    network = await TestNetwork.create({
+      dbPostgresSchema: 'bsky_views_subscriptions',
```
dholms, line 17: we wanna make sure we use a different `dbPostgresSchema` for this test suite than the one we use in `tests/subscriptions.test.ts`

PR #3162, opened 2024-12-02
deps: bump express to 4.21.1 and bump compression to 1.7.5

bjohansebas:
From the package lock, there are vulnerable versions of Express that are susceptible to known vulnerabilities and can be exploited by any attacker.


bjohansebas:
I haven't updated Express in xrpc-server because there seems to be an error in the types of some package related to Express

PR #3164, opened 2024-12-02
:zap: Add composite index on reviewState and lastReportedAt columns

foysalit:
Dropping the index on `reviewState` column in favor of a composite index on `reviewState` and `lastReportedAt` columns since the `lastReportedAt` column is used as cursor and is present in almost all queries on the `moderation_subject_status` table.


foysalit:
We don't need this anymore. @dholms created a better index.

PR #3171, opened 2024-12-03
✨ Optionally ack all open subjects from the author with acknowledge event

foysalit:
This PR brings the same flag as takedown event from here https://github.com/bluesky-social/atproto/pull/2793 for the acknowledge event. This helps clean up moderation queue of unnecessary subjects that won't require individual reviews if the author's account is being acked.


matthieusieben (approved):
Looks good

PR #3173, opened 2024-12-04
PDS: fix account deactivation when behind entryway by using entryway auth

### LINKED ISSUE #3149: Can't deactivate account after migration, com.atproto.server.deactivateAccount BadJwtSignature jwt signature does not match jwt issuer
I migrated my account from a bluesky PDS to a third party PDS with the [goat](https://github.com/bluesky-social/indigo/tree/main/cmd/goat) tool, finishing the PLC migration [manually](https://whtwnd.com/bnewbold.net/3l5ii332pf32u) because my token expired.

Now I cannot deactivate the old account. I even tried manually with this script:

```js
fetch(`https://shiitake.us-east.host.bsky.network/xrpc/com.atproto.server.createSession`, {
	body: JSON.stringify({
		identifier: "(email)",
		password: "(password)"
	}),
	headers: {
		"Content-Type": "application/json"
	},
	method: "POST"
}).then(res => {
	console.log(res.status);
	res.json().then(json => {
		console.log(json);
		fetch(`https://shiitake.us-east.host.bsky.network/xrpc/com.atproto.server.deactivateAccount`, {
			body: "{}",
			headers: {
				"Content-Type": "application/json",
				"Authorization": "Bearer "+ json.accessJwt
			},
			method: "POST"
		}).then(res => {
			console.log(res.status);
			res.text().then(console.log);
		});
	});
});
```

And got output:

```
200
{
  did: 'did:plc:u4gngygg2w5egsigxu5g7byu',
  handle: 'lamp.bsky.social',
  email: '(email)',
  emailConfirmed: true,
  emailAuthFactor: false,
  accessJwt: '(token)',
  refreshJwt: '(token)',
  active: true
}
401
{"error":"BadJwtSignature","message":"jwt signature does not match jwt issuer"}
```

I thought leaving it as is would be ok but apparently it causes an issue where [other people on the same PDS can't reply to my newer posts that don't exist in this old account](https://github.com/bluesky-social/atproto/issues/3152), they get a post deleted error.

How to deactivate account?

devinivy:
Addresses an issue where a user attempts to deactivate their account after their signing key has changed in an account migration.  The problem is that the user's signing key changes, and then their old PDS becomes unable to perform service auth on their behalf.  This breaks the flow on PDSes that are a behind an entryway service (e.g. bsky.social), since service auth was used between the old PDS and its entryway service.

Resolves #3149

### DIFF
--- a/packages/pds/src/api/com/atproto/server/activateAccount.ts
+++ b/packages/pds/src/api/com/atproto/server/activateAccount.ts
@@ -1,29 +1,21 @@
-import assert from 'node:assert'
-
 import { CidSet } from '@atproto/repo'
 import { INVALID_HANDLE } from '@atproto/syntax'
 import { InvalidRequestError } from '@atproto/xrpc-server'
 
 import AppContext from '../../../../context'
 import { Server } from '../../../../lexicon'
-import { ids } from '../../../../lexicon/lexicons'
 import { assertValidDidDocumentForService } from './util'
+import { authPassthru } from '../../../proxy'
 
 export default function (server: Server, ctx: AppContext) {
   server.com.atproto.server.activateAccount({
     auth: ctx.authVerifier.accessFull(),
-    handler: async ({ auth }) => {
+    handler: async ({ req, auth }) => {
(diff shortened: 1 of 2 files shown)


dholms (approved):
yup yup ✅

PR #3177, opened 2024-12-04
Drop axios dependency

matthieusieben:


packages/bsky/src/util/retry.ts
```diff
@@ -1,38 +1,14 @@
...
 
-export async function retryHttp<T>(
-  fn: () => Promise<T>,
-  opts: RetryOptions = {},
-): Promise<T> {
-  return retry(fn, { retryable: retryableHttp, ...opts })
-}
+export const RETRYABLE_HTTP_STATUS_CODES = [
+  408, 425, 429, 500, 502, 503, 504, 522, 524,
+]
```
devinivy, line 6: Was there a particular reason to move away from the `Set`?

packages/bsky/src/config.ts
```diff
@@ -356,6 +382,34 @@ export class ServerConfig {
...
+    return this.cfg.disableSsrfProtection ?? false
+  }
+
+  get proxyAllowHTTP2(): boolean {
+    return this.cfg.proxyAllowHTTP2 ?? false
+  }
+
+  get proxyHeadersTimeout(): number {
+    return this.cfg.proxyHeadersTimeout ?? 10e3
+  }
```
devinivy, line 396: Just a headsup that this change in timeout was disruptive at least for the PDS use case.  Maybe we should set both defaults to something less conservative, like 30s.

packages/bsky/src/config.ts
```diff
@@ -144,6 +152,22 @@ export class ServerConfig {
     const maxThreadDepth = process.env.BSKY_MAX_THREAD_DEPTH
       ? parseInt(process.env.BSKY_MAX_THREAD_DEPTH || '', 10)
       : undefined
+
+    const disableSsrfProtection =
+      process.env.NODE_ENV === 'development' || process.env.NODE_ENV === 'test'
```
devinivy, line 157: Could we just configure this option in those environments rather than infer it?  Or base it on the value of `debugMode`?

packages/oauth/jwk/src/keyset.ts
```diff
@@ -162,6 +162,7 @@ export class Keyset<K extends Key = Key> implements Iterable<K> {
 
       // Skip negotiation if a specific "alg" was provided
       if (typeof alg === 'string') return [key, alg]
+      else if (alg?.length === 1) return [key, alg[0]!]
```
devinivy, line 165: Is this perhaps from a different line of work?

packages/bsky/src/util/retry.ts
devinivy: Ah I see now how it's used with `RetryAgent` 👍

packages/bsky/src/image/server.ts
```diff
@@ -1,57 +1,86 @@
...
+  res: ServerResponse,
+  next: (err?: unknown) => void,
+) => void
+
+export function createMiddleware(
+  { cfg, dataplane, blobDispatcher }: AppContext,
+  { prefix = '/' }: { prefix?: string } = {},
+): Middleware {
+  if (!prefix.startsWith('/') && !prefix.endsWith('/')) {
+    throw new TypeError('Prefix must start and end with a slash')
```
devinivy, line 48:
```suggestion
  if (!prefix.startsWith('/') || !prefix.endsWith('/')) {
    throw new TypeError('Prefix must start and end with a slash')
```

packages/internal/xrpc-utils/src/accept.ts
```diff
@@ -0,0 +1,143 @@
+import { ResponseType } from '@atproto/xrpc'
```
devinivy, line 1: I am assuming this is a direct port from packages/pds/src/pipethrough.ts, so not reviewing in detail.

.changeset/cool-hairs-tap.md
```diff
@@ -0,0 +1,5 @@
+---
+"@atproto/bsky": patch
```
devinivy, line 2: I believe we'll need a changeset for xrpc-utils.

packages/bsky/src/api/blob-resolver.ts
```diff
@@ -1,160 +1,307 @@
...
+  const takendownPromise = dataplane.getBlobTakedown({ did, cid: cidStr })
+  const identityPromise = dataplane.getIdentityByDid({ did }, { signal })
+
+  // Avoid "uncaught promise rejection" error (we will await later, if we care)
+  identityPromise.catch(() => {})
+
+  const { takenDown } = await takendownPromise
+  // If the blob is taken down, don't wait for the identity promise
   if (takenDown) {
+    abortController.abort() // We don't care about the identity response
```
devinivy, line 187: The amount of time we hit this takedown path, let alone on an identity that isn't known/cached, should be quite tiny!  At a glance this seems a bit unrelated from the axios move—was there something specific on your mind with this change?

packages/bsky/src/api/blob-resolver.ts
```diff
@@ -1,160 +1,307 @@
...
+  url: URL,
+): Generator<[string, string], void, unknown> {
+  if (bypassKey && bypassHostname) {
+    const matchesUrl = bypassHostname.startsWith('.')
+      ? url.hostname.endsWith(bypassHostname)
+      : url.hostname === bypassHostname
+
+    if (matchesUrl) {
+      yield ['x-ratelimit-bypass', bypassKey]
     }
```
devinivy, line 228: I wonder if it would be a bit more straightforward to build an array then return it.

packages/bsky/src/api/blob-resolver.ts
```diff
@@ -1,160 +1,307 @@
...
+ */
+export function createCidVerifier(
+  cid: CID,
+  encoding?: string | string[],
+): Duplex {
+  // If the upstream content is compressed, we do not want to return a
+  // de-compressed stream here. Indeed, the "compression" middleware will
+  // compress the response before it is sent downstream, if it is not already
+  // compressed. Because of this, it is preferable to keep the return the
+  // content as-is to avoid re-compressing it.
```
devinivy, line 245: Okay, cool!

packages/bsky/src/api/blob-resolver.ts
```diff
@@ -1,160 +1,307 @@
...
+      yield ['x-ratelimit-bypass', bypassKey]
     }
   }
-  return {}
+}
+
+/**
+ * This function creates a passthrough stream that will decompress (if needed)
+ * and verify the CID of the input stream. The output data will be identical to
+ * the input data. For that reason, iy you need the
```
devinivy, line 235: Looks like this comment got truncated.

packages/bsky/src/image/server.ts
```diff
@@ -66,71 +95,133 @@ export class ImageProcessingServer {
...
+
+      const dispatcherOptions: Dispatcher.RequestOptions = {
+        method: 'GET',
+        origin: url.origin,
+        path: url.pathname + url.search,
+        headers: Object.fromEntries([
+          ...getBlobHeaders(cfg, url),
+          ['accept-encoding', acceptEncoding],
+        ]),
       }
```
devinivy, line 108: The implementation gets a little complicated in order to get everything streaming end-to-end.  In case it simplifies the implementation here, this code isn't performance critical and not used in production.  The primary use is to run a self-contained local dev environment which is suitable for UI work, without needing to run an additional service such as imgproxy.

packages/internal/xrpc-utils/src/accept.ts
matthieusieben: Just a copy paste

packages/bsky/src/api/blob-resolver.ts
matthieusieben: Reverted to the previous behavior

packages/bsky/src/config.ts
matthieusieben: Changed the definition of `debugMode` so that is is "opt-in" instead of defaulting to true. Added the ability to override the value by setting the `BSKY_DISABLE_SSRF_PROTECTION` env var.

packages/bsky/src/config.ts
matthieusieben: As it stands, this will only impact `getBlob` calls from the "blob-resolver". The current timeout there is actually 5 seconds (see bellow).

https://github.com/bluesky-social/atproto/blob/c72145dbeb2d67068bc28c00a13447e0d382d121/packages/bsky/src/api/blob-resolver.ts#L133C1-L133C59

Should I:
1) Lower the timeout to match current behavior ?
2) Increase to timeout to future proof (when these config become used for other PDS proxying purposes) ?
3) Leave it like that?

packages/bsky/src/util/retry.ts
matthieusieben: Still reverted this change as `set.has` is better than `array.includes` and `RetryAgent` is only instantiated once.

packages/oauth/jwk/src/keyset.ts
matthieusieben: yup. reverted.

packages/ozone/src/util.ts
```diff
@@ -83,3 +83,17 @@ export const formatLabelerHeader = (parsed: ParsedLabelers): string => {
   )
   return parts.join(',')
 }
+
+export async function allFulfilled(
```
devinivy, line 87: This is fine, but there's a util in common `handleAllSettledErrors()` that does something similar.

devinivy (approved):

PR #3182, opened 2024-12-05
[subs 1] RevenueCat sync in bsync

ghost:
This supersedes https://github.com/bluesky-social/atproto/pull/3119, by putting the subscription sync with RevenueCat in `bsync`.

A future PR will add a "scan" endpoint for consumers to read the individual subscriptions operations.

### DIFF
--- a/packages/bsync/src/config.ts
+++ b/packages/bsync/src/config.ts
@@ -93,10 +93,10 @@ export const readEnv = (): ServerEnvironment => {
     // secrets
     apiKeys: envList('BSYNC_API_KEYS'),
     // revenue cat
-    revenueCatV1ApiKey: envStr('BSKY_REVENUE_CAT_V1_API_KEY'),
-    revenueCatV1ApiUrl: envStr('BSKY_REVENUE_CAT_V1_API_URL'),
+    revenueCatV1ApiKey: envStr('BSYNC_REVENUE_CAT_V1_API_KEY'),
+    revenueCatV1ApiUrl: envStr('BSYNC_REVENUE_CAT_V1_API_URL'),
     revenueCatWebhookAuthorization: envStr(
-      'BSKY_REVENUE_CAT_WEBHOOK_AUTHORIZATION',
+      'BSYNC_REVENUE_CAT_WEBHOOK_AUTHORIZATION',
     ),
   }
 }


packages/bsync/src/db/migrations/20241205T030533572Z-subs.ts
```diff
@@ -0,0 +1,24 @@
...
+    .addColumn('createdAt', 'timestamptz', (col) =>
+      col.notNull().defaultTo(sql`CURRENT_TIMESTAMP`),
+    )
+    .execute()
+  await db.schema
+    .createTable('subs_item')
+    .addColumn('actorDid', 'varchar', (col) => col.primaryKey())
+    .addColumn('entitlements', 'jsonb', (col) => col.notNull())
+    .addColumn('fromId', 'bigint', (col) => col.notNull())
+    .execute()
```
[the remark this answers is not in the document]
dholms: We could, but it would require a supporting index on the operations table which I'm not sure we want to do because we want to keep that table as lightweight as possible

We could always reconstruct those values through a migration, but we wouldn't be able to quickly query them. So if we need to quickly query them, I'd recommend materializing them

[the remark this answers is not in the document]
dholms: Altho these could just be materialized in the dataplane from subscribing to the event stream 🤔 I'm not sure bsync needs them materialized

packages/bsync/src/subscriptions/revenueCatWebhookHandler.ts
```diff
@@ -0,0 +1,174 @@
...
+  ) {
+    res.statusCode = 403
+    return res.end(
+      JSON.stringify({
+        error: 'Forbidden: invalid authentication for RevenueCat webhook',
+      }),
+    )
+  }
+
+  if (req.method !== 'POST') {
```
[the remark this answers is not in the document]
dholms: nah i like it 👍

packages/bsync/src/subscriptions/revenueCatWebhookHandler.ts
```diff
@@ -0,0 +1,174 @@
...
+      // create subs op
+      const id = await createSubsOp(txn, actorDid, entitlements)
+      // update subs state
+      await updateSubsItem(txn, id, actorDid, entitlements)
+      return id
+    })
+
+    res.statusCode = 200
+    res.end(JSON.stringify({ success: true, operationId: id }))
+  } catch (error) {
```
dholms, line 124: we probably oughta log this error

dholms (approved):
nice lookin great 👍

packages/bsync/src/db/migrations/20241205T030533572Z-subs.ts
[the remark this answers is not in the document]
devinivy: When I built bsync I materialized both mute and notification state just because it was easy to do and might be useful, to avoid backfilling from the beginning of time to bootstrap a new service.  But not critical, so I think it's alright either way.

packages/bsync/src/db/schema/subs_op.ts
```diff
@@ -0,0 +1,17 @@
...
+  createdAt: GeneratedAlways<Date>
+}
+
+export type SubsOpEntry = Selectable<SubsOp>
+
+export const tableName = 'subs_op'
+
+export type PartialDB = { [tableName]: SubsOp }
+
+export const createSubsOpChannel = 'subs_op_create' // used with listen/notify
```
devinivy, line 17: We may want to wire this up here as well: https://github.com/bluesky-social/atproto/blob/d634f9d6d858dc26c69443e9c2e1f1e20ed381a3/packages/bsync/src/index.ts#L102-L109

packages/bsync/src/subscriptions/revenueCatWebhookHandler.ts
```diff
@@ -0,0 +1,174 @@
...
+export const isRevenueCatWebhookUrl = (urlStr: string | undefined) => {
+  if (!urlStr) return false
+  const url = new URL(urlStr, 'http://host')
+  return url.pathname === '/webhooks/revenuecat'
+}
+
+const parseBody = async (req: IncomingMessage) =>
+  new Promise((resolve, reject) => {
+    let body = ''
+    req.on('data', (chunk) => (body += chunk))
```
devinivy, line 17: I believe this is implicitly converting the chunk from a buffer to a string then concatenating.  I wonder if you could get some funk here depending on how the chunks break up the string bytes— would recommend collecting the buffers, and then concatenating them and turning them into a string.

packages/bsync/src/subscriptions/revenueCatWebhookHandler.ts
```diff
@@ -0,0 +1,174 @@
...
+
+  if (!revenueCatClient) {
+    res.statusCode = 501
+    return res.end(
+      JSON.stringify({
+        error:
+          'Not Implemented: bsync is being served without RevenueCat support',
+      }),
+    )
+  }
```
estrattonbailey, line 54: Can we get an alert log or something here if we start to see spikes in this? Maybe that's already here for free with datadog

estrattonbailey: I commented here, but really I mean, we want to track any sort of failures from this webhook handler, and know what's gone wrong if spikes occur

packages/bsync/src/subscriptions/revenueCatWebhookHandler.ts
estrattonbailey: oh should have scrolled down, yeah I think good idea 👍

packages/bsync/src/subscriptions/revenueCatWebhookHandler.ts
```diff
@@ -0,0 +1,174 @@
...
+    req.on('data', (chunk) => (body += chunk))
+    req.on('end', () => {
+      try {
+        resolve(JSON.parse(body))
+      } catch (err) {
+        reject(err)
+      }
+    })
+    req.on('error', reject)
+  })
```
devinivy, line 26: Two things here:
a. if the response is compressed then this would be the compressed body.
b. could probably simplify this and avoid worrying about e.g. setting-up and cleaning-up event handlers:

```js
const chunks: Buffer[] = []
for await (const chunk of req) {
  chunks.push(chunk)
}
return Buffer.concat(chunks).toString()
```

packages/bsync/src/subscriptions/revenueCatWebhookHandler.ts
devinivy, line 54: I wonder if we could get in trouble if we hit these error cases and never read/drain the request body.

packages/bsync/src/subscriptions/revenueCatWebhookHandler.ts
```diff
@@ -0,0 +1,174 @@
...
+      JSON.stringify({
+        error:
+          'Bad request: body must be JSON with Content-Type: application/json',
+      }),
+    )
+  }
+
+  let body: RevenueCatEventBody
+  try {
+    body = (await parseBody(req)) as RevenueCatEventBody
```
devinivy, line 89: I'd encourage validating the input, even though it's authenticated.  Despite the trust, bugs or unexpected behavior could happen on either end of the contract.  We often use zod for this kind of input validation.

packages/bsync/src/subscriptions/revenueCatWebhookHandler.ts
devinivy: I think `compression` may just be for compression of responses, but I don't see the harm in it so long as it doesn't conflict with the connect middleware.

packages/bsync/src/index.ts
```diff
@@ -39,20 +46,25 @@ export class BsyncService {
...
-        res.statusCode = 200
-        res.setHeader('content-type', 'application/json')
-        return res.end(JSON.stringify({ version: cfg.service.version }))
-      }
-      handler(req, res)
-    })
-    return new BsyncService({ ctx, server, ac })
+
+    const app = express()
+    app.use(cors({ maxAge: DAY / SECOND }))
```
devinivy, line 51: I don't think we need to support CORS on bsync since it's an implementation detail of the appview and end users don't end-up interacting with it.

devinivy (approved):
Nice!

devinivy (approved):

PR #3160, opened 2024-12-02
Add refreshSubscriptionCache

ghost:
This adds the `app.bsky.subscription.refreshSubscriptionCache` lexicon, which can be called:
1. By a regular user, to clear their own cache.
2. By an admin, to call any user's cache.

This builds on top of things added in https://github.com/bluesky-social/atproto/pull/3119.

PR #3196, opened 2024-12-06
[subs 2] Add `purchase` endpoints

ghost:
(supersedes https://github.com/bluesky-social/atproto/pull/3156 and https://github.com/bluesky-social/atproto/pull/3160)

This PR adds 4 endpoints to `bsky` and the necessary underlying calls to `bsync` to implement them:
1. `app.bsky.purchase.getSubscriptions`:
	1. Used to fetch the active subscriptions for a user. It is called roughly once per app load. 
	2. This calls `bsync`, which calls RevenueCat.
2. `app.bsky.purchase.getFeatures`:
	1. Used to fetch what paid features the client app should enable for the user (example: enable selecting a custom color).
	2. Reads cached data from the dataplane, doesn't reach out to RevenueCat.
	3. The dataplane read path is still to be done in a future PR. It will be done through `bsync` providing a `scan...` GRPC endpoint that the indexing infrastructure will call. That `scan...` method will serve data that `bsync` either obtained by being triggered by a webhook event, or by being called in the `app.bsky.purchase.refreshCache` flow.
3. `app.bsky.purchase.getSubscriptionGroup`:
	1. This is used when the user navigates to the screen to subscribe and we need to show the possible options (depends on the platform, etc).
	2. This calls `bsync`. In this case, `bsync` doesn't need to call RevenueCat, it replies with in-memory data. The return data here is basically just a configuration. Still, we do it in `bsync` so `bsky` can remain completely agnostic regarding any purchase-related configuration.
4. `app.bsky.purchase.refreshCache`:
	1. Use by the user (and possibly support agents) to refresh the dataplane cache. Refreshing this cache possibly affects the data returned by `app.bsky.purchase.getFeatures` (if that was out-of-date, this is what makes it be refreshed).
	2. This calls `bsync`, which calls RevenueCat.

*You can skip the `codegen` commit when reviewing.*

Notable points:
1. `bsky` knows nothing about RevenueCat.
2. `bsync` knows about RevenueCat, but only in a few contained points.
	1. In the webhook listener, which is RevenueCat-specific.
	2. Some configs have "revenue cat" in their names, as they are used to build the `RevenueCatClient`.
	3. Previously we only had `RevenueCatClient`, which was used in multiple places in `bsync`. In this PR I abstracted `RevenueCatClient` away into a `PurchasesClient`, that does the mapping between RevenueCat formats and our formats. `PurchasesClient` is now called in multiple places instead of `RevenueCatClient`. Note also that `PurchasesClient` is what does the mappings between entitlements, products, etc. Since those rely on configs for the Stripe identifiers, I preferred to put that logic inside a class than in individual functions (like it was in the POC), so I can inject the config.

### DIFF
--- a/packages/bsync/tests/purchases.test.ts
+++ b/packages/bsync/tests/purchases.test.ts
@@ -25,7 +25,7 @@ import {
   RcGetSubscriberResponse,
 } from '../src/purchases'
 
-const revenueCatWebhookAuthorization = 'Bearer any-token'
+const revenueCatWebhookAuthorization = 'any-token'
 
 describe('purchases', () => {
   let bsync: BsyncService
@@ -760,7 +760,7 @@ const callWebhook = async (
     method: 'POST',
     body: JSON.stringify(body),
     headers: {
-      Authorization: revenueCatWebhookAuthorization,
+      Authorization: `Bearer ${revenueCatWebhookAuthorization}`,
       'Content-Type': 'application/json',
     },
   })
--- a/packages/bsync/src/purchases/revenueCatClient.ts
+++ b/packages/bsync/src/purchases/revenueCatClient.ts
@@ -98,6 +98,6 @@ export class RevenueCatClient {
   }
 
   isWebhookAuthorizationValid(authorization: string | undefined): boolean {
-    return authorization === this.webhookAuthorization
+    return authorization === `Bearer ${this.webhookAuthorization}`
   }
 }


lexicons/app/bsky/purchase/getSubscriptions.json
```diff
@@ -0,0 +1,41 @@
...
+          }
+        }
+      }
+    },
+    "subscription": {
+      "type": "object",
+      "required": [],
+      "properties": {
+        "status": { "type": "string" },
+        "renewalStatus": { "type": "string" },
```
estrattonbailey, line 31: Nit: could add some `knownValues` fields here so we get more strict types

packages/bsync/src/context.ts
```diff
@@ -42,16 +42,27 @@ export class AppContext {
...
+    if (cfg.purchases) {
+      purchasesClient = new PurchasesClient({
+        revenueCatV1ApiKey: cfg.purchases.revenueCatV1ApiKey,
+        revenueCatV1ApiUrl: cfg.purchases.revenueCatV1ApiUrl,
+        revenueCatWebhookAuthorization:
+          cfg.purchases.revenueCatWebhookAuthorization,
+        stripePriceIdMonthly: cfg.purchases.stripePriceIdMonthly,
+        stripePriceIdAnnual: cfg.purchases.stripePriceIdAnnual,
+        stripeProductIdMonthly: cfg.purchases.stripeProductIdMonthly,
+        stripeProductIdAnnual: cfg.purchases.stripeProductIdAnnual,
```
estrattonbailey, line 55: Is the reason these are configurable and the other platforms are not because these values will change depending on if we're in test mode or production?

packages/bsync/src/purchases/purchasesClient.ts
```diff
@@ -0,0 +1,269 @@
...
+
+    this.STRIPE_PRODUCTS = {
+      coreMonthly: opts.stripeProductIdMonthly,
+      coreAnnual: opts.stripeProductIdAnnual,
+    }
+
+    this.PRODUCTS = {
+      web: {
+        coreMonthly: opts.stripePriceIdMonthly,
+        coreAnnual: opts.stripePriceIdAnnual,
```
estrattonbailey, line 94: I think this is totally reasonable (see other comment about test/prod) but keep in mind: as we add more products, we'll end up with a lot of config. I'd maybe argue that it's safe to hard-code these values in here too, and add a single piece of top-level config like `isTestMode`. There may be other handling we only want in test mode too, so might be good to narrow the `opts` interface and just make this more verbose here.

lexicons/app/bsky/purchase/getSubscriptions.json
dholms: yup you can 👍 
it's basically a documentation mechanism for string fields that says what the expected values are. but you can always put any string in it

lexicons/app/bsky/purchase/getFeatures.json
```diff
@@ -0,0 +1,33 @@
+{
+  "lexicon": 1,
+  "id": "app.bsky.purchase.getFeatures",
+  "defs": {
+    "main": {
+      "type": "query",
+      "description": "Enumerate unlocked paid feature flags for the requesting account. Requires auth.",
+      "output": {
```
dholms, line 8: let's put empty `input`s on all these schemas

it's a little goofy but basically going from "no input" to "some input" is in this lexicon gray area. so this is future proofing it to the possible addition of some input in the future

packages/bsky/src/api/app/bsky/purchase/getFeatures.ts
```diff
@@ -0,0 +1,50 @@
...
+const defaultFeatures: Features = {
+  customProfileColor: false,
+}
+
+const coreEntitlementFeatures: Features = {
+  customProfileColor: true,
+}
+
+const getFeaturesForViewerEntitlements = async (
+  viewerDid: string,
```
dholms, line 32: nit, but i usually do the `ctx` as the first param of functions like this since you can kinda think of it as "parameterizing" the function

packages/bsync/src/routes/add-purchase-operation.ts
```diff
@@ -0,0 +1,35 @@
...
+    if (!isValidDid(actorDid)) {
+      throw new ConnectError(
+        'actor_did must be a valid did',
+        Code.InvalidArgument,
+      )
+    }
+
+    const entitlements = await purchasesClient.getEntitlements(actorDid)
+
+    await addPurchaseOperation(db, actorDid, entitlements)
```
dholms, line 31: I think we should check the current state for a user before adding an operation to the log. Otherwise, someone can keep pinging bsync to basically fill it up with a bunch of no-op logs. 

And in that framing, maybe we actually want to change the proto to something like `refresh-purchases`.

packages/bsync/proto/bsync.proto
```diff
@@ -66,7 +68,45 @@ message ScanNotifOperationsResponse {
...
+  string renewalStatus = 2;
+  string group = 3;
+  string platform = 4;
+  string offering = 5;
+  google.protobuf.Timestamp periodEndsAt = 6;
+  google.protobuf.Timestamp periodStartsAt = 7;
+  google.protobuf.Timestamp purchasedAt = 8;
+}
+
+message GetSubscriptionsRequest {
```
dholms, line 88: I forget, is there as reason we use "subscription" here but "add" in other locations?

packages/bsync/src/routes/get-subscriptions.ts
```diff
@@ -0,0 +1,44 @@
...
+    const { actorDid } = req
+    if (!isValidDid(actorDid)) {
+      throw new ConnectError(
+        'actor_did must be a valid did',
+        Code.InvalidArgument,
+      )
+    }
+
+    const { email, subscriptions } =
+      await purchasesClient.getSubscriptionsAndEmail(actorDid)
```
dholms, line 30: could be a good time to check our local state and make sure it's in sync

dholms: and if not, add an op to the log

packages/bsync/src/purchases/purchasesClient.ts
dholms: i think either is probably fine tbh. I'd be inclined to just leave it like this for now & switch out how we do it as the config gets more complex

packages/bsky/src/api/app/bsky/purchase/getFeatures.ts
```diff
@@ -0,0 +1,50 @@
+import { Server } from '../../../../lexicon'
+import AppContext from '../../../../context'
+import { Features } from '@atproto/api/dist/client/types/app/bsky/purchase/getFeatures'
```
estrattonbailey, line 3: I'm not positive on this, but I think the move here is to reference this type from `packages/bsky/src/lexicon/types/app/bsky/purchase/getFeatures.ts`, which is local to the app view package.

packages/bsky/src/api/app/bsky/purchase/getFeatures.ts
```diff
@@ -0,0 +1,50 @@
...
+  ctx: AppContext,
+  viewerDid: string,
+): Promise<Features> => {
+  const { purchaseEntitlements } = await ctx.dataplane.getPurchaseEntitlements({
+    dids: [viewerDid],
+  })
+
+  if (purchaseEntitlements?.length === 0) {
+    return defaultFeatures
+  }
```
estrattonbailey, line 41: Not sure if `purchaseEntitlements` can actually be `undefined`, but we might want something like `!purchaseEntitlements || purchaseEntitlements.length === 0` to avoid referencing `undefined[0]`.

packages/bsync/src/context.ts
estrattonbailey: Well we can still hard-code them because they won't change after we create them, but I see your point. We can deal with more config when the time comes, or move to the hard-coded option then too if we want.

packages/bsync/src/purchases/addPurchaseOperation.ts
```diff
@@ -7,11 +7,13 @@ export const addPurchaseOperation = async (
   actorDid: string,
   entitlements: string[],
 ) => {
+  const sortedEntitlements = [...entitlements].sort()
```
estrattonbailey, line 10: May want to mention here that they're sorted for later comparison. I can imagine someone in the future deleting this without knowing the purpose haha.

packages/bsync/src/purchases/purchasesClient.ts
```diff
@@ -0,0 +1,269 @@
+import { Subscription } from '../proto/bsync_pb'
+import { RcSubscription, RevenueCatClient } from './revenueCatClient'
+
+type PlatformId = 'web' | 'ios' | 'android'
+
+type SubscriptionGroupId = 'core'
+
+type SubscriptionOfferingId = 'coreMonthly' | 'coreAnnual'
```
estrattonbailey, line 8: Ya know what might be good? Calling these `core:monthly` and `core:annual`, just to hammer home the relationship. Not a big deal, your call if you wanna do that right now.

packages/bsync/src/config.ts
```diff
@@ -92,12 +117,16 @@ export const readEnv = (): ServerEnvironment => {
     dbMigrate: envBool('BSYNC_DB_MIGRATE'),
     // secrets
     apiKeys: envList('BSYNC_API_KEYS'),
-    // revenue cat
+    // purchases
     revenueCatV1ApiKey: envStr('BSYNC_REVENUE_CAT_V1_API_KEY'),
     revenueCatV1ApiUrl: envStr('BSYNC_REVENUE_CAT_V1_API_URL'),
     revenueCatWebhookAuthorization: envStr(
       'BSYNC_REVENUE_CAT_WEBHOOK_AUTHORIZATION',
     ),
```
devinivy, line 125: This is the full HTTP header—how would you feel about making the configuration just receive the credential itself, the bearer token.

packages/bsync/src/purchases/purchasesClient.ts
```diff
@@ -0,0 +1,272 @@
...
+    this.revenueCatClient = new RevenueCatClient({
+      v1ApiKey: opts.revenueCatV1ApiKey,
+      v1ApiUrl: opts.revenueCatV1ApiUrl,
+      webhookAuthorization: opts.revenueCatWebhookAuthorization,
+    })
+
+    this.STRIPE_PRODUCTS = {
+      'core:monthly': opts.stripeProductIdMonthly,
+      'core:annual': opts.stripeProductIdAnnual,
+    }
```
devinivy, line 89: I bet we can get some tighter types on these with some `as const`.
```suggestion
    this.STRIPE_PRODUCTS = {
      'core:monthly': opts.stripeProductIdMonthly,
      'core:annual': opts.stripeProductIdAnnual,
    } as const
```

devinivy: Ah I see now you define the tight types above— that works too!

packages/bsync/src/purchases/purchasesClient.ts
```diff
@@ -0,0 +1,272 @@
...
+    const now = new Date()
+    const expiresAt = new Date(s.expires_date)
+
+    let status = 'unknown'
+    if (s.auto_resume_date) {
+      if (now > expiresAt) {
+        status = 'paused'
+      }
+    } else if (now > expiresAt) {
+      status = 'expired'
```
devinivy, line 194: Does this seem accurate?
```suggestion
    } else if (now >= expiresAt) {
      status = 'expired'
```

packages/bsync/src/purchases/purchasesClient.ts
```diff
@@ -0,0 +1,272 @@
...
+      }
+    } else if (now > expiresAt) {
+      renewalStatus = 'will_not_renew'
+    }
+
+    let periodEndsAt = s.expires_date
+    if (s.auto_resume_date) {
+      if (now > expiresAt) {
+        periodEndsAt = s.auto_resume_date
+      }
```
devinivy, line 219:
```suggestion
    } else if (now >= expiresAt) {
      renewalStatus = 'will_not_renew'
    }

    let periodEndsAt = s.expires_date
    if (s.auto_resume_date) {
      if (now >= expiresAt) {
        periodEndsAt = s.auto_resume_date
      }
```

packages/bsync/src/config.ts
devinivy: Can look to `BSYNC_API_KEYS` for a similar situation.

packages/bsync/src/routes/get-subscription-group.ts
```diff
@@ -0,0 +1,30 @@
...
+    const { group, platform } = req
+    try {
+      const offerings = purchasesClient.getSubscriptionGroup(group, platform)
+
+      return new GetSubscriptionGroupResponse({
+        offerings,
+      })
+    } catch (error) {
+      throw new ConnectError((error as Error).message, Code.InvalidArgument)
+    }
```
devinivy, line 28: Not sure I totally understand why this error gets rethrown like this.  I would be a little concerned that in the case of a developer error the details would leak out here through `error.message`.

packages/bsync/tests/purchases.test.ts
```diff
@@ -1,32 +1,60 @@
...
   RcEventBody,
   RcGetSubscriberResponse,
 } from '../src/purchases'
+import { Code, ConnectError } from '@connectrpc/connect'
+import {
+  GetSubscriptionsResponse,
+  PurchaseOperation,
+} from '../src/proto/bsync_pb'
+import { Timestamp } from '@bufbuild/protobuf'
+import { DAY, wait } from '@atproto/common'
```
devinivy, line 23: Could we hoist these up to below the `get-port` import (per convention described in an earlier comment)?

packages/bsync/tests/purchases.test.ts
```diff
@@ -77,18 +114,26 @@ describe('purchases', () => {
...
       expect(response.status).toBe(400)
-      expect(response.json()).resolves.toMatchObject({
-        error: 'Bad request: invalid DID in app_user_id',
-      })
+      const body = await response.json()
+      expect(body).toStrictEqual(
+        structuredClone({
+          error: 'Bad request: invalid DID in app_user_id',
+          success: false,
+        }),
```
devinivy, line 135: Does this need to be cloned?

devinivy: (Some other examples like this in here too.)

devinivy (commented):
Looking great, just left a handful of small comments.

packages/bsync/src/config.ts
devinivy: Yeah, I do still lean that way.  (And this does seem like a bearer token to me 👍)

devinivy (approved):

PR #3203, opened 2024-12-07
fix(crypto): fix error message of `bytesToMultibase`

tesaguri:
The error message was printing the `Uint8Array` instead of the name of the unsupported encoding.


devinivy (approved):
Thanks!

PR #3209, opened 2024-12-08
pds: support config for email reply-to addresses

devinivy:
The pds can be configured with email reply-to addresses, useful in particular if it should be different from the sender.  The new optional environment variables are:
 - `PDS_EMAIL_REPLY_TO_ADDRESS`
 - `PDS_MODERATION_EMAIL_REPLY_TO_ADDRESS`

### DIFF
--- a/.changeset/slow-brooms-scream.md
+++ b/.changeset/slow-brooms-scream.md
@@ -1,0 +1,5 @@
+---
+"@atproto/pds": patch
+---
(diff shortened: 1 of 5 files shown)


foysalit (approved):
looks good!

devinivy:
Didn't end-up being needed, closing out to keeps things simple.

PR #3222, opened 2024-12-10
Add filter param to listNotifications

gaearon:
This is what I imagine I'll need. The `mentions` name is not exactly reflective of the purpose (it'll be *any* text — so replies, quotes, mentions all count). Could maybe call that "posts" idk. But in the UI we'll probably call it Mentions for familiarity.

### DIFF
--- a/lexicons/app/bsky/notification/listNotifications.json
+++ b/lexicons/app/bsky/notification/listNotifications.json
@@ -8,6 +8,12 @@
       "parameters": {
         "type": "params",
         "properties": {
+          "filter": {
+            "type": "string",
+            "description": "Notification types to include in response.",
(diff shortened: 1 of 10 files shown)


lexicons/app/bsky/notification/listNotifications.json
```diff
@@ -8,6 +8,14 @@
       "parameters": {
         "type": "params",
         "properties": {
+          "filter": {
```
matthieusieben, line 11: Wouldn't `reason` be less generic and reflect what is being filtered?

dholms: yeah I like that 👍

PR #3224, opened 2024-12-10
Notification filtering

dholms:


gaearon (approved):
lgtm

gaearon:
rolled into https://github.com/bluesky-social/atproto/pull/3222

gaearon:
Seeing questions about what this is for — this is groundwork for adding a separate Mentions tab to the app. The only difference between *Mentions* and *All* is that Mentions would omit likes and follows (but it *would* show mentions, quotes, replies, and other notifications that contain text).

PR #3236, opened 2024-12-12
Add account details to queryStatuses

matthieusieben:
- [x] Add support for pre-existing view indexes in the migration (allowing the views and indexes to be created manually instead of through the migration)
- [x] Refresh materialized views from the worker (note: refresh order matters!)
- [x] Determine which columns to expose as filter
  - [x] For those: add an index on their respective materialized views
  - [x] Adapt the code to support the filtering


packages/ozone/src/lexicon/lexicons.ts
```diff
@@ -11179,6 +11179,69 @@ export const schemaDict = {
               type: 'string',
             },
           },
+          accountStats: {
```
foysalit, line 11182: Instead of having them spread all across the `subjectStatusView` object, we should consider a container object for `repoStats` or `repoModerationStats` ?

packages/ozone/src/lexicon/lexicons.ts
```diff
@@ -11179,6 +11179,69 @@ export const schemaDict = {
...
+            ref: 'lex:tools.ozone.moderation.defs#recordsStats',
+          },
+        },
+      },
+      accountStats: {
+        description:
+          'Statistics about a particular account subject on the labeller',
+        type: 'object',
+        properties: {
+          suspendedCount: {
```
foysalit, line 11200: maybe an additional few properties at the account level
1. `reportCount` -> total number of reports the account received. if we could have those broken down by report reason, that would be amazing! but the report reasons are kinda dynamic so might be a pain to store and expose. we do use `json` columns to hold arbitrary objects like this quite a lot so that might be an option?

matthieusieben: done

packages/oauth/oauth-client-browser/package.json
```diff
@@ -19,12 +19,12 @@
     "directory": "packages/oauth/oauth-client-browser"
   },
   "type": "commonjs",
-  "main": "dist/index.js",
+  "main": "dist/index.cjs",
```
matthieusieben, line 22: This is probably a mistake

packages/ozone/src/mod-service/views.ts
```diff
@@ -545,54 +542,37 @@ export class ModerationViews {
...
+            return clause.orWhere((qb) =>
+              qb
+                .where('mss.did', '=', sub.did)
+                .where('mss.recordPath', '=', sub.recordPath ?? ''),
+            )
+          },
+          // If the array is empty, no result should ne returned
+          clause.where(sql`false`),
+        )
+      },
```
matthieusieben, line 561: This is a fix for `TODO: Fix the typing here?`

lexicons/tools/ozone/moderation/defs.json
```diff
@@ -182,6 +182,76 @@
...
+        },
+        "recordsStats": {
+          "description": "Statistics related to the record subjects authored by the subject's account",
+          "type": "ref",
+          "ref": "#recordsStats"
+        }
+      }
+    },
+    "accountStats": {
+      "description": "Statistics about a particular account subject on the labeller",
```
dholms, line 199: nit, but generally we refer to these as "mod service" instead of labeler since it's duties are wider than just labeling. alternately, could just leave off the "on the labeler"

packages/ozone/src/db/migrations/20241220T144630860Z-stats-materialized-views.ts
```diff
@@ -0,0 +1,187 @@
...
+            sql<number>`COUNT(*) FILTER(
+              WHERE ${eb.ref('action')} = 'tools.ozone.moderation.defs#modEventReport'
+              AND ${eb.ref('meta')} ->> 'reportType' = ${REASONAPPEAL}
+            )`.as('appealCount'),
+        ])
+        .groupBy('subjectDid'),
+    )
+    .execute()
+
+  // TODO try/catch to ignore existing
```
dholms, line 56: migrations are meant to be run once and nearly all of our migrations will fail if re-run (for instance `createTable`)

basically, i don't think you need to worry about a try/catch here

packages/ozone/src/mod-service/status.ts
```diff
@@ -203,6 +203,24 @@ const getSubjectStatusForRecordEvent = ({
   return {}
 }
 
+export const moderationSubjectStatusQueryBuilder = (db: DatabaseSchema) => {
+  return db
+    .selectFrom('moderation_subject_status as mss')
+    .selectAll('mss')
```
dholms, line 209: I do get the appeal of the shortened versions of these table names, but they leak out into the calling functions and I find it hard to balance the mental state of them. Generally I try to only use abbreviated names like this when the variable is confined to some scoped function & the context is contained within a dozen or two lines. Even though it's wordier, I'm inclined to just use the table names for these.

packages/ozone/src/mod-service/index.ts
```diff
@@ -312,19 +313,6 @@ export class ModerationService {
       .executeTakeFirst()
   }
 
-  async getCurrentStatus(
```
dholms, line 315: This work might just be in-progress, but I think this method is still needed. Specifically, I see it used in `createReport`

packages/ozone/src/db/migrations/20241220T144630860Z-stats-materialized-views.ts
matthieusieben: Because I manually created those view in the prod db. But Devin tells me that migrations are manually performed anyways so this does not matter indeed.

packages/ozone/src/mod-service/index.ts
matthieusieben: This is a mistake. will revert 👍

packages/ozone/src/mod-service/views.ts
```diff
@@ -556,54 +553,37 @@ export class ModerationViews {
...
-
-        return clause
-      })
-      .selectAll()
+  async getSubjectStatus(subjects: string[]) {
+    const parsedSubjects = subjects.map(parseSubjectId)
+
+    const builder = moderationSubjectStatusQueryBuilder(this.db.db).where(
+      (clause) => {
+        return parsedSubjects.reduce(
```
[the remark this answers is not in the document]
matthieusieben: Ideally we would use [this pattern](https://kysely.dev/docs/examples/where/or-where) but that requires updating Kysely (while is non trivial).


But reading the docs of the current version we use I read:


> Even the first `where` can be an `orWhere`. This is useful if you are looping through a set of conditions:


So we don't actually need the index check.

packages/pds/src/index.ts
```diff
@@ -99,7 +99,13 @@ export class PDS {
     app.use(loggerMiddleware)
     app.use(compression())
     app.use(authRoutes.createRouter(ctx)) // Before CORS
-    app.use(cors({ maxAge: DAY / SECOND }))
+    app.use(
+      cors({
+        maxAge: DAY / SECOND,
+        credentials: true,
+        origin: (origin, callback) => callback(null, origin),
```
foysalit, line 106: want to make sure these are intended changes?

packages/ozone/src/mod-service/views.ts
```diff
@@ -559,51 +556,39 @@ export class ModerationViews {
...
-    }
-
-    const builder = this.db.db
-      .selectFrom('moderation_subject_status')
-      .where((clause) => {
-        parsedSubjects.forEach((subject, i) => {
-          const applySubjectFilter = filterForSubject(
-            subject.did,
-            subject.recordPath,
+    if (!subjects.length) return new Map()
```
foysalit, line 559: nice! looks so much simpler!

foysalit (approved):
looking good!

packages/pds/src/index.ts
matthieusieben: Yes and no. Will move in separate PR.

dholms (approved):
Looks great 🙇

PR #3238, opened 2024-12-12
Validate profile records for views

dholms:
Correctly validate profile records so that we don't serve the client broken views. 

Note: this is what we do for all other record types, I think we just didn't do it for profiles because they're wrapped in the actor type and are structured a bit differently

### DIFF
--- a/packages/bsky/src/hydration/actor.ts
+++ b/packages/bsky/src/hydration/actor.ts
@@ -5,7 +5,6 @@ import {
   HydrationMap,
   RecordInfo,
   parseRecord,
-  parseRecordBytes,
   parseString,
   safeTakedownRef,
 } from './util'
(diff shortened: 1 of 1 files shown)


packages/bsky/src/hydration/actor.ts
```diff
@@ -116,18 +115,19 @@ export class ActorHydrator {
...
       return acc.set(did, {
         did,
         handle: parseString(actor.handle),
-        profile: parseRecordBytes<ProfileRecord>(profile?.record),
+        profile: profile?.record,
         profileCid: profile?.cid,
-        profileTakedownRef: safeTakedownRef(profile),
-        sortedAt: profile?.sortedAt?.toDate(),
-        indexedAt: profile?.indexedAt?.toDate(),
+        profileTakedownRef: profile?.takedownRef,
```
devinivy, line 128: Any reason to drop usage of `safeTakedownRef()` here?  It has a fallback for `profile?.takedownRef`, so the logic is a little different.

devinivy (approved):

packages/bsky/src/hydration/actor.ts
dholms: `parseRecord` does exactly that 👍 
similar to how we got to drop `toDate` on sortedAt/indexedAt

PR #3249, opened 2024-12-16
More explicit error messages when uploading corrupt images

### LINKED ISSUE #3151: Can't upload text/html blob with contents that contain "<svg" + any character
**Describe the bug**

com.atproto.repo.uploadBlob fails with a payload containing "<svg" followed by any other characters, with a mimeType of "text/html".

**To Reproduce**

1. Create a file (test.txt) with contents:
```
<svgQ
```
2. Use postman/atfile/any other tool to upload test.txt.
3. This causes a `500: Internal Server Error`.

**Expected behavior**

The blob successfully uploads.

**Additional context**
* Any number of characters can precede the "<svg"
* Removing the character after "<svg" causes a successful upload.

db3000:
Fixes #3151 

Uploading an attachment using `com.atproto.repo.uploadBlob` will fail with a 500 Internal Server Error if the contents of the upload has the signature of an image (as determined by sharp via ImageMagick) but is not fully parsable as the corresponding image type.

While it is possible to handle such uploads by silencing this error, this PR errs on the side of caution and continues to reject corrupted images but with a more explicit error. It seems to me that the risk of allowing people to upload malformed images outweighs the benefit of more flexible uploads. I can change the code to silently accept such uploads though if that is desirable instead.

### DIFF
--- a/packages/pds/src/image/index.ts
+++ b/packages/pds/src/image/index.ts
@@ -2,6 +2,7 @@ import { Readable } from 'stream'
 import { pipeline } from 'stream/promises'
 import sharp from 'sharp'
 import { errHasMsg } from '@atproto/common'
+import { InvalidRequestError } from '@atproto/xrpc-server'
 
 export async function maybeGetInfo(
   stream: Readable,
@@ -15,10 +16,12 @@ export async function maybeGetInfo(
     ])
     metadata = result
   } catch (err) {
+    // If the buffer doesn't have any identifiable image in it, return null
     if (errHasMsg(err, 'Input buffer contains unsupported image format')) {
       return null
     }
-    throw err
+    // Otherwise, the buffer has a corrupt image in it
+    throw new InvalidRequestError("Image detected but content is unparsable")
   }
   const { size, height, width, format } = metadata
   if (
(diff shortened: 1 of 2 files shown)


jakelazaroff:
Agree with @Saturn-VI — I think if the PDS wants to try to detect extra metadata that's fine, but ultimately it should just fall back to treating the blob as opaque.

db3000:
Changes the PR as per comments here and in #3151 , this PR now changes it so the PDS no longer validates the image but just continues if it is malformed

packages/pds/src/image/index.ts
```diff
@@ -15,10 +14,8 @@ export async function maybeGetInfo(
     ])
     metadata = result
   } catch (err) {
-    if (errHasMsg(err, 'Input buffer contains unsupported image format')) {
-      return null
-    }
-    throw err
+    // The buffer has a corrupted image or no image at all.
+    return null
```
matthieusieben: This is not a proper fix as it will also mute any `pipeline(stream, processor)` error.

db3000: Good point, I will limit the exceptions that are ignored here

packages/common-web/src/util.ts
```diff
@@ -145,7 +145,7 @@ export const isErrnoException = (
 }
 
 export const errHasMsg = (err: unknown, msg: string): boolean => {
-  return !!err && typeof err === 'object' && err['message'] === msg
+  return !!err && typeof err === 'object' && err['message'].startsWith(msg)
```
db3000, line 148: This was necessary as the error is actually:
> Input buffer has corrupt header: glib: XML parse error: Error domain 1 code 73 on line 1 column 6 of data: Couldn't find end of Start Tag svgQ line 1

packages/pds/src/image/index.ts
db3000: I put back the previous code and just added another specific error message that gets ignored

PR #3250, opened 2024-12-16
add trending topic lex, wire in agent

haileyok:


packages/bsky/src/api/app/bsky/graph/getTrendingTopics.ts
```diff
@@ -0,0 +1,97 @@
...
+  SkeletonFnInput,
+} from '../../../../pipeline'
+import { HydrateCtx, Hydrator } from '../../../../hydration/hydrator'
+import { Views } from '../../../../views'
+import { QueryParams } from '../../../../lexicon/types/app/bsky/unspecced/getTrendingTopics'
+import AtpAgent from '@atproto/api'
+import { TrendingTopic } from '../../../../lexicon/types/app/bsky/unspecced/defs'
+import { InternalServerError } from '@atproto/xrpc-server'
+
+export default function (server: Server, ctx: AppContext) {
```
dholms, line 18: gotta mount this route in `src/api/index.ts`

packages/bsky/src/context.ts
```diff
@@ -52,6 +53,10 @@ export class AppContext {
     return this.opts.suggestionsAgent
   }
 
+  get topicsAgent(): AtpAgent | undefined {
```
dholms, line 56: is this going out to Discover? tempted to just re-use the suggestionsAgent in the meantime, unless we have a reason to introduce a new one

not very concerned either way though. Happy to keep it as is as well

dholms (approved):
looks great 🙌

packages/bsky/src/api/app/bsky/graph/getTrendingTopics.ts
```diff
@@ -0,0 +1,97 @@
...
+  const getTrendingTopics = createPipeline(
+    skeleton,
+    hydration,
+    noBlocksOrMutes,
+    presentation,
+  )
+  server.app.bsky.unspecced.getTrendingTopics({
+    auth: ctx.authVerifier.standard,
```
dholms, line 26: oh, do we want this to be able to be hit unauthed? if so, we can use `standardOptional` here

packages/bsky/src/context.ts
haileyok: have it on a separate box rn which is why i created the separate agent. we might combine it sooner than later at which point we could def drop this.

packages/bsky/src/api/app/bsky/graph/getTrendingTopics.ts
haileyok: ah nice, ty

PR #3251, opened 2024-12-16
:sparkles: Allow appeals on takendown account

foysalit:
TODO: The entryway also needs to be synced with these changes but going to wait until an approval on this approach before porting it over there.

### // Deprecated
This PR adds a new endpoint `com.atproto.server.appealAccountAction` that allows takendown users to send an appeal event to moderation service.

First approach i tried was to make `com.atproto.server.createSession` respond with a custom token that can only be used to send a `com.atproto.moderation.createReport` request. while it felt risky to change the response shape there, i was able to make it work. However, a few major issues with that approach are

1. Using service token to make requests pipe through PDS to ozone doesn't work because the PDS expect a generic access token.
2. Even if we made (a) work, we would still run into attack vectors that allow takendown accounts to make reports since the lxm only allows us to specify which endpoint the token can call which opens up the entire reporting endpoint for all takendown users.

A major downside with this current approach may be that we are requiring the users to send auth credentials to a new endpoint besides `createSession` which doesn't seem ... ideal? 
### // Deprecated 

This PR adds a new optional param `allowTakendown: boolean` to the `createSession` endpoint and when set, the endpoint now issues a valid access token with the scope `com.atproto.takendown`. This token allows the user to appeal the takedown decision.


packages/pds/src/pipethrough.ts
```diff
@@ -23,12 +23,23 @@ import {
 import AppContext from './context'
 import { ids } from './lexicon/lexicons'
 import { httpLogger } from './logger'
+import { AuthScope } from './auth-verifier'
+
+const getStandardAccess = (ctx: AppContext, path: string) => {
```
dholms, line 28: I think what I'd prefer to do here is just add a route for handling `createReport` to the PDS with the correct auth scope on it. This will keep it from falling back to this generic proxy logic.

Two benefits here:
- we keep this route-specific complexity out of the proxy code & can keep it more generic
- we avoid running the `getStandardAccess` fn on every proxy request which will keep them a bit lighter

We can still use the `pipethrough` function from this file to help with actually proxying the request - you can find examples in `getFeed` and `getRecord`

packages/pds/src/account-manager/index.ts
```diff
@@ -211,15 +211,21 @@ export class AccountManager
...
     did: string,
     appPassword: password.AppPassDescript | null,
+    isSoftDeleted = false,
   ) {
     const { accessJwt, refreshJwt } = await auth.createTokens({
       did,
       jwtKey: this.jwtKey,
       serviceDid: this.serviceDid,
-      scope: auth.formatScope(appPassword),
+      scope: isSoftDeleted
```
dholms, line 220: For a second, I thought we needed to replicate this logic on `rotateRefreshToken`, but `refreshSession` itself prevents session refresh when you're account is takendown.

I actually like that & don't think you should/need to be refreshing your session when taken down. Just making sure my understanding of this matches yours

packages/pds/src/account-manager/index.ts
```diff
@@ -211,15 +211,21 @@ export class AccountManager
...
+        ? AuthScope.Takendown
+        : auth.formatScope(appPassword),
```
dholms, line 222: can we pass through `isSoftDeleted` to `formatScope` instead of doing the ternary logic here?

packages/pds/tests/auth.test.ts
```diff
@@ -1,6 +1,6 @@
 import * as jose from 'jose'
 import { AtpAgent } from '@atproto/api'
-import { TestNetworkNoAppView, SeedClient } from '@atproto/dev-env'
+import { SeedClient, TestNetworkNoAppView } from '@atproto/dev-env'
 import { createRefreshToken } from '../src/account-manager/helpers/auth'
 
 describe('auth', () => {
```
dholms, line 6: It'd be nice to have a quick test in here that verifies that a "takendown" access token isn't able to make normal calls - for instance `createRecord`

packages/pds/src/account-manager/index.ts
foysalit: yep, we are actively making sure that refresh tokens for taken down accounts are not stored and as you said, we already fail attempts to rotate refreshtoken for takendown accounts so it can just be left alone.

packages/pds/tests/auth.test.ts
foysalit: sure, added the test in `takedown-appeal` which calls createRecord and asserts that `bad token scope` error is thrown.

dholms (approved):
lookin good!

PR #3254, opened 2024-12-17
Update README.md to add some missing details in examples

kidGodzilla:
I built an [example](https://github.com/kidGodzilla/bsky-oauth-example) and ran into a few small omissions in the examples here, that might make it difficult for a new developer not incredibly familiar with OAuth internals to figure out why their example doesn't work.

My recommended changes to `README.md`:

Each `JoseKey.fromImportable` call requires a key ID (parameter 2). Recommended change:

```
            JoseKey.fromImportable(process.env.PRIVATE_KEY_1, 'key1'),
            JoseKey.fromImportable(process.env.PRIVATE_KEY_2, 'key2'),
            JoseKey.fromImportable(process.env.PRIVATE_KEY_3, 'key3'),
```

Otherwise, it fails silently in testing (localhost) and fails with an error in production (because kid is missing on each key in jwks.json).

---

Scope requires `atproto transition:generic` for the code example that follows, would recommend adding the scope line back in as:

```
scope: 'atproto transition:generic',
```

Otherwise, the example that follows this (login) will not work, transition:generic is required for that.

---

This is required in the example given:

```
token_endpoint_auth_signing_alg: 'RS256',
```

Otherwise an error is thrown. Recommend adding this line.

---

Not included, if you wanted to include it, is a little in-memory store to complete the example.

```js
// Simple In-memory session store (OAuth requirement)
class InMemoryStore {
    constructor() {
        this.storeData = {};
    }

    // stateStore methods
    async set(key, internalState) {
        this.storeData[key] = internalState;
    }

    async get(key) {
        return this.storeData[key] || undefined;
    }

    async del(key) {
        delete this.storeData[key];
    }
}
const stateStore = new InMemoryStore();
const sessionStore = new InMemoryStore();
```

But, I did not include this, you might have an existing library in mind or not want to give an easier option than the redis session store example.

Feel free to add if you think it improves things.

### DIFF
--- a/packages/oauth/oauth-client-node/README.md
+++ b/packages/oauth/oauth-client-node/README.md
@@ -38,9 +38,11 @@ const client = new NodeOAuthClient({
     policy_uri: 'https://my-app.com/policy',
     redirect_uris: ['https://my-app.com/callback'],
     grant_types: ['authorization_code', 'refresh_token'],
+    scope: 'atproto transition:generic',
     response_types: ['code'],
     application_type: 'web',
     token_endpoint_auth_method: 'private_key_jwt',
+    token_endpoint_auth_signing_alg: 'RS256',
     dpop_bound_access_tokens: true,
     jwks_uri: 'https://my-app.com/jwks.json',
   },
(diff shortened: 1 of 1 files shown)


bnewbold (approved):
these look good to me!

@matthieusieben could you give this a double-check as well? and maybe comment on whether to include the `InMemoryStore`, either in the README or even in the package itself (for development).

PR #3260, opened 2024-12-17
:sparkles: Add subject tags with report type

foysalit:
This PR adds report reason tags to subjects which enables moderators to filter queue items based on the subjects' report types.

### DIFF
--- a/packages/ozone/src/api/moderation/emitEvent.ts
+++ b/packages/ozone/src/api/moderation/emitEvent.ts
@@ -164,7 +164,7 @@ const handleModerationEvent = async ({
     )
     const initialTags = isModEventReport(event)
       ? [getTagForReport(event.reportType)]
-      : []
+      : undefined
     await tagService.evaluateForSubject(initialTags)
 
     if (subject.isRepo()) {
--- a/packages/ozone/src/api/moderation/emitEvent.ts
+++ b/packages/ozone/src/api/moderation/emitEvent.ts
@@ -7,6 +7,7 @@ import {
   isModEventEmail,
   isModEventLabel,
   isModEventMuteReporter,
+  isModEventReport,
   isModEventReverseTakedown,
   isModEventTag,
   isModEventTakedown,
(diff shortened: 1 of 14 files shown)


packages/ozone/tests/report-muting.test.ts
```diff
@@ -69,7 +69,7 @@ describe('report-muting', () => {
     })
 
     // Verify that a subject status was not created for bob's post since the reporter was muted
-    await assertSubjectStatus(bobsPostSubject.uri, undefined)
+    await assertSubjectStatus(bobsPostSubject.uri, REVIEWNONE)
```
foysalit, line 72: This only needed changing because we are now emitting an additional tag event because of report which causes the status to change to reviewNone which is similar to not having a reviewstate so there is no functional change.

packages/ozone/src/tag-service/index.ts
```diff
@@ -22,15 +22,18 @@ export class TagService {
     ]
   }
 
-  async evaluateForSubject() {
+  // Allow the caller to seed the initial tags
+  async evaluateForSubject(initialTags: string[] = []) {
     try {
-      const tags: string[] = []
+      const tags = new Set([...initialTags])
```
matthieusieben, line 28: You don't need to copy here (the `Set` will do that for you). This also allows to type the argument in a broader way `initialTags?: Iterable<string>` instead of defaulting to an empty array.

```suggestion
      const tags = new Set(initialTags)
```

packages/ozone/src/api/moderation/emitEvent.ts
```diff
@@ -160,7 +162,10 @@ const handleModerationEvent = async ({
       ctx.cfg.service.did,
       moderationTxn,
     )
-    await tagService.evaluateForSubject()
+    const initialTags = isModEventReport(event)
+      ? [getTagForReport(event.reportType)]
+      : []
```
matthieusieben, line 167: Nit: No need to create empty unused array here.
```suggestion
      : undefined
```

matthieusieben (approved):

packages/ozone/src/tag-service/index.ts
foysalit: cleaned it up as recommended!

PR #3261, opened 2024-12-17
:sparkles: Add && separator for tags filter param on queryStatuses endpoint

foysalit:
This PR adds a special separator character in the `tags` filter param array on `tools.ozone.moderation.queryStatuses` endpoint which allows moderators to filter queue items for multiple tags using the AND operator instead of the default OR operator which is applied between different items in the array.

So, currently, there is no way for a moderator to find all subjects that have both `embed:image` and `lang:ja` tags because passing `tags: ['lang:ja', 'embed:image']` will return all items with either of the tags. This PR makes way for us to apply the expected filter using the syntax `tags: ['lang:ja && embed:image']`

### DIFF
--- a/packages/ozone/src/mod-service/index.ts
+++ b/packages/ozone/src/mod-service/index.ts
@@ -797,27 +797,32 @@ export class ModerationService {
   ) => {
     const { ref } = this.db.db.dynamic
     // Build an array of conditions
-    const conditions = tags.map((tag) => {
-      if (tag.includes('&&')) {
-        // Split by '&&' for AND logic
-        const subTags = tag
-          .split('&&')
-          // Make sure spaces on either sides of '&&' are trimmed
-          .map((subTag) => subTag.trim())
-          // Remove empty strings after trimming is applied
-          .filter(Boolean)
-        return sql`(${sql.join(
-          subTags.map(
-            (subTag) =>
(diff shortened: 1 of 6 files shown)


packages/ozone/src/mod-service/index.ts
```diff
@@ -791,6 +791,38 @@ export class ModerationService {
...
+          .map((subTag) => subTag.trim())
+          // Remove empty strings after trimming is applied
+          .filter(Boolean)
+        return sql`(${sql.join(
+          subTags.map(
+            (subTag) =>
+              sql`${ref('moderation_subject_status.tags')} ? ${subTag}`,
+          ),
+          sql` AND `,
+        )})`
```
matthieusieben, line 815: Do we ignore empty queries ? Or maybe even throw an error ?

```suggestion
        // False is neutral in a union ("OR")
        if (!subTags.length) return sql`FALSE`

        return sql`(${sql.join(
          subTags.map(
            (subTag) =>
              sql`${ref('moderation_subject_status.tags')} ? ${subTag}`,
          ),
          sql` AND `,
        )})`
```

matthieusieben (approved):

packages/ozone/src/mod-service/index.ts
foysalit: good catch! handled these edge cases through boolean filters.

PR #3262, opened 2024-12-18
fix missing limit param in getTrendingTopics

haileyok:
didn't spread params here, so missing the `limit` in the request.

### DIFF
--- a/packages/bsky/src/api/app/bsky/unspecced/getTrendingTopics.ts
+++ b/packages/bsky/src/api/app/bsky/unspecced/getTrendingTopics.ts
@@ -48,7 +48,7 @@ const skeleton = async (input: SkeletonFnInput<Context, Params>) => {
   if (ctx.topicsAgent) {
     const res = await ctx.topicsAgent.app.bsky.unspecced.getTrendingTopics(
       {
-        ...params,
+        limit: params.limit,
         viewer: params.viewer,
       },
       {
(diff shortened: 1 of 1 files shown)


packages/bsky/src/api/app/bsky/unspecced/getTrendingTopics.ts
```diff
@@ -48,6 +48,7 @@ const skeleton = async (input: SkeletonFnInput<Context, Params>) => {
   if (ctx.topicsAgent) {
     const res = await ctx.topicsAgent.app.bsky.unspecced.getTrendingTopics(
       {
+        ...params,
```
devinivy, line 51: If I'm reading correctly, this will include some values we don't want to try to pass through such as `headers`.  Can we pass through the specific fields accepted by `unspecced.getTrendingTopics` instead?

haileyok: oh, yea. is this actually how you all usually structure these? (i.e., in my head was just thinking `params` was "request params" and didnt look closely. ill push change for now)

devinivy (approved):

packages/bsky/src/api/app/bsky/unspecced/getTrendingTopics.ts
devinivy: I'm personally paranoid and try to be explicit at service boundaries, but I wouldn't be surprised at all if there are exceptions to that written by me in this codebase 😅

PR #3266, opened 2024-12-18
Fix avatar path resolution in dev-env

ghost:
The avatar path is currently dependent on the CWD. So it might not work if executed from a different place, e.g., when using `@atproto/dev-env` as a dependency on another project.

A related but separate issue is that the `img` folder is not copied to `dist/` on build.

This PR moves the images to an `assets` folder outside `src` and accesses it in a non-CWD-dependent way.

### DIFF
--- a/packages/pds/tests/crud.test.ts
+++ b/packages/pds/tests/crud.test.ts
@@ -156,9 +156,7 @@ describe('crud operations', () => {
   })
 
   it('attaches images to a post', async () => {
-    const file = await fs.readFile(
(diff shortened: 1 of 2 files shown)


matthieusieben (approved):
I typically like to put "assets" in an `assets` folder (next to `src`). This way, if `src` gets discarded from the published package, the assets can still be bundled.

devinivy (approved):
Seems like a reasonable way to get this working in the short/med-term 👍

matthieusieben (approved):

PR #3270, opened 2024-12-19
Add codegen to support deprecated labels

estrattonbailey:
The `gore` label is deprecated in favor of `graphic-media`. However, some moderation actions still result in the `gore` label being applied. In our frontend codebase, [we manually migrate](https://github.com/bluesky-social/social-app/blob/main/src/lib/moderatePost_wrapped.ts) the `gore` label to `graphic-media`.

Work is being done to remove the usage of this old label, but we have no definite timeline for complete removal and migration to fully eradicate historical data.

Once we do that migration, the code added in this PR will not be necessary, but:
- 3p devs likely already have hit this problem, or they likely will
- we may have more need of this handling in the future

The first commit here ports our existing logic directly into the SDK. The second commit removes that and replaces it with a new `gore` label definition that duplicates `graphic-media`. The third is what I think is the "correct" approach, even if naming and such needs to change after we talk about this.

### This approach

`gore` needs to duplicate `graphic-media`, but we should leave some paper trail of why this is. I've added an `alias` field to the definitions JSON file so that `graphic-media` can define `gore` as an alias. Codegen then uses this `alias` array to generate additional definitions that duplicate the root def.

Because we don't intent to support aliases, and prefer the root def, I'm also generating a tsdoc `@deprecation` comment above the alias definition.


pfrazee (approved):
Nice, lgtm

PR #3271, opened 2024-12-19
:sparkles: Add a policy property to takedown events

foysalit:
This PR adds an optional `policy` property to all takedown events that allows an arbitrary string value and allows moderators to query for all takedowns by specifying a `policy` param through th `queryEvents` endpoint.

### DIFF
--- a/lexicons/tools/ozone/moderation/defs.json
+++ b/lexicons/tools/ozone/moderation/defs.json
@@ -224,6 +224,10 @@
         "acknowledgeAccountSubjects": {
           "type": "boolean",
           "description": "If true, all other reports on content authored by this account will be resolved (acknowledged)."
+        },
+        "policy": {
+          "type": "string",
+          "description": "Name/Keyword of the policy that drove the decision."
         }
       }
     },
--- a/lexicons/tools/ozone/moderation/defs.json
+++ b/lexicons/tools/ozone/moderation/defs.json
@@ -225,9 +225,11 @@
           "type": "boolean",
           "description": "If true, all other reports on content authored by this account will be resolved (acknowledged)."
         },
-        "policy": {
-          "type": "string",
-          "description": "Name/Keyword of the policy that drove the decision."
+        "policies": {
+          "type": "array",
+          "maxLength": 5,
+          "items": { "type": "string" },
+          "description": "Names/Keywords of the policies that drove the decision."
         }
       }
     },
--- a/lexicons/tools/ozone/moderation/queryEvents.json
+++ b/lexicons/tools/ozone/moderation/queryEvents.json
@@ -106,6 +106,10 @@
               "type": "string"
             }
           },
+          "policy": {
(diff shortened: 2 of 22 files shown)


lexicons/tools/ozone/moderation/defs.json
```diff
@@ -224,6 +224,10 @@
         "acknowledgeAccountSubjects": {
           "type": "boolean",
           "description": "If true, all other reports on content authored by this account will be resolved (acknowledged)."
+        },
+        "policy": {
+          "type": "string",
+          "description": "Name/Keyword of the policy that drove the decision."
         }
```
devinivy, line 231: Is there any merit to sprinkling this onto other event types such as `modEventLabel`?

lexicons/tools/ozone/moderation/queryEvents.json
```diff
@@ -106,6 +106,10 @@
               "type": "string"
             }
           },
+          "policy": {
+            "type": "string",
+            "description": "If specified, only events where the policy matches the given policy are returned"
+          },
```
devinivy, line 112: Is there a decent chance we end-up wanting to list multiple policies here in the future?  If so, we could proactively make this an array.

devinivy (approved):
Left a couple small comments, looking good!  Also may want to add a changeset for @atproto/api and @atproto/ozone before merging.

lexicons/tools/ozone/moderation/defs.json
foysalit: Labels themselves are quite descriptive of the "why" part of the action and confirmed from T&S folks that this won't be necessary any time soon.

lexicons/tools/ozone/moderation/queryEvents.json
foysalit: T&S expressed that this may be nice to have at some point but not anytime soon. I'll get it in the backend but the UI side will mostly keep it 1:1

packages/ozone/src/mod-service/views.ts
```diff
@@ -139,11 +139,11 @@ export class ModerationViews {
 
     if (
       event.action === 'tools.ozone.moderation.defs#modEventTakedown' &&
-      event.meta?.policy
+      typeof event.meta?.policies === 'string'
     ) {
       eventView.event = {
         ...eventView.event,
-        policy: event.meta.policy,
+        policies: event.meta.policies.split(' '),
```
devinivy, line 146: Look out for the empty string, which likes to split into `['']`.

devinivy: In other words:
```js
const policies = []
const serialized = policies.join(' ')
const deserialized = serialized.split(' ')
// we want policies and deserialized to both be `[]`
// but policies is `[]` while deserialized is `['']`
```

devinivy (approved):

PR #3286, opened 2024-12-23
lexicon schema for lexicon schema records (meta)

bnewbold:
See Lexicon Resolution spec discussion: https://github.com/bluesky-social/atproto/discussions/3074

This needs a codegen and maybe a patch-level changeset before merging; looking for review on just the schema to start.

### DIFF
--- a/lexicons/com/atproto/lexicon/schema.json
+++ b/lexicons/com/atproto/lexicon/schema.json
@@ -1,0 +1,35 @@
+{
+  "lexicon": 1,
+  "id": "com.atproto.lexicon.schema",
+  "defs": {
+    "main": {
+      "type": "record",
+      "description": "Lexicon schema for when schemas themselves are represented as atproto records. Note that the full schema language is not defined in Lexicon. This schema is protocol-layer, and designed to be flexible and forwards-compatible, and does not follow normal schema best practices.",
+      "key": "nsid",
+      "record": {
+        "type": "object",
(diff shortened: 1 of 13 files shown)


lexicons/com/atproto/lexicon/schema.json
```diff
@@ -0,0 +1,35 @@
...
+            "format": "nsid",
+            "description": "The NSID of this overall set of Lexicon definitions. Must match the record key of this record. Must not include a fragment (separated by a #), even '#main'."
+          },
+          "description": {
+            "type": "string",
+            "description": "Short top-level description of the Lexicon (distinct from descriptions of any sub-fields)"
+          },
+          "defs": {
+            "type": "array",
+            "items": { "type": "unknown" },
```
devinivy, line 28: I believe we want this whole thing to be `"type": "unknown"`, without the array.

bnewbold: The current specs say that an `unknown` field must be an object (not an array, string, etc): https://atproto.com/specs/lexicon

bnewbold: I was mistaken, it is a map of def name to lexicon, not an array!

bnewbold:
Updated to simplify down to just the 'lexicon' field.

I also removed the codegen (via force-push) until we are settled on the schema file.

devinivy (approved):

PR #3306, opened 2024-12-31
[bsky did 1] Add did:web to bsky

ghost:
Docs: https://github.com/bluesky-social/bsky-docs/pull/302

bsky already had a DID web but it was misconfigured. For it to return the endpoint properly, it required the public URL hostname and DID web hostname to match. That was no longer the case after we have different PoPs, because the public URIs would be different on each PoP, but the DID web hostname should still be the same in both, so we always returned the 404 in the `if`.

That was now removed and as long as a DID web config is present, the DID doc will be returned. A new service was also added.

## Testing

1. Checkout this branch
2. Apply this:
	```diff
	diff --git a/packages/dev-env/src/bsky.ts b/packages/dev-env/src/bsky.ts
	index fd61bb7c9..d64b8c7de 100644
	--- a/packages/dev-env/src/bsky.ts
	+++ b/packages/dev-env/src/bsky.ts
	@@ -26,13 +26,14 @@ export class TestBsky {
	 
	     const port = cfg.port || (await getPort())
	     const url = `http://localhost:${port}`
	-    const serverDid = await plcClient.createDid({
	+    let serverDid = await plcClient.createDid({
	       signingKey: serviceKeypair.did(),
	       rotationKeys: [serviceKeypair.did()],
	       handle: 'bsky.test',
	       pds: `http://localhost:${port}`,
	       signer: serviceKeypair,
	     })
	+    serverDid = 'did:web:api.bsky.app'
	 
	     // shared across server, ingester, and indexer in order to share pool, avoid too many pg connections.
	     const db = new bsky.Database({
	```
3. Build `dev-env` and `bsky` (cd to the package, `pnpm build`)
4. In `atproto`, `make run-dev-env`.
5. `curl localhost:2584/.well-known/did.json | jq`:
	```json
	{
	  "@context": [
	    "https://www.w3.org/ns/did/v1"
	  ],
	  "id": "did:web:api.bsky.app",
	  "verificationMethod": [
	    {
	      "id": "did:web:api.bsky.app#atproto",
	      "type": "Multikey",
	      "controller": "did:web:api.bsky.app",
	      "publicKeyMultibase": "zQ3shge2SL3u8mT82WEer9sesfa34HbnN7DgvhiQkwkq4XVMK"
	    }
	  ],
	  "service": [
	    {
	      "id": "#bsky_notif",
	      "type": "BskyNotificationService",
	      "serviceEndpoint": "https://api.bsky.app"
	    },
	    {
	      "id": "#bsky_appview",
	      "type": "BskyAppView",
	      "serviceEndpoint": "https://api.bsky.app"
	    }
	  ]
	}
	```

### DIFF
--- a/packages/bsky/src/api/well-known.ts
+++ b/packages/bsky/src/api/well-known.ts
@@ -28,8 +28,8 @@ export const createRouter = (ctx: AppContext): express.Router => {
             serviceEndpoint,
           },
           {
-            id: '#bsky_app_view',
-            type: 'BskyAppViewService',
+            id: '#bsky_appview',
+            type: 'BskyAppView',
             serviceEndpoint,
           },
         ],


packages/bsky/src/api/well-known.ts
```diff
@@ -4,31 +4,38 @@ import AppContext from '../context'
...
+          },
+        ],
+        service: [
+          {
+            id: '#bsky_notif',
+            type: 'BskyNotificationService',
+            serviceEndpoint,
+          },
+          {
+            id: '#bsky_app_view',
```
bnewbold, line 31: My thinking on this was `#bsky_api` for the `id`, and `BskyAppView` for the `type`.

Maybe that is inconsistent though? I do like the terseness and generality of "API" instead of "AppView" in the `id`. If we stuck with AppView terminology, at least removing the second underline (so `#bsky_appview`) reads clearer to me.

For the "type", I feel like we have settled on "AppView" being a noun that is a type of service, so the extra suffix is unnecessary.

bnewbold (commented):
Really excited to get this out!

I had proposed doing PoP-specific DIDs and docs in addition to the generic `did:web:api.bsky.app`, but we could handle that later. This is a big unblock as-is.

Left some comments about specific `id` and `type` terminology.

packages/bsky/src/api/well-known.ts
devinivy: How do we feel about splitting the difference and a. keeping the consistency between `id` and `type` as we have elsewhere while also b. adopting @bnewbold's suggestions to trim things down.  To that end I'd propose:
```json
  "id": "#bsky_appview",
  "type": "BskyAppView",
```

devinivy:
@bnewbold just to close the loop on your comment— I intentionally had us leave out the pop-specific DIDs, but we can talk it over more later.  As you mentioned, I also figured this would do the trick for our immediate purposes.

devinivy (approved):
Nice!

PR #3313, opened 2024-12-31
[threadContext 2] Add threadContext to threadViewPost - backend

ghost:
Implements the backend for https://github.com/bluesky-social/atproto/pull/3235, on top of the improvement suggested in https://github.com/bluesky-social/atproto/pull/3318.

It is easier to review each commit:
1. Just move some code around, I find it easier to follow it that way.
2. The implementation.
3. An optimization in fetching thread posts, although it shouldn't yield massive results (only for things like a reply that embeds another post in the thread). I just properly used an existing mechanism, so nothing major. I can easily remove this one if we don't want it.
4. Update the other affected snapshots.

### DIFF
--- a/packages/bsky/src/hydration/hydrator.ts
+++ b/packages/bsky/src/hydration/hydrator.ts
@@ -58,7 +58,7 @@ import {
   Threadgates,
   Postgates,
   FeedItem,
-  ThreadContextStates,
+  ThreadContexts,
   ThreadRef,
 } from './feed'
 import { ParsedLabelers } from '../util'
@@ -91,7 +91,7 @@ export type HydrationState = {
   posts?: Posts
   postAggs?: PostAggs
   postViewers?: PostViewerStates
-  threadContextStates?: ThreadContextStates
+  threadContexts?: ThreadContexts
   postBlocks?: PostBlocks
   reposts?: Reposts
   follows?: Follows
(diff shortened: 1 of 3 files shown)


packages/bsky/src/hydration/hydrator.ts
```diff
@@ -442,6 +470,9 @@ export class Hydrator {
       ctx.viewer
         ? this.feed.getPostViewerStates(threadRefs, ctx.viewer)
         : undefined,
+      ctx.includeThreadRootAuthorState
+        ? this.feed.getThreadRootAuthorStates(threadRefs)
```
devinivy, line 474: I wonder if it would be more straightforward leave `getThreadRootAuthorStates()` out of generic post hydration.  It is only needed on posts within threads, so perhaps it could be added within `hydrateThreadPosts()`.

packages/bsky/src/api/app/bsky/feed/getPostThread.ts
```diff
@@ -36,6 +36,7 @@ export default function (server: Server, ctx: AppContext) {
       const hydrateCtx = await ctx.hydrator.createContext({
         labelers,
         viewer,
+        includeThreadRootAuthorState: true,
```
devinivy, line 39: The main reason `includeTakedowns` and `include3pBlocks` are context options is that they are fairly far-reaching through the hydration and view call stacks.  The usage of `includeThreadRootAuthorState` is quite localized though.  If it needs to stick around, I wonder if it could become an option/argument of the hydration method where it's used.

packages/bsky/src/views/index.ts
```diff
@@ -799,6 +799,7 @@ export class Views {
             opts.depth,
           )
         : undefined,
+      isLikedByRootAuthor: !!state.threadRootAuthors?.get(post.uri)?.like,
```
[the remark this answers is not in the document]
devinivy: I agree, that sounds like a good direction.  Could you make a proposal for updated lexicons on #3235, and once we finalize those lexicons together we can update this PR?

PR #3318, opened 2025-01-02
[threadContext 1] Add threadContext to threadViewPost - lex

ghost:
Proposal on top of https://github.com/bluesky-social/atproto/pull/3235

### DIFF
--- a/lexicons/app/bsky/feed/defs.json
+++ b/lexicons/app/bsky/feed/defs.json
@@ -48,6 +48,13 @@
         "pinned": { "type": "boolean" }
       }
     },
+    "threadContext": {
+      "type": "object",
+      "description": "Metadata about this post within the context of the thread it is in.",
+      "properties": {
+        "rootAuthorLike": { "type": "string", "format": "at-uri" }
+      }
+    },
     "feedViewPost": {
       "type": "object",
       "required": ["post"],
@@ -108,7 +115,8 @@
             "type": "union",
             "refs": ["#threadViewPost", "#notFoundPost", "#blockedPost"]
           }
-        }
+        },
+        "threadContext": { "type": "ref", "ref": "#threadContext" }
       }
     },
     "notFoundPost": {
--- a/packages/api/src/client/lexicons.ts
+++ b/packages/api/src/client/lexicons.ts
@@ -5545,6 +5545,17 @@ export const schemaDict = {
           },
         },
       },
+      threadContext: {
+        type: 'object',
+        description:
+          'Metadata about this post within the context of the thread it is in.',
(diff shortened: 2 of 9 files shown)


lexicons/app/bsky/feed/defs.json
```diff
@@ -109,7 +116,7 @@
             "refs": ["#threadViewPost", "#notFoundPost", "#blockedPost"]
           }
         },
-        "isLikedByRootAuthor": { "type": "boolean" }
+        "rootAuthor": { "type": "ref", "ref": "#rootAuthorState" }
```
devinivy, line 119: I have a small issue with the name, which is that "author" fields are typically profile views, as in the case of `replyRef.grandparentAuthor` or `postView.author`.

What do you think of reframing this field as "thread-specific state" for the post.  Other state that could end up there would be various post rankings within a thread, whether the post is highlighted within a thread, etc.  In that case it might look like:
```json5
{
  "post": {/*...*/},
  "parent": {/*...*/},
  "replies": [{/*...*/}],
  "threadContext": {
    "like": "at://..."
  }
}
```

If we need another set of eyes or a tie-breaker on this, Daniel has opinions and good sensibilities around this kind of thing too.

dholms: I like the reframing to `threadContext`. mainly because i think it's likely that we'll continue to add additional pieces of threadContext

I also like the further clarification of `like`. `opLike` is a little to jargony I think, and prefer to use something with `root` in it. I like `rootAuthorLike` 👍

devinivy: Sweet I dig that, let's do that then if Rafael is also down with it:
```json5
{
  "post": {/*...*/},
  "parent": {/*...*/},
  "replies": [{/*...*/}],
  "threadContext": {
    "rootAuthorLike": "at://..."
  }
}
```

lexicons/app/bsky/feed/defs.json
```diff
@@ -48,6 +48,13 @@
         "pinned": { "type": "boolean" }
       }
     },
+    "threadContextState": {
```
devinivy, line 51: If someone asked me what the object was I think I would say it's "the post's thread context."  So I am not sure "state" is pulling much weight here—see any issue shortening this to just `threadContext`?

davidsainez:
Any chance the root post URI can also be returned in this context? Seems like the perfect place for this metadata. AFAICT it is needed in `agent.post`. Having it in threadContext would avoid querying for unnecessary data when constructing a reply.

devinivy (approved):

PR #3323, opened 2025-01-03
[bsky did 2] Accept audience service in bsky

ghost:
This accepts a new audience in bsky auth that is composed of the DID + the service ID.

The service ID is optional for backward compatibility. In the future, we'll make it required.

### DIFF
--- a/packages/bsky/src/auth-verifier.ts
+++ b/packages/bsky/src/auth-verifier.ts
@@ -92,7 +92,10 @@ export class AuthVerifier {
   ) {
     this.ownDid = opts.ownDid
     this.standardAudienceDids = new Set([
+      // @TODO: Temporarily we still accept the own DID by itself.
+      // Later, we will only accept with the service id fragment.
       opts.ownDid,
+      `${opts.ownDid}#bsky_appview`,
       ...opts.alternateAudienceDids,
     ])
     this.modServiceDid = opts.modServiceDid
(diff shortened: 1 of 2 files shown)


packages/bsky/src/auth-verifier.ts
```diff
@@ -134,12 +134,29 @@ export class AuthVerifier {
...
+        const [audDid, audServiceId] = aud.split('#')
+
+        if (!opts.skipAudCheck && !this.standardAudienceDids.has(audDid)) {
           throw new AuthRequiredError(
             'jwt audience does not match service did',
             'BadJwtAudience',
           )
         }
+
+        const hasServiceId = aud.includes('#')
```
devinivy, line 147: It seems that in parsing out the service id we've already done the work to know whether the audience "has a service id".
```suggestion
        const hasServiceId = audServiceId !== undefined
```

packages/bsky/src/auth-verifier.ts
```diff
@@ -134,12 +134,29 @@ export class AuthVerifier {
           iss: null,
           aud: null,
         })
-        if (!opts.skipAudCheck && !this.standardAudienceDids.has(aud)) {
+
+        const [audDid, audServiceId] = aud.split('#')
+
+        if (!opts.skipAudCheck && !this.standardAudienceDids.has(audDid)) {
```
devinivy, line 140: We'd only like to accept one new audience: `ownDid + '#bsky_appview'`, but this is a bit looser than that.

packages/bsky/src/auth-verifier.ts
```diff
@@ -134,12 +134,29 @@ export class AuthVerifier {
...
           throw new AuthRequiredError(
             'jwt audience does not match service did',
             'BadJwtAudience',
           )
         }
+
+        const hasServiceId = aud.includes('#')
+        if (
+          audDid === this.ownDid &&
```
devinivy, line 149: I think `opts.skipAudCheck` should be taken into account here, as above.

PR #3344, opened 2025-01-09
Properly dispose of unused undici HTTP responses

matthieusieben:


matthieusieben:
Merging without approval because 1) this is a very small change 2) this breaks tests on other branches.

PR #3348, opened 2025-01-09
Allow resetting a route's rate limits

ghost:
Adds `resetRouteRateLimits` to the context, allowing a handler to reset the rate limits for that route.

It only resets the route rate limits, not global or shared.

### DIFF
--- a/.changeset/early-keys-fetch.md
+++ b/.changeset/early-keys-fetch.md
@@ -1,0 +1,5 @@
+---
+"@atproto/xrpc-server": patch
+---
+
+Add resetRouteRateLimits to req context


packages/xrpc-server/src/rate-limiter.ts
```diff
@@ -109,6 +110,30 @@ export class RateLimiter implements RateLimiterI {
       }
     }
   }
+
+  delete = async (
```
dholms, line 114: I might call this `reset` rather than `delete` since `delete` is a keyword. i also find `reset` to be a bit more understandable (which you use in `server.ts`)

packages/xrpc-server/src/rate-limiter.ts
```diff
@@ -109,6 +110,30 @@ export class RateLimiter implements RateLimiterI {
...
+  ): Promise<void> => {
+    const key = opts?.calcKey ? opts.calcKey(ctx) : this.calcKey(ctx)
+    if (key === null) {
+      return
+    }
+
+    try {
+      await this.limiter.delete(key)
+    } catch (err) {
+      logger.error(
```
dholms, line 126: Since this doesn't get handled as middleware, i wonder if it's better to let the error bubble up to the caller? Otherwise the caller has no way of knowing if the limit was actually reset

packages/xrpc-server/tests/rate-limiter.test.ts
```diff
@@ -208,6 +242,22 @@ describe('Parameters', () => {
     await expect(makeCall).rejects.toThrow('Rate Limit Exceeded')
   })
 
+  it('can reset route rate limits', async () => {
```
dholms, line 245: beautiful

dholms (approved):
looks great ✨

PR #3353, opened 2025-01-10
oauth-client: recommend verifying state parameter before processing oauth callback

ThisIsMissEm:
Given that the state parameter is often used as a nonce in many clients (e.g., a random value) and not as a "this is the user account in our system" type value, you'd want to verify the state parameters match before actually completing the oauth callback handling.

Unless there's something I'm missing about the `client.callback` method that some how asserts that state matches before doing an oauth authorization code for token exchange?

### DIFF
--- a/packages/oauth/oauth-client/README.md
+++ b/packages/oauth/oauth-client/README.md
@@ -147,11 +147,13 @@ Make user visit `url`. Then, once it was redirected to the callback URI, perform
 // Parse the query params from the callback URI
 const params = new URLSearchParams('code=...&state=...')
 
+if (params.state !== session.state) {
+  // fail authentication
+  return
+}
+
 // Process the callback using the OAuth client
 const result = await client.callback(params)
-
-// Verify the state (e.g. to link to an internal user)
-result.state === '434321' // true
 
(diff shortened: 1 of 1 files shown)


matthieusieben:
Hey @ThisIsMissEm ,

The `state` parameter is internally managed by the `client` implementation.

The way it works, the client implementation generates a random value, and the actual state data must be persisted in the client somehow (the state data will be stored in the indexdb in the browser, a storage implementation must be provided when used from a back-end).

The client implementation will also remove the persisted state data directly after it is first read (preventing replays).

Thanks to this, users of the lib do not need to actually verify the `state` data returned as part of the login flow, as this is a different value than the random string that was used with the oauth server.

The example in the readme should probably be made more explicit in that regard. Thanks for bringing this to our attention 🙏

matthieusieben:
> Oh! So `result.state` isn't actually the oAuth state parameter?

No, it isn't. Naming things is hard 😅

bnewbold:
I created a tracking issue for this documentation need: https://github.com/bluesky-social/atproto/issues/3876

Going to close this PR.

PR #3364, opened 2025-01-13
Add recId to getSuggestionsSkeleton

iwsmith:
Add recids to getSuggestion response


lexicons/app/bsky/unspecced/getSuggestionsSkeleton.json
```diff
@@ -45,6 +45,10 @@
               "type": "string",
               "format": "did",
               "description": "DID of the account these suggestions are relative to. If this is returned undefined, suggestions are based on the viewer."
+            },
+            "recId": {
```
dholms, line 49: Is the intention to pass this down to the client? If so, we'll need to add `recId` to `getSuggestions` and `getSuggestedFollowsByActor` as well

dholms (approved):
looks good 👍

PR #3370, opened 2025-01-15
Appview: permit redirects during blob resolution

devinivy:
Followup to changes in #3177, ensure we still follow redirects during blob resolution.

### DIFF
--- a/packages/bsky/src/api/blob-resolver.ts
+++ b/packages/bsky/src/api/blob-resolver.ts
@@ -193,6 +193,7 @@ export async function streamBlob(
         path: url.pathname + url.search,
         headers,
         signal: options.signal,
+        maxRedirections: 10,
       },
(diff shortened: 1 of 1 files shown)


matthieusieben (approved):
Yep, This seems like the way to go !

snarfed:
Thanks @devinivy! Looks like this hit Bridgy Fed too, we redirect `getBlob` requests too, and our images stopped working some time last night. Definitely looking forward to seeing this deployed!

devinivy:
It was deployed today, should be resolved!

snarfed:
Woo, thank you!

devinivy:
With #3379 I think we're good to go— @snarfed appreciate your patience as we ironed this out!

snarfed:
@devinivy totally ok, thanks for the quick fixes!

PR #3377, opened 2025-01-16
use ubuntu 22.04 on github actions

ghost:
`ubuntu-latest` used to mean `ubuntu-22.04` but around yesterday it was changed to mean `ubuntu-24.04`, which is breaking our builds because of something related to `puppeteer`

This changes it to explicitly rely on `ubuntu-22.04` until we can make our setup work on `ubuntu-24.04`.

Relevant links:
- a failing build caused by that: https://github.com/bluesky-social/atproto/actions/runs/12810950081/job/35719885873?pr=3376
- the image version rollout explained: https://github.com/actions/runner-images/issues/10636
- the changed documentation merged 2 days ago: https://github.com/actions/runner-images/pull/11332

### DIFF
--- a/.github/workflows/build-and-push-bsky-aws.yaml
+++ b/.github/workflows/build-and-push-bsky-aws.yaml
@@ -12,7 +12,7 @@ env:
 jobs:
   bsky-container-aws:
     if: github.repository == 'bluesky-social/atproto'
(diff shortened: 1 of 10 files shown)


devinivy (approved):
Thanks!

PR #3379, opened 2025-01-16
fetch-node: ensure unicast checks allow psl domains

devinivy:
The unicast check in here was unintentionally only passing through domains that are not themselves public suffix domains.  We want to use the logic from the `psl` package to check that the domain is valid and on a public TLD, but we did not want to omit domains on the PSL such as `s3-us-east-2.amazonaws.com` or `github.io`.

Some details here: https://github.com/lupomontero/psl/issues/258

Should resolve @snarfed's issue here: https://github.com/bluesky-social/atproto/pull/3370#issuecomment-2596713118

### DIFF
--- a/.changeset/silver-birds-run.md
+++ b/.changeset/silver-birds-run.md
@@ -1,0 +1,5 @@
+---
+"@atproto-labs/fetch-node": patch
+---
+
+Unicast checks should permit PSL domains.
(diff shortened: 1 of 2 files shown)


dholms (approved):
yup yup 👍

snarfed:
Aha, thank you!!!

PR #3391, opened 2025-01-18
Lexicon updates for Sync v1.1 (inductive firehose)

bnewbold:
This will probably get applied on top of https://github.com/bluesky-social/atproto/pull/3449

The overall picture here has been in the works since the holidays. We plan to have a public proposal soon.


lexicons/com/atproto/sync/subscribeRepos.json
```diff
@@ -206,6 +242,10 @@
         "cid": {
           "type": "cid-link",
           "description": "For creates and updates, the new record CID. For deletions, null."
+        },
+        "prev": {
+          "type": "cid-link",
+          "description": "For updates and deletes, the previous record CID (required). For creations, null."
```
devinivy, line 248: This mentions that `prev` is `null` for creations.  Two things here to poke at: a. it's not currently marked as nullable, b. marking it as nullable would make it optional-nullable, which is a pattern we want to try to avoid.  I don't think there's a perfect path forward, but my inclination is to leave the lexicon as-is and omit it for creation ops.

lexicons/com/atproto/sync/subscribeRepos.json
```diff
@@ -102,9 +103,44 @@
...
           "items": {
             "type": "cid-link",
-            "description": "List of new blobs (by CID) referenced by records in this commit."
+            "description": "DEPRECATED -- will soon always be empty. List of new blobs (by CID) referenced by records in this commit."
           }
         },
+        "prevData": {
+          "type": "cid-link",
+          "description": "The root CID of the MST tree for the previous commit from this repo (indicated by the 'since' revision field in this message). Corresponds to the 'data' field in the repo commit object. NOTE: this field is effectively required for the 'inductive' version of firehose."
+        },
```
devinivy, line 112: I'm alright with this, but I do have an inclination to name it for its purpose rather than its content.  In particular, how the inductive checks work in the future might not be the same as they do today.  So I'd call it something like `check`, i.e. it's "the event's check CID".

lexicons/com/atproto/sync/subscribeRepos.json
```diff
@@ -19,6 +19,7 @@
           "type": "union",
           "refs": [
             "#commit",
+            "#resync",
```
devinivy, line 22: As I recall we prospectively landed on `#sync` here.

lexicons/com/atproto/sync/subscribeRepos.json
```diff
@@ -102,9 +103,44 @@
...
+    },
+    "resync": {
+      "type": "object",
+      "description": "Updates the repo to a new state, without necessarily including that state on the firehose. Used to recover from broken commit streams, data loss incidents, or in situations where upstream host does not know recent state of the repository.",
+      "required": [
+        "seq",
+        "did",
+        "commit",
+        "time"
+      ],
```
devinivy, line 128: I'd lean towards also including `rev`.

lexicons/com/atproto/sync/subscribeRepos.json
```diff
@@ -102,9 +103,44 @@
...
+        "did": {
+          "type": "string",
+          "format": "did",
+          "description": "The account this repo event corresponds to."
+        },
+        "commit": {
+          "type": "bytes",
+          "description": "CBOR block encoding the commit object (including signature).",
+          "maxLength": 100000
+        },
```
devinivy, line 143: As I recall this was going to be changed to the CAR.

devinivy: We should be a little careful setting the max size since it's hard to change.

bnewbold: yup! updated

lexicons/com/atproto/sync/subscribeRepos.json
bnewbold: I added `rev`, and also `data`. I don't think we have consensus on `data` (@dholms didn't include it in his most recent proposal). I could go either way on it... `rev` is the more important I think.

lexicons/com/atproto/sync/subscribeRepos.json
bnewbold: yep! updated

lexicons/com/atproto/sync/subscribeRepos.json
bnewbold: Ah yep, meant for these to stay "optional" not "nullable". Changed the wording to "For creations, field should not be defined."

Also updated from "(required)" to "(required for inductive firehose)".

lexicons/com/atproto/sync/subscribeRepos.json
```diff
@@ -102,9 +103,47 @@
...
+        "did": {
+          "type": "string",
+          "format": "did",
+          "description": "The account this repo event corresponds to. Must match that in the commit object."
+        },
+        "blocks": {
+          "type": "bytes",
+          "description": "CAR file containing the commit, as a block. The CAR header must include the commit block CID as the first 'root'.",
+          "maxLength": 10000
```
bnewbold, line 137: flagging this for discussion; @dholms had it as 4k in a proposal doc. It is maybe hard to change this value so I gave it more breathing room... though probably 4k is fine/good?

lexicons/com/atproto/sync/getRepoStatus.json
```diff
@@ -27,7 +27,14 @@
             "status": {
               "type": "string",
               "description": "If active=false, this optional field indicates a possible reason for why the account is not active. If active=false and no status is supplied, then the host makes no claim for why the repository is no longer being hosted.",
-              "knownValues": ["takendown", "suspended", "deactivated"]
+              "knownValues": [
+                "takendown",
+                "suspended",
+                "deleted",
+                "deactivated",
+                "broken",
```
bnewbold, line 35: flagging for discussion: didn't seem like folks don't love `broken` (though I don't mind it). earlier term we were using was `dirty`.

No great words have come up yet... `fixme`, `out-of-sync`, `unsynchronized`, `disjoint`?

lexicons/com/atproto/sync/subscribeRepos.json
```diff
@@ -102,9 +103,47 @@
...
+        },
+        "rev": {
+          "type": "string",
+          "description": "The rev of the commit. This value must match that in the commit object."
+        },
+        "data": {
```
bnewbold, line 143: flagging for discussion: should we actually break this out top-level, or leave it in the commit? `rev` is the important one to have top-level.

lexicons/com/atproto/sync/getRepoStatus.json
bnewbold: going with `desynchronized`

lexicons/com/atproto/sync/subscribeRepos.json
bnewbold: Note: we discussed today and decided to stick with `prevData`

Issue #3386, opened 2025-01-17
Add skeletonReasonSponsored to app.bsky.feed.getFeedSkeleton


peat:
Hello bsky! The Graze Social team is working on a monetization platform for injecting sponsored posts into feeds. Our goal with this request is to:

- Make it clear which posts are being sent due to paid sponsorship, to allow for filtering/blocking/etc.
- Comply with FTC/EU regulations about disclosing advertising content.

The change adding `skeletonReasonSponsored` to the existing list of reasons in `getFeedSkeleton`

pfrazee:
https://github.com/bluesky-social/social-app/pull/7518#issuecomment-2603464561

peat:
I think it would be helpful to clarify attribution instead of just saying it's a sponsored post. I would like to make it clear who is paying for the post, and who is promoting it.

For example, if a cat food company wants one of their posts to show up in my feed, we should disclose that [feed owner DID] is promoting [post DID] from [cat food company DID]

That's an interesting social/financial data point, discloses the origin and reason for the content, and provides enough information to disambiguate appview disclosures so that the appview owner isn't blamed for the ad.

PR #3419, opened 2025-01-20
Feat: adds app.bsky.graph.defs#skeletonReasonSponsored lexicon.

aendra-rininsland:
See #3386.

I'm also working on a PR for bluesky-social/social-app to add UI for these. Creating this as a draft in the interim.

### DIFF
--- a/lexicons/app/bsky/feed/defs.json
+++ b/lexicons/app/bsky/feed/defs.json
@@ -185,7 +185,11 @@
         "post": { "type": "string", "format": "at-uri" },
         "reason": {
           "type": "union",
-          "refs": ["#skeletonReasonRepost", "#skeletonReasonPin"]
+          "refs": [
+            "#skeletonReasonRepost",
+            "#skeletonReasonPin",
+            "#skeletonReasonSponsored"
+          ]
         },
         "feedContext": {
           "type": "string",
@@ -202,6 +206,10 @@
       }
     },
     "skeletonReasonPin": {
+      "type": "object",
+      "properties": {}
+    },
+    "skeletonReasonSponsored": {
       "type": "object",
       "properties": {}
     },


aendra-rininsland:
Marking as "ready" given this needs to be merged before bluesky-social/social-app#7518.

lexicons/app/bsky/feed/defs.json
```diff
@@ -205,6 +209,15 @@
       "type": "object",
       "properties": {}
     },
+    "skeletonReasonSponsored": {
+      "type": "object",
+      "properties": {
+        "info": {
+          "type": "string",
+          "description": "Extended information about sponsored content, provided by the feed generator for transparency purposes."
```
bnewbold, line 217: Can probably hold off on changes until the `social-app` PR gets a reply, but quick notes:

- a description on the overall object clarifying that it existing at all means the feed item is sponsored content (we don't have good descriptions like this for `skeletonReasonRepost` or `skeletonReasonPin`, but we should)
- we are slowly but surely trying to lock down every `string` field with a `maxLength` to prevent shenanigans. maybe 2000 bytes (like `feedContext`)?

I'd be interested in considering if this should eventually be an AT-URI field pointing at a "context" record of some type. We are thinking about a generic "annotation" / "more context" mechanism for things like labels. That is still speculative though, and this is an optional field so there is a transition path.

aendra-rininsland: Thanks for the feedback! I've added a description and a maxLength as requested in the meantime while waiting for a response on bluesky-social/social-app#7518.

aendra-rininsland:
I've added two more fields, promotionCreator and promotionManager, to show who is running and benefitting from a promotional campaign. I've also renamed it from `skeletonReasonSponsored` to `skeletonReasonPromoted` given it's more generic language. See https://github.com/bluesky-social/social-app/pull/7518#issuecomment-2671965102

DGaffney:
Bumping on this thread - we're looking at adding subscriber tiers for feeds as an option on Graze, and using the `skeletonReason*` would be likely super helpful to denote who is a subscriber / sponsor in the feed, who helps pay for the feed etc

PR #3420, opened 2025-01-20
expose resetRouteRateLimits to the req handler

ghost:
Exposes in the generated type the functionality introduced in https://github.com/bluesky-social/atproto/pull/3348


dholms (approved):
yup 👍

PR #3447, opened 2025-01-23
[SH] Change keyBy return type from Record to Map

ghost:
This gives us more type-safety since record entries are always understood as existing on the record, while map entries are expected to be checked against `undefined`.

### DIFF
--- a/.changeset/honest-windows-appear.md
+++ b/.changeset/honest-windows-appear.md
@@ -1,0 +1,5 @@
+---
+"@atproto/common-web": minor
+---
+
+Make keyBy typing stricter


packages/common-web/src/arrays.ts
```diff
@@ -1,11 +1,8 @@
-export const keyBy = <T>(arr: T[], key: string): Record<string, T> => {
-  return arr.reduce(
-    (acc, cur) => {
-      acc[cur[key]] = cur
-      return acc
-    },
-    {} as Record<string, T>,
-  )
+export const keyBy = <T>(arr: T[], key: string): Map<string, T> => {
```
matthieusieben, line 1: Note that this is a breaking change, requiring a `minor` version update to the changeset.

matthieusieben, line 1: You also need to change the key type. When using `object[someValue]` in javascript, the key will indeed be casted to string. It is not the case with `Map`s.

```suggestion
export function keyBy<T, K extends keyof T>(arr: readonly T[], key: K): Map<T[K], T> {
```

Note that this is a breaking change, requiring a `minor` version update to the changeset (which is probably fine given the low number of [dependents](https://www.npmjs.com/package/@atproto/common-web?activeTab=dependents)).

matthieusieben (changes requested):
- changeset for `common-web` missing
- `keyBy` is not properly typed

matthieusieben (approved):

PR #3448, opened 2025-01-23
add watch to interactions, send feed with interactions

haileyok:


lexicons/app/bsky/feed/defs.json
```diff
@@ -243,13 +243,18 @@
...
+            "app.bsky.feed.defs#interactionShare",
+            "app.bsky.feed.defs#interactionVideoWatch"
           ]
         },
         "feedContext": {
           "type": "string",
           "description": "Context on a feed item that was originally supplied by the feed generator on getFeedSkeleton.",
           "maxLength": 2000
+        },
+        "videoWatchMs": {
```
matthieusieben, line 255: Would it make sense to use a more generic name here ?

For example: `interactionDurationMs` could also be for `interactionSeen`. This metric could also be usefull if we add an interaction that describes how long the user spent reading the responses threads.

lexicons/app/bsky/feed/sendInteractions.json
```diff
@@ -9,14 +9,19 @@
         "encoding": "application/json",
         "schema": {
           "type": "object",
-          "required": ["interactions"],
+          "required": ["interactions", "feed"],
```
matthieusieben, line 12: This change is not backwards compatible. not sure if that matters though.

PR #3449, opened 2025-01-23
Send prevs on firehose

dholms:
Add an experimental field `subscribeRepos` commit ops which includes the cid of the `prev` record in the case of updates/deletes. 

This is intended for testing of a new sync semantic that allows firehose consumers to do fully authenticated sync without keeping around a repo store.

Note on the implementation: I moved ops into the `CommitData` because they need a reference to the `prev` which can only be gotten from the transactional update of the user's repository, not in the "write preparation" step since two writes at the same time will cause one transaction to fail and replay while write preparation happens outside of the transaction.


lexicons/com/atproto/sync/subscribeRepos.json
```diff
@@ -206,6 +206,10 @@
         "cid": {
           "type": "cid-link",
           "description": "For creates and updates, the new record CID. For deletions, null."
+        },
+        "prev": {
+          "type": "cid-link",
+          "description": "EXPERIMENTAL. For deletes and updates, the CID  of the previous record. For creates, undefined."
```
matthieusieben, line 212: Should we use a `nullable` to be able to determine if the implementation supports this or not ?

devinivy: I had the same thought about detecting support.  Daniel pointed out that we don't need to worry about that yet for the experimental phase.  The tricky part here will be that we want to avoid optional, nullable fields.

bnewbold:
to cross-reference, I have a similar set of Lexicon changes proposed in this PR: https://github.com/bluesky-social/atproto/pull/3391

lexicons/com/atproto/sync/subscribeRepos.json
bnewbold: just to note, `nullable` and `optional` combined is a no-go (for golang), and that is especially relevant for these wire protocol details.

lexicons/com/atproto/sync/subscribeRepos.json
```diff
@@ -105,6 +105,10 @@
             "description": "List of new blobs (by CID) referenced by records in this commit."
           }
         },
+        "prevData": {
```
devinivy, line 108: Spoke about this possibly being the right moment to finalize deprecation of `prev` by removing it, also disambiguating `prevData`.

packages/pds/src/api/com/atproto/server/activateAccount.ts
```diff
@@ -53,7 +56,7 @@ export default function (server: Server, ctx: AppContext) {
         requester,
         account.handle ?? INVALID_HANDLE,
       )
-      await ctx.sequencer.sequenceCommit(requester, commitData, [])
+      await ctx.sequencer.sequenceCommit(requester, commitData)
```
devinivy, line 59: Seems like this is just blasting out the commit object—will this become a `#sync`?

packages/pds/src/repo/types.ts
```diff
@@ -42,6 +42,19 @@ export type PreparedDelete = {
   swapCid?: CID | null
 }
 
+export type CommitOp = {
+  action: 'create' | 'update' | 'delete'
+  path: string
+  cid: CID | null
+  prev?: CID
+}
```
devinivy, line 50: Wonder if it could be useful for correctness to discriminate on `action`, e.g.:
```ts
export type CommitOp =
  | {
      action: 'create'
      path: string
      cid: CID
      prev: null
    }
  | {
      action: 'update'
      path: string
      cid: CID
      prev: CID
    }
  | {
      action: 'delete'
      path: string
      cid: null
      prev: CID
    }
```

packages/pds/src/scripts/rebuild-repo.ts
```diff
@@ -71,10 +71,13 @@ export const rebuildRepo = async (ctx: AppContext, args: string[]) => {
...
       relevantBlocks: newBlocks,
       removedCids: toDelete,
+      ops: [],
+      blobs: new CidSet(),
+      prevData: null,
     }
   })
   await ctx.accountManager.updateRepoRoot(did, commit.cid, rev)
-  await ctx.sequencer.sequenceCommit(did, commit, [])
+  await ctx.sequencer.sequenceCommit(did, commit)
```
devinivy, line 80: Might make for a good `#sync` event.

packages/pds/src/sequencer/events.ts
```diff
@@ -1,74 +1,61 @@
...
   blocksToSend.addMap(commitData.newBlocks)
   blocksToSend.addMap(commitData.relevantBlocks)
 
-  // max 200 ops or 1MB of data
-  if (writes.length > 200 || blocksToSend.byteSize > 1000000) {
-    tooBig = true
+  let evt: CommitEvt
+
+  // If event is too big (max 200 ops or 1MB of data)
+  if (commitData.ops.length > 200 || blocksToSend.byteSize > 1000000) {
```
devinivy, line 19: Just a note, think we're still anticipating the byte size check and how it gets handled to be updated.

packages/pds/src/sequencer/events.ts
```diff
@@ -1,74 +1,61 @@
...
+      rebase: false,
+      tooBig: false,
+      repo: did,
+      commit: commitData.cid,
+      prev: commitData.prev,
+      rev: commitData.rev,
+      since: commitData.since,
+      blocks: await blocksToCarFile(commitData.cid, blocksToSend),
+      ops: commitData.ops,
+      blobs: commitData.blobs.toList(),
```
devinivy, line 50: I believe this can be left empty as part of deprecation.

packages/repo/src/mst/mst.ts
```diff
@@ -544,27 +544,46 @@ export class MST {
...
     for (let i = index; i < entries.length; i++) {
       const entry = entries[i]
       if (entry.isLeaf()) {
         yield entry
       } else {
-        for await (const e of entry.walkLeavesFrom(key)) {
-          yield e
+        for await (const node of entry.walkFrom(key)) {
+          yield node
         }
```
devinivy, line 574: I think there's a case that it's clearer as written, but I believe there is syntax specifically for this case if desired:
```ts
yield* entry.walkFrom(key)
```

packages/repo/src/mst/mst.ts
```diff
@@ -755,17 +774,33 @@ export class MST {
...
+    return blocks
+  }
+
+  async preceedingKey(key: string): Promise<string> {
+    const index = await this.findGtOrEqualLeafIndex(key)
     const prev = await this.atIndex(index - 1)
-    if (prev && prev.isTree()) {
-      await prev.addBlocksForPath(key, blocks)
+    if (!prev) {
+      return ''
```
devinivy, line 799: Just want to double-check that this can only happen on the leftmost key and not e.g. in some bad case like a missing `key`.  From reading it seems that is the case, just looking for a double-check.  (For example, one unintended result could be that `getCoveringProof()` ends-up being very large by starting from the left edge of the tree in a bad case.)

devinivy (approved):

packages/pds/src/repo/types.ts
dholms: It might be more correct, but it leads to really annoying verbose code in `repo/transactor`

This type is a direct reflection of the type in the firehose stream, and this data is only ever used in that context so I feel pretty comfortable about it

packages/pds/src/scripts/rebuild-repo.ts
dholms: agreed 👍

packages/pds/src/sequencer/events.ts
dholms: I'll do that in the follow up PR where we fully switch over to sync1.1 semantics

packages/repo/src/mst/mst.ts
dholms: oh nice didn't know this syntax!

packages/repo/src/mst/mst.ts
dholms: Yup that's right 👍

PR #3450, opened 2025-01-23
Appview: add user-agent for feeds and resolving blobs

devinivy:
Appivew trying to be a friendly netizen.  In particular, some servers reject requests without a user-agent, which can impact the appview's ability to resolve blobs.


matthieusieben (approved):
Looks pretty straight forward.

PR #3454, opened 2025-01-24
Fix did:web encoding/decoding

matthieusieben:
Current implementation does not properly encode/decode did webs as per spec:

https://w3c-ccg.github.io/did-method-web/#read-resolve

### DIFF
--- a/.changeset/gorgeous-lies-warn.md
+++ b/.changeset/gorgeous-lies-warn.md
@@ -1,0 +1,5 @@
+---
+"@atproto-labs/did-resolver": patch
+---
+
+Improve error descriptions
--- a/.changeset/lovely-news-prove.md
+++ b/.changeset/lovely-news-prove.md
@@ -1,0 +1,5 @@
+---
+"@atproto/did": patch
+---
+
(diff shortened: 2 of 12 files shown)


matthieusieben:
Good point. I added some tests in.

bnewbold:
didn't see this when it first went by, but note that we have a bunch of atproto-specific restrictions on `did:web` in the specs:

> In the context of atproto, only hostname-level did:web DIDs are supported: path-based DIDs are not supported. The same restrictions on top-level domains that apply to handles (eg, no .arpa) also apply to did:web domains. The special localhost hostname is allowed, but only in testing and development environments. Port numbers (with separating colon hex-encoded) are only allowed for localhost, and only in testing and development.

https://atproto.com/specs/did

I'm pretty confident about those restrictions, and that we don't want to support the full did:web functionality. I don't have a direct citation, but some W3C DID folks have also expressed that path-based `did:web` isn't particularly safe or recommended, and that `did:web` is still just a draft, so there isn't very strong pressure to implement it exactly as described in the draft doc.

Issue #3479, opened 2025-01-31
ValidationError: Message/seq must be an integer


maple-buice:
**Describe the bug**

At some point in the past few hours, `seq` values in Subscription messages started exceeding the max integer value (example value I just saw: `4308367499n`) and causing them to fail validation.

```
ValidationError: Message/seq must be an integer
    at integer (/Users/maple/Repos/feed-generator/node_modules/@atproto/lexicon/src/validators/primitives.ts:101:14)
    at validate2 (/Users/maple/Repos/feed-generator/node_modules/@atproto/lexicon/src/validators/complex.ts:24:14)
    at validateOneOf (/Users/maple/Repos/feed-generator/node_modules/@atproto/lexicon/src/util.ts:71:9)
    at object (/Users/maple/Repos/feed-generator/node_modules/@atproto/lexicon/src/validators/complex.ts:127:25)
    at validateOneOf (/Users/maple/Repos/feed-generator/node_modules/@atproto/lexicon/src/util.ts:70:9)
    at assertValidOneOf (/Users/maple/Repos/feed-generator/node_modules/@atproto/lexicon/src/util.ts:95:15)
    at assertValidXrpcMessage (/Users/maple/Repos/feed-generator/node_modules/@atproto/lexicon/src/validation.ts:64:12)
    at Lexicons.assertValidXrpcMessage (/Users/maple/Repos/feed-generator/node_modules/@atproto/lexicon/src/lexicons.ts:211:12)
    at Object.validate (/Users/maple/Repos/feed-generator/src/util/subscription.ts:26:27)
    at Subscription.[Symbol.asyncIterator] (/Users/maple/Repos/feed-generator/node_modules/@atproto/xrpc-server/src/stream/subscription.ts:42:32)
```

**To Reproduce**

1. Clone [feed-generator](https://github.com/bluesky-social/feed-generator)
2. `yarn install`
3. `yarn start`
4. Gaze upon a firehose of errors

**Expected behavior**

Validation to pass and the feed-generator Subscription to function properly, as it did a few hours ago.

**Details**

- Operating system: macOS
- Node version: 18.17.0

**Additional context**

matthieusieben:
`4308367499n` it is a BigInt. That value should be a `number`, not a `bigint`.

matthieusieben:
The issue is probably with https://github.com/bluesky-social/feed-generator ?

Bossett:
[...]

PR #3481, opened 2025-01-31
CBOR: parse safe uint64 as number

dholms:
`cbor-x` decodes `uint64`s as bigints even if they are safe to represent as javascript numbers (less than `2^57-1`). 

This was causing unnecessary validation errors in lexicon.

Fixes https://github.com/bluesky-social/atproto/issues/3479

### DIFF
--- a/.changeset/three-papayas-begin.md
+++ b/.changeset/three-papayas-begin.md
@@ -1,0 +1,5 @@
+---
+"@atproto/common": patch
+---
+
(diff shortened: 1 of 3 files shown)


DavidBuchanan314:
I assume `2^57-1` is a typo of `2^53-1`?

Eboyjax1212:
60467158-d739-40cf-b5e4-f930f120aade

PR #3495, opened 2025-02-04
:sparkles: Add a priority score value for ozone subjects

foysalit:
This PR allows moderators to set an arbitrary numeric value to any subject, as its priority score. Then, when querying statuses, subjects can be sorted by this value.


matthieusieben (approved):
looks pretty straight forward

packages/ozone/src/mod-service/index.ts
```diff
@@ -1098,12 +1104,22 @@ export class ModerationService {
       )
     }
 
+    if (minPriorityScore != null && minPriorityScore > 0) {
+      builder = builder.where(
+        'moderation_subject_status.priorityScore',
+        '>=',
+        minPriorityScore,
+      )
+    }
```
matthieusieben, line 1113: Are you sure we don't want to allow `priorityScore` to be `0`?

This would allow to distinguish "items with a priority score defined" from items "without a priority score (priority unknown)"

foysalit: yeah good point, I don't see a use-case for this right now but doesn't hurt to allow it.

PR #3496, opened 2025-02-04
Add followerRule threadgate

ghost:


packages/bsky/tests/views/threadgating.test.ts
```diff
@@ -242,8 +247,72 @@ describe('views with thread gating', () => {
...
+      { uri: post.ref.uriStr },
+      {
+        headers: await network.serviceHeaders(
+          sc.dids.dan,
+          ids.AppBskyFeedGetPostThread,
+        ),
+      },
+    )
+    assert(isThreadViewPost(danThread))
+    expect(danThread.post.viewer?.replyDisabled).toBe(true)
```
estrattonbailey, line 291: For sake of completeness, might as well add a check to confirm that Dan's reply wasn't accepted and included in the thread response

packages/bsky/tests/views/threadgating.test.ts
```diff
@@ -440,6 +516,23 @@ describe('views with thread gating', () => {
...
+      data: { thread: eveThread },
+    } = await agent.api.app.bsky.feed.getPostThread(
+      { uri: post.ref.uriStr },
+      {
+        headers: await network.serviceHeaders(
+          sc.dids.eve,
+          ids.AppBskyFeedGetPostThread,
+        ),
+      },
+    )
```
estrattonbailey, line 530: General note on this file: these tests are getting kinda hard to understand. The block/follow relationships are not obvious and the order of these tests matters to some extent.

Optional, but if it's helpful to you, we could consider creating a new seed with users specific to each test case, [like this one.](https://github.com/bluesky-social/atproto/blob/threadgate/1-followers/packages/bsky/tests/seed/feed-hidden-replies.ts) That way we can have explicitly named users and write some tests for new rules that are easier to understand. Those tests can live in a separate file alongside this one.

estrattonbailey (commented):
Looking good so far!

packages/bsky/tests/views/threadgating.test.ts
estrattonbailey: Ok I see, it's implicitly checked with the length check there. That's what I mean by these tests being kinda messy 😛 all good for now, thanks for the explainer 👍

.changeset/silver-horses-rush.md
estrattonbailey, line 1: Need to include `api` in here too

estrattonbailey (approved):
LGTM! Just need to add `api` to the changeset 👍

PR #3510, opened 2025-02-08
Label expiration

ghost:
This adds some handling of expired labels on the data-plane. It filters out expired labels.

I'm not very happy with the existing `neg` implementation, it just checks whether a label has `neg` and disregards the label in case it does. But the behavior should be a bit more specific, where a `neg` label removes a previously-added label that has the same `src`, `uri` and `val`. So just dropping all `neg` labels doesn't cut it. But I couldn't do anything better because it seems we are not indexing labels (we have a `label` table but we never write to it other than in tests), so I can't really handle the `neg` label there.

On the prod data-plane I implemented the `neg` as I described above (as well as `exp` with TTL), so the appview should always receive records without the expired labels and with the `neg`ed ones properly removed.

### DIFF
--- a/packages/bsky/src/data-plane/server/routes/labels.ts
+++ b/packages/bsky/src/data-plane/server/routes/labels.ts
@@ -1,5 +1,5 @@
 import { ServiceImpl } from '@connectrpc/connect'
-import { Selectable } from 'kysely'
+import { Selectable, sql } from 'kysely'
 import * as ui8 from 'uint8arrays'
 import { noUndefinedVals } from '@atproto/common'
 import { Service } from '../../../proto/bsky_connect'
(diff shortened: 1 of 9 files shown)


foysalit:
LGTM! might wanna have @devinivy take a final look tho

packages/bsky/src/data-plane/server/routes/labels.ts
```diff
@@ -14,10 +14,14 @@ export default (db: Database): Partial<ServiceImpl<typeof Service>> => ({
...
       return { labels: [] }
     }
+
     const res: LabelRow[] = await db.db
       .selectFrom('label')
       .where('uri', 'in', subjects)
       .where('src', 'in', issuers)
+      .where((qb) =>
+        qb.where('exp', 'is', null).orWhere(sql`exp::timestamp > now()`),
+      )
```
devinivy, line 24: Should we omit not only expired but also negated labels here?  The negation and expiration feel like they are part of the same kind of logic to me.  Though most importantly is really just that this has the same behavior as the production dataplane 👍

devinivy (approved):

PR #3520, opened 2025-02-11
Improve indexing for chat.bsky.moderation.getMessageContext

matthieusieben:
Currently, the `moderation_event_message_id_index` index covers the entire table. Because of this, every entry inserted in the `moderation_event` table will trigger an update of the index. This unnecessarily slows down inserts.

This PR changes that index for a partial index covering only the needed entries.


packages/ozone/src/db/migrations/20250211T132135150Z-moderation-event-message-partial-idx.ts
```diff
@@ -0,0 +1,24 @@
+import { Kysely, sql } from 'kysely'
+
+// support lookup for chat.bsky.moderation.getMessageContext
+
+export async function up(db: Kysely<unknown>): Promise<void> {
+  await db.schema.dropIndex('moderation_event_message_id_index')
```
devinivy, line 6: Just missing an `execute()` here.

packages/ozone/src/db/migrations/20250211T132135150Z-moderation-event-message-partial-idx.ts
```diff
@@ -0,0 +1,24 @@
...
+  await db.schema
+    .createIndex('moderation_event_message_id_idx')
```
devinivy, line 8: In practice this will need to be done `concurrently`.

devinivy: (Probably want to do the same with the `drop index` too.)

packages/ozone/src/db/migrations/20250211T132135150Z-moderation-event-message-partial-idx.ts
```diff
@@ -0,0 +1,24 @@
...
+    .on('moderation_event')
+    // https://github.com/kysely-org/kysely/issues/302
+    .expression(
+      sql`"subjectMessageId") WHERE ("subjectMessageId" IS NOT NULL AND "action" = 'tools.ozone.moderation.defs#modEventReport'`,
+    )
+    .execute()
+}
+
+export async function down(db: Kysely<unknown>): Promise<void> {
+  await db.schema.dropIndex('moderation_event_message_id_idx')
```
devinivy, line 18: Just missing an execute() here.

devinivy (approved):

PR #3524, opened 2025-02-12
com.atproto.sync.listReposByCollection Lexicon, for collections directory

bnewbold:
Pairs with: https://github.com/bluesky-social/indigo/pull/929

Have not run codegen in this repo; usually do that just before merge (after approval).

There was some back and forth on whether we should return array-of-objects or just array-of-strings (DIDs). I think it is good to have this more consistent with `listRepos`, which returns array-of-objects.

@brianolson had a minimum bound on the `limit` parameter (50). For consistency with other endpoints, I made the Lexicon minimum here 1.

This is just the `listReposByCollection` endpoint; I didn't do a Lexicon for `com.atproto.sync.countReposByCollection`.

### DIFF
--- a/lexicons/com/atproto/sync/listReposByCollection.json
+++ b/lexicons/com/atproto/sync/listReposByCollection.json
@@ -14,8 +14,8 @@
             "type": "integer",
             "description": "Maximum size of response set. Recommend setting a large maximum (1000+) when enumerating large DID lists.",
             "minimum": 1,
-            "maximum": 5000,
-            "default": 100
+            "maximum": 2000,
+            "default": 500
           },
           "cursor": { "type": "string" }
         }
--- a/lexicons/com/atproto/sync/listReposByCollection.json
+++ b/lexicons/com/atproto/sync/listReposByCollection.json
@@ -1,0 +1,45 @@
+{
+  "lexicon": 1,
+  "id": "com.atproto.sync.listReposByCollection",
+  "defs": {
+    "main": {
+      "type": "query",
+      "description": "Enumerates all the DIDs which have records with the given collection NSID.",
+      "parameters": {
+        "type": "params",
+        "required": ["collection"],
+        "properties": {
+          "collection": { "type": "string", "format": "nsid" },
+          "limit": {
+            "type": "integer",
+            "minimum": 1,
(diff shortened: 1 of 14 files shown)


lexicons/com/atproto/sync/listReposByCollection.json
```diff
@@ -0,0 +1,45 @@
...
+      "parameters": {
+        "type": "params",
+        "required": ["collection"],
+        "properties": {
+          "collection": { "type": "string", "format": "nsid" },
+          "limit": {
+            "type": "integer",
+            "minimum": 1,
+            "maximum": 500,
+            "default": 50
```
brianolson, line 17: I think the default should be much higher, paging through millions at 50/call would be bad. The db access is super cheap, this is just about what a reasonable network blob is. 32 byte did * 1000 would only be 32kB of data so at least that. Maybe default 500 or 1000, upper limit 5000 or 10000.

bnewbold: I think a common case is going to be folks just iterating through hundreds of accounts, not millions, right? eg, looking for all feed generator declarations, or statusphere users. I think folks doing this bulk should configure the limit; it is nice to have it smaller for devs poking around with `curl`.

I'll bump to default=100 and max=5000

(I don't feel super strongly about this and could bump higher as well)

devinivy: My inclination would be to follow `sync.listRepos` which has a pretty similar use case.  The number of repos on a host vs. within a collection seems like they both span large and small lists.

devinivy (approved):

devinivy:
@bnewbold if you do a codegen, please also add a changeset for `@atproto/api` to ensure the sdk gets marked for release.  I usually do this with `pnpm changeset`.

brianolson (approved):

lexicons/com/atproto/sync/listReposByCollection.json
bnewbold: `sync.listRepos` has min=1, default=500, max=1000.

I went with min=1, default=500, max=2000 to be mostly-aligned

PR #3525, opened 2025-02-12
Rework of authorization request hooks

matthieusieben:
This PR essentially adds three new "hooks" in the `OAuthProvider`.

- `onAuthorized` which is triggered when the user "authorizes" a client (a `code` is issued). This allows to retrieve the user's IP address through the `deviceMetadata` hook data.
- `onTokenCreated` which is triggered when the code is exchanged, by the OAuth client, for a token.
- `onTokenRefreshed` which is triggered when a refresh token is exchanged, by the OAuth client, for a new access token

Since these hooks are "events" and do not allow to manipulate any of the authorization data, they were prefixed with "`on`".

In order to distinguish hooks that do allow to extend the authorization data, existing hooks where renamed by replacing their prefix with "`get`". This is a breaking change that causes a `minor` version change.

Because we want to be able to retrieve the user's IP address from the `onAuthorized` hook, this PR introduces the ability to configure the `DeviceManager` options (which used to use a hard-coded config). This is especially useful to be able to set the `trustProxy` config, that determines if the `x-forwarded-*` headers can be trusted to determine the user's IP address.

Finally, this PR removes the `onClientInfo` hook configured in the PDS. This hook did not have any effect and its code is unlikely to stay the same when the Bluesky App actually starts using OAuth credentials.


packages/pds/src/oauth/provider.ts
```diff
@@ -46,12 +46,6 @@ export class PdsOAuthProvider extends OAuthProvider {
       // the PDS can use tokenId as access tokens. This allows the PDS to
       // always use up-to-date token data from the token store.
       accessTokenType: AccessTokenType.id,
-
-      onClientInfo: (clientId) => ({
-        isFirstParty: clientId === 'https://bsky.app/',
-        // @TODO make client client list configurable:
-        isTrusted: undefined,
```
devinivy, line 53: Could use a little context for this change.  If we intend to land it, also could use a changeset.

matthieusieben: This is currently not in use and does not match the clientId that the app should have (it should be a url to a JSON file).

Until we actually do have a clientId for our app, I figured it is better not to use this at all.

packages/oauth/oauth-provider/src/device/device-details.ts
```diff
@@ -1,42 +0,0 @@
-import { IncomingMessage } from 'node:http'
```
matthieusieben, line 1: These utils where moved in `lib/http/request.ts` (see bellow)

packages/oauth/oauth-provider/src/oauth-provider.ts
```diff
@@ -134,7 +134,7 @@ export {
 export type RouterOptions<
   Req extends IncomingMessage = IncomingMessage,
   Res extends ServerResponse = ServerResponse,
-> = {
+> = DeviceManagerOptions & {
```
matthieusieben, line 137: Allows configuring the `DeviceManager` options when creating a the provider's middleware.

packages/pds/src/oauth/provider.ts
devinivy: Ah, got it! If there's no change in behavior then I am ambivalent about whether there's a changeset for it.

PR #3532, opened 2025-02-13
Appview: apply needs-review to individual records

devinivy:
These labels previously only applied at the account level.  This provides more granular application at the record level.

### DIFF
--- a/packages/bsky/tests/views/labels-needs-review.test.ts
+++ b/packages/bsky/tests/views/labels-needs-review.test.ts
@@ -43,127 +43,272 @@ describe('bsky needs-review labels', () => {
     await network.processAll()
 
     AtpAgent.configure({ appLabelers: [network.ozone.ctx.cfg.service.did] })
-    await network.bsky.db.db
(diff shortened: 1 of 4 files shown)


packages/bsky/tests/views/labels-needs-review.test.ts
```diff
@@ -43,127 +43,272 @@ describe('bsky needs-review labels', () => {
...
-        return notif.reason === 'quote' && notif.author.did === sc.dids.geoff
-      }),
-    ).toBe(false)
-    expect(
-      notifications.some((notif) => {
-        return notif.reason === 'mention' && notif.author.did === sc.dids.geoff
-      }),
-    ).toBe(false)
-  })
+    it('applies to thread replies.', async () => {
```
devinivy, line 75: I recommend reviewing this file with whitespace ignored.  The same tests are essentially run twice, first with geoff labeled, and then all of geoff's posts labeled.

dholms (approved):
yup yup ✅

PR #3553, opened 2025-02-18
Add x-forwarded-for to PDS-proxied requests to entryway

devinivy:
As advertised: adding x-forwarded-for to PDS-proxied requests to entryway.

Can get rid of some minor noise in review by ignoring whitespace changes.

### DIFF
--- a/packages/pds/src/api/proxy.ts
+++ b/packages/pds/src/api/proxy.ts
@@ -36,19 +36,17 @@ export function authPassthru(req: IncomingMessage) {
   }
 }
 
-export const forwardFor = (
+// @NOTE this function may mutate its params input
+// future improvement here would be to forward along all untrusted ips rather than just the first (req.ip)
+export const forwardedFor = (
   req: express.Request,
   params: HeadersParam | undefined,
 ) => {
   const result: HeadersParam = params ?? { headers: {} }
-  let forwarded: string[] = []
-  if (req.headers['x-forwarded-for']) {
-    forwarded = forwarded.concat(req.headers['x-forwarded-for'])
+  const ip = req.ip
+  if (ip) {
+    result.headers['x-forwarded-for'] = ip
   }
-  if (req.socket.remoteAddress) {
-    forwarded.push(req.socket.remoteAddress)
-  }
-  result.headers['x-forwarded-for'] = forwarded.join(',')
   return result
 }
 
--- a/packages/pds/src/context.ts
+++ b/packages/pds/src/context.ts
@@ -26,7 +26,7 @@ import {
 } from '@atproto-labs/fetch-node'
 import { AccountManager } from './account-manager'
 import { ActorStore } from './actor-store/actor-store'
-import { authPassthru, forwardFor } from './api/proxy'
+import { authPassthru, forwardedFor } from './api/proxy'
 import {
   AuthVerifier,
   createPublicKeyObject,
@@ -407,11 +407,11 @@ export class AppContext {
       this.cfg.entryway.did,
       lxm,
     )
-    return forwardFor(req, headers)
+    return forwardedFor(req, headers)
   }
 
   entrywayPassthruHeaders(req: express.Request) {
-    return forwardFor(req, authPassthru(req))
+    return forwardedFor(req, authPassthru(req))
   }
 
   async serviceAuthHeaders(did: string, aud: string, lxm: string) {
--- a/packages/pds/src/index.ts
+++ b/packages/pds/src/index.ts
@@ -122,7 +122,7 @@ export class PDS {
     server = API(server, ctx)
 
     const app = express()
-    app.set('trust proxy', true)
+    app.set('trust proxy', ['loopback', 'linklocal', 'uniquelocal'])
     app.use(loggerMiddleware)
     app.use(compression())
     app.use(authRoutes.createRouter(ctx)) // Before CORS


packages/pds/src/api/com/atproto/server/util.ts
```diff
@@ -117,3 +118,26 @@ const assertValidDocContents = async (
...
+
+export const forwardIp: IForwardIp = <T extends HeadersParam>(
+  req: express.Request,
+  params: T | undefined,
+) => {
+  const ip = req.ip
+  const result: T | HeadersParam = params ?? { headers: {} }
+  if (ip) {
+    result.headers['x-forwarded-for'] = ip
+  }
```
matthieusieben, line 130: Wouldn't it be more accurate (in terms of semantics of that header), and easier, to simply add the `req.socket.remoteAddress` to the already provided list of `x-forwarded-for` ?

matthieusieben (commented):
Looks like the catchall handler (pipethrough) does not apply this. Is this on purpose?

packages/pds/src/api/com/atproto/admin/sendEmail.ts
```diff
@@ -28,11 +29,13 @@ export default function (server: Server, ctx: AppContext) {
...
-              ctx.cfg.entryway.did,
-              ids.ComAtprotoAdminSendEmail,
-            ),
+            await ctx
+              .serviceAuthHeaders(
+                recipientDid,
+                ctx.cfg.entryway.did,
+                ids.ComAtprotoAdminSendEmail,
+              )
+              .then((x) => forwardIp(req, x)),
```
matthieusieben, line 38: Would it make sense to add a special `ctx.entrywayServiceAuthHeaders(req, did, lxm)` method that wraps this ? It makes it less likely, in the future, that someone forgets to forward the ip when proxying fo entryway, don't you think ?

matthieusieben: The following lines, for example, should probably be using the same logic:

https://github.com/bluesky-social/atproto/blob/75c05d50b6a7283555f01eb783fc968b29d62d6b/packages/pds/tests/entryway.test.ts#L120-L124

matthieusieben (commented):
It looks like you missed one spot:

https://github.com/bluesky-social/atproto/blob/75c05d50b6a7283555f01eb783fc968b29d62d6b/packages/pds/src/api/com/atproto/server/deleteSession.ts#L11

packages/pds/src/api/com/atproto/admin/sendEmail.ts
devinivy: Yupyup good call, should be set now.

devinivy:
> Looks like the catchall handler (pipethrough) does not apply this. Is this on purpose?

Yeah, not needed at the moment for service proxying and if we do enable it there it may call for some careful configuration.

packages/pds/src/api/com/atproto/server/util.ts
devinivy: Yeah that seems like a good call, should be resolved now.

packages/pds/src/api/proxy.ts
```diff
@@ -45,9 +32,24 @@ export function authPassthru(req: IncomingMessage, withEncoding?: boolean) {
...
+) => {
+  const result: HeadersParam = params ?? { headers: {} }
+  let forwarded: string[] = []
+  if (req.headers['x-forwarded-for']) {
+    forwarded = forwarded.concat(req.headers['x-forwarded-for'])
   }
+  if (req.socket.remoteAddress) {
+    forwarded.push(req.socket.remoteAddress)
+  }
+  result.headers['x-forwarded-for'] = forwarded.join(',')
```
matthieusieben, line 51: nit: Do we need to worry that this mutates its input `params` ?

matthieusieben (approved):
Love it when the diff has a negative delta 🤩

I just have two concerns:
- The PDS uses `true` as `trust proxy`, which it shouldn't
- The PDS currently exposes its trusted ips when proxying to entryway

With this change, the PDS will expose **all** the IPS (including those of its internal deployment setup: load balancer, etc.) to entryway. This means that if the `trust proxy` addresses on the entryway does not match the addresses trusted by the PDS, the entryway will end-up deducing the wrong client ip.

We could consider that the PDS and Entryway are part of the same "deployment", in which case this would be fine. However, since the PDS can be deployed as stand-alone (self-hosted), I think that we **must** setup the `trust proxy` setting, at least for the sake of correctly computing `req.ip`, wherever we rely on it (rate limit, `OauthProvider` session management).

Since we would be setting `trust proxy` on the PDS, I thing it might make sense to filter out, from the proxied list of ips, those trusted by the PDS. Even if we do consider the PDS and Entryway as part of the same deployment.

Pseudo code here:

```ts
import forwarded from 'forwarded'


// ...


// Compute the "x-forwarded-for" header when proxying towards entryway by
// removing the ips from our local deployment setup (load balancer, etc.)

// @NOTE Computed by Express when setting "trust proxy"
const trust = req.app.get('trust proxy fn')

// @NOTE parses addresses backwards (closest to farthest)
const addrs = forwarded(req)

// Find the first untrusted IP
const untrustedIdx = addrs.findIndex(ip => !trust(ip))

// @NOTE This is *not* the same as `req.ips`. `req.ips` only contains the
// trusted ips + the first untrusted ip. Here we want to keep only the untrusted
// ips
const untrustedIps = untrustedIdx === -1 ? undefined : addrs.slice(untrustedIdx)

const newForwardedFor = untrustedIps?.reverse().join(',')

// Either the the trusted proxies do not set `x-forwarded-for`, or the 'trust
// proxy' setting is too lax. This can lead to abuse (un-reliable logs, unable
// to comply to legal requests, unable to properly enforce rate limiting) which
// is why we require a proper proxy configuration here.
if (!newForwardedFor) throw new Error('Invalid proxy configuration')
```

PR #3554, opened 2025-02-18
:sparkles: Allow filtering ozone team list by role and status

foysalit:


packages/ozone/src/team/index.ts
```diff
@@ -18,17 +18,37 @@ export class TeamService {
...
     limit?: number
+    disabled?: boolean
+    roles?: string[]
   }): Promise<{ members: Selectable<Member>[]; cursor?: string }> {
     let builder = this.db.db.selectFrom('member').selectAll()
     if (cursor) {
       builder = builder.where('createdAt', '>', new Date(cursor))
     }
+    if (roles?.length) {
+      builder = builder.where('role', 'in', this.getRolesList(roles))
```
matthieusieben, line 47: It looks like the filtering of invalid roles will not have any effect and could be simplified like so ?

```suggestion
      builder = builder.where('role', 'in', roles)
```

matthieusieben (approved):
Looks good

PR #3557, opened 2025-02-19
OAuthProvider: Update "trustProxy" options to allow function

matthieusieben:


packages/oauth/oauth-provider/src/lib/http/request.ts
```diff
@@ -223,5 +243,8 @@ function extractPort(
     }
   }
 
-  return req.socket.remotePort
+  const port = req.socket.remotePort
+  if (port != null) return port
+
+  throw new Error('Could not determine port')
```
devinivy, line 249: This feels like it is bound to happen in some cases, e.g. in the case `extractPort()` is called when a socket has been destroyed, and might be prone to escalate to a crash.

I see that there should not be any regression here compared to how it's worked before, so I am not sure we need to address this here, but perhaps worth a closer look.

devinivy (approved):

PR #3567, opened 2025-02-22
remove deprecated firehose message types

bnewbold:
So far this is just lexicon JSON updates and codegen.

Doesn't fully remove all the relevant code paths; probably makes sense to do that part along with implementing `#sync` events? As many of the same bits are touched. And some care is probably needed with sequencer, which might contain old message types.


bnewbold:
This got done in: https://github.com/bluesky-social/atproto/pull/3585

PR #3568, opened 2025-02-22
:bug: Add $type to ozone subject status

foysalit:


matthieusieben (changes requested):
the `$type` property is, per spec, only needed in order to discriminate when there is a "doubt" about the possible value: in unions and in records.

We should not make explicit the `$type` from places where there is no doubt.

Issue #3292, opened 2024-12-25
OAuth client in CloudFlare environments


JoviDeCroock:
**Describe the bug**

There's a few issues with implementing the OAuth client in environments like CloudFlare and other WinterCG environments.

I have been able to circumvent these myself with a few tricks

- Fetch.cache is not available, in `@atproto/oauth-client` we are using `cache: 'no-cache'` which isn't supported in CloudFlare, we could circumvent this by using `cache-control: no-cache` instead.
- Fetch.redirect: error is not available, in `@atproto/did-resolver` we are using `redirect: 'error'` which does not work, we need to use manual or follow.
- When we use a tool like `KV` we have to restore the `state.dpopKey` to a `JoseKey` (more a docs issue I guess) which can be done like `kvResult.dpopKey = new JoseKey(kvResult.dpopKey.jwks);`
- The handle-resolvers currently available all require DNS which isn't supported, instead I've made
  ```ts
    handleResolver: {
      resolve: async (handle) => {
        const result = await fetch(
          "https://bsky.social/xrpc/com.atproto.identity.resolveHandle?handle=" +
            handle,
          {
            cf: {
              cacheEverything: true,
            },
          },
        ).then((x) => x.json<{ did: ResolvedHandle }>());
        return result.did as ResolvedHandle;
      },
    },
  ```

**Expected behaviour**

I mainly wanted to describe the issues and discuss them before jumping on pull requests. I'm not sure how to replace `redirect: error` however the other three look feasible.

kaytwo:
I did something very similar - basically monkey patched fetch and `Request`, super unclean but got the job done: https://gist.github.com/kaytwo/d5e553a6fce20e28f6d5573a520fb525

As for a pull request - I haven't looked at their code for it, but I vaguely recall when working on this that the difference was a non-issue, because neither `redirect: error` nor `redirect: manual` would lead to a success state in the existence of a redirect, so just switching to `redirect: manual` might be fine.

zairex:
It looks like the switch from axios to `fetch` was recent (https://github.com/bluesky-social/atproto/pull/3177) so ideally there's still room for changes. I wonder if (and would propose that) the `fetch` implementation could be an optional argument to some of these utilities, so that an alternate version could be provided in cases like these. The native `fetch` would still be the default, but it would be easier to implement workarounds like the ones @kaytwo suggested in their gist.

I would be happy to implement a PR myself if that sounded like a reasonable change.

[...]

PR #3569, opened 2025-02-23
feat(identity): allow custom fetch implementation

molisani:
Implements the suggestion proposed by https://github.com/bluesky-social/atproto/issues/3292#issuecomment-2617130922 which provides an easier way to solve the issue manually.

As part of #3177 the dependency on `axios` was removed and HTTP requests in this package were migrated to use `fetch`. However, the current implementation calls the global `fetch` (indirectly accessing `globalThis`). If there is a need to replace the `fetch` implementation with a custom one, this package requires that the `globalThis.fetch` implementation is replaced globally.

This change adds `fetch?: typeof globalThis.fetch` to the options for `IdentityResolver`, `HandleResolver`, and `DidResolver` so that users can provide an alternate implementation of `fetch`. If no alternate implementation is provided then these options default to `globalThis.fetch`, maintaining the exact same behavior as the current implementation.

### DIFF
--- a/packages/identity/src/did/did-resolver.ts
+++ b/packages/identity/src/did/did-resolver.ts
@@ -9,11 +9,11 @@ export class DidResolver extends BaseResolver {
 
   constructor(opts: DidResolverOpts) {
     super(opts.didCache)
-    const { timeout = 3000, plcUrl = 'https://plc.directory' } = opts
+    const { fetch = globalThis.fetch, timeout = 3000, plcUrl = 'https://plc.directory' } = opts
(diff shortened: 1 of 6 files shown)


molisani:
To any reviewers: it looks like the tests for this package work by spinning up a local server to process the requests. Those tests continue to work with my changes, so I didn't write any new tests. Please let me know if it makes sense to extend the test suite to exercise this new functionality (ex: provide a mock `fetch` that directly returns the expected result without going to the local server).

Issue #3368, opened 2025-01-14
Missing sort keys in various responses that return lists of `ProfileView`


tunjid:
**Is your feature request related to a problem? Please describe.**

I'm building an open-source offline-first client for bluesky uing Compose Multiplatform: https://github.com/tunjid/heron.

Requests typically load data from a local database first, with calls to the API to refresh the local database. Certain endpoints do not return information that allow local sorting however, making combining local requests with updates from the network impossible.

A summary of endpoints without the desired behavior and those with the behavior are summarized in the table below:

| Endpoint      | Model Type     | sort key  | Returned in API Response |
|---------------|----------------|-----------|--------------------------|
| getFollows    | `ProfileView`  | ??        | No                       |
| getFollowers  | `ProfileView`  | ??        | No                       |
| getRepostedBy | `ProfileView`  | ??        | No                       |
| getLikes      | `GetLikesLike` | indexedAt | Yes                      |
| getTimeline   | `FeedViewPost` | indexedAt | Yes                      |
| getFeed       | `FeedViewPost` | indexedAt | Yes                      |
| getListFeed   | `FeedViewPost` | indexedAt | Yes                      |
| getAuthorFeed | `FeedViewPost` | indexedAt | Yes                      |

**Describe the solution you'd like**

It would appear the issues tend to be with endpoints that return a `ProfileView`. I assume:

`getFollows`: Sorted by the time a profile was followed
`getFollowers`: Sorted by the time a profile added a follower
`getRepostedBy`: Sorted by the time a post was reposted.

All of these are similar to the case of `getLikes`, however `getLikes` wraps the `ProfileView` instances returned in the following structure with the requisite information:

``` kotlin
public data class GetLikesLike(
  public val indexedAt: Timestamp,
  public val createdAt: Timestamp,
  public val actor: ProfileView,
)
```

This can be applied to `getFollows`, `getFollowers` and `getRepostedBy`, allowing for local sorting of the query to resolve with the network response when available.

**Describe alternatives you've considered**

Coming up with a best guess heuristic to approximate network sort keys locally.

**Additional context**

Similar and related issues:

https://github.com/bluesky-social/atproto/issues/1574
https://github.com/bluesky-social/atproto/issues/3342

PR #3570, opened 2025-02-24
Update lexicons that return a list of profile views to include their sort keys

tunjid:
This is a:

* Fix for #3368 
* Lexicon update

This PR would like to update the lexicon for a few requests that return `ProfileView` lists to also include the time based sort key in an `indexedAt` field for the backing model that the list is generated from, as well as a `createdAt` field.

In other words, make the changed requests match their `getLikes` sibling which provides the requisite information.

This allows offline first clients to seamlessly combine cached local responses with updates from the server.

As this is a breaking change, other alternatives include:

* Adding a new field in the #ProfileView lexicon definition for a `sourceModel` or something that has the:
    * `cid` for the associated model
    * `indexedAt` for the associated model
    * `createdAt` for the associated model
    
* Adding new methods to the API (`getBlockedActors`, `getLikingActors`, e.t.c) and a new def for that has a `ProfileView` embedded and a field for the requested metadata
    * `cid` for the associated model
    * `indexedAt` for the associated model
    * `createdAt` for the associated model

### DIFF
--- a/lexicons/app/bsky/feed/getRepostedBy.json
+++ b/lexicons/app/bsky/feed/getRepostedBy.json
@@ -41,11 +41,20 @@
               "type": "array",
               "items": {
                 "type": "ref",
-                "ref": "app.bsky.actor.defs#profileView"
+                "ref": "#repostedBy"
               }
             }
           }
         }
+      }
+    },
+    "repostedBy": {
+      "type": "object",
+      "required": ["indexedAt", "createdAt", "actor"],
+      "properties": {
+        "indexedAt": { "type": "string", "format": "datetime" },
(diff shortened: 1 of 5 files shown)


mary-ext:
this would be a breaking change unfortunately

tunjid:
@mary-ext yes it unfortunately is, which significantly reduces the chances of it happening but its worth noting:

* Without this, is is impossible to create an offline first client, there is just no way to make local sorting match that on the server.
* Keeps the lexicon inconsistent. `GetLikes` will be the only request with the desired behavior. Why is it special?

tunjid:
Closing this in favor of https://github.com/bluesky-social/atproto/pull/3842

PR #3573, opened 2025-02-24
Notification cursor delay

ghost:
This PR formalizes the `app.bsky.notification.listNotifications` cursor to be a datetime string and allow a configurable delay to be applied to it.

It is easier to review the commits in order.

### DIFF
--- a/packages/bsky/src/api/app/bsky/notification/listNotifications.ts
+++ b/packages/bsky/src/api/app/bsky/notification/listNotifications.ts
@@ -105,15 +105,10 @@ export const delayCursor = (
   delayMs: number,
 ): string => {
   const nowMinusDelay = Date.now() - delayMs
-  if (cursorStr === undefined) {
-    return new Date(nowMinusDelay).toISOString()
-  }
-  try {
-    const cursor = new Date(cursorStr).getTime()
-    return new Date(Math.min(cursor, nowMinusDelay)).toISOString()
-  } catch (err) {
-    return cursorStr
-  }
+  if (cursorStr === undefined) return new Date(nowMinusDelay).toISOString()
+  const cursor = new Date(cursorStr).getTime()
+  if (isNaN(cursor)) return cursorStr
+  return new Date(Math.min(cursor, nowMinusDelay)).toISOString()
 }
 
 const skeleton = async (


lexicons/app/bsky/notification/listNotifications.json
```diff
@@ -33,7 +33,7 @@
           "type": "object",
           "required": ["notifications"],
           "properties": {
-            "cursor": { "type": "string" },
+            "cursor": { "type": "string", "format": "datetime" },
```
devinivy, line 36: In general the cursor here may be opaque—we choose to use a datetime in our appview implementation. So we don't need to adopt this breaking lexicon change as part of this work.

packages/bsky/src/api/app/bsky/notification/listNotifications.ts
```diff
@@ -15,11 +15,11 @@ import {
 import { Notification } from '../../../../proto/bsky_pb'
 import { uriToDid as didFromUri } from '../../../../util/uris'
 import { Views } from '../../../../views'
-import { clearlyBadCursor, resHeaders } from '../../../util'
+import { resHeaders } from '../../../util'
 
 export default function (server: Server, ctx: AppContext) {
   const listNotifications = createPipeline(
-    skeleton,
+    skeleton(ctx.cfg.notificationsDelayMs),
```
devinivy, line 22: We don't need to wrap `skeleton` in a function like this, as `ctx` is already plumbed through the pipeline.  You'll just want to reflect `cfg` on the `Context` type.  See `getPostThread` for an example.

packages/bsky/tests/views/notifications.test.ts
```diff
@@ -497,17 +498,150 @@ describe('notification views', () => {
...
+    let delayAgent: AtpAgent
+    let delaySc: SeedClient
+    let delayAlice: string
+
+    beforeAll(async () => {
+      delayNetwork = await TestNetwork.create({
+        bsky: {
+          notificationsDelayMs: delay3s,
         },
+        dbPostgresSchema: 'bsky_views_notifications',
```
devinivy, line 516: These will be the same tables as the rest of above notification tests.  I wonder if that was unintentional, since I see the seed gets played on it again.

packages/bsky/src/api/app/bsky/notification/listNotifications.ts
```diff
@@ -93,24 +94,45 @@ const paginateNotifications = async (opts: {
...
+ */
+export const delayCursor = (
+  cursorStr: string | undefined,
+  delayMs: number,
+): string => {
+  const nowMinusDelay = Date.now() - delayMs
+  if (cursorStr === undefined) {
+    return new Date(nowMinusDelay).toISOString()
+  }
+  const cursor = new Date(cursorStr).getTime()
```
devinivy, line 111: I'd propose we pass the value through rather than manipulate it if it does not parse as a date.

packages/bsky/tests/views/notifications.test.ts
```diff
@@ -497,17 +498,152 @@ describe('notification views', () => {
...
+          },
+        )
+
+      // @NOTE: This is not perfectly synchronized. It relies on it taking less than 3s to run between
+      // the seeds and the first queries. This assumption should be true in most test runs.
+      expect(fullBeforeDelay.data.notifications.length).toEqual(0)
+      expect(results(paginatedAllBeforeDelay)).toEqual(
+        results([fullBeforeDelay.data]),
+      )
+      await new Promise((resolve) => setTimeout(resolve, delay3s))
```
devinivy, line 576: There is a `wait()` helper that does this in `@atproto/common`.

packages/bsky/tests/views/notifications.test.ts
```diff
@@ -497,17 +498,152 @@ describe('notification views', () => {
...
+          {
+            headers: await delayNetwork.serviceHeaders(
+              delayAlice,
+              ids.AppBskyNotificationListNotifications,
+            ),
+          },
+        )
+
+      // @NOTE: This is not perfectly synchronized. It relies on it taking less than 3s to run between
+      // the seeds and the first queries. This assumption should be true in most test runs.
```
devinivy, line 571: The timing dependence and 3sec wait is a little rough, hmmm.  I take it using the timer mocks didn't work out in these tests?

packages/bsky/src/api/app/bsky/notification/listNotifications.ts
```diff
@@ -93,24 +94,49 @@ const paginateNotifications = async (opts: {
...
+  const nowMinusDelay = Date.now() - delayMs
+  if (cursorStr === undefined) {
+    return new Date(nowMinusDelay).toISOString()
+  }
+  try {
+    const cursor = new Date(cursorStr).getTime()
+    return new Date(Math.min(cursor, nowMinusDelay)).toISOString()
+  } catch (err) {
+    return cursorStr
+  }
```
devinivy, line 116: The logic behind the try/catch here is a little circuitous:
 1. `cursorStr` is not a valid date string so `cursor = new Date(cursorStr).getTime()` is `NaN`
 2. `Math.min(cursor, nowMinusDelay)` is `NaN` since `cursor` was `NaN`.
 3. `new Date(Math.min(cursor, nowMinusDelay)).toISOString()` throws on the method call since the date object is `new Date(NaN)` which is an `Invalid Date`.

How would you feel about handling the bad date string early on instead?
```suggestion
    const cursor = new Date(cursorStr).getTime()
    if (isNaN(cursor)) return cursorStr
    return new Date(Math.min(cursor, nowMinusDelay)).toISOString()
```

devinivy (approved):

PR #3574, opened 2025-02-24
nobody uses notifyOfUpdate, make it official

brianolson:


bnewbold (commented):
Seems good to me.

I'm not sure what the status of this is overall; eg is it getting called by the PDS? Any impact of *not* having this endpoint implemented?

brianolson:
this seems like a thing that the relay should have implemented, but it was always just a no-op stub there

bnewbold (approved):

bnewbold:
Before merging this, you should run:

```
make deps
make build
make codegen
```

which should only update comments in typescript code (no actual code/behavior change). The reason to do that codegen is to reduce confusion for whoever the next person is that runs codegen and sees unexpected changes. For now, we don't need a changelog updae ("changeset" file).

PR #3585, opened 2025-02-25
Wrap sync semantics

dholms:
Followup to https://github.com/bluesky-social/atproto/pull/3391 and https://github.com/bluesky-social/atproto/pull/3449

This PR wraps up some loose threads on the switch to [Sync v1.1](https://github.com/bluesky-social/proposals/tree/main/0006-sync-iteration) semantics.

Namely:
- introduce new `sync` event and emit on accountActivation and repo rebuild
- deprecate handle & tombstone events and remove from lexicons
- deprecate `tooBig` events (though we still need to include the flag set to `false` on all commit events for lexicon backwards-compatibility)
- increase the size limit on commit events from 1MB to 2MB


packages/bsky/src/lexicon/lexicons.ts
```diff
@@ -1860,16 +1860,6 @@ export const schemaDict = {
...
-            rkeyStart: {
-              type: 'string',
-              description:
-                'DEPRECATED: The lowest sort-ordered rkey to start from (exclusive)',
-            },
-            rkeyEnd: {
-              type: 'string',
-              description:
-                'DEPRECATED: The highest sort-ordered rkey to stop at (exclusive)',
-            },
```
devinivy, line 1872: Were these removed from `lexicons/` in a different PR that didn't include codegen?

dholms: oh yeah looks like it - I didn't check the codegen close enough: https://github.com/bluesky-social/atproto/pull/2506

packages/pds/src/api/com/atproto/sync/subscribeRepos.ts
```diff
@@ -66,13 +66,6 @@ export default function (server: Server, ctx: AppContext) {
           time: evt.time,
           ...evt.evt,
         }
-      } else if (evt.type === 'tombstone') {
-        yield {
-          $type: '#tombstone',
-          seq: evt.seq,
-          time: evt.time,
-          ...evt.evt,
-        }
```
devinivy, line 75: What's the intended interaction between old data in the sequencer table and this new logic.  Looks like this should just ignore data that has already been stored for the deprecated event types?

packages/pds/src/sequencer/sequencer.ts
```diff
@@ -163,13 +161,6 @@ export class Sequencer extends (EventEmitter as new () => SequencerEmitter) {
           time: row.sequencedAt,
           evt: evt as AccountEvt,
         })
-      } else if (row.eventType === 'tombstone') {
-        seqEvts.push({
-          type: 'tombstone',
-          seq: row.seq,
-          time: row.sequencedAt,
-          evt: evt as TombstoneEvt,
-        })
```
devinivy, line 172: Oops kind of a dupe, but applies here too— what's the intended interaction between old data in the sequencer table and this new logic? Looks like this should just ignore data that has already been stored for the deprecated event types?

packages/pds/src/sequencer/sequencer.ts
```diff
@@ -222,8 +213,9 @@ export class Sequencer extends (EventEmitter as new () => SequencerEmitter) {
     return await this.sequenceEvt(evt)
   }
 
-  async sequenceHandleUpdate(did: string, handle: string): Promise<number> {
-    const evt = await formatSeqHandleUpdate(did, handle)
+  // @TODO
```
devinivy, line 216: What is the remaining TODO here (or can we nix this)?

packages/pds/src/sequencer/events.ts
```diff
@@ -13,41 +14,23 @@ export const formatSeqCommit = async (
...
-  if (commitData.ops.length > 200 || blocksToSend.byteSize > 1000000) {
-    const justRoot = new BlockMap()
-    const rootBlock = blocksToSend.get(commitData.cid)
-    if (rootBlock) {
-      justRoot.set(commitData.cid, rootBlock)
-    }
+  // If event is too big (max 200 ops or 2MB of data)
+  if (blocksToSend.byteSize > 2000000) {
+    throw new InvalidRequestError('Too many writes. Max event size: 2MB')
+  }
```
devinivy, line 20: The write could have already been made to the repo in this case, I believe, so this is sort of like an assertion to ensure we don't send bad data over the firehose.  I think it would be fine if this surfaced as a 500 to be honest, as this shouldn't ever happen.

Did we ever end-up enforcing a 1mb max on the record blocks within the write handlers?  I think ideally there error would come from around there, during the repo write.

packages/pds/src/sequencer/events.ts
```diff
@@ -13,41 +14,23 @@ export const formatSeqCommit = async (
...
-    }
+  const evt = {
+    repo: did,
+    commit: commitData.cid,
+    rev: commitData.rev,
+    since: commitData.since,
+    blocks: await blocksToCarFile(commitData.cid, blocksToSend),
+    ops: commitData.ops,
+    prevData: commitData.prevData ?? undefined,
+    // deprecated (but still required) fileds
```
devinivy, line 30:
```suggestion
    // deprecated (but still required) fields
```

devinivy (approved):

packages/pds/src/api/com/atproto/sync/subscribeRepos.ts
dholms: yeah exactly, it'll just skip over it

packages/pds/src/sequencer/sequencer.ts
dholms: yup same as above 👍

packages/pds/src/sequencer/sequencer.ts
dholms: oops left it straggling

packages/pds/src/sequencer/events.ts
dholms: Ah good call, I like just moving this all to the write handler

PR #3587, opened 2025-02-25
:sparkles: Add team member handle and displayName

foysalit:
This PR adds `handle` and `displayName` columns to the member table and keeps those values in sync with member profile through a new daemon that runs every 24hrs by default.

This allows searching for team members using handle/display name.


packages/ozone/src/daemon/context.ts
```diff
@@ -67,6 +70,17 @@ export class DaemonContext {
       appviewAgent,
       createAuthHeaders,
     )
+    const teamService = TeamService.creator()
+    const teamProfileSynchronizer = new TeamProfileSynchronizer(
+      backgroundQueue,
+      {
+        createAuthHeaders: (method: string) =>
```
foysalit, line 77: This reads a bit weird but basically can't come up with a better, more context aware name that would work elsewhere as well as here.

packages/ozone/src/team/index.ts
```diff
@@ -1,15 +1,22 @@
...
 import { Database } from '../db'
 import { Member } from '../db/schema/member'
 import { ids } from '../lexicon/lexicons'
 import { ProfileViewDetailed } from '../lexicon/types/app/bsky/actor/defs'
 import { Member as TeamMember } from '../lexicon/types/tools/ozone/team/defs'
 import { httpLogger } from '../logger'
+import { AuthHeaders } from '../mod-service/views'
 
 export type TeamServiceCreator = (db: Database) => TeamService
+export type AppviewCreator = {
```
foysalit, line 16: Refactoring away from passing the whole `AppContext` since we want to be able to run this from both app and daemon.

packages/ozone/src/daemon/team-profile-synchronizer.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export class TeamProfileSynchronizer extends PeriodicBackgroundTask {
+  constructor(
+    backgroundQueue: BackgroundQueue,
+    appviewCreator: AppviewCreator,
+    teamService: TeamService,
+    interval = 24 * HOUR,
+  ) {
+    super(backgroundQueue, interval, async (_, signal) => {
+      if (signal.aborted) return
```
matthieusieben, line 13: There is no need to check if the signal is aborted here. This check is only needed between tasks (if multiple statements, or during loops)

```suggestion
```

packages/ozone/src/team/index.ts
matthieusieben: Since you take the opportunity to refactor, you should change this from being a method  argument to a constructor argument:

```ts

class TeamService {
  constructor(
    public db: Database,
    protected readonly agent: AtpAgent,
    // Uses the same signature as existing "createAuthHeaders" methods
    protected readonly createAuthHeaders: (aud: string, lxm: string) => Promise<{ headers: Record<string, string> }>
  ) {}

  async getProfiles(dids: string[]) {
    // uses this.agent and this.createAuthHeaders
  }
}

```

packages/ozone/src/daemon/context.ts
devinivy: I like that pretty well too 👍

packages/ozone/src/team/index.ts
```diff
@@ -211,13 +230,48 @@ export class TeamService {
...
 
+  async syncMemberProfiles(appviewCreator: AppviewCreator): Promise<void> {
+    const { count } = await this.db.db
+      .selectFrom('member')
+      .select(this.db.db.fn.count<number>('did').as('count'))
+      .executeTakeFirstOrThrow()
+
+    let lastDid = ''
+    // Max 25 profiles can be fetched at a time so let's pull 25 members at a time from the db and update their profile details
+    for (let i = 0; i < count; i += 25) {
```
devinivy, line 241: Rather than fetching the `count` and relying on it not changing, you could alternately _only_ use the `lastDid` to iterate through.  I think it's sort of nice since you're already iterating by `lastDid`.  It would look more like:
```ts
let lastDid = ''
do {
  // ...
  lastDid = lastDid = dids.at(-1) || ''
} while (lastDid)
```

packages/ozone/src/api/team/addMember.ts
```diff
@@ -26,14 +26,23 @@ export default function (server: Server, ctx: AppContext) {
           )
         }
 
+        const appviewCreator = {
+          agent: ctx.appviewAgent,
+          createAuthHeaders: ctx.appviewAuth,
+        }
+        const profiles = await teamService.getProfiles([did], appviewCreator)
```
devinivy, line 33: It would be kind of nice if adding a member would still complete even if fresh profile information couldn't be fetched.

packages/ozone/tests/team.test.ts
```diff
@@ -11,6 +11,9 @@ describe('team management', () => {
   beforeAll(async () => {
     network = await TestNetwork.create({
       dbPostgresSchema: 'ozone_team_test',
+      ozone: {
+        dbTeamProfileRefreshIntervalMs: 1000,
```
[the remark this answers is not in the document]
devinivy: Since the information is fetched when the member is added, I don't think the refresh interval is ever really used (or perhaps I missed something).  Could we disable it altogether?  (Also—isn't it only used by the daemon and not ozone proper?)

packages/ozone/src/team/index.ts
matthieusieben: (this is the same comment as above)

packages/ozone/src/db/migrations/20250221T132135150Z-member-details.ts
```diff
@@ -0,0 +1,24 @@
...
+  await db.schema
+    .createIndex('member_display_name_idx')
+    .on('member')
+    .column('displayName')
+    .execute()
+  await db.schema
+    .createIndex('member_handle_idx')
+    .on('member')
+    .column('handle')
+    .execute()
```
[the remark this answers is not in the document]
foysalit: oh right. I'm gonna go with trgm, i think i saw an example of that somewhere else.

packages/ozone/src/api/team/addMember.ts
foysalit: yeah `teamService.getProfiles` handles failures internally and won't throw error if fetching profile fails.

packages/ozone/tests/team.test.ts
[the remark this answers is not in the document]
foysalit: i believe the daemon processes are run when `processAll()` is called in test network and for this test, we are creating the members with `handle` and `displayName` null to kinda implicitly test that the background sync works.

devinivy (approved):

PR #3605, opened 2025-03-06
:bug: Add missing union ref to ozone mod subject

foysalit:


devinivy (approved):
Looks good, probably just want to add a changeset for @atproto/api.

PR #3619, opened 2025-03-08
fix: allow handle format with tld shorter than 3 characters

Johannes-Andersen:
When trying to sign up in the oAuth flow with a tld or domain shorter than 3 char, you end up getting blocked by zod validation against the post body.

Example of failing ones:
"hello.norwegian.no"
"hello.vg.com"

Ref implementation done in https://github.com/bluesky-social/atproto/pull/2945

### DIFF
--- a/packages/oauth/oauth-provider/src/account/account-store.ts
+++ b/packages/oauth/oauth-provider/src/account/account-store.ts
@@ -19,9 +19,18 @@ export const newPasswordSchema = z.string().min(8)
 export const tokenSchema = z.string().regex(/^[A-Z2-7]{5}-[A-Z2-7]{5}$/)
 export const handleSchema = z
   .string()
-  .min(3)
-  .max(30)
-  .regex(/^[a-z0-9][a-z0-9-]+[a-z0-9](?:\.[a-z0-9][a-z0-9-]+[a-z0-9])+$/)
+  .refine((value) => value.split('.').length >= 3, {
+    message: 'Handle must be on a subdomain (handle.domain.tld)',
+  })
+  .refine((value) => value.split('.')[0].length >= 3, {
+    message: 'Handle must be at least 3 characters long',
+  })
+  .refine((value) => value.split('.')[0].length <= 18, {
+    message: 'Handle must be at most 18 characters long',
+  })
+  .refine((value) => /^[a-z0-9]+$/.test(value.split('.')[0]), {
+    message: 'Handle must only contain lowercase letters and numbers',
+  })
 export const emailSchema = z
   .string()
   .email()
--- a/.changeset/dirty-snakes-wave.md
+++ b/.changeset/dirty-snakes-wave.md
@@ -1,0 +1,5 @@
+---
+"@atproto/oauth-provider": patch
+---
+
+Allow handle format with tld shorter than 3 characters


packages/oauth/oauth-provider/src/account/account-store.ts
```diff
@@ -19,9 +19,18 @@ export const newPasswordSchema = z.string().min(8)
...
-  .max(30)
-  .regex(/^[a-z0-9][a-z0-9-]+[a-z0-9](?:\.[a-z0-9][a-z0-9-]+[a-z0-9])+$/)
+  .refine((value) => value.split('.').length >= 3, {
+    message: 'Handle must be on a subdomain (handle.domain.tld)',
+  })
+  .refine((value) => value.split('.')[0].length >= 3, {
+    message: 'Handle must be at least 3 characters long',
+  })
+  .refine((value) => value.split('.')[0].length <= 18, {
+    message: 'Handle must be at most 18 characters long',
```
Johannes-Andersen, line 29: Copied length check from https://github.com/Johannes-Andersen/atproto/blob/2759ef1a0043134d5bc228c431fd276bbe76c4a3/packages/pds/src/handle/index.ts#L43-L44

packages/oauth/oauth-provider/src/account/account-store.ts
```diff
@@ -19,9 +19,18 @@ export const newPasswordSchema = z.string().min(8)
...
+  })
+  .refine((value) => /^[a-z0-9]+$/.test(value.split('.')[0]), {
+    message: 'Handle must only contain lowercase letters and numbers',
+  })
```
Johannes-Andersen, line 33: Alternative for regex:
```suggestion
.regex(/^[a-z0-9]{3,18}\.[a-z0-9-]+\.[a-z]{2,}$/)
```

packages/oauth/oauth-provider/src/account/account-store.ts
```diff
@@ -19,9 +19,18 @@ export const newPasswordSchema = z.string().min(8)
...
+    message: 'Handle must be on a subdomain (handle.domain.tld)',
+  })
+  .refine((value) => value.split('.')[0].length >= 3, {
+    message: 'Handle must be at least 3 characters long',
+  })
+  .refine((value) => value.split('.')[0].length <= 18, {
+    message: 'Handle must be at most 18 characters long',
+  })
+  .refine((value) => /^[a-z0-9]+$/.test(value.split('.')[0]), {
+    message: 'Handle must only contain lowercase letters and numbers',
```
matthieusieben, line 32: Dashes are allowed too (but not in first or last position)

matthieusieben:
Thank you for your contribution. This looks almost right 👍

bnewbold:
would it be clearer to validate the overall handle syntax using the `@atproto/syntax` library, and then check the additional "name" constraints after that? Instead of having checks for digits, dashes, etc duplicated here.

Johannes-Andersen:
> would it be clearer to validate the overall handle syntax using the `@atproto/syntax` library, and then check the additional "name" constraints after that? Instead of having checks for digits, dashes, etc duplicated here.

I agree!
I’ll have a go at it later in the week to see if I can get something up for review.
Just need to figure out how to properly get it all running locally and wrap my head around the codebase and packages 😄

Johannes-Andersen:
Thank you Matthieu! Seems to be getting close to a working oAuth version now 🙌🏻 
Only issue noticed in on-boarding flow at the moment is hCaptcha failing to load. Created an issue for it :D 

https://github.com/bluesky-social/atproto/issues/3625

PR #3620, opened 2025-03-09
did: Allow Ed25519

lorenzleutgeb:


lorenzleutgeb:
@foysalit, kindly asking you to approve workflows for me. Thanks!

bnewbold:
This is something we are open to considering over time, but it should start as a protocol conversation before implementation work.

Ed25519 is a popular system, and we considered it when selecting the original 2x cryptographic systems. If we are going to add the complexity of a third system, which every implementation of atproto would be expected to adopt, we would want to make sure there are clear and unique/novel advantages to that additional system, however.

lorenzleutgeb:
Please also note my other, linked PR which restricts rotation keys and verification methods for `"atproto"` to P-256 or secp256k1, even if `@atproto/crypto` would support more `did:key` algorithms.

The reason for allowing Ed25519 is because I would like to link my `did:plc` with a verification method that simply does not support either P-256 nor secp256k1, but does support Ed25519.

Just adding an Ed25519 verification method with ID `"radicle"` would enable interoperability with radicle.xyz. That is, it would allow `did:plc` to be used not only in an ATProtocol context but also in a Radicle context.

From what I understand, such openness is a goal of `did:plc`.

bnewbold:
Apologies for the long reply time on this!

I assumed you were pushing to include support for Ed25519 more broadly through atproto, for use in signing commit messages and things like that.

Instead it sounds like you simply want to be able to include Ed25519 public keys in `did:plc` DID documents, to use with protocols other than atproto.

That sounds great! We definitely want folks to be able to do this sort of thing. `did:plc` should be relatively agnostic to applications and use-cases.

Just to double-check, are you certain that a patch like what you are proposing is needed? My assumption would be that things "just work" and you could add additional pubkeys as encoded strings, without touching the `@atproto/crypto` package. I'm not as familiar with the TypeScript codebase and could be wrong. If the package doesn't support this usecase today, what we might want to do is update it to support arbitrary key verification methods as strings, instead of adding full cryptographic system support.

lorenzleutgeb:
> Apologies for the long reply time on this!

Don't worry. Of course, anyone prefers their feature requests worked on sooner rather than later, but I am aware that my suggestion is "from the outside"/"drive by", and you have planned to work on other things for which you see a higher priority or urgency. Thank you for your consideration!

> I assumed you were pushing to include support for Ed25519 more broadly through atproto, for use in signing commit messages and things like that.
> Instead it sounds like you simply want to be able to include Ed25519 public keys in `did:plc` DID documents, to use with protocols other than atproto.

Exactly. Even though I would like to be able to use Ed25519 more broadly, I do realize that this would have much further reaching consequences, so with this PR and with https://github.com/did-method-plc/did-method-plc/pull/93 I am trying to get Ed25519 in "minimally".

> Just to double-check, are you certain that a patch like what you are proposing is needed? My assumption would be that things "just work" and you could add additional pubkeys as encoded strings, without touching the `@atproto/crypto` package. I'm not as familiar with the TypeScript codebase and could be wrong. If the package doesn't support this usecase today, what we might want to do is update it to support arbitrary key verification methods as strings, instead of adding full cryptographic system support.

As of today, from what I understand, there is no separation between "full cryptographic" and "encoded string" support. My requests hinge on the fact that the following check would fail without a plugin for Ed25519:

https://github.com/bluesky-social/atproto/blob/c777ba6d68d4be0df2efffffb957f46ce1e032e5/packages/crypto/src/did.ts#L13-L16

To be honest, I did not consider splitting support into "full cryptograpic" and "encoded string", for me it was natural to just add Ed25519 at the same level of support by `@atproto/crypto` as the other two curves. This would also make it easier to support Ed25519 in more use-cases in the future, if so desired. Then, my idea was to restrict usage of these curves at higher layers, which is how https://github.com/did-method-plc/did-method-plc/pull/93 came to be.

I am willing to work on this PR, introducing such separation, i.e., thinking of how an "encoded string" fallback could look like, if there is no plugin for a `did:key`. However, since I already did the work for an Ed25519 plugin, I would like to suggest to keep it, and add "encoded string" capabilities later in another PR.

tesaguri:
Seconding this!

I'd love to experiment with PLC using my hardware security key, but I don't want to replace my existing Ed25519 keys just for it.

lorenzleutgeb:
@bnewbold what are the chances for moving this forward in the next weeks? Could you please approve the GitHub Actions workflow, and consider reviewing? Happy to also do this synchronously with you on a call.

PR #3626, opened 2025-03-10
feat(oauth): add Pwned Passwords check

Johannes-Andersen:
Integrate https://haveibeenpwned.com/API/v3#PwnedPasswords breached/compromised password check in oAuth account registry flow.

Should help with ATO attacks by enforcing better security hygiene.

Screenshot 2025-03-11 at 00 26 54

### DIFF
--- a/.changeset/great-seals-act.md
+++ b/.changeset/great-seals-act.md
@@ -1,0 +1,8 @@
+---
+"@atproto/oauth-provider-ui": minor
+"@atproto/oauth-provider": minor
+"@atproto/dev-env": patch
+"@atproto/pds": patch
+---
+
+Add compromised passwords check to oAuth signup
--- a/packages/dev-env/src/pds.ts
+++ b/packages/dev-env/src/pds.ts
@@ -55,6 +55,7 @@ export class TestPds {
       // colors are required from a config perspective.
       warningColor: undefined,
       successColor: '#02c39a',
+      enableHibpCheck: true,
       logoUrl:
         // Using a "data:" instead of a real URL to avoid making CORS requests in dev.
         // License: https://uxwing.com/license/
--- a/packages/oauth/oauth-provider-ui/src/components/utils/error-message.tsx
+++ b/packages/oauth/oauth-provider-ui/src/components/utils/error-message.tsx
@@ -2,6 +2,7 @@ import { Trans } from '@lingui/react/macro'
 import { ReactNode, memo } from 'react'
 import {
   AccessDeniedError,
+  CompromisedPasswordError,
   EmailTakenError,
   HandleUnavailableError,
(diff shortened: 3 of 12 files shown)


tom-sherman:
Is a high-entropy pwned password more secure than a low-entropy unique one? What about when the password is salted?

My concern with this is that it risks regressing the average security of new accounts, especially if the check doesn't include email. There's obviously privacy concerns with adding hibp to the PDS like this, but even more so if the check sends email+password to the hipb API (not even sure if the API supports this actually).

I think trying to increase the security of new accounts is a noble and valid goal, but I think a better option would be to adopt passkeys and add 2FA compared to trying (and perhaps failing) to increase the security of passwords.

Johannes-Andersen:
> Is a high-entropy pwned password more secure than a low-entropy unique one? What about when the password is salted?
> 
> My concern with this is that it risks regressing the average security of new accounts, especially if the check doesn't include email. There's obviously privacy concerns with adding hibp to the PDS like this, but even more so if the check sends email+password to the hipb API (not even sure if the API supports this actually).
> 
> I think trying to increase the security of new accounts is a noble and valid goal, but I think a better option would be to adopt passkeys and add 2FA compared to trying (and perhaps failing) to increase the security of passwords.

Valid questions and concerns. One way people have implemented this, and I debating adding it as well, is a limit on how many breaches before it should trigger. 
So it a password has been seen 5, 10 or 100 times it will be marked as compromised and rejected. 

Any password that has been breached and is available as a plain text password in hibp would be a bad one, no matter how much the entropy. Password lists are generated and ran thru in no time. 

« sends email+password to the hipb» it is worth adding we are never sending the password nor email. I recommend reading https://www.troyhunt.com/understanding-have-i-been-pwneds-use-of-sha-1-and-k-anonymity/ and https://blog.cloudflare.com/validating-leaked-passwords-with-k-anonymity/

I also agree 2fa and passkeys are the way to go, but unfortunately passwords will be around for some time, and people love reusing passwords added on all kinds of password lists 🥲

Johannes-Andersen:
«Fun» thing we noticed at my old job is that suddenly we had a massive influx of reports of accounts getting hacked.

After investigating, it was found that all of them where in a specific external data breach when we looked up their email on hibp.

Someone had setup a botnet to attempt the email and password combination for all those accounts on our customers site of that unrelated breach. 

We fixed it by emailing all users who was our clients customer, and on the breached list, asking them to change passwords due to the known trend. And on login, if we detected the password was bad, lock their account and ask them to reset their password via an email link.

Looking at the graphs, we could see passwords being attempted by the hotbed only breached less than 8 times being the most impactful ones (high value accounts) (1-10 or rounded to nearest 10 to prevent tracking)

PR #3632, opened 2025-03-11
@atproto/lex-cli: add support for JSONC and JSONL

strass:
I would like to add support for importing additional types of JSON file, [json with comments](https://github.com/komkom/jsonc) and [json lines](https://jsonlines.org/).

Use for JSONC: I want to be able to make inline notes in my schema files
Use for JSONL: convenient way to group multiple related schema entries

I'm not yet familiar with the dev environment for the packages so I have not yet tested the PR. Before I do testing I want to make sure this is a change you'd accept

### DIFF
--- a/packages/lex-cli/src/util.ts
+++ b/packages/lex-cli/src/util.ts
@@ -5,10 +5,16 @@ import { ZodError, type ZodFormattedError } from 'zod'
 import { type LexiconDoc, parseLexiconDoc } from '@atproto/lexicon'
 import { type FileDiff, type GeneratedAPI } from './types'
 
+const isJsonLike = (path: string) => {
+  return (
+    path.endsWith('.json') || path.endsWith('.jsonc') || path.endsWith('.jsonl')
+  )
+}
+
 export function readAllLexicons(paths: string[]): LexiconDoc[] {
   const docs: LexiconDoc[] = []
   for (const path of paths) {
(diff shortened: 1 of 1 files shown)


packages/lex-cli/src/util.ts
```diff
@@ -20,17 +26,33 @@ export function readAllLexicons(paths: string[]): LexiconDoc[] {
...
+): Record<string, unknown> | Record<string, unknown>[] => {
+  if (path.endsWith('.json')) {
+    return JSON.parse(str)
+  }
+  if (path.endsWith('.jsonc')) {
+    return jsonc.parse(str)
+  }
+  if (path.endsWith('.jsonl')) {
+    return jsonl.parse(str)
+  }
```
strass, line 41: TODO: need to pick parsers, does the org have any licensing requirements for dependencies?

packages/lex-cli/src/util.ts
```diff
@@ -41,6 +63,9 @@ export function readLexicon(path: string): LexiconDoc {
     typeof (obj as LexiconDoc).lexicon === 'number'
   ) {
     try {
+      if (Array.isArray(obj)) {
+        return obj.map((o) => parseLexiconDoc(o))
+      }
```
strass, line 68: TODO: have parent deal with possibility of arrays

PR #3645, opened 2025-03-17
OAuth: Allow `onSignedUp` hook to access hcaptcha result data

matthieusieben:
This change reworks the way sign-up hooks are triggered:

- The `onSignupAttempt` is now triggered before hcaptcha verification. The reason for an earlier check is to prevent sing-ups based, for example, on ip addresses.
- The `onSignedUp` hook and store's `createAccount` method now receives the hcaptcha verification result, allowing them to store hcaptcha results
- Remove `onSignupHcaptchaResult` hook

This change also fixes hcaptcha score validation

### DIFF
--- a/.changeset/slimy-flowers-punch.md
+++ b/.changeset/slimy-flowers-punch.md
@@ -2,4 +2,4 @@
 "@atproto/oauth-provider": patch
 ---
 
-Improve HCaptocha error reporting
+Improve HCaptcha error reporting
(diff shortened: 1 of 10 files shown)


.changeset/slimy-flowers-punch.md
```diff
@@ -0,0 +1,5 @@
+---
+"@atproto/oauth-provider": patch
+---
+
+Improve HCaptocha error reporting
```
devinivy, line 5:
```suggestion
Improve HCaptcha error reporting
```

devinivy (approved):

PR #3649, opened 2025-03-18
api: add some metrics to topics response

haileyok:


mozzius (approved):
🔥

lexicons/app/bsky/unspecced/defs.json
```diff
@@ -30,7 +30,11 @@
         "topic": { "type": "string" },
         "displayName": { "type": "string" },
         "description": { "type": "string" },
-        "link": { "type": "string" }
+        "link": { "type": "string" },
+        "startTime": { "type": "string", "format": "datetime" },
+        "postCount": { "type": "integer" },
+        "category": { "type": "string" },
+        "images": { "type": "array", "items": { "type": "string" } }
```
haileyok, line 37: this has a generic `images` name, because we may end up sticking other types of images in here (i.e. for starter packs in the suggestions response). the server will remove images of profiles that are labeled, taken down, etc. at an interval of 30 seconds.

with that said, the better option might be to have the TT engine send down a list of dids which the appview can then hydrate w/ viewer info (so that labelers/blocks/etc. will also work). i'll get this going, though in the meantime for client work we'll send down just the raw URLs.

PR #3650, opened 2025-03-18
Update profile.json

seboslaw:
Added the optional properties:

- pronouns
- location
- birthday
- customLinks (max. 4)
- tipLink


lexicons/app/bsky/actor/profile.json
```diff
@@ -45,6 +45,46 @@
             "type": "ref",
             "ref": "com.atproto.repo.strongRef"
           },
+          "pronouns": {
```
ThisIsMissEm, line 48: Pronouns has been added in https://github.com/bluesky-social/atproto/pull/4232

PR #3651, opened 2025-03-19
:sparkles: Add getSubjects endpoint to ozone for full view of subjects

foysalit:


lexicons/tools/ozone/moderation/defs.json
```diff
@@ -203,6 +203,36 @@
...
+      "type": "object",
+      "required": ["type", "subject"],
+      "properties": {
+        "type": {
+          "type": "ref",
+          "ref": "com.atproto.moderation.defs#subjectType"
+        },
+        "subject": {
+          "type": "string"
+        },
```
devinivy, line 217: I assume that `subject` is intended to contain the DID or AT-URI, is that right?

Would it be desirable to get tighter typings by doing what we do elsewhere, making this a union of a repo ref or record/strong ref?  Don't feel too strongly, this just looked like a new pattern to me and it does have the side effect of making the types a bit looser.

packages/ozone/src/api/moderation/getSubjects.ts
```diff
@@ -0,0 +1,91 @@
...
+
+      for (const [subject, status] of missingSubjectStatuses) {
+        const subjectView = subjectWithDetails.get(subject)
+        if (subjectView)
+          subjectView.status = modViews.formatSubjectStatus(status)
+      }
+
+      return {
+        encoding: 'application/json',
+        body: { subjects: Array.from(subjectWithDetails.values()) },
```
devinivy, line 87: It would be a nice property for the order of the response to preserve the order of the params.

lexicons/tools/ozone/moderation/defs.json
```diff
@@ -203,6 +203,36 @@
...
+          "ref": "#subjectStatusView"
+        },
+        "repo": {
+          "type": "ref",
+          "ref": "#repoViewDetail"
+        },
+        "profile": {
+          "type": "ref",
+          "ref": "app.bsky.actor.defs#profileViewDetailed"
+        },
```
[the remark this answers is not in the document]
foysalit: hmmm... interesting. hadn't thought of that tbh. this is what we have been doing on the clients already and moving this into the lexicons just give it a tightly integrated feeling i guess and i think the bsky profiles carry really strong signals so ozone integrating with that feels useful to me.

lexicons/tools/ozone/moderation/defs.json
foysalit: > I assume that subject is intended to contain the DID or AT-URI, is that right?

yep. 

it is a different pattern than what we have been using but it's just soo much easier to pass around and handle compared to repo ref or strong ref.

lexicons/tools/ozone/moderation/defs.json
```diff
@@ -203,6 +203,36 @@
...
+          "type": "ref",
+          "ref": "com.atproto.moderation.defs#subjectType"
+        },
+        "subject": {
+          "type": "string"
+        },
+        "status": {
+          "type": "ref",
+          "ref": "#subjectStatusView"
+        },
```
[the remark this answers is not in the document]
foysalit: im kinda intentionally duplicating this because 
1. in cases where repo or record may have been deleted but we have subject status, this will be the carrier of that status
2. in cases where duplicate is found in repo or record, it's just much easier to always know that if status exists, it will be available here without traversing `record.moderation.subjectStatus` - this is just a nice to have tho and #1 is more crucial

lexicons/tools/ozone/moderation/getSubjects.json
```diff
@@ -0,0 +1,40 @@
...
+      "type": "query",
+      "description": "Get details about subjects.",
+      "parameters": {
+        "type": "params",
+        "required": ["subjects"],
+        "properties": {
+          "subjects": {
+            "type": "array",
+            "maxLength": 100,
+            "minLength": 1,
```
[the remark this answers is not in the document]
devinivy: Depends what you mean by "make sense", ha!  If it allowed an empty list then callers could just call this with any array of subjects, without having to do a special check to ensure there is at least one subject.  I think it's a nice simplification to be a little permissive.  Many callers might perform the check as a way to avoid an unnecessary roundtrip.  Just my take, all cool either way.

I actually just realized that `subjects` is required which also enforces the `minLength` anyway, and apparently other methods like `app.bsky.actor.getProfiles` has the same behavior.  So I am cool with this as-is.

packages/ozone/tests/__snapshots__/get-subjects.test.ts.snap
```diff
@@ -0,0 +1,445 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`admin get multiple subjects with all relevant details gets multiple subjects with records 1`] = `
+Array [
+  Object {
+    "profile": Object {
+      "$type": "app.bsky.actor.defs#profileViewDetailed",
+    },
```
devinivy, line 8: This doesn't look quite right—I think `profile` should be omitted in this case.

packages/ozone/src/api/moderation/getSubjects.ts
```diff
@@ -0,0 +1,101 @@
...
+            : repo?.moderation.subjectStatus
+
+        subjectWithDetails.set(subject, {
+          type,
+          repo,
+          record,
+          profile: {
+            $type: 'app.bsky.actor.defs#profileViewDetailed',
+            ...profile,
+          },
```
devinivy, line 69: I think we want something along these lines since `profile` can be missing.
```suggestion
          profile: profile && {
            $type: 'app.bsky.actor.defs#profileViewDetailed',
            ...profile,
          },
```

devinivy (approved):
One final small bug to address (see comments), otherwise looking good!

packages/ozone/tests/__snapshots__/get-subjects.test.ts.snap
foysalit: yep yep!

PR #3659, opened 2025-03-21
Account management page

matthieusieben:
This PR introduces a private API for the authorization server that enables account management. These API endpoints are documented in the new `@atproto/oauth-provider-api` package (shared lib for backend and frontend).

The new routes are implemented in [packages/oauth/oauth-provider/src/router/api-router.ts](https://github.com/bluesky-social/atproto/blob/msi/oauth-account-api/packages/oauth/oauth-provider/src/router/api-router.ts).

The API routes were previously implemented from the `OAuthProvider` class. That file was getting too big (> 2k LoC) so the middleware code was split in three parts:

- [api-router.ts](https://github.com/bluesky-social/atproto/blob/msi/oauth-account-api/packages/oauth/oauth-provider/src/router/api-router.ts) - API endpoints
- [oauth-router.ts](https://github.com/bluesky-social/atproto/blob/msi/oauth-account-api/packages/oauth/oauth-provider/src/router/oauth-router.ts) - OAuth endpoints
- [authorize-router.ts](https://github.com/bluesky-social/atproto/blob/msi/oauth-account-api/packages/oauth/oauth-provider/src/router/authorize-page/authorize-router.ts) - Authorization page endpoints

Only the code from the API endpoints was significantly changed (and dependencies: `TokenManager`, `AccountManager`). The OAuth Router (standard OAuth endpoints) and Authorize Router (serving of authorization page & redirects to client) were mostly moved. The authorization coded does contain some other changes linked to the way frontend assets are interacted with.

## Common types package

Types that are common to the front and backend were moved into `@atproto/oauth-provider-api` to ensure proper coordination between both codebases.

## API routes

API endpoints can be triggered from two different contexts:
- From an authorization flow (a `RequestURI` is available)
- From the account management page (no `RequestURI`)

## Device Account

User's "Device" (browser) and account are linked in a 1-1 relationship, allowing to authenticate users based on their cookies.

```
 ┌─────────────┐    ┌────────────────────┐    ┌────────┐ 
 │Device       │1  n│DeviceAccount       │n  1│Account │ 
 │             ├────┤                    ├────┤        │ 
 │- ip         │    │                    │    │- did   │ 
 │- user-agent │    │                    │    │        │ 
 └─────────────┘    └────────────────────┘    └────────┘ 
```

When a user chooses not to use the "remember me" feature, the Device-Account  entry is not created. Instead an ephemeral token is created allowing to temporarily authenticate the requests.

These ephemeral tokens are bound to a particular request (if any) preventing their re-use in other contexts.

## Token mode

All tokens are now wrapped inside JWTs. The reason for this is that before, the `TokenId` string was used directly as access-token. However, these `TokenId` are now exposed through the account management API.

In order to avoid exposing valid access-tokens to the user, the `TokenId` are now useless unless wrapped inside a JWT.

The use of stateful, non JWT, `TokenId` was motivated by two reasons:
- no need to verify a signature if a database lookup can be made.
- a random string is significantly shorter than a JWT

The first benefit was abandoned with this change. This will avoid having to perform a DB lookup when the token is expired, at the expense of performing more computation, which is an acceptable trade-off IMO.

The ability to generate shorter tokens was preserved though the `AccessTokenMode.light` option. That option


packages/oauth/oauth-provider-ui/src/backend-data.ts
```diff
@@ -1,35 +0,0 @@
-import type {
```
matthieusieben, line 1: This file was no longer used.

packages/oauth/oauth-provider/src/access-token/access-token-type.ts
```diff
@@ -1,5 +0,0 @@
-export enum AccessTokenType {
```
matthieusieben, line 1: All tokens are now JWT. see `access-token-mode.ts`

packages/oauth/oauth-provider/src/access-token/access-token-mode.ts
```diff
@@ -0,0 +1,4 @@
+export enum AccessTokenMode {
```
matthieusieben, line 1: All tokens are now wrapped inside JWTs. The reason for this is that before, the `TokenId` string was used directly as access-token. However, these `TokenId` are now exposed through the account management API.

In order to avoid exposing valid access-tokens to the user, the `TokenId` are now useless unless wrapped inside a JWT.

packages/oauth/oauth-provider/src/lib/http/request.ts
```diff
@@ -106,77 +103,44 @@ export function validateReferer(
...
-) {
-  const csrfToken = randomBytes(8).toString('hex')
-  appendHeader(
-    res,
-    'Set-Cookie',
-    serializeCookie(cookieName, csrfToken, {
-      secure: true,
-      httpOnly: false,
-      sameSite: 'lax',
-      path: req.url?.split('?', 1)[0] || '/',
```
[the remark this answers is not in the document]
matthieusieben: CSRF cookie needs to be accessible from the JS to serve their purpose.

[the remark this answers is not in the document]
matthieusieben: https://github.com/bluesky-social/atproto/blob/2113c59985ba4ab5152477b7634159a384930e43/packages/oauth/oauth-provider/src/router/csrf.ts#L11

matthieusieben:
## Csrf changes

Cookies now need to be valid on two different pages (Authorization page and account management).

Because these pages are on different paths (and don't share a common root path), every cookie now has to be set with a path of `/`. The CSRF cookie mechanism was simplified to use only a single value instead of one per request URI.

packages/oauth/oauth-provider/src/lib/util/ui8.ts
```diff
@@ -0,0 +1,14 @@
+export function parseUi8Hex(v: string) {
+  return asUi8(parseInt(v, 16))
+}
+
+export function parseUi8Dec(v: string) {
+  return asUi8(parseInt(v, 10))
+}
+
+export function asUi8(v: number) {
+  if (v >= 0 && v <= 255 && v === (v | 0)) return v
```
devinivy, line 10: Is `v === (v | 0)` used instead of `Number.isInteger(v)` as an optimization?

packages/pds/tests/db.test.ts
```diff
@@ -1,5 +1,12 @@
 import { TestNetworkNoAppView } from '@atproto/dev-env'
-import { AccountDb } from '../src/account-manager/db'
+
+// @NOTE Not using an import from "../src" because this results in types from
+// "src" and "dst" being mixed (because of the circular dependency between
+// dev-end and pds). That mixing of types is usually not a problem (as they are
+// considered as two different implementation of the same interfaces by TS),
+// except when `unique symbol` are used, which is the case for
+// `jsonEncodedType` used by DB definitions.
```
devinivy, line 8: In that case this is a bit of a sharp edge around `jsonEncodedType`, I could imagine this being difficult to catch if you bump into it.  Is a unique symbol needed as opposed to a global one?

packages/oauth/oauth-provider/src/lib/http/request.ts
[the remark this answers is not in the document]
devinivy: Ah got it—for some reason I assumed it was coming down in the response body.  Pulling it out of the cookie works too 👍

packages/oauth/oauth-provider/src/lib/util/ui8.ts
matthieusieben: This was an optimization of `Nunber.floor(v) === v`, yeah. I didn't know about `Number.isInteger`. Gonna use that instead.

packages/pds/tests/db.test.ts
matthieusieben: A `unique symbol` is basically the same as doing `typeof someSymbolVariable`. The issue here is that we mix symbols from `src` and `dist`, which doesn't work whether they are declared as `unique symbol` or as a variable.

packages/oauth/oauth-provider/src/router/csrf.ts
```diff
@@ -0,0 +1,65 @@
...
+  path: `/`,
+}
+
+export async function setupCsrfToken(
+  req: IncomingMessage,
+  res: ServerResponse,
+): Promise<string> {
+  const token =
+    getCookieCsrf(req) ||
+    (await randomBuffer(TOKEN_BYTE_LENGTH)).toString('hex')
```
devinivy, line 30: I think this can be replaced by `randomHexId(TOKEN_BYTE_LENGTH)`.

packages/oauth/oauth-provider/src/types/email.ts
```diff
@@ -0,0 +1,17 @@
+import { isEmailValid } from '@hapi/address'
+import { isDisposableEmail } from 'disposable-email-domains-js'
+import { z } from 'zod'
+
+export const emailSchema = z
+  .string()
+  .email()
+  // @NOTE using @hapi/address here, in addition to the email() check to ensure
```
devinivy, line 8: Is there a reason not to only use @hapi/address?

packages/oauth/oauth-provider/src/types/email-otp.ts
```diff
@@ -0,0 +1,5 @@
+import { z } from 'zod'
+
+export const emailOtpSchema = z
+  .string()
+  .regex(/^[A-Z2-7]{5}-[A-Z2-7]{5}$/, 'Invalid token format')
```
devinivy, line 5: Does this require compatibility with the pds's current email token format?

packages/oauth/oauth-provider/src/token/token-manager.ts
```diff
@@ -33,59 +32,72 @@ import { RequestMetadata } from '../lib/http/request.js'
...
+      iat: dateToEpoch(options.now),
+      cnf: parameters.dpop_jkt ? { jkt: parameters.dpop_jkt } : undefined,
 
-    return this.accessTokenType === AccessTokenType.jwt
+      ...(this.accessTokenMode !== AccessTokenMode.stateless && {
+        aud: account.aud,
+        scope: parameters.scope,
+        // https://datatracker.ietf.org/doc/html/rfc8693#section-4.3
+        client_id: client.id,
+      }),
```
devinivy, line 91: I would expect these to be part of the stateless token but not the light token.

packages/oauth/oauth-provider/src/token/token-manager.ts
```diff
@@ -33,59 +32,72 @@ import { RequestMetadata } from '../lib/http/request.js'
...
+    account: Account,
+    client: Client,
+    parameters: OAuthAuthorizationRequestParameters,
+    options: {
+      now: Date
+      expiresAt: Date
+    },
+  ): Promise<OAuthAccessToken> {
+    return this.signer.createAccessToken({
+      sid: tokenId,
```
devinivy, line 80: I take it that over time many JWTs are created with the same token id, which is why this is `sid` and not `jti`—sound right?

devinivy: (I.e. this is not suitable for replay detection.)

packages/oauth/oauth-provider/src/token/token-manager.ts
```diff
@@ -477,41 +430,65 @@ export class TokenManager {
...
+    return data.updatedAt.getTime() + inactivityTimeout < Date.now()
+  }
+
   /**
    * @see {@link https://datatracker.ietf.org/doc/html/rfc7009#section-2.2 | RFC7009 Section 2.2}
    */
-  async revoke(token: string): Promise<void> {
+  async revoke(token: string): Promise<undefined | TokenInfo> {
     switch (true) {
       case isTokenId(token): {
```
devinivy, line 460: Not directly related to this work, but the `switch (true)` is a little bit of an odd construction.  No love for the good ole `if`s 😆 ?

packages/oauth/oauth-provider/src/token/token-manager.ts
```diff
@@ -546,9 +523,6 @@ export class TokenManager {
 
   protected async findTokenInfo(token: string): Promise<TokenInfo | null> {
     switch (true) {
-      case isTokenId(token):
-        return this.store.readToken(token)
-
```
devinivy, line 551: This case looks correct at a glance, and it makes the method comprehensive.  Was this removed because there's no code path that hits it?  It is just a little hard to tell whether there might be such a codepath when the types can't back it up in this case.

packages/oauth/oauth-provider/src/token/token-manager.ts
```diff
@@ -581,54 +555,72 @@ export class TokenManager {
...
-    const claims: TokenClaims = {
-      aud: tokenInfo.account.aud,
-      sub: tokenInfo.account.sub,
-      exp: dateToEpoch(tokenInfo.data.expiresAt),
-      iat: dateToEpoch(tokenInfo.data.updatedAt),
-      scope: tokenInfo.data.parameters.scope,
-      client_id: tokenInfo.data.clientId,
+    const claims: SignedTokenPayload = {
+      iss: this.signer.issuer,
+      jti: tokenId,
```
devinivy, line 593: Same here—I do not see a `jti` claim on the access tokens.

packages/oauth/oauth-provider/src/token/token-manager.ts
devinivy, line 80: I think `SignedTokenPayload` wants a `jti` but I don't see one passed here—any idea why we're not getting a type error?

packages/oauth/oauth-provider/src/oauth-provider.ts
```diff
@@ -357,20 +341,36 @@ export class OAuthProvider extends OAuthVerifier {
     return this.keyset.publicJwks
   }
 
-  protected loginRequired(
-    client: Client,
+  /**
+   * @returns true is the user's consent is required for the requested scopes
```
devinivy, line 345:
```suggestion
   * @returns true if the user's consent is required for the requested scopes
```

packages/oauth/oauth-provider/src/router/ui-router/assets.ts
```diff
@@ -0,0 +1,34 @@
+import type { HydrationData as FeHydrationData } from '@atproto/oauth-provider-frontend/hydration-data'
+import type { HydrationData as UiHydrationData } from '@atproto/oauth-provider-ui/hydration-data'
+import { combineMiddlewares } from '../../lib/http/middleware.js'
+import { Simplify } from '../../lib/util/type.js'
+import { parseAssetsManifest } from './assets-manifest.js'
+
+// If the "ui" and "frontent" packages are ever unified, this can be replaced
```
devinivy, line 7:
```suggestion
// If the "ui" and "frontend" packages are ever unified, this can be replaced
```

packages/oauth/oauth-provider/src/token/token-manager.ts
matthieusieben: right !

packages/oauth/oauth-provider/src/token/token-manager.ts
matthieusieben: `tokenId`s are actually rotated.

Your comment is the reason why I switched to using `sid` instead at some point. But that had other side effects (e.g. w.r.t backwards compatibility) that I thought was not ideal.

The best solution here would probably be to use `tokenId` as a static identifier, and use another identifier for the "current token identifier" (`jti`).

This solution, however, was more complex to put in place (requires DB migrations, and change in code logic).

matthieusieben: I figured it out. `Omit<SomeType, 'someKey'>` when `SomeType` has a `{ someKey: string; [x: string]: unknown }` signature turns the whole thing into a `{ [x:string]: unknown }`... -_-

Fixed that by using `OmitKey` (non std) instead.

packages/pds/tests/db.test.ts
matthieusieben: Note that we _could_ use a `string` instead of a `symbol` here (e.g. `__encodedType`) but that would mean that the following would be allowed:

```ts
const foo: JsonEncoded<{ foo: number }> = '{"foo":42}'
foo.__encodedType.foo // TS: 42; Runtime: Error
```

The use of a `unique symbol` is the proper TS way of implementing tagged types. The caveat here is only needed because of the circular dependency between `dev-env` and `pds`.

packages/oauth/oauth-provider/src/token/token-manager.ts
matthieusieben: Haha I love this construct. But I can change to ifs.

packages/oauth/oauth-provider/src/types/email.ts
matthieusieben: Internally, `zod` uses a regexp for validation. This validation strategy *could* be less permissive in some (edge) cases than `@hapi/address` as the latter uses an algorithm based on the spec. Truth is, I don't really know if the set of emails allowed by `@hapi/address` is covered by the set of emails allowed by `zod`, nor do I think we should investigate it.

Because of this uncertainty, and because other part of our codebases rely solely on `zod`, this code only allows emails that are valid according to both libraries ensuring that we never encounter a case where an email allowed here is in a format that would be rejected by other parts of our systems.

matthieusieben: Note that this behavior was not changed with this PR (code was moved)

packages/oauth/oauth-provider/src/token/token-manager.ts
matthieusieben: This code is used through OAuth's token introspection endpoint, which is supposed to only be called with either an access token or a refresh token. Since access tokens are always JWTs, we do not need to support introspecting a token id, besides for legacy reasons.

I've added the behavior back (for legacy)

packages/oauth/oauth-provider/src/token/token-manager.ts
matthieusieben: Should be fixed

packages/pds/tests/db.test.ts
matthieusieben: Note that I changed the implementation to use an import from `dist`, which does fix the issue.

packages/oauth/oauth-provider/src/token/token-manager.ts
devinivy: Yeah I'd either add the behavior back or just leave a comment explaining that.

packages/oauth/oauth-provider/src/router/api-router.ts
```diff
@@ -0,0 +1,899 @@
...
+      try {
+        // Prevent caching of API routes
+        res.setHeader('Cache-Control', 'no-store')
+        res.setHeader('Pragma', 'no-cache')
+
+        // Prevent CORS requests
+        validateFetchMode(req, ['same-origin'])
+        validateFetchSite(req, ['same-origin'])
+        validateOrigin(req, issuerOrigin)
+        const referer = validateReferer(req, { origin: issuerOrigin })
```
devinivy, line 724: The HTTP `Referer` header misspells the word "referrer".  It's not consequential here, but we could consider using the normal spelling, e.g.:
```ts
const referrer = validateReferrer(req, { origin: issuerOrigin })
```

packages/oauth/oauth-provider/src/router/api-router.ts
```diff
@@ -0,0 +1,899 @@
...
+
+        // Ensure we are one the right page
+        if (
+          referer.pathname !== '/oauth/authorize' &&
+          referer.pathname !== '/account' &&
+          !referer.pathname.startsWith(`/account/`)
+        ) {
+          throw createHttpError(400, `Invalid referer ${referer}`)
+        }
```
devinivy, line 733: I am not sure if trailing slashes are allowed on those pages, but if they are we might consider permitting them here.

packages/oauth/oauth-provider/src/router/api-router.ts
```diff
@@ -0,0 +1,899 @@
...
+  Object.freeze({
+    expires: undefined, // session
+    sameSite: 'strict',
+    secure: true,
+    httpOnly: true,
+    path: '/',
+  })
+
+function ephemeralCookieValue(requestUri?: RequestUri) {
+  return requestUri ? `sec-${requestUri}` : 'sec-no-request'
```
devinivy, line 777: What does `sec` stand for?

devinivy (commented):
Have now reviewed everything outside -ui and -frontend.  Feeling quite solid! 👏

packages/oauth/oauth-provider/src/account/account-manager.ts
```diff
@@ -113,108 +116,236 @@ export class AccountManager {
...
+        account,
+        deviceId,
+        deviceMetadata,
+        requestId,
+      })
+
+      return { account, ephemeralCookie }
+    } catch (err) {
+      await this.removeDeviceAccount(deviceId, account.sub)
+
```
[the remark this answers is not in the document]
matthieusieben: Hooks should normally not throw unless they actually want to prevent what happens next. Not sure what the right call is here.

packages/oauth/oauth-provider/src/router/api-router.ts
matthieusieben: We actually don't. Added a comment.

packages/oauth/oauth-provider/src/router/api-router.ts
```diff
@@ -0,0 +1,899 @@
...
+              deviceId,
+              deviceMetadata,
+            )
+
+            const clientData = authorizedClients.get(clientId)
+            if (server.checkConsentRequired(parameters, clientData)) {
+              const scopes = new Set(clientData?.authorizedScopes)
+
+              // Add the newly accepted scopes to the authorized scopes
+              for (const s of parameters.scope?.split(' ') ?? []) scopes.add(s)
```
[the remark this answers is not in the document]
matthieusieben: This is not permitted (per spec) and actually enforced by the `oauthScopeSchema`

packages/oauth/oauth-provider/src/router/api-router.ts
matthieusieben: Changed to something more explicit (this will likely disappear)

packages/pds/tests/db.test.ts
```diff
@@ -1,5 +1,5 @@
 import { TestNetworkNoAppView } from '@atproto/dev-env'
-import { AccountDb } from '../src/account-manager/db'
+import { AccountDb } from '../dist/account-manager/db'
```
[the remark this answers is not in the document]
matthieusieben: This is an intentional work around the `unique symbol` issue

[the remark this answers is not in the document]
matthieusieben: added a comment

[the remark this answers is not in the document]
matthieusieben: See this comment for more info https://github.com/bluesky-social/atproto/pull/3659#discussion_r2032730491

packages/oauth/oauth-provider-frontend/src/routes/account/_appLayout/$sub.tsx
```diff
@@ -90,9 +90,9 @@ export function AccountHome() {
         />
       ) : accountSessions.length > 0 ? (
         <div className="space-y-3">
-          {accountSessions.map((session, i) => (
+          {accountSessions.map((session) => (
             <AccountSessionCard
-              key={session.deviceId + session.deviceMetadata.lastSeenAt}
+              key={`${sub}@${session.deviceId}`}
```
estrattonbailey, line 97: Is it impossible to have multiple sessions with the same deviceId?

PR #3673, opened 2025-03-26
v2 trends endpoint

haileyok:
~there's some borked codegen in here that needs fixing. i'd really like it if we could just commit everything that codegen is giving me, but not sure if that's an option.~ (fixed)


lexicons/app/bsky/unspecced/defs.json
```diff
@@ -32,6 +32,58 @@
...
+    "trendView": {
+      "type": "object",
+      "required": [
+        "topic",
+        "displayName",
+        "link",
+        "startedAt",
+        "postCount",
+        "hot"
+      ],
```
devinivy, line 71: Any special reason to omit `actors` here?

lexicons/app/bsky/unspecced/defs.json
```diff
@@ -32,6 +32,58 @@
...
+    "skeletonTrend": {
+      "type": "object",
+      "required": [
+        "topic",
+        "displayName",
+        "link",
+        "startedAt",
+        "postCount",
+        "hot"
+      ],
```
devinivy, line 45: Same here— any reason to omit `dids`?

lexicons/app/bsky/unspecced/defs.json
```diff
@@ -32,6 +32,58 @@
...
+      "properties": {
+        "topic": { "type": "string" },
+        "displayName": { "type": "string" },
+        "link": { "type": "string" },
```
devinivy, line 49: What goes in here?  If it's a URL we could use the `uri` string format.

haileyok: it's not actually a valid uri, but rather just a a path. it _could_ be a uri at some point in the future but probably won't be. totally could just rename this though for clarity

lexicons/app/bsky/unspecced/defs.json
haileyok: we probably won't always need to send them all the time is the main reason (i.e. we might only send them in certain cases)

lexicons/app/bsky/unspecced/defs.json
haileyok: same as above

lexicons/app/bsky/unspecced/defs.json
haileyok: cc @mozzius to verify here. do we think we _always_ want them to come down?

estrattonbailey: I think we should always send them, but if that's been talked about I must have missed it

lexicons/app/bsky/unspecced/defs.json
```diff
@@ -32,6 +32,46 @@
...
+          }
+        }
+      }
+    },
+    "trendView": {
+      "type": "object",
+      "required": ["topic", "displayName", "link", "startedAt", "postCount"],
+      "properties": {
+        "topic": { "type": "string" },
+        "displayName": { "type": "string" },
```
estrattonbailey, line 61: Posting here for sake of completeness: what's the distinction between these two values now? Is this vestigial from v1 when these were search queries? Maybe we can just use `topic` in that case, like we do for v1.

haileyok: i personally like having the distinction between the exact topic (the exact keyword/hashtag that is used for the topic) and a display name for that topic (which searching for might not even return any values of). the former isn't actually used anywhere right now, but we could in the future (i.e. ui features to like...search for more). also unsure if any other apps are taking advantage of that (there are a few that have implemented topics though)

estrattonbailey: Cool yeah that's fair. So in v2 we'll actually populate both fields, but use `displayName` for now 👍

lexicons/app/bsky/unspecced/defs.json
estrattonbailey: Unless we change the UI I'd anticipate us always wanting avatars, unless there's some performance reason to rug it. So in that case, a fallback of an empty array is fine imo. But if you're anticipating other use cases where we don't need to pull all that data and just pull the trend itself, then I think leaving it optional is more correct.

lexicons/app/bsky/unspecced/defs.json
haileyok: i think that's what we are already doing. there's a `displayName` in the old lexicon that we should be using for displaying in the sidebar i believe

lexicons/app/bsky/unspecced/getTrends.json
```diff
@@ -0,0 +1,42 @@
...
+    "main": {
+      "type": "query",
+      "description": "Get the current trends on the network",
+      "parameters": {
+        "type": "params",
+        "properties": {
+          "viewer": {
+            "type": "string",
+            "format": "did",
+            "description": "DID of the account making the request (not included for public/unauthenticated queries)."
```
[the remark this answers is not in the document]
haileyok: ah i copy pasted here. i suepect we want this only on the skeleton endpoint?

lexicons/app/bsky/unspecced/defs.json
haileyok: the client can send people to anywhere it wants within the app. so i.e. it might be `/profile/hailey.at/feed/rkey`, which will result in that page/screen being opened

PR #3684, opened 2025-03-27
Deterministic CAR files on sync.getRepo

dholms:


devinivy (approved):
Would be cool to have a test in the pds to document the behavior, but this does look good to me.

dholms:
Unfortunately this rewrite involves running to sqlite for every block and makes the route prohibitively slow. Needs to be refactored before merging

PR #3687, opened 2025-03-27
auth scope scratch work

dholms:


lexicons/app/bsky/auth/base.json
```diff
@@ -0,0 +1,26 @@
+{
+  "lexicon": 1,
+  "id": "app.bsky.auth.base",
+  "defs": {
+    "main": {
+      "type": "auth-bundle",
+      "description": "Create and delete data in your Bluesky account including your profile, posts, follows, lists, blocks, and viewing feeds.",
```
dholms, line 7: wordsmithing a high-level scope like this is tough 😅

lexicons/tools/ozone/auth/reporter.json
```diff
@@ -0,0 +1,13 @@
+{
+  "lexicon": 1,
+  "id": "tools.ozone.auth.reporter",
+  "defs": {
+    "main": {
+      "type": "auth-bundle",
+      "description": "Report content and view status of reports",
+      "permissions": [
+        "com.atproto.resource.authentication?aud=*&lxm=tools.ozone.moderation.createReport"
```
dholms, line 9: implies moving this method into the ozone namespace which I think we're down with (can keep around the com.atproto method for now as well)

lexicons/com/atproto/resource/account.json
```diff
@@ -0,0 +1,40 @@
+{
+  "lexicon": 1,
+  "id": "com.atproto.resource.account",
```
dholms, line 3: I'm not sure about having these all in the same "resource". We're kinda mixing together the scariest permissions (deleteAccount) with the permissions that basically every app will be requesting access to (viewEmail).

An alternative would be to split this into several resources such as:
- com.atproto.resource.email
- com.atproto.resource.account
- com.atproto.resource.app_password

lexicons/com/atproto/resource/authentication.json
```diff
@@ -0,0 +1,24 @@
+{
+  "lexicon": 1,
+  "id": "com.atproto.resource.authentication",
```
dholms, line 3: went with `authentication` here instead of `serviceAuth` since i felt it was cleaner & matched the other schemas better stylistically

lexicons/com/atproto/resource/account.json
matthieusieben: +1 for splitting.

What do you think about:

- `com.atproto.account.deletion`
- `com.atproto.account.activation` // No params (allows to activate/deactivate)
- `com.atproto.account.email?view=true&update=false`
- `com.atproto.account.appPassword?create=false&revoke=true&list=true`

lexicons/com/atproto/resource/authentication.json
```diff
@@ -0,0 +1,24 @@
...
+  "defs": {
+    "main": {
+      "type": "auth-resource",
+      "description": "Grants access to authentication for third party servcies (service auth)",
```
matthieusieben, line 7: Should we forbid the use of `aud=*&lxm=*`?

lexicons/com/atproto/resource/account.json
dholms: yeah I dig that quite a bit. I think it's easier to reason about & prevents long possibly confusing query strings

lexicons/com/atproto/resource/authentication.json
dholms: I don't necessarily think so. We have resources that are more powerful (for instance full write access to your repo I would consider more powerful). I think we can just flag something like that pretty aggressively in the auth screen

PR #3699, opened 2025-03-31
Improve logging of XRPC errors

matthieusieben:
Currently, "known errors" are logged with the following info:

```
message = "error in xrpc method <nsid>"

{
  dd: { }
  hostname: 'abc'
  level: 50
  name: 'xrpc-server'
  pid: 42
  status: 400
  time: 1743382556683
}
```

This is not enough information to allow useful debugging of un-expected behaviors in the case of known errors.

In addition, the `customErrorName` optional property will override the logger's `name` when set, resulting in inconsistent logs. Indeed `name` is usually `xrpc-server`, unless the error has an `customErrorName`, in which case `name` is another value, like `InvalidPhoneVerification`.

https://github.com/bluesky-social/atproto/blob/713a52d4dc9a242fd066e0cbd21934ba0367366e/packages/xrpc-server/src/server.ts#L555

This PR adds more details to the logged events:

- `nsid` (for easy filtering)
- `req` details (id, timing, etc.) -- if `pino-http` is used 
- `err` for all logs, stripping the `stack` for expected errors, in production enviroments
- response `payload`, ensuring that the XRPC error name sent to the browser is always available.

This PR changes the following log event properties:

- `name` will always be `xrpc-server`
- `customErrorName` is now available through `payload.error` instead of `name`
- `message` is now available through `err.message`


packages/xrpc-server/src/server.ts
```diff
@@ -542,24 +542,68 @@ function createErrorMiddleware({
...
+        // @NOTE Computation of error stack is an expensive operation, so
+        // we strip it for expected errors.
+        err:
+          isInternalError || process.env.NODE_ENV === 'development'
+            ? err
+            : toSimplifiedErrorLike(err),
+
+        // XRPC specific properties, for easier browsing of logs
+        nsid: locals?.nsid,
+        status: xrpcError.type,
```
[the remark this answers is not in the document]
matthieusieben: I agree, this was done before your recent changes ;-)

packages/xrpc-server/src/server.ts
```diff
@@ -542,24 +542,68 @@ function createErrorMiddleware({
...
+} {
+  return typeof (req as { log?: any }).log?.error === 'function'
+}
+
+function toSimplifiedErrorLike(err: unknown): unknown {
+  if (err instanceof Error) {
+    // Transform into an "ErrorLike" for pino's std "err" serializer
+    return {
+      ...err,
+      // Cary over non-enumerable properties
```
matthieusieben, line 596:
```suggestion
      // Carry over non-enumerable properties
```

packages/xrpc-server/src/server.ts
matthieusieben, line 565:
```suggestion
        type: xrpcError.type,
        status: xrpcError.statusCode,
```

PR #3701, opened 2025-03-31
Add `client` to `onSignedIn`, `onSignedUp` and `onSignupAttempt` hooks

matthieusieben:


matthieusieben:
Closing because:
- This requires loading the client metadata before the user is singed-in, which we want to avoid
- This requires loading client metadata whether we'll need them or not
- Replacing in favor of exposing the `requestId` in https://github.com/bluesky-social/atproto/pull/3659

PR #3714, opened 2025-04-04
new host status com.atproto.sync lexicons

bnewbold:
The immediate use-case here is to help backfill new full-network relay instances, using the host list of an existing relay. It could also help relays stay in sync (eg, periodically pull an updated list of hosts), though this isn't really thought-through for that use case.

A side-effect is making it easier to see how large the network is (number of PDS instances, rough estimate of total accounts). FWIW, both bigsky and the new relay keep per-host counts pre-computed, so this is a simple/cheap single-table queries. That field is also left optional in case other relay implementations don't keep track in the same way.

This introduces a concept of "host status", similar to "account status". My thinking on these:

- `active`: actively connected (or reconnecting)
- `idle`: disconnected due to lack of activity (but would retry on `requestCrawl`)
- `offline`: disconnected and failed to connect after retries (but would retry on `requestCrawl`)
- `throttled`: some host-level quota has been exceeded: for example, total number of accounts, or events-per-hour. may or may not be connected.
- `banned`: disconnected indefinitely, and not allowed to reconnect

Open to cutting down or iterating on this list.

Also added a `getHostStatus` which could be used for easier direct fetching of info for a single hostname. In a fuzzy sense this could help debug various connectivity issues, though the short-term motivation is to just avoid folks iterating over the whole list if they are only interested in a single host.

Things for review:
- is `com.atproto.sync.*` the right place for these? I think so.
- would an AppView or something like jetstream implement these endpoints? I think it is fine if they do, but not expected/required
- the host metadata schema is the same between the two for now. should they be unified? might want to allow `getHostStatus` to evolve in to a more "detailed" set of fields
- I wiped this off pretty quickly and haven't done codegen or implemented these yet

### DIFF
--- a/lexicons/com/atproto/sync/listHosts.json
+++ b/lexicons/com/atproto/sync/listHosts.json
@@ -47,8 +47,8 @@
         },
         "accountCount": { "type": "integer" },
         "status": {
-          "type": "string",
-          "knownValues": ["active", "idle", "offline", "throttled", "banned"]
+          "type": "ref",
+          "ref": "com.atproto.sync.defs#hostStatus"
         }
       }
     }
--- a/lexicons/com/atproto/sync/listHosts.json
+++ b/lexicons/com/atproto/sync/listHosts.json
@@ -1,0 +1,56 @@
+{
+  "lexicon": 1,
+  "id": "com.atproto.sync.listHosts",
+  "defs": {
+    "main": {
+      "type": "query",
+      "description": "Enumerates upstream hosts (eg, PDS or relay instances) that this service consumes from. Implemented by relays.",
+      "parameters": {
+        "type": "params",
+        "properties": {
+          "limit": {
+            "type": "integer",
+            "minimum": 1,
+            "maximum": 1000,
+            "default": 200
+          },
+          "cursor": { "type": "string" }
+        }
+      },
+      "output": {
+        "encoding": "application/json",
+        "schema": {
+          "type": "object",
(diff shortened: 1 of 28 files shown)


lexicons/com/atproto/sync/listHosts.json
```diff
@@ -0,0 +1,56 @@
...
+          "type": "integer",
+          "description": "Recent repo stream event sequence number. May be delayed from actual stream processing (eg, persisted cursor not in-memory cursor)."
+        },
+        "accountCount": { "type": "integer" },
+        "status": {
+          "type": "string",
+          "knownValues": ["active", "idle", "offline", "throttled", "banned"]
+        }
+      }
+    }
```
devinivy, line 54: Any merit in sharing this "type" between `listHosts` and `getHostStatus` within `com.atproto.sync.defs`?

bnewbold:
updated with `listHosts` broken out in to `com.atproto.sync.defs`, and added a patch-level changeset on `@atproto/api`.

If this gets approved, i'll run codegen and merge.

lexicons/com/atproto/sync/listHosts.json
devinivy: I see you broke out `hostStatus`.  I was actually picturing this whole host shape `{ hpstname, seq, accountCount, status }`, but no biggie either way!

devinivy (approved):

PR #3715, opened 2025-04-04
fix lex-cli record name conflicts

### LINKED ISSUE #3711: lex cli gen-api doesn't work when two different record types have the same ending
**Describe the bug**

the lex-cli's gen-api function generates a central index.ts file at the top level of the folder it's generating into. In it, for each record lexicon that is used, a class is exported named <lexicon-name-ending>Record (ex. for `com.whwnd.blog.entry`, the class would be named `EntryRecord`. 

This is problematic because it means that no two record types with the same ending can be used without errors of the same class being defined twice. Many lexicons share endings, such as `post` (used by bsky.app, sprk.so, flashes.blue, and more), `like` (used by bsky.app and sprk.so) and `follow` (used by bsky.app and tangled.sh)

This bug makes it impossible to use multiple of these at once in a project that uses lex-cli

**To Reproduce**

1. Open [Statusphere React](https://github.com/mozzius/statusphere-react), a project that uses lex-cli's gen-api function
2. Duplicate the file `status.json` in lexicons/xyz/statusphere and rename the new file to `profile.json`
3. Change the `id` on line 3 to `xyz.statusphere.profile`

You have now made a new profile record type in the `xyz/statusphere` folder. Another profile record type exists in the `app/bsky/actor` folder, but these should not conflict with each other because their ids are completely different.

4. Run `pnpm run lexgen` in the root project folder
5. Open `packages/lexicon/src/index.ts`, Ctrl+F for `ProfileRecord`. Notice there are two declarations and exports of the same class. 

**Expected behavior**

Records should be named with their full names and not just their endings, given that they are exported outside of the export of their domain's lexicon queries and procedures.

For instance, xyz.statusphere.profile should be named `XyzStatusphereProfileRecord` and not simply `ProfileRecord`. This will make it impossible for different lexicons to conflict in this way.

knotbin:
This pull request fixes a bug that made the lex-cli's gen-api command not work with any lexicon that had the same ID ending (className).

Currently, only the className is used for the record name in the root index file, so `app.bsky.feed.like` would simply be LikeRecord, and if your app were to support both `app.bsky.feed.like` and `so.sprk.feed.like`, the gen-api command would error out. This is very bad as it makes interoperability with multiple lexicon schemas with the same class name impossible.

In this PR, records generated into the `index.ts` file outputted at the root of the output are now named with their full names and not just their endings, given that they are exported outside of the export of their domain's lexicon queries and procedures.

For instance, xyz.statusphere.profile is now named XyzStatusphereProfileRecord and not simply ProfileRecord. This will make it impossible for different lexicons to conflict in this way.

Closes #3711


matthieusieben (commented):
That looks nice. This would require a `minor` patch, which you can generate with `pnpm changeset`.

knotbin:
Done!

knotbin:
> That looks nice. This would require a `minor` version bump, which you can generate with `pnpm changeset`.

Done! Let me know if anything else is needed

knotbin:
@matthieusieben any feedback on this?

knotbin:
@matthieusieben any way you could get back to this? it's blocking some work for me

matthieusieben (approved):

matthieusieben:
And thank you for your contribution 🙏

PR #3729, opened 2025-04-08
Msi/oauth account api alt

matthieusieben:
This PR changes the way authentication is done when the user does not press the "remember me" checkbox in the sign-in flow.

In those instance, signing-in or signing-up will result in an "api token" to be returned to the front-end. That token can then be used as `Bearer` that will cause the back-end to skip the verification for a valid `device_account` DB row.

The motivation behind this change is a a big simplification in the management of device-account sessions, both in terms of code as well as in terms of UI/UX. This change also makes these ephemeral session safer as the credentials are not stored in cookies.

The tokens are bound to the `sub`, `deviceId`, `requestId` and can only be used in the same context as the one they were created for (a token created for a particular request flow can only be used in the context of that request flow).

The reason for using a token in the JS memory is that the token will be "forgotten" by the device as soon as the page is left (as opposed to a session cookie that can stay active for hours after a page was left). The "back-forward cache" could still be used to "revive" the browser's JS memory but the back-forward cache is actively being busted ([here](https://github.com/bluesky-social/atproto/blob/e4f4a62ea9e0ce6dff8f0348b1fdd60d20eccb35/packages/oauth/oauth-provider/src/router/send-redirect.ts#L124-L126)), so that shouldn't be an issue.


packages/oauth/oauth-provider-api/src/api-endpoints.ts
```diff
@@ -13,12 +13,12 @@ export type ApiEndpoints = {
   '/sign-up': {
     method: 'POST'
     input: SignUpInput
-    output: { account: Account }
+    output: { account: Account; token?: string }
```
devinivy, line 16: Since there are lots of "tokens" floating around the auth flows, I wonder if it may be nice to give these a more specific name to indicate what they are for / what their lifetime is.

packages/oauth/oauth-provider/src/account/account-manager.ts
```diff
@@ -144,30 +137,15 @@ export class AccountManager {
...
-
-    const ephemeralCookie = remembered ? null : await randomHexId(32)
-
-    await this.upsertDeviceAccount(
-      deviceId,
-      account,
-      remembered,
-      requestId,
-      ephemeralCookie,
-    )
```
devinivy, line 159: I see we no longer upsert a device-account in account creation but we do still remove a device-account in the case of a failure.  Just double-checking these are both the intended behavior.

packages/pds/src/account-manager/db/migrations/005-oauth-account-management.ts
```diff
@@ -109,61 +83,6 @@ export async function up(
     .column('did')
     .execute()
 
-  await db.schema
-    .createTable('account_device_request')
```
devinivy, line 113: We should remove this from the `down()` as well.

devinivy (approved):
This is indeed a nice simplification.  And looks like we don't need any new secrets, which is neat.

My main feedback after reading through is that we could document the scope/lifetime/purpose of this new token, and might be able to do better self-documentation around it e.g. by being more specific with naming.  A short comment in the code describing how the token comes into play would also probably be useful.

PR #3733, opened 2025-04-09
Appview: skip actor cache for viewer

devinivy:
This applies new cache control behavior to dataplane calls to `GetActors` within the appview.  In short, we want to ensure the user always sees fresh data about themselves.  There are a few other places where we skip the cache, mostly in moderation and admin-specific settings.


jcalabro (approved):
🐎🐎🐎 

Not sure I'm the best person to be reviewing yet, but this is great!

ericvolp12 (approved):
💎 🔥 💎

PR #3735, opened 2025-04-10
Ozone: add metadata endpoint

devinivy:
The ozone distribution uses a [metadata endpoint](https://github.com/bluesky-social/ozone/blob/b6ad1e5e9102373b2c2e94dd99b758c0362b54ad/service/index.js#L34-L40) which allows ozone to help get itself configured through the UI.  Moving this up into the ozone backend itself is convenient when the backend and frontend aren't deployed together through the ozone distribution.

We can remove the endpoint from the ozone distribution once these changes are incorporated there.

### DIFF
--- a/.changeset/tall-rice-sparkle.md
+++ b/.changeset/tall-rice-sparkle.md
@@ -1,0 +1,5 @@
+---
+"@atproto/ozone": patch
+---
+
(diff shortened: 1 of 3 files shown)


bnewbold (approved):
code looks good, and this has been working as deployed

PR #3741, opened 2025-04-11
Update DM reaction length on chat lexicon

ghost:


.changeset/strange-donuts-clap.md
```diff
@@ -0,0 +1,8 @@
+---
+"@atproto/ozone": patch
+"@atproto/bsky": patch
+"@atproto/api": patch
+"@atproto/pds": patch
```
devinivy, line 5: We don't need to explicitly list the pds, ozone, or bsky in these, I don't think, as it doesn't actually change any of their behavior.  The pds version will still get bumped since its dep is updated. In the changelog it will appear as a dependency update rather than a specific line item for the reaction limit being updated.

devinivy (approved):

PR #3761, opened 2025-04-16
Delegated verification

ghost:


packages/api/src/client/lexicons.ts
```diff
@@ -4564,6 +4564,10 @@ export const schemaDict = {
             type: 'string',
             format: 'datetime',
           },
+          verification: {
```
devinivy, line 4567: Can we add a changeset for the @atproto/api package?

packages/bsky/src/views/index.ts
```diff
@@ -385,6 +389,62 @@ export class Views {
...
+        }
+      },
+    )
+    const hasValidVerification = verifications.some((v) => v.isValid)
+
+    const verifiedStatus = hasValidVerification
+      ? isImpersonation
+        ? 'invalid'
+        : 'valid'
+      : 'none'
```
devinivy, line 425: I sort of expected the logic here to look something like this:
```suggestion
    const verifiedStatus = verifications.length ?
      ? hasValidVerification && !isImpersonation
        ? 'valid'
        : 'invalid'
      : 'none'
```

devinivy (approved):
Looking solid 👍

PR #3765, opened 2025-04-17
❇️ Verifications manager in ozone

foysalit:
This PR introduces verification related functionalities in ozone. Namely:

1. Indexing and viewing verifications in ozone using various filters like verifier, verified subjects, specific timeframe etc.
2. Verifying users in bulk
3. Revoking verifications from users in bulk

A few moving pieces to pay attention to

1. We've introduced a new role `roleVerifier` that's similar to `roleTriage` but also allows interacting with verification endpoints. 
2. To implement jetstream listener, we are now exposing `WebSocketKeepAlive` from `xrpc-server` and spinning up the listener inside the ozone daemon. The types in this are a bit scrappy and not full-proof.


packages/ozone/src/daemon/verification-listener.ts
```diff
@@ -0,0 +1,142 @@
...
+      this.cursor = Number(cursor)
+    }
+    return this.cursor
+  }
+
+  async updateCursor(cursor: number) {
+    // Assuming cursors are always incremental, if we have processed an event with higher value cursor, let's not update to a lower value
+    if (this.cursor && this.cursor >= cursor) {
+      return
+    }
```
devinivy, line 93: Nice 👍

packages/ozone/src/daemon/verification-listener.ts
```diff
@@ -0,0 +1,142 @@
...
+
+    await this.verificationService.updateFirehoseCursor(`${cursor}`)
+    this.cursor = cursor
+  }
```
devinivy, line 97: As written this method can still race and cause the cursor to move backwards.
1. Current cursor is `4`
2. Update to cursor `5` and update to cursor `6` are kicked off near the same time.
3. They both make it past the check `this.cursor >= cursor` since `this.cursor` is currently `4`.
4. The update lands for cursor `6`.
5. The update lands for cursor `5`.

packages/ozone/src/db/migrations/20250415T201720309Z-verification.ts
```diff
@@ -0,0 +1,25 @@
+import { Kysely, sql } from 'kysely'
+
+export async function up(db: Kysely<unknown>): Promise<void> {
+  await db.schema
+    .createTable('verification')
+    .addColumn('uri', 'text', (col) => col.primaryKey())
```
devinivy, line 6: Since we don't keep the full record contents around, keeping around the `cid` in here would be a nice backstop.

packages/ozone/src/db/migrations/20250415T201720309Z-verification.ts
```diff
@@ -0,0 +1,25 @@
...
+    .addColumn('issuer', 'text', (col) => col.notNull())
+    .addColumn('subject', 'text', (col) => col.notNull())
+    .addColumn('handle', 'text', (col) => col.notNull())
+    .addColumn('displayName', 'text', (col) => col.notNull())
+    .addColumn('revokeReason', 'text')
+    .addColumn('revokedBy', 'text')
+    .addColumn('revokedAt', 'text')
+    .addColumn('createdAt', 'text', (col) =>
+      col.defaultTo(sql`now()`).notNull(),
+    )
```
devinivy, line 16: Is this the `createdAt` time contained in the record or ozone's initial indexed time?

packages/ozone/src/db/schema/verification.ts
```diff
@@ -0,0 +1,18 @@
...
+  uri: string
+  issuer: string
+  subject: string
+  handle: string
+  displayName: string
+  revokeReason: string | null
+  revokedBy: string | null
+  revokedAt: string | null
+  createdAt: string
+  updatedAt?: string
```
devinivy, line 13: We generally don't want to mark fields in the model as optional, since they always are present when reading data.  If you're looking for the convenience of not including the field on writes, you may want to use kysely's `Generated` type helper.

packages/ozone/src/jetstream/service.ts
```diff
@@ -0,0 +1,110 @@
...
+      if (parsedMessage.kind === 'commit') {
+        const { collection, operation, record } = parsedMessage.commit || {}
+
+        if (operation === 'create') {
+          options.onCreate?.[collection]?.(
+            parsedMessage as CommitCreateEvent<typeof record>,
+          )
+        } else if (operation === 'delete') {
+          options.onDelete?.[collection]?.(parsedMessage as CommitDeleteEvent)
+        }
```
devinivy, line 99: We should ensure that the record contents on the firehose gets lexicon-validated, as it's not guaranteed to be valid.

packages/ozone/tests/verification-listener.test.ts
```diff
@@ -0,0 +1,90 @@
...
+      kind: 'commit',
+      did: sc.dids.bob,
+      time_us: 123456789,
+      commit: {
+        rev: 'xyz',
+        operation: 'create',
+        collection: 'app.bsky.graph.verification',
+        rkey: 'abcdefg',
+        record: {
+          subject: sc.dids.alice,
```
devinivy, line 60:
```suggestion
        record: {
          $type: 'app.bsky.graph.verification',
          subject: sc.dids.alice,
```

packages/ozone/tests/verification-listener.test.ts
```diff
@@ -0,0 +1,90 @@
...
+          handle: sc.accounts[sc.dids.alice].handle,
+          displayName: 'Alice',
+          createdAt: new Date().toISOString(),
+        },
```
devinivy, line 64: Could type check the record contents with `satisfies Verification.Record`.

packages/ozone/tsconfig.build.json
```diff
@@ -3,6 +3,10 @@
   "compilerOptions": {
     "rootDir": "./src",
     "outDir": "./dist",
+    "module": "CommonJS",
+    "moduleResolution": "Node",
+    "esModuleInterop": true,
+    "target": "ES2020",
```
devinivy, line 9: I wonder if this was left over from trying to get the jetstream sdk to work.

packages/xrpc-server/src/stream/index.ts
```diff
@@ -3,3 +3,4 @@ export * from './frames'
 export * from './stream'
 export * from './subscription'
 export * from './server'
+export * from './websocket-keepalive'
```
devinivy, line 6: Let's get a changeset in for this so the published version of `@atproto/ozone` is able to use the functionality.

packages/ozone/tests/verification-listener.test.ts
```diff
@@ -0,0 +1,90 @@
...
+        rev: 'yza',
+        operation: 'delete',
+        collection: 'app.bsky.graph.verification',
+        rkey: 'abcdefg',
+      },
+    }
+    relay.send(JSON.stringify(createEvent))
+    relay.send(JSON.stringify(deleteEvent))
+    // Give the processor enough time to handle the events
+    await new Promise((resolve) => setTimeout(() => resolve(true), 500))
```
devinivy, line 81: May be able to sidestep this by polling the consumer until its cursor moves past the delete event's cursor.

packages/ozone/src/auth-verifier.ts
```diff
@@ -82,14 +85,13 @@ export class AuthVerifier {
 
   moderator = async (reqCtx: ReqCtx): Promise<ModeratorOutput> => {
     const creds = await this.standard(reqCtx)
-    if (!creds.credentials.isTriage) {
+    if (!creds.credentials.isTriage && !creds.credentials.isVerifier) {
       throw new AuthRequiredError('not a moderator account')
     }
```
devinivy, line 90: This previously was the check that the user had at least triage access, and everywhere else I expect it was implicit.  Do we need to add explicit `isTriage` checks to any endpoints to restrict members who are verifiers but not triage?

packages/ozone/src/daemon/context.ts
```diff
@@ -88,6 +91,16 @@ export class DaemonContext {
       cfg.db.materializedViewRefreshIntervalMs,
     )
 
+    const verificationService = VerificationService.creator()
+    const verificationListener = cfg.verifier?.jetstreamUrl
+      ? new VerificationListener(
+          verificationService(db),
+          new BackgroundQueue(db, { concurrency: 1 }),
```
devinivy, line 98: Since this background queue isn't shared, perhaps it can be created internally to `VerificationListener`.  That would also allow `VerificationListener` to ensure `concurrency: 1` if it relies on that.

packages/ozone/src/api/verification/grant.ts
```diff
@@ -0,0 +1,67 @@
...
+
+export default function (server: Server, ctx: AppContext) {
+  server.tools.ozone.verification.grant({
+    auth: ctx.authVerifier.modOrAdminToken,
+    handler: async ({ input, auth, req }) => {
+      if (!ctx.cfg.verifier) {
+        throw new InvalidRequestError('Verifier not configured')
+      }
+
+      if (!auth.credentials.isAdmin && !auth.credentials.isVerifier) {
```
devinivy, line 14: If I'm understanding correctly admins have `isVerifier` flipped to `true`, so may be able to simplify the check.

devinivy: All good either way though if this is clearer 👍

packages/ozone/src/verification/issuer.ts
```diff
@@ -0,0 +1,119 @@
...
+
+export class VerificationIssuer {
+  constructor(private verifierConfig: VerifierConfig) {}
+
+  static creator() {
+    return (verifierConfig: VerifierConfig) =>
+      new VerificationIssuer(verifierConfig)
+  }
+
+  // @TODO: Probably shouldn't login for every req?
```
devinivy, line 25: Agree, ideally we'd probably login once and then let it maintain its session.

packages/ozone/src/verification/issuer.ts
```diff
@@ -0,0 +1,119 @@
...
+            throw new Error(`Only verification records can be revoked`)
+          }
+
+          if (atUri.host !== this.verifierConfig.did) {
+            throw new Error(
+              `Cannot revoke verification record ${uri} not issued by ${this.verifierConfig.did}`,
+            )
+          }
+
+          const agent = await this.getAgent()
```
devinivy, line 103: Given that `await this.getAgent()` currently logs in for every call, then we should at least share a single agent for the whole batch.  I.e. not create 100 sessions for a single batch of 100.

packages/ozone/src/team/index.ts
```diff
@@ -197,11 +198,13 @@ export class TeamService {
       isAdmin || member?.role === 'tools.ozone.team.defs#roleModerator'
     const isTriage =
       isModerator || member?.role === 'tools.ozone.team.defs#roleTriage'
+    const isVerifier = member?.role === 'tools.ozone.team.defs#roleVerifier'
```
devinivy, line 201: Should admins also have this flag?  That would be consistent with `adminToken()` in `auth-verifier`.

packages/ozone/src/verification/service.ts
```diff
@@ -0,0 +1,192 @@
...
+
+  async create(
+    verifications: Pick<
+      Verification,
+      'uri' | 'issuer' | 'subject' | 'handle' | 'displayName' | 'createdAt'
+    >[],
+  ) {
+    // @TODO: We should probably handle errors here?
+    await this.db.transaction(async (tx) => {
+      return Promise.allSettled(
```
devinivy, line 30: Ideally use the transaction so that they all create or all fail together.

devinivy: Actually you could just do a single insert with all the verifications, that is probably simplest!

packages/ozone/src/verification/service.ts
```diff
@@ -0,0 +1,192 @@
...
+    revokeReason,
+  }: {
+    uris: string[]
+    revokedBy?: string
+    revokedAt?: string
+    revokeReason?: string
+  }) {
+    const now = new Date().toISOString()
+    return Promise.allSettled(
+      uris.map(async (uri) => {
```
devinivy, line 55: It looks like callers of `markRevoked()` are ignoring errors surfaced here.  Perhaps it's more straightforward to use a transaction and allow all the updates to fail or succeed together, then surface the error if they fail.

packages/ozone/src/verification/service.ts
```diff
@@ -0,0 +1,192 @@
...
+
+    if (createdAfter) {
+      qb = qb.where('createdAt', '>', createdAfter)
+    }
+
+    if (createdBefore) {
+      qb = qb.where('createdAt', '>', createdBefore)
+    }
+
+    const keyset = new CreatedAtUriKeyset(ref(`createdAt`), ref('uri'))
```
devinivy, line 118: We'll want to support this pagination scheme with an index.

packages/ozone/src/verification/service.ts
```diff
@@ -0,0 +1,192 @@
...
+    isRevoked?: boolean
+    limit?: number
+  }) {
+    const { ref } = this.db.db.dynamic
+
+    let qb = this.db.db.selectFrom('verification').selectAll()
+
+    if (issuers.length) {
+      qb = qb.where('issuer', 'in', issuers)
+    }
```
devinivy, line 100: if querying by ozone's verifier did as an issuer (as I suspect it might?), maybe it would make sense to have an index supporting that too.

packages/ozone/src/verification/service.ts
```diff
@@ -0,0 +1,192 @@
...
+      qb = qb.where('subject', 'in', subjects)
+    }
+
+    if (createdAfter) {
+      qb = qb.where('createdAt', '>', createdAfter)
+    }
+
+    if (createdBefore) {
+      qb = qb.where('createdAt', '>', createdBefore)
+    }
```
devinivy, line 116: There's a fix in here for `createdBefore`.  Also making it inclusive, just to be consistent with `getEvents()`.
```suggestion
    if (createdAfter) {
      qb = qb.where('createdAt', '>=', createdAfter)
    }

    if (createdBefore) {
      qb = qb.where('createdAt', '<=', createdBefore)
    }
```

devinivy (commented):
Dropped some comments.  There are a handful but nothing major that I can see.  Good stuff!

packages/ozone/src/daemon/verification-listener.ts
```diff
@@ -0,0 +1,142 @@
...
+  async start() {
+    await this.getCursor()
+
+    this.jetstream = new Jetstream({
+      endpoint: this.jetstreamUrl,
+      cursor: this.cursor || undefined,
+      wantedCollections: [this.collection],
+      wantedDids: this.verifierIssuersToIndex?.length
+        ? this.verifierIssuersToIndex
+        : undefined,
```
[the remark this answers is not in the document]
foysalit: yeah this isn't really meant for us. I was mostly thinking of 3p labelers wanting to index verifications for only their own verifier of trust.

packages/ozone/src/config/env.ts
```diff
@@ -37,6 +37,11 @@ export const readEnv = (): OzoneEnvironment => {
     signingKeyHex: envStr('OZONE_SIGNING_KEY_HEX'),
     blobDivertUrl: envStr('OZONE_BLOB_DIVERT_URL'),
     blobDivertAdminPassword: envStr('OZONE_BLOB_DIVERT_ADMIN_PASSWORD'),
+    verifierUrl: envStr('OZONE_VERIFIER_URL'),
+    verifierDid: envStr('OZONE_VERIFIER_DID'),
+    verifierPassword: envStr('OZONE_VERIFIER_PASSWORD'),
+    verifierIssuersToIndex: envList('OZONE_VERIFIER_ISSUERS_TO_INDEX'),
+    jetstreamUrl: envStr('OZONE_VERIFIER_JETSTREAM_URL'),
```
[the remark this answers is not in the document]
foysalit: oh yeah that's what i was going for which is why the property isn't prefixed with verifier but left it in the env var name 🤦🏽‍♂️

packages/ozone/src/db/migrations/20250415T201720309Z-verification.ts
foysalit: this is meant to be ozone's index time but i see the record itself contains a `createdAt` timestamp so I'll use that.

packages/ozone/src/verification/issuer.ts
foysalit: yep yep. i had that outside of the loop for the record creator but missed this one.

packages/xrpc-server/src/stream/index.ts
foysalit: do i need to do that separately or just including `@atproto/xrpc-server` in the changeset of this PR would take care of that?

packages/ozone/src/daemon/verification-listener.ts
foysalit: hm yeah. moving the cursor value check inside the query, that should take care of it.

packages/ozone/src/verification/service.ts
foysalit: yh definitely easier. i was thinking we could allow partial success but makes things easier to just fail the entire transaction

packages/ozone/src/jetstream/service.ts
foysalit: hmm... does it make sense to do the validation here? this would allow listening for any records so we probably should be doing the validation in the specific handlers for the records right? if not, is there an example somewhere of validating generic records based on collection?

packages/ozone/src/auth-verifier.ts
foysalit: I don't believe there are any actions that triage mods can perform that we'd want to block verifiers from doing but I'll get this confirmed.

packages/ozone/tests/verification-listener.test.ts
foysalit: can you clarify this a bit please? do you mean like querying the db to check the cursor and waiting until it gets to the right value? like

```

    let hasCursorUpdated = false
    let attempt = 0
    do {
      const cursor = await verificationService.getFirehoseCursor()
      hasCursorUpdated = cursor === 123456799
      // cooldown between each check
      // await new Promise((resolve) => setTimeout(() => resolve(true), 10))
      attempt++
    } while (!hasCursorUpdated && attempt < 20)
```

on one hand, I think it adds value by adding a check for cursor update on the other hand, i feel like the additional complexity feels a bit unnecessary for the test. or may be i misunderstood the ask entirely?

foysalit:
@devinivy thank you! left some questions and addressed everything else you pointed out.
main things I'd like your eyes on
1. jetstream record validation approach
2. jetstream cursor race condition 
3. atp agent's session manager - I'm not sure if the current implementation will take care of auto refreshing the session?

packages/ozone/src/db/migrations/20250415T201720309Z-verification.ts
devinivy: Over time the same record may have many CIDs but at a given time it just has one, so we wouldn't need to make it a composite key.

packages/xrpc-server/src/stream/index.ts
devinivy: I would create two changesets in this PR: the existing one for "Add verification lexicons to ozone" and then a second one just for xrpc-server describing the change there.  That way in the xrpc-server changelog it will talk about the change relevant to that package rather than talking about verifications.

packages/ozone/tests/verification-listener.test.ts
devinivy: The intention is to not rely on an arbitrary amount of "wait" time in the tests, which may be flaky.  By polling regularly for the cursor update, we don't need to specify an arbitrary wait time in the tests.  This is similar to how `network.processAll()` functions to reliably wait for writes to propagate.

lexicons/tools/ozone/verification/revokeVerifications.json
```diff
@@ -0,0 +1,59 @@
...
+              }
+            },
+            "failedRevocations": {
+              "type": "array",
+              "description": "List of verification uris that couldn't be revoked, including failure reasons",
+              "items": {
+                "type": "string",
+                "format": "at-uri"
+              }
+            }
```
devinivy, line 53: Should this have room for a reason for the failure, similar to `#grantError` in the `grantVerifications` endpoint?

.changeset/dull-bobcats-beam.md
```diff
@@ -0,0 +1,9 @@
+---
+"@atproto/dev-env": patch
+"@atproto/ozone": patch
+"@atproto/api": patch
+"@atproto/pds": patch
+"@atproto/xrpc-server": patch
```
devinivy, line 6: I think ideally the change to xrpc-server would be its own separate changeset, with a description that will make sense in context of the xrpc-server changelog.

packages/ozone/src/verification/service.ts
```diff
@@ -0,0 +1,211 @@
...
+  }
+
+  async updateFirehoseCursor(cursor: number) {
+    const updated = await this.db.db
+      .updateTable('firehose_cursor')
+      .set({ cursor })
+      .where('service', '=', 'verification')
+      .where((qb) =>
+        qb.where('cursor', '<', cursor).orWhere('cursor', 'is', null),
+      )
```
devinivy, line 205: Nice!

lexicons/tools/ozone/verification/revokeVerifications.json
foysalit: yep yep!

packages/ozone/src/verification/issuer.ts
```diff
@@ -0,0 +1,122 @@
...
+  private agent = new Agent(this.session)
+  constructor(private verifierConfig: VerifierConfig) {}
+
+  static creator() {
+    return (verifierConfig: VerifierConfig) =>
+      new VerificationIssuer(verifierConfig)
+  }
+
+  async getAgent() {
+    if (!this.session.hasSession) {
```
foysalit, line 28: @devinivy is there any documentation i can check out to confirm if this works? i see internally, we try to refresh session if requests fail with auth token but wondering if we need to add a token persistence implementation ourselves here to make sure we dont login while there's a valid session?

packages/ozone/src/daemon/verification-listener.ts
```diff
@@ -119,6 +120,19 @@ export class VerificationListener {
...
         [this.collection]: async (e: CommitCreateEvent<VerificationRecord>) => {
+          const recordValidity = lexicons.validate(
+            this.collection,
+            e.commit.record,
+          )
+
+          if (!recordValidity.success) {
+            verificationLogger.error(
+              recordValidity.error,
+              'Invalid verification record in the fireshose',
```
devinivy, line 131:
```suggestion
              'Invalid verification record in the firehose',
```

packages/ozone/src/verification/issuer.ts
devinivy: I don't think we do need to do that—that would be for e.g. persisting session details to local storage in the browser setting.

devinivy (approved):
Good stuff!

PR #3772, opened 2025-04-17
Return verification views with empty display name values

estrattonbailey:
It's currently possible to issue verifications for users without display names (which equates to empty strings). However, we do not currently hydrate verifications without a display name, which means they are not returned to the client, and therefore the verifier cannot revoke and re-issue a verification for the user (after they update their display name).

This PR removes that check so that the client can present the verifier with the option to clear this invalid verification and thereby issue a fresh one.

### DIFF
--- a/packages/bsky/src/hydration/actor.ts
+++ b/packages/bsky/src/hydration/actor.ts
@@ -145,7 +145,6 @@ export class ActorHydrator {
         Object.entries(actor.verifiedBy),
         ([actorDid, verificationMeta]) => {
           if (
-            verificationMeta.displayName &&
             verificationMeta.handle &&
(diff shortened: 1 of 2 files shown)


packages/bsky/src/hydration/actor.ts
```diff
@@ -145,7 +145,7 @@ export class ActorHydrator {
         Object.entries(actor.verifiedBy),
         ([actorDid, verificationMeta]) => {
           if (
-            verificationMeta.displayName &&
+            typeof verificationMeta.displayName === 'string' &&
```
estrattonbailey, line 148: I think the dataplane will always return a string, but can't hurt

PR #3773, opened 2025-04-17
Verification notifications

ghost:


packages/dev-env/src/seed/client.ts
```diff
@@ -583,6 +583,19 @@ export class SeedClient<
     return this.verifications[by][subject]
   }
 
+  async unverify(by: string, subject: string) {
+    const verification = this.verifications[by][subject]
```
devinivy, line 587: In case we don't even know the verifier:
```suggestion
    const verification = this.verifications[by]?.[subject]
```

packages/bsky/tests/views/__snapshots__/notifications.test.ts.snap
```diff
@@ -1652,3 +1652,97 @@ Array [
...
+      },
+    },
+    "cid": "cids(0)",
+    "indexedAt": "1970-01-01T00:00:00.000Z",
+    "isRead": false,
+    "labels": Array [],
+    "reason": "verified",
+    "record": Object {
+      "$type": "app.bsky.notification.defs#recordDeleted",
+    },
```
devinivy, line 1697: I expected in this case that we'd include the verification record.

packages/bsky/src/views/index.ts
```diff
@@ -1426,6 +1429,14 @@ export class Views {
...
     } else if (uri.collection === ids.AppBskyGraphFollow) {
       recordInfo = state.follows?.get(notif.uri)
+    } else if (uri.collection === ids.AppBskyGraphVerification) {
+      recordInfo = {
+        record: {
+          $type: 'app.bsky.notification.defs#recordDeleted',
+        },
+        // Pre-computed CID for the dummy record above.
+        cid: 'bafyreidad6nyekfa4a67yfb573ptxiv6s7kyxyg2ra6qbbemcruadvtuim',
+      }
```
devinivy, line 1439: I might recommend pulling these values out into constants.  Also, in the case of a verification I think we'd want the details of the verification record to appear here.

devinivy (approved):

estrattonbailey (approved):

PR #3789, opened 2025-04-21
Don't use unsafe values for HTTP codes

ghost:
`xrpc-server` reused `ResponseType` from the `xrpc`, but it needed to use the response types as HTTP codes.

Not all of the `ResponseType` from `xrpc` are valid HTTP codes, so we created a new `ResponseType` enum in `xrpc-server` that only contains the valid HTTP codes.

### DIFF
--- a/packages/xrpc/src/types.ts
+++ b/packages/xrpc/src/types.ts
@@ -187,7 +187,9 @@ export class XRPCInvalidResponseError extends XRPCError {
   ) {
     super(
       ResponseType.InvalidResponse,
-      ResponseTypeNames[ResponseType.InvalidResponse],
+      // @NOTE: This is probably wrong and should use ResponseTypeNames instead.
+      // But it would mean a breaking change.
+      ResponseTypeStrings[ResponseType.InvalidResponse],
       `The server gave an invalid response and may be out of date.`,
       undefined,
       { cause: validationError },
--- a/packages/xrpc-server/tests/errors.test.ts
+++ b/packages/xrpc-server/tests/errors.test.ts
@@ -243,7 +243,7 @@ describe('Errors', () => {
       expect(e).toBeInstanceOf(XRPCError)
       expect(e).toBeInstanceOf(XRPCInvalidResponseError)
       expect(e.success).toBeFalsy()
-      expect(e.error).toBe('InvalidResponse')
+      expect(e.error).toBe('Invalid Response')
       expect(e.message).toBe(
         'The server gave an invalid response and may be out of date.',
       )


packages/xrpc/src/types.ts
```diff
@@ -24,7 +24,9 @@ export const errorResponseBody = z.object({
 export type ErrorResponseBody = z.infer<typeof errorResponseBody>
 
 export enum ResponseType {
+  // Network issue, unable to get response from the server.
```
matthieusieben, line 27: JS Doc will have better IDE integration (tooltip will contain message on hover).

```suggestion
  /**
   * Network issue, unable to get response from the server.
   */
```

packages/xrpc-server/src/types.ts
```diff
@@ -235,7 +290,7 @@ export class XRPCError extends Error {
     return {
       error: this.customErrorName ?? this.typeName,
       message:
-        this.type === ResponseType.InternalServerError
+        this.type === ClientResponseType.InternalServerError
```
matthieusieben, line 293: Since `this.type` is of type `ResponseType`, It seems more logical to compare this with `ResponseType.InternalServerError`:

```suggestion
        this.type === ResponseType.InternalServerError
```

packages/xrpc-server/src/types.ts
```diff
@@ -219,7 +219,62 @@ export type XRPCStreamHandlerConfig = {
...
+      return ResponseType.NotAcceptable
+    case ClientResponseType.PayloadTooLarge:
+      return ResponseType.PayloadTooLarge
+    case ClientResponseType.UnsupportedMediaType:
+      return ResponseType.UnsupportedMediaType
+    case ClientResponseType.RateLimitExceeded:
+      return ResponseType.RateLimitExceeded
+    case ClientResponseType.InvalidResponse:
+    case ClientResponseType.InternalServerError:
+      return ResponseType.InternalServerError
```
matthieusieben, line 246: If the upstream returns a 500, this server should return a 503

```suggestion
      return ResponseType.UpstreamFailure
```

packages/xrpc-server/src/types.ts
```diff
@@ -219,7 +219,62 @@ export type XRPCStreamHandlerConfig = {
...
+      return ResponseType.XRPCNotSupported
+    case ClientResponseType.NotAcceptable:
+      return ResponseType.NotAcceptable
+    case ClientResponseType.PayloadTooLarge:
+      return ResponseType.PayloadTooLarge
+    case ClientResponseType.UnsupportedMediaType:
+      return ResponseType.UnsupportedMediaType
+    case ClientResponseType.RateLimitExceeded:
+      return ResponseType.RateLimitExceeded
+    case ClientResponseType.InvalidResponse:
```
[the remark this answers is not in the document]
matthieusieben: An alternative that would not require mapping all these cases would be to **not** introduce a new `ResponseType` enum, an only map relevant types.

```ts

/**
 * Converts an upstream XRPC {@link ResponseType} into a down stream {@link ResponseType}.
 */
function responseTypeFromClient (type: ResponseType): ResponseType {
  switch (type) {
    case ClientResponseType.InvalidResponse:
      // Upstream server returned an XRPC response that is not compatible with our internal typings for that XRPC method
      return ResponseType.InternalServerError
    case ClientResponseType.InternalServerError:
      // Upstream encountered an internal error. We choose to proxy the 500 here instead or returning a 503, as per semantics of the XRPC protocol
      return ResponseType.InternalServerError
    case ClientResponseType.Unknown:
      // Typically a network error / unknown host
      return ResponseType.UpstreamFailure
    default:
      return type
  }
}

```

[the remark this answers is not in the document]
matthieusieben: 1) This has the advantage of being a non breaking change
2) This does **not** make the server `XRPCError`'s `type` property any less safe. (even with your change, it is still possible to do `new ServerXRPCError(-1 as number, '')` due to Typescript's handling of enums.

packages/xrpc-server/src/types.ts
```diff
@@ -221,6 +221,35 @@ export type XRPCStreamHandlerConfig = {
...
+      // Upstream server returned an XRPC response that is not compatible with our internal lexicon definitions for that XRPC method.
+      // @NOTE This could be reflected as both a 500 ("we" are at fault) and 502 ("they" are at fault). Let's be gents about it.
+      return ResponseType.InternalServerError
+    case ResponseType.InternalServerError:
+      // Upstream encountered an internal error. We choose to proxy the 500 here, instead of returning a 503, as per semantics of the XRPC protocol.
+      // @NOTE this is less ideal as logs, for example, will wrongly reflect the amount of internal vs. upstream errors.
+      return ResponseType.InternalServerError
+    case ResponseType.Unknown:
+      // Typically a network error / unknown host
+      return ResponseType.UpstreamFailure
```
matthieusieben, line 239:
```suggestion
      return ResponseType.InternalServerError
```

packages/xrpc-server/src/types.ts
```diff
@@ -221,6 +221,35 @@ export type XRPCStreamHandlerConfig = {
...
+    default:
+      // Because of a clunky behavior in TS, it is possible to call
+      // `new XRPCError(123 as number, '')` and get no error. This code
+      // acts as fool-proofing against such cases, ensuring that an actual
+      // ResponseType is always returned.
+      if (!Object.hasOwn(ResponseType, type)) {
+        return ResponseType.UpstreamFailure
+      }
+
```
matthieusieben, line 248:
```suggestion
```

packages/xrpc-server/src/types.ts
```diff
@@ -221,6 +221,23 @@ export type XRPCStreamHandlerConfig = {
...
 export { ResponseType }
 
+/**
+ * Converts an upstream XRPC {@link ResponseType} into a downstream {@link ResponseType}.
+ */
+function responseTypeFromClient(type: ResponseType): ResponseType {
+  switch (type) {
+    case ResponseType.InvalidResponse:
+      // Upstream server returned an XRPC response that is not compatible with our internal lexicon definitions for that XRPC method.
+      // @NOTE This could be reflected as both a 500 ("we" are at fault) and 502 ("they" are at fault). Let's be gents about it.
```
devinivy, line 231: 😆

devinivy (approved):

devinivy (commented):
If there's an easy way to add a test for this in xrpc-server, that would be great.

devinivy (approved):

packages/xrpc/src/types.ts
```diff
@@ -181,7 +187,7 @@ export class XRPCInvalidResponseError extends XRPCError {
   ) {
     super(
       ResponseType.InvalidResponse,
-      ResponseTypeStrings[ResponseType.InvalidResponse],
+      ResponseTypeNames[ResponseType.InvalidResponse],
```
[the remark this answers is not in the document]
devinivy: Hmm, I think that this probably was a mistake since as you pointed out, the second argument to `XRPCError` is a "name."  It would be a breaking change which ends-up affecting `@atproto/api`.  So my inclination would be to handle this change independently from the rest of this work.

devinivy (approved):

PR #3791, opened 2025-04-22
Ozone: fix correctness and perf issues with reporter stats

devinivy:
This fixes a couple issues with ozone reporter stats.  We were counting distinct accounts rather than records for takedowns and labels on `takendownRecordCount` and `labeledRecordCount`.  The join in that query also found all account actions rather than actions particular to the reported record.  In order to make the queries simpler to read I hoisted some logic up so that the queries themselves did not need to contain conditionals.  Lastly, we now handle the case where reporter stats are requested for an empty set of DIDs.


foysalit (approved):
Looks good! i reviewed the original PR and missed the numeric discrepancies in the tests 🤦🏽‍♂️

PR #3805, opened 2025-04-25
Fix format for createdAt

drasticactions:
The createdAt format field is incorrect. It should have datetime attached to it to match the other createdAt types. This broke my source generator since it was treating it as a string.


drasticactions:
Closing https://github.com/bluesky-social/atproto/pull/3829

PR #3806, opened 2025-04-25
OAuth scopes

matthieusieben:


packages/pds/src/util/http.ts
```diff
@@ -0,0 +1,15 @@
+import { ServerResponse } from 'node:http'
+
+export function appendVary(res: ServerResponse, value: string) {
+  const current = res.getHeader('Vary')
+  if (current == null || typeof current === 'number') {
+    res.setHeader('Vary', value)
+  } else {
+    const alreadyIncluded = Array.isArray(current)
+      ? current.some((value) => value.includes(value))
+      : current.includes(value)
```
devinivy: Due to the way inclusion is checked as a substring, I think this usage wont have the expected result:
```ts
appendVary(res, 'atproto-accept-labelers') // Vary: atproto-accept-labelers
appendVary(res, 'accept')                  // Vary: atproto-accept-labelers
```

packages/pds/src/oauth/oauth-scopes.ts
```diff
@@ -0,0 +1,8 @@
+export function isValidPermission(value) {
```
devinivy: I think this may be an implicit any.
```suggestion
export function isValidPermission(value: unknown) {
```

packages/oauth/oauth-provider/src/client/client-manager.ts
```diff
@@ -259,14 +259,6 @@ export class ClientManager {
       throw new InvalidClientMetadataError(`Duplicate scope "${dupScope}"`)
     }
 
-    for (const scope of scopes) {
-      // Note, once we have dynamic scopes, this check will need to be
-      // updated to check against the server's supported scopes.
-      if (!this.serverMetadata.scopes_supported?.includes(scope)) {
-        throw new InvalidClientMetadataError(`Unsupported scope "${scope}"`)
-      }
-    }
```
devinivy, line 268: What is the future of this check?

packages/pds/src/util/http.ts
```diff
@@ -0,0 +1,30 @@
...
+) {
+  switch (typeof headerValue) {
+    case 'string':
+      return separator
+        ? headerValue.split(separator).some((v) => v.trim() === searchValue)
+        : headerValue.trim() === searchValue
+    case 'number':
+      return String(headerValue) === searchValue
+    case 'object':
+      return headerValue.some((h) => headerContains(h, searchValue, separator))
```
devinivy: I see this safely checks that `headerValue` is an array thanks to type narrowing, but it was a little unintuitive to read, took me a minute!

packages/pds/src/util/http.ts
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Set or appends a value to the `Vary` header in the response, only if the
+ * value is not already present.
+ */
+export function appendVary(res: ServerResponse, value: string) {
+  if (!headerContains(res.getHeader('Vary'), value, ',')) {
+    res.appendHeader('Vary', value)
+  }
+}
```
devinivy: I expect this may work just fine in practice, but a few small things stood out to me:
 - if the vary header has the wildcard value `*` we'll still append to it.  I believe this is technically still correct though.
 - `appendHeader()` will either add a new header or concatenate an existing one using the `;` character (i.e. not `,`) depending on a `res` setting.  Should be alright so long as we're always in the former case.
 - `headerContains()` wouldn't dedupe case-insensitively even though the values in the list are case-insensitive.

packages/pds/src/auth-verifier.ts
```diff
@@ -755,32 +657,17 @@ export const parseBasicAuth = (
...
-    const alreadyIncluded = Array.isArray(current)
-      ? current.some((value) => value.includes(value))
-      : current.includes(value)
-    if (!alreadyIncluded) {
-      res.appendHeader('Vary', value)
-    }
-  }
+function setAuthHeaders(res: ServerResponse) {
+  res.setHeader('Cache-Control', 'private')
+  appendVary(res, 'Authorization')
```
devinivy, line 669: There's no need to vary over the authorization header, as it's well-specified how intermediaries should treat it.

devinivy: Ah I see we are just preserving some existing behavior, so that's fine for now 👍

packages/pds/src/util/http.ts
devinivy: I see now we're just preserving some existing behavior, so that should be fine for now 👍

packages/pds/src/permissions/permissions-oauth.ts
```diff
@@ -0,0 +1,69 @@
...
+      // @NOTE PRIVILEGED_METHODS is a superset of CHAT_BSKY_METHODS so we check
+      // it after CHAT_BSKY_METHODS.
+      return false
+    }
+    if (lxm && lxm === `${ids.AppBskyActorGetPreferences}Full`) {
+      return false
+    }
+    if (lxm && lxm === `${ids.AppBskyActorPutPreferences}Full`) {
+      return false
+    }
```
devinivy: I am still wrapping my head around the workaround we have going on here.  I think it's important logic though, could we add a comment here with a bit of context?

packages/pds/src/api/com/atproto/repo/putRecord.ts
```diff
@@ -46,6 +46,14 @@ export default function (server: Server, ctx: AppContext) {
         swapCommit,
         swapRecord,
       } = input.body
+
+      if (auth.credentials.type === 'permissions') {
+        auth.credentials.permissions.assertRepo({
+          action: 'update',
+          collection,
+        })
+      }
```
devinivy: I think we want to ensure both `'create'` and `'update'` here.

packages/pds/src/api/com/atproto/repo/deleteRecord.ts
```diff
@@ -29,6 +29,14 @@ export default function (server: Server, ctx: AppContext) {
     ],
     handler: async ({ input, auth }) => {
       const { repo, collection, rkey, swapCommit, swapRecord } = input.body
+
+      if (auth.credentials.type === 'permissions') {
+        auth.credentials.permissions.assertRepo({
+          action: 'delete',
+          collection,
+        })
+      }
```
devinivy: Just confirming my understanding—this lives in the handler rather than auth step because it relies on the `input`, is that right?

packages/pds/src/api/com/atproto/moderation/createReport.ts
```diff
@@ -1,14 +1,15 @@
...
 import { Server } from '../../../../lexicon'
 import { ids } from '../../../../lexicon/lexicons'
 import { parseProxyInfo } from '../../../../pipethrough'
 
 export default function (server: Server, ctx: AppContext) {
   server.com.atproto.moderation.createReport({
-    auth: ctx.authVerifier.accessStandard({
+    auth: ctx.authVerifier.authorization({
       additional: [AuthScope.Takendown],
+      authorize: (permissions) => permissions.assertIdentity({ report: true }),
```
devinivy: I am not 100% sure why `createReport` has its own handler rather than using service proxying 🤔 .  Huh!  In any case, I would expect the permissions to be based on rpc rather than identity.

packages/pds/src/api/com/atproto/repo/importRepo.ts
```diff
@@ -17,14 +17,20 @@ import { Server } from '../../../../lexicon'
...
 export default function (server: Server, ctx: AppContext) {
   server.com.atproto.repo.importRepo({
-    auth: ctx.authVerifier.accessFull({
+    auth: ctx.authVerifier.authorization({
       checkTakedown: true,
+      // Scope "repo:*" needed to import a repo
+      authorize: (permissions) => {
+        permissions.assertRepo({ action: '*', collection: '*' })
+      },
     }),
```
devinivy: Do we need the `ACCESS_FULL` scope here?

packages/pds/src/api/com/atproto/identity/requestPlcOperationSignature.ts
```diff
@@ -1,12 +1,18 @@
...
 export default function (server: Server, ctx: AppContext) {
   server.com.atproto.identity.requestPlcOperationSignature({
-    auth: ctx.authVerifier.accessFull({ additional: [AuthScope.Takendown] }),
+    auth: ctx.authVerifier.authorization({
+      // @NOTE Reflect any change in signPlcOperation
+      additional: [AuthScope.Takendown],
+      authorize: (permissions) => {
+        permissions.assertIdentity({ plcOpRequest: true })
+      },
+    }),
```
devinivy: Do `accessFull()` need to be converted to use the `ACCESS_FULL` scope?  If so, I believe there are a handful of these to take a look at.

packages/pds/src/api/app/bsky/notification/registerPush.ts
```diff
@@ -11,15 +11,18 @@ export default function (server: Server, ctx: AppContext) {
...
   if (!bskyAppView) return
 
+  const aud = `${bskyAppView.did}#bsky_appview`
+
   server.app.bsky.notification.registerPush({
-    auth: ctx.authVerifier.accessStandard({
+    auth: ctx.authVerifier.authorization({
       additional: [AuthScope.SignupQueued],
+      authorize: (permissions) => {
+        permissions.assertRpc({ aud, lxm: ids.AppBskyNotificationRegisterPush })
```
devinivy: Taking a closer look below, I think the audience check should be based on `input.body.serviceDid`.  This endpoint predated the generic service proxying, but it's the same idea.

packages/pds/src/api/app/bsky/feed/getTimeline.ts
```diff
@@ -8,10 +9,18 @@ import {
...
   server.app.bsky.feed.getTimeline({
-    auth: ctx.authVerifier.accessStandard(),
+    auth: ctx.authVerifier.authorization({
+      authorize: (permissions) => {
+        permissions.assertRpc({
+          aud: `${bskyAppView.did}#bsky_appview`,
+          lxm: ids.AppBskyFeedGetTimeline,
+        })
+      },
+    }),
```
devinivy: I believe the audience here should be based on service proxying info—the requests don't necessarily go to our appview.

packages/pds/src/api/app/bsky/feed/getPostThread.ts
```diff
@@ -25,10 +25,18 @@ import {
...
-  if (!ctx.bskyAppView) return
+  const { bskyAppView } = ctx
+  if (!bskyAppView) return
 
   server.app.bsky.feed.getPostThread({
-    auth: ctx.authVerifier.accessStandard(),
+    auth: ctx.authVerifier.authorization({
+      authorize: (permissions) => {
+        permissions.assertRpc({
+          aud: `${bskyAppView.did}#bsky_appview`,
```
devinivy: I believe the audience here and should be based on service proxying info.

packages/pds/src/api/app/bsky/feed/getFeed.ts
```diff
@@ -10,7 +10,14 @@ export default function (server: Server, ctx: AppContext) {
...
   server.app.bsky.feed.getFeed({
-    auth: ctx.authVerifier.accessStandard(),
+    auth: ctx.authVerifier.authorization({
+      authorize: (permissions) => {
+        permissions.assertRpc({
+          aud: `${bskyAppView.did}#bsky_appview`,
+          lxm: ids.AppBskyFeedGetFeed,
+        })
+      },
+    }),
```
devinivy: I believe the audience here and should be based on service proxying info.

packages/pds/src/api/app/bsky/feed/getAuthorFeed.ts
```diff
@@ -9,10 +10,18 @@ import {
...
-  if (!ctx.bskyAppView) return
+  const { bskyAppView } = ctx
+  if (!bskyAppView) return
 
   server.app.bsky.feed.getAuthorFeed({
-    auth: ctx.authVerifier.accessStandard(),
+    auth: ctx.authVerifier.authorization({
+      authorize: (permissions) => {
+        permissions.assertRpc({
+          aud: `${bskyAppView.did}#bsky_appview`,
```
devinivy: (Just going to leave a comment on each of the read-after-write endpoint for good measure even though it's the same issue.)

I believe the audience here and should be based on service proxying info.

packages/pds/src/api/app/bsky/feed/getActorLikes.ts
```diff
@@ -8,10 +9,18 @@ import {
...
-  if (!ctx.bskyAppView) return
+  const { bskyAppView } = ctx
+  if (!bskyAppView) return
 
   server.app.bsky.feed.getActorLikes({
-    auth: ctx.authVerifier.accessStandard(),
+    auth: ctx.authVerifier.authorization({
+      authorize: (permissions) => {
+        permissions.assertRpc({
+          aud: `${bskyAppView.did}#bsky_appview`,
```
devinivy: I believe the audience here and should be based on service proxying info.

packages/pds/src/api/app/bsky/actor/putPreferences.ts
```diff
@@ -19,12 +30,20 @@ export default function (server: Server, ctx: AppContext) {
...
       }
+
+      // @NOTE This is a "hack" that uses a fake lxm to allow for full access
+      const fullAccess =
+        auth.credentials.type === 'access'
+          ? auth.credentials.scope === AuthScope.Access
+          : auth.credentials.permissions.allowsRpc({
+              aud: `${bskyAppView.did}#bsky_appview`,
+              lxm: `${ids.AppBskyActorPutPreferences}Full`,
+            })
```
devinivy: I think I would expect this to use the same logic as including the `scopes: FULL_ACCESS` option of `authVerifier.authorization()`.  If that's the case I believe it would end-up being something like:
```ts
const fullAccess =
  auth.credentials.type === 'access' &&
  FULL_ACCESS.includes(auth.credentials.scope)
```

packages/pds/src/api/app/bsky/actor/getProfiles.ts
```diff
@@ -8,10 +9,18 @@ import {
...
-  if (!ctx.bskyAppView) return
+  const { bskyAppView } = ctx
+  if (!bskyAppView) return
 
   server.app.bsky.actor.getProfiles({
-    auth: ctx.authVerifier.accessStandard(),
+    auth: ctx.authVerifier.authorization({
+      authorize: (permissions) => {
+        permissions.assertRpc({
+          aud: `${bskyAppView.did}#bsky_appview`,
```
devinivy: I believe the audience here and should be based on service proxying info.

packages/pds/src/api/app/bsky/actor/getProfile.ts
```diff
@@ -8,10 +9,18 @@ import {
...
-  if (!ctx.bskyAppView) return
+  const { bskyAppView } = ctx
+  if (!bskyAppView) return
 
   server.app.bsky.actor.getProfile({
-    auth: ctx.authVerifier.accessStandard(),
+    auth: ctx.authVerifier.authorization({
+      authorize: (permissions) => {
+        permissions.assertRpc({
+          aud: `${bskyAppView.did}#bsky_appview`,
```
devinivy: I believe the audience here and should be based on service proxying info.

packages/pds/src/api/com/atproto/server/getServiceAuth.ts
```diff
@@ -1,31 +1,34 @@
...
 import { Server } from '../../../../lexicon'
 import { ids } from '../../../../lexicon/lexicons'
 import { PRIVILEGED_METHODS, PROTECTED_METHODS } from '../../../../pipethrough'
 
 export default function (server: Server, ctx: AppContext) {
   server.com.atproto.server.getServiceAuth({
-    auth: ctx.authVerifier.accessStandard({
+    auth: ctx.authVerifier.authorization({
       additional: [AuthScope.Takendown],
+      authorize: (permissions, { params }) => permissions.assertRpc(params),
```
devinivy: This is cool!

packages/pds/src/api/com/atproto/server/getServiceAuth.ts
```diff
@@ -50,10 +53,17 @@ export default function (server: Server, ctx: AppContext) {
...
         }
-        if (!auth.credentials.isPrivileged && PRIVILEGED_METHODS.has(lxm)) {
-          throw new InvalidRequestError(
-            `insufficient access to request a service auth token for the following method: ${lxm}`,
-          )
+
+        if (
+          auth.credentials.type === 'access' &&
+          auth.credentials.scope !== AuthScope.Access &&
+          auth.credentials.scope !== AuthScope.AppPassPrivileged
```
devinivy: Could use the `ACCESS_PRIVILEGED` constant here.

packages/pds/src/api/com/atproto/server/getSession.ts
```diff
@@ -65,23 +63,16 @@ export default function (server: Server, ctx: AppContext) {
...
-
-    default:
-      // @ts-expect-error
-      throw new Error(`Unknown credentials type: ${credentials.type}`)
+  if (
+    credentials.type === 'permissions' &&
+    !credentials.permissions.allowsAccount({ email: true })
+  ) {
+    const { email, emailAuthFactor, emailConfirmed, ...rest } = data
+    return rest
```
devinivy: Nice 👍

packages/pds/src/api/com/atproto/server/getSession.ts
```diff
@@ -2,27 +2,25 @@ import { ComAtprotoServerGetSession } from '@atproto/api'
...
-                req,
-                auth.credentials.did,
-                'com.atproto.server.getSession',
-              )
-            : ctx.entrywayPassthruHeaders(req)
+        const headers = await ctx.entrywayAuthHeaders(
+          req,
+          auth.credentials.did,
+          'com.atproto.server.getSession',
+        )
```
devinivy, line 25: This is a bit consequential—will want to remember this change is in here when we release.  Do you happen to remember why this wasn't changed before?  Just want to make sure we're not missing/forgetting something.

packages/pds/src/pipethrough.ts
```diff
@@ -50,13 +56,21 @@ export const proxyHandler = (ctx: AppContext): CatchallHandler => {
...
+      const { url: origin, did: aud } = await parseProxyInfo(ctx, req, lxm)
+
+      const authResult = await performAuth({ req, res, params: { lxm, aud } })
+
+      const { credentials } = excludeErrorResult(authResult)
+
+      if (
+        credentials.type === 'access' &&
+        credentials.scope !== AuthScope.Access &&
+        credentials.scope !== AuthScope.AppPassPrivileged &&
```
devinivy: Could use `ACCESS_PRIVILEGED` const here.

packages/pds/src/auth-verifier.ts
```diff
@@ -143,360 +112,291 @@ export class AuthVerifier {
...
-        ctx,
-        [
-          AuthScope.Access,
-          AuthScope.AppPassPrivileged,
-          AuthScope.AppPass,
-          ...(opts.additional ?? []),
-        ],
-        opts,
-      )
+  public null: MethodAuthVerifier<NullOutput> = (ctx) => {
```
devinivy: `public null` really tickles me 😆 I think it can be a little confusing to encounter though, and using the keyword does impact its usage.  E.g. you could destructure `{ adminToken }` but not `{ null }` from the auth verifier.

devinivy: Would something like `none()` or `missing()` do?

devinivy: Ah, I see now it was already called this!  (I couldn't tell from the diff.)  I think just seeing it divorced from the method parens (`protected null()` vs `public null:`) made it look more like it was being used as a keyword.  Would still recommend we consider renaming, but all good if we don't address that as part of this work 👍

packages/pds/src/auth-verifier.ts
```diff
@@ -510,99 +410,122 @@ export class AuthVerifier {
...
+      .catch((cause) => {
+        throw cause instanceof jose.errors.JWTExpired
+          ? new InvalidRequestError('Token has expired', 'ExpiredToken', {
+              cause,
+            })
+          : new InvalidRequestError(
+              'Token could not be verified',
+              'InvalidToken',
+              { cause },
+            )
```
devinivy: I think we can make this a little more accessible / commonplace:
```suggestion
        if (cause instanceof jose.errors.JWTExpired) {
          throw new InvalidRequestError('Token has expired', 'ExpiredToken', {
            cause,
          })
        } else {
          throw new InvalidRequestError(
            'Token could not be verified',
            'InvalidToken',
            { cause },
          )
        }
```

packages/pds/src/auth-verifier.ts
```diff
@@ -510,99 +410,122 @@ export class AuthVerifier {
...
-        did,
-        scope,
-        isPrivileged,
-      },
+    const { sub, aud, scope, lxm, cnf, jti } = payload
+
+    if (typeof lxm !== 'undefined') {
+      // Service auth tokens are not allowed here
+      throw new InvalidRequestError('Malformed token', 'InvalidToken')
+    }
```
devinivy: I wonder if there's a better name out there for the `verifyBearerJwt()` method, since of course service auth tokens are also bearer jwts.  Hmmmm, nothing quickly to mind 🤔

devinivy: P.S. just confirming—this check is defensive and not strictly necessary in addition to the `typ` check for service auth tokens.

packages/pds/src/auth-verifier.ts
```diff
@@ -143,360 +112,291 @@ export class AuthVerifier {
...
+  }
 
-    // https://datatracker.ietf.org/doc/html/rfc9449#section-8.2
-    if (res) {
+  public oauth<P extends Params>(
+    verifyStatusOptions: VerifiedOptions = {},
+  ): MethodAuthVerifier<OAuthOutput, P> {
+    const verifyTokenOptions: VerifyTokenClaimsOptions = {
+      audience: [this.dids.pds],
+      scope: ['atproto'],
```
devinivy, line 339: I may be misremembering, but I expected that `authenticateRequest()` would always require this scope.  In the previous it looks like this wasn't passed—is there a behavior change or just being more explicit?

packages/pds/src/auth-verifier.ts
```diff
@@ -732,12 +627,19 @@ const isBasicToken = (req: IncomingMessage): boolean => {
   return type === AuthType.BASIC
 }
 
+const isUserServiceAuth = (req: IncomingMessage): boolean => {
+  const token = bearerTokenFromReq(req)
+  if (!token) return false
+  const payload = jose.decodeJwt(token)
+  return payload['lxm'] != null
+}
```
devinivy: Ah, I really wish we set the `typ` from the beginning!  Technically `lxm` isn't required in service auth tokens right now, but it is required for the purposes used for here (which I believe ends-up only being `uploadBlob`).  Maybe we just add a comment to clarify and leave the implementation as-is?

packages/oauth/oauth-provider/src/client/client-manager.ts
devinivy: I believe I found this, essentially moved into the AS's `getClientInfo()`.  Should we add a changeset for this?

devinivy (commented):
Well-wrangled!  I think you struck a pretty nice balance here between the new permissions system and existing system with all its little exceptions— I think we can leave it in this state for now 👍 

There are a lot of comments but many are duplicates, questions I ended up answering for myself during review, or ended-up being about existing code that I discovered was just moved around.  So I don't think it should lead to very many changes, which is good!

One question I had is that I noticed no tests were added during the development process.  On one hand that makes sense since functionality shouldn't have really changed.  On the other hand, I wonder if you feel like you've hit the level of confidence in these changes that you were hoping for, without additional testing.  Authorization is sensitive functionality and this is the biggest refactor that has happened around authorization for some time, so figure it's worth asking / looking into.

packages/oauth/oauth-provider/src/client/client-manager.ts
matthieusieben: The main reason here is that we can't possibly list every single possible scope supported by the server. I've added a changeset

packages/pds/src/util/http.ts
matthieusieben: Simplified the code a bit

packages/pds/src/api/com/atproto/repo/deleteRecord.ts
matthieusieben: Yep

matthieusieben: (added a comment)

packages/pds/src/api/com/atproto/server/getSession.ts
matthieusieben: I think it was not changed before as a precautionary measures, to avoid introducing breaks on existing code path.

packages/pds/src/api/com/atproto/server/getServiceAuth.ts
matthieusieben: Was a lot of efforts just for this :D

packages/pds/src/api/com/atproto/server/getSession.ts
matthieusieben: And also to allow for a smooth deployment as the support for service auth on this endpoint was added to entryway in parallel.

packages/pds/src/auth-verifier.ts
matthieusieben: renamed to `unauthenticated`

packages/pds/src/auth-verifier.ts
matthieusieben: Just being more explicit

packages/pds/src/auth-verifier.ts
matthieusieben: The `typ` check is disabled if the header does not have a `typ` claim, which is the case of service auth tokens. This is mainly to protect against the hypotetical case where a PDS would sign tokens without `typ` with its private key somehow.

packages/pds/src/auth-verifier.ts
matthieusieben: Note that this does not introduce any change.

packages/pds/src/auth-verifier.ts
matthieusieben: Renamed to `unauthenticated`

.changeset/thick-zebras-roll.md
```diff
@@ -0,0 +1,5 @@
+---
+"@atproto/oauth-provider": patch
+---
+
+The scopes allowed in the client metadata dowument is no longer constraint by the `scopes_supported` in the server metadata.
```
devinivy:
```suggestion
The scopes allowed in the client metadata document is no longer constrained by the `scopes_supported` in the server metadata.
```

packages/pds/src/api/app/bsky/feed/getFeed.ts
```diff
@@ -3,19 +3,18 @@ import { AtUri } from '@atproto/syntax'
...
     auth: ctx.authVerifier.authorization({
-      authorize: (permissions) => {
-        permissions.assertRpc({
-          aud: `${bskyAppView.did}#bsky_appview`,
-          lxm: ids.AppBskyFeedGetFeed,
-        })
+      authorize: (permissions, { req }) => {
+        const lxm = ids.AppBskyFeedGetFeed
+        const aud = computeProxyTo(ctx, req, lxm)
+        permissions.assertRpc({ aud, lxm })
```
devinivy, line 17: I believe we want to check the RPC call to the feed generator here as well.  It's an awkward case.

packages/pds/src/api/app/bsky/notification/registerPush.ts
```diff
@@ -11,19 +11,27 @@ export default function (server: Server, ctx: AppContext) {
...
+        // assert permissions here.
       },
     }),
     handler: async ({ auth, input }) => {
       const { serviceDid } = input.body
       const { did } = auth.credentials
 
+      if (auth.credentials.type === 'permissions') {
+        auth.credentials.permissions.assertRpc({
+          aud: `${serviceDid}#bsky_appview`,
```
devinivy: This is another funny case— you'll find the `getEndpoint()` function used in here actually uses the DID's `#bsky_notif` service.  I think for now we should be consistent with that.  Ideally we fix this by migrating the endpoint to use service proxying.
```suggestion
          aud: `${serviceDid}#bsky_notif`,
```

packages/pds/src/api/com/atproto/repo/createRecord.ts
```diff
@@ -17,6 +17,9 @@ export default function (server: Server, ctx: AppContext) {
     auth: ctx.authVerifier.authorization({
       checkTakedown: true,
       checkDeactivated: true,
+      authorize: () => {
+        // always allow
```
devinivy: This one isn't always allowed, rather checked below based on `input` which contains the collection.

devinivy:
```suggestion
        // Performed in the handler as it requires the request body
```

packages/pds/src/api/com/atproto/repo/uploadBlob.ts
```diff
@@ -8,7 +8,9 @@ export default function (server: Server, ctx: AppContext) {
   server.com.atproto.repo.uploadBlob({
     auth: ctx.authVerifier.authorizationOrUserServiceAuth({
       checkTakedown: true,
-      authorize: (permissions) => permissions.assertBlob(),
+      authorize: () => {
+        // Checked in the handler, as it requires the mime type
```
devinivy: It would be great if we could validate that they at least have some access to upload blobs before performing an upload.  Maybe it wouldn't be so bad to sniff the mime before initiating the upload— all it does is peek the first several bytes of the stream.

packages/pds/src/auth-verifier.ts
```diff
@@ -112,16 +112,25 @@ export class AuthVerifier {
...
-  public null: MethodAuthVerifier<NullOutput> = (ctx) => {
+  public unauthenticated: MethodAuthVerifier<UnauthenticatedOutput> = (ctx) => {
     setAuthHeaders(ctx.res)
+
+    // @NOTE this auth method is typically used as fallback when no other auth
+    // method is applicable. This means that the presence of an "authorization"
+    // header means that that header is invalid (as it did not match any of the
+    // other auth methods).
+    if (ctx.req.headers['authorization']) {
+      throw new AuthRequiredError('Invalid authorization header', 'AuthMissing')
```
devinivy:
```suggestion
      throw new AuthRequiredError('Invalid authorization header')
```

packages/pds/src/permissions/scope-syntax.ts
```diff
@@ -0,0 +1,228 @@
...
+ * The syntax is defined as follows:
+ * ```nbf
+ * scope := resource [':' positional] ['?' params]
+ * params := param ['&' param]*
+ * param := name '=' value
+ * ```
+ * Where "positional" can be used as short-hand (i.e. not used in combination
+ * with) for a specific parameter.
+ */
+export class ParsedResourceScope<R extends string = string> {
```
devinivy: This is both load bearing and straightforward to test—we should get some unit tests going in here.

packages/pds/src/permissions/scope-syntax.ts
```diff
@@ -0,0 +1,228 @@
...
+    const colonIdx = scope.indexOf(':')
+
+    const resourceEnd = minIdx(paramIdx, colonIdx)
+
+    const resource = (
+      resourceEnd !== -1 ? scope.slice(0, resourceEnd) : scope
+    ) as R
+    const positional =
+      colonIdx !== -1
+        ? // There is a positional parameter, extract it
```
devinivy: The colon alternately may have been part of the query params.

packages/pds/src/permissions/scope-syntax.ts
```diff
@@ -0,0 +1,228 @@
...
+        ? // There is a (non-empty) query string, parse it
+          new URLSearchParams(scope.slice(paramIdx + 1))
+        : undefined
+
+    return new ParsedResourceScope(resource, positional, params)
+  }
+}
+
+const minIdx = (a: number, b: number): number =>
+  a === -1 ? b : b === -1 ? a : Math.min(a, b)
```
devinivy: We can make this a bit easier to read while still being fairly terse.
```suggestion
const minIdx = (a: number, b: number): number => {
  if (a === -1) return b
  if (b === -1) return a
  return Math.min(a, b)
}
```

packages/pds/src/permissions/scope-syntax.ts
```diff
@@ -0,0 +1,228 @@
...
+
+/**
+ * Format a scope string for a resource with parameters
+ *
+ * @param resource - The resource name (e.g. `rpc`, `repo`, etc.)
+ * @param params - The list of parameters. The first parameter will be formatted
+ * as a positional parameter, if possible (if ti has only one value).
```
devinivy:
```suggestion
 * as a positional parameter, if possible (if it has only one value).
```

devinivy: Two thoughts on this— first, a given resource may not necessarily support a positional param but would have no great way to opt out here.  Second, I think the first param being implicitly positional may make it less clear what's going on when reading code on the caller side.

Perhaps we can pass all the params and optionally name which one is positional—this would make the usage explicit and would also allow us to make positional params optional for the resource.

devinivy: Maybe not a big deal since the function is used internally, but it is able to generate invalid scopes when the first param name is used twice.

packages/pds/src/permissions/resources/identity-scope.ts
```diff
@@ -0,0 +1,63 @@
...
+export type IdentityScopeMatch = {
+  feature: IdentityScopeFeatures
+}
+
+export class IdentityScope {
+  constructor(
+    public readonly features?: readonly [
+      IdentityScopeFeatures,
+      ...IdentityScopeFeatures[],
+    ],
```
devinivy: We can use `NeRoArray<IdentityScopeFeatures>` here for consistency.

packages/pds/src/permissions/resources/account-scope.ts
```diff
@@ -0,0 +1,53 @@
...
+const ALLOWED_PARAMS = Object.freeze(['feature'] as const)
+
+export class AccountScope {
+  constructor(public readonly features: NeRoArray<AccountFeature>) {}
+
+  matches(options: AccountScopeMatch): boolean {
+    if (!options.feature) return true
+    if (!this.features) return false
+    return this.features.includes(options.feature)
+  }
```
devinivy: There are a few differences from `identity`'s handling of its `feature` param—is that intentional?  This looks like it thinks `this.features` is optional, but it is always present.  If it were missing then there would be no match— whereas the `identity` resource makes `this.features` optional, and when it's missing it matches permissively.

packages/pds/src/permissions/resources/blob-scope.ts
```diff
@@ -0,0 +1,67 @@
...
+    value.includes('/') &&
+    !value.startsWith('/') &&
+    !value.endsWith('/') &&
+    !value.includes(' ')
+  )
+}
+
+const ALLOWED_PARAMS = Object.freeze(['accept'] as const)
+
+function matchAccept(accept: Accept, mime: string): boolean {
```
devinivy: We don't have to do it now, but we should eventually unify this logic with the accept validation of blobs (found in `packages/pds/src/actor-store/blob/transactor.ts`).

packages/pds/src/permissions/resources/account-scope.ts
matthieusieben: Adapted to make them more coherent

packages/oauth/oauth-provider-ui/src/components/utils/scope-description.tsx
```diff
@@ -0,0 +1,529 @@
...
+  if (permissions.scopes.some((s) => RpcScope.fromString(s) != null)) {
+    return (
+      <DescriptionCard
+        role="listitem"
+        image={<AuthenticateIcon className="size-6" />}
+        title={t`Authenticate`}
+        description={t`Perform authenticated actions on your behalf`}
+      >
+        <Trans>
+          The ATproto network uses authenticated inter service communications to
```
devinivy: To be consistent with atproto.com docs, "atproto" or "ATProto":
```suggestion
          The ATProto network uses authenticated inter service communications to
```

PR #3818, opened 2025-04-29
Remove dependency on the Public Suffix List

matthieusieben:
The PSL is currently used for two purposes:

- check that the `client_uri` from a client metadata document is a valid internet URL
- check that any `fetch()` performed by the server using user provided urls is performed against a valid URL.

The first check is not necessary as it's not part of the Atproto OAuth spec.
The second use is not necessary as the code still ensures that the IP is a unicast IP.

A basic heuristic was left in place to easily detect attempts to access local addresses.


packages/internal/fetch-node/src/unicast.ts
```diff
@@ -183,12 +182,13 @@ export function unicastLookup(
...
 
-// see lupomontero/psl#258 for context on psl usage.
-// in short, this ensures a structurally valid domain
-// plus a "listed" tld.
-function isValidDomain(domain: string) {
-  const parsed = pslParse(domain)
-  return !parsed.error && parsed.listed
+export function isInternetDomain(domain: string): boolean {
+  for (const tld of ['.test', '.local', '.localhost', '.invalid', '.example']) {
+    if (domain.endsWith(tld)) {
```
devinivy, line 187: Is case sensitivity an issue here?

devinivy (approved):
A couple questions, but looks good.

packages/internal/fetch-node/src/unicast.ts
matthieusieben: might be. Updated.

PR #3819, opened 2025-04-29
Fix potential memory leak

matthieusieben:
In the `unicastFetchWrap`, if the `didLookup` is false, the function will throw without canceling the response body, causing a memory leak. This change fixes that.

### DIFF
--- a/packages/internal/fetch-node/src/unicast.ts
+++ b/packages/internal/fetch-node/src/unicast.ts
@@ -115,25 +115,20 @@ export function unicastFetchWrap<C = FetchContext>({
             },
           })
 
-          const headers = new Headers(init?.headers)
-          headers.set('connection', 'close') // Proactively close the connection
-
           try {
-            return await fetch.call(this, input, {
+            const headers = new Headers(init?.headers)
+            headers.set('connection', 'close') // Proactively close the connection
+
+            const response = await fetch.call(this, input, {
               ...init,
               headers,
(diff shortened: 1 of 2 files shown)


packages/internal/fetch-node/src/unicast.ts
```diff
@@ -115,25 +115,20 @@ export function unicastFetchWrap<C = FetchContext>({
...
-          } finally {
-            // Free resources (we cannot await here since the response was not
-            // consumed yet).
-            void dispatcher.close().catch((err) => {
-              // No biggie, but let's still log it
-              console.warn('Failed to close dispatcher', err)
-            })
 
             if (!didLookup) {
+              await response.body?.cancel()
```
devinivy, line 130: Do you know why in the undici examples they drain the body with `for await ...` rather than `cancel()`? https://undici.nodejs.org/#/?id=garbage-collection

devinivy: In the docs they do say to "consume or cancel" so I am sure this is fine, just a bit odd they don't use `cancel()` since it's so much more direct!

devinivy (approved):

packages/internal/fetch-node/src/unicast.ts
matthieusieben: consuming will keep the socket open (for future re-use). My guess is that `cancel()` causes the socket to be closed, which is what we actually want.

PR #3820, opened 2025-04-29
Allow proxying of `getSession` using OAuth session

matthieusieben:


packages/pds/src/api/com/atproto/server/getSession.ts
```diff
@@ -14,12 +14,25 @@ export default function (server: Server, ctx: AppContext) {
...
     handler: async ({ auth, req }) => {
       if (ctx.entrywayAgent) {
-        return resultPassthru(
-          await ctx.entrywayAgent.com.atproto.server.getSession(
-            undefined,
-            ctx.entrywayPassthruHeaders(req),
-          ),
+        // Allow proxying of dpop bound requests by using service auth instead
+        const headers =
+          req.headers.authorization?.split(' ')[0].toLowerCase() === 'dpop'
```
devinivy, line 19: This check here is a bit dense.  Could we break it up a little, or perhaps use `credentials.type` if that's able to do the trick.

packages/pds/src/api/com/atproto/server/getSession.ts
```diff
@@ -37,16 +50,38 @@ export default function (server: Server, ctx: AppContext) {
 
       return {
         encoding: 'application/json',
-        body: {
+        body: render(auth, {
```
devinivy, line 53: At a glance the term "render" made me wonder if some other representation was being constructed, e.g. html.  Not sure if others would have the same association, though.  I also like terms "presentation" (a term used in the appview) or "output" (from lexicon parlance).

devinivy (approved):

packages/pds/src/api/com/atproto/server/getSession.ts
```diff
@@ -14,12 +14,25 @@ export default function (server: Server, ctx: AppContext) {
...
-            ctx.entrywayPassthruHeaders(req),
-          ),
+        // Allow proxying of dpop bound requests by using service auth instead
+        const headers =
+          req.headers.authorization?.split(' ')[0].toLowerCase() === 'dpop'
+            ? await ctx.entrywayAuthHeaders(
+                req,
+                auth.credentials.did,
+                'com.atproto.server.getSession',
+              )
```
devinivy, line 24: Since `com.atproto.server.getSession` is proxied only administratively, we may want to add it to `PROTECTED_METHODS` to avoid other usage: https://github.com/bluesky-social/atproto/blob/f476003709d43b5e2474218cd48a9e1d7ebf3089/packages/pds/src/pipethrough.ts#L493-L497

packages/pds/src/api/com/atproto/server/getSession.ts
matthieusieben: 💯

PR #3824, opened 2025-05-01
Actor status lexicon and views

ghost:


lexicons/app/bsky/actor/status.json
```diff
@@ -0,0 +1,37 @@
...
+            "description": "The duration of the status in minutes. Applications can choose to limit the duration.",
+            "minimum": 1
+          },
+          "createdAt": { "type": "string", "format": "datetime" }
+        }
+      }
+    },
+    "live": {
+      "type": "token",
+      "description": "Advertises and account as currently offering live content."
```
fry69, line 34: Typo? and -> an

packages/bsky/src/data-plane/server/indexing/plugins/status.ts
```diff
@@ -0,0 +1,62 @@
...
+): PluginType => {
+  const processor = new RecordProcessor(db, background, {
+    lexId,
+    insertFn,
+    findDuplicate,
+    deleteFn,
+    notifsForInsert,
+    notifsForDelete,
+  })
+  processor.assertValidRecord = () => null
```
devinivy, line 58: I think we probably want to get rid of this, as we only want to index valid status records.  I assume this came from the chat-declaration plugin, which has a todo to use lexicon validation—maybe that plugin was written before the lexicon was finalized, or something like that.

packages/bsky/src/data-plane/server/routes/profile.ts
```diff
@@ -102,6 +114,7 @@ export default (db: Database): Partial<ServiceImpl<typeof Service>> => ({
         priorityNotifications: row?.priorityNotifs ?? false,
         trustedVerifier: row?.trustedVerifier ?? false,
         verifiedBy,
+        statusRecord: status,
```
devinivy, line 117: Perhaps we could add the new tag fields here, mostly for self-documentation purposes.

packages/bsky/src/views/index.ts
```diff
@@ -453,6 +455,33 @@ export class Views {
...
+    const maxDuration = 3 * HOUR
+    const expiresAtMs = record.durationMinutes
+      ? sortedAt.getTime() +
+        Math.min(record.durationMinutes * MINUTE, maxDuration)
+      : undefined
+    const expiresAt = expiresAtMs
+      ? new Date(expiresAtMs).toISOString()
+      : undefined
+
+    const isActive = expiresAtMs ? expiresAtMs > Date.now() : undefined
```
devinivy, line 474: I don't think we should necessarily do this as part of this work, but I am wondering if we should eventually add the current time into `HydrationState` so that all time-dependent view logic consistently hinges off the same time.

packages/bsky/tests/views/__snapshots__/profile.test.ts.snap
```diff
@@ -469,3 +469,55 @@ Object {
   },
 }
 `;
+
+exports[`pds profile views status returns active status when within the duration 1`] = `
+Object {
+  "embed": Object {
+    "$type": "app.bsky.embed.external#external",
```
devinivy, line 476: This is supposed to be an `app.bsky.embed.external#view`.  I don't think that `app.bsky.embed.external#external` exists, actually!

packages/bsky/src/views/index.ts
```diff
@@ -453,6 +455,33 @@ export class Views {
...
+
+    return {
+      record: record,
+      status: record.status,
+      embed: record.embed,
```
devinivy, line 479: Going to want to transform this into the external embed view, so that it contains info useful to the client such as CDN urls.

devinivy (commented):
Just a couple final thoughts, otherwise looking good to me.

packages/bsky/tests/views/profile.test.ts
```diff
@@ -225,6 +227,154 @@ describe('pds profile views', () => {
     })
   })
 
+  describe('status', () => {
+    const embed: AppBskyEmbedExternal.External = {
+      $type: 'app.bsky.embed.external#external',
```
devinivy, line 232: My understanding is that this is intended to be a `AppBskyEmbedExternal.Main` with `$type: 'app.bsky.embed.external'`.

devinivy (approved):

PR #3825, opened 2025-05-01
Add `app.bsky.unspecced.getPostThreadV2`

ghost:
This starts to move the thread-building logic to the backend.

Notable changes:
1. This does fewer intermediate steps than the current client logic.
2. This annotates OP threads in the same tree-building pass, while on the client logic it does a separate pass.
3. This introduces `branchingFactor`. The idea is that currently we return "all" (within the lexicon limit) replies to the anchor post, but inside those (and further down) we trim to `branchingFactor`. This is flexible enough to allow us to implement linear view (branching factor `1`) and tree view (probably a large branching factor) on the client while fetching minimal data for each case.
	1. When we start introducing pagination, a new param will be needed (e.g., `limit` or `pageSize`).
4. I reduced the maximum values in the lexicons for `above` and `below` to prevent abuse.

### DIFF
--- a/.changeset/wicked-olives-live.md
+++ b/.changeset/wicked-olives-live.md
@@ -1,0 +1,5 @@
+---
+"@atproto/api": patch
+---
+
+Add app.bsky.unspecced.getPostThreadHiddenV2


packages/bsky/src/views/index.ts
```diff
@@ -1111,6 +1131,462 @@ export class Views {
...
+    const hasReplies = (postView.replyCount ?? 0) > 0
+    const hiddenByThreadgate =
+      viewer !== authorDid &&
+      this.replyIsHiddenByThreadgate(uri, rootUri, state)
+    const mutedByViewer = this.viewerMuteExists(authorDid, state)
+
+    type Annotation = [boolean, ThreadItemPost['annotations'][number]]
+    const annotations: Annotation[] = [
+      [hasReplies, 'app.bsky.unspecced.getPostThreadV2#hasReplies'],
+      [
```
estrattonbailey, line 1512: Drive-by thought: it'd be great to have `isAnchor` here too, esp if we ever change depth calculation on the backend.

lexicons/app/bsky/unspecced/getPostThreadV2.json
```diff
@@ -0,0 +1,144 @@
...
+            "default": 80,
+            "minimum": 0,
+            "maximum": 100
+          },
+          "below": {
+            "type": "integer",
+            "description": "How many levels of replies to include below the anchor.",
+            "default": 6,
+            "minimum": 0,
+            "maximum": 100
```
devinivy, line 29: In my view this max for `below` could be much lower.  And even more so if it's implicit that self-threads get special treatment.

lexicons/app/bsky/unspecced/getPostThreadV2.json
```diff
@@ -0,0 +1,144 @@
...
+            "format": "at-uri",
+            "description": "Reference (AT-URI) to post record."
+          },
+          "above": {
+            "type": "integer",
+            "description": "How many levels of parent (and grandparent, etc) to include above the anchor.",
+            "default": 80,
+            "minimum": 0,
+            "maximum": 100
+          },
```
devinivy, line 23: I'm still a little skeptical we need this parameter at all, or that if we do need it, it could be a boolean to include all ancestors (up to an app-defined limit).

lexicons/app/bsky/unspecced/getPostThreadV2.json
```diff
@@ -0,0 +1,144 @@
...
+            "default": 10,
+            "minimum": 0,
+            "maximum": 100
+          },
+          "prioritizeFollowedUsers": {
+            "type": "boolean",
+            "description": "Whether to prioritize posts from followed users. It only has effect when the user is authenticated.",
+            "default": false
+          },
+          "sorting": {
```
devinivy, line 43: I think in other APIs that allow sorting we use the terms `sort` or `sortBy`.  Would recommend keeping with one of those—probably `sort` since that's the one common to `app.bsky` lexicons.

lexicons/app/bsky/unspecced/getPostThreadV2.json
```diff
@@ -0,0 +1,144 @@
...
+            "type": "string",
+            "description": "Sorting for the thread replies.",
+            "knownValues": [
+              "app.bsky.unspecced.getPostThreadV2#newest",
+              "app.bsky.unspecced.getPostThreadV2#oldest",
+              "app.bsky.unspecced.getPostThreadV2#top"
```
devinivy, line 49: This only needs to be tokens if we want to leave it open to extension by other lexicon namespaces.  It might be a case where we can get by with regular strings.  There is some precedence for this in `threadViewPref.sort`.

lexicons/app/bsky/unspecced/getPostThreadV2.json
```diff
@@ -0,0 +1,144 @@
...
+          "type": "array",
+          "description": "Annotations on this post that might be used by clients to customize experiences.",
+          "items": {
+            "type": "string",
+            "knownValues": [
+              "app.bsky.unspecced.getPostThreadV2#hasReplies",
+              "app.bsky.unspecced.getPostThreadV2#hiddenByThreadgate",
+              "app.bsky.unspecced.getPostThreadV2#mutedByViewer",
+              "app.bsky.unspecced.getPostThreadV2#opThread"
+            ]
```
devinivy, line 123: Just my gut, but this feels a little explicit to go from `{ hasReplies: true }` to `{ annotations: ['app.bsky.unspecced.getPostThreadV2#hasReplies'] }`

devinivy: Another way to look at it—the issue with using the field `hasReplies` wasn't that it was closed to extension. As I understand it we were just looking for a tidier way to model a list of booleans.

packages/bsky/tests/views/__snapshots__/thread-v2.test.ts.snap
```diff
@@ -0,0 +1,1103 @@
...
+    },
+  ],
+}
+`;
+
+exports[`appview thread views v2 simple thread returns thread anchored on 2_0 1`] = `
+Object {
+  "thread": Array [
+    Object {
+      "depth": -2,
```
devinivy, line 244: Haha negative depth is a funny quirk!  I wonder how it will get treated by consumers—e.g. maybe they will try first to split off the section of ancestors by filtering on `depth < 0` then treat those specially.

packages/bsky/src/api/app/bsky/unspecced/getPostThreadV2.ts
```diff
@@ -0,0 +1,154 @@
...
+        result = await getPostThread({ ...params, hydrateCtx }, ctx)
+      } catch (err) {
+        const repoRev = await ctx.hydrator.actor.getRepoRevSafe(viewer)
+        if (repoRev) {
+          res.setHeader(ATPROTO_REPO_REV, repoRev)
+        }
+        throw err
+      }
+
+      const repoRev = await ctx.hydrator.actor.getRepoRevSafe(viewer)
```
devinivy, line 54: This is used for read-after-write.  No harm including it, but if the client and pds aren't going to use it, we may be able to start by omitting it.

packages/bsky/src/api/app/bsky/unspecced/getPostThreadV2.ts
```diff
@@ -0,0 +1,154 @@
...
+    viewer: params.hydrateCtx.viewer,
+  })
+
+  const anchor = thread.find((i) => i.depth === 0)
+  if (!anchor || isThreadItemNotFound(anchor.value)) {
+    throw new InvalidRequestError(
+      `Post not found: ${skeleton.anchor}`,
+      'NotFound',
+    )
+  }
```
devinivy, line 122: Currently in the app when I try to view a thread with anchor by someone I block, I get a screen "Post hidden: You have blocked the author or you have been blocked by the author."

How will the app check for that state with this new API?

packages/bsky/src/api/app/bsky/unspecced/getPostThreadV2.ts
```diff
@@ -0,0 +1,154 @@
...
+  const thread = ctx.views.threadV2(skeleton, hydration, {
+    above: params.above,
+    below: calculateBelow(ctx, skeleton.anchor, params),
+    branchingFactor: params.branchingFactor,
+    prioritizeFollowedUsers: params.prioritizeFollowedUsers,
+    sorting: params.sorting,
+    viewer: params.hydrateCtx.viewer,
+  })
+
+  const anchor = thread.find((i) => i.depth === 0)
```
devinivy, line 116: No big deal here, but it feels like there should be a more direct way to check whether the thread was "found" other than scanning the thread items for the anchor post.

I also wonder if it's a bit more direct to check based on the anchor's uri rather than knowing that it's the unique item with `depth: 0`.

packages/bsky/src/api/app/bsky/unspecced/getPostThreadV2.ts
```diff
@@ -0,0 +1,154 @@
...
+  if (!anchor || isThreadItemNotFound(anchor.value)) {
```
devinivy, line 117: Not finding the `anchor` in the thread item list may not be an issue once this API is paginated, since the anchor wont necessarily be in the response.

packages/bsky/src/views/threadsV2.ts
```diff
@@ -0,0 +1,294 @@
+import { AppBskyUnspeccedGetPostThreadV2, asPredicate } from '@atproto/api'
```
devinivy, line 1: Small style thing with this file name—we would typically use kebab-case here by convention.

packages/bsky/src/views/index.ts
```diff
@@ -1111,6 +1131,462 @@ export class Views {
...
+      postView: PostView
+      rootUri: string
+      uri: string
+      viewer: HydrateCtx['viewer']
+    },
+    state: HydrationState,
+  ): ThreadItemValuePost {
+    const authorDid = creatorFromUri(uri)
+
+    const hasReplies = (postView.replyCount ?? 0) > 0
```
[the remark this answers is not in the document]
devinivy: (And we probably wont want to lean on the reply count for that unless it can be imperfect, due to e.g. blocks.)

packages/bsky/src/views/index.ts
devinivy: Just curious, did yall ever talk about something like `{ ancestors, anchor, replies }`?

packages/bsky/src/views/threadsV2.ts
```diff
@@ -0,0 +1,294 @@
...
+    branchingFactor,
+    fetchedAt,
+    opDid,
+    prioritizeFollowedUsers,
+    sorting,
+    viewer,
+  } = opts
+
+  if (node.replies) {
+    node.replies.sort((aTree: ThreadTree, bTree: ThreadTree) => {
```
devinivy, line 103: I'm not opposed to this function `sortTrimThreadTree()` mutating its input `node`, but I think it would be nice to call out at the top of the function as a headsup.

packages/bsky/src/views/threadsV2.ts
```diff
@@ -0,0 +1,294 @@
...
+  direction,
+}: {
+  tree: ThreadTree
+  direction: 'up' | 'down'
+}): Generator<ThreadItem, void> {
+  if (isThreadNoUnauthenticatedNode(tree)) {
+    if (direction === 'up') {
+      if (tree.parent) {
+        // Unfold all parents above.
+        yield* flattenThree(tree.parent)
```
devinivy, line 274: I was a little confused about this branch—mind helping me understand what's up with the no-unauthenticated case?

packages/bsky/src/views/threadsV2.ts
```diff
@@ -0,0 +1,294 @@
...
+  // Fallback to newest.
+  return b.post.indexedAt.localeCompare(a.post.indexedAt)
+}
+
+function topSortValue(likeCount: number, hasOPLike: boolean): number {
+  return Math.log(3 + likeCount) * (hasOPLike ? 1.45 : 1.0)
+}
+
+function flattenThree(tree: ThreadTree) {
+  return Array.from([
```
devinivy, line 241: Do we need this outer `Array.from()`?

packages/bsky/src/views/index.ts
```diff
@@ -145,6 +158,13 @@ export class Views {
     return false
   }
 
+  noUnauthenticatedPost(viewer: HydrateCtx['viewer'], post: PostView): boolean {
+    const isNoUnauthenticated = !!post.author.labels?.find(
```
devinivy, line 162: I tend to use `some()` in these cases when just checking for existence and the item itself isn't needed.

packages/bsky/src/views/index.ts
```diff
@@ -1111,6 +1131,462 @@ export class Views {
...
+          viewer,
+          opDid,
+          rootUri,
+          state,
+          above: opts.above,
+          depth: -1,
+        })
+      : undefined
+
+    const { tree: parent, isOPThread: isOpThreadFromRootToParent } =
```
devinivy, line 1208: Small consistency thing—probably best to stick with either `isOPThread` or `isOpThread`.

devinivy: (Similar with `isOpPost`.)

packages/bsky/src/views/index.ts
```diff
@@ -1111,6 +1131,462 @@ export class Views {
...
+    // Reached the `above` limit.
+    if (Math.abs(depth) > above) {
+      return undefined
+    }
+
+    // Not found.
+    const uri = state.posts?.get(childUri)?.record.reply?.parent.uri
+    if (!uri) {
+      return undefined
+    }
```
devinivy, line 1291: One thing we couldn't do in `getPostThread` is include a deleted post but maintain the ancestor chain, e.g.:
```
0         # anchor's parent's parent appears in full above deleted post
0_0✗      # anchor's parent is deleted
0_0_0     # anchor
0_0_0_0   # reply to anchor
```

Might be nice to introduce!  But we can always do it in a future pass.

packages/bsky/src/views/index.ts
```diff
@@ -1111,6 +1131,462 @@ export class Views {
...
+        return undefined
+      }
+
+      // Blocked (1p and 3p for replies).
+      const has1pBlock = this.viewerBlockExists(authorDid, state)
+      const has3pBlock =
+        !state.ctx?.include3pBlocks && state.postBlocks?.get(uri)?.parent
+      if (has1pBlock || has3pBlock) {
+        return undefined
+      }
```
devinivy, line 1428: It's not obvious to me whether we want to show 1p and 3p blocks identically—and I think they are treated differently in threads v1—but these are fairly sensitive behaviors to get right.  So just flagging for a double-check 👍

packages/bsky/src/api/app/bsky/unspecced/getPostThreadV2.ts
devinivy: I'd recommend not doing it if we can get away with it, since we are trying to move away from app logic in the pds.  The endpoint is also still unspecced, potentially prone to change.

devinivy (commented):
Looking pretty darn good, and the tests for this came out quite nice.

packages/bsky/src/views/index.ts
devinivy: I mean't the response structure having three distinct parts (`ancestors`, `anchor`, `replies`), each individual part flat.

lexicons/app/bsky/unspecced/getPostThreadV2.json
estrattonbailey: May need to revisit this once we do some testing with some prod data

packages/bsky/src/views/index.ts
devinivy: Ah, I was looking at this and my interpretation was that 3p replies were omitted while 1p replies were included as `#blockedPost`: https://github.com/bluesky-social/atproto/blob/e939f8e466979680a6949b9851754b4e0e8e5704/packages/bsky/src/views/index.ts#L1080-L1093

packages/bsky/tests/views/__snapshots__/thread-v2.test.ts.snap
devinivy: No specific concerns, just thinking through how folks will likely interact with these details of the API.  The reason I ask here is because I think generally the ancestors aren't presented nested despite their `depth`.  It's a UI concern, but also maybe a common/anticipated use case.

packages/bsky/src/api/app/bsky/unspecced/getPostThreadV2.ts
devinivy: Looking at that test, I suppose the guarantee is that there will always be a single item in the response, i.e. ancestors are never included above a blocked post.  So the client will detect this using something like `thread.length === 1 && threadItemBlocked(thread[0])`.  Or should clients not rely on that behavior and instead try to `find()` the item in the response with `depth: 0`, with the knowledge that only the anchor can have a `depth` of `0`.

A related question is whether a not-found vs. a blocked item should be detected in different ways by the client—the former by an error and the latter through inspecting the success response.  I ask because these are fairly similar situations for consumers. 
 There is actually a [TODO](https://github.com/bluesky-social/atproto/blob/e939f8e466979680a6949b9851754b4e0e8e5704/packages/bsky/src/api/app/bsky/feed/getPostThread.ts#L111-L113) around this in the old endpoint.

Trying to imagine how each case will be detected once pagination is in play is also interesting.  E.g. perhaps it's expected that the result set for a paginated query could be empty, which would mean the client doesn't have a way to check whether the anchor is blocked.

devinivy (approved):
Good stuff!  Let's just ensure to get a changeset in there for `@atproto/api`.

packages/bsky/src/api/app/bsky/unspecced/getPostThreadV2.ts
estrattonbailey: I think both cases could be returned as a "valid" response, but with a different `threadItem*` shape as the item with `depth === 0`. That was our intention with `threadItemNotFound` or whatever it's called now.

PR #3838, opened 2025-05-08
lexicons: add account management to describeServer links

bnewbold:
Before merge:

- [ ] changeset (?)
- [ ] codegen

This allows PDS instances (and implementations) specify a URL for account self-management. For example, the PDS reference implementation puts this at `https://<hostname>/account` for self-hosters, but `https://bsky.social/account` for folks on the Bluesky mushroom PDS instances. Other PDS implementations might put them at a different path or hostname. The intended use case is for client apps to provide a link within the app to account settings.

Haven't deeply though this through, but tossing over this PR as a placeholder/reminder that we need a hook like this. Maybe should clarify whether this is not (or could be?) for account signup versus existing account management.

### DIFF
--- a/lexicons/com/atproto/server/describeServer.json
+++ b/lexicons/com/atproto/server/describeServer.json
@@ -26,7 +26,7 @@
             },
             "links": {
               "type": "ref",
-              "description": "URLs of service policy documents.",
+              "description": "URLs of service interfaces and policy documents.",
               "ref": "#links"
             },
             "contact": {
@@ -45,6 +45,11 @@
     "links": {
       "type": "object",
       "properties": {
+        "accountManagement": {
+          "type": "string",
+          "format": "uri",
+          "description": "web interface for account self-management"
+        },
         "privacyPolicy": { "type": "string", "format": "uri" },
         "termsOfService": { "type": "string", "format": "uri" }
       }


bnewbold:
Some more comments to follow-up on:

- does OIDC have a metadata field like this? or OAuth Auth Server metadata files?
- any security concerns we need to lock down?
- should requests to whatever endpoint is included here include a `login_hint` query param, to make "which account" flow smoother when coming from apps?
- should relative URLs be allowed (`"/account"`), or only full URLs

raiderss:
The documentation in this repository is excellent! Very clear and comprehensive.

matthieusieben:
- I skimmed through https://www.iana.org/assignments/oauth-parameters/oauth-parameters.xhtml and could not find mention of a url for user account management.

- Regarding security: That endpoint should (*must*) implement strong CSP policy to avoid being embedded in an iframe. If should also implement CSRF prevention mechanisms.

- I think so. If we are adding it to `describeServer` (as opposed to a more "OAuth"y endpoint), I think we should use a more "atproto"y naming, like `did`.

- only full URLs (with mandatory `https` scheme) should be allowed IMO

matthieusieben (approved):

PR #3847, opened 2025-05-10
Properly validate auth during refresh

matthieusieben:
Ensure that the credentials used during a refresh correspond to those used to create the OAuth tokens.

### DIFF
--- a/packages/oauth/oauth-provider/src/client/client.ts
+++ b/packages/oauth/oauth-provider/src/client/client.ts
@@ -1,6 +1,6 @@
 import {
+  type JWTHeaderParameters,
   type JWTPayload,
-  type JWTVerifyGetKey,
   type JWTVerifyOptions,
   type JWTVerifyResult,
   type KeyLike,
@@ -28,6 +28,7 @@ import { InvalidParametersError } from '../errors/invalid-parameters-error.js'
 import { InvalidRequestError } from '../errors/invalid-request-error.js'
 import { InvalidScopeError } from '../errors/invalid-scope-error.js'
 import { compareRedirectUri } from '../lib/util/redirect-uri.js'
+import { Awaitable } from '../lib/util/type.js'
 import { ClientAuth, authJwkThumbprint } from './client-auth.js'
 import { ClientId } from './client-id.js'
 import { ClientInfo } from './client-info.js'
@@ -40,7 +41,9 @@ export class Client {
    */
   static readonly AUTH_METHODS_SUPPORTED = ['none', 'private_key_jwt'] as const
 
-  private readonly keyGetter: JWTVerifyGetKey
+  private readonly keyGetter: (
+    protectedHeader: JWTHeaderParameters,
+  ) => Awaitable<KeyLike | Uint8Array>
 
   constructor(
     public readonly id: ClientId,
@@ -205,13 +208,10 @@ export class Client {
         return false
       }
       try {
-        const key = await this.keyGetter(
-          {
-            kid: clientAuth.kid,
-            alg: clientAuth.alg,
-          },
-          { payload: '', signature: '' },
-        )
+        const key = await this.keyGetter({
+          kid: clientAuth.kid,
+          alg: clientAuth.alg,
+        })
         const jtk = await authJwkThumbprint(key)
 
         return jtk === clientAuth.jkt
--- a/packages/oauth/oauth-provider/src/oauth-provider.ts
+++ b/packages/oauth/oauth-provider/src/oauth-provider.ts
@@ -855,15 +855,17 @@ export class OAuthProvider extends OAuthVerifier {
   ) {
     // > The authorization server first validates the client credentials (in
     // > case of a confidential client)
-    const [client, clientAuth] = await this.authenticateClient(credentials)
+    const [client] = await this.authenticateClient(credentials)
 
     const tokenInfo = await this.tokenManager.findToken(token)
 
     // > [...] and then verifies whether the token was issued to the client
-    // > making the revocation request. If this validation fails, the request is
-    // > refused and the client is informed of the error by the authorization
-    // > server as described below.
-    await this.tokenManager.validateAccess(client, clientAuth, tokenInfo)
+    // > making the revocation request.
+    if (client.id !== tokenInfo.data.clientId) {
+      // > If this validation fails, the request is refused and the client is
+      // > informed of the error by the authorization server as described below.
+      throw new InvalidGrantError(`Token was not issued to this client`)
+    }
 
     // > In the next step, the authorization server invalidates the token. The
     // > invalidation takes place immediately, and the token cannot be used
--- a/packages/oauth/oauth-provider/src/token/token-manager.ts
+++ b/packages/oauth/oauth-provider/src/token/token-manager.ts
@@ -305,11 +305,32 @@ export class TokenManager {
       throw new InvalidGrantError(`Token was not issued to this client`)
     }
 
-    if (tokenInfo.data.clientAuth.method !== clientAuth.method) {
-      throw new InvalidGrantError(`Client authentication method mismatch`)
+    const initialClientAuth = tokenInfo.data.clientAuth
+
+    switch (initialClientAuth.method) {
+      case CLIENT_ASSERTION_TYPE_JWT_BEARER:
+        if (
+          clientAuth.method !== initialClientAuth.method ||
+          clientAuth.jkt !== initialClientAuth.jkt
+        ) {
+          throw new InvalidGrantError(`Client authentication mismatch`)
+        }
+        break
+      case 'none':
+        if (clientAuth.method !== initialClientAuth.method) {
+          throw new InvalidGrantError(`Client authentication mismatch`)
+        }
+        break
+      default:
+        throw new InvalidGrantError(
+          // @ts-expect-error (future proof, backwards compatibility)
+          `Invalid method "${initialClientAuth.method}"`,
+        )
     }
 
-    if (!(await client.validateClientAuth(tokenInfo.data.clientAuth))) {
+    // Ensure that the initial auth is still valid (public key still present).
+    // No need
+    if (!(await client.validateClientAuth(initialClientAuth))) {
       throw new InvalidGrantError(`Client authentication mismatch`)
     }
   }
--- a/.changeset/rare-pillows-search.md
+++ b/.changeset/rare-pillows-search.md
@@ -1,0 +1,5 @@
+---
+"@atproto/oauth-provider": minor
+---
+
+Ensure that the credentials used during a refresh correspond to those used to create the OAuth tokens.


packages/oauth/oauth-types/src/oauth-client-metadata.ts
```diff
@@ -33,8 +34,10 @@ export const oauthClientMetadataSchema = z.object({
     // > "authorization_code" Grant Type.
     .default(['authorization_code']),
   scope: oauthScopeSchema.optional(),
+  // https://www.rfc-editor.org/rfc/rfc7591.html#section-2
   token_endpoint_auth_method: oauthEndpointAuthMethod
-    .default('none')
+    // > If unspecified or omitted, the default is "client_secret_basic" [...].
+    .default('client_secret_basic')
```
matthieusieben, line 40: > [!NOTE]
> This is a change that will break all existing clients that do not explicitly define the `token_endpoint_auth_method` config.

packages/oauth/oauth-client/src/oauth-server-factory.ts
```diff
@@ -17,19 +21,47 @@ export class OAuthServerFactory {
...
+    dpopKey: Key,
+    options?: GetCachedOptions,
+  ) {
     const serverMetadata = await this.resolver.getAuthorizationServerMetadata(
       issuer,
       options,
     )
-    return this.fromMetadata(serverMetadata, dpopKey)
+
+    if (authMethod === 'legacy') {
```
matthieusieben, line 44: Because we were previously not storing the `authMethod` in the session data, we provide a backwards compatible implementation by computing it here.

packages/oauth/oauth-provider/src/token/token-manager.ts
```diff
@@ -305,11 +305,31 @@ export class TokenManager {
...
 
-    if (tokenInfo.data.clientAuth.method !== clientAuth.method) {
-      throw new InvalidGrantError(`Client authentication method mismatch`)
+    const initialClientAuth = tokenInfo.data.clientAuth
+
+    switch (initialClientAuth.method) {
+      case CLIENT_ASSERTION_TYPE_JWT_BEARER:
+        if (
+          clientAuth.method !== initialClientAuth.method ||
+          clientAuth.jkt !== initialClientAuth.jkt
```
matthieusieben, line 314: This is the "big" change, server side.

packages/oauth/oauth-client-node/src/node-dpop-store.ts
```diff
@@ -11,9 +11,10 @@ type ToDpopJwkValue<V extends { dpopKey: Key }> = Omit<V, 'dpopKey'> & {
  * Utility function that allows to simplify the store interface by exposing a
  * JWK (JSON) instead of a Key instance.
  */
-export function toDpopKeyStore<K extends string, V extends { dpopKey: Key }>(
-  store: SimpleStore<K, ToDpopJwkValue<V>>,
-): SimpleStore<K, V> {
+export function toDpopKeyStore<
+  K extends string,
+  V extends { dpopKey: Key; dpopJwk?: never },
```
matthieusieben, line 16: Improved type safety

packages/oauth/oauth-client/src/oauth-client-auth.ts
```diff
@@ -0,0 +1,146 @@
...
+
+  throw new Error(
+    `The ATProto OAuth spec requires that client use either "none" or "client_secret_jwt" authentication method.` +
+      (method === 'client_secret_basic'
+        ? ' You might want to explicitly set "token_endpoint_auth_method" to one of those values in the client metadata document.'
+        : ` You set "${method}" which is not allowed.`),
+  )
+}
+
+export type ClientCredentialsGetter = () => Awaitable<OAuthClientCredentials>
```
matthieusieben, line 70: Maybe this would be better named `ClientCredentialsFactory` ?

packages/oauth/oauth-client/src/oauth-client-auth.ts
```diff
@@ -0,0 +1,146 @@
...
+      client_id: clientMetadata.client_id,
+    })
+  }
+
+  if (authMethod.method === 'client_secret_jwt') {
+    try {
+      // The client used to be a confidential client but no longer has a keyset.
+      if (!keyset) throw new Error('A keyset is required for client_secret_jwt')
+
+      const [key, alg] = keyset.findPrivateKey({
```
matthieusieben, line 102:
```suggestion
      // @NOTE throws if no matching key can be found
      const [key, alg] = keyset.findPrivateKey({
```

packages/oauth/oauth-client/src/oauth-client.ts
```diff
@@ -425,6 +437,7 @@ export class OAuthClient extends CustomEventTarget<OAuthClientEventMap> {
 
       const server = await this.serverFactory.fromIssuer(
         stateData.iss,
+        stateData.authMethod ?? 'legacy',
```
matthieusieben, line 440: Using the literal `"legacy"` if the `authMethod` is not defined (because `stateData` was created through an old version of this lib)

.changeset/quiet-pans-fix.md
```diff
@@ -0,0 +1,5 @@
+---
+"@atproto/oauth-provider": patch
+---
+
+Verify the presence of a "kid" in singed JAR headers
```
devinivy, line 5:
```suggestion
Verify the presence of a "kid" in signed JAR headers
```

packages/oauth/oauth-provider/src/client/client-auth-check.ts
```diff
@@ -0,0 +1,53 @@
...
+    case CLIENT_ASSERTION_TYPE_JWT_BEARER:
+      if (clientAuth.method !== previous.clientAuth.method) {
+        throw new InvalidGrantError(
+          `Client authentication method mismatch (expected ${previous.clientAuth.method})`,
+        )
+      }
+      if (
+        clientAuth.kid !== previous.clientAuth.kid ||
+        clientAuth.alg !== previous.clientAuth.alg ||
+        clientAuth.jkt !== previous.clientAuth.jkt
```
devinivy, line 33: Just confirming: the client key's `jkt` is calculated by oauth-provider and doesn't come from the JWK itself, right?  I know this as a DPoP mechanism so just making sure I understand its usage here.

matthieusieben: That's right. Computed here:

https://github.com/bluesky-social/atproto/blob/a1feaf84963afbaa323aaa79138551d1e1e8841b/packages/oauth/oauth-provider/src/client/client.ts#L248

packages/oauth/oauth-provider/src/client/client-auth-check.ts
```diff
@@ -0,0 +1,54 @@
...
+          `The session was initiated with a different key than the client assertion currently used`,
+        )
+      }
+      break
+    case 'none':
+      if (clientAuth.method !== previous.clientAuth.method) {
+        throw new InvalidGrantError(
+          `Client authentication method mismatch (expected ${previous.clientAuth.method})`,
+        )
+      }
```
matthieusieben: Note: this also allows to support the legacy case where non PAR auth requests were created with `{ method: "none" }` (now represented as `null`). This would only concern sessions created before the deployment of this code, and "token exchanged" right after, which is why I do not add a comment about that in the code here.

.changeset/cuddly-yaks-hide.md
```diff
@@ -0,0 +1,5 @@
+---
+"@atproto/oauth-client": minor
+---
+
+Use `private_key_jwt` instead of incorrect `client_secret_jwt` as authentication method for confidential clients
```
devinivy: I don't think `client_secret_jwt` was ever listed as a valid authentication method for confidential clients, at least not in a version that landed in `main` or was published.

packages/oauth/oauth-client/src/oauth-client-auth.ts
```diff
@@ -0,0 +1,152 @@
...
+}
+
+function supportedMethods(serverMetadata: OAuthAuthorizationServerMetadata) {
+  return serverMetadata['token_endpoint_auth_methods_supported']
+}
+
+function supportedAlgs(serverMetadata: OAuthAuthorizationServerMetadata) {
+  return (
+    serverMetadata['token_endpoint_auth_signing_alg_values_supported'] ?? [
+      // @TODO Make sure that this fallback is reflected in the spec
```
devinivy: Just calling out to ensure the TODO has been captured.

packages/oauth/oauth-client/src/oauth-client-auth.ts
```diff
@@ -0,0 +1,152 @@
...
+        client_assertion_type: CLIENT_ASSERTION_TYPE_JWT_BEARER,
+        client_assertion: await key.createJwt(
+          { alg },
+          {
+            iss: clientMetadata.client_id,
+            sub: clientMetadata.client_id,
+            aud: serverMetadata.issuer,
+            jti: await runtime.generateNonce(),
+            iat: Math.floor(Date.now() / 1000),
+          },
```
devinivy: I think per the spec this must have an `exp`.

matthieusieben: Are you sure? I don't see it here:
https://datatracker.ietf.org/doc/html/rfc9449#section-4.2

devinivy: As I understand it this is the client assertion.  Check out requirement 4 here: https://www.rfc-editor.org/rfc/rfc7523.html#section-3

(Also makes some intuitive sense to me since this is used for authentication and it's a simple mitigation for replays.)

packages/oauth/oauth-provider/src/token/token-manager.ts
```diff
@@ -363,7 +365,6 @@ export class TokenManager {
     const { parameters } = data
 
     try {
-      await this.validateAccess(client, clientAuth, tokenInfo)
       await this.validateRefresh(client, clientAuth, tokenInfo)
```
devinivy: Was the access validation never needed here?

matthieusieben: It's now performed from the `validateRefresh()`

matthieusieben: I moved it to `validateRefresh` and now moved it back here.

packages/oauth/oauth-provider/src/client/client.ts
```diff
@@ -151,27 +143,22 @@ export class Client {
...
-      return { clientAuth }
+      return { method: 'none' }
     }
 
     if (method === 'private_key_jwt') {
-      if (!('client_assertion_type' in input)) {
+      if (!('client_assertion' in input)) {
         throw new InvalidRequestError(
-          `client_assertion_type required for "${method}"`,
+          `client authentication method "${method} required an "client_assertion"`,
```
devinivy:
```suggestion
          `client authentication method "${method}" required a "client_assertion"`,
```

packages/oauth/oauth-provider/src/request/request-manager.ts
```diff
@@ -457,53 +456,26 @@ export class RequestManager {
...
-          `The code was not issued to client "${client.id}"`,
-        )
-      }
-
-      if (data.expiresAt < new Date()) {
-        throw new InvalidGrantError('This code has expired')
-      }
+    // A "code" can only be used once. This should already have been ensured
+    // by the store implementation, but let's be extra safe.
+    await this.store.deleteRequest(id)
```
devinivy: On the flip side, would this patch over an incorrect implementation and not surface the underlying issue?  I wonder if a warning log and/or assertion in dev mode might be more appropriate.

packages/oauth/oauth-provider/src/request/request-store-redis.ts
```diff
@@ -19,14 +19,18 @@ export class RequestStoreRedis implements RequestStore {
   }
 
   async readRequest(id: RequestId): Promise<RequestData | null> {
-    const data = await this.redis.get(id)
-    return data ? JSON.parse(data) : null
+    const value = await this.redis.get(id)
+    if (!value) return null
+
+    const data = decode(value)
+    if (!data) await this.redis.del(id)
```
devinivy: What is the purpose of this addition?  I wonder if it's necessary—when reads end-up having an effect it can break expectations of consumers, sometimes cause unexpected races, so I would lean towards only introducing it if it serves a known/specific function.

packages/oauth/oauth-provider/src/request/request-store-redis.ts
```diff
@@ -19,14 +19,18 @@ export class RequestStoreRedis implements RequestStore {
...
+
+    return data
   }
 
   async createRequest(id: RequestId, data: RequestData): Promise<void> {
-    const timeFrame = data.expiresAt.getTime() - Date.now()
-    await this.redis.set(id, JSON.stringify(data), 'PX', timeFrame)
-    if (data.code) await this.redis.set(data.code, id, 'PX', timeFrame)
+    await this.redis.set(id, encode(data), 'PX', px(data.expiresAt))
+    if (data.code) await this.redis.set(data.code, id, 'PX', px(data.expiresAt))
```
devinivy: It is slightly awkward that `px(data.expiresAt)` will take a different value in each call here.  We could compute it once, maybe even additionally pipeline these two calls.

packages/oauth/oauth-provider/src/token/validate-refresh-token.ts
```diff
@@ -0,0 +1,43 @@
...
+  const [sessionLifetime, refreshLifetime] =
+    clientAuth.method !== 'none' || client.info.isFirstParty
+      ? [
+          CONFIDENTIAL_CLIENT_SESSION_LIFETIME,
+          CONFIDENTIAL_CLIENT_REFRESH_LIFETIME,
+        ]
+      : [PUBLIC_CLIENT_SESSION_LIFETIME, PUBLIC_CLIENT_REFRESH_LIFETIME]
+
+  const sessionAge = Date.now() - tokenInfo.data.createdAt.getTime()
+  const refreshAge = Date.now() - tokenInfo.data.updatedAt.getTime()
```
devinivy: Would recommend computing these ages from the same "now", especially in this case since they get compared to each other.

packages/oauth/oauth-provider/src/token/validate-refresh-token.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  if (sessionAge > sessionLifetime) {
+    throw new InvalidGrantError(`Session expired`)
+  } else if (refreshAge > refreshLifetime) {
+    throw new InvalidGrantError(`Refresh token expired`)
+  } else if (refreshAge > sessionAge) {
+    throw new InvalidGrantError(`Refresh token is older than session`)
```
devinivy: When does this happen?  Or just catching this as a failed invariant (which "should" never happen)?

packages/oauth/oauth-provider/src/token/token-manager.ts
```diff
@@ -299,165 +190,80 @@ export class TokenManager {
...
-      )
+    await this.store.rotateToken(tokenInfo.id, nextTokenId, nextRefreshToken, {
+      updatedAt: now,
+      expiresAt,
+      // @NOTE Normally, the clientAuth not change over time. There are two
+      // exceptions:
+      // - Upgrade from a legacy representation of client authentication to
+      //   a modern one.
+      // - Allow clients to become "confidential" if they were previously
+      //   "public
```
devinivy:
```suggestion
      //   "public"
```

packages/oauth/oauth-provider/src/token/token-manager.ts
```diff
@@ -475,30 +281,39 @@ export class TokenManager {
...
+    token: RefreshToken,
+  ): Promise<null | TokenInfo> {
+    return this.store.findTokenByRefreshToken(token)
+  }
 
-    if (!tokenInfo) {
-      throw new InvalidRequestError(`Invalid refresh token`)
-    }
+  public async consumeRefreshToken(token: RefreshToken): Promise<TokenInfo> {
+    // @NOTE concurrent refreshes of the same refresh token could theoricaly
```
devinivy:
```suggestion
    // @NOTE concurrent refreshes of the same refresh token could theoretically
```

packages/oauth/oauth-provider/src/oauth-provider.ts
```diff
@@ -438,7 +459,11 @@ export class OAuthProvider extends OAuthVerifier {
     dpopProof: null | DpopProof,
   ): Promise<OAuthParResponse> {
     try {
-      const [client, clientAuth] = await this.authenticateClient(credentials)
+      const { client, clientAuth } = await this.authenticateClient(
+        credentials,
+        dpopProof,
+        { allowMissingDpopProof: true },
```
devinivy, line 476: Probably worth a comment describing why the proof is allowed to be missing.

packages/oauth/oauth-provider/src/request/request-store-redis.ts
matthieusieben: This actually a wanted behavior. We basically want both entries to expire together. So if the first call takes 10ms, the second entry should have a 10ms shorter TTL.

There might be better ways of doing this with Redis but I did not spent too much time investigating  (the `mset()` method does not have type definitions that allows `PX`).

packages/oauth/oauth-provider/src/request/request-store-redis.ts
matthieusieben: reverted

packages/oauth/oauth-provider/src/request/request-store-redis.ts
matthieusieben: adapted code to use absolute expiracy date instead

packages/oauth/oauth-provider/src/token/validate-refresh-token.ts
matthieusieben: just catching this as a failed invariant. Added comment

packages/oauth/oauth-client/src/oauth-client-auth.ts
matthieusieben: I probably needed more coffee -_-

PR #3853, opened 2025-05-13
Transitional Identity OAuth Scope

knotbin:
**DEPENDS ON #3820** 

Adds a transitional Identity OAuth scope

### DIFF
--- a/.changeset/early-trains-share.md
+++ b/.changeset/early-trains-share.md
@@ -1,0 +1,5 @@
+---
+"@atproto/oauth-client-browser-example": patch
+---
+
+Add `com.atproto.server.getSession` query.
--- a/.changeset/five-toes-clean.md
+++ b/.changeset/five-toes-clean.md
@@ -3,4 +3,4 @@
 "@atproto/pds": minor
 ---
 
-add transitional full access oauth scope
+add transitional identity oauth scope for account actions
--- a/.changeset/funny-moles-explain.md
+++ b/.changeset/funny-moles-explain.md
@@ -1,0 +1,5 @@
+---
+"@atproto/pds": patch
+---
+
+Allow access to `com.atproto.server.getSession` when authenticated using oauth credentials. Email info (`email`, `emailConfirmed`) will only be exposed if the credentials were issued with the `transition:email` scope.
--- a/.changeset/shy-moose-fail.md
+++ b/.changeset/shy-moose-fail.md
@@ -1,0 +1,7 @@
+---
+"@atproto/oauth-provider-ui": patch
+"@atproto/oauth-provider": patch
+"@atproto/pds": patch
+---
+
+Add support for `transition:email` oauth scope
--- a/.changeset/silver-scissors-laugh.md
+++ b/.changeset/silver-scissors-laugh.md
@@ -1,0 +1,5 @@
+---
+"@atproto/oauth-client-browser-example": patch
+---
+
+Allow configuration of loopback client id scope through query param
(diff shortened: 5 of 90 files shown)


knotbin:
Context: I am working on a web account migration tool that uses OAuth. I need to be able to make PLC requests, sign PLC operations, and deactivate accounts with OAuth.

I figure it's better for the user to go through OAuth and be warned that the app will get full access than to use password auth where the app can just not include those warnings and then do malicious things with that access.

matthieusieben:
Hey, @knotbin thank you for contributing.

There are two issues with this change:

1) We want to avoid the situation where scopes overlap with each other. With the change you introduced, `transitional:access` is a superset of `transitional:generic`.

2) We want to avoid users giving too broad powers to apps without knowing it.

*IF* we were to do this, it would be using a scope that does not overlap, in functionality, with other scopes (e.g. `transitional:identity`). We would also need a UX that clearly explains the dangers to the user in the OAuth flow.

Given the current state of thing, I am not sure this is a quite as simple change. For it to be a simple change, we would have to have a clean mapping of the various combination of `transitional:*` scopes to `AuthScope.*`. I don't think this is feasible as-is.

Note that I am [working](https://github.com/bluesky-social/atproto/pull/3820) (WIP) on a `transitional:email` scope to expose the user's email. The logic there might help properly structuring the changes needed to support another `transitional:identity` scope.

knotbin:
Got it. One question: I know that PR allows OAuth to use getSession, which provides email and verification status, but it doesn't seem to introduce any new `transitional:email` scope.
Correct me if I'm wrong, but it seems like it's a separate method of getting the session agent.
Is there a separate branch that has a new scope? Would be helpful to see the structure used for scopes that aren't perfectly matched to `AuthScope.*`.

matthieusieben:
The work was not pushed yet.

knotbin:
Got it. I think I found a way to add `transition:identity` as a non-overlapping scope with an `AuthScope.*` equivalent that can combine with `transition:chat.bsky`.

I completely understand if you want this to be refactored or changed based on your current work in #3820.

I also understand if this transitional scope is just not within your "scope" right now (I'll see myself out) and you want to wait until OAuth scopes are fully implemented to add this ability.

matthieusieben:
I'd rather not do it the way you did as we would need to create an `AuthScope` for every possible combination of oauth scope. Actually, moving forward, we will probably get rid of the `AuthScope` enum.

knotbin:
Got it, I'll implement that.

I changed this PR to merge into the branch #3820 is using (GH will change that to main once that branch is merged and deleted) so I can use the `OAuthOutput` type and keep working on this without creating conflicts, but obviously don't merge until that branch is merged, and I'll add a

knotbin:
> Hey @knotbin, here is how the oauth scopes can be retrieved from the auth data:
> 
> https://github.com/bluesky-social/atproto/blob/891b3c36912aa9e99a1ea8c0062991f8ab156453/packages/pds/src/api/com/atproto/server/getSession.ts#L75-L81

I used the same method to retrieve the scopes but edited the AuthVerifier to add an `additionalOauthScopes` property, as the methods I needed were all inaccessible without `AuthScope.Access`, whereas in the email change, OAuth already had permission as `AuthScope.Standard`, and only _further_ verification of the scope had to be added.

This change will provide a potential alternative to (and help eventually eliminate) the `authScopeEquivalent` property.

knotbin:
Is this unplanned or is it just merging to the wrong branch?

PR #3864, opened 2025-05-18
Correctly validate unions with refs mapping to record types

essential-randomness:
According [to the spec](https://atproto.com/specs/lexicon#union),`record` should be a valid type of `ref` in a `union`:

> The schema definitions pointed to by a union are objects or types with a clear mapping to an object, like a record.

However, lexicon validation currently fails for `record` fields in `union`.

> ValidationError: Unexpected lexicon type: record

This PR fixes the issue by adding validation for `record`.

### DIFF
--- a/packages/lexicon/src/validators/complex.ts
+++ b/packages/lexicon/src/validators/complex.ts
@@ -1,6 +1,7 @@
 import { Lexicons } from '../lexicons'
 import {
   LexArray,
+  LexRecord,
   LexRefVariant,
   LexUserType,
   ValidationError,
@@ -25,6 +26,8 @@ export function validate(
       return array(lexicons, path, def, value)
     case 'blob':
       return blob(lexicons, path, def, value)
+    case 'record':
+      return record(lexicons, path, def, value)
     default:
       return validatePrimitive(lexicons, path, def, value)
   }
@@ -152,6 +155,24 @@ export function object(
   }
 
   return { success: true, value: resultValue }
+}
+
+export function record(
+  lexicons: Lexicons,
+  path: string,
+  def: LexRecord,
+  value: unknown,
+): ValidationResult {
+  if (!isObj(value) || !('$type' in value)) {
+    return {
+      success: false,
+      error: new ValidationError(
+        `${path} must be a record with a "$type" property`,
+      ),
+    }
+  }
+  const propPath = `${path}/record`
+  return validateOneOf(lexicons, propPath, def.record, value)
 }
 
 export function validateOneOf(
--- a/packages/lexicon/tests/general.test.ts
+++ b/packages/lexicon/tests/general.test.ts
@@ -352,6 +352,18 @@ describe('General validation', () => {
     })
     expect(result.success).toBe(true)
 
+    result = lexicon.validate('com.example.testUnionWithRecordRef', {
+      union: {
+        $type: 'com.example.recordType',
+        // Test that the record type is validated against the record schema
+        test: 123,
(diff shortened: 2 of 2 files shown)


packages/lexicon/tests/general.test.ts
```diff
@@ -254,6 +254,124 @@ describe('General validation', () => {
...
+      union: {
+        $type: 'com.example.recordType',
+        // Test that the record type is validated against the record schema
+        test: 123,
+      },
+    })
+    expect(result.success).toBe(false)
+    expect(result['error']?.message).toBe(
+      'Object/union/record/test must be a string',
+    )
```
devinivy: I believe we want this to just read as the path down the object to the bad value:
```suggestion
    expect(result['error']?.message).toBe(
      'Object/union/test must be a string',
    )
```

packages/lexicon/src/validators/complex.ts
```diff
@@ -154,6 +157,24 @@ export function object(
...
+  if (!isObj(value) || !('$type' in value)) {
+    return {
+      success: false,
+      error: new ValidationError(
+        `${path} must be a record with a "$type" property`,
+      ),
+    }
+  }
+  const propPath = `${path}/record`
+  return validateOneOf(lexicons, propPath, def.record, value)
```
devinivy: Left a related comment in the tests—I believe we want the path here to reflect the path down the validated object itself.
```suggestion
  return validateOneOf(lexicons, path, def.record, value)
```

packages/lexicon/src/validators/complex.ts
```diff
@@ -154,6 +157,24 @@ export function object(
   return { success: true, value: resultValue }
 }
 
+export function record(
+  lexicons: Lexicons,
+  path: string,
+  def: LexRecord,
+  value: unknown,
+): ValidationResult {
+  if (!isObj(value) || !('$type' in value)) {
```
devinivy: We can use this helper:
```suggestion
  if (!isDiscriminatedObject(value)) {
```

essential-randomness:
> but for my own understanding, can you share the usecase that drove this PR?

In the demo app we're building for ATgeo, users can associate a geographic marker with a series of AtURIs, pointing to records from different lexicons (the demo features `app.bsky.feed.post` and `com.whtwnd.blog.entry`). [See here for the lexicon.](https://github.com/ATProtocol-Community/atgeo-marker/blob/main/lexicons/community/atprotocol/geomarker.json#L25)

In the AppView, I've then created [a `getMarkers` endpoint](https://github.com/ATProtocol-Community/atgeo-marker/blob/main/lexicons/community/atprotocol/defs.json#L54) that explicitly hydrates Lexicons from known supported types rather than just returning them as AtURIs.

Mostly I wanted to:
- Cache record reads at the AppView level rather than at each App's level, to make implementations easier
- See if it was possible and supported by the spec (which it seemed to be)

I haven't done this yet so I haven't thought too deeply about it, but I was also considering pairing the AppView with an installable UI component that would provide a default view for any record type supported. A union of record types makes which type of records the AppView can return explicit in a way that AtURIs alone don't, so it makes it easier to implement against it (even if non-exhaustively).

pfrazee:
Devin dropped in some review notes that I think are accurate. If you could fix those up, we'll get this merged! <3

essential-randomness:
Fixed them up! Thank you so much, excited to remove my patch from our demo app :)

I was also surprised you hadn't run into this, but happy to be a trailblazer!

PR #3873, opened 2025-05-20
Transitional Identity OAuth Scope

knotbin:
Builds on changes in @matthieusieben's #3820

(originally #3853 but branch that was merging into was merged and apparently nested PRs don't work for non-nested branches)


lorenzleutgeb:
Now that https://github.com/did-method-plc/did-method-plc/pull/101 was merged (also see #3928), I would like to implement a public web client with OAuth to append to the [`verificationMethods`](https://web.plc.directory/spec/v0.1/did-plc) of a `did:plc`. I have a proof of concept in Python, linked in https://github.com/did-method-plc/did-method-plc/issues/92 which calls the XRPC methods [`com.atproto.identity.requestPlcOperationSignature`](https://docs.bsky.app/docs/api/com-atproto-identity-request-plc-operation-signature) and [`com.atproto.identity.signPlcOperation`](https://docs.bsky.app/docs/api/com-atproto-identity-sign-plc-operation) in succession to achieve this. That Python script works, but it is not a web client!

However, for the web client, in response to `c.a.i.requestPlcOperationSignature` I get an HTTP response with status 400 and the body

```json
{
  "error": "InvalidToken",
  "message": "DPoP access token cannot be used for this request"
}
```

It is not entirely clear to me, as I am not so familiar with this codebase, but it appears that this PR might make it possible to call the above endpoints (assuming that the web client requests and the the user confirms the scope `transition:identity`, of course) from a web client.

It would be great if you could please help me clarify whether this is the case or not. And if it is the case, please consider my comment as an endorsement. I would like to use such feature.

(Mentioning @bnewbold and @DavidBuchanan314 as they have been working on the above, and this PR seems to be the next logical step to me.)

knotbin:
Yes this would fix that error by allowing DPoP tokens to be used for identity requests but I believe the right person to tag is @matthieusieben as he was the initial reviewer of this PR and the person who afaik does the most in the OAuth side of the codebase.

He had some specific requests which I fixed but prefaced them with a strong "_IF_ we were to do this" so I don't know if the lack of a review on this PR is a matter of having other things to do or a matter of deciding not to implement this before full OAuth scopes are added.

It would be very nice to get clarity on if this is a feature they are willing to merge or not so I know whether I should keep maintaining the PR.

@lorenzleutgeb I was in the same situation and I made the decision to just add password auth in addition to OAuth, and not show the OAuth option until it works. I would recommend you do the same if this PR continues to get no responses.

blowdart:
Right now it works for me if I go direct to the PDS, no proxy header or anything needed, just the auth dpop and the dpop header.

knotbin:
Right now _what_ works? PLC operations?

PR #3874, opened 2025-05-21
lexicon: validate 'uri' string length

bnewbold:
Link to relevant part of spec: https://atproto.com/specs/lexicon#uri

This sort of change can be fraught to introduce, because it is going to make existing records invalid. I think in this case, that will be rare enough that we can go ahead.

The go SDK does this check already during dynamic Lexicon validation (though we don't use that codepath anywhere yet; eg the appview doesn't do any real lexicon validation).

### DIFF
--- a/packages/lexicon/src/validators/formats.ts
+++ b/packages/lexicon/src/validators/formats.ts
@@ -1,6 +1,6 @@
 import { isValidISODateString } from 'iso-datestring-validator'
 import { CID } from 'multiformats/cid'
-import { validateLanguage } from '@atproto/common-web'
+import { validateLanguage, utf8Len } from '@atproto/common-web'
 import {
   ensureValidAtUri,
   ensureValidDid,
@@ -33,6 +33,12 @@ export function uri(path: string, value: string): ValidationResult {
     return {
       success: false,
       error: new ValidationError(`${path} must be a uri`),
+    }
+  }
+  if (utf8Len(value) > 8192) {
+    return {
+      success: false,
+      error: new ValidationError(`${path} URI must be 8 KBytes or less`),
     }
   }
   return { success: true, value }
--- a/.changeset/tiny-drinks-explode.md
+++ b/.changeset/tiny-drinks-explode.md
@@ -1,0 +1,5 @@
+---
+"@atproto/lexicon": patch
+---
+
+validate 'uri' string length


matthieusieben:
Any reason why that limit is 8192 here when RFCs [*recommend*](https://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers/417184#417184) using a limit of 8000 ?

mary-ext:
it's also the only string format using UTF-8 length checks AFAIK

packages/lexicon/src/validators/formats.ts
```diff
@@ -35,6 +35,12 @@ export function uri(path: string, value: string): ValidationResult {
       error: new ValidationError(`${path} must be a uri`),
     }
   }
+  if (utf8Len(value) > 8192) {
```
matthieusieben, line 38:
```suggestion
  if (new URL(value).href.length > 8192) {
```

mary-ext:
with Node.js and web having differing behaviors on non-HTTP URLs I do wonder if that's a viable approach, especially when `uri` syntax is pretty lax with its validation

packages/lexicon/src/validators/formats.ts
```diff
@@ -35,6 +35,12 @@ export function uri(path: string, value: string): ValidationResult {
...
+    return {
+      success: false,
+      error: new ValidationError(`${path} URI must be 8 KBytes or less`),
+    }
+  }
   return { success: true, value }
```
matthieusieben, line 44:
```ts
export function uri(path: string, value: string): ValidationResult {
  try {
    // normalize
    const url = new URL(value).href
    
    if (url.length > 8192) {
      return {
        success: false,
        error: new ValidationError(`${path} URI must be 8 KBytes or less`),
      }
    }
    
    // return the normalized value
    return { success: true, value: url }
  } catch {
    return {
      success: false,
      error: new ValidationError(`${path} must be a uri`),
    }
  }
}
```

bnewbold:
I don't think we should allow 8k grapheme clusters or whatever; folks can use whatever valid unicode in URIs, but I think counting UTF-8 bytes is reasonable and prevents abuse. RFC limit mentions octets, which means (encoded) bytes. I believe the other string formats we check length against (eg, handles, NSIDs) are required to be ASCII so all the flavors of string length will be the same for those (unlike the `uri` format which is more flexible).

I'm kind of ambivalent about 8000 vs 8192. The RFC recommendation is:

> It is RECOMMENDED that all senders and recipients support, at a minimum, URIs with lengths of 8000 octets in protocol elements. 

I'd be fine doing 8000 instead, but if we do that we should update the spec and go SDK (consistency is most important). I'm kind of inclined to just go ahead with 8192.

We definitely should not validate these as URLs; AT-URIs are not considered valid URLs! And by many parsers not valid IETF URIs, for that matter.

The `uri` string format is much more informal than the other formats; it is really expressing intent as much as anything. But it is important that we lock down the overall length so folks don't stuff 50k or 200k byte strings in there (for example)

Issue #3846, opened 2025-05-10
DPoP Spec Violation: atproto oauth clients include query parameters in `htu` field of DPoP JWT


iameli:
**Describe the bug**

Here's a DPoP JWT produced by `@atproto/oauth-client-browser`:

```
eyJhbGciOiJFUzI1NiIsInR5cCI6ImRwb3Arand0IiwiandrIjp7Imt0eSI6IkVDIiwiY3J2IjoiUC0yNTYiLCJ4IjoiM2dLTkV3Q3RkemJOakxDdDVzTGp0aHU2LUF3VWlaRl9McVdIZmdGVUc2TSIsInkiOiJLR01kQlpNNzR6NUlXNTdwcTAwWi1DYm9ZTnRSTF9qRTM1aUZ3WmFHMFZFIn0sImtpZCI6IjEwMzAxYzk5LTljZTItNDg2NS04NDRlLTgxZjIwMjk2ZDE2YSJ9.eyJpc3MiOiJodHRwczovL3N0cmVhbS5wbGFjZS9hcGkvYXRwcm90by1vYXV0aC93ZWIiLCJpYXQiOjE3NDY4NDQxNDMsImp0aSI6ImowbnVqdDB1bXRkIiwiaHRtIjoiR0VUIiwiaHR1IjoiaHR0cHM6Ly9taWxrY2FwLnVzLXdlc3QuaG9zdC5ic2t5Lm5ldHdvcmsveHJwYy9hcHAuYnNreS5hY3Rvci5nZXRQcm9maWxlP2FjdG9yPWRpZCUzQXBsYyUzQWRraDRyd2FmZGNkYTRrbzdsZXdlNDNtbCIsIm5vbmNlIjoiR05BSlhEWEpxS20yelhob3pwUmc4UTFEWGxWVEU0eGdMTEpzaXg4RDlqZyIsImF0aCI6ImhlMTEyamxvMlFYNXJZMUktT21fbk5Sd3VRcjVnN0x5V19lWGNXZlpNSlkifQ.M9Waa_v6XJnEUUWWuEaRLkCJlbSpbVjqdbtm4eZTfVH4Wu0rtKwYDfRh0CFUzmIdFwDg2sQQiLfDEEbU16BBYA
```

Decoded body:
```json
{
  "iss": "https://stream.place/api/atproto-oauth/web",
  "iat": 1746844143,
  "jti": "j0nujt0umtd",
  "htm": "GET",
[...]

PR #3879, opened 2025-05-21
Deprecate query & fragment in DPoP proof `htu` claim

matthieusieben:
This PR deprecates the use of query & fragment in DPoP "htu" proofs and paves the way for removing support altogether.

This PR also makes the DPoP validation result (`dpopResult`) available in the return value from `authenticateRequest` allowing PDS to log the use of invalid `htu` claims.

Related to: https://github.com/bluesky-social/atproto/issues/3846
Related to: https://github.com/bluesky-social/atproto/commit/36d0d370c24498f74c243ebfb01564e5050c672d (client side fix)

### DIFF
--- a/packages/oauth/oauth-provider/src/dpop/dpop-manager.ts
+++ b/packages/oauth/oauth-provider/src/dpop/dpop-manager.ts
@@ -1,9 +1,11 @@
 import { createHash } from 'node:crypto'
 import { EmbeddedJWK, calculateJwkThumbprint, errors, jwtVerify } from 'jose'
 import { z } from 'zod'
+import { ValidationError } from '@atproto/jwk'
 import { DPOP_NONCE_MAX_AGE } from '../constants.js'
 import { InvalidDpopProofError } from '../errors/invalid-dpop-proof-error.js'
 import { UseDpopNonceError } from '../errors/use-dpop-nonce-error.js'
+import { ifURL } from '../lib/util/cast.js'
 import {
   DpopNonce,
   DpopSecret,
@@ -58,27 +60,15 @@ export class DpopManager {
       throw new TypeError('HTTP method is required')
     }
 
-    const proof =
-      Array.isArray(httpHeaders['dpop']) && httpHeaders['dpop'].length === 1
-        ? httpHeaders['dpop'][0]!
-        : httpHeaders['dpop']
-
-    if (proof === undefined) {
-      return null
-    }
-
-    if (!proof || typeof proof !== 'string') {
-      throw new InvalidDpopProofError('DPoP proof required')
-    }
+    const proof = extractProof(httpHeaders)
+    if (!proof) return null
 
     const { protectedHeader, payload } = await jwtVerify(proof, EmbeddedJWK, {
       typ: 'dpop+jwt',
       maxTokenAge: 10, // Will ensure presence & validity of "iat" claim
       clockTolerance: DPOP_NONCE_MAX_AGE / 1e3,
     }).catch((err) => {
-      throw err instanceof JOSEError
-        ? new InvalidDpopProofError(err.message, err)
-        : new InvalidDpopProofError('Invalid DPoP proof', err)
+      throw newInvalidDpopProofError('Failed to verify DPoP proof', err)
     })
 
     // @NOTE For legacy & backwards compatibility reason, we cannot use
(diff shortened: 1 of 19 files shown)


packages/oauth/oauth-provider/src/oauth-verifier.ts
```diff
@@ -135,35 +138,37 @@ export class OAuthVerifier {
...
-        method,
-        url,
+      const dpopResult = await this.checkDpopProof(
+        httpMethod,
+        httpUrl,
+        httpHeaders,
         token,
       )
 
-      if (tokenType === 'DPoP' && !dpopJkt) {
```
matthieusieben, line 162: > [!NOTE]
> This check is covered within verifyToken

packages/oauth/oauth-provider/src/dpop/dpop-manager.ts
```diff
@@ -47,110 +48,153 @@ export class DpopManager {
...
+        ? httpHeaders['dpop'][0]!
+        : httpHeaders['dpop']
+
+    if (proof === undefined) {
+      return null
     }
 
     if (!proof || typeof proof !== 'string') {
       throw new InvalidDpopProofError('DPoP proof required')
     }
```
devinivy: Should we offer a different error message if `proof` is an array?  That would indicate multiple proofs were received.

packages/pds/src/auth-verifier.ts
```diff
@@ -490,19 +491,32 @@ export class AuthVerifier {
...
+        await this.oauthVerifier.authenticateRequest(
+          req.method || 'GET',
+          url,
+          req.headers,
+          { audience: [this.dids.pds] },
+        )
+
+      // @TODO drop this once oauth provider no longer accepts DPoP proof with
+      // query or fragment in "htu" claim.
+      if (dpopResult?.htu.match(/[?#]/)) {
```
devinivy: Can we get away with only logging in entryway and not worrying about the pdses?  I _think_ the token endpoint there should be comprehensive.

packages/oauth/oauth-provider/src/token/verify-token-claims.ts
```diff
@@ -17,48 +18,50 @@ export type VerifyTokenClaimsResult = {
...
   if (expectedTokenType !== tokenType) {
     throw new InvalidTokenError(expectedTokenType, `Invalid token type`)
   }
-  if (tokenType === 'DPoP' && !dpopJkt) {
-    throw new InvalidDpopProofError(`jkt is required for DPoP tokens`)
+  if (tokenType === 'DPoP' && !dpopResult) {
+    throw new InvalidDpopProofError(`DPoP proof required`)
   }
-  if (claimsJkt !== dpopJkt) {
+  if (tokenClaims.cnf?.jkt !== dpopResult?.jkt) {
```
devinivy: This looks a little bit like the intent is to permit a match when they are both missing, though I expect that isn't the case.  And the fact that these were both `null` and are now both `undefined` when missing is messing with me a little.  Would it make sense to explicitly rule out the missing cases?
```ts
if (tokenType === 'DPoP') {
  assert(tokenClaims.cnf?.jkt, 'cnf.jkt claim must exist via expected token type check')
  if (!dpopResult) {
    throw new InvalidDpopProofError(`DPoP proof required`)
  }
  if (tokenClaims.cnf.jkt !== dpopResult.jkt) {
    throw new InvalidDpopKeyBindingError()
  }
}
```

packages/oauth/oauth-provider/src/dpop/dpop-manager.ts
```diff
@@ -47,110 +48,153 @@ export class DpopManager {
...
-        err instanceof JOSEError ? err.message : 'Failed to calculate jkt'
-      throw new InvalidDpopProofError(message, err)
-    }
+    // @NOTE we can assert there is a jwk because the jwtVerify used the
+    // EmbeddedJWK key getter mechanism.
+    const jwk = protectedHeader.jwk!
+    const jkt = await calculateJwkThumbprint(jwk, 'sha256').catch((err) => {
+      throw err instanceof JOSEError
+        ? new InvalidDpopProofError(err.message)
+        : new InvalidDpopProofError('Failed to calculate jkt', err)
```
devinivy: There's a lot going on here, I'd recommend a more explicit construction.
```suggestion
      if (err instanceof JOSEError) {
        throw new InvalidDpopProofError(err.message)
      } else {
        throw new InvalidDpopProofError('Failed to calculate jkt', err)
      }
```

packages/oauth/oauth-provider/src/dpop/dpop-manager.ts
```diff
@@ -47,110 +48,153 @@ export class DpopManager {
...
-      requiredClaims: ['iat', 'jti'],
     }).catch((err) => {
-      const message =
-        err instanceof JOSEError
-          ? `Invalid DPoP proof (${err.message})`
-          : 'Invalid DPoP proof'
-      throw new InvalidDpopProofError(message, err)
+      throw err instanceof JOSEError
+        ? new InvalidDpopProofError(err.message, err)
+        : new InvalidDpopProofError('Invalid DPoP proof', err)
```
devinivy: Would also recommend a more explicit construction here.

packages/oauth/oauth-provider/src/dpop/dpop-manager.ts
```diff
@@ -47,110 +48,153 @@ export class DpopManager {
...
+
+  const endIdx =
+    queryIdx === -1
+      ? fragmentIdx
+      : fragmentIdx === -1
+        ? queryIdx
+        : Math.min(queryIdx, fragmentIdx)
+
+  return endIdx === -1 ? urlNorm : urlNorm.slice(0, endIdx)
+}
```
devinivy: Where does `url.origin + url.pathname` fall down?

PR #3881, opened 2025-05-21
Add expanded moderation report reasons

estrattonbailey:
_See [published proposal ](https://github.com/bluesky-social/proposals/tree/main/0009-mod-report-granularity)and [RFC discussion thread](https://github.com/bluesky-social/atproto/discussions/3880)._ 

This PR adds lexicon definitions for an expanded set of reporting reasons for labelers. The goal of this change is to provide users and moderators with additional granularity to aid in the processing of reports. For users, we want to ensure their concerns are represented in the options. For moderators, we want to give them as much context as possible.

We anticipate presenting users with a multi-step flow where they choose a category and subsequently a single report reason to include in their report.

As you can see, these new definitions are being made on the `tools.ozone.*` namespace instead of the lower-level `com.atproto.*` namespace. See the linked RFC for more information.

A couple of key observations:
- Not all report reasons will be forwarded to community labelers. Broadly this is on the basis of reports on illegal content which is sensitive enough that uncareful investigation could pose a legal risk. The specific reasons which are not forwarded are listed in our RFC discussion thread [TODO].
- We are not replacing or deprecating our existing report reasons with this expanded vocabulary. Instead, we’re asking developers to adopt the new report reasons where possible or otherwise applicable. Existing report reasons will continue to function. Notes will be added to the lexicons providing preferred usage as an optional, but recommended, migration path.

A list of existing report reasons and preferred equivalents:
- `com.atproto.moderation.defs#reasonSpam` → `tools.ozone.report.defs#reasonMisleadingSpam`
- `com.atproto.moderation.defs#reasonViolation` → `tools.ozone.report.defs#reasonRuleOther`
- `com.atproto.moderation.defs#reasonMisleading` → `tools.ozone.report.defs#reasonMisleadingOther`
- `com.atproto.moderation.defs#reasonSexual` → `tools.ozone.report.defs#reasonSexualUnlabeled`
- `com.atproto.moderation.defs#reasonRude` → `tools.ozone.report.defs#reasonHarassmentOther`
- `com.atproto.moderation.defs#reasonOther` → `tools.ozone.report.defs#reasonRuleOther`

Please refer to the [RFC discussion thread](https://github.com/bluesky-social/atproto/discussions/3880) for questions and comments.

### DIFF
--- a/packages/ozone/src/db/migrations/20250718T150931000Z-update-appeal-reason-stats.ts
+++ b/packages/ozone/src/db/migrations/20250718T150931000Z-update-appeal-reason-stats.ts
@@ -8,6 +8,8 @@ import * as recordEventsStats from '../schema/record_events_stats'
 export async function up(db: Kysely<any>): Promise<void> {
   // Drop and recreate materialized views to update appeal reason counting
   // to include both REASONAPPEAL and OZONE_APPEAL_REASON_TYPE
+  // The primary difference between the old and new query is that we were using = and != operators
+  // to match against the meta->>'reportType' field and now we use IN and NOT IN
 
   // Drop existing materialized views in reverse dependency order
   await db.schema


packages/ozone/src/config/config.ts
```diff
@@ -15,6 +15,7 @@ export const envToCfg = (env: OzoneEnvironment): OzoneConfig => {
     did: env.serverDid,
     version: env.version,
     devMode: env.devMode,
+    serviceRecordCacheTTL: env.serviceRecordCacheTTL ?? 5 * 60 * 1000, // default 5 mins
```
devinivy: The other TTLs in here are using the units from @atproto/common—may be nice to unify on.  (Also pretty self-documenting!)
```suggestion
    serviceRecordCacheTTL: env.serviceRecordCacheTTL ?? 5 * MINUTE,
```

packages/ozone/src/api/report/createReport.ts
```diff
@@ -21,21 +18,24 @@ export default function (server: Server, ctx: AppContext) {
...
         throw new ForbiddenError('You cannot appeal this report')
       }
 
       const db = ctx.db
 
-      await assertValidReporter(ctx.modService(db), reasonType, requester)
+      await Promise.all([
+        assertValidReporter(ctx.modService(db), reasonType, requester),
+        ctx.moderationServiceProfile().validateReasonType(reasonType),
+      ])
```
devinivy, line 30: Only slightly funny thing about doing these in parallel is that if both are invalid it's not deterministic which error you'd receive.

packages/ozone/src/context.ts
```diff
@@ -139,6 +144,10 @@ export class AppContext {
     const settingService = SettingService.creator()
     const verificationService = VerificationService.creator()
     const verificationIssuer = VerificationIssuer.creator()
+    const moderationServiceProfile = ModerationServiceProfile.creator(
+      cfg,
+      appviewAgent,
+    )
```
devinivy, line 156: Any special reason why this isn't a part of the existing `modService`?  No biggie either way, mostly just curious.

packages/ozone/src/mod-service/profile.ts
```diff
@@ -0,0 +1,141 @@
...
+export class ModerationServiceProfile {
+  private cache: CacheEntry | null = null
+  private CACHE_TTL: number
+
+  constructor(
+    private cfg: OzoneConfig,
+    private appviewAgent: AtpAgent,
+    cacheTTL?: number,
+  ) {
+    this.CACHE_TTL = cacheTTL || cfg.service.serviceRecordCacheTTL || 5 * MINUTE
```
devinivy: There are defaults for this in a couple places, perhaps we could just use a single one.  I think that `cfg.service.serviceRecordCacheTTL` is always present, so perhaps no need to fall through to the default here at all.

packages/ozone/src/config/config.ts
```diff
@@ -124,6 +125,7 @@ export type ServiceConfig = {
   did: string
   version?: string
   devMode?: boolean
+  serviceRecordCacheTTL?: number // in ms, default 5 mins
```
devinivy: This will always be set with default applied, so we can type it that way:
```suggestion
  serviceRecordCacheTTL: number // in ms
```

packages/ozone/src/mod-service/profile.ts
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  static creator(
+    cfg: OzoneConfig,
+    appviewAgent: AtpAgent,
+  ): ModerationServiceProfileCreator {
+    return () => new ModerationServiceProfile(cfg, appviewAgent)
+  }
+
+  async validateReasonType(reasonType: string): Promise<string> {
```
devinivy: This method is a bit difficult to read through because the caching logic and other logic are mixed together.  I think we could improve the factoring a bit by extracting the caching details into their own method.

```ts
    const profile = await this.getProfile() // All caching logic is in here
    const supportedReasonTypes = profile?.reasonTypes
    if (
      !Array.isArray(supportedReasonTypes) ||
      supportedReasonTypes.includes(reasonType) ||
      supportedReasonTypes.includes(NEW_TO_OLD_REASON_MAPPING[reasonType])
    ) {
      return reasonType
    }
    throw new InvalidRequestError(`Invalid reason type: ${reasonType}`)
```

packages/ozone/src/mod-service/status.ts
```diff
@@ -333,7 +333,7 @@ export const adjustModerationSubjectStatus = async (
 
   const isAppealEvent =
     action === 'tools.ozone.moderation.defs#modEventReport' &&
-    meta?.reportType === REASONAPPEAL
+    isAppealReport(`${meta?.reportType}`)
```
devinivy: It's a little awkward that the string `'undefined'` gets passed through.  Could `isAppealReport()` accept the argument as `unknown` or `undefined`?

packages/ozone/src/db/migrations/20250618T150931000Z-update-appeal-reason-stats.ts
```diff
@@ -0,0 +1,309 @@
+import { Kysely, sql } from 'kysely'
+import { OZONE_APPEAL_REASON_TYPE } from '../../api/util'
+import { REASONAPPEAL } from '../../lexicon/types/com/atproto/moderation/defs'
+import { DatabaseSchemaType } from '../schema'
+import * as modEvent from '../schema/moderation_event'
+import * as recordEventsStats from '../schema/record_events_stats'
+
+export async function up(db: Kysely<any>): Promise<void> {
```
devinivy: When I read through this I was trying to identify the "diff" from the current materialized view, I think we could leave some comments that make it clearer what the specific differences are.

packages/ozone/src/db/migrations/20250618T150931000Z-update-appeal-reason-stats.ts
```diff
@@ -0,0 +1,309 @@
+import { Kysely, sql } from 'kysely'
```
devinivy: I think we just need to plan the migration path for ourselves and self-hosters.  E.g. what is the impact of this migration as written?

Also, do we need to do any perf testing of the updated materialized views or are we already pretty high confidence that queries wont be significantly affected.

What did the alternative path of migrating the underlying data look like as an alternative?

devinivy: Just double-checking: are there any partial/covering indexes that exist based around the old report reason values?

packages/ozone/src/api/report/createReport.ts
foysalit: true but i think it's worth the perf gain? i dont think it matters a lot to the end user if both were to fail.

packages/ozone/src/context.ts
foysalit: mostly thinking that 
1. it doesn't feel like it belongs in that class
2.  i can see us needing it elsewhere without all the baggage of the modService instance

packages/ozone/src/mod-service/status.ts
foysalit: it already accepts undefined but meta properties are losely typed to allow numbers, booleans etc. so casting it to string is kinda necessary unless we open up the input type to all of those which doesn't make a lot of sense. i guess we could just add an undefined check first and then pass as string?

packages/ozone/src/db/migrations/20250618T150931000Z-update-appeal-reason-stats.ts
foysalit: good shout. i wasn't aware of any but double checked the db and here are all the indices we have 

| Index Name | Type | Definition |
|------------|------|------------|
| `moderation_event_pkey` | PRIMARY KEY | `CREATE UNIQUE INDEX moderation_event_pkey ON public.moderation_event USING btree (id)` |
| `moderation_event_created_at_idx` | INDEX | `CREATE INDEX moderation_event_created_at_idx ON public.moderation_event USING btree ("createdAt")` |
| `moderation_event_subject_did_idx` | INDEX | `CREATE INDEX moderation_event_subject_did_idx ON public.moderation_event USING btree ("subjectDid")` |
| `moderation_event_mod_tool_name_idx` | INDEX | `CREATE INDEX moderation_event_mod_tool_name_idx ON public.moderation_event USING btree ((("modTool" ->> 'name'::text)))` |
| `moderation_event_account_reports_idx` | CONDITIONAL INDEX | `CREATE INDEX moderation_event_account_reports_idx ON public.moderation_event USING btree ("createdBy", "subjectDid", "createdAt") WHERE (("subjectUri" IS NULL) AND ((action)::text = 'tools.ozone.moderation.defs#modEventReport'::text))` |
| `moderation_event_record_reports_idx` | CONDITIONAL INDEX | `CREATE INDEX moderation_event_record_reports_idx ON public.moderation_event USING btree ("createdBy", "subjectDid", "subjectUri", "createdAt") WHERE (("subjectUri" IS NOT NULL) AND ((action)::text = 'tools.ozone.moderation.defs#modEventReport'::text))` |
| `moderation_event_account_actions_ids` | CONDITIONAL INDEX | `CREATE INDEX moderation_event_account_actions_ids ON public.moderation_event USING btree ("subjectDid", action, "createdAt") WHERE (("subjectUri" IS NULL) AND ((action)::text = ANY ((ARRAY['tools.ozone.moderation.defs#modEventTakedown'::character varying, 'tools.ozone.moderation.defs#modEventLabel'::character varying])::text[])))` |
| `moderation_event_message_id_idx` | CONDITIONAL INDEX | `CREATE INDEX moderation_event_message_id_idx ON public.moderation_event USING btree ("subjectMessageId") WHERE (("subjectMessageId" IS NOT NULL) AND ((action)::text = 'tools.ozone.moderation.defs#modEventReport'::text))` |
| `moderation_event_record_actions_ids` | CONDITIONAL INDEX | `CREATE INDEX moderation_event_record_actions_ids ON public.moderation_event USING btree ("subjectDid", "subjectUri", action, "createdAt") WHERE (("subjectUri" IS NOT NULL) AND ((action)::text = ANY ((ARRAY['tools.ozone.moderation.defs#modEventTakedown'::character varying, 'tools.ozone.moderation.defs#modEventLabel'::character varying])::text[])))` |


we only index the action column and not the report reasons which is stored under JSON meta column

packages/ozone/src/db/migrations/20250618T150931000Z-update-appeal-reason-stats.ts
foysalit: i was hoping the diff is covered by the first comment 
>   // Drop and recreate materialized views to update appeal reason counting
>   // to include both REASONAPPEAL and OZONE_APPEAL_REASON_TYPE

that's all it does really. previously, we were counting all reports with appeal type as an appeal, not we just need to include a new type as appeal so instead of an equality check with `=/!=` against report types, we are now doing `IN/NOT IN` operators. let me know if the updated comment will suffice?

packages/ozone/src/db/migrations/20250618T150931000Z-update-appeal-reason-stats.ts
foysalit: the primary change being `IN` operator instead of `=` and `NOT IN` operator instead of `!=`, i think the perf diff shouldn't be noteworthy, wdyt?

packages/ozone/tests/report-reason.test.ts
```diff
@@ -0,0 +1,154 @@
...
+        rkey: 'self',
+        record: {
+          policies: { labelValues: [] },
+          reasonTypes: [REASONSPAM, REASONRUDE],
+          createdAt: new Date().toISOString(),
+        },
+      })
+      await network.processAll()
+
+      await new Promise((resolve) => setTimeout(resolve, 500))
```
devinivy, line 133: There is a `wait()` util in `@atproto/common` that can help here.

devinivy (approved):

PR #3882, opened 2025-05-22
Referred likes lexicon changes

mozzius:
- [x] Add `via` field to like
- [x] Add `via` field to repost
- [x] Update utility methods in `Agent`
- [x] Add `uri` and `cid` to `#reasonRepost`
- [x] Add `like-via-repost` and `repost-via-repost` to notification reasons
- [x] Serve `uri` and `cid` in `reasonRepost` view


lexicons/app/bsky/notification/listNotifications.json
```diff
@@ -71,7 +71,9 @@
             "quote",
             "starterpack-joined",
             "verified",
-            "unverified"
+            "unverified",
+            "referred-like",
```
jcalabro, line 75: `via-like` and `via-repost`? Not actually sure. Might be nice to keep the verb consistent?

mozzius: ah yeah, maybe this is a good reason to switch back to `referrer` because `via-like` is a bit strange

Unless we do `like-via-repost` and `repost-via-repost`?

jcalabro: Happy to take your suggestion! I'm personally down with `like-via-repost` and `repost-via-repost`

mozzius: I'll go with `like-via-repost`, I think that's the clearest of our 3 options

mary-ext:
shouldn't `listNotifications` return a `viaRepost` field instead so that out-of-date clients could still display referred reposts as a regular repost?

lexicons/app/bsky/feed/defs.json
```diff
@@ -98,6 +98,8 @@
       "required": ["by", "indexedAt"],
       "properties": {
         "by": { "type": "ref", "ref": "app.bsky.actor.defs#profileViewBasic" },
+        "uri": { "type": "string", "format": "at-uri" },
+        "cid": { "type": "string", "format": "cid" },
```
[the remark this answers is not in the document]
mozzius: These are for making a strongRef to the repost record when making a like/repost record

mozzius:
@mary-ext decided against it - the new notification is meaningfully different enough from the old one that such fallback behaviour would be a breaking change for old clients - showing the new notif type should be opt-in

lexicons/app/bsky/feed/defs.json
```diff
@@ -98,6 +98,8 @@
       "required": ["by", "indexedAt"],
```
mozzius, line 98: Made the new fields not required - to be determined if that's the right move

PR #3886, opened 2025-05-23
Refactor route rate limiter builder

matthieusieben:
Currently, `setupRouteRateLimits` creates an entry in `Server.routeRateLimiters` that is then used (only) from `createHandler`. This PR changes the behavior by not using an instance property as intermediary variable storage between two distinct methods.

This PR also addresses the following issues:

 - `keyPrefix` being the same for every endpoint defining custom rate limits (caused by [this](https://github.com/bluesky-social/atproto/blob/d1e3e68dd9eb7bed13d9023bc0e4ce3c448eabf5/packages/xrpc-server/src/server.ts#L470))
 - The global rate limiters not being applied if local rate limiter are used (cause by [this](https://github.com/bluesky-social/atproto/blob/d1e3e68dd9eb7bed13d9023bc0e4ce3c448eabf5/packages/xrpc-server/src/server.ts#L448))
 - The bypass condition is now checked once instead of for every individual limiter (for every request), through a callback function (instead of hard coded secret / ip logic).


packages/xrpc-server/src/server.ts
```diff
@@ -432,76 +403,35 @@ export class Server {
...
-                  calcKey,
-                  calcPoints,
-                }),
-              reset: (ctx: XRPCReqContext) =>
-                rateLimiter.reset(ctx, {
-                  calcKey,
-                }),
-            })
-          }
+          return rateLimits.creator({ ...options, keyPrefix: `${nsid}-${i}` })
```
matthieusieben, line 427: Fixes an issue where the same prefix (`"nsid-0"`, `"nsid-1"`, ...) was used for every method defining a custom `rateLimit` config.

https://github.com/bluesky-social/atproto/blob/d1e3e68dd9eb7bed13d9023bc0e4ce3c448eabf5/packages/xrpc-server/src/server.ts#L470

packages/xrpc-server/src/server.ts
```diff
@@ -432,76 +405,41 @@ export class Server {
...
+      if (isShared(options)) {
+        const rateLimiter = this.sharedRateLimiters.get(options.name)
+
+        // The route config references a shared rate limiter that does not
+        // exist. This is a configuration error.
+        assert(rateLimiter, `Shared rate limiter "${options.name}" not found`)
+
+        return updateRateLimiterOptions(rateLimiter, options)
+      } else {
+        return rateLimits.creator({ ...options, keyPrefix: `${nsid}-${i}` })
```
matthieusieben, line 431: Fixes an issue where the same prefix (`"nsid-0"`, `"nsid-1"`, ...) was used for every method defining a custom `rateLimit` config.

https://github.com/bluesky-social/atproto/blob/d1e3e68dd9eb7bed13d9023bc0e4ce3c448eabf5/packages/xrpc-server/src/server.ts#L470

packages/xrpc-server/src/rate-limiter.ts
```diff
@@ -142,48 +129,128 @@ export const formatLimiterStatus = (
...
+    reset: async (ctx, opts) => {
+      const promises: ReturnType<RateLimiterReset>[] = []
+      for (const rl of rateLimiters) promises.push(rl.reset(ctx, opts))
+      await Promise.all(promises)
+    },
+  }
+}
+
+export type RouteRateLimiterOptions = {
+  bypass?: (ctx: XRPCReqContext) => boolean
```
matthieusieben, line 214: This allows to `bypass` the whole rate limiting check, once per route / per request (instead of each individual limiter previously computing the same bypass checks).

This also allows the user of the `xrpc-server` lib to define their own bypass strategy since the `x-ratelimit-bypass` is not part of the spec.


Hidden motive: remove dependency on Express by moving Express dependent logic (`req.ip`) outside of this module.

packages/xrpc-server/src/rate-limiter.ts
```diff
@@ -64,17 +62,8 @@ export class RateLimiter implements RateLimiterI {
...
   ): Promise<RateLimiterStatus | RateLimitExceededError | null> {
-    if (
-      this.bypassSecret &&
-      ctx.req.header('x-ratelimit-bypass') === this.bypassSecret
-    ) {
-      return null
-    }
-    if (this.bypassIps && this.bypassIps.includes(ctx.req.ip)) {
-      return null
-    }
```
matthieusieben, line 77: Instead of each individual consumer computing the same thing; this was replaced with a `bypass` option on the "root limiter", computed only once per request (instead of once per request per limiter).

It is still possible to implement a specific consumer level bypass by having the `calcKey` function return `null`.

packages/xrpc-server/src/server.ts
```diff
@@ -432,76 +411,43 @@ export class Server {
...
+    config: XRPCHandlerConfig,
+  ): RouteRateLimiter | undefined {
+    // No route specific rate limiting configured, use the global rate limiter.
+    if (!config.rateLimit) return this.globalRateLimiter
 
-    if (config.rateLimit) {
-      const limits = Array.isArray(config.rateLimit)
-        ? config.rateLimit
-        : [config.rateLimit]
-      this.routeRateLimiters[nsid] = []
```
matthieusieben, line 447: this line prevents global limiters (set above) from being applied if a route specific `rateLimit` is set. If that route specific config is an empty array, no rate limit at all would be set.

This PR changes that behavior:
- By returning the global rate limiter if the route has none
- By always combining the global rate limiter with the route specific ones.

packages/xrpc-server/src/server.ts
```diff
@@ -432,76 +411,43 @@ export class Server {
...
+
+    const { creator, bypass } = rateLimits
+
+    const rateLimiters = asArray(config.rateLimit).map((options, i) => {
+      if (isShared(options)) {
+        const rateLimiter = this.sharedRateLimiters?.get(options.name)
+
+        // The route config references a shared rate limiter that does not
+        // exist. This is a configuration error.
+        assert(rateLimiter, `Shared rate limiter "${options.name}" not defined`)
```
matthieusieben, line 434: This is a breaking change: Before, missing shared limiters would silently be ignored.

packages/xrpc-server/src/server.ts
```diff
@@ -177,7 +178,6 @@ export class Server {
       middleware.push(this.middleware.json)
       middleware.push(this.middleware.text)
     }
-    this.setupRouteRateLimits(nsid, config)
```
matthieusieben, line 180: This line was particularly error prone:
- `setupRouteRateLimits` must be called *before* `createHandler` otherwise rate limits would silently be ignored
- `setupRouteRateLimits` sets a value in `this.routeRateLimiters[nsid]`. That value is read from `createHandler` and never used again.

packages/pds/src/index.ts
```diff
@@ -91,9 +94,25 @@ export class PDS {
 
         return XRPCError.fromError(err)
       },
-      rateLimits: ctx.ratelimitCreator
+      rateLimits: rateLimits.enabled
         ? {
-            creator: ctx.ratelimitCreator,
+            creator:
+              rateLimits.mode === 'redis'
+                ? (opts) => RateLimiter.redis(ctx.redisScratch!, opts)
```
devinivy: What happens in case of a misconfiguration, where `rateLimits.mode === 'redis'` but `ctx.redisScratch` isn't configured?

devinivy (approved):

packages/pds/src/index.ts
matthieusieben: There is an assert in the config parsing but you're right

PR #3887, opened 2025-05-23
Add Generic Bsync Proto

jcalabro:
Idea here is to use arbitrary key-value style data in bsync, modeled loosely after existing atproto shapes. Will leave this up to @rafaelbsky to codegen, but we live coded on these protobuf definitions together.

### DIFF
--- a/packages/bsync/proto/bsync.proto
+++ b/packages/bsync/proto/bsync.proto
@@ -67,6 +67,38 @@ message ScanNotifOperationsResponse {
 }
 
 
+enum Method {
+  METHOD_UNSPECIFIED = 0;
+  METHOD_CREATE = 1;
+  METHOD_UPDATE = 2;
+  METHOD_DELETE = 3;
+}
+
+message Operation {
+  string collection = 1;
+  string actor_did = 2;
+  string id = 3;
+  Method method = 4;
+  bytes payload = 5;
+}
+
+message PutOperationRequest {
+  Operation operation = 1;
+}
+
+message PutOperationResponse {
+}
+
+message ScanOperationsRequest {
+  string cursor = 1;
+  int32 limit = 2;
+}
+
+message ScanOperationsResponse {
+  repeated Operation operations = 1;
+  string cursor = 2;
+}
+
 
 // Ping
 message PingRequest {}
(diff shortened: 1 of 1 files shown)


packages/bsync/proto/bsync.proto
```diff
@@ -67,6 +67,44 @@ message ScanNotifOperationsResponse {
...
+message Operation {
+  string id = 1;
+  string collection = 2;
+  string actor_did = 3;
+  string rkey = 4;
+  Method method = 5;
+  bytes payload = 6;
+}
+
+message PutOperationRequest {
```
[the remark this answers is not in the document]
jcalabro: `id` is not passed along on the request, so using `Operation` would be inappropriate. `id` is assigned by the server, not the client

packages/bsync/src/db/migrations/20250527T022203400Z-add-op.ts
```diff
@@ -0,0 +1,20 @@
+import { Kysely, sql } from 'kysely'
+
+export async function up(db: Kysely<unknown>): Promise<void> {
+  await db.schema
+    .createTable('op')
```
jcalabro, line 5: This may be a nit, but can we use the full table name `operation`?

jcalabro: I think my preference would be the full name `operation`. I'd be wondering what `\d` would yield in a postgres shell

packages/bsync/proto/bsync.proto
```diff
@@ -67,6 +67,44 @@ message ScanNotifOperationsResponse {
...
+  METHOD_CREATE = 1;
+  METHOD_UPDATE = 2;
+  METHOD_DELETE = 3;
+}
+
+message Operation {
+  string id = 1;
+  string collection = 2;
+  string actor_did = 3;
+  string rkey = 4;
```
devinivy, line 81: Small thought here.  If we're trying to evoke the structure of repositories with did/collection/rkey, we might consider the actors the top level of the hierarchy and list it first.
```suggestion
  string actor_did = 2;
  string collection = 3;
  string rkey = 4;
```

devinivy (approved):

PR #3892, opened 2025-05-26
:bug: Fetch record from pds if appview fails to find it for ozone

foysalit:
This should help ozone display records from custom lexicons.

### DIFF
--- a/packages/ozone/src/mod-service/views.ts
+++ b/packages/ozone/src/mod-service/views.ts
@@ -314,18 +314,19 @@ export class ModerationViews {
       return record
     } catch (err) {
       if (err instanceof RecordNotFoundError) {
-        const { agent: pdsAgent } = await getPdsAgentForRepo(
-          this.idResolver,
-          params.repo,
-          this.devMode,
-        )
-        if (!pdsAgent) {
-          return null
-        }
-
         // If pds fetch fails, just return null regardless of the error
         try {
-          return pdsAgent.com.atproto.repo.getRecord(params)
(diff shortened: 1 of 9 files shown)


packages/ozone/src/mod-service/views.ts
```diff
@@ -294,28 +300,57 @@ export class ModerationViews {
...
+  ) {
+    try {
+      const record = await this.appviewAgent.com.atproto.repo.getRecord(
+        params,
+        appviewAuth,
+      )
+      return record
+    } catch (err) {
+      if (
+        (err as Error).message?.startsWith('Record not found') &&
```
devinivy, line 316: There is a specific error type `RecordNotFound` for this in the `repo.getRecord` endpoint—these kinds of checks should be able to use that error code.  If the appview doesn't already use it, we should probably update it!

packages/ozone/src/mod-service/views.ts
```diff
@@ -1,7 +1,13 @@
 import { sql } from 'kysely'
-import { AppBskyActorDefs, AtpAgent } from '@atproto/api'
+import {
+  AppBskyActorDefs,
+  AtpAgent,
+  ComAtprotoRepoGetRecord,
+} from '@atproto/api'
+import { RecordNotFoundError } from '@atproto/api/dist/client/types/com/atproto/repo/getRecord'
```
devinivy, line 7: I think you should be able to avoid dipping into `dist/` and instead use `ComAtprotoRepoGetRecord.RecordNotFoundError`.
```suggestion
```

devinivy (approved):

PR #3893, opened 2025-05-26
fix: oauth footer links not working

Johannes-Andersen:
Fixes issue where https://bsky.social/account footer is only linking to https://bsky.social/account/sign-in

This is due to `@tanstack/react-router`'s Link component not liking the use of href to external (non router managed paths), `to` is used everywhere else to link to other router managed pages, so they work fine.

Before patch:
Screenshot 2025-05-26 at 23 39 07
After:
Screenshot 2025-05-26 at 23 42 57

Not that familiar with react-router, so there may be better ways to tell the Link component to be external

cc: @matthieusieben

### DIFF
--- a/.changeset/rude-shoes-applaud.md
+++ b/.changeset/rude-shoes-applaud.md
@@ -1,0 +1,5 @@
+---
+"@atproto/oauth-provider-frontend": patch
+---
+
+Fix oauth footer links not working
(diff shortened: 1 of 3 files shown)


matthieusieben:
Hey @Johannes-Andersen, thank you again for your interest in our project !

I've take a different approach in https://github.com/bluesky-social/atproto/pull/3934 so I'll close this one.

PR #3894, opened 2025-05-26
Add `app.bsky.unspecced.getPostThreadHiddenV2`

ghost:
This is a proposal on top of #3825 .

### DIFF
--- a/packages/bsky/src/views/index.ts
+++ b/packages/bsky/src/views/index.ts
@@ -1427,42 +1427,22 @@ export class Views {
     const childrenUris = childrenByParentUri[parentUri] ?? []
     let hasHiddenReplies = false
     const replies = mapDefined(childrenUris, (uri) => {
-      // Not found.
-      const post = state.posts?.get(uri)
-      if (post?.violatesThreadGate) {
+      const replyInclusion = this.checkThreadV2ReplyInclusion(
+        uri,
+        rootUri,
+        state,
+      )
+      if (!replyInclusion) {
         return undefined
       }
-      const postView = this.post(uri, state)
-      if (!post || !postView) {
-        return undefined
-      }
-      const authorDid = postView.author.did
-      if (rootUri !== getRootUri(uri, post)) {
-        // outside thread boundary
-        return undefined
-      }
-
-      // Blocked (1p and 3p for replies).
-      const has1pBlock = this.viewerBlockExists(authorDid, state)
-      const has3pBlock =
-        !state.ctx?.include3pBlocks && state.postBlocks?.get(uri)?.parent
-      if (has1pBlock || has3pBlock) {
-        return undefined
-      }
-      if (!this.viewerSeesNeedsReview({ uri, did: authorDid }, state)) {
-        return undefined
-      }
-
-      // No unauthenticated.
-      if (this.noUnauthenticatedPost(state, postView)) {
-        return undefined
-      }
+      const { authorDid, postView } = replyInclusion
 
       // Hidden.
       const { hiddenByThreadgate, mutedByViewer } = this.isHiddenThreadPost(
         { rootUri, uri },
         state,
       )
+      // Is hidden reply.
       if (hiddenByThreadgate || mutedByViewer) {
         // Only care about anchor replies
         if (depth === 1) {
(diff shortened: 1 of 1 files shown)


lexicons/app/bsky/unspecced/getPostThreadHiddenV2.json
```diff
@@ -0,0 +1,74 @@
...
+          "format": "at-uri"
+        },
+        "depth": {
+          "type": "integer",
+          "description": "The nesting level of this item in the thread. Depth 0 means the anchor item. Items above have negative depths, items below have positive depths."
+        },
+        "value": {
+          "type": "ref",
+          "ref": "#threadHiddenItemPost"
+        }
```
devinivy, line 55: I notice that the analogous property from `getPostThreadV2` was a union.  How likely is it that this will become a union?  If it's likely, we might get ahead of it now.  I suppose this is unspecced so we have some wiggle-room, low-stakes.

packages/bsky/src/api/app/bsky/unspecced/getPostThreadHiddenV2.ts
```diff
@@ -0,0 +1,118 @@
...
+  uris: string[]
+}
+
+const calculateBelow = (ctx: Context, anchor: string, below: number) => {
+  let maxDepth = ctx.cfg.maxThreadDepth
+  if (ctx.cfg.bigThreadUris.has(anchor) && ctx.cfg.bigThreadDepth) {
+    maxDepth = ctx.cfg.bigThreadDepth
+  }
+  return maxDepth ? Math.min(maxDepth, below) : below
+}
```
devinivy, line 118: I think this is shared by 2-3 endpoints now.  Could be nice to factor out somewhere.

packages/bsky/src/views/index.ts
```diff
@@ -1606,6 +1595,264 @@ export class Views {
...
+    state: HydrationState,
+    {
+      below,
+      branchingFactor,
+      viewer,
+    }: {
+      below: number
+      branchingFactor: number
+      viewer: HydrateCtx['viewer']
+    },
```
devinivy, line 1609: These places where `viewer` is passed as a separate parameter—I believe we should be able to get that out of `state.ctx` (and `state` is already passed).

packages/bsky/src/views/index.ts
```diff
@@ -1606,6 +1595,264 @@ export class Views {
...
+    {
+      rootUri,
+      uri,
+      viewer,
+    }: {
+      rootUri: string
+      uri: string
+      viewer: HydrateCtx['viewer']
+    },
+    state: HydrationState,
```
devinivy, line 1817: Another example here where I believe we can pull `viewer` from `state.ctx`.  I think there are others we could look at but figured I'd just tag a couple to illustrate.

packages/bsky/src/views/index.ts
```diff
@@ -1606,6 +1595,264 @@ export class Views {
...
+        return undefined
+      }
+      if (!this.viewerSeesNeedsReview({ uri, did: authorDid }, state)) {
+        return undefined
+      }
+
+      // No unauthenticated.
+      if (this.noUnauthenticatedPost(viewer, postView)) {
+        return undefined
+      }
```
devinivy, line 1708: There is a decent amount of common logic needed by both replies and hidden replies.  I wonder if it's worth leaving a note in `threadHiddenV2Replies()` and `threadV2Replies()` to help ensure it stays in sync.

packages/bsky/src/views/index.ts
```diff
@@ -1606,6 +1595,264 @@ export class Views {
...
+
+  private threadHiddenV2ItemPostAnchor({
+    depth,
+    uri,
+  }: {
+    depth: number
+    uri: string
+  }): ThreadHiddenAnchorPostNode['item'] {
+    return {
+      $type: 'app.bsky.unspecced.getPostThreadHiddenV2#threadHiddenItem',
```
devinivy, line 1765: Ideally we'd avoid placing this in the endpoint's output— a. partly just because it's more precise, but also b. because otherwise implementations can come to rely on it and break if we opt to remove it.

My understanding is that some code relies on this being present—maybe we should reflect that through the types.  I believe the property could be removed here without causing any type errors, so it could be hard to track down an issue here.  I think adding `$Typed` within `ThreadHiddenAnchorPostNode` might do the trick.

packages/bsky/src/views/threads-v2.ts
```diff
@@ -57,40 +92,50 @@ const isThreadNoUnauthenticatedNode = (
 const isThreadPostNode = (node: ThreadTree): node is ThreadPostNode =>
   AppBskyUnspeccedGetPostThreadV2.isThreadItemPost(node.item.value)
 
-export type ThreadTree =
+const isThreadHiddenPostNode = (
+  node: ThreadTree,
+): node is ThreadHiddenPostNode =>
+  AppBskyUnspeccedGetPostThreadHiddenV2.isThreadHiddenItem(node.item)
```
devinivy, line 98: Rather than duck-typing `node` based on its inner contents (e.g. `node.item.$type`), what if `ThreadTree` items themselves offered a form of reflection?  I wonder if this could reduce some dependencies on implementation details such as presence of `$type` in API output.  Just an idea, not critical.

E.g. `node.isThreadHiddenPostNode` and `node.isThreadHiddenPostNode`.

devinivy (approved):
Looking good 👍

PR #3900, opened 2025-05-28
Add max length limit to passwords

matthieusieben:


packages/pds/src/account-manager/helpers/password.ts
```diff
@@ -27,6 +27,15 @@ export const verifyAppPassword = async (
...
 ): Promise<AppPassDescript | null> => {
+  // Fast check for the password format
+  if (
+    password.length !== 19 ||
+    password[4] !== '-' ||
+    password[9] !== '-' ||
+    password[14] !== '-'
+  ) {
+    return null
+  }
```
devinivy, line 38: I am a bit concerned about this ending up being too strict and biting us, e.g. if the format changes in the future or has ever changed in the past.  I don't think it will harm us for for app password checks to be equally efficient to regular password checks.

packages/pds/src/account-manager/account-manager.ts
```diff
@@ -172,6 +172,10 @@ export class AccountManager {
     deactivated?: boolean
     refreshJwt?: string
   }) {
+    if (password && password.length > scrypt.NEW_PASSWORD_MAX_LENGTH) {
+      throw new InvalidRequestError('Password too long')
```
devinivy, line 176: How will these kinds of xrpc errors surface up to the oauth UI?  I wondered if that may rely on some different error class, or other explicit error handling.

devinivy (approved):

packages/pds/src/account-manager/account-manager.ts
matthieusieben: This is already handled here:

https://github.com/bluesky-social/atproto/blob/897ee3748bb40ae54970af3db5cb62ac479165de/packages/pds/src/account-manager/oauth-store.ts#L190-L196

matthieusieben: Also, the OAuth will already have validated that particular constraint through it's own Zod validation schema.

PR #3901, opened 2025-05-28
Notification preferences V2 endpoints

mozzius:
~~Note: codegen is broken for arrays of strings of known values~~ Fixed!

### DIFF
--- a/packages/bsync/tests/operations.test.ts
+++ b/packages/bsync/tests/operations.test.ts
@@ -16,13 +16,9 @@ describe('operations', () => {
   let bsync: BsyncService
   let client: BsyncClient
 
-  const validPayload0 = Buffer.from(
-    new TextEncoder().encode(JSON.stringify({ value: 0 })),
-  )
-  const validPayload1 = Buffer.from(
-    new TextEncoder().encode(JSON.stringify({ value: 1 })),
-  )
-  const invalidPayload = Buffer.from('{invalid json}', 'utf-8')
+  const validPayload0 = Buffer.from(JSON.stringify({ value: 0 }))
+  const validPayload1 = Buffer.from(JSON.stringify({ value: 1 }))
+  const invalidPayload = Buffer.from('{invalid json}')
 
   beforeAll(async () => {
     bsync = await BsyncService.create(


lexicons/app/bsky/notification/getPreferences.json
```diff
@@ -0,0 +1,89 @@
+{
+  "lexicon": 1,
+  "id": "app.bsky.notification.getPreferences",
+  "defs": {
+    "main": {
+      "type": "query",
+      "description": "Set notification-related preferences for an account. Requires auth.",
```
jcalabro, line 7: Typo? `Set` -> `Get`?

jcalabro (approved):
One note, otherwise I think this looks good. Might be worth getting a second opinion though as I'm far from being a lexicon guru

lexicons/app/bsky/notification/getPreferences.json
mozzius: Oops yep

lexicons/app/bsky/notification/putPreferences.json
```diff
@@ -4,7 +4,7 @@
   "defs": {
     "main": {
       "type": "procedure",
-      "description": "Set notification-related preferences for an account. Requires auth.",
+      "description": "DEPRECIATED: Use putPreferencesV2. Requires auth.",
```
surfdude29, line 7:
```suggestion
      "description": "DEPRECATED: Use putPreferencesV2. Requires auth.",
```
nit: I think `deprecated` might fit better in this context?

mozzius: Ah shoot I thought it looked wrong. Thanks

lexicons/app/bsky/notification/putPreferences.json
```diff
@@ -4,7 +4,7 @@
   "defs": {
     "main": {
       "type": "procedure",
-      "description": "Set notification-related preferences for an account. Requires auth.",
+      "description": "DEPRECATED: Use putPreferencesV2. Requires auth.",
```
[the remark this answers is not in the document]
mozzius: Sure

lexicons/app/bsky/notification/putPreferencesV2.json
```diff
@@ -0,0 +1,145 @@
...
+            "quoteNotification",
+            "starterpackJoinedNotification",
+            "verifiedNotification",
+            "unverifiedNotification",
+            "likeViaRepostNotification",
+            "repostViaRepostNotification",
+            "subscribedPostNotification",
+            "chatNotification"
+          ],
+          "properties": {
```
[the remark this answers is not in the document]
mozzius: Good points! I think the reason I opted for two different objects is that one has all required fields, and one has all optional fields. We can definitely unify them into just all-optional or all-required (maybe all-optional, and agree on what the default settings are)

lexicons/app/bsky/notification/defs.json
```diff
@@ -5,6 +5,23 @@
...
+      "type": "object",
+      "required": ["filter", "channels"],
+      "properties": {
+        "channels": {
+          "type": "array",
+          "items": {
+            "type": "string",
+            "knownValues": ["in-app", "push"]
+          }
+        },
```
mozzius, line 19: Hmm, more I think about it the less I like this. Should probably just be
```json
{
  "inApp": true,
  "push": false
}
```

lexicons/app/bsky/notification/defs.json
```diff
@@ -5,6 +5,23 @@
...
+        "filter": {
+          "type": "string",
+          "knownValues": ["all", "follows"]
```
ghost: Do we want to have a `none`?

mozzius: I was thinking turning off all the channels was equivalent

ghost: Yeah, that should be enough. I just found a bit weird but it would be weird any way around it.

lexicons/app/bsky/notification/putPreferencesV2.json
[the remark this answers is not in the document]
ghost: We decided for option 1. It is more aligned in the direction we'll go with BFF.

lexicons/app/bsky/notification/defs.json
ghost: I'll make the change, between other changes I'm working on.

ghost: How are those preferences stored currently? We need to think how to port the existing data.

packages/bsky/src/vault.ts
```diff
@@ -0,0 +1,57 @@
+import { BsyncClient } from './bsync'
+import { Method } from './proto/bsync_pb'
+
+export const createVaultClient = (bsyncClient: BsyncClient): VaultClient => {
+  return new VaultClient(bsyncClient)
+}
+
+// VaultClient is an abstraction over the BsyncClient, that uses the bsync `PutOperation` RPC
+// to store private data, which can be indexed by the dataplane and queried by the appview.
+export class VaultClient {
```
ghost: I'm not settled on this name, but I do like the idea of having a named abstraction for this. Suggestions are welcome (either naming or not having a separate concept at all).

packages/bsky/src/api/app/bsky/notification/putPreferencesV2.ts
```diff
@@ -0,0 +1,74 @@
...
+    auth: ctx.authVerifier.standard,
+    handler: async ({ auth, input }) => {
+      const actorDid = auth.credentials.iss
+      const [preferences, exists] = await computePreferences(
+        ctx,
+        actorDid,
+        input,
+      )
+
+      // @TODO: lexicon validation?
```
ghost: I intend to do this in this PR, I just didn't have time yet, and it won't prevent reviews.

packages/bsky/src/vault.ts
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  private async putOperation(method: Method, input: PutOperationInput) {
+    const { actorDid, namespace, key, payload } = input
+    await this.bsyncClient.putOperation({
+      actorDid,
+      namespace,
+      key,
+      method,
+      payload: payload ? Buffer.from(JSON.stringify(payload)) : undefined,
```
ghost: I should also add the validation that payload is JSON on bsync.

Want to confirm first whether this is the format we want to work with.

packages/bsky/proto/bsky.proto
```diff
@@ -753,6 +753,49 @@ message GetBlocklistSubscriptionsResponse {
...
+  bool push = 2;
+}
+
+message NotificationChannelsPush {
+  bool push = 1;
+}
+
+message NotificationPreferenceFull {
+  NotificationChannels channels = 1;
+  string filter = 2;
```
jcalabro: Could we make an enum for all instances of `filter` rather than using a string? It should have the values `all` and `follows` like the lexicon

packages/bsky/proto/bsky.proto
```diff
@@ -753,6 +753,49 @@ message GetBlocklistSubscriptionsResponse {
...
+message GetNotificationPreferencesRequest {
+  string actor_did = 1;
+}
+
+message NotificationChannels {
+  bool list = 1;
+  bool push = 2;
+}
+
+message NotificationChannelsPush {
```
jcalabro: This structure took me kind of a long time to grok to figure out what the intent was, though I guess it makes sense

I wonder if this would be preferable. It leaves us more room in the future to add values to each individual notification channel rather than having them just be a bool. It is more self-documenting. It's also more composable because the `chat = 1;` preference on the response shape could embed the `NotificationChannelPush` exactly the same as `NotificationChannels` rather than having it be a bespoke type.

```proto
message NotificationChannels {
  NotificationChannelList list = 1;
  NotificationChannelPush push = 2;
  // can add more channels in the future
}

message NotificationChannelList {
  bool enabled = 1;
  // can add more options specific to in-app notifs in the future
}

message NotificationChannelPush {
  bool enabled = 1;
    // can add more options specific to push notifs in the future
}
```

packages/bsky/proto/bsky.proto
```diff
@@ -753,6 +753,49 @@ message GetBlocklistSubscriptionsResponse {
...
+  bool push = 1;
+}
+
+message NotificationPreferenceFull {
```
jcalabro: I'm thinking this should be called `NotificationPreference` and the other one that can't be filtered should be `RequiedNotificationPreference` or similar? I'm not in love with that name, but I think it's better than what we have. Any other suggestions?

packages/bsky/src/vault.ts
devinivy: I like having the named abstraction 👍  I think "vault" is pretty good, my only reservation is that it makes it sound like it may be designed with strong security properties e.g. encryption.

A few alternatives that jump to mind (none of them especially exciting): "private storage", "private state", "actor storage".

packages/bsky/src/vault.ts
```diff
@@ -0,0 +1,57 @@
...
+    })
```
devinivy: I might recommend this vault abstraction being where lexicon validation comes into play based on `namespace`, and working with the lexicon data model in the vault's public interface.

That would probably involve passing a `LexValue` as the `payload`, and using `lexToJson()` or `stringifyLex()` in here from the `@atproto/lexicon` package.  This would be consistent with typical handling of e.g. record values, xrpc body/input values, etc.

packages/bsky/src/vault.ts
devinivy: Would recommend our typical JSON or DAG-CBOR.  Upside of JSON is that it's easier to work with and debug, upside of DAG-CBOR is that it's more compact.  (I'd probably lean JSON, personally.)

packages/bsky/src/vault.ts
devinivy: You also may want to consider whether the payload explicitly should or shouldn't include `$type`.  In the public repository we do enforce `$type` on records, partly so that if you come across a free-floating record out of context you can determine what it represents and validate it.

lexicons/app/bsky/notification/defs.json
```diff
@@ -5,6 +5,94 @@
...
+        "push": { "type": "boolean", "default": true }
+      }
+    },
+    "channelsPush": {
+      "type": "object",
+      "required": ["push"],
+      "properties": {
+        "push": { "type": "boolean", "default": true }
+      }
+    },
```
devinivy: I don't think we need defaults on required fields since the defaults will not have a chance to be applied.  It looks like there may be some additional cases like this below.
```suggestion
      "required": ["list", "push"],
      "properties": {
        "list": { "type": "boolean" },
        "push": { "type": "boolean" }
      }
    },
    "channelsPush": {
      "type": "object",
      "required": ["push"],
      "properties": {
        "push": { "type": "boolean" }
      }
    },
```

lexicons/app/bsky/notification/putPreferencesV2.json
```diff
@@ -0,0 +1,83 @@
+{
+  "lexicon": 1,
+  "id": "app.bsky.notification.putPreferencesV2",
```
devinivy, line 3: I assume this replaces the functionality of `app.bsky.notification.putPreferences`.  Is the `priority` setting from that endpoint modeled here somewhere?

packages/bsky/src/api/app/bsky/notification/getPreferences.ts
```diff
@@ -0,0 +1,42 @@
...
+  try {
+    const res = await ctx.dataplane.getNotificationPreferences({
+      actorDid,
+    })
+    return ensurePreferences(res)
+  } catch (err) {
+    if (err instanceof ConnectError && err.code !== Code.NotFound) {
+      throw new UpstreamFailureError(
+        'cannot get current notification preferences',
+      )
```
devinivy: We don't have a super solid established practice of doing this, but it could be nice to preserve the error chain w/ `cause` particularly in unexpected cases like this:
```suggestion
      throw new UpstreamFailureError(
        'cannot get current notification preferences',
        undefined,
        { cause: err },
      )
```

packages/bsky/src/api/app/bsky/notification/putPreferencesV2.ts
devinivy: I threw a comment in the vault class about this—I wonder if lexicon validation should be functionality of the vault itself, based on the lexicon `namespace`.

packages/bsky/src/api/app/bsky/notification/putPreferencesV2.ts
```diff
@@ -0,0 +1,74 @@
...
+  input: HandlerInput,
+): Promise<{ preferences: Un$Typed<Preferences>; exists: boolean }> => {
+  let preferences: Preferences
+  let exists = false
+  try {
+    const res = await ctx.dataplane.getNotificationPreferences({
+      actorDid,
+    })
+    const currentPreferences = ensurePreferences(res)
+    preferences = { ...currentPreferences, ...input.body }
```
devinivy: `input.body` may have unknown fields—do we want those to make their way in?  Could imagine either way but worth having a "take" on.

packages/bsky/src/api/app/bsky/notification/putPreferencesV2.ts
```diff
@@ -0,0 +1,74 @@
...
+    exists = true
+  } catch (err) {
+    if (err instanceof ConnectError && err.code !== Code.NotFound) {
+      throw new UpstreamFailureError(
+        'cannot get current notification preferences',
+      )
```
devinivy:
```suggestion
        'cannot get current notification preferences',
        undefined,
        { cause: err },
      )
```

packages/bsky/src/api/app/bsky/notification/putPreferencesV2.ts
devinivy: I know having a single preferences object that we overwrite has created some pitfalls for the client team in the past.  I think part of the issue there was other clients clobbering preferences, which wont be an issue here.  Just want to make sure @mozzius was able to look at this through that lens.

packages/bsky/src/data-plane/bsync/index.ts
```diff
@@ -138,6 +139,63 @@ const createRoutes = (db: Database) => (router: ConnectRouter) =>
...
+        await db.db
+          .updateTable('private_data')
+          .where('actorDid', '=', actorDid)
+          .where('namespace', '=', namespace)
+          .where('key', '=', key)
+          .set({
+            payload: Buffer.from(payload).toString('utf8'),
+            updatedAt: new Date().toISOString(),
+          })
+          .execute()
```
devinivy: Can we have a single time for "now" within this method?  It is always a little awkward when a creation and update time are off by some milliseconds when the row has never been updated.
```suggestion
      const now = new Date()
      if (method === Method.CREATE) {
        await db.db
          .insertInto('private_data')
          .values({
            actorDid,
            namespace,
            key,
            payload: Buffer.from(payload).toString('utf8'),
            indexedAt: now.toISOString(),
            updatedAt: now.toISOString(),
          })
          .execute()
      } else if (method === Method.UPDATE) {
        await db.db
          .updateTable('private_data')
          .where('actorDid', '=', actorDid)
          .where('namespace', '=', namespace)
          .where('key', '=', key)
          .set({
            payload: Buffer.from(payload).toString('utf8'),
            updatedAt: now.toISOString(),
          })
          .execute()
```

packages/bsky/src/data-plane/bsync/index.ts
```diff
@@ -138,6 +139,63 @@ const createRoutes = (db: Database) => (router: ConnectRouter) =>
...
+          .where('namespace', '=', namespace)
+          .where('key', '=', key)
+          .execute()
+      } else {
+        throw new Error(`Unsupported method: ${method}`)
+      }
+
+      // @NOTE: This operation ID is opaque to the client, and is used as a cursor to scan.
+      // Since we don't implement scanning for dev-env, it can random, doesn't have to be sequential.
+      const operationId = crypto.randomStr(16, 'hex')
```
devinivy: Could use `TID.nextStr()` and make it sequential pretty easily though!

packages/bsky/src/data-plane/server/routes/private-data.ts
```diff
@@ -0,0 +1,25 @@
...
+      .where('key', '=', 'self')
+      .executeTakeFirst()
+    if (!res) {
+      throw new ConnectError(
+        'Notification preferences not found',
+        Code.NotFound,
+      )
+    }
+
+    return JSON.parse(res.payload) as GetNotificationPreferencesResponse
```
devinivy: Oh wow, does the lexicon representation of preferences match the protobuf representation that closely? 👀

devinivy (commented):
Just dropped a few thoughts, looking good!

lexicons/app/bsky/notification/putPreferencesV2.json
mozzius: It's entirely superseded by setting reply/mention/quote notifs to `filter: "follows"`. We will run a migration

packages/bsky/src/api/app/bsky/notification/putPreferencesV2.ts
mozzius: Should be good - these are required fields, so if you forget them you'll get an error. The problem with the generic prefs is that it's far too easy to set it to `[]` but that's not an issue here

packages/bsky/src/vault.ts
ghost: > my only reservation is that it makes it sound like it may be designed with strong security properties e.g. encryption.

I agree, I'll try something else.

packages/bsky/src/api/app/bsky/notification/putPreferencesV2.ts
ghost: > these are required fields, so if you forget them you'll get an error.

@mozzius that's not quite true here. The input type for `putPreferencesV2` has all properties optional.

packages/bsky/src/data-plane/server/routes/private-data.ts
ghost: Yeah, I wasn't sure about how we want to model that in the dataplane. For now I did them basically identical, but if @jcalabro has a preference on how to represent that on the dataplane we might want to change this.

jcalabro: I made some suggestions to change the protobuf shapes, but actually those changes could also be reflected in the lexicon? I'm not sure they have to be 1:1, and if they are great, but if not, I don't view that as the end of the world

packages/bsky/src/vault.ts
ghost: Named `StashClient` in 920f74381

packages/bsky/src/vault.ts
ghost: 6d7624f4d

packages/bsky/src/data-plane/bsync/index.ts
ghost: dfee5aaaf

packages/bsky/src/data-plane/bsync/index.ts
ghost: https://github.com/bluesky-social/atproto/commit/dfee5aaaf072e31348a4c3a9c8fb0a8d218c338d

lexicons/app/bsky/notification/defs.json
ghost: a488df2a7

packages/bsky/src/api/app/bsky/notification/getPreferences.ts
ghost: ff3ef1f1d

packages/bsky/proto/bsky.proto
ghost: One small issue I have with this is that not necessarily all cases will have all possible values of the enum.

Imagine the enum one day has 3 values, but some cases only accept 2 of those possible values. How do we represent that?

I might be overthinking this, though.

packages/bsky/src/api/app/bsky/notification/putPreferencesV2.ts
ghost: https://github.com/bluesky-social/atproto/commit/ff3ef1f1dde257e65cdeb1f9d295d6d430da6b4d

packages/bsky/src/data-plane/server/routes/private-data.ts
ghost: I think the main point was whether we wanted to model the dataplane and lex definitions identically. I think it shouldn't be a requirement, but if it happens to be, it is also not an issue.

packages/bsky/src/api/app/bsky/notification/putPreferencesV2.ts
ghost: I'd rather not allow unknown fields. Do we have a way to ensure only known properties based on the lexicon using our libraries and/or zod somehow?

packages/bsky/src/api/app/bsky/notification/putPreferencesV2.ts
ghost: Will handle it on the other comment

packages/bsky/proto/bsky.proto
ghost: Can you give me an example of what this would look like?

> It's also more composable because the chat = 1; preference on the response shape could embed the NotificationChannelPush exactly the same as NotificationChannels rather than having it be a bespoke type.

packages/bsky/proto/bsky.proto
ghost: I'm fine with calling this `NotificationPreference`. I don't like much calling the other `RequiedNotificationPreference` (the `Required` implies something here that I don't think is correct).

packages/bsky/proto/bsky.proto
ghost: Created the enum in 2801f2cea. Note that I needed 2 different enums.

packages/bsky/proto/bsky.proto
ghost: Addressed this based on our call 2801f2cea

packages/bsky/proto/bsky.proto
ghost: Renamed in 2801f2cea

packages/bsky/src/data-plane/server/routes/private-data.ts
ghost: They got a bit different in 2801f2cea and I'm doing an explicit mapping

packages/bsky/src/vault.ts
ghost: I've added the `$type` in ea86c2572.

packages/bsky/proto/bsky.proto
```diff
@@ -753,6 +753,66 @@ message GetBlocklistSubscriptionsResponse {
 // Notifications
 //
 
+message GetNotificationPreferencesRequest {
```
jcalabro, line 756: After reviewing again, these protos do look a lot better to me 👍

lexicons/app/bsky/notification/putPreferencesV2.json
ghost: Yup, resolving.

packages/bsky/src/vault.ts
ghost: Validated lexicon in 82d2848557fd97b8bd061b26a6048a94ad545976

packages/bsky/src/api/app/bsky/notification/putPreferencesV2.ts
ghost: We've discussed this and are allowing unknown fields for compatibility. The dataplane will store them, even if we might not return them.

mozzius: Ah ok. but if you leave out the options, it just won't change them, right? Whereas with generic prefs if you leave them out they get wiped

ghost: Correct. Resolving

packages/bsky/src/data-plane/bsync/index.ts
```diff
@@ -138,7 +143,110 @@ const createRoutes = (db: Database) => (router: ConnectRouter) =>
...
+      if (
+        method !== Method.CREATE &&
+        method !== Method.UPDATE &&
+        method !== Method.DELETE
+      ) {
+        throw new Error(`Unsupported method: ${method}`)
+      }
+
+      const now = new Date().toISOString()
+      if (namespace === 'app.bsky.notification.defs#preferences') {
```
ghost, line 157: The idea here is to allow a custom handling for cases that require it, while still having a generally low-lift approach to private data on the TS dataplane (having a unified flow and table for most cases).

This might change as we figure out new cases but for now it seems reasonable.

packages/bsky/src/api/app/bsky/notification/getPreferences.ts
```diff
@@ -0,0 +1,50 @@
...
+      { cause: err },
+    )
+  }
+
+  if (res.preferences.length !== 1) {
+    throw new InternalServerError(
+      `expected exactly one preferences entry, got ${res.preferences.length}`,
+      'NotificationPreferencesWrongResult',
+    )
+  }
```
devinivy: I am not sure that we want to expose to consumers which unexpected condition took place to cause the 500, as a fairly general practice.  If we make an assertion here then it should turn into a 500 and the error details should get logged appropriately.
```suggestion
  assert(
    res.preferences.length === 1,
    `expected exactly one preferences entry, got ${res.preferences.length}`,
  )
```

(To adopt this, would need to import `assert`)

packages/bsky/src/api/app/bsky/notification/putPreferencesV2.ts
```diff
@@ -0,0 +1,63 @@
...
+      { cause: err },
+    )
+  }
+
+  if (res.preferences.length !== 1) {
+    throw new InternalServerError(
+      `expected exactly one preferences entry, got ${res.preferences.length}`,
+      'NotificationPreferencesWrongResult',
+    )
+  }
```
devinivy: Could apply a similar assertion here.

packages/bsync/src/routes/put-operation.ts
```diff
@@ -82,6 +82,17 @@ const validateOp = (req: PutOperationRequest): Operation => {
     throw new ConnectError('operation method is invalid', Code.InvalidArgument)
   }
 
+  if (req.method === Method.CREATE || req.method === Method.UPDATE) {
+    try {
+      JSON.parse(new TextDecoder().decode(req.payload))
```
devinivy, line 87: I think elsewhere in the PR I saw `.toString('utf8')` used.  Would that work just fine?

packages/bsync/tests/operations.test.ts
```diff
@@ -16,6 +16,14 @@ describe('operations', () => {
   let bsync: BsyncService
   let client: BsyncClient
 
+  const validPayload0 = Buffer.from(
+    new TextEncoder().encode(JSON.stringify({ value: 0 })),
+  )
+  const validPayload1 = Buffer.from(
+    new TextEncoder().encode(JSON.stringify({ value: 1 })),
+  )
+  const invalidPayload = Buffer.from(new TextEncoder().encode('{invalid json}'))
```
devinivy: Would using e.g. `Buffer.from('{invalid json}', 'utf8')` do the trick?  (utf8 is the default so technically that second arg is not needed, but would be a good way to be explicit if you like.)

devinivy (approved):

packages/bsync/src/routes/put-operation.ts
ghost: Do you mean `req.payload.toString('utf8')`? The `toString` doesn't accept an argument, and it does not do what I want (it concatenates the array of numbers for the bytes).

packages/bsync/tests/operations.test.ts
ghost: Done 08ba67892

packages/bsky/src/api/app/bsky/notification/getPreferences.ts
ghost: 08ba67892

packages/bsky/src/api/app/bsky/notification/putPreferencesV2.ts
ghost: 08ba67892

packages/bsync/tests/operations.test.ts
ghost: Actually forgot to handle all cases, fixed in 63ec9ba56

packages/bsync/src/routes/put-operation.ts
devinivy: Ah, I had assumed that `req.payload` was a `Buffer`.  I see it's a `Uint8Array`.  You'll find we use the `uint8arrays` library pretty liberally for working with those, or on the server they are easy to wrap in a `Buffer`.

PR #3905, opened 2025-05-28
`app.bsky.unspecced.getPostThreadV2` filtering

ghost:


packages/bsky/src/views/index.ts
```diff
@@ -1149,12 +1149,16 @@ export class Views {
...
+      threadTagsBumpDown,
+      threadTagsHide,
     }: {
       above: number
       below: number
       branchingFactor: number
       prioritizeFollowedUsers: boolean
       sort: GetPostThreadV2QueryParams['sort']
+      threadTagsBumpDown: readonly string[]
+      threadTagsHide: readonly string[]
```
devinivy, line 1161: These get piped around quite a bit, but they are also static to the server.  I wonder if there's room to hoist it up into config for the `Views` class and avoid all the plumbing.

devinivy (approved):

PR #3906, opened 2025-05-29
Fixed codegen for arrays of strings with known values

mozzius:
**Stacked on #3901**

Fixes codegen for arrays of strings with known values by overzealously adding parentheses. Most of these will be stripped by prettier, but in the case of #3901 it fixes the incorrect type generation

Before:

```ts
export interface Preference {
  $type?: 'app.bsky.notification.defs#preference'
  channels: 'in-app' | 'push' | (string & {})[]
  filter: 'all' | 'follows' | (string & {})
}
```

After:

```ts
export interface Preference {
  $type?: 'app.bsky.notification.defs#preference'
  channels: ('in-app' | 'push' | (string & {}))[]
  filter: 'all' | 'follows' | (string & {})
}
```


devinivy (approved):
Looks good, mind just adding a changeset for @atproto/lex-cli and @atproto/api?

mozzius:
Rebased to main, so it's just the change to @atproto/lex-cli now

PR #3908, opened 2025-05-29
Subscribe to activity lexicons

mozzius:
Adds new lexicons

- `app.bsky.notification.subscribe`
- `app.bsky.notification.unsubscribe`
- `app.bsky.notification.listSubscriptions`
- Adds `!no-subscribe` label
- Extends `app.bsky.actor.defs#viewerState` with `subscription`

### DIFF
--- a/lexicons/app/bsky/actor/defs.json
+++ b/lexicons/app/bsky/actor/defs.json
@@ -162,6 +162,10 @@
         "knownFollowers": {
           "type": "ref",
           "ref": "#knownFollowers"
+        },
+        "subscription": {
+          "type": "ref",
+          "ref": "#subscription"
         }
       }
     },
@@ -227,6 +231,15 @@
           "type": "string",
           "description": "Timestamp when the verification was created.",
           "format": "datetime"
+        }
+      }
+    },
+    "subscription": {
+      "type": "object",
+      "properties": {
+        "activity": {
+          "type": "string",
+          "knownValues": ["posts_no_replies", "posts_with_replies"]
         }
       }
     },
--- a/lexicons/app/bsky/notification/subscribe.json
+++ b/lexicons/app/bsky/notification/subscribe.json
@@ -1,0 +1,24 @@
+{
+  "lexicon": 1,
+  "id": "app.bsky.notification.subscribe",
+  "defs": {
+    "main": {
+      "type": "procedure",
+      "description": "Subscribe to activity for an account. Requires auth.",
+      "input": {
+        "encoding": "application/json",
+        "schema": {
+          "type": "object",
+          "required": ["subject", "activity"],
+          "properties": {
+            "subject": { "type": "string", "format": "did" },
+            "activity": {
+              "type": "string",
+              "knownValues": ["posts_no_replies", "posts_with_replies"]
+            }
+          }
+        }
+      }
+    }
+  }
+}
--- a/packages/api/definitions/labels.json
+++ b/packages/api/definitions/labels.json
@@ -83,6 +83,15 @@
     }
   },
   {
+    "identifier": "!no-subscribe",
+    "configurable": false,
+    "defaultSetting": "warn",
+    "flags": ["no-override"],
+    "severity": "none",
+    "blurs": "none",
+    "behaviors": {}
+  },
+  {
     "identifier": "porn",
     "configurable": true,
     "defaultSetting": "hide",
--- a/lexicons/app/bsky/notification/listSubscriptions.json
+++ b/lexicons/app/bsky/notification/listSubscriptions.json
@@ -1,0 +1,47 @@
+{
+  "lexicon": 1,
+  "id": "app.bsky.notification.listSubscriptions",
+  "defs": {
+    "main": {
+      "type": "procedure",
+      "description": "List subscriptions to all accounts. Requires auth.",
+      "parameters": {
+        "type": "params",
+        "properties": {
+          "limit": {
+            "type": "integer",
+            "minimum": 1,
+            "maximum": 100,
+            "default": 50
+          },
+          "cursor": { "type": "string" }
+        }
+      },
+      "output": {
+        "encoding": "application/json",
+        "schema": {
+          "type": "object",
+          "required": ["subscriptions"],
+          "properties": {
+            "cursor": { "type": "string" },
+            "subscriptions": {
+              "type": "array",
+              "items": { "type": "ref", "ref": "#subscription" }
+            }
+          }
+        }
+      }
+    },
+    "subscription": {
+      "type": "object",
+      "required": ["subject"],
+      "properties": {
+        "subject": { "type": "string", "format": "did" },
+        "activity": {
+          "type": "string",
+          "knownValues": ["posts_no_replies", "posts_with_replies"]
+        }
+      }
+    }
+  }
+}
--- a/lexicons/app/bsky/notification/unsubscribe.json
+++ b/lexicons/app/bsky/notification/unsubscribe.json
@@ -1,0 +1,20 @@
+{
+  "lexicon": 1,
+  "id": "app.bsky.notification.unsubscribe",
+  "defs": {
+    "main": {
+      "type": "procedure",
+      "description": "Unsubscribe from activity for an account. Requires auth.",
+      "input": {
+        "encoding": "application/json",
+        "schema": {
+          "type": "object",
+          "required": ["subject"],
+          "properties": {
+            "subject": { "type": "string", "format": "did" }
+          }
+        }
+      }
+    }
+  }
+}
--- a/packages/api/src/client/index.ts
+++ b/packages/api/src/client/index.ts
@@ -128,8 +128,8 @@ import * as AppBskyFeedGetFeedGenerators from './types/app/bsky/feed/getFeedGene
 import * as AppBskyFeedGetFeedSkeleton from './types/app/bsky/feed/getFeedSkeleton.js'
 import * as AppBskyFeedGetLikes from './types/app/bsky/feed/getLikes.js'
 import * as AppBskyFeedGetListFeed from './types/app/bsky/feed/getListFeed.js'
-import * as AppBskyFeedGetPostThread from './types/app/bsky/feed/getPostThread.js'
 import * as AppBskyFeedGetPosts from './types/app/bsky/feed/getPosts.js'
+import * as AppBskyFeedGetPostThread from './types/app/bsky/feed/getPostThread.js'
 import * as AppBskyFeedGetQuotes from './types/app/bsky/feed/getQuotes.js'
 import * as AppBskyFeedGetRepostedBy from './types/app/bsky/feed/getRepostedBy.js'
 import * as AppBskyFeedGetSuggestedFeeds from './types/app/bsky/feed/getSuggestedFeeds.js'
@@ -176,8 +176,11 @@ import * as AppBskyLabelerService from './types/app/bsky/labeler/service.js'
 import * as AppBskyNotificationDefs from './types/app/bsky/notification/defs.js'
 import * as AppBskyNotificationGetUnreadCount from './types/app/bsky/notification/getUnreadCount.js'
 import * as AppBskyNotificationListNotifications from './types/app/bsky/notification/listNotifications.js'
+import * as AppBskyNotificationListSubscriptions from './types/app/bsky/notification/listSubscriptions.js'
 import * as AppBskyNotificationPutPreferences from './types/app/bsky/notification/putPreferences.js'
 import * as AppBskyNotificationRegisterPush from './types/app/bsky/notification/registerPush.js'
+import * as AppBskyNotificationSubscribe from './types/app/bsky/notification/subscribe.js'
+import * as AppBskyNotificationUnsubscribe from './types/app/bsky/notification/unsubscribe.js'
 import * as AppBskyNotificationUpdateSeen from './types/app/bsky/notification/updateSeen.js'
 import * as AppBskyRichtextFacet from './types/app/bsky/richtext/facet.js'
 import * as AppBskyUnspeccedDefs from './types/app/bsky/unspecced/defs.js'
@@ -391,8 +394,8 @@ export * as AppBskyFeedGetFeedGenerators from './types/app/bsky/feed/getFeedGene
 export * as AppBskyFeedGetFeedSkeleton from './types/app/bsky/feed/getFeedSkeleton.js'
 export * as AppBskyFeedGetLikes from './types/app/bsky/feed/getLikes.js'
 export * as AppBskyFeedGetListFeed from './types/app/bsky/feed/getListFeed.js'
-export * as AppBskyFeedGetPostThread from './types/app/bsky/feed/getPostThread.js'
 export * as AppBskyFeedGetPosts from './types/app/bsky/feed/getPosts.js'
+export * as AppBskyFeedGetPostThread from './types/app/bsky/feed/getPostThread.js'
 export * as AppBskyFeedGetQuotes from './types/app/bsky/feed/getQuotes.js'
 export * as AppBskyFeedGetRepostedBy from './types/app/bsky/feed/getRepostedBy.js'
 export * as AppBskyFeedGetSuggestedFeeds from './types/app/bsky/feed/getSuggestedFeeds.js'
@@ -439,8 +442,11 @@ export * as AppBskyLabelerService from './types/app/bsky/labeler/service.js'
 export * as AppBskyNotificationDefs from './types/app/bsky/notification/defs.js'
 export * as AppBskyNotificationGetUnreadCount from './types/app/bsky/notification/getUnreadCount.js'
 export * as AppBskyNotificationListNotifications from './types/app/bsky/notification/listNotifications.js'
+export * as AppBskyNotificationListSubscriptions from './types/app/bsky/notification/listSubscriptions.js'
 export * as AppBskyNotificationPutPreferences from './types/app/bsky/notification/putPreferences.js'
 export * as AppBskyNotificationRegisterPush from './types/app/bsky/notification/registerPush.js'
+export * as AppBskyNotificationSubscribe from './types/app/bsky/notification/subscribe.js'
+export * as AppBskyNotificationUnsubscribe from './types/app/bsky/notification/unsubscribe.js'
 export * as AppBskyNotificationUpdateSeen from './types/app/bsky/notification/updateSeen.js'
 export * as AppBskyRichtextFacet from './types/app/bsky/richtext/facet.js'
 export * as AppBskyUnspeccedDefs from './types/app/bsky/unspecced/defs.js'
@@ -2107,6 +2113,13 @@ export class AppBskyFeedNS {
       })
   }
 
+  getPosts(
+    params?: AppBskyFeedGetPosts.QueryParams,
+    opts?: AppBskyFeedGetPosts.CallOptions,
+  ): Promise<AppBskyFeedGetPosts.Response> {
+    return this._client.call('app.bsky.feed.getPosts', params, undefined, opts)
+  }
+
   getPostThread(
     params?: AppBskyFeedGetPostThread.QueryParams,
     opts?: AppBskyFeedGetPostThread.CallOptions,
@@ -2116,13 +2129,6 @@ export class AppBskyFeedNS {
       .catch((e) => {
         throw AppBskyFeedGetPostThread.toKnownErr(e)
       })
-  }
-
-  getPosts(
-    params?: AppBskyFeedGetPosts.QueryParams,
-    opts?: AppBskyFeedGetPosts.CallOptions,
-  ): Promise<AppBskyFeedGetPosts.Response> {
-    return this._client.call('app.bsky.feed.getPosts', params, undefined, opts)
   }
 
   getQuotes(
@@ -3367,6 +3373,18 @@ export class AppBskyNotificationNS {
     )
   }
 
+  listSubscriptions(
+    data?: AppBskyNotificationListSubscriptions.InputSchema,
+    opts?: AppBskyNotificationListSubscriptions.CallOptions,
+  ): Promise<AppBskyNotificationListSubscriptions.Response> {
+    return this._client.call(
+      'app.bsky.notification.listSubscriptions',
+      opts?.qp,
+      data,
+      opts,
+    )
+  }
+
   putPreferences(
     data?: AppBskyNotificationPutPreferences.InputSchema,
     opts?: AppBskyNotificationPutPreferences.CallOptions,
@@ -3385,6 +3403,30 @@ export class AppBskyNotificationNS {
   ): Promise<AppBskyNotificationRegisterPush.Response> {
     return this._client.call(
       'app.bsky.notification.registerPush',
+      opts?.qp,
+      data,
+      opts,
+    )
+  }
+
+  subscribe(
+    data?: AppBskyNotificationSubscribe.InputSchema,
+    opts?: AppBskyNotificationSubscribe.CallOptions,
+  ): Promise<AppBskyNotificationSubscribe.Response> {
+    return this._client.call(
+      'app.bsky.notification.subscribe',
+      opts?.qp,
+      data,
+      opts,
+    )
+  }
+
+  unsubscribe(
+    data?: AppBskyNotificationUnsubscribe.InputSchema,
+    opts?: AppBskyNotificationUnsubscribe.CallOptions,
+  ): Promise<AppBskyNotificationUnsubscribe.Response> {
+    return this._client.call(
+      'app.bsky.notification.unsubscribe',
       opts?.qp,
       data,
       opts,
--- a/packages/api/src/client/lexicons.ts
+++ b/packages/api/src/client/lexicons.ts
@@ -4779,6 +4779,10 @@ export const schemaDict = {
             type: 'ref',
             ref: 'lex:app.bsky.actor.defs#knownFollowers',
           },
+          subscription: {
+            type: 'ref',
+            ref: 'lex:app.bsky.actor.defs#subscription',
+          },
         },
       },
       knownFollowers: {
@@ -4851,6 +4855,15 @@ export const schemaDict = {
             type: 'string',
             description: 'Timestamp when the verification was created.',
             format: 'datetime',
+          },
+        },
+      },
+      subscription: {
+        type: 'object',
+        properties: {
+          activity: {
+            type: 'string',
+            knownValues: ['posts_no_replies', 'posts_with_replies'],
           },
         },
       },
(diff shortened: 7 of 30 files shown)


lexicons/app/bsky/notification/subscribe.json
```diff
@@ -0,0 +1,24 @@
...
+        "schema": {
+          "type": "object",
+          "required": ["subject", "activity"],
+          "properties": {
+            "subject": { "type": "string", "format": "did" },
+            "activity": {
+              "type": "string",
+              "knownValues": ["posts_no_replies", "posts_with_replies"]
+            }
+          }
```
mozzius, line 19: Note: I set `activity` to be required here. Still not 100% sure what the ideal shape should be, given it should be extensible enough to support Live Now in future

lexicons/app/bsky/notification/subscribe.json
```diff
@@ -0,0 +1,24 @@
...
+        "encoding": "application/json",
+        "schema": {
+          "type": "object",
+          "required": ["subject", "activity"],
+          "properties": {
+            "subject": { "type": "string", "format": "did" },
+            "activity": {
+              "type": "string",
+              "knownValues": ["posts_no_replies", "posts_with_replies", "none"]
+            }
```
mozzius, line 18: setting `activity` to `none` should delete the subscription. This is future-proofing - in the future, we'll probably want to subscribe to `none` + "live now" notifications.

In `app.bsky.actor.defs#viewerState`, we should delete the object rather than return `none`, but it's in the lexicon so that if we can support `none` + "live now" in the future

lexicons/app/bsky/actor/defs.json
```diff
@@ -230,6 +234,16 @@
         }
       }
     },
+    "subscription": {
+      "type": "object",
+      "required": ["activity"],
+      "properties": {
+        "activity": {
+          "type": "string",
+          "knownValues": ["posts_no_replies", "posts_with_replies", "none"]
```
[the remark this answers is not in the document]
jcalabro: Nvm, this was answered by https://github.com/bluesky-social/atproto/pull/3908/files#r2114029683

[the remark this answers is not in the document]
mozzius: in the future, we might serve something like:

```json
{
  "activity": "none",
  "goesLive": true,
}
```

and thus we should support the "none" case now so we don't have to add it in future, which would be awkward from a lexicon evolution standpoint

packages/api/definitions/labels.json
```diff
@@ -82,6 +82,15 @@
...
   },
+  {
+    "identifier": "!no-subscribe",
+    "configurable": false,
+    "defaultSetting": "warn",
+    "flags": ["no-override"],
+    "severity": "none",
+    "blurs": "none",
+    "behaviors": {}
+  },
```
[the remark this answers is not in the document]
mozzius: No longer using labels

packages/bsky/src/data-plane/server/routes/notifs.ts
```diff
@@ -151,4 +167,82 @@ export default (db: Database): Partial<ServiceImpl<typeof Service>> => ({
       .onConflict((oc) => oc.doNothing())
       .executeTakeFirst()
   },
+
+  async getNotificationPreferences(req) {
```
ghost, line 172: Moved here from the deleted `private-data.ts` file, I'd rather keep the endpoints grouped by feature than by which tables they access.

lexicons/app/bsky/notification/putActivitySubscription.json
```diff
@@ -0,0 +1,35 @@
...
+        "encoding": "application/json",
+        "schema": {
+          "type": "object",
+          "required": ["subject", "activitySubscription"],
+          "properties": {
+            "key": { "type": "string", "format": "record-key" },
+            "subject": { "type": "string", "format": "did" },
+            "activitySubscription": {
+              "type": "ref",
+              "ref": "app.bsky.notification.defs#activitySubscription"
```
ghost: @mozzius Note that this means we need to pass all properties in the input. Do you mind this, or is it better to make each property optional? That probably makes the implementation a bit more complicated.

packages/bsky/src/data-plane/server/db/migrations/20250611T140649895Z-add-activity-subscription.ts
```diff
@@ -0,0 +1,29 @@
+import { Kysely, sql } from 'kysely'
+
+export async function up(db: Kysely<unknown>): Promise<void> {
+  await db.schema
+    .createTable('activity_subscription')
+    .addColumn('creator', 'varchar', (col) => col.notNull())
+    .addColumn('subjectDid', 'varchar', (col) => col.notNull())
```
ghost, line 7: Here's a case where I couldn't comfortably use the `private_data` generic table, because I needed to index by `subjectDid`, which in `private_data` would be burried deep into a JSON column.

packages/bsky/proto/bsky.proto
```diff
@@ -473,6 +473,7 @@ message ActorInfo {
   // Tags being applied to the profile record
   repeated string profile_tags = 16;
   Record status_record = 17;
+  string allow_activity_subscriptions_from = 18;
```
jcalabro, line 484: How about `allowed_activity_subscriptions` or `activity_subscriptions_preference`?

packages/bsky/proto/bsky.proto
```diff
@@ -503,6 +504,11 @@ message GetRelationshipsRequest {
   repeated string target_dids = 2;
 }
 
+message ActivitySubscription {
+  bool post = 1;
```
jcalabro: Wondering if we want these to be their own objects so there's some room to breathe for changes we may want to make in the future. Something like this?

```proto
message ActivitySubscription {
  PostActivitySubscription posts = 1;
  ReplyActivitySubscription replies = 2;
}

message PostActivitySubscription {
  bool enabled = 1;
  // can add more fields later
}

message ReplyActivitySubscription {
  bool enabled = 1;
  // can add more fields later
}
```

lexicons/app/bsky/notification/putActivitySubscription.json
mozzius: I think that's fine frontend implementation wise

packages/bsky/proto/bsky.proto
jcalabro: Alternative could be this, implying that if the optional messages are non-null, then the user wants that type of notification.

```proto
message ActivitySubscription {
  optional PostActivitySubscription posts = 1;
  optional ReplyActivitySubscription replies = 2;
}

message PostActivitySubscription {
  // can add more fields later
}

message ReplyActivitySubscription {
  // can add more fields later
}
```

ghost: I've changed it to match the second comment, in 76ddcda3e

packages/bsky/proto/bsky.proto
ghost: We discussed this, I'm keeping the name for consistency with the existing `allow_incoming_chats_from`, but changed it to an enum in 76ddcda3e

packages/bsky/src/stash.ts
```diff
@@ -1,8 +1,26 @@
...
 
+export const NamespaceAppBskyNotificationDefsPreferences: Exclude<
+  Preferences['$type'],
+  undefined
+> = 'app.bsky.notification.defs#preferences'
+
+export const NamespaceAppBskyNotificationDefsSubjectActivitySubscription: Exclude<
+  SubjectActivitySubscription['$type'],
+  undefined
+> = 'app.bsky.notification.defs#subjectActivitySubscription'
```
ghost: @jcalabro Now I have here a more explicit set of constants with all the stash (private data in bsync) namespaces I'm using.

lexicons/app/bsky/notification/defs.json
```diff
@@ -63,6 +63,26 @@
...
+    "activitySubscription": {
+      "type": "object",
+      "required": ["post", "reply"],
+      "properties": {
+        "post": { "type": "boolean" },
+        "reply": { "type": "boolean" }
+      }
+    },
+    "subjectActivitySubscription": {
+      "description": "Object used to store activity subscription data in stash (bsync).",
```
devinivy: Bsync is an implementation detail of our appview, so we probably don't need to mention it in lexicon descriptions.

lexicons/app/bsky/notification/putActivitySubscription.json
```diff
@@ -0,0 +1,35 @@
...
+      "output": {
+        "encoding": "application/json",
+        "schema": {
+          "type": "object",
+          "required": ["key"],
+          "properties": {
+            "key": { "type": "string", "format": "record-key" }
+          }
+        }
+      }
```
devinivy: Not an issue, but re: API design I am personally a fan of when APIs hand me back the entity I created/updated.

packages/bsky/src/data-plane/server/db/migrations/20250611T140649895Z-add-activity-subscription.ts
```diff
@@ -0,0 +1,29 @@
...
+        .stored()
+        .notNull(),
+    )
+    .addColumn('post', 'boolean', (col) => col.notNull())
+    .addColumn('reply', 'boolean', (col) => col.notNull())
+    .addPrimaryKeyConstraint('activity_subscription_pkey', [
+      'creator',
+      'subjectDid',
+    ])
+    .addUniqueConstraint('activity_subscription_unique_key', ['key'])
```
devinivy: Under the repo's semantics the key only really needs to be unique in the context of the DID + namespace.  Is that the same here, or different?

packages/bsky/src/data-plane/server/routes/activity-subscription.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export default (db: Database): Partial<ServiceImpl<typeof Service>> => ({
+  async getActivitySubscriptions(req) {
+    const { actorDid, cursor, limit } = req
+
+    let builder = db.db
+      .selectFrom('activity_subscription')
+      .select('subjectDid')
+      .where('creator', '=', actorDid)
+      .orderBy('indexedAt', 'desc')
```
devinivy: Perhaps it's not necessary, but I didn't see an index supporting this pagination scheme.

packages/bsky/src/data-plane/server/routes/activity-subscription.ts
```diff
@@ -0,0 +1,28 @@
...
+
+    const { ref } = db.db.dynamic
+    const key = new IsoSortAtKey(ref('activity_subscription.sortAt'))
```
devinivy: Should these both perhaps use `sortAt`?

packages/bsky/src/data-plane/server/routes/notifs.ts
```diff
@@ -151,4 +167,82 @@ export default (db: Database): Partial<ServiceImpl<typeof Service>> => ({
...
+      .where('key', '=', 'self')
+      .execute()
+
+    const byDid = keyBy(res, 'actorDid')
+    const preferences = dids.map((did) => {
+      const row = byDid.get(did)
+      if (!row) {
+        return {}
+      }
+      const p: Preferences = JSON.parse(row.payload)
```
devinivy: Probably doesn't make a difference in this case, but might recommend using `jsonStringToLex()` for consistency in handling lexicon data.

packages/bsky/src/api/app/bsky/notification/listActivitySubscriptions.ts
```diff
@@ -0,0 +1,123 @@
...
+    handler: async ({ params, auth, req }) => {
+      const viewer = auth.credentials.iss
+      const labelers = ctx.reqLabelers(req)
+      const hydrateCtx = await ctx.hydrator.createContext({
+        labelers,
+        viewer,
+      })
+
+      const result = await listActivitySubscriptions(
+        { ...params, hydrateCtx: hydrateCtx.copy({ viewer }) },
```
devinivy, line 34: At a glance I don't think the `copy()` is needed here.
```suggestion
        { ...params, hydrateCtx },
```

devinivy: Oh maybe it's a typing thing!

packages/bsky/src/api/app/bsky/notification/listActivitySubscriptions.ts
```diff
@@ -0,0 +1,123 @@
...
+  return followState
+}
+
+const noBlocks = (input: RulesFnInput<Context, Params, SkeletonState>) => {
+  const { skeleton, params, hydration, ctx } = input
+  const viewer = params.hydrateCtx.viewer
+  skeleton.dids = skeleton.dids.filter((followUri) => {
+    const followerDid = didFromUri(followUri)
+    return (
+      !hydration.followBlocks?.get(followUri) &&
```
devinivy: I believe `followBlocks` are used for the case when you're just looking at follow records of other users, as in `getFollows` or `getFollowers` endpoints.

packages/bsky/src/api/app/bsky/notification/listActivitySubscriptions.ts
```diff
@@ -0,0 +1,123 @@
...
+      (!viewer || !ctx.views.viewerBlockExists(followerDid, hydration))
```
devinivy: Two questions here:
 - is `viewer` ever missing?
 - if I subscribe to someone but block them, are we sure they should not be returned here?  E.g. I might want to be able to unsubscribe from them.  It's a product question, just want to make sure we gave it a look.

packages/bsky/src/api/app/bsky/notification/listActivitySubscriptions.ts
```diff
@@ -0,0 +1,123 @@
...
+const presentation = (
+  input: PresentationFnInput<Context, Params, SkeletonState>,
+) => {
+  const { ctx, hydration, skeleton, params } = input
+  const { dids, cursor } = skeleton
+  const isNoHosted = (did: string) => ctx.views.actorIsNoHosted(did, hydration)
+
+  const subscriptions = mapDefined(dids, (followUri) => {
+    const followerDid = didFromUri(followUri)
+    if (!params.hydrateCtx.includeTakedowns && isNoHosted(followerDid)) {
```
devinivy: As far as I can tell `includeTakedowns` isn't in play on this endpoint, i.e. wasn't set on the hydration context.

packages/bsky/src/data-plane/bsync/index.ts
```diff
@@ -210,6 +221,61 @@ const handleNotificationPreferencesOperation = async (
...
 }
 
+const handleSubjectActivitySubscriptionOperation = async (
+  db: Database,
+  req: PutOperationRequest,
+  now: string,
+) => {
+  const { actorDid, key, method, payload } = req
+
+  const parsed: SubjectActivitySubscription = JSON.parse(
```
devinivy: May consider using `jsonStringToLex()` here.

packages/bsky/src/data-plane/bsync/index.ts
```diff
@@ -210,6 +221,61 @@ const handleNotificationPreferencesOperation = async (
...
+  if (method === Method.UPDATE && enabled) {
+    return db.db
+      .updateTable('activity_subscription')
+      .where('creator', '=', actorDid)
+      .where('subjectDid', '=', parsed.subject)
+      .where('key', '=', key)
+      .set({
+        post,
+        reply,
+      })
```
devinivy: Perhaps update the indexed time?

packages/bsky/src/data-plane/server/routes/profile.ts
```diff
@@ -97,6 +110,28 @@ export default (db: Database): Partial<ServiceImpl<typeof Service>> => ({
...
 
+      const activitySubscription = () => {
+        const record =
+          parseRecordBytes<AppBskyNotificationActivitySubscriptionDeclaration.Record>(
+            activitySubscriptionDeclarations.records[i].record,
+          )
+
+        if (typeof record?.allowSubscriptions !== 'string') {
+          return ActivitySubscriptionsFrom.UNSPECIFIED
+        }
```
devinivy: Just to clarify: the default is applied in the appview and not the dataplane.

packages/bsky/src/stash.ts
devinivy: These string constants can also be pulled out of the `ids` export from `./lexicon/lexicons`.

packages/bsky/src/data-plane/bsync/index.ts
```diff
@@ -210,6 +221,61 @@ const handleNotificationPreferencesOperation = async (
   return handleGenericOperation(db, req, now)
 }
 
+const handleSubjectActivitySubscriptionOperation = async (
```
devinivy: I found the control flow in here a little difficult to follow, i.e. interactions between `method`, `enabled`, early return vs. falling through.

packages/bsky/src/data-plane/bsync/index.ts
```diff
@@ -210,6 +221,61 @@ const handleNotificationPreferencesOperation = async (
...
+    }
+    return
+  }
+
+  if (method === Method.UPDATE && enabled) {
+    return db.db
+      .updateTable('activity_subscription')
+      .where('creator', '=', actorDid)
+      .where('subjectDid', '=', parsed.subject)
+      .where('key', '=', key)
```
devinivy: `(creator, subjectDid)` and `key` are each unique.  Is there a reason both are needed?  What happens if an update for a given `key` changes the `creator` or `subjectDid`?

packages/bsky/src/data-plane/bsync/index.ts
```diff
@@ -210,6 +221,61 @@ const handleNotificationPreferencesOperation = async (
...
+          reply,
+        })
+        .execute()
+    }
+    return
+  }
+
+  if (method === Method.UPDATE && enabled) {
+    return db.db
+      .updateTable('activity_subscription')
```
devinivy: Should this throw if it doesn't impact any rows?

devinivy (commented):
Done with my review pass 👍 coming along pretty nicely.

lexicons/app/bsky/notification/defs.json
ghost: Do you mean only removing "(bsync)" or removing the part about "stash" as well (basically removing the whole description)?

I find it useful to keep some of this information in the description, since this type is not used anywhere in lexicons. I thought it might be confusing to someone looking at the lexicons and not seeing it being used anywhere.

packages/bsky/src/stash.ts
ghost: But only `app.bsky.notification.defs` is there, not `app.bsky.notification.defs#subjectActivitySubscription`.

Do you mean I should use the NSID and concatenate?

lexicons/app/bsky/notification/putActivitySubscription.json
ghost: I see your point. My main question would be: what is the entity created here? I'm not sure what should be considered its canonical representation.

packages/bsky/src/api/app/bsky/notification/listActivitySubscriptions.ts
ghost: Yeah, it is a typing thing.

packages/bsky/src/data-plane/bsync/index.ts
ghost: I don't know, should I?

packages/bsky/src/data-plane/server/db/migrations/20250611T140649895Z-add-activity-subscription.ts
ghost: You are right, it doesn't need to be unique. Here only the combination of `['creator', 'subjectDid']` needs to be unique. Removed it in 1afeb14dd.

packages/bsky/src/data-plane/server/routes/activity-subscription.ts
ghost: I first implemented it without pagination (with the orderBy), and then added pagination but forgot to remove the `orderBy`. Removed it, and added the index for `['creator', 'sortAt']` in 1afeb14dd.

packages/bsky/src/data-plane/server/routes/activity-subscription.ts
ghost: I removed the `orderBy`, as explained in https://github.com/bluesky-social/atproto/pull/3908#discussion_r2142956475

packages/bsky/src/data-plane/server/routes/notifs.ts
ghost: Fixed in 1afeb14dd

packages/bsky/src/data-plane/bsync/index.ts
ghost: Fixed in 1afeb14dd

packages/bsky/src/data-plane/bsync/index.ts
devinivy: The `indexedAt` indicates the most recent time that there was an change to the content causing it to be indexed.  I believe we are indexing it here, so I would expect it to be updated.

packages/bsky/src/stash.ts
devinivy: Ah interesting—you could do that, may be a useful way to help keep things in sync.  But I actually had an assumption that the hash references like `#subjectActivitySubscription` were contained in there too.

lexicons/app/bsky/notification/defs.json
devinivy: I meant removing the detail about bsync.

packages/bsky/src/data-plane/bsync/index.ts
ghost: Ah, I wasn't sure if it should be updated on updates. 79bca52ef

packages/bsky/src/api/app/bsky/notification/listActivitySubscriptions.ts
ghost: Right, I forgot to adapt this part, my bad. 50d283ace

packages/bsky/src/api/app/bsky/notification/listActivitySubscriptions.ts
ghost: - No. Fixed in 50d283ace.
- I'm checking with Product.

What about noHosted, should that apply here?

packages/bsky/src/api/app/bsky/notification/listActivitySubscriptions.ts
ghost: Removed it in 50d283ace.

packages/bsky/src/stash.ts
ghost: In this case, I prefer not concatenating, so searching for the full string `app.bsky.notification.defs#subjectActivitySubscription` finds all ocurrences.

lexicons/app/bsky/notification/defs.json
ghost: 3018c4e42

packages/bsky/src/data-plane/server/routes/profile.ts
ghost: I'll check how chat declaration works

packages/bsky/src/data-plane/bsync/index.ts
ghost: We discussed this, `key` is no longer unique.

lexicons/app/bsky/notification/putActivitySubscription.json
ghost: It returns an `activitySubscription` in ca70ab997

packages/bsky/src/data-plane/server/routes/profile.ts
ghost: After discussing it with @jcalabro , we decided the dataplane will enforce the default.

Done in 498eca6ec.

That affected a lot of snapshots though, I just didn't like that part. Including the PDS (424433fad) and Ozone (ea0421799).

packages/bsky/src/data-plane/bsync/index.ts
ghost: This was changed quite a bit, resolving.

packages/bsky/src/data-plane/bsync/index.ts
ghost: We discussed, and this should not throw. Ideally we'll rework it to work async and never throw (unless input validation fails).

packages/bsky/src/api/app/bsky/notification/putActivitySubscription.ts
```diff
@@ -10,39 +11,60 @@ export default function (server: Server, ctx: AppContext) {
...
-      const key = existingKey ?? TID.nextStr()
+      const existingKey = await getExistingKey(ctx, actorDid, subject)
+      const enabled = isActivitySubscriptionEnabled(activitySubscription)
+
       const stashInput = {
         actorDid,
         namespace: NamespaceAppBskyNotificationDefsSubjectActivitySubscription,
-        payload,
-        key,
+        payload: <SubjectActivitySubscription>{
```
devinivy: This is a type cast similar to using the `as` keyword.  It's a little uncommon so just making sure that was intended.

packages/bsky/src/api/app/bsky/notification/putActivitySubscription.ts
```diff
@@ -10,39 +11,60 @@ export default function (server: Server, ctx: AppContext) {
...
+          await ctx.stashClient.delete(stashInput)
+        }
       } else {
-        await ctx.stashClient.create(stashInput)
+        if (enabled) {
+          await ctx.stashClient.create(stashInput)
+        } else {
+          throw new InvalidRequestError(
+            'Cannot create a disabled activity subscription',
+          )
```
devinivy: I would lean towards treating this as idempotent, especially since the whole thing works asynchronously.

packages/bsky/src/data-plane/bsync/index.ts
```diff
@@ -228,51 +228,45 @@ const handleSubjectActivitySubscriptionOperation = async (
...
         post,
         reply,
       })
       .execute()
   }
 
-  // A DELETE, or an UPDATE with subscription disabled.
   return db.db
-    .deleteFrom('activity_subscription')
+    .updateTable('activity_subscription')
```
devinivy, line 262: Uniqueness constraint on (creator, subject) might fail.  Let's at least ensure the write to the stash doesn't fail for the caller.

packages/bsky/src/data-plane/bsync/index.ts
```diff
@@ -228,51 +228,45 @@ const handleSubjectActivitySubscriptionOperation = async (
...
-
-  if (method === Method.UPDATE && enabled) {
     return db.db
-      .updateTable('activity_subscription')
-      .where('creator', '=', actorDid)
-      .where('subjectDid', '=', parsed.subject)
-      .where('key', '=', key)
-      .set({
+      .insertInto('activity_subscription')
+      .values({
```
devinivy: Uniqueness constraint on (creator, subject) might fail. Let's at least ensure the write to the stash doesn't fail for the caller.

packages/bsky/src/api/app/bsky/notification/putActivitySubscription.ts
ghost: Yeah, I don't remember my reasoning for adding this. Removed it in bf112f2d6

.changeset/flat-countries-shake.md
```diff
@@ -0,0 +1,8 @@
+---
+"@atproto/ozone": patch
+"@atproto/bsky": patch
+"@atproto/api": patch
+"@atproto/pds": patch
+---
+
+Add activity subscription lexicons
```
ghost: No need for this anymore if we merge https://github.com/bluesky-social/atproto/pull/3990

devinivy (approved):

packages/bsky/proto/bsky.proto
```diff
@@ -453,13 +453,6 @@ message GetActorsRequest {
   repeated string skip_cache_for_dids = 2;
 }
 
-enum ActivitySubscriptionsFrom {
```
jcalabro: Why remove this? Enums > stringly-typed. It makes it more clear what the valid values are.

ghost: This came from a private convo with @ericvolp12 . They felt it is more flexible this way.
We can discuss this more if y'all want, especially including some lexicon expert (since this comes from a lexicon value that itself can be extended).

ghost: We've discussed this, we're sticking with strings.

PR #3916, opened 2025-06-02
Show account picker instead of welcome screen when user already signed-in

matthieusieben:
Whenever at least one account is signed in on the device, the welcome screen is replaced with the account picker view.

Capture d’écran 2025-06-02 à 16 19 28

### DIFF
--- a/packages/oauth/oauth-provider-ui/src/views/authorize/authorize-view.tsx
+++ b/packages/oauth/oauth-provider-ui/src/views/authorize/authorize-view.tsx
@@ -176,7 +176,7 @@ export function AuthorizeView({
             ? undefined
             : () => {
                 selectSub(null)
-                setView(sessions.length ? View.SignIn : View.Welcome)
+                showHome()
               }
         }
       />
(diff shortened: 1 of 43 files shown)


packages/oauth/oauth-provider-ui/src/views/authorize/authorize-view.tsx
```diff
@@ -160,7 +170,7 @@ export function AuthorizeView({
         account={session.account}
         scopeDetails={authorizeData.scopeDetails}
         onAccept={() => doAccept(session.account.sub)}
-        onReject={() => doReject()}
+        onReject={doReject}
```
devinivy, line 173: A few lines below it reads:
```ts
setView(sessions.length ? View.SignIn : View.Welcome)
```

Just double-checking that we don't want this to be `showHome()`.

matthieusieben: Good catch !

Issue #3723, opened 2025-04-07
@atproto/oauth-client sends PAR as JSON


tom-sherman:
**Describe the bug**

@atproto/oauth-client sends pushed authorization request as JSON when it should send it as form encoded.

**To Reproduce**

1. Send a PAR using the oauth client to an atproto auth server

**Expected behavior**

Body is sent form encoded

**Details**

- Operating system:
- Node version:

**Additional context**

From RFC9126: "2. Pushed Authorization Request Endpoint  [...]  2.1. Request [...] The client constructs the message body of an HTTP POST request with parameters formatted with x-www-form-urlencoded"

matthieusieben:
Actually, it's not only PAR, but all of the requests that should be using form encoding.

PR #3919, opened 2025-06-02
Use Form encoded body instead of JSON for OAuth requests

matthieusieben:
As reported by @tom-sherman our OAuth requests are not using the proper encoding (JSON instead of Form encoded). This fixes that.

Fixes: #3723

### DIFF
--- a/packages/oauth/oauth-client/src/oauth-server-agent.ts
+++ b/packages/oauth/oauth-client/src/oauth-server-agent.ts
@@ -207,10 +207,17 @@ export class OAuthServerAgent {
 
     const auth = await this.buildClientAuth(endpoint)
 
+    // https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1-13#section-3.2.2
+    // https://datatracker.ietf.org/doc/html/rfc7009#section-2.1
+    // https://datatracker.ietf.org/doc/html/rfc7662#section-2.1
+    // https://datatracker.ietf.org/doc/html/rfc9126#section-2
     const { response, json } = await this.dpopFetch(url, {
       method: 'POST',
-      headers: { ...auth.headers, 'Content-Type': 'application/json' },
-      body: JSON.stringify({ ...payload, ...auth.payload }),
+      headers: {
+        ...auth.headers,
+        'Content-Type': 'application/x-www-form-urlencoded',
+      },
+      body: wwwFormUrlEncode({ ...payload, ...auth.payload }),
     }).then(fetchJsonProcessor())
 
     if (response.ok) {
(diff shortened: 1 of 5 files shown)


packages/oauth/oauth-client/src/oauth-server-agent.ts
```diff
@@ -294,3 +301,37 @@ export class OAuthServerAgent {
...
+    case 'string':
+      return entry as [string, string]
+    case 'number':
+    case 'boolean':
+      return [name, String(value)]
+    default: {
+      const enc = JSON.stringify(value)
+      if (enc === undefined) {
+        throw new Error(`Unsupported value type for ${name}: ${String(value)}`)
+      }
```
devinivy, line 333: I don't think `enc` is allowed to be `undefined`, given that it's output from `JSON.stringify()`.  It is possible you were intending to use a "safe stringify" or something like that?
```suggestion
      const enc = JSON.stringify(value)
```

packages/oauth/oauth-client/src/oauth-server-agent.ts
```diff
@@ -294,3 +301,37 @@ export class OAuthServerAgent {
...
+  return entry[1] !== undefined
+}
+
+function stringifyEntryValue(entry: [string, unknown]): [string, string] {
+  const name = entry[0]
+  const value = entry[1]
+
+  switch (typeof value) {
+    case 'string':
+      return entry as [string, string]
```
devinivy: Is the purpose here not to allocate the array?  It is a small thing, but this seems more straightforward to me:
```suggestion
      return [name, value]
```

devinivy (approved):

packages/oauth/oauth-client/src/oauth-server-agent.ts
matthieusieben: `JSON.stringify` is not properly typed in TS's lib and can actually return `undefined`:

```ts
JSON.stringify(undefined) // undefined
JSON.stringify(Symbol.iterator) // undefined
JSON.stringify(() => {}) // undefined
```

packages/oauth/oauth-client/src/oauth-server-agent.ts
matthieusieben: That was the purpose indeed.

packages/oauth/oauth-client/src/oauth-server-agent.ts
matthieusieben: See https://github.com/microsoft/TypeScript/issues/18879#issuecomment-1399758565

PR #3927, opened 2025-06-05
❇️ Ozone events timeline

foysalit:
Oftentimes, for relatively problematic accounts, the mod event history gets quite noisy while not painting a full picture for moderators. This PR exposes the all relevant event history (not just moderation events) in a timeline view grouped by day to help moderators quickly grasp if an account is manifesting or has manifested problematic behavior.


lexicons/tools/ozone/moderation/getAccountTimeline.json
```diff
@@ -0,0 +1,73 @@
...
+    },
+    "accountTimelineSummary": {
+      "type": "object",
+      "required": ["eventSubjectType", "eventType", "count"],
+      "properties": {
+        "eventSubjectType": {
+          "type": "string",
+          "knownValues": ["account", "record", "chat"]
+        },
+        "eventType": {
```
foysalit, line 64: keeping this super generic since we want to be able to blend PLC audit log, mod events and account history from entryway at the moment but also leave room for additional event types in the future.
instead of keeping this flat, we could also make this an object where primary types could be `modEvent`, `plcLog`, `accountHistory` and then have multiple subtypes for each type of event types?

packages/ozone/src/mod-service/index.ts
```diff
@@ -1451,6 +1451,25 @@ export class ModerationService {
...
+        'subjectUri',
+        'action',
+        sql<number>`count(*)`.as('count'),
+      ])
+      .groupBy(['day', 'subjectDid', 'subjectUri', 'action'])
+      .orderBy('day', 'desc')
+      .execute()
+    return result
+  }
 }
```
devinivy: Since `subjectDid` is fixed, does it need to be part of the select, group-by, etc?  Perhaps it's there for convenience of the caller.

Also just double-checking that we're not too worried about usage of this query cause any issues.  I am imagining some accounts may have many moderation events.  If that is a problem, one option would be to paginate it e.g. by week.

lexicons/tools/ozone/moderation/getAccountTimeline.json
devinivy: Yeah, I think it's a bit difficult a. not having documentation here about which timeline event types may come down, and b. mixing different namespaces such as `'plc_operation'` vs. `'tools.ozone.moderation.defs#modEventTakedown'`.

I'd be pretty satisfied to keep it flat, which is nice and simple, and just continue to use tokens for all the types.  E.g. `'tools.ozone.moderation.defs#timelineEventPlcOperation'` and `'tools.ozone.moderation.defs#modEventTakedown'`.

lexicons/tools/ozone/moderation/getAccountTimeline.json
```diff
@@ -0,0 +1,73 @@
...
+          }
+        }
+      },
+      "errors": [
+        {
+          "name": "RepoNotFound"
+        }
+      ]
+    },
+    "accountTimeline": {
```
devinivy: My initial reaction here is that this looks more like a `timelineItem`.  I think of a "timeline" as being a list/sequence.

packages/ozone/src/api/moderation/getAccountTimeline.ts
```diff
@@ -0,0 +1,151 @@
...
+      auth,
+    )
+    cursor = data.cursor
+    for (const event of data.events) {
+      // This should never happen and the check is here only because typescript screams at us otherwise
+      if (!event.$type) {
+        continue
+      }
+
+      const day = new Date(event.createdAt).toISOString().split('T')[0]
```
devinivy: I think extracting the logic to get `YYYY-MM-DD` will probably make this a little more readable, and will be pretty self-documenting as to what's going on.

packages/ozone/src/api/moderation/getAccountTimeline.ts
```diff
@@ -0,0 +1,151 @@
...
+
+  const result = await ctx.plcClient.getAuditableLog(did)
+  for (const event of result) {
+    const day = new Date(event.createdAt).toISOString().split('T')[0]
+    if (!events[day]) {
+      events[day] = {}
+    }
+
+    events[day][event.operation.type] =
+      (events[day][event.operation.type] || 0) + 1
```
devinivy: This looks all good to me, just tossing out a stylistic idea:
```suggestion
    events[day] ??= {}
    events[day][event.operation.type] ??= 0
    events[day][event.operation.type]++
```

packages/ozone/src/api/moderation/getAccountTimeline.ts
```diff
@@ -0,0 +1,151 @@
...
+
+  const auth = await ctx.pdsAuth(ids.ToolsOzoneHostingGetAccountHistory)
+  let cursor: string | undefined = undefined
+
+  do {
+    const { data } = await ctx.pdsAgent.tools.ozone.hosting.getAccountHistory(
+      { did },
+      auth,
+    )
+    cursor = data.cursor
```
devinivy: I expected to see the `cursor` used in the request.

packages/ozone/src/api/moderation/getAccountTimeline.ts
```diff
@@ -0,0 +1,151 @@
...
+          modService.getAccountTimeline(did),
+          getAccountHistory(ctx, did),
+          getPlcHistory(ctx, did),
+        ])
+      const timelineByDay = new Map<
+        string,
+        ToolsOzoneModerationGetAccountTimeline.AccountTimelineSummary[]
+      >()
+
+      if (movEventHistory.status === 'fulfilled') {
```
devinivy: If this one fails we probably want the error to surface.

packages/ozone/src/mod-service/index.ts
```diff
@@ -1451,6 +1451,25 @@ export class ModerationService {
     // Convert map values to an array and return
     return Array.from(statsMap.values())
   }
+
+  async getAccountTimeline(did: string) {
+    const result = await this.db.db
+      .selectFrom('moderation_event')
+      .where('subjectDid', '=', did)
+      .select([
+        sql<string>`TO_CHAR(CAST(${sql.ref('createdAt')} AS TIMESTAMP), 'YYYY-MM-DD')`.as(
```
devinivy: Perhaps following the same logic used in typescript to extract the date would be a nice way to keep things consistent:
```suggestion
        sql<string>`SPLIT_PART(${sql.ref('createdAt')}, 'T', 1)`.as(
```


Could even keep these utils next to each other:
```ts
const dateFromDatetime = (date: Date) => {
  const [date] = date.toISOString().split('T')
  return date
}

const dateFromDbDatetime = (dateRef: DbRef) => {
  return sql<string>`SPLIT_PART(${dateRef}, 'T', 1)`;
}
```

packages/ozone/src/mod-service/index.ts
foysalit: checked some numbers and looks like the highest number of event rows per subjectDid is ~450K while the 2nd highest is ~130K with only a few over 50K. running this query over 450K entries take nearly 6s which is quite slow but for 130K rows it drops to 1s. it's not great but also doesn't seem to bad TBH.

however, it looks like the most expensive part of the query is the `order by day desc` clause and without that, the execution time drops to 0.5s for 450K rows. so i think we can either move the sorting to the handler or return unsorted results and let the browser handle it? quite sure the sorting operation would be a lot faster on either side than pg, wdyt?

lexicons/tools/ozone/moderation/getAccountTimeline.json
foysalit: sorry, i could use some clarification on this. im with you on using token style strings instead of arbitrary values like `plc_operation`. however, the `eventType` property in timelineItemSummary is defined as a string and does it suffice to add all these common values as `knownValues` on that prop or are you saying we should create new tokens for PLC events?

foysalit:
I believe I've addressed everything except for the naming convention of mixed event types.

lexicons/tools/ozone/moderation/getAccountTimeline.json
devinivy: If we're going to use a value like `tools.ozone.moderation.defs#timelineEventPlcOperation` here then it's probably best to document it as a token.

packages/ozone/src/api/moderation/getAccountTimeline.ts
```diff
@@ -0,0 +1,156 @@
...
+          }
+          if (day) {
+            day.push(summary)
+            timelineByDay.set(row.day, day)
+          } else {
+            timelineByDay.set(row.day, [summary])
+          }
+        }
+      } else {
+        throw new Error(movEventHistory.reason)
```
devinivy: I believe the reason would be the error itself.
```suggestion
        throw movEventHistory.reason
```

packages/ozone/src/api/moderation/getAccountTimeline.ts
```diff
@@ -0,0 +1,156 @@
...
+import { dateFromDatetime } from '../../mod-service/util'
+
+export default function (server: Server, ctx: AppContext) {
+  server.tools.ozone.moderation.getAccountTimeline({
+    auth: ctx.authVerifier.modOrAdminToken,
+    handler: async ({ params }) => {
+      const { did } = params
+      const db = ctx.db
+      const modService = ctx.modService(db)
+      const [movEventHistory, accountHistory, plcHistory] =
```
devinivy: I wonder if this is a small typo:
```suggestion
      const [modEventHistory, accountHistory, plcHistory] =
```

packages/ozone/src/api/moderation/getAccountTimeline.ts
```diff
@@ -0,0 +1,156 @@
...
+  if (!ctx.plcClient) {
+    return events
+  }
+
+  const result = await ctx.plcClient.getAuditableLog(did)
+  for (const event of result) {
+    const day = dateFromDatetime(new Date(event.createdAt))
+    events[day] ??= {}
+    const eventType =
+      PLC_OPERATION_MAP[event.operation.type] || event.operation.type
```
devinivy, line 154: If it's an unknown operation should we still log it at all?  I wonder if it's not in the `PLC_OPERATION_MAP` if it's best to skip over the unknown data.

packages/ozone/tests/__snapshots__/get-account-timeline.test.ts.snap
```diff
@@ -0,0 +1,36 @@
...
+  Object {
+    "count": 1,
+    "eventSubjectType": "account",
+    "eventType": "tools.ozone.moderation.defs#modEventTakedown",
+  },
+  Object {
+    "count": 1,
+    "eventSubjectType": "account",
+    "eventType": "plc_operation",
+  },
```
devinivy: Not sure if it's an issue with the snapshots or if there's a small bug, but still seeing `plc_operation` used as an `eventType`.

packages/ozone/src/api/moderation/getAccountTimeline.ts
foysalit: hmm... idk how often we expect event types to change in PLC log but i guess keeping it this way means we capture newly introduced/updated events without a backend change in ozone?

packages/ozone/tests/__snapshots__/get-account-timeline.test.ts.snap
foysalit: yep, needed to update the snapshot.

packages/ozone/src/mod-service/index.ts
foysalit: we moved this to a nested query to help pg with sorting perf.

PR #3942, opened 2025-06-08
Modify parsing URLs and hashtags

lamrongol:
This is an extended PR of https://github.com/bluesky-social/atproto/pull/3142 

# Original PR description
This fixes the parsing of URLs that are prefixed by a dot (.), i.e. when the user forgot a space between the end of a sentence and a link.

image

The fix is done by preventing the second alternative capture group from having a domain end in .https:// (which would be an invalid domain anyway), since that would capture out.https in the example message Check this out.https://afterdot.com.

------------------------
This PR also fixes hashtags and not only ".", for almost patterns like following:
【Website】https://example.com/


tpaulshippy:
How many PRs does it take to fix a bug? Maybe third time is the charm?

PR #3945, opened 2025-06-10
:sparkles: Adds safelink module

foysalit:


packages/ozone/src/db/migrations/20250609T110704000Z-safelink.ts
```diff
@@ -0,0 +1,71 @@
...
+      col.defaultTo(sql`now()`).notNull(),
+    )
+    .addColumn('updatedAt', 'timestamptz', (col) =>
+      col.defaultTo(sql`now()`).notNull(),
+    )
+    .addColumn('comment', 'text')
+    .addUniqueConstraint('safelink_url_pattern_key', ['url', 'pattern'])
+    .execute()
+
+  await db.schema
```
foysalit, line 31: we need better indexing here to support domain search queries.

packages/ozone/src/db/migrations/20250609T110704000Z-safelink.ts
```diff
@@ -0,0 +1,71 @@
...
+    .addColumn('reason', 'varchar', (col) => col.notNull())
+    .addColumn('createdBy', 'varchar', (col) => col.notNull())
+    .addColumn('createdAt', 'varchar', (col) =>
+      col.defaultTo(sql`now()`).notNull(),
+    )
+    .addColumn('comment', 'text')
+    .execute()
+
+  await db.schema
+    .createTable('safelink')
```
devinivy: Perhaps `safelink_rule` for consistency with the parlance used in the lexicon, etc?

lexicons/tools/ozone/safelink/defs.json
```diff
@@ -0,0 +1,155 @@
...
+        },
+        "comment": {
+          "type": "string",
+          "description": "Optional comment about the decision"
+        }
+      }
+    },
+    "eventType": {
+      "type": "string",
+      "knownValues": ["#addRule", "#updateRule", "#removeRule"]
```
devinivy: If you want to use the token values, those strings should be fully-qualified e.g. `tools.ozone.safelink.defs#addRule`.  If you don't need the safelink to be extensible by systems outside of `tools.ozone`, you also may not need tokens.

devinivy: Same goes for some of the other token-using values in here such as `patternType`.

It looks like some other ozone lexicons use unqualified tokens within the lexicon, but then the implementation uses the full token value.  E.g. `reviewState` on `subjectStatusView` lists the known value `#reviewOpen`, but the actual string value used in production is `tools.ozone.moderation.defs#reviewOpen`.

packages/ozone/src/api/safelink/queryRules.ts
```diff
@@ -0,0 +1,57 @@
...
+      })
+
+      return {
+        encoding: 'application/json',
+        body: {
+          cursor: result.cursor,
+          rules: result.rules.map((rule) => ({
+            url: rule.url,
+            pattern: rule.pattern,
+            action: rule.action,
```
devinivy, line 47: Could we use `formatEvent()` here?

packages/ozone/src/db/schema/safelink.ts
```diff
@@ -0,0 +1,53 @@
...
+export const safelinkEventTableName = 'safelink_event'
+
+export type SafelinkEventType =
+  | 'tools.ozone.safelink.defs#addRule'
+  | 'tools.ozone.safelink.defs#updateRule'
+  | 'tools.ozone.safelink.defs#removeRule'
+
+export type SafelinkPatternType =
+  | 'tools.ozone.safelink.defs#domain'
+  | 'tools.ozone.safelink.defs#url'
```
devinivy: The lexicon constants can be used, it would look like this:
```suggestion
export type SafelinkPatternType =
  | typeof DOMAIN
  | typeof URL
```

We might consider colocating some of these types with the logic/helpers that also deal with them in `packages/ozone/src/api/util.ts`.

packages/ozone/src/db/schema/safelink.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export interface SafelinkRule {
+  id: GeneratedAlways<number>
+  url: string
+  pattern: SafelinkPatternType
+  action: SafelinkActionType
+  reason: SafelinkReasonType
+  createdBy: string
+  createdAt: string
+  updatedAt: GeneratedAlways<string>
```
devinivy: I think in other ozone models we tend to control the `updatedAt` time in the app code (sets, verification, settings), so this may end up being an exception.

packages/ozone/src/db/migrations/20250609T110704000Z-safelink.ts
```diff
@@ -0,0 +1,71 @@
...
+    .createTable('safelink_event')
+    .addColumn('id', 'bigserial', (col) => col.primaryKey())
+    .addColumn('eventType', 'varchar', (col) => col.notNull())
+    .addColumn('url', 'varchar', (col) => col.notNull())
+    .addColumn('pattern', 'varchar', (col) => col.notNull())
+    .addColumn('action', 'varchar', (col) => col.notNull())
+    .addColumn('reason', 'varchar', (col) => col.notNull())
+    .addColumn('createdBy', 'varchar', (col) => col.notNull())
+    .addColumn('createdAt', 'varchar', (col) =>
+      col.defaultTo(sql`now()`).notNull(),
```
devinivy: When `createdAt` is a varchar with a datetime (`now()`) as a default, how does postgres cast the datetime to a string?  That may not come out the way we want it, may be inconsistent with datetime formats generated by the application itself.

packages/ozone/src/db/migrations/20250609T110704000Z-safelink.ts
```diff
@@ -0,0 +1,71 @@
...
+    .createIndex('safelink_reason_idx')
+    .on('safelink_rule')
+    .column('reason')
+    .execute()
+
+  await db.schema
+    .createIndex('safelink_updated_at_idx')
+    .on('safelink_rule')
+    .column('updatedAt')
+    .execute()
```
devinivy:
```suggestion
  await db.schema
    .createIndex('safelink_rule_action_idx')
    .on('safelink_rule')
    .column('action')
    .execute()

  await db.schema
    .createIndex('safelink_rule_reason_idx')
    .on('safelink_rule')
    .column('reason')
    .execute()

  await db.schema
    .createIndex('safelink_rule_updated_at_idx')
    .on('safelink_rule')
    .column('updatedAt')
    .execute()
```

packages/ozone/src/db/migrations/20250609T110704000Z-safelink.ts
```diff
@@ -0,0 +1,71 @@
...
+  await db.schema
+    .createIndex('safelink_event_url_idx')
+    .on('safelink_event')
+    .column('url')
+    .execute()
+
+  await db.schema
+    .createIndex('safelink_event_id_created_at_idx')
+    .on('safelink_event')
+    .columns(['id', 'createdAt'])
```
devinivy: it looks like event pagination uses `(createdAt, id)`, where the index here is `(id, createdAt)`.  Would just using `id` for pagination be sufficient?

packages/ozone/src/db/migrations/20250609T110704000Z-safelink.ts
```diff
@@ -0,0 +1,71 @@
...
+    .addColumn('pattern', 'varchar', (col) => col.notNull())
+    .addColumn('action', 'varchar', (col) => col.notNull())
+    .addColumn('reason', 'varchar', (col) => col.notNull())
+    .addColumn('createdBy', 'varchar', (col) => col.notNull())
+    .addColumn('createdAt', 'varchar', (col) =>
+      col.defaultTo(sql`now()`).notNull(),
+    )
+    .addColumn('updatedAt', 'varchar', (col) =>
+      col.defaultTo(sql`now()`).notNull(),
+    )
```
devinivy: Same here—we have a postgres datetime cast into a varchar.  I just looked it up and I believe the string format it will use is `2025-07-01 16:38:47.749+00`.  Maybe easiest to just manage the datetime strings in the application.

packages/ozone/src/db/migrations/20250609T110704000Z-safelink.ts
devinivy: I assume this is intended to support pagination—I believe the way pagination is currently written it uses `(createdAt, id)`.

packages/ozone/src/db/migrations/20250609T110704000Z-safelink.ts
```diff
@@ -0,0 +1,71 @@
+import { Kysely, sql } from 'kysely'
```
devinivy: I am interested in the various use cases for the filter params on rules and events, which ones need to be cheap and which can be expensive to run.  I don't think we have indexes that support them at the moment.

For example, the index on `safelink_rule.action` may be used to narrow down the result set to many thousands of results but the pagination index wouldn't be able to be used at that point.  If there are filter parameters we're supporting that don't have a specific use case, removing those might help narrow the space a bit.

packages/ozone/src/db/migrations/20250609T110704000Z-safelink.ts
foysalit: right yeah i think i dropped the idea of paginating by updatedAt along the way. will get rid of it.

packages/ozone/src/db/migrations/20250609T110704000Z-safelink.ts
foysalit: i think we need both since we may need to bulk import rules at some point and same rows may end up with same value. however, we dont really have such bulk endpoint at the moment so this might be an overkill?

packages/ozone/src/api/safelink/queryRules.ts
foysalit: rule and event are ever so slightly different and this is the only place where we return a rule so im not sure if it makes sense to try to fit the formatEvent helper for this case or create a separate helper just for this.

lexicons/tools/ozone/safelink/defs.json
foysalit: hmm yeah i never considered the cross namespace case. i think these can safely be just knownValues of strings and dont need to be tokens.

packages/ozone/src/db/migrations/20250609T110704000Z-safelink.ts
foysalit: for filtering rules, we mostly want to be able to answer these questions 
1. do we have rules for these domains: of.com, facebook.com etc. or these urls: https://abc.xyz, https://def.xyz - these will mostly be used by mods to verify what rules, if any, are in place for certain domains and urls
2. filter for all the rules that have a certain action type (ban, whitelist etc.) and perhaps by reason type among each subgroup action type.

for filtering events, i think we just need to be able to paginate efficiently and then filter down to events for a certain url-patternType combination.

packages/ozone/src/db/migrations/20250609T110704000Z-safelink.ts
devinivy: > same rows may end up with same value

I can see how that would be the case with `createdAt`, but the `id` is a primary key so it should be unique.  If that works, one upside is that additional indexes wouldn't be needed on events or rules for pagination.

packages/ozone/src/api/safelink/queryRules.ts
devinivy: Oh yeah, oops!

devinivy (approved):
Looking good!

Still may need to take a closer look at indexes, particularly for `safelink_rule` which doesn't have an index supporting pagination.  But sounds like you're aware and want to cross that bridge when we get there 👍

packages/ozone/src/db/migrations/20250609T110704000Z-safelink.ts
foysalit: ahh you're right! yeah i think we can just id in this case.

PR #3953, opened 2025-06-13
Improve error reporting and logging

matthieusieben:


packages/oauth/oauth-provider/src/lib/util/zod-error.ts
```diff
@@ -2,12 +2,29 @@ import { ZodError } from 'zod'
...
+
+function extractMessage({ message }: { message: string }): string | undefined {
+  // message serialization defaults to JSON.stringify(err.issues)
+  if (!message.startsWith('[') && !message.endsWith(']')) {
+    // If the error message does not end with ']', it means it is not a
+    // serialization of the issues, so we can return it directly.
+    return message
   }
 
   return undefined
```
matthieusieben: Example of message:

```
Validation of "" body parameter failed: Invalid input: [
  {
    "code": "invalid_union",
    "unionErrors": [
      {
        "issues": [
          {
            "code": "invalid_type",
            "expected": "string",
            "received": "undefined",
            "path": [
              "body",
              "client_id"
            ],
            "message": "Required"
          },
          {
            "code": "invalid_literal",
            "expected": "urn:ietf:params:oauth:client-assertion-type:jwt-bearer",
            "path": [
              "body",
              "client_assertion_type"
            ],
            "message": "Invalid literal value, expected \"urn:ietf:params:oauth:client-assertion-type:jwt-bearer\""
          },
          {
            "code": "invalid_type",
            "expected": "string",
            "received": "undefined",
            "path": [
              "body",
              "client_assertion"
            ],
            "message": "Required"
          }
        ],
        "name": "ZodError"
      },
      {
        "issues": [
          {
            "code": "invalid_type",
            "expected": "string",
            "received": "undefined",
            "path": [
              "body",
              "client_id"
            ],
            "message": "Required"
          },
          {
            "code": "invalid_type",
            "expected": "string",
            "received": "undefined",
            "path": [
              "body",
              "client_secret"
            ],
            "message": "Required"
          }
        ],
        "name": "ZodError"
      },
      {
        "issues": [
          {
            "code": "invalid_type",
            "expected": "string",
            "received": "undefined",
            "path": [
              "body",
              "client_id"
            ],
            "message": "Required"
          }
        ],
        "name": "ZodError"
      }
    ],
    "path": [
      "body"
    ],
    "message": "Invalid input"
  }
]
```

packages/oauth/oauth-provider/src/lib/util/zod-error.ts
```diff
@@ -2,12 +2,31 @@ import { ZodError } from 'zod'
...
+    const message = extractMessage(err)
+    if (message) return `Invalid request: ${message}`
+  }
+
+  return undefined
+}
+
+function extractMessage({ message }: { message: string }): string | undefined {
+  // message serialization defaults to JSON.stringify(err.issues)
+  if (!message.startsWith('[') && !message.endsWith(']')) {
```
devinivy: If the purpose of the logic is to detect a serialized JSON array, I believe you'd want to use `||` here.

packages/oauth/oauth-provider/src/router/create-api-middleware.ts
```diff
@@ -452,8 +452,7 @@ export function createApiMiddleware<
           }
 
           if (err instanceof AccessDeniedError && err.parameters.redirect_uri) {
-            // Prefer logging the cause
-            onError?.(req, res, err.cause ?? err, 'Authorization failed')
+            onError?.(req, res, err, 'Authorization request not acceptable')
```
devinivy: Here it says "not acceptable" and below it says "rejected."  Is there a difference?  If not, I'd recommend adopting the latter.

devinivy: There is another that says "denied."  I can't tell if they are intended to be distinguished or not, they all roughly seem to mean the same thing.

packages/pds/src/auth-routes.ts
```diff
@@ -35,12 +40,34 @@ export const createRouter = ({ oauthProvider, cfg }: AppContext): Router => {
...
 }
+
+function ignoreError(err: unknown): boolean {
+  if (err instanceof InvalidGrantError) {
+    return (
+      err.error_description === 'Refresh token expired' ||
+      err.error_description === 'Refresh token exceeded inactivity timeout' ||
+      err.error_description === 'Invalid refresh token'
+    )
+  }
```
devinivy: I am less sure about these ones, since this indicates users are getting logged out, right?  That may be interesting to be able to measure.

packages/oauth/oauth-provider/src/lib/util/zod-error.ts
matthieusieben: this code was updated to rely on the issue messages instead of error messages, removing the need for this heuristic.

packages/oauth/oauth-provider/src/router/create-api-middleware.ts
```diff
@@ -451,20 +471,21 @@ export function createApiMiddleware<
...
 
-            return { url }
+              return { json: { url } }
+            } catch {
+              // Unable to build redirect URL, ignore
+            }
           }
 
-          throw err
+          return buildErrorJsonResponse(err)
```
matthieusieben, line 489: Because we already logged the error (`onError` above), we don't want to re-throw here (to avoid double logging of the error)

packages/oauth/oauth-provider/src/router/create-api-middleware.ts
```diff
@@ -507,22 +528,25 @@ export function createApiMiddleware<
...
```
matthieusieben, line 550: Because we already logged the error (`onError` above), we don't want to re-throw here (to avoid double logging of the error)

PR #3966, opened 2025-06-17
Rename `filter` -> `include`

mozzius:
Just renamed the lexicon then renamed things until typecheck passed

Does not change the protos

### DIFF
--- a/packages/bsky/tests/views/notifications.test.ts
+++ b/packages/bsky/tests/views/notifications.test.ts
@@ -925,7 +925,7 @@ describe('notification views', () => {
 
     // Defaults
     const fp: FilterablePreference = {
-      filter: 'all',
+      include: 'all',
       list: true,
       push: true,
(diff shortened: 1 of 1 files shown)


lexicons/app/bsky/notification/defs.json
```diff
@@ -8,17 +8,17 @@
...
       "type": "object",
-      "required": ["filter", "push"],
+      "required": ["include", "push"],
       "properties": {
-        "filter": { "type": "string", "knownValues": ["all", "accepted"] },
+        "include": { "type": "string", "knownValues": ["all", "accepted"] },
         "push": { "type": "boolean" }
       }
     },
     "filterablePreference": {
```
devinivy, line 17: Think we should change this name, now that we're not saying that it applies a "filter"?

devinivy (approved):

PR #3971, opened 2025-06-18
Replace slice() with subarray() in car file parsing

pfrazee:
See https://bsky.app/profile/did:plc:ragtjsm2j2vknwkz3zp4oxrd/post/3lru2skt7wk2m

Somewhere around Node 16, slice became deprecated and its semantics changed so that it copies into new buffers. On large CAR files which have been read into buffers, this can cause parsing to take a very long time. The `subarray()` method avoids the allocations.

I'd suggest reviewing the entire codebase for this. I did find some other usages, but felt a little gunshy about just doing a search & replace. Uint8Array subarray support is broad so I don't think this creates compat issues in eg web or RN (though RN is a wildcard on this kind of thing).

### DIFF
--- a/.changeset/gold-mice-burn.md
+++ b/.changeset/gold-mice-burn.md
@@ -1,0 +1,5 @@
+---
+"@atproto/repo": patch
+---
+
+improve performance of reading repos from car files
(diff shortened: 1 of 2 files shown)


devinivy (approved):
Nice find!

PR #3973, opened 2025-06-18
OAuth: Improve error reporting

matthieusieben:
Motivations are:
1) To return more accurate `error` responses during authorization flows
2) To return more accurate `error` responses during PAR
3) To rework the class hyerarchy around a new `AuthorizationError` parent class (instead of `AccessDeniedError`, which is actually a special case of authorization error)


packages/oauth/oauth-provider/src/oauth-provider.ts
```diff
@@ -522,7 +523,7 @@ export class OAuthProvider extends OAuthVerifier {
       // > Since initial processing of the pushed authorization request does not
       // > involve resource owner interaction, error codes related to user
       // > interaction, such as "access_denied", are never returned.
-      if (err instanceof AccessDeniedError) {
+      if (err instanceof AuthorizationError && !isPARResponseError(err.error)) {
```
matthieusieben, line 526: Prefer throwing the original error if it is a valid PAR error code, improving error responses for clients.

packages/oauth/oauth-provider/src/errors/login-required-error.ts
```diff
@@ -1,22 +1,12 @@
...
-
-  static from(
-    parameters: OAuthAuthorizationRequestParameters,
-    cause?: unknown,
-    fallbackError?: string,
-  ): LoginRequiredError {
-    if (cause instanceof LoginRequiredError) return cause
-
-    return new LoginRequiredError(parameters, fallbackError, cause)
-  }
```
matthieusieben, line 21: unused code

packages/oauth/oauth-provider/src/errors/invalid-parameters-error.ts
```diff
@@ -1,12 +0,0 @@
...
-
-export class InvalidParametersError extends AccessDeniedError {
-  constructor(
-    parameters: OAuthAuthorizationRequestParameters,
-    error_description: string,
-    cause?: unknown,
-  ) {
-    super(parameters, error_description, 'invalid_request', cause)
-  }
-}
```
matthieusieben, line 12: Redundent with `AuthorizationError`

packages/oauth/oauth-provider/src/errors/authorization-error.ts
```diff
@@ -0,0 +1,45 @@
...
+    return new AuthorizationError(
+      parameters,
+      payload.error_description,
+      isAuthorizationResponseError(payload.error)
+        ? payload.error // Propagate "error" derived from the cause
+        : rootCause(cause) instanceof OAuthError
+          ? 'invalid_request'
+          : 'server_error',
+      cause,
+    )
```
matthieusieben, line 36: per spec, AuthorizationError are only allowed to contain specific `error` values.

This "new" implementation tries to keep the original `error` (`buildErrorPayload` will turn any JS `Error` -- including `OAuthError`s -- into an `error` code).

packages/oauth/oauth-provider/src/errors/access-denied-error.ts
```diff
@@ -1,39 +1,12 @@
...
-  // OAuth but we allow them because they provide better feedback to the client
-  // (in particular when SSO is used).
-  | OidcAuthenticationErrorResponse
-  // This error is defined by rfc9396 (not part of the OAuth 2.1 or OIDC). But
-  // since, in ATproto flavoured OAuth, client registration is a dynamic part of
-  // the authorization process, we allow it.
-  | 'invalid_authorization_details'
-
-export class AccessDeniedError extends OAuthError {
+export class AccessDeniedError extends AuthorizationError {
```
matthieusieben, line 4: `AccessDeniedError` used to be defined as parent class for all Authroization Request Error.

The class hyerarchy was changed by introducing a new `AuthorizationError` as common ancestror. `AccessDeniedError` is now a simple special case of `AuthorizationError`, better reflecting reality.

packages/oauth/oauth-provider/src/errors/authorization-error.ts
```diff
@@ -0,0 +1,45 @@
...
+  }
+}
+
+function rootCause(err: unknown): unknown {
+  while (err instanceof Error && err.cause != null) {
+    err = err.cause
```
bricefriha, line 42: Would we be able to yield here instead, do you think?

```suggestion
    yield err.cause;
```

matthieusieben: I'm not interested in intermediary causes, only the root one. I like generator functions very much but I'm not sure how to conveniently use one here.

PR #3974, opened 2025-06-18
❇️ Support new reporting categories in ozone

foysalit:
This PR removes report type validation based on static reasons and replaces it with validation based on reason types defined in the labeler's service profile.


packages/ozone/src/mod-service/profile.ts
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    if (
+      this.cache?.profile?.reasonTypes &&
+      Array.isArray(this.cache.profile.reasonTypes)
+    ) {
+      const supportedReasonTypes = this.cache.profile.reasonTypes
+
+      // Check if the reason type is directly supported
+      if (supportedReasonTypes.includes(reasonType)) {
```
bricefriha, line 126: Personally, I feel like this entire section can be simplified to:

```suggestion
      const mappedOldReason = NEW_TO_OLD_REASON_MAPPING[reasonType];
      const supportedReasonTypes = this.cache.profile.reasonTypes;

      if (!supportedReasonTypes.includes(reasonType) && 
          !mappedOldReason &&
          !supportedReasonTypes.includes(mappedOldReason)) {
              throw new InvalidRequestError(`Invalid reason type: ${reasonType}`);
      }
      return reasonType;
```

But I also see the point of detailing each part

PR #3976, opened 2025-06-19
Re-export all types & utilities needed to instantiate an OAuth client

matthieusieben:
## Improved exports

In order to instantiate a NodeJS based oauth client today, one must import stuff from 
`@atproto/jwk` , `@atproto/jwk-jose` and `@atproto/oauth-client-node`. In addition to being non-intuitive, this also requires users of our SDK to maintain proper versions of all the dependencies (NPM's resolution algorithm could result in different versions of these to be used, in particular when declared accross multiple `packages.json` files).

These changes make it so that `@atproto/oauth-client-node` re-exports all that is required to instantiate a client.

## Improved customizability

The `handleResolver` option can now be overridden when instantiating a `NodeOAuthClient`


packages/oauth/oauth-client/src/oauth-client.ts
```diff
@@ -207,14 +214,23 @@ export class OAuthClient extends CustomEventTarget<OAuthClientEventMap> {
...
+            ),
+        handleResolver instanceof CachedHandleResolver && !handleCache
+          ? handleResolver
+          : new CachedHandleResolver(
+              typeof handleResolver === 'string' ||
+              handleResolver instanceof URL
+                ? new XrpcHandleResolver(handleResolver, { fetch })
+                : handleResolver,
+              handleCache,
+            ),
```
devinivy, line 233: This is fairly dense, if you think there's a straightforward way to make it more readable that might be nice.

PR #3977, opened 2025-06-19
Perform a bi-directional check when resolving identity from did

matthieusieben:


packages/internal/identity-resolver/src/identity-resolver.ts
```diff
@@ -50,15 +48,38 @@ export class IdentityResolver {
...
+      const did = await this.handleResolver.resolve(handle, options)
+      if (did !== document.id) {
+        throw new TypeError(
+          `Did document for "${did}" does not match the handle "${handle}"`,
+        )
+      }
+    }
+
+    return document
   }
```
devinivy: I don't think this is the behavior we want for atproto DID resolution.  The status of your handle doesn't make your identity any more or less legitimate, and we perform many operations on your identity without a valid handle (e.g. validating your records). One reason we use strong links based on DID is to avoid reliance on handle status.

A "resolved handle" is a DID, but a DID isn't a resolved handle.  So I wonder if the more appropriate input type here might be `AtprotoDid`.

packages/internal/identity-resolver/src/identity-resolver.ts
```diff
@@ -50,15 +48,38 @@ export class IdentityResolver {
...
-    return this.didResolver.resolve(did, options)
+    const document = await this.didResolver.resolve(did, options)
+
+    const handle = getValidHandle(document)
+    if (handle) {
+      options?.signal?.throwIfAborted()
+
+      const did = await this.handleResolver.resolve(handle, options)
+      if (did !== document.id) {
+        throw new TypeError(
```
devinivy: I am not sure `TypeError` is a fit for this case since all the values are used appropriately, it's just the state of the network that doesn't line up.  Not a big deal though.

(I also try to use `TypeError` with care since it is used by libraries distinguishing system errors from application/library/operational errors, as in bounce for example https://github.com/hapijs/bounce/blob/c2ee4d7c71d339e740d89e7f84664a17655ddcb9/lib/index.js#L9-L29)

packages/internal/identity-resolver/src/identity-resolver.ts
devinivy: Okay sorry!  Realizing that "identity resolution" here is distinct from atproto DID resolution.  Taking a closer look, and sort of interested to get a better sense for the use cases where we start with a DID lookup.

packages/internal/identity-resolver/src/identity-resolver.ts
matthieusieben: did this to match existing behavior. Adapted to add explicit error classes.

packages/internal/identity-resolver/src/identity-resolver.ts
matthieusieben: When signing in, the user can either provide a `handle`, `did` or `pdsOrigin`. In case they provide a `did`, this change will ensure that the did, if it has a handle, resolves bi-dirrectionaly.

This method can also be used as a reliable way to determine a `did -> handle` mapping

devinivy: This is a tricky one since looking up an identity by handle essentially requires (by definition) having a valid, bidirectionally-resolving handle—so that explains why it happens in `getDocumentFromHandle()`.  Looking up an identity by DID does not require that, of course, so it's not obvious whether it's necessary to add here.

The `resolve()` method always has the ability to report the special value defined in the protocol, `'handle.invalid'`, to provide all identity information while indicating the user's handle is invalid.  Apps leverage this so that the user isn't erased from the network just because something went wrong with their handle, or the server is having a temporary issue performing handle resolution (which would be very brittle).

I do think that this is more in line with how the protocol usually treats identity: you don't need to have a valid/bidirectional handle to have an active identity, but we also qualify that by saying the identity currently has an "invalid handle."  My inclination would be to stick to that.

Another consideration is that if we _are_ going to enforce two rules for identity (a. valid/bidi handle, b. has a pds) then we might want each public method to always either a. enforce all or b. enforce minimal necessary.  Right now `getDocument()` will enforce one, and `resolve()` will enforce both.

I am interested to get your thoughts!  What issue did you run into that motivated this?

matthieusieben: I basically wanted to be able to reliably rely on `handle`. That value is used in the `oauth-client`, and I wanted to be able to also rely on it from the statusphere example app (to compute the did -> handle map).

My latest commit should better show my intent, the way it's used, and should be more compliant with the way things are supposed to work.

matthieusieben: https://github.com/bluesky-social/atproto/pull/3977/commits/cb880febbba501421d55754ee626a2cb0c2720c0

packages/internal/identity-resolver/src/identity-resolver.ts
```diff
@@ -80,9 +96,9 @@ export class IdentityResolver {
 
     const document = await this.didResolver.resolve(did, options)
 
-    // Ensure that the handle is included in the document
-    if (!document.alsoKnownAs?.includes(`at://${handle}`)) {
-      throw new TypeError(
+    // Enforce bi-directional resolution
+    if (handle !== getHandle(document)) {
```
devinivy: Ahh I didn't notice we've been treating handles as case-sensitive here.  We probably want to fix that, but fine to do so in separate work if you'd like, as it's unrelated to these changes.

I think you've also fixed a bug here, where we now use the first `at://` entry rather than any matching `at://` entry.

devinivy (approved):

PR #3982, opened 2025-06-21
Identity resolver's `resolve()` method returns valud consistent with `resolveIdentity`

matthieusieben:


packages/internal/identity-resolver/src/identity-resolver.ts
```diff
@@ -13,14 +13,19 @@ import {
 import { IdentityResolverError } from './identity-resolver-error'
 import { asNormalizedHandle, extractNormalizedHandle } from './util'
 
-export type ResolvedIdentity = {
-  document: DidDocument<AtprotoIdentityDidMethods>
```
devinivy, line 17: I wonder if moving to `didDoc` also comes with the small benefit of not conflicting with the browser global `document` 👍

.changeset/few-rivers-chew.md
```diff
@@ -0,0 +1,5 @@
+---
+"@atproto-labs/identity-resolver": minor
+---
+
+Identity resolver's `resolve()` method returns valud consistent with `com.atproto.identity.resolveIdentity`
```
devinivy:
```suggestion
Identity resolver's `resolve()` method returns value consistent with `com.atproto.identity.resolveIdentity`
```

packages/internal/identity-resolver/src/identity-resolver.ts
```diff
@@ -58,22 +63,24 @@ export class IdentityResolver {
       : undefined
 
     return {
-      document,
-      handle: resolvedDid === did ? handle : undefined,
+      did: document.id,
+      didDoc: document,
+      handle: handle && resolvedDid === did ? handle : 'handle.invalid',
```
devinivy: Perhaps at some point we should look into exporting the special value as a constant somewhere.

PR #3991, opened 2025-06-23
❇️ Ozone user agent

foysalit:
This PR adds an optional `userAgent` property to moderation events, allowing for better tracing of where events are coming from. The primary property of the userAgent object is just a `name` string while it allows for any additional metadata passing via the `extra` object.


lexicons/tools/ozone/moderation/defs.json
```diff
@@ -894,6 +896,21 @@
...
+    "userAgent": {
+      "type": "object",
+      "description": "User agent information for tracing the source of the action",
+      "required": ["name"],
+      "properties": {
+        "name": {
+          "type": "string",
+          "description": "Name/identifier of the source (e.g., 'automod', 'ozone/workspace')"
+        },
+        "extra": {
```
devinivy: `extra` works for me, though I also wonder if something more neutral about the info's "extra-ness" like `details` could also be good.

lexicons/tools/ozone/moderation/defs.json
```diff
@@ -894,6 +896,21 @@
...
+    },
+    "userAgent": {
+      "type": "object",
+      "description": "User agent information for tracing the source of the action",
+      "required": ["name"],
+      "properties": {
+        "name": {
+          "type": "string",
+          "description": "Name/identifier of the source (e.g., 'automod', 'ozone/workspace')"
+        },
```
devinivy, line 908: I assume also using an actual user-agent string (in [this](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/User-Agent) sense) would also be welcome.

lexicons/tools/ozone/moderation/emitEvent.json
```diff
@@ -44,9 +44,19 @@
...
             },
-            "createdBy": { "type": "string", "format": "did" }
+            "createdBy": {
+              "type": "string",
+              "format": "did"
+            },
+            "userAgent": {
+              "type": "ref",
+              "ref": "tools.ozone.moderation.defs#userAgent"
+            }
```
devinivy: My only confusion about this kind of usage is that there is also the `user-agent` header which will be present in these requests.

packages/dev-env/src/moderator-client.ts
```diff
@@ -1,5 +1,6 @@
 import {
```
devinivy, line 1: May be worth a changeset for this dev-env change.

packages/ozone/src/mod-service/views.ts
```diff
@@ -141,6 +141,12 @@ export class ModerationViews {
       createdAt: row.createdAt,
       subjectHandle: row.subjectHandle ?? undefined,
       creatorHandle: row.creatorHandle ?? undefined,
+      userAgent: row.userAgent
+        ? {
+            name: row.userAgent.name,
+            extra: row.userAgent.extra as { [x: string]: unknown } | undefined,
```
devinivy: Could this cast be addressed by updating the type in the `moderation_event` schema?

packages/ozone/src/db/migrations/20250618T180246000Z-add-user-agent-to-moderation-event.ts
```diff
@@ -0,0 +1,21 @@
...
+  await db.schema
+    .alterTable('moderation_event')
+    .addColumn('userAgent', 'jsonb')
+    .execute()
+
+  await db.schema
+    .createIndex('moderation_event_user_agent_name_idx')
+    .on('moderation_event')
+    .expression(sql`("userAgent" ->> 'name')`)
+    .execute()
```
devinivy: If many events end-up being created by the same system, with the same user agent name, even this index may not save us!  Do we think that is likely?  What is the most similar use case in `queryEvents` (where I assume this index is used) and how is it indexed?

foysalit: i think that's very likely. to start with, automod accounts for the majority of report and label events. then, i think for in-app reporting, we'd wanna fill in modTool name as something like `bsky-app` which would account for the second largest part of the reports.

lexicons/tools/ozone/moderation/emitEvent.json
foysalit: decided to change it to `modTool` instead.

lexicons/tools/ozone/moderation/defs.json
foysalit: moved this to meta

packages/ozone/src/db/migrations/index.ts
```diff
@@ -25,3 +25,4 @@ export * as _20250221T132135150Z from './20250221T132135150Z-member-details'
 export * as _20250404T201720309Z from './20250404T201720309Z-subject-status-sort-idxs'
 export * as _20250415T201720309Z from './20250415T201720309Z-verification'
 export * as _20250417T201720309Z from './20250417T201720309Z-firehose-cursor'
+export * as _20250618T180246000Z from './20250618T180246000Z-add-mod-tool-to-moderation-event'
```
devinivy, line 29: I know there are a couple PRs with migrations in flight—just a reminder to make sure that when they land the migrations land in order that they'll run.

devinivy (approved):

packages/ozone/src/db/migrations/20250618T180246000Z-add-user-agent-to-moderation-event.ts
devinivy: Got it—then just worth keeping in mind that this index can help filter the result set, but then postgres may not be able to efficiently paginate in some cases.

PR #3999, opened 2025-06-26
xrpc-server & lex-cli rework

matthieusieben:
## Context

We want to be able to validate auth based on query params & input. At its core, this PR adds `params` and `input` as properties in the context used for the auth verifiers.

This behavio required that the `lex-cli` code be adapted to support the new types. This was done by removing the had-oc types created in the `lexicon/index.ts`, and replacing them with equivalent generic types exported by `xrpc-server`.

`lex-cli` no longer generates a `Handler` and `HandlerReqCtx` type for every method. Instead, it relies on a generic type exposed by `xrpc-server`:

```ts
import { MethodHandler } from '@atproto/xrpc-server'

type MyMethodHandler = MethodHandler<
  Auth, // extends { credentials: unknown }
  Params, // extends Record<string, undefined | Primitive | Primitive[]>
  Input, // extends void | HandlerInput
  Output, // extends void | ErrorResult | HandlerSuccess 
>
```

thanks to this generic type, the params, input and auth value can be properly bound together.

the use of this generic type in the methods of the `XrpcServer` class also means that `addMethod({ auth, handler })` now properly types the handler's context (`auth` prop) using the return value of the `auth` verifier.

The biggest change in terms of runtime is the way the handler middleware is created. The handler used to consist of a bunch of middlewares depending on each other through a shared `kRequestLocals` property in `req.locals`. This is no longer needed as all the request processing steps (except body parsing) are now performed from a single middleware (`XrpcServer.createHandler()`).

## Notes

- Because we now need to be able to use the code-gen `QueryParams` as generic in the `MethodHandler`, we needed to change it from an `interface` to a `type` (see [here](https://www.typescriptlang.org/play/?#code/C4TwDgpgBACgTgSwLYOAgbtAvFAzsRAOwHMoAfKQgVyQCMI5ypaB7FgGwgENCAoUSLC5wuSXFBwAlCAGMWcACYAefEWIAaWIhRpMTeMlQYIAbQC6TKoQUQAZgkIQFAPl79w0ABI8FnOEpgoCAAPYAhrcRhhUXEcKJExZwkoAAo5QjDQgC4oAG8oMGixHMCAXwBKCST0FgQFNwFoAFkQeJjk3N4oKAU6nNUHYl5S3hCweWAoRqgW72s-ZLnfBiUWtsS3BzC4Wy4ZZtai3AAmPK6evrwCQeG3MYmpjxmQJb9TnFeVtaPjpIB6P5QACicDg8hyAElrCE8AhiIQuMAqHBoLZ5I9BAByAYkTFQBDiFC4XCDfGEDHQTHfBInTFAA))
- There used to be a lot of inter-dependencies in `xrpc-server/src/types`. I reduced these by 1) moving RateLimit specific types in `rate-limiter.ts`, using a more generic type as `RateLimiterContext` since the RateLimiter's *internal code* does not really care about `auth` props. 2) I moved error classes in a new `errors.ts`.
- Because I introduced new breaking types for `MethodHandler` and `StreamHandler`, I also renamed a bunch of existing types that are public (exported) but (almost) not used in the rest of the codebase.

### DIFF
--- a/.changeset/lazy-falcons-repeat.md
+++ b/.changeset/lazy-falcons-repeat.md
@@ -1,0 +1,5 @@
+---
+"@atproto/xrpc-server": patch
+---
+
+Ignore `__proto__` in query string params.


.changeset/nasty-icons-peel.md
```diff
@@ -0,0 +1,5 @@
+---
+"@atproto/xrpc-server": patch
+---
+
+`catchall` handler wil not be triggered if the XRPC method is not a valid NSID
```
devinivy:
```suggestion
`catchall` handler will not be triggered if the XRPC method is not a valid NSID
```

.changeset/swift-flowers-invent.md
```diff
@@ -0,0 +1,5 @@
+---
+"@atproto/xrpc-server": patch
+---
+
+Make `params` and `input` available during auth
```
devinivy: Just a reminder to update this in case we do change the availability of `input`.  Sounds like that was a small open question.

.changeset/twenty-trainers-drive.md
```diff
@@ -0,0 +1,5 @@
+---
+"@atproto/pds": minor
+---
+
+Replace export of `SkeletonHandler` with an export of the `AppBskyFeedGetFeedSkeleton` namespace
```
devinivy: There is a bit of snag with a minor semver bump on the PDS.

Right now the docker tag on the PDS distribution is the same as the PDS package version.  The compose that we ship uses the `0.4` tag, and it receives nightly updates—the idea is that `0.5` will need to be manually adopted by self-hosters, but we can ship patches over the course of `0.4` automatically.  So if we bump the semver on the PDS, either the docker tags will start to be out of sync or we'll break the distribution's auto-updating.

I think the simplest option might be to preserve the `SkeletonHandler` export from the PDS package.

packages/xrpc-server/tests/auth.test.ts
```diff
@@ -115,7 +115,7 @@ describe('Auth', () => {
     expect(validated.lxm).toEqual(lxm)
   })
 
-  it('fails on bad auth before invalid request payload.', async () => {
+  it('fails on invalid request payload before bad auth.', async () => {
```
devinivy: Sounds like we're going to revert this change in behavior, which I support 👍

packages/xrpc-server/src/rate-limiter.ts
```diff
@@ -1,68 +1,126 @@
...
-  XRPCReqContext,
-} from './types'
-import { setHeaders } from './util'
-
-export type RateLimiterOpts = {
+
+// @NOTE Do not depend (directly or indirectly) on "./types" here, as it would
+// create a circular dependency.
+
+export interface RateLimierContext {
```
devinivy:
```suggestion
export interface RateLimiterContext {
```

packages/xrpc-server/src/rate-limiter.ts
```diff
@@ -241,35 +311,77 @@ export class RouteRateLimiter implements RateLimiterI {
...
-// when using a proxy, ensure headers are getting forwarded correctly: `app.set('trust proxy', true)`
-// https://expressjs.com/en/guide/behind-proxies.html
-const defaultKey: CalcKeyFn = (ctx: XRPCReqContext) => ctx.req.ip
+/**
+ * @note when using a proxy, ensure headers are getting forwarded correctly:
+ * `app.set('trust proxy', true)`
+ *
+ * @see {@link https://expressjs.com/en/guide/behind-proxies.html}
+ */
+const defaultKey: CalcKeyFn = ({ req }) => {
```
devinivy: The change to `defaultKey` is fairly significant, and I am not sure the new behavior is what we want to happen.  I'd propose we revert the change—perhaps we log a warning if `req.ip` is simply not present.

packages/xrpc-server/src/util.ts
```diff
@@ -81,93 +79,88 @@ export function decodeQueryParam(
   }
 }
 
-export function getQueryParams(url = ''): Record<string, string | string[]> {
-  const { searchParams } = new URL(url ?? '', 'http://x')
-  const result: Record<string, string | string[]> = {}
+export type QueryParams = Record<string, undefined | string | string[]>
+export function getQueryParams(url = ''): QueryParams {
+  const result: QueryParams = Object.create(null)
```
devinivy, line 85: What is the story with the changes to this helper?

packages/xrpc-server/src/util.ts
```diff
@@ -81,93 +79,88 @@ export function decodeQueryParam(
...
-      result[key] = result[key][0]
-    }
+    const values = searchParams.getAll(key)
+    result[key] = values.length === 1 ? values[0] : values
   }
   return result
 }
 
-export function validateInput(
+export function createInputVerifier(
```
devinivy, line 110: If I am understanding correctly there are no changes of behavior intended in here.  I am pretty sure that is the case, but could you help confirm?

I think we gained a lot of clarity in the implementation here overall 👍

packages/xrpc-server/src/util.ts
```diff
@@ -81,93 +79,88 @@ export function decodeQueryParam(
...
   nsid: string,
-  def: LexXrpcProcedure | LexXrpcQuery,
-  req: express.Request,
+  def: LexXrpcProcedure,
   opts: RouteOpts,
   lexicons: Lexicons,
-): HandlerInput | undefined {
-  // request expectation
+): (req: Request) => Input {
+  const { input } = def
```
devinivy: One case I think was being handled by the previous implementation but was dropped here is lexicons that erroneously contain `input` should not respect that setting.  For example, even if a `query` lexicon contains `input`, the request body should still not be present.

devinivy: Ah nevermind, I see now how this is handled 👍

devinivy (commented):
Just a few notes—pretty pumped about this!

packages/xrpc-server/src/rate-limiter.ts
matthieusieben: The hidden goal here is to reduce dependency on Express. I can either restore the dependency on `express.Request` and `express.Response` or throw an error in here if the user's IP cannot be determined. WDYT ?

packages/xrpc-server/src/util.ts
matthieusieben: This is a hot path and the change is mostly an optimization: Why instantiate a full `URL` when you only need a `URLSearchParams`, when there is an actual `?` in the `request.url`.

packages/xrpc-server/src/util.ts
matthieusieben: Yep, the goal was to keep the functionality exactly the same but "compile" the verifier instead of performing `ifs` with static conditions on every requests.

packages/xrpc-server/src/rate-limiter.ts
devinivy: When we're not using express there just isn't that much utility in the default—all the useful logic is really behind that `ip` getter and the `trust proxy` setting.  For many deployments using the socket address without checking x-forwarded-for or similar is likely to have unintended consequences, and hard to track down.

Up to you, though I lean away somewhat from throwing an error.  Logging a warning is also a valid option—or throwing in dev and warning in production.  If this is truly meant to be used outside express then we probably need the user to determine the default key behavior in non-express settings.

packages/xrpc-server/tests/auth.test.ts
matthieusieben: Reverted

.changeset/twenty-trainers-drive.md
matthieusieben: Reverted to exporting an `SkeletonHandler`

packages/xrpc-server/src/util.ts
```diff
@@ -91,21 +92,28 @@ export function getQueryParams(url = ''): QueryParams {
 
   const searchParams = new URLSearchParams(queryString)
   for (const key of searchParams.keys()) {
+    if (key === '__proto__') {
+      // Prevent prototype pollution
```
devinivy, line 96: This isn't a big or breaking change, but I think it's significant enough of a change in behavior to merit a changeset.

devinivy (approved):

PR #4000, opened 2025-06-26
Improve stash types for namespaces

ghost:


estrattonbailey (approved):
LGTM, talked about an object as const enum on Slack

PR #4009, opened 2025-07-01
bump PDS JSON request limit to 1MB

bnewbold:
We've generally discussed a 1MB limit for records. I didn't realize that we actually cap request sizes at 150KB in the PDS reference implementation.

This PR is a bit of a provocation: maybe increasing this limit would lead to abuse? It is about an order of magnitude larger limit, so it increases attack surface but not to a totally crazy degree. Potentially we should reduce the state best practice around record size to ~150KB, not increase this limit?

### DIFF
--- a/packages/pds/src/index.ts
+++ b/packages/pds/src/index.ts
@@ -62,7 +62,7 @@ export class PDS {
     const server = createServer({
       validateResponse: false,
       payload: {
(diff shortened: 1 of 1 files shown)


dholms:
Closing in favor of https://github.com/bluesky-social/atproto/pull/4989

PR #4010, opened 2025-07-01
Loosen unspecced constraint on saved feed IDs

mozzius:
`app.bsky.actor.defs#savedFeed` looks like this - `id` is just specified as a `string`

```json
"savedFeed": {
  "type": "object",
  "required": ["id", "type", "value", "pinned"],
  "properties": {
    "id": {
      "type": "string"
    },
    "type": {
      "type": "string",
      "knownValues": ["feed", "list", "timeline"]
    },
    "value": {
      "type": "string"
    },
    "pinned": {
      "type": "boolean"
    }
  }
```

Then, the helper methods for updating saved feeds preferences validate the items, which includes ensuring the `id` is a valid TID. Unfortunately, since this constraint is not encoded in the lexicon, other clients have used other values such as UUIDs.

Since we probably shouldn't tighten the lexicon retroactively and the bad data is already out there, I think we should just loosen the constraint. There's no reason this ID actually needs to be a TID.

Alternatively, our helper methods could detected invalid `id`s and replace them with TIDs. If that's preferred, I can do that instead.

### DIFF
--- a/.changeset/sixty-ducks-clean.md
+++ b/.changeset/sixty-ducks-clean.md
@@ -1,0 +1,5 @@
+---
+"@atproto/api": patch
+---
+
+loosen contraints for saved feed preferences
--- a/.changeset/sixty-ducks-clean.md
+++ b/.changeset/sixty-ducks-clean.md
@@ -2,4 +2,4 @@
 "@atproto/api": patch
 ---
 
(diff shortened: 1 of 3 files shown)


devinivy (approved):
Looks good, let's just add a changeset to the API package.

PR #4014, opened 2025-07-02
PDS: account deletion robustness

devinivy:
Removing a directory and its contents isn't an atomic operation, and can fail partially.  The actor store directory removal step of account deletion was able to fail with an `ENOTEMPTY`, `EBUSY`, or other generic filesystem error.  This adds support for retries using node's builtin functionality, requires a small patch to @atproto/common.

### DIFF
--- a/packages/common/src/fs.ts
+++ b/packages/common/src/fs.ts
@@ -1,4 +1,4 @@
-import { constants } from 'node:fs'
+import { constants, RmOptions } from 'node:fs'
 import fs from 'node:fs/promises'
 import { isErrnoException } from '@atproto/common-web'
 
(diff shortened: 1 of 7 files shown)


packages/pds/src/util/fs.ts
```diff
@@ -0,0 +1,7 @@
+import { RmOptions } from 'node:fs'
```
matthieusieben, line 1: Wouln't it be more consistent to use an import form `@atproto/common` here ? (meaning that we'd export `RmOptions` from `@atproto/common`)

matthieusieben (approved):
Looks good. Left small comment.

packages/common/src/fs.ts
```diff
@@ -29,10 +29,11 @@ export const readIfExists = async (
 
 export const rmIfExists = async (
   filepath: string,
-  recursive = false,
+  options: RmOptions | boolean = false,
 ): Promise<void> => {
   try {
-    await fs.rm(filepath, { recursive })
+    const opts = typeof options === 'boolean' ? { recursive: options } : options
+    await fs.rm(filepath, opts)
```
matthieusieben, line 36: As I am re-visinting this, I was wondering why we don't "just" do:

```suggestion
    await fs.rm(filepath, {
      recursive,
      force: true, // do not error if the file does not exist
      maxRetries: 5,
      retryDelay: 50,
    })
```

devinivy:
Replaced by #5080

PR #4027, opened 2025-07-09
Fix json and text uploads: don't parse bodies with input encoding of */*

### LINKED ISSUE #4026: Uploading plaintext to the PDS fails.
**Describe the bug**

When uploading a plaintext file to the PDS (i.e. a 1mb string) the upload fails, either with a "PayloadTooLargeError" or an "InternalServerError."
 
**To Reproduce**

Run the code below:

```typescript
import AtpAgent from "@atproto/api";

const agent = new AtpAgent({
  service: "https://bsky.social/",
});

await agent.login({
  identifier: "ID",
  password: "PASSWORD",
});

const longString = "a".repeat(1024 * 1024); // 1 MB of 'a's

const assetUploadResponse = await agent.uploadBlob(longString, {
  encoding: "text/plain"
});

console.log("Asset upload response:", assetUploadResponse);
```

**Expected behavior**

The file is uploaded and available as an `application/octet-stream`.

**Additional context**

This happens on both the selfhosted PDS and Bsky PDS instances. When encoding the file as utf-8 it uploads fine, somehow.

devinivy:
Uploads of json and text content were failing, as xrpc-server would consume the request body by attempting to parse them.  This step could also apply lower text/json-specific size limitations.  This has been addressed by skipping body parsing when the lexicon's input encoding is `*/*`, which indicates that the handler accepts any content-type and will determine how to process the body.

Resolves #4026

### DIFF
--- a/packages/xrpc-server/src/util.ts
+++ b/packages/xrpc-server/src/util.ts
@@ -129,22 +129,9 @@ export function createInputVerifier(
   // Lexicon definition expects a request body
 
   const { input } = def
-  const { blobLimit, jsonLimit, textLimit } = options
-  const jsonParser = json({ limit: jsonLimit })
-  const textParser = text({ limit: textLimit })
+  const { blobLimit } = options
 
-  // Transform json and text parser middlewares into a single function
-  const bodyParser = (req: Request, res: Response) => {
-    return new Promise<void>((resolve, reject) => {
-      jsonParser(req, res, (err) => {
-        if (err) return reject(XRPCError.fromError(err))
-        textParser(req, res, (err) => {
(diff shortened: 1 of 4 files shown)


packages/xrpc-server/src/util.ts
```diff
@@ -164,7 +164,9 @@ export function createInputVerifier(
       )
     }
 
-    await bodyParser(req, res)
+    if (input.encoding !== ENCODING_ANY) {
+      await bodyParser(req, res)
```
matthieusieben, line 155: We could improve memory usage by not creating a `bodyParser` at all!

matthieusieben (approved):
Awesome! Left a small nit but looks great!

PR #4029, opened 2025-07-10
Ozone support for age assurance compliance events

foysalit:
In the UK, the Online Safety Act requires that online platforms take specific steps to reduce the risk of children seeing harmful content. Since certain parts of that act come into effect later this month, we're making some changes for people in the UK.

See our blog post for more information: https://bsky.social/about/blog/07-10-2025-age-assurance

This PR adds visibility into age assurance info from users through a new event type in ozone and allows managing age assurance states for accounts.


packages/ozone/src/mod-service/index.ts
```diff
@@ -295,6 +299,11 @@ export class ModerationService {
         .where('modTool', 'is not', null)
         .where(sql`("modTool" ->> 'name')`, 'in', modTool)
     }
+    if (ageAssuranceState) {
+      builder = builder
+        .where('action', '=', 'tools.ozone.moderation.defs#ageAssuranceEvent')
+        .where(sql`meta->>'status'`, '=', ageAssuranceState)
```
devinivy: I notice `action` isn't usually specified in these queries.  Is it needed here because `meta->>'status'` is used for multiple action types?

packages/ozone/src/mod-service/status.ts
```diff
@@ -426,6 +435,30 @@ export const adjustModerationSubjectStatus = async (
...
+    }
+  }
+
+  if (action === 'tools.ozone.moderation.defs#ageAssuranceOverrideEvent') {
+    if (typeof meta?.status === 'string') {
+      newStatus.ageAssuranceState = meta.status
+      subjectStatus.ageAssuranceState = meta.status
+    }
+    newStatus.ageAssuranceUpdatedBy = 'admin'
+    subjectStatus.ageAssuranceUpdatedBy = 'admin'
```
devinivy: I see the logic is a little different than in `#ageAssuranceEvent` re: whether `ageAssuranceUpdatedBy` is set when `meta?.status` is/isn't a string.  Is that different in behavior intentional or can we sync the two up?

devinivy (approved):

packages/ozone/src/mod-service/index.ts
foysalit: yep! i dont think we use meta->>status for other action types right now but eventually, if we do, scoping it down to event type seems safer.

packages/ozone/src/mod-service/status.ts
foysalit: yeah makes sense to only update last updated by when status is set for admin events too.

PR #4038, opened 2025-07-12
[sync

knotbin:


knotbin:
@dholms could you please take a quick look at this if you have a moment? It's a blocker for me right now.

bnewbold:
I didn't initially catch that this is all client-side, not a change to the firehose protocol itself. We are working on other backfill/sync dev tooling, but it would still be nice to have this client package in a good place.

@dholms wdyt?

PR #4041, opened 2025-07-14
Endpoint for unregistering push notifications

ghost:
Supersedes and continues work from https://github.com/bluesky-social/atproto/pull/3903

This will be called when users sign off of the app.


mozzius (approved):
good stuff. lgtm

PR #4046, opened 2025-07-15
dev-env, identity, ozone: upgrade express 5

bjohansebas:
Following up on https://github.com/bluesky-social/atproto/pull/4045


bnewbold:
As a heads up, many of the folks who would review this are on vacation for the next couple weeks.

PR #4048, opened 2025-07-15
External ID in ozone events to help dedupe events

foysalit:


packages/ozone/src/api/moderation/emitEvent.ts
```diff
@@ -162,6 +163,20 @@ const handleModerationEvent = async ({
     await assertTagAuth(settingService, ctx.cfg.service.did, event, auth)
   }
 
+  if (externalId) {
+    const existingEvent = await moderationService.getEventByExternalId(
```
jcalabro: should this occur within the transaction?

lexicons/tools/ozone/moderation/emitEvent.json
```diff
@@ -72,6 +76,10 @@
       "errors": [
         {
           "name": "SubjectHasAction"
+        },
+        {
+          "name": "DuplicateExternalId",
```
jcalabro, line 81: I was thinking about this a bit more, in ordinary http I'd expect a http 409 (status conflict), but the [xrpc docs](https://atproto.com/specs/xrpc#summary-of-http-status-codes) don't list that as a respected status code 🤔 

this error condition generally is fine, just thinking out loud

jcalabro (approved):
couple small notes but otherwise 👍

packages/ozone/src/api/moderation/emitEvent.ts
foysalit: hmm.. yeah i guess it makes sense to move it inside the transaction since duplicate events will probably be fired very close to each other? i was hoping to avoid the transaction overhead for this check but good shout!

PR #4055, opened 2025-07-17
Fixes on handle suggestions

ghost:


estrattonbailey (approved):
Talked offline, we'll continue to refine before this goes live in the app 👍

PR #4060, opened 2025-07-19
Single-sign-on

avdb13:
Implementation of https://github.com/bluesky-social/atproto/discussions/2796.


packages/pds/src/api/com/atproto/sso/getCallback.ts
```diff
@@ -0,0 +1,561 @@
...
+      if (ctx.entrywayAgent) {
+        return resultPassthru(
+          await ctx.entrywayAgent.com.atproto.sso.getCallback(
+            params,
+            ctx.entrywayPassthruHeaders(req),
+          ),
+        )
+      }
+
+      const { code, state } = params;
```
ThisIsMissEm: OIDC can return back `error` and `error_description` parameters, and it currently looks like you don't handle any errors from the OIDC provider.

ghost: > ```json
> "parameters": {
>   "type": "union",
>   "refs": ["#successParams", "#errorParams"],
>   "closed": true
> }
> ```
> 
> … is invalid for a query because parameters is strictly expected to be a single params schema. That’s why all the codegen packages complain about refs and closed.
> 
> Implication: For OIDC callbacks, you need either:
>   1. A single params object containing all optional fields (success + error fields), with your handler distinguishing them at runtime.
>   2. Split into two separate endpoints, one for success, one for error.

I guess I will have to go for the following:

```json
"parameters": {
  "type": "params",
  "required": ["state"],
  "properties": {
    "code": { "type": "string" },
    "error": { "type": "string" },
    "error_description": { "type": "string" },
    "state": { "type": "string" }
  }
}
```

packages/pds/src/api/com/atproto/sso/getCallback.ts
```diff
@@ -0,0 +1,555 @@
...
+} from "@atproto/oauth-provider";
+import { DidDocument } from "@atproto/common";
+import { validateInputsForLocalPds } from "../server/createAccount";
+import { didDocForSession, safeResolveDidDoc } from "../server/util";
+import {
+  AccountStatus,
+  formatAccountStatus,
+} from "../../../../account-manager/account-manager";
+import { syncEvtDataFromCommit } from "../../../../sequencer";
+import { NullOutput, UserServiceAuthOutput } from "../../../../auth-verifier";
```
Minion3665: This doesn't seem to exist, causing my build to fail... (I can believe it works if you're ignoring types?)?

We've found the following to work better (*when references are also changed):


```suggestion
import { UserServiceAuthOutput, UnauthenticatedOutput } from "../../../../auth-output";
```

Minion3665 (commented):
A comment about a build failure here - as well as you seem not to have run `pnpm codegen` which causes further failures

I think your code probably(?) all works, but it definitely doesn't build in any mode where full type checking is happening. I'll comment more once I've got this up and running and tested this a bit...

packages/pds/src/api/com/atproto/sso/getCallback.ts
```diff
@@ -0,0 +1,555 @@
...
+
+          callbackId = cookies["atproto-callback"];
+        } catch (err) {
+          throw new InvalidRequestError(`Invalid cookie header: ${err}`);
+        }
+      }
+
+      if (!callbackId) {
+        throw new InvalidRequestError("Missing callback cookie");
+      }
```
ThisIsMissEm: You don't really need the cookie here, since you have that from the `state` parameter.

ghost: I rewrote the cookie logic to do encryption, so that the state inside the cookie can provide CSRF protection.

PR #4067, opened 2025-07-24
don't allow importRepo with a CAR intended for another identity

char:
patches up a footgun in the PDS from e.g. a bad migration tool that supplies the wrong carfile.
`verifyDiff` is only used by `importRepo` afaict so this shouldn't regress anything unrelated

ref: https://github.com/knotbin/airport/issues/6

### DIFF
--- a/packages/repo/src/sync/consumer.ts
+++ b/packages/repo/src/sync/consumer.ts
@@ -68,6 +68,11 @@ export const verifyDiff = async (
     did,
     signingKey,
   )
+  if (repo?.did && repo.did !== updated.did) {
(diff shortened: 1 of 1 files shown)


char:
I'm happy to have a "don't validate" flag, but for the case where you just want to migrate to a new DID I can imagine some tool that mutates your carfile to rewrite the `commit.did` and also additionally rewrite self-linking strongRefs (you can recompute the AT URIs and even recursively recompute the CIDs) - this car-prime would be accepted by even a strict-validating `importRepo` implementation.

PR #4069, opened 2025-07-25
Lexicon resolver package

devinivy:
This adds a new package @atproto/lexicon-resolver, which implements lexicon resolution as specified in [Lexicon Publication and Resolution](https://atproto.com/specs/lexicon#lexicon-publication-and-resolution).

A couple other additions came along the way:
 - Added an option to the CAR helpers in @atproto/repo for verifying CID -> content mappings within the CAR.
 - Made lexicon document validation permissive of an optional `$type` field for compatibility with published lexicon schema records.

### DIFF
--- a/.changeset/orange-phones-type.md
+++ b/.changeset/orange-phones-type.md
@@ -2,4 +2,4 @@
 "@atproto/repo": patch
 ---
 
-Add option to verify CIDs found withing CARs
+Add option to verify CIDs found withing CARs.
--- a/.changeset/orange-phones-type.md
+++ b/.changeset/orange-phones-type.md
@@ -1,0 +1,5 @@
+---
+"@atproto/repo": patch
+---
+
+Add option to verify CIDs found withing CARs
--- a/packages/lexicon-resolver/jest.config.js
+++ b/packages/lexicon-resolver/jest.config.js
@@ -1,0 +1,7 @@
+/** @type {import('jest').Config} */
+module.exports = {
+  displayName: 'Common',
+  transform: { '^.+\\.(t|j)s$': '@swc/jest' },
+  setupFiles: ['<rootDir>/../../jest.setup.ts'],
+  moduleNameMapper: { '^(\\.\\.?\\/.+)\\.js$': ['$1.ts', '$1.js'] },
+}
--- a/packages/lexicon-resolver/src/client/index.ts
+++ b/packages/lexicon-resolver/src/client/index.ts
@@ -1,0 +1,67 @@
+/**
+ * GENERATED CODE - DO NOT MODIFY
+ */
+import {
+  XrpcClient,
+  type FetchHandler,
+  type FetchHandlerOptions,
+} from '@atproto/xrpc'
+import { schemas } from './lexicons.js'
+import { CID } from 'multiformats/cid'
+import { type OmitKey, type Un$Typed } from './util.js'
+import * as ComAtprotoSyncGetRecord from './types/com/atproto/sync/getRecord.js'
+
+export * as ComAtprotoSyncGetRecord from './types/com/atproto/sync/getRecord.js'
+
+export class AtpBaseClient extends XrpcClient {
+  com: ComNS
+
+  constructor(options: FetchHandler | FetchHandlerOptions) {
+    super(options, schemas)
+    this.com = new ComNS(this)
+  }
+
+  /** @deprecated use `this` instead */
+  get xrpc(): XrpcClient {
+    return this
+  }
+}
+
+export class ComNS {
+  _client: XrpcClient
+  atproto: ComAtprotoNS
+
+  constructor(client: XrpcClient) {
+    this._client = client
+    this.atproto = new ComAtprotoNS(client)
+  }
+}
+
+export class ComAtprotoNS {
+  _client: XrpcClient
+  sync: ComAtprotoSyncNS
+
+  constructor(client: XrpcClient) {
+    this._client = client
+    this.sync = new ComAtprotoSyncNS(client)
+  }
+}
+
+export class ComAtprotoSyncNS {
+  _client: XrpcClient
+
+  constructor(client: XrpcClient) {
+    this._client = client
+  }
+
+  getRecord(
+    params?: ComAtprotoSyncGetRecord.QueryParams,
+    opts?: ComAtprotoSyncGetRecord.CallOptions,
+  ): Promise<ComAtprotoSyncGetRecord.Response> {
+    return this._client
+      .call('com.atproto.sync.getRecord', params, undefined, opts)
+      .catch((e) => {
+        throw ComAtprotoSyncGetRecord.toKnownErr(e)
+      })
+  }
+}
--- a/packages/lexicon-resolver/src/lexicon.ts
+++ b/packages/lexicon-resolver/src/lexicon.ts
@@ -10,24 +10,37 @@ const DNS_PREFIX = 'did='
 export const LEXICON_SCHEMA_NSID = 'com.atproto.lexicon.schema'
 
 export type ResolveLexiconOptions = ResolveRecordOptions & {
-  authorityDid?: string
+  didAuthority?: string
 }
 
 export type LexiconResolution = {
   commit: Commit
   uri: AtUri
   cid: CID
+  nsid: NSID
   lexicon: LexiconDoc & LexiconSchemaRecord
 }
 
+/**
+ * Resolve a lexicon from the network, verifying its authenticity.
+ * @param nsidStr NSID or string representing one for the Lexicon that will be resolved.
+ * @param options
+ * @returns
+ */
 export async function resolveLexicon(
   nsidStr: NSID | string,
   options: ResolveLexiconOptions = {},
 ): Promise<LexiconResolution> {
   const nsid = typeof nsidStr === 'string' ? NSID.parse(nsidStr) : nsidStr
-  const did = options.authorityDid ?? (await getLexiconAuthorityDid(nsid))
-  if (!did) {
-    throw new Error(`Could not resolve a DID authority for NSID: ${nsid}`)
+  let did: string | undefined
+  if (options.didAuthority) {
+    ensureValidDid(options.didAuthority)
+    did = options.didAuthority
+  } else {
+    did = await getLexiconDidAuthority(nsid)
+    if (!did) {
+      throw new Error(`Could not resolve a DID authority for NSID: ${nsid}`)
+    }
   }
   const verified = await resolveRecord(
     AtUri.make(did, LEXICON_SCHEMA_NSID, nsid.toString()),
@@ -37,15 +50,35 @@ export async function resolveLexicon(
   if (!isLexiconSchemaRecord(lexicon)) {
     throw new Error('Invalid lexicon schema record')
   }
+  if (lexicon.id !== nsid.toString()) {
+    throw new Error(
+      `Lexicon schema record id does not match NSID: ${lexicon.id}`,
+    )
+  }
   const { uri, cid, commit } = verified
-  return { commit, uri, cid, lexicon }
+  return { commit, uri, cid, nsid, lexicon }
 }
 
-export async function getLexiconAuthorityDid(nsidStr: NSID | string) {
+/**
+ *
+ * @param nsidStr NSID or string representing one for which to lookup its lexicon DID authority.
+ * @param options
+ * @returns
+ */
+export async function getLexiconDidAuthority(nsidStr: NSID | string) {
   const nsid = typeof nsidStr === 'string' ? NSID.parse(nsidStr) : nsidStr
   const did = await resolveDns(nsid.authority)
-  if (did != null) ensureValidDid(did)
+  if (did == null || !isValidDid(did)) return
   return did
+}
+
+function isValidDid(did: string) {
+  try {
+    ensureValidDid(did)
+    return true
+  } catch {
+    return false
+  }
 }
 
 async function resolveDns(authority: string): Promise<string | undefined> {
--- a/packages/lexicon-resolver/src/lexicon.ts
+++ b/packages/lexicon-resolver/src/lexicon.ts
@@ -1,0 +1,73 @@
+import dns from 'node:dns/promises'
+import { CID } from 'multiformats/cid'
+import { LexiconDoc, parseLexiconDoc } from '@atproto/lexicon'
+import { Commit } from '@atproto/repo'
+import { AtUri, NSID, ensureValidDid } from '@atproto/syntax'
+import { ResolveRecordOptions, resolveRecord } from './record.js'
+
+const DNS_SUBDOMAIN = '_lexicon'
+const DNS_PREFIX = 'did='
+export const LEXICON_SCHEMA_NSID = 'com.atproto.lexicon.schema'
+
+export type ResolveLexiconOptions = ResolveRecordOptions & {
+  authorityDid?: string
+}
+
+export type LexiconResolution = {
+  commit: Commit
+  uri: AtUri
+  cid: CID
+  lexicon: LexiconDoc & LexiconSchemaRecord
+}
+
+export async function resolveLexicon(
+  nsidStr: NSID | string,
+  options: ResolveLexiconOptions = {},
+): Promise<LexiconResolution> {
+  const nsid = typeof nsidStr === 'string' ? NSID.parse(nsidStr) : nsidStr
+  const did = options.authorityDid ?? (await getLexiconAuthorityDid(nsid))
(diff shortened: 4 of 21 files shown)


packages/lexicon-resolver/README.md
```diff
@@ -0,0 +1,79 @@
...
+})
+
+const resolved = await resolveLexicon('app.bsky.feed.post', {
+  idResolver,
+})
+```
+
+### With DID authority override
+
+You may specify a specific DID authority you'd like to use to perform a Lexicon resolution, overriding the ATProto's DNS-based authority over Lexicons. This is described in some more detail in [Authority and Control](https://atproto.com/specs/lexicon#authority-and-control).
```
devinivy:
```suggestion
You may specify a specific DID authority you'd like to use to perform a Lexicon resolution, overriding ATProto's DNS-based authority over Lexicons. This is described in some more detail in [Authority and Control](https://atproto.com/specs/lexicon#authority-and-control).
```

packages/lexicon-resolver/jest.config.js
```diff
@@ -0,0 +1,7 @@
+/** @type {import('jest').Config} */
+module.exports = {
+  displayName: 'Common',
```
devinivy:
```suggestion
  displayName: 'Lexicon Resolver',
```

.changeset/orange-phones-type.md
```diff
@@ -0,0 +1,5 @@
+---
+"@atproto/repo": patch
+---
+
+Add option to verify CIDs found withing CARs.
```
devinivy:
```suggestion
Add option to verify CIDs found within CARs.
```

packages/lexicon-resolver/src/lexicon.ts
```diff
@@ -0,0 +1,120 @@
...
+export type LexiconResolution = {
+  commit: Commit
+  uri: AtUri
+  cid: CID
+  nsid: NSID
+  lexicon: LexiconDoc & LexiconSchemaRecord
+}
+
+/**
+ * Resolve a lexicon from the network, verifying its authenticity.
```
devinivy:
```suggestion
 * Resolve a Lexicon from the network, verifying its authenticity.
```

packages/lexicon-resolver/src/lexicon.ts
```diff
@@ -0,0 +1,120 @@
...
+    throw new LexiconResolutionError(
+      `Lexicon schema record id does not match NSID: ${lexicon.id}`,
+    )
+  }
+  const { uri, cid, commit } = verified
+  return { commit, uri, cid, nsid, lexicon }
+}
+
+/**
+ *
```
devinivy:
```suggestion
 * Resolve the DID authority for a Lexicon from the network using DNS, based on its NSID.
```

packages/lexicon-resolver/src/lexicon.ts
```diff
@@ -0,0 +1,120 @@
...
+ * @param nsidStr NSID or string representing one for which to lookup its lexicon DID authority.
```
devinivy:
```suggestion
 * @param nsidStr NSID or string representing one for which to lookup its Lexicon DID authority.
```

packages/lexicon-resolver/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+### With DID authority override
+
+You may specify a specific DID authority you'd like to use to perform a Lexicon resolution, overriding ATProto's DNS-based authority over Lexicons. This is described in some more detail in [Authority and Control](https://atproto.com/specs/lexicon#authority-and-control).
+
+```ts
+import { resolveLexicon } from '@atproto/lexicon-resolution'
+
+const resolved = await resolveLexicon('app.bsky.feed.post', {
+  didAuthority: 'did:plc:...',
```
bnewbold, line 68: if this is the "resolution override" mechanism, should this be an array of DIDs, not just one?

packages/lexicon-resolver/README.md
```diff
@@ -0,0 +1,79 @@
...
+})
+
+const resolved = await resolveLexicon('app.bsky.feed.post', {
+  idResolver,
+})
+```
+
+### With DID authority override
+
+You may specify a specific DID authority you'd like to use to perform a Lexicon resolution, overriding ATProto's DNS-based authority over Lexicons. This is described in some more detail in [Authority and Control](https://atproto.com/specs/lexicon#authority-and-control).
```
bnewbold, line 62: the Auth Scopes proposal might be a better description than the current lexicon resolution spec: https://github.com/bluesky-social/proposals/tree/main/0011-auth-scopes#lexicon-resolution-overrides

packages/repo/src/car.ts
```diff
@@ -101,35 +108,40 @@ export const readCarStream = async (
...
-      // @NOTE the "finally" block of the async generator won't be called
-      // if the iteration was never started so we need to manually close here.
-      await reader.close()
-    }
+const readCarBlocksIter = (
+  reader: BufferedReader,
+  opts?: ReadCarOptions,
+): CarBlockIterable => {
+  let generator = readCarBlocksIterGenerator(reader)
+  if (opts?.verifyCids) {
```
matthieusieben: I would default this to `true`:

```suggestion
  if (opts?.verifyCids !== false) {
```

We could also name this `skipCidVerification` to make it more obvious that it needs to be set for cid verification to be disabled.

matthieusieben: This would also help with backwards compatibility: current dependents might implicitly assume that cids are being validated.

.changeset/orange-phones-type.md
```diff
@@ -0,0 +1,5 @@
+---
+"@atproto/repo": patch
+---
+
+Add option to verify CIDs found within CARs.
```
matthieusieben: Would it make sense to make this the default ?


> CIDs found within CARs are now being verified (unless explicilety disabled).

packages/lexicon-resolver/README.md
devinivy: Actually this isn't quite the override mechanism, currently it's just a way to specify the authority you want to pull the lexicon from.  I think it's nice to have for completeness, and we'll also use it for the recursive lexicon resolution to use cached authority lookups.

packages/lexicon-resolver/README.md
devinivy: This doesn't implement the lexicon override functionality, it just sets the authority to use instead of the DNS mapping.  For example, there is no fallback mechanism from the DID authority specified to the DNS-based authority.  However, one might use this functionality to implement the full override mechanism themselves.

.changeset/orange-phones-type.md
DavidBuchanan314: This is the right thing to do but I wonder if there's code that accidentally relies on the old (non-verifying) behaviour - something to watch on deployment maybe.

devinivy: That is why I originally opted to go with the non-verifying version by default—on balance it's actually not obvious which is "worse" because there are legitimate use cases where you don't want to be so strict with CAR contents.  Performance regressions can also be a real, practical issue and we're at risk of that here too.

After getting Matthieu's feedback I did a more thorough overview of all of our usage, went through our dependents according to npm, and I think that there is a low enough chance of an issue that we can roll this out.  My main concern is with `importRepo` performance which is already an issue for some repos, but I do think we want to ensure we only import clean repos.

If we want to be more conservative but verify by default, we could also just call this a breaking change of the @atproto/repo package, I don't think that would be a major deal.

packages/lexicon-resolver/src/client/index.ts
```diff
@@ -0,0 +1,67 @@
+/**
```
matthieusieben, line 1: You should also add these files to `.gitattributes` so they get ignored in Guthub diffs.

packages/lexicon-resolver/src/lexicon.ts
```diff
@@ -0,0 +1,120 @@
...
+/**
+ * Resolve the DID authority for a Lexicon from the network using DNS, based on its NSID.
+ * @param nsidStr NSID or string representing one for which to lookup its Lexicon DID authority.
+ * @param options
+ * @returns
+ */
+export async function resolveLexiconDidAuthority(nsidStr: NSID | string) {
+  const nsid = typeof nsidStr === 'string' ? NSID.parse(nsidStr) : nsidStr
+  const did = await resolveDns(nsid.authority)
+  if (did == null || !isValidDid(did)) return
```
matthieusieben, line 100: nit: I think it's safer to explicit the return type of publicly exposed function to avoid any unexpected break in the future.

packages/lexicon-resolver/src/record.ts
```diff
@@ -0,0 +1,125 @@
...
+    .resolveAtprotoData(did, forceRefresh)
+    .catch((err) => {
+      throw new RecordResolutionError('Could not resolve DID identity data', {
+        cause: err,
+      })
+    })
+  const client = new Client(
+    typeof rpc === 'function'
+      ? rpc
+      : { service: identity.pds, fetch: safeFetch, ...rpc },
```
matthieusieben: I don't really like the spreading pattern when we want to apply defaults. It can be error prone:

If `rpc = { fetch: undefined }`, we won't get any typing error (because `fetch` is optional in `Partial<BuildFetchHandlerOptions> `) but we will end up *not* using safe fetch, as the `fetch: undefined` will end up overriding the `fetch: safeFetch` default.

I usually prefer enumerating properties (as this will allow better v8 optimizations), or, if we really want to spread:

```ts
{
  ...rpc,
  fetch: rpc.fetch ?? safeFetch,
  service: rpc.service ?? identity.pds,
}
```

or even

```ts
const { fetch = safeFetch, service = identity.pds, ...rest } = rpc
return { fetch, service, ...rest } // Safe now
```

matthieusieben:
I think that we might be better served by exposing an `interface` (and at lease one `class` implementation), as I believe it provides a better DX.

What I mean by that is that the piece of code that wants to do the resolution will generally not care about the global "options" (like the `didAuthority`, or the fact that, under the hood, a lexhub type of service is being used):

```ts
// somewhere in oauth-provider

const result = await this.lexiconResolver.resolveLexicon(nsid)
```

There are two kinds of options:
- Those that matter at the time of individual calls (like `forceRefresh`)
- Those that are configuration

The `interface` pattern neatly allows to wrap the configuration options inside a class while only exposing to the relevant interface to the users of that class.

It also allows to neatly combine / wrap / and use alternative implementation completely transparently.


```ts
interface LexiconResolver {
  resolve(nsid: string): Promise<ResolvedLexicon>
}

class LexiconResolverBase implements LexiconResolver {
  constructor(private options: { fetch, dnsLookup }) {}
  resolve(nsid: string) { /* dns + getRecord */ }
}

class LexiconResolverCache implements LexiconResolver {
  constructor(private resolver: LexiconResolver) {}
  resolve(nsid: string) { /* cache or this.resolver */ }
}

class LexiconResolverOverride implements LexiconResolver {
  constructor(didAuthority: string, private resolver: LexiconResolver) {
    this.overrideResolver = new LexiconResolverBase({ dnsLookup: (nsid) => didAuthority })
  }
  resolve(nsid: string) {
    // check in didAuthority's repo first
    // If not found, fallback to this.resolver.resolve()
  }
}

class LexiconResolverLexhub implements LexiconResolver {
  constructor(private serviceUrl: URL) {}
  resolve(nsid: string) {
    // TBD
  }
}

const myResolver = new LexiconResolverCache(
  new LexiconResolverOverride(
    'did:web:foo.com',
    new LexiconResolverLexhub('https://lexhub.com')
  )
)
```

My take is that, at some point, we *will* be able to compose resolver logic in that way, but we won't want to update the code of the users of that logic because the options might have changed. WDYT?

devinivy:
@matthieusieben I was able to address all your feedback 👍 

Want to explain a little bit about the direction I went with the interfaces.

First, I totally agree that it's inconvenient to pass `rpc` and `idResolver` on individual calls to `resolveLexicon()`, and I wanted to fix that here.

I offered a `LexiconResolver` interface as you proposed, and I am interested in what level of reuse we'll end up getting from it—the contents of the return value `LexiconResolution` is quite low-level to the spec.  That could mean a. it may not be what will get cached for many use cases, b. it's not compatible with totally different forms of resolution, such as looking up lexicons stored on disk.  But this intended to be is a low-level reference implementation, and I am inclined to keep the return value as it is (we may later even want to include the whole proof in the output, as certain tooling might want to use it for integrity checks).

I also wanted to include `didAuthority` on individual calls—this doesn't make sense in the generic interface, but it does make sense here because you may use it to implement something lower-level such as authority pinning / trust-on-first-use, which could vary per lexicon.

I tried to meet in the middle by offering two interfaces: `LexiconResolver` and `AtprotoLexiconResolver`.  The former is just `(nsid) => Promise<LexiconResolution>` and the latter is identical but supports an `options` arg with settings that only make sense in the context of the spec, but not generically.  Since `AtprotoLexiconResolver` satisfies `LexiconResolver`, we still get that generic interface you were looking for.

One of the main reasons I reached for a builder function rather than a class was a naming issue.  Since this is a reference implementation and "Lexicon resolution" is something very specific in the context of the protocol, I didn't want to clutter the name of the implementation.  So if the interface were `LexiconResolver` I didn't want the class to be named `LexiconResolverBase` or something like that.  And I couldn't think of a better name for the interface—e.g. I didn't want to reach for `ILexiconResolver`.  I also don't mind that I was able to export a single function as a "default lexicon resolver" since I feel this is pretty friendly to just import and start using without any setup.

It was hard to decide on these compromises but overall I am pretty happy with how it turned out, and I believe/hope I was able to stick to the spirit of your feedback above!

packages/lexicon-resolver/src/record.ts
```diff
@@ -0,0 +1,150 @@
...
+      })
+    const client = new Client(
+      typeof rpc === 'function'
+        ? rpc
+        : {
+            ...rpc,
+            service: rpc?.service ?? identity.pds,
+            fetch: rpc?.fetch ?? safeFetch,
+          },
+    )
```
matthieusieben, line 77: It's too bad we end up creating one `Client` instance on every call. `XrpcClient` was built as being "single service". We should maybe investigate how we could improve the way `XrpcClient` works with multiple services (as a separate line of work obviously)

packages/lexicon-resolver/src/record.ts
```diff
@@ -0,0 +1,150 @@
...
+        did,
+        collection: uri.collection,
+        rkey: uri.rkey,
+      })
+      .catch((err) => {
+        throw new RecordResolutionError('Could not fetch record proof', {
+          cause: err,
+        })
+      })
+    const verified = await verifyRecordProof(proofBytes, {
```
matthieusieben, line 89: Looks like some code path in `buildRecordResolver` could possibly throw errors other than `RecordResolutionError` (e.g. if `readCarWithRoot` throws). Is that acceptable (because of the low probability?), or do we want this to always throw `RecordResolutionError` ?

I've been using the following pattern in oauth stuff, to complement the pattern you already adopted (`.catch` block that wraps specific errors):

```ts
function fetchRecord(atUri: string) {
  try {
    // You're already doing this:
    await foo().catch(cause => {
      throw new RecordResolutionError('Failed to foo', { cause })
    })
    
    // ...
  } catch (err) {
    // This ensures that only RecordResolutionError are ever thrown, improving debug traces.
    throw RecordResolutionError.from(err, `Unable to fetch record at ${uriStr}`)
  }
}

class RecordResolutionError {
  // ...

  static from(err: unknown, message = "Failed to resolve record"): RecordResolutionError {
    if (err instanceof RecordResolutionError) return err
    return new RecordResolutionError(messgage, { cause: err })
  }
}

```

packages/lexicon-resolver/src/record.ts
devinivy: We have the same thing some other places too where we call out to folks pds's dynamically.  It would be nice, though I don't think it's going to cause any issues here.

packages/lexicon-resolver/src/record.ts
devinivy: The way I see it there's not a goal to _always_ throw a `RecordResolutionError` necessarily, but I would like to map all the meaningful parts of record resolution into these errors.  Especially for unexpected/programmer errors or precondition/postcondition failures, I like when those bubble up as-is since they require a different kind of handling versus. operational errors.  I think for now I will just do a final pass and try to cover any remaining gaps.

devinivy: Another way I think of it is that `RecordResolutionError` are errors semantic to record resolution, versus "any error that happens within the record resolution function".  Hopefully those are mostly the same, but they may not be a perfect match.

PR #4072, opened 2025-07-28
Cleanup app.bsky.unspecced.checkHandleAvailability

ghost:
The implementation needs to live elsewhere, not in appview, otherwise we can't check reserved handles.

Commits:
1. Removes the old `app.bsky.unspecced.checkHandleAvailability` (lex + implementation).
2. Adds `com.atproto.temp.checkHandleAvailability` (lex only).

### DIFF
--- a/.changeset/thin-panthers-sparkle.md
+++ b/.changeset/thin-panthers-sparkle.md
@@ -1,5 +1,5 @@
 ---
-"@atproto/api": patch
+"@atproto/api": minor
 ---
 
 Remove app.bsky.unspecced.checkHandleAvailability, add com.atproto.temp.checkHandleAvailability


.changeset/thin-panthers-sparkle.md
```diff
@@ -0,0 +1,5 @@
+---
+"@atproto/api": patch
```
matthieusieben: If this was previoulsy released, it should be a `minor` patch.

matthieusieben (approved):
The mechanism is not implemented on PDS's which will prevent the APP from using suggestions when creating account on self hosted PDS.

This is fine if the APP does not crash when the server answers with a `MethodNotImplemented` error, and/or completely avoids calling this endpoint if the service URL is not Bluesky's entryway.

.changeset/thin-panthers-sparkle.md
ghost: Fixed in 9a917520e

PR #4073, opened 2025-07-29
Small xrpc-server optimizations

matthieusieben:
This commit is part of #3806 and was extracted for clarity.

The main purpose was to expose `parseReqEncoding`. While doing that, I realized there was an optimization opportunity in the was the encoding is checked both for input and output:
- `validateOutput` performs several times the same check on `output`, which is now covered by a single `if`.
- `createInputVerifier` was parsing the definition's allowed encoding on every request. That value is now parsed a single time (at startup), and the check is only performed if needed.

### DIFF
--- a/packages/xrpc-server/src/util.ts
+++ b/packages/xrpc-server/src/util.ts
@@ -5,6 +5,7 @@ import { Request, Response, json, text } from 'express'
 import { contentType } from 'mime-types'
 import { MaxSizeChecker, createDecoders } from '@atproto/common'
 import {
+  LexXrpcBody,
   LexXrpcProcedure,
   LexXrpcQuery,
   LexXrpcSubscription,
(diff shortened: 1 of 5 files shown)


packages/xrpc-server/src/util.ts
```diff
@@ -180,64 +181,73 @@ export function validateOutput(
...
-      `A response body is expected but none was provided`,
-    )
-  }
+    // Fool-proofing (should not be necessary due to type system)
+    const result = handlerSuccess.safeParse(output)
+    if (!result.success) {
+      throw new InternalServerError(`Invalid handler output`, undefined, {
+        cause: result.error,
+      })
+    }
```
matthieusieben, line 199: Replaces previous `handlerSuccess.parse(output)`

packages/xrpc-server/src/util.ts
```diff
@@ -180,64 +182,81 @@ export function validateOutput(
   output: HandlerSuccess | void,
   lexicons: Lexicons,
 ): void {
-  // initial validation
-  if (output) {
-    handlerSuccess.parse(output)
-  }
+  if (def.output) {
```
ghost, line 185: Nit: you could check `!def.output` first and then early return or throw depending on `output` being present.

Then the bulk of the function can have 1 fewer indentation level.

ghost, line 185: You could also consider (not here, can be in a separate PR) validating the output as we validate the inputs, where we create a validator function. So you might perform some work upfront only once, not on every request.

There isn't _that_ much work to be done upfront, but at least parsing the encodings from the definition could be done on startup.

matthieusieben: I considered it but realized that output validation is only enabled in dev / test

PR #4076, opened 2025-07-30
Hook up bsync service in dev-env

ghost:
Before this PR:
1. We didn't run the `bsync` service in `dev-env` (for dev and tests).
2. We had a mocked `bsync` implementation that `bsky` ran in `dev-env`. It was somewhat correct, but it did the operations on the dataplane synchronously, which was quite misleading and caused a few wrong usages.

After this PR:
1. We removed the mocked `bsync`.
2. `bsync` is now run as part of `dev-env` as an independent service, using the actual implementation.
3. `bsky` talks to the actual `bsync` implementation via gRPC (as it did with the mocked `bsync`).
4. `bsky` implemented as part of its TS dataplane a `BsyncSubscription`, which consumes from the `bsync` scan endpoints and writes to the TS dataplane.

This flow is now much more aligned with the production flow.

### DIFF
--- a/packages/bsky/src/data-plane/server/bsync-subscription.ts
+++ b/packages/bsky/src/data-plane/server/bsync-subscription.ts
@@ -1,4 +1,5 @@
 import assert from 'node:assert'
+import { EventEmitter, once } from 'node:events'
 import { jsonStringToLex } from '@atproto/lexicon'
 import { AtUri } from '@atproto/syntax'
 import { BsyncClient, authWithApiKey, createBsyncClient } from '../../bsync'
@@ -7,21 +8,26 @@ import { ids } from '../../lexicon/lexicons'
 import { SubjectActivitySubscription } from '../../lexicon/types/app/bsky/notification/defs'
 import { AgeAssuranceEvent } from '../../lexicon/types/app/bsky/unspecced/defs'
 import { subLogger as log } from '../../logger'
-import { Method, MuteOperation_Type, Operation } from '../../proto/bsync_pb'
+import {
+  Method,
+  MuteOperation,
+  MuteOperation_Type,
+  NotifOperation,
+  Operation,
+} from '../../proto/bsync_pb'
 import { Namespaces } from '../../stash'
-import { BackgroundQueue } from './background'
 import { Database } from './db'
 import { excluded } from './db/util'
 
-export class BsyncSubscription {
+export class BsyncSubscription extends EventEmitter {
   private ac: AbortController | undefined
-  private background: BackgroundQueue
+  private pendingScans = new Set<Promise<void>>()
   private bsyncClient: BsyncClient
   private db: Database
 
   constructor(public opts: { db: Database; config: ServerConfig }) {
+    super()
     const { config, db } = opts
-    this.background = new BackgroundQueue(db)
 
     this.db = db
     this.bsyncClient = createBsyncClient({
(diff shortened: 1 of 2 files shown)


packages/bsky/src/api/app/bsky/graph/muteActor.ts
```diff
@@ -9,6 +10,7 @@ export default function (server: Server, ctx: AppContext) {
     handler: async ({ auth, input }) => {
       const { actor } = input.body
       const requester = auth.credentials.iss
+      assert(requester !== actor, 'cannot mute yourself')
```
ghost: This has to be done sync in the endpoint, we were relying on the mocked `bsync` for doing it.

packages/bsky/src/data-plane/server/bsync-subscription.ts
```diff
@@ -0,0 +1,336 @@
...
+      })
+  }
+
+  private scanMuteOperations() {
+    this.scanBsyncEndpoint((cursor?: string) => {
+      return this.bsyncClient
+        .scanMuteOperations({ limit: 100, cursor }, { signal: this.ac.signal })
+        .then(async (res) => {
+          this.background.add(async () => {
+            for await (const op of res.operations) {
```
ghost: The logic for inserting operations was copied from `packages/bsky/src/data-plane/bsync/index.ts`, only the logic of consuming the `bsync` scan endpoints was added here.

So basically everything inside each of the 3 `this.background.add()` calls was copied.

packages/dev-infra/docker-compose.yaml
```diff
@@ -1,4 +1,3 @@
-version: '3.8'
```
ghost, line 1: This is not directly related to this PR, but it causes an annoying warning. We should remove this everywhere.

packages/bsky/src/api/app/bsky/graph/muteActor.ts
matthieusieben: This should throw an `InvalidRequestError`, not a 500.

packages/bsync/src/index.ts
```diff
@@ -39,6 +39,8 @@ export class BsyncService {
     overrides?: Partial<AppContextOptions>,
   ): Promise<BsyncService> {
     const ac = new AbortController()
+    // Prevents unhelpful warnings.
+    setMaxListeners(Infinity, ac.signal)
```
matthieusieben: If you are not able to find an upper bound, it means that you have a memory leak. Listeners should always be removed when no longer needed.

packages/bsky/src/data-plane/server/bsync-subscription.ts
```diff
@@ -0,0 +1,336 @@
...
+      post,
+      reply,
+    })
+    .execute()
+}
+
+const handleAgeAssuranceEventOperation = async (
+  db: Database,
+  op: Operation,
+  _now: string,
```
matthieusieben: Is there a particular reason we keep this if its unused?

packages/bsky/src/data-plane/server/bsync-subscription.ts
```diff
@@ -0,0 +1,336 @@
...
+  }
+
+  async destroy() {
+    if (this.destroyed) return
+    this.destroyed = true
+    this.ac.abort()
+    await this.processAll()
+  }
+
+  private scanBsyncEndpoint(
```
matthieusieben: I would call this `startScaning`

packages/bsky/src/data-plane/server/bsync-subscription.ts
```diff
@@ -0,0 +1,336 @@
...
+    callFn: (cursor?: string) => Promise<string>,
+    cursor?: string,
+  ) {
+    if (this.destroyed) return
+
+    callFn(cursor)
+      .then((nextCursor) => {
+        this.scanBsyncEndpoint(callFn, nextCursor)
```
matthieusieben: Should we add some delay here ?

packages/bsky/src/data-plane/server/bsync-subscription.ts
```diff
@@ -0,0 +1,336 @@
...
+import { SubjectActivitySubscription } from '../../lexicon/types/app/bsky/notification/defs'
+import { AgeAssuranceEvent } from '../../lexicon/types/app/bsky/unspecced/defs'
+import { subLogger as log } from '../../logger'
+import { Method, MuteOperation_Type, Operation } from '../../proto/bsync_pb'
+import { Namespaces } from '../../stash'
+import { BackgroundQueue } from './background'
+import { Database } from './db'
+import { excluded } from './db/util'
+
+export class BsyncSubscription {
```
matthieusieben: There is a concurrency issue with the current implementation.

`restart()` can cause multiple versions of the same `scanMuteOperations`, `scanNotifOperations` and `scanOperations` processes to run in parallel. This is due to the fact that `this.background.add()` is called as nested operation. Because of that, it is possible that `scanMuteOperations` is doing something, but did not use `background.add()` yet. If `destroy()` gets called at that moment, it will not be waiting for that task to complete (since the task will not have been added to the background queue yet) before calling `start()`. `start()` will start a new process in parallel of the one still running (and that did not complete yet). The still running process will then complete and check for `this.destroyed` to determine if it should continue, which will be the case, causing both processes to be running.

In addition to the concurrency issue, the `AbortController` is not re-created after a `restart`.

The simple solution is to remove the `restart()` functionality if we don't need it.

Otherwise, I recommend doing these:
1) Improve the state tracking (deriving `stopped`/`running`/`stopping` from `this.ac` in a getter instead of using redundent `destroyed: boolean`)
2) Binding every running process to the current `AbortSignal` **instance** instead of using `this.ac` (that could have been instantiated for another run).
3) Avoid using `BackgroundQueue` as it's purpose does not properly match this class's behavior.

packages/bsky/src/data-plane/server/bsync-subscription.ts
```diff
@@ -0,0 +1,336 @@
...
+      .catch((err) => {
+        if (this.destroyed) return
+        log.error({ err }, 'error in bsync notif scan')
+      })
+  }
+
+  private scanMuteOperations() {
+    this.scanBsyncEndpoint((cursor?: string) => {
+      return this.bsyncClient
+        .scanMuteOperations({ limit: 100, cursor }, { signal: this.ac.signal })
```
matthieusieben: If `destroy()` is called while `bsyncClient.scanMuteOperations()` is still running, `processAll()` (in destroy) will *not* be waiting for the background task, since it will only be added *after* (in the `then()` bellow). This can cause concurrency issues, in particular in case of a restart.

packages/bsky/src/data-plane/server/bsync-subscription.ts
ghost: This is a long-polling, so it waits for 10s while polling if it doesn't have results.

So if there's never a result, it will call it once every 10s, which to me feels reasonable. Adding a delay here will just make it slower in cases where it might have data to fetch.

packages/bsky/src/api/app/bsky/graph/muteActor.ts
ghost: bf5112225

packages/bsky/src/data-plane/server/bsync-subscription.ts
ghost: No, I removed in bf5112225

packages/bsync/src/index.ts
ghost: I didn't need it to be Infinity, I've set it to 100 as we have in other places. bf5112225

packages/bsky/src/data-plane/server/bsync-subscription.ts
ghost: bf5112225

packages/bsky/src/data-plane/server/bsync-subscription.ts
ghost: I did some improvements in 50e1afd60.

I wanted to stop using the background queue, but I wasn't sure how to implement `processAll` without it. I'd like some help there if you can.

packages/bsky/src/data-plane/server/bsync-subscription.ts
ghost: I did some improvements in 50e1afd60, what do you think?

matthieusieben:
There are at least two things that still need to be done for proper concurrency control:

1) you should **not** read the signal from `this.ac` after the task has already started. This is because you might end up reading a signal from another controller, causing the first task not to stop.
2) `background.add` should be called directly, and not in a callback or `processAll` might not await on that particular task.

the following code should allow you to do that (and even drops the use of a background task completely):

```ts
  private scanningTasks: Promise[] = []
  
  public async processAll() {
    await Priomise.all(this.scanningTasks)
  }

  private startScanning(
    callFn: (cursor: undefined | string, signal: AbortSignal) => Promise<string | undefined>
  ) {
    // @NOTE important: read this.ac only once, and at the very begining,
    // to avoid reading from another AbortController instance if the current
    // one gets aborted and another one gets created during the processing

    const { signal } = this.ac!
    signal.throwIfAborted() // fool-proofing

    const taskPromise = (async () => {
      try {
        let cursor

        while (!signal.aborted) {
          cursor = await callFn(cursor, signal)
        }
      } catch (err) {
        if (!signal.aborted) {
          log.error({ err }, 'error in bsync scan')
        }
      }
    })

    this.scanningTasks.add(taskPromise)
    void taskPromise.finally(() => {
      this.scanningTasks.remove(taskPromise)
    })
  }

```

Note that if you adapt your code using this, `scanMuteOperations` (& co.) should use the `signal` provided in the callback rather than `this.ac`. Tasks using `this.background.add` can also be inlined.

PR #4083, opened 2025-08-05
Enforce max size on importRepo

dholms:


packages/pds/src/api/com/atproto/repo/importRepo.ts
```diff
@@ -20,13 +20,22 @@ export default function (server: Server, ctx: AppContext) {
...
+      if (
+        ctx.cfg.service.maxImportSize &&
+        size > ctx.cfg.service.maxImportSize
+      ) {
+        throw new InvalidRequestError('Import size exceeds maximum allowed')
+      }
+
       await ctx.actorStore.transact(did, (store) =>
-        importRepo(store, input.body),
+        importRepo(store, input.body, size),
```
matthieusieben: Is there a particular reason not to use the max here ? If a request is using chuncked encoding, the `size` would be `0` (`Number('')`), disabling the check.

```suggestion
        importRepo(store, input.body, ctx.cfg.service.maxImportSize),
```

packages/pds/src/api/com/atproto/repo/importRepo.ts
```diff
@@ -20,13 +20,22 @@ export default function (server: Server, ctx: AppContext) {
...
         throw new InvalidRequestError('Service is not accepting repo imports')
       }
+      const contentLength = Number(req.header('content-length'))
+      const size = isNaN(contentLength) ? 0 : contentLength
+      if (
+        ctx.cfg.service.maxImportSize &&
+        size > ctx.cfg.service.maxImportSize
+      ) {
+        throw new InvalidRequestError('Import size exceeds maximum allowed')
+      }
```
matthieusieben: Note that `Number('')` is `0`. Which means that a client "streaming" (chunk encoding) would not be properly limited as the code would check against `0`.

```suggestion
      // Fail early if request is not using chuncked encoding
      const contentLength = req.header('content-length')
      if (contentLength != null && ctx.cfg.service.maxImportSize) {
        const size = Number(contentLength || 'NaN')
        if (
          Number.isNaN(size) ||
          size > ctx.cfg.service.maxImportSize
        ) {
          throw new InvalidRequestError('Import size exceeds maximum allowed')
        }
      }
      
      // Otherwise fail while reading stream through importRepo's maxSize arg
```

matthieusieben (changes requested):
Nice feature to have! It looks like the current logic could be abused to bypass the check though (by using chencked encoding)

PR #4086, opened 2025-08-07
Make codegen locale-independent (maybe?)

DavidBuchanan314:
Codegen inadvertently relied on shell globbing order, which is locale-dependent. People on different systems were getting different import orders etc (see https://github.com/bluesky-social/atproto/pull/4085).

This PR standardises on `LC_COLLATE=C`.

Should get a few people on different systems to test it before merging, not 100% sure this is a valid fix.

I considered fixing by adding a `paths.sort()` at https://github.com/bluesky-social/atproto/blob/9a1746a025f82239b891ff6938d2db1a5ae3e708/packages/lex-cli/src/util.ts#L8, however this ends up being a different sort - and while it should be consistent it'd end up being a big diff.

### DIFF
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -17,7 +17,7 @@
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
   "scripts": {
-    "codegen": "node ./scripts/generate-code.mjs && lex gen-api --yes ./src/client ../../lexicons/com/atproto/*/* ../../lexicons/app/bsky/*/* ../../lexicons/chat/bsky/*/* ../../lexicons/tools/ozone/*/*",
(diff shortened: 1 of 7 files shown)


packages/api/package.json
```diff
@@ -17,7 +17,7 @@
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
   "scripts": {
-    "codegen": "node ./scripts/generate-code.mjs && lex gen-api --yes ./src/client ../../lexicons/com/atproto/*/* ../../lexicons/app/bsky/*/* ../../lexicons/chat/bsky/*/* ../../lexicons/tools/ozone/*/*",
+    "codegen": "node ./scripts/generate-code.mjs && LC_COLLATE=C sh -c 'lex gen-api --yes ./src/client ../../lexicons/com/atproto/*/* ../../lexicons/app/bsky/*/* ../../lexicons/chat/bsky/*/* ../../lexicons/tools/ozone/*/*'",
     "build": "tsc --build tsconfig.build.json",
```
DavidBuchanan314, line 21: ^ the actual change

DavidBuchanan314:
Apparently this didn't work, presumably different shells still have different sort rules...

Second attempt: https://github.com/bluesky-social/atproto/pull/4087

PR #4090, opened 2025-08-08
Incremental Lexicon Codegen (prototype)

DavidBuchanan314:
`pnpm codegen` takes ~15 seconds now (down from multiple minutes).

To enable this:
- I refactored the generated `index.ts` file into separate files per-namespace, with all the original exports propagated back up to the main `index.ts` so as not to break the API.
- Implemented incremental rebuilds (via basic timestamp checking logic)

This is just a prototype, and I hacked up the lex-gen code a lot to make it work - it needs some refactoring, especially related to passing around state associated with the change-tracking/timestamp-comparison logic.

During lexicon codegen, the majority of CPU time was being spent in side `ts-morph`, the library used to manipulate the TS AST. `ts-morph` is great but it's not very fast, and there isn't much we can do to speed it up. So with incremental rebuilds, we limit the amount of AST manipulation we actually need to do at once.

Since we commit generated code to git, almost all builds can be incremental builds.

Known issues:
- If you delete a lexicon schema, the generated code will persist until you manually delete it or do a non-incremental build (`pnpm codegen --rebuild`) - to fix this, a smarter approach to tracking changes is needed.

### DIFF
--- a/packages/lex-cli/src/codegen/client.ts
+++ b/packages/lex-cli/src/codegen/client.ts
@@ -260,7 +260,7 @@ const nsIndexTs = (
         ])
 
       // necessary imports only
-      let importNsids = [
+      const importNsids = [
         ...new Set(
           ns.userTypes.map((ut) => ut.nsid).concat(Object.values(ATP_METHODS)),
         ),
@@ -277,7 +277,7 @@ const nsIndexTs = (
       }
 
       // recurse
-      for (let child of genNamespaceCls(file, ns)) {
+      for (const child of genNamespaceCls(file, ns)) {
         api.files.push(
           await nsIndexTs(
             api,
@@ -296,7 +296,7 @@ const nsIndexTs = (
   )
 
 function genNamespaceCls(file: SourceFile, ns: DefTreeNode): DefTreeNode[] {
-  let children: DefTreeNode[] = []
+  const children: DefTreeNode[] = []
 
   //= export class {ns}NS {...}
   const cls = file.addClass({
--- a/packages/lex-cli/src/index.ts
+++ b/packages/lex-cli/src/index.ts
@@ -10,9 +10,9 @@ import {
   applyFileDiff,
   genFileDiff,
   genTsObj,
+  getTSTimestamps,
   printFileDiff,
   readAllLexicons,
-  getTSTimestamps,
 } from './util'
 
 const program = new Command()
--- a/packages/lex-cli/src/util.ts
+++ b/packages/lex-cli/src/util.ts
@@ -45,7 +45,7 @@ function walk(dir: string): Promise<[string, fs.Stats][]> {
       }
       Promise.all(
         files.map((file) => {
-          return new Promise<[string, fs.Stats][]>(async (resolve, reject) => {
+          return new Promise<[string, fs.Stats][]>((resolve, reject) => {
             const filepath = join(dir, file)
             fs.stat(filepath, async (error, stats) => {
               if (error) {
--- a/packages/lex-cli/src/codegen/common.ts
+++ b/packages/lex-cli/src/codegen/common.ts
@@ -1,9 +1,9 @@
+import { relative } from 'node:path'
 import { Options as PrettierOptions, format } from 'prettier'
 import { Project, SourceFile, VariableDeclarationKind } from 'ts-morph'
 import { type LexiconDoc } from '@atproto/lexicon'
-import { ModificationTimes, type GeneratedFile } from '../types'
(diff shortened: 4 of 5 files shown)


package.json
```diff
@@ -19,7 +19,7 @@
     "verify:lint": "pnpm lint",
     "verify:types": "tsc --build tsconfig.json",
     "format": "pnpm lint:fix && pnpm style:fix",
-    "precodegen": "pnpm run --recursive --stream --filter '@atproto/lex-cli...' build --force",
+    "precodegen": "pnpm run --recursive --stream --filter '@atproto/lex-cli...' build",
```
DavidBuchanan314, line 22: I'm not sure how necessary this was - removing it saves about 10 seconds.

matthieusieben: It's useful when working on `lex-cli` to make sure the dependencies are re-built. In my experience, it sometimes happens that `tsc` will not re-build a file unless `--force` is added (though I've never been able to determine why, maybe git branch swaps ?).

It can be disabled here as long as devs working on `lex-cli` are aware that they may need to run `p run precodegen --force && p codegen` before they issue their last commit.

packages/lex-cli/src/codegen/client.ts
```diff
@@ -137,6 +162,27 @@ const indexTs = (
         name: ns.propName,
         type: ns.className,
       })
+
+      file
+        .addImportDeclaration({
+          moduleSpecifier: `./ns/${ns.propName}/index.js`,
```
matthieusieben, line 168: Does it make sense to use distinct file path for `client` and `server`, instead of `/ns/` for both ? I can see a future in which an app would generate both client and server definitions.

packages/lex-cli/src/types.ts
```diff
@@ -1,14 +1,16 @@
...
 }
 
 export interface FileDiff {
-  act: 'add' | 'mod' | 'del'
+  act: 'add' | 'mod' | 'del' | 'leave'
   path: string
   content?: string
 }
+
+export type ModificationTimes = Record<string, number>
```
matthieusieben, line 16: I feel like a `Map<string, number>` are more accurate when working with key-value store. They are also "safer" than plain objects used as key value store (although, in this instance, file names are arguably unlikely to be named `toString` or `__proto__`).

the typing also better reflect missing entries: `map.get('bar')` is typed `V | undefined` whereas `({} as Record<string, number>).bar` will incorrectly by typed as as `number` (actual value is `undefined`)

packages/lex-cli/src/util.ts
```diff
@@ -1,23 +1,95 @@
...
+  paths = [...paths].sort() // incoming path order may have come from locale-dependent shell globs
+  const lastModified: ModificationTimes = {}
   const docs: LexiconDoc[] = []
   for (const path of paths) {
     if (!path.endsWith('.json') || !fs.statSync(path).isFile()) {
       continue
     }
     try {
-      docs.push(readLexicon(path))
+      const ts = fs.statSync(path).mtime.getTime()
```
matthieusieben, line 23: It would be nice to be able to trigger a "force" re-build, in which case we don't need the `mtime` at all.

packages/lex-cli/src/util.ts
```diff
@@ -1,23 +1,95 @@
...
+                folderContents: [string, fs.Stats][],
+              ) => all.concat(folderContents),
+              [],
+            ),
+          )
+        })
+        .catch(reject)
+    })
+  })
+}
```
matthieusieben, line 84: you should avoid creating your own promises when possible. Node exposes a promisified version of the `fs` module that should allow for simpler code.

```ts
import * as fsPromises from "node:fs/promises"
```

packages/lex-cli/src/index.ts
```diff
@@ -10,6 +10,7 @@ import {
   applyFileDiff,
   genFileDiff,
   genTsObj,
+  getTSTimestamps,
```
matthieusieben, line 13: For consistency (with `genTsObj`), would this better be named `getTsTimestamps` ?

package.json
DavidBuchanan314: Hm that makes sense, it could plausibly be an issue with the incremental codegen in this pr too, after a branch swap or merge or whatever it is that causes it. Maybe a strategic git hook could `touch` relevant source files to force full rebuilds after certain actions.

matthieusieben: I actually remember what causes this: if a dependency is changed, the dependents don't get rebuilt as the typescript compiler won't check for timestamp of files not in `src`.

This is only an issue if the changes in the dependencies influence the build of the codegen, which they usually shouldn't. So your change here should be fine.

PR #4101, opened 2025-08-13
Remove requirement for JWK to define either `use` or `key_ops`

matthieusieben:


packages/oauth/oauth-provider/src/client/client-manager.ts
```diff
@@ -75,6 +80,13 @@ export class ClientManager {
       return jwks
     }, clientJwksCache)
 
+    const fetchMetadataHandler = pipe(
```
matthieusieben, line 83: Move here for consistency

DavidBuchanan314 (approved):
LGTM - I assume we'll do some dev comms at some point in the future before we flip the config into not-legacy mode

(Edit: for people in the future, this comment refers to an earlier approach and is no longer relevant!)

packages/oauth/jwk/src/jwk.ts
```diff
@@ -122,19 +122,20 @@ export const jwkSchema = z.union([
...
-  )
+export const jwkValidator = jwkSchema.refine(
+  (k) =>
+    !k.use ||
+    !k.key_ops ||
+    k.key_ops.every(
+      k.use === 'sig'
+        ? (o) => o === 'sign' || o === 'verify'
+        : (o) => o === 'encrypt' || o === 'decrypt',
+    ),
```
DavidBuchanan314: With the caveat that I'm unfamiliar with zod, this looks like it does what we want.

Although, `use` is only for public keys - is it "consistent" to allow the `key_ops` values of "sign" and "decrypt"? (since a public key can only verify or encrypt)

("consistent" as required at the end of https://datatracker.ietf.org/doc/html/rfc7517#section-4.3 )

DavidBuchanan314: > The "use" parameter is employed to indicate whether a public key is used for encrypting data or verifying the signature on data.

packages/oauth/jwk/src/jwk.ts
```diff
@@ -122,12 +122,13 @@ export const jwkSchema = z
...
+      // https://datatracker.ietf.org/doc/html/rfc7517#section-4.3
+      // > The "use" parameter is employed to indicate whether a public key is
+      // > used for encrypting data or verifying the signature on data.
       !k.use ||
       !k.key_ops ||
       k.key_ops.every(
-        k.use === 'sig'
-          ? (o) => o === 'sign' || o === 'verify'
-          : (o) => o === 'encrypt' || o === 'decrypt',
+        k.use === 'sig' ? (o) => o === 'verify' : (o) => o === 'decrypt',
```
DavidBuchanan314, line 131: looking good :+1:

PR #4103, opened 2025-08-13
Allow `"use"` claims only in public jwk

matthieusieben:
This pr addresses invalid assumptions that was made when the `jwk` tools were created:

- `use` and `key_ops` SHOULD NOT (per spec) be used together
- `use` MUST NOT be used for private keys

### DIFF
--- a/packages/oauth/jwk-jose/src/jose-key.ts
+++ b/packages/oauth/jwk-jose/src/jose-key.ts
@@ -23,6 +23,7 @@ import {
   SignedJwt,
   VerifyOptions,
   VerifyResult,
+  isPrivateJwk,
   jwkSchema,
   jwtHeaderSchema,
   jwtPayloadSchema,
@@ -163,12 +164,12 @@ export class JoseKey<J extends Jwk = Jwk> extends Key<J> {
     allowedAlgos: string[] = ['ES256'],
     kid?: string,
     options?: Omit<GenerateKeyPairOptions, 'extractable'>,
-  ) {
+  ): Promise<JoseKey> {
     const kp = await this.generateKeyPair(allowedAlgos, {
       ...options,
       extractable: true,
     })
-    return this.fromImportable(kp.privateKey, kid)
+    return this.fromKeyLike(kp.privateKey, kid)
   }
 
   static async fromImportable(
@@ -239,8 +240,16 @@ export class JoseKey<J extends Jwk = Jwk> extends Key<J> {
     if (!jwk || typeof jwk !== 'object') throw new JwkError('Invalid JWK')
 
     const kid = either(jwk.kid, inputKid)
-    const use = jwk.use || 'sig'
 
-    return new JoseKey(jwkSchema.parse({ ...jwk, kid, use }))
+    // Backwards compatibility with old behavior
+    if (jwk.use != null && isPrivateJwk(jwk)) {
+      console.warn(
+        'Deprecation warning: Private JWK with a "use" property will be rejected in the future. Please remove replace "use" with (valid) "key_ops".',
+      )
+      jwk.key_ops ??= jwk.use === 'sig' ? ['sign'] : ['encrypt']
+      delete jwk.use
+    }
+
+    return new JoseKey<Jwk>(jwkSchema.parse({ ...jwk, kid }))
   }
 }
--- a/packages/oauth/jwk/src/jwk.ts
+++ b/packages/oauth/jwk/src/jwk.ts
@@ -1,45 +1,96 @@
 import { z } from 'zod'
+import { isLastOccurrence } from './util'
 
-export const keyUsageSchema = z.enum([
+export const PUBLIC_KEY_USAGE = ['verify', 'encrypt', 'wrapKey'] as const
+export const publicKeyUsageSchema = z.enum(PUBLIC_KEY_USAGE)
+export type PublicKeyUsage = (typeof PUBLIC_KEY_USAGE)[number]
+export function isPublicKeyUsage(usage: unknown): usage is PublicKeyUsage {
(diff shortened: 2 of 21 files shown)


packages/oauth/jwk/src/key.ts
```diff
@@ -84,6 +111,54 @@ export abstract class Key<J extends Jwk = Jwk> {
...
+    if (opts.usage) {
+      if (this.ops) {
+        if (opts.usage === 'verify' && this.ops.includes('sign')) {
+          // allowed
+        } else if (opts.usage === 'decrypt' && this.ops.includes('encrypt')) {
+          // allowed
+        } else if (!this.ops.includes(opts.usage)) {
+          return false // no match
+        }
+      }
```
matthieusieben: When ops is defined, only matching ops should be allowed. The provided key should contain both `verify` and `sign` if that's their intended use.

```suggestion
      if (this.ops && !this.ops.includes(opts.usage)) {
        return false // no match
      }
```

packages/oauth/jwk/src/key.ts
```diff
@@ -84,6 +111,48 @@ export abstract class Key<J extends Jwk = Jwk> {
...
+        if (opts.usage === 'verify' && this.use === 'sig') {
+          // allowed
+        } else if (opts.usage === 'encrypt' && this.use === 'enc') {
+          // allowed
+        } else {
+          return false // no match
+        }
+      }
+
+      if (opts.usage === 'sign' || opts.usage === 'encrypt') {
```
matthieusieben:
```suggestion
      if (opts.usage === 'sign' || opts.usage === 'decrypt') {
```

packages/oauth/jwk/src/key.ts
```diff
@@ -1,32 +1,38 @@
...
 import { cachedGetter } from './util.js'
 
 const jwkSchemaReadonly = jwkSchema.readonly()
 
-export abstract class Key<J extends Jwk = Jwk> {
-  constructor(protected readonly jwk: Readonly<J>) {
-    // @TODO "use" is actually only for public keys. We should allow missing
-    // "use" here and automatically add it to the exposed `publicJwk`
+function isPublicKeyUsage(usage: KeyUsage): boolean {
+  return usage === 'verify' || usage === 'decrypt'
```
matthieusieben:
```suggestion
  return usage === 'verify' || usage === 'encrypt'
```

packages/oauth/jwk/src/key.ts
```diff
@@ -39,10 +45,27 @@ export abstract class Key<J extends Jwk = Jwk> {
...
+      ? Array.from(
+          new Set(
+            this.jwk.key_ops
+              .map((op) =>
+                op === 'sign' ? 'verify' : op === 'encrypt' ? 'decrypt' : op,
+              )
+              .filter(isPublicKeyUsage),
+          ),
+        )
+      : undefined
```
matthieusieben:
```suggestion
    const publicOps = this.jwk.key_ops?.filter(isPublicKeyUsage)
```

packages/oauth/jwk-jose/src/jose-key.ts
```diff
@@ -168,7 +168,7 @@ export class JoseKey<J extends Jwk = Jwk> extends Key<J> {
       ...options,
       extractable: true,
     })
-    return this.fromImportable(kp.privateKey, kid)
+    return this.fromKeyLike(kp.privateKey, kid)
```
matthieusieben, line 172: This is just a "shortcut" (`fromImportable` would end up calling `fromKeyLike`)

packages/oauth/jwk/src/jwk.ts
```diff
@@ -98,54 +148,110 @@ export const jwkOkpKeySchema = jwkBaseSchema.extend({
...
   alg: z.enum(['HS256', 'HS384', 'HS512']).optional(),
 
   k: z.string(), // Key Value (base64url encoded)
 })
 
-export const jwkUnknownKeySchema = jwkBaseSchema.extend({
-  kty: z
-    .string()
-    .refine((v) => v !== 'RSA' && v !== 'EC' && v !== 'OKP' && v !== 'oct'),
-})
```
matthieusieben, line 112: There is no point in supporting keys of unknown `kty` without supporting any additional property those might contain.

This _could_ be implemented by adding a `.passthrough()`, however this is not ideal as it causes typing errors downstream.

Since this lib is pretty exhaustive in terms of recognized `kty` (even if only a small subset is actually used but Atproto code), it's just simpler to get rid of this.

This is technically a breaking change, though in practice it isn't really since only shady behaviors would have actually relied on this. For this reason, not marking a minor version bump.

packages/oauth/jwk/src/jwk.ts
```diff
@@ -98,54 +148,110 @@ export const jwkOkpKeySchema = jwkBaseSchema.extend({
...
-      !k.use ||
-      !k.key_ops ||
-      k.key_ops.every(
-        k.use === 'sig' ? (o) => o === 'verify' : (o) => o === 'decrypt',
-      ),
+    (k): boolean =>
+      k.use == null ||
+      k.key_ops == null ||
+      (k.use === 'sig' && k.key_ops.every(isSigKeyUsage)) ||
+      (k.use === 'enc' && k.key_ops.every(isEncKeyUsage)),
```
DavidBuchanan314: https://datatracker.ietf.org/doc/html/rfc7517#section-4.3

> Other values [of `key_ops`] MAY be used.

Perhaps we're being overly strict here? I think we should disallow known-incompatible values, but allow unrecognized `key_ops`.

DavidBuchanan314: Although I suppose if a key had an unrecognized `key_ops` we wouldn't know what to do with it anyway

PR #4104, opened 2025-08-13
Fix putRecord auth check

DavidBuchanan314:
Related thread: https://bsky.app/profile/bnewbold.net/post/3lwcopaeubs2y

The check in `putRecord` assumed `repo` is a DID, but it may also be a handle. This PR uses the same logic as before (before OAuth scopes were introduced), and also mirrors the checks in `createRecord`.

Note: It builds, but not yet tested!

### DIFF
--- a/.changeset/nine-plants-peel.md
+++ b/.changeset/nine-plants-peel.md
@@ -1,0 +1,5 @@
+---
+"@atproto/pds": minor
+---
+
+Fix putRecord auth check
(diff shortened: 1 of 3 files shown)


.changeset/nine-plants-peel.md
```diff
@@ -0,0 +1,5 @@
+---
+"@atproto/pds": minor
```
devinivy:
```suggestion
"@atproto/pds": patch
```

PR #4107, opened 2025-08-14
Perf: Avoid fetching account data twice in `putRecord`

matthieusieben:


DavidBuchanan314 (approved):
Looks good. I added a test case earlier for putRecord with repo==handle in https://github.com/bluesky-social/atproto/pull/4104/commits/ffa3e12a54b0fd89ea51257db0fb21145175178b , we should perhaps test create/delete/applyWrites also

packages/pds/src/api/com/atproto/repo/applyWrites.ts
```diff
@@ -84,6 +87,16 @@ export default function (server: Server, ctx: AppContext) {
         }
       }
 
+      const account = await ctx.authVerifier.findAccount(repo, {
+        checkDeactivated: true,
+        checkTakedown: true,
+      })
```
devinivy: Moving this code down here now performs the authorization checks after status verification, where these usually occur in the opposite order.  I think for consistency we should probably switch it back around.

devinivy: Just looking ahead, I believe there's a similar case with the other endpoints that were updated.  For example, if I am takendown I should fail to authenticate and my write permissions should not be checked.

devinivy (approved):
Tiny piece of feedback, but generally looks good!

PR #4108, opened 2025-08-14
Msi/permission set

matthieusieben:


.changeset/loud-fishes-cover.md
```diff
@@ -0,0 +1,5 @@
+---
+"@atproto-labs/simple-store": minor
+---
+
+Allow `Key` type to be anything (except nullable values)
```
devinivy, line 5: I will read ahead, but could you summarize how this capability gets used?  It is certainly more complicated for implementers to support any key type, and in the past I've almost been more inclined to reign it back to just `string` instead of `string | number`.

.changeset/cool-books-roll.md
```diff
@@ -0,0 +1,5 @@
+---
+"@atproto/lexicon": minor
+---
+
+Add support for `permission-set` in lexicon documents
```
devinivy, line 5: What made this minor vs. patch (pre-v1)?  Or nothing in particular, just opting for minor as a marker?

packages/dev-env/src/mock/index.ts
```diff
@@ -542,21 +541,23 @@ export async function generateMockSetup(env: TestNetwork) {
...
-  await seedThreadV2.annotateOP(sc)
-  await seedThreadV2.sort(sc)
-  await seedThreadV2.bumpOpAndViewer(sc)
-  await seedThreadV2.bumpGroupSorting(sc)
-  await seedThreadV2.bumpFollows(sc)
-  await seedThreadV2.blockDeletionAuth(sc, env.bsky.ctx.cfg.modServiceDid)
-  await seedThreadV2.mutes(sc)
-  await seedThreadV2.threadgated(sc)
-  await seedThreadV2.tags(sc)
+  // @TODO the following should be optimized as it makes dev-env start very slow (>10 sec)
```
devinivy, line 545: Is this a TODO that needs to be resolved in this PR?  Also what is the plan—perhaps add a flag or turn it into a seed?

packages/pds/bin/migration-create.ts
```diff
@@ -1,38 +0,0 @@
-#!/usr/bin/env ts-node
```
devinivy, line 1: Ah, had this become outdated?

packages/syntax/src/nsid.ts
```diff
@@ -53,48 +48,102 @@ export class NSID {
...
+
+    // Hot path: check frequent chars first
+    if (
+      (charCode >= 97 && charCode <= 122) /* a-z */ ||
+      (charCode >= 65 && charCode <= 90) /* A-Z */
+    ) {
+      // All good
+    } else if (charCode >= 48 && charCode <= 57 /* 0-9 */) {
+      if (i === 0) {
+        return 'NSID first part may not start with a digit'
```
devinivy: Ah, I actually find myself missing the previous implementation quite a bit!  While less performant, it was just very linear, concise, and self-documenting.  Is there something in the rest of the work that you found necessitated the change?

devinivy: (For what it's worth, I think moving away from throwing and instead returning the error strings to better enable `isValidNsid()` is nice 👍 )

packages/pds/src/config/env.ts
```diff
@@ -143,6 +143,9 @@ export const readEnv = (): ServerEnvironment => {
     // fetch
     fetchMaxResponseSize: envInt('PDS_FETCH_MAX_RESPONSE_SIZE'),
 
+    // lexicon resolver
+    lexiconDidAuthority: envStr('PDS_LEXICON_RESOLVER_DID_AUTHORITY'),
```
devinivy: It's possible that this functionality may not need to be configurable via env var—perhaps just part of the config for local testing?

packages/pds/src/context.ts
```diff
@@ -292,22 +294,60 @@ export class AppContext {
...
-        responseMaxSize: cfg.fetch.maxResponseSize,
-        ssrfProtection: !cfg.fetch.disableSsrfProtection,
-      }),
-      logRequest: ({ method, url }) => {
-        fetchLogger.debug({ method, uri: url }, 'fetch')
-      },
-      logResponse: false,
-      logError: false,
+    //
+    // @NOTE **DO NO** wrap this fetch with any logging or other transforms as
```
devinivy:
```suggestion
    // @NOTE **DO NOT** wrap this fetch with any logging or other transforms as
```

devinivy: Is this whole comment relevant to the pds's usage specifically, a general issue with `allowImplicitRedirect: false`, or something else?  I am having a little trouble understanding the cause and impact.

packages/pds/src/account-manager/helpers/lexicon.ts
```diff
@@ -0,0 +1,51 @@
...
+    lastSucceededAt: toDateISO(lastSucceededAt),
+    uri,
+    lexicon: toJson(lexicon),
+  }
+
+  await db.db
+    .insertInto('lexicon')
+    .values({ nsid, ...updates })
+    .onConflict((oc) => oc.column('nsid').doUpdateSet(updates))
+    .execute()
```
devinivy: As I recall we generally want to execute writes with sqlite retry logic, i.e. with `db.executeWithRetry()`.  It's been a while since I've had to muck around in here, but does that sound right/familiar?

packages/internal/fetch-node/src/unicast.ts
```diff
@@ -13,32 +13,55 @@ import { isUnicastIp } from './util.js'
...
+   * On older NodeJS version, the `dispatcher` init option is ignored when
+   * creating a new Request instance. It can only be passed through the fetch
+   * function directly.
+   *
+   * Since this is a security feature, we need to ensure that the unicastLookup
+   * function is called to resolve the hostname to a unicast IP address.
+   *
+   * However, in the case a custom "fetch" function is passed here (fetch !==
+   * globalThis.fetch), we have no guarantee that the dispatcher will be used to
+   * make the request. Because of this, in such a case, we will use a on-time
```
devinivy:
```suggestion
   * make the request. Because of this, in such a case, we will use a one-time
```

packages/internal/simple-store/src/simple-store.ts
```diff
@@ -1,15 +1,15 @@
-export type Awaitable<V> = V | PromiseLike<V>
+import { Awaitable } from './util.js'
 
-export type Key = string | number
+export type Key = NonNullable<unknown>
 export type Value = NonNullable<unknown> | null
 
-export type GetOptions = { signal?: AbortSignal }
+export type GetStoredOptions = { signal?: AbortSignal }
```
devinivy: Is this intended to be `GetStoreOptions` instead of `GetStoredOptions`?

packages/internal/simple-store/src/simple-store.ts
```diff
@@ -1,15 +1,15 @@
-export type Awaitable<V> = V | PromiseLike<V>
+import { Awaitable } from './util.js'
 
-export type Key = string | number
+export type Key = NonNullable<unknown>
```
devinivy, line 3: I had a question about this change above in the changesets, just noting again here for good measure.

packages/lexicon/src/types.ts
```diff
@@ -1,7 +1,15 @@
 import { z } from 'zod'
-import { NSID } from '@atproto/syntax'
+import { isValidNsid } from '@atproto/syntax'
 import { requiredPropertiesRefinement } from './util'
 
+export const localeSchema = z
+  .string()
+  .regex(/^[a-z]{2,3}(-[A-Z]{2})?$/, 'Invalid locale')
```
devinivy: There is a `validateLanguage()` in `@atproto/common` and `@atproto/common-web` which validates well-formed BCP47.

packages/oauth/oauth-provider-ui/src/components/utils/scope-description.tsx
```diff
@@ -87,6 +101,89 @@ export function ScopeDescription({
...
+  includeScopes: IncludeScope[]
+  permissionSets: PermissionSets
+}) {
+  if (!includeScopes.length) return null
+
+  return (
+    <>
+      {includeScopes.map((includeScope, i) => (
+        <IncludeScopePermissions
+          key={i}
```
devinivy, line 117: Numeric indexes aren't usually preferred for `key` due to the way it preserves component state when elements get reordered—is it our best option here?

packages/oauth/oauth-provider-ui/src/components/utils/scope-description.tsx
```diff
@@ -762,3 +808,121 @@ function scopeEnablesPrivateBskyAppMethods(scope: string): boolean {
...
+    <Nsid {...attrs} nsid={lxm} />
+  )
+}
+
+type AudProps = Override<
+  Omit<HTMLAttributes<HTMLDivElement>, 'children' | 'title'>,
+  { aud: AudParam }
+>
+function Aud({ aud, ...attrs }: AudProps) {
+  if (aud.startsWith('did:web:bsky.app#')) {
```
devinivy:
```suggestion
  if (aud.startsWith('did:web:api.bsky.app#')) {
```

packages/oauth/oauth-provider-ui/src/components/utils/scope-description.tsx
```diff
@@ -762,3 +808,121 @@ function scopeEnablesPrivateBskyAppMethods(scope: string): boolean {
...
+    return (
+      <ItemDescription {...attrs} title={aud}>
+        <Trans>Bluesky App servers</Trans>
+      </ItemDescription>
+    )
+  }
+  if (aud.startsWith('did:web:bsky.chat#')) {
```
devinivy:
```suggestion
  if (aud.startsWith('did:web:api.bsky.chat#')) {
```

packages/oauth/oauth-provider-ui/src/components/utils/scope-description.tsx
```diff
@@ -762,3 +808,121 @@ function scopeEnablesPrivateBskyAppMethods(scope: string): boolean {
...
+
+type AudProps = Override<
+  Omit<HTMLAttributes<HTMLDivElement>, 'children' | 'title'>,
+  { aud: AudParam }
+>
+function Aud({ aud, ...attrs }: AudProps) {
+  if (aud.startsWith('did:web:bsky.app#')) {
+    return (
+      <ItemDescription {...attrs} title={aud}>
+        <Trans>Bluesky App servers</Trans>
```
devinivy:
```suggestion
        <Trans>Bluesky App services</Trans>
```

packages/oauth/oauth-provider-ui/src/components/utils/scope-description.tsx
```diff
@@ -762,3 +808,121 @@ function scopeEnablesPrivateBskyAppMethods(scope: string): boolean {
...
+      </ItemDescription>
+    )
+  }
+  if (aud.startsWith('did:web:bsky.chat#')) {
+    return (
+      <ItemDescription {...attrs} title={aud}>
+        <Trans>Bluesky Chat servers</Trans>
```
devinivy:
```suggestion
        <Trans>Bluesky Chat services</Trans>
```

packages/oauth/oauth-provider/src/lexicon/lexicon-manager.ts
```diff
@@ -0,0 +1,106 @@
...
+
+function parseScope(scope?: string) {
+  const otherScopes: string[] = []
+  const includeScopes: IncludeScope[] = []
+
+  if (scope) {
+    for (const scopeValue of scope.split(' ')) {
+      const parsed = IncludeScope.fromString(scopeValue)
+      if (parsed) includeScopes.push(parsed)
+      else otherScopes.push(scopeValue)
```
devinivy:
```suggestion
      if (parsed) {
        includeScopes.push(parsed)
      } else {
        otherScopes.push(scopeValue)
      }
```

packages/oauth/oauth-provider/src/lexicon/lexicon-resolution-error.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export class LexiconResolutionError extends Error {
+  constructor(
+    public readonly nsid: string,
+    options?: ErrorOptions,
+  ) {
+    const cause = options?.cause
+    // Make sure to always return the most informative message possible,
+    // including the nsid, which @atproto/lexicon-resolver doesn't consistently
+    // do, so that the client devs can properly debug issues.
```
devinivy: Reaching into `cause.message` to patch over this feels a little funny—if it's important to devex but a small change could we update the error messages in lexicon-resolver instead?

packages/oauth/oauth-provider/src/lib/util/locale.ts
```diff
@@ -5,14 +5,8 @@ export const localeSchema = z
   .regex(/^[a-z]{2,3}(-[A-Z]{2})?$/, 'Invalid locale')
```
devinivy, line 5: There is a helper for validating well-formed BCP47 language tags in @atproto/common and @atproto/common-web.

packages/oauth/oauth-scopes/src/lib/nsid.ts
```diff
@@ -1,6 +1,103 @@
...
-  !value.endsWith('.')
+export type Nsid = `${string}.${string}.${string}`
+export const isNsid = (v: unknown): v is Nsid =>
+  typeof v === 'string' && isValidNsid(v)
+
+// @NOTE the code below is a copy-paste from @atproto/syntax. The reason for
+// this is that this package (as well as @atproto/syntax) gets compiled as an
+// commonjs module, preventing tree-shaking of un-necessary @atproto/syntax code
+// when bundled by vite. This can be replaced with an import once all packages
+// use esm modules.
```
devinivy: Ah, bummer!  Should we leave a note in @atproto/syntax so that if there are changes there they are brought into here as well?

packages/oauth/oauth-scopes/src/scopes/blob-permission.ts
```diff
@@ -0,0 +1,106 @@
+import { Matchable } from '../lib/matchable.js'
```
devinivy: Any idea why these didn't appear as moved files?  Would appreciate any interesting call outs, since these permission classes largely look unchanged.

packages/oauth/oauth-scopes/src/scopes/include-scope.ts
```diff
@@ -0,0 +1,147 @@
...
+function isUnderAuthority(this: IncludeScope, itemNsid: '*' | Nsid) {
+  if (itemNsid === '*') return false
+
+  const authorityNamespace = extractNsidNamespace(this.nsid)
+  const itemNamespace = extractNsidNamespace(itemNsid)
+  return (
+    itemNamespace === authorityNamespace ||
+    itemNamespace.startsWith(`${authorityNamespace}.`)
+  )
+}
```
devinivy: @bnewbold tagging just to double-check you're on the same page about this logic for determining which `lxm` and `collection` used within a permission set, based on their nsids.

packages/oauth/oauth-scopes/src/scopes/include-scope.ts
```diff
@@ -0,0 +1,147 @@
+import { AtprotoAudience, isAtprotoAudience } from '@atproto/did'
```
devinivy, line 1: It would be nice to get some tests on this, esp. for limits on `lxm` and `collection`.

devinivy (commented):
Did a pass of review on this, looking pretty darn close 👍

.changeset/loud-fishes-cover.md
matthieusieben: This was initially done to support Keying using `NSID` (which are objects) as keys.

Implementers can *still* define how much they restrict the value.

Any place accepting a simple store can constrain it like so:

```ts
function foo(store: SimpleStore<NSID, LexiconDoc>) {}
```

Any place implementing the interface can still constrain further the types they support:

```ts
// Redis only supports string as key and values
class RedisStore<K extends string, V extends string> implements SimpleStore<K, V> {
}
```

It can be used as a mechanism to convert to simpler types:

```ts
// convert a SimpleStore<string, string> (e.g. RedisStore) into a SimpleStore<NSID, LexiconDoc>
class StringBasedLexiconStore implements SimpleStore<NSID, LexiconDoc> {
  constructor(
    private internal: SimpleStore<string, string>,
  ) {}

  async get(nsid: NSID, options) {
    const res = await this.internal.get(nsid.toString(), options)
    if (res === undefined) return undefined
    return JSON.parse(res) as LexiconDoc
  }

  async set(nsid: NSID, lexiconDoc: LexiconDoc) {
    await this.internal.set(nsid.toString(), JSON.stringify(lexiconDoc))
  }

  async del(nsid: NSID) {
    await this.internal.del(nsid.toString())
  }
}
```

It was useful to be able to do:

```ts
class LexiconStore implements SimpleStore<NSID, LexiconDoc> {
  // ...
}

const store = new LexiconStore()
const getter = new CachedGetter<NSID, LexiconDoc>(lexiconResolver, store)

const nsid = NSID.parse('foo.bar.baz')
const doc = await getter.get(nsid)
```

I ended up changing some of that code to use `Nsid` (extends string) but I figured the mechanism is still worth supporting at an abstract level.

.changeset/cool-books-roll.md
matthieusieben: It's minor because now, `lexicon.get('nsid').defs.main` can return a `LexPermissionSet` which was not the case before. So any code doing things like: 

```ts
const { type } = lexicon.get('nsid').defs.main
switch (type) {
  case "record":
  case "...":
    console.log('all good');
    break
  default:
    throw new Error(`unexpected type ${type}`)
}

```

now has to account for that possibility

packages/dev-env/src/mock/index.ts
matthieusieben: We agreed internally to comment this for now and that more work would be done to improve seeding for threads. This is not being actively used by any team member at the moment.

packages/pds/bin/migration-create.ts
matthieusieben: yes. The migrations generated are not valid and didn't properly work. I could also have fixed the script but I find it simpler to copy paste a previous migration rather than `cd packages/pds; pnpm what-script-do-i-need-again`

matthieusieben: Because it was obviously never used to generate migration in that directory, I didn't fix the script. I can revert this change if you want.

packages/syntax/src/nsid.ts
matthieusieben: This is becoming quite a hot path (being now used on every auth request, refresh, and in the auth UI, in addition to the previous places: lexicon validation, lexicon resolution, at uri validation, etc.). 

The previous implementation was allocating string arrays (without re-purposing the result), executing regexps (on average, about 9 per nsid!), all of which consume processing and memory (+additional GC).

IMO the complexity it worth it, especially since this is so well covered by tests (note that reworking this allowed me to add missing cases in tests)

packages/pds/src/config/env.ts
matthieusieben: One way to look at it is that a service could allow implementing an override mechanism based on the `com.atproto.repo.getRecord` endpoint.

If such a service existed, all that would be needed for a PDS to start using it, would be to set that ENV var.

If you think that is too far fetched, or that any override mechanism would be implemented through another lexicon xrpc, then I can remove this.

packages/pds/src/context.ts
matthieusieben: The difference is that `allowImplicitRedirect` can support `redirect: "follow"` when called like so (explicit redirect):

```ts
safeFetch(url, {
  redirect: "follow"
})
```

But that is only true if the `fetch` function returned by `safeFetchWrap` is called with two arguments.

If we do:

```ts
const req = new Request(url, { redirect: "follow" })

safeFetch(req)
```

then the `allowImplicitRedirect` code cannot distinguish between an explicit redirect (as is actually the case here) and an implicit redirect (default value being used `req = new Request(url)`).

The logging wrapper was doing that (wrapping the fetch args into a `new Request(input, init)`), preventing the `allowImplicitRedirect` code from distinguishing explicit redirects, causing errors.

packages/internal/simple-store/src/simple-store.ts
matthieusieben: answered above 👍

packages/oauth/oauth-provider-ui/src/components/utils/scope-description.tsx
matthieusieben: Technically `includeScope.toString()` is indeed better. Because of the way scopes are defined, it would still be possible to end up with the same key being used twice.

For example `include:foo.bar?aud=did:web:example.com#aaa` and `include:foo.bar?aud=did:web:example.com%23aaa` are considered different scope values but would end up being stringified identically here.

matthieusieben: In other words, scope values are not "logically" deduped.

matthieusieben: Note that in this case, I don't expect any issue as the value will not dynamically change in the UI anyways...

packages/oauth/oauth-scopes/src/scopes/blob-permission.ts
matthieusieben: The change is mostly aestetical. I added a `Matchable` (which I should probably rename to `ResourceScope` or similar) and moved the parsed logic to be a static prop. Nothing else changed.

packages/syntax/src/nsid.ts
matthieusieben: I've mostly reverted to the previous implementation, but I kept some optimizations:
- `new NSID` will only cause `nsid.split('.')` to be called once
- Avoid Regexp when simple enough
- use for-of loop
- Check first and last part outside of the loop

With all those, I was able to get a ~33% speed improvement.
Note that the "full optimized" version was giving a ~75% speed improvement.

packages/pds/src/context.ts
matthieusieben: I've adapted the comments both here an of `safeFetchWrap` to better explain the issue.

packages/oauth/oauth-scopes/src/lib/nsid.ts
matthieusieben: Ended up being able to do it the right way

packages/oauth/oauth-scopes/src/scopes/include-scope.ts
matthieusieben: https://github.com/bluesky-social/atproto/blob/6c6cd4d186b9cea18419a85a9de18195849c3d19/packages/oauth/oauth-scopes/src/scopes/include-scope.ts#L88-L99

packages/oauth/oauth-provider-ui/src/components/utils/scope-description.tsx
devinivy: Sounds good 👍

.changeset/loud-fishes-cover.md
devinivy: That all makes sense, though at the same time I have also seen us write a generic redis store that kept track of key types since both numbers and strings were technically supported by the `SimpleStore` interface.  So I think this approach can sometimes tempt further abstract implementations in order to speak to the full interface.  Though maybe that is actually less likely now since it's become open-ended enough that it's no longer as practical to try to support all use cases.

Also— it's actually great if we _do_ we have concrete implementations like `RedisStore` that we can reuse.  This is where aligning on `string` as a primitive for reusable implementations can be really useful.  As you've illustrated above with `StringBasedLexiconStore`, it is quite practical to build on top of an implementation that supports string keys.

In any case, it seems alright to keep this change to me 👍

.changeset/cool-books-roll.md
devinivy: Hmmm!  Got it 👍

packages/pds/bin/migration-create.ts
devinivy: That's fine, we can remove it if it's out of date and not being used.

packages/syntax/src/nsid.ts
devinivy: ~I don't think I'm seeing the partial reversion—perhaps not pushed yet?~ Oops—found it!  I really like how it turned out.

> The previous implementation was allocating string arrays (without re-purposing the result), executing regexps (on average, about 9 per nsid!), all of which consume processing and memory (+additional GC).

I see these differences between the two, but that on its own doesn't make it obvious to me what the impact is of the optimizations. Clarity and readability are important to balance against, especially in reference implementations.  I think there are often middle paths that balance readability and performance that are worth investigating, since we don't have to optimize fully for one versus the other.  It sounds like you did exactly that, and I'm excited to see what it looks like!

devinivy: (P.S. you also mentioned this is now in the path of every authed request.  Could you help me see where that comes into play?  I was wondering if that was now the case, and I'd like to take a closer look.)

packages/pds/src/config/env.ts
devinivy: I'd recommend nixing the env for now, and we can reintroduce it easily if the use case crops up.  I think we're still learning about the use cases and override behaviors.

packages/oauth/oauth-scopes/src/lib/nsid.ts
devinivy: Oh sick!

devinivy (approved):
Whew, lotta work in here!  This is looking good 👍 

👉 Let's just coordinate prior to merge.  I think there is some work we should land and publish ahead of this one.  Shouldn't cause any conflicts or anything like that.

PR #4113, opened 2025-08-17
fix(identity/did): allow redirects for did:web json fetch

lizthegrey:
Fixes regression in #3177; it should be permissible to have a "reasonable" number of redirects in the chain both for `.well-known/did-proto` and for `.well-known/did.json`.

### DIFF
--- a/packages/identity/src/did/web-resolver.ts
+++ b/packages/identity/src/did/web-resolver.ts
@@ -35,7 +35,6 @@ export class DidWebResolver extends BaseResolver {
     return timed(this.timeout, async (signal) => {
       const res = await fetch(url, {
         signal,
-        redirect: 'error',
         headers: { accept: 'application/did+ld+json,application/json' },
       })
 


lizthegrey:
> I'm about to be traveling, but will ping other folks. I'm not entirely sure that we should support redirects for either of these; we probably discussed in the past but i'm not remembering what the reasoning was. I think folks often set up redirects between entire domains (vs explicitly for individual paths), and that could lead to weird handle and `did:web` scenarios. I guess with `did:web` it is possible to check the identifier in the document against the originally resolved domain, but i'm not sure we are doing that anywhere.

ah. well, for reference see did:web:lizthegrey.com (fuckin squarespace not letting me just upload files to the .well-known directory)

packages/identity/src/did/web-resolver.ts
```diff
@@ -35,7 +35,6 @@ export class DidWebResolver extends BaseResolver {
     return timed(this.timeout, async (signal) => {
       const res = await fetch(url, {
         signal,
-        redirect: 'error',
```
matthieusieben, line 38: Instead of just removing the line, please use `redirect: "follow"`. This makes it explicit that we want to allow follows there, which is required for our SSRF protection tools.

lizthegrey: done.

packages/identity/src/did/web-resolver.ts
```diff
@@ -35,7 +35,7 @@ export class DidWebResolver extends BaseResolver {
...
+        redirect: 'follow',
```
ThisIsMissEm, line 38: @matthieusieben should this actually be using safeFetch here with a max redirects count?

There's also similar code here which does have errors too (that's the `/.well-known/atproto-did` to resolved handle logic): https://github.com/bluesky-social/atproto/blob/9d40ccbb69103fae9aae7e3cec31e9b3116f3ba2/packages/internal/handle-resolver/src/internal-resolvers/well-known-handler-resolver.ts

matthieusieben: Yeah this should definitely be using safe fetch. This code was implemented before SSRF protection was added. It's probably gonna be a little painful to get a `safeFetch` all the way to here in all our code paths.

matthieusieben:
> I guess with `did:web` it is possible to check the identifier in the document against the originally resolved domain, but i'm not sure we are doing that anywhere.

@bnewbold In every case (any did method) we do check that the did document's `id` matches that `did` that was originally used to perform the resolution. So from that stand point, it should be fine:

https://github.com/bluesky-social/atproto/blob/76bea37146ebef7476a427b1bff669be2e274ca6/packages/identity/src/did/base-resolver.ts#L22-L25

As you mentioned though, I am not sure whether we should allow to follow redirects. 1) because it is not mentioned in the did specs, 2) because it would be a non backwards compatible change :/

PR #4122, opened 2025-08-19
@atproto/lexicon: relax validation of lexicon documents

DavidBuchanan314:
Allow unknown fields to be present in lexicon documents

### DIFF
--- a/packages/lexicon/src/types.ts
+++ b/packages/lexicon/src/types.ts
@@ -5,27 +5,23 @@ import { requiredPropertiesRefinement } from './util'
 // primitives
 // =
 
-export const lexBoolean = z
-  .object({
-    type: z.literal('boolean'),
-    description: z.string().optional(),
-    default: z.boolean().optional(),
-    const: z.boolean().optional(),
-  })
-  .strict()
+export const lexBoolean = z.object({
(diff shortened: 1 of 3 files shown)


packages/lexicon/tests/general.test.ts
```diff
@@ -100,21 +100,22 @@ describe('General validation', () => {
...
   })
-  it('fails when unknown fields are present', () => {
+  it('allows unknown fields to be present', () => {
     const schema = {
       lexicon: 1,
       id: 'com.example.unknownFields',
       defs: {
         test: {
           type: 'object',
+          properties: {},
```
DavidBuchanan314, line 110: `properties` field is required, so I had to add it to allow for a successful parse

packages/lexicon/src/types.ts
```diff
@@ -422,7 +382,6 @@ export const lexiconDoc = z
     description: z.string().optional(),
     defs: z.record(lexUserType),
   })
-  .strict()
```
devinivy, line 425: I think we can remove the line above for `$type`—it was originally only added to be loose enough for usage with lexicon resolution.

devinivy (approved):

PR #4139, opened 2025-08-26
Fix #4136: Support multiple redirect URIs in @atproto/oauth-client

### LINKED ISSUE #4136: oauthClient.callback assumes a singular redirect URI, breaks clients using multiple redirect URIs.
**Describe the bug**

In the `callback` method of `OAuthClient` it calls `exchangeCode` from `OAuthServerAgent`. The `exchangeCode` method [does not support multiple redirect URIs](https://github.com/bluesky-social/atproto/blob/main/packages/oauth/oauth-client/src/oauth-server-agent.ts#L99), even though `OAuthClient`'s `authorize` does support multiple redirect URIs.

This was discovered by a user on the [AT Proto Touchers discord](https://discord.com/channels/1097580399187738645/1097580399187738648/1409887534116245658), with investigation into root cause by myself.

The example code to trigger it can be found here: https://pastebin.com/0TZf3fm6

**To Reproduce**

1. Create an application with multiple redirect URIs
2. Try to authorize the second redirect URI, and then perform the callback.

**Expected behavior**

Token exchange should be successful.

**Actual behaviour**

Token exchange fails, because the incorrect redirectURI was used (the first one, not the second)

**Details**

- Operating system: n/a
- Node version: n/a

**Additional context**

I can open up a pull request to fix this.

ThisIsMissEm:
Previously the callback method assumed a singular `redirect_uris` value, and enforced only performing the callback with the first registered redirect URI (via `exchangeCode` being hardcoded to pick the first value from `redirect_uris`). This change allows passing the actual redirect URI to the `callback` method, much like the `authorize` method supports.

Fixes #4136

**Note:** I have not added support for multiple redirect URIs to `@atproto/oauth-client-browser` since the methods exposed there (`signInCallback` via `init`) does not currently support any additional options besides a `refresh` boolean. `readCallbackParams` would probably need to be modified to "discover" the correct redirect URI for which to use and subsequently pass to the `callback` method.

### DIFF
--- a/packages/oauth/oauth-client/src/oauth-client.ts
+++ b/packages/oauth/oauth-client/src/oauth-client.ts
@@ -47,7 +47,7 @@ import {
   SessionStore,
 } from './session-getter.js'
 import { InternalStateData, StateStore } from './state-store.js'
-import { AuthorizeOptions, ClientMetadata } from './types.js'
+import { AuthorizeOptions, CallbackOptions, ClientMetadata } from './types.js'
 import { CustomEventTarget } from './util.js'
 import { validateClientMetadata } from './validate-client-metadata.js'
 
@@ -364,7 +364,10 @@ export class OAuthClient extends CustomEventTarget<OAuthClientEventMap> {
     // @TODO investigate actual necessity & feasibility of this feature
   }
 
-  async callback(params: URLSearchParams): Promise<{
+  async callback(
+    params: URLSearchParams,
+    options: CallbackOptions = {},
+  ): Promise<{
     session: OAuthSession
     state: string | null
   }> {
@@ -413,6 +416,13 @@ export class OAuthClient extends CustomEventTarget<OAuthClientEventMap> {
         stateData.dpopKey,
       )
 
+      const redirectUri =
+        options?.redirect_uri ?? server.clientMetadata.redirect_uris[0]
+      if (!server.clientMetadata.redirect_uris.includes(redirectUri)) {
+        // The server will enforce this, but let's catch it early
+        throw new TypeError('Invalid redirect_uri')
+      }
+
       if (issuerParam != null) {
         if (!server.issuer) {
           throw new OAuthCallbackError(
@@ -438,7 +448,11 @@ export class OAuthClient extends CustomEventTarget<OAuthClientEventMap> {
         )
       }
 
-      const tokenSet = await server.exchangeCode(codeParam, stateData.verifier)
+      const tokenSet = await server.exchangeCode(
+        codeParam,
+        stateData.verifier,
+        redirectUri,
+      )
       try {
         await this.sessionGetter.setStored(tokenSet.sub, {
           dpopKey: stateData.dpopKey,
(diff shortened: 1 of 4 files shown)


packages/oauth/oauth-client/src/oauth-client.ts
```diff
@@ -413,6 +416,13 @@ export class OAuthClient extends CustomEventTarget<OAuthClientEventMap> {
         stateData.dpopKey,
       )
 
+      const redirectUri =
+        options?.redirect_uri ?? server.clientMetadata.redirect_uris[0]
+      if (!server.clientMetadata.redirect_uris.includes(redirectUri)) {
+        // The server will enforce this, but let's catch it early
+        throw new TypeError('Invalid redirect_uri')
```
ThisIsMissEm, line 423: `TypeError` feels most right here, even though all other errors are `OAuthCallbackError`'s. We could also let this cascade through to the server, if we don't want this explicit check.

packages/oauth/oauth-client/src/oauth-server-agent.ts
```diff
@@ -91,12 +91,18 @@ export class OAuthServerAgent {
     }
   }
 
-  async exchangeCode(code: string, codeVerifier?: string): Promise<TokenSet> {
+  async exchangeCode(
+    code: string,
+    codeVerifier?: string,
+    redirectUri?: string,
```
ThisIsMissEm: Whilst I cannot see any other usages of `exchangeCode` in the codebase, I marked the `redirectUri` parameter as optional as to avoid any breaking changes in API.

packages/oauth/oauth-client/src/oauth-client.ts
matthieusieben: Shouldn't the comparison here account for possible variable port number for loopback redirect uris?

There is a function that does that in OAuthProvider.

Alternatively, we can optimistically suppose the client does the right thing and let the oauth provider make the check (which might be why there was no check before).

ThisIsMissEm: Well, the comparison for `redirect_uris` on the authorization server side is simple string comparison: https://www.rfc-editor.org/rfc/rfc6749.html#section-3.1.2.3

So this comparison is correct, it's also the same comparison used in the `authorize` method.

ThisIsMissEm:
@matthieusieben just fixed the type error that caused the workflow to fail.

packages/oauth/oauth-client/src/oauth-client.ts
matthieusieben: I'm making reference to: https://datatracker.ietf.org/doc/html/rfc8252#section-8.4

> Authorization servers MUST require clients to register their complete redirect URI (including the path component) and reject authorization requests that specify a redirect URI that doesn't exactly match the one that was registered; the exception is loopback redirects, where an exact match is required except for the port URI component.

But you're right, the behavior is already present in `authorize()` anyways

matthieusieben:
Looks good. Thank you for tackling this issue 💙

matthieusieben:
I agree with the suggestion you made: `readCallbackParams` sounds like a good place to try to infer the redirect uri.

PR #4140, opened 2025-08-26
adjust notifications to filter thread hidden tags

haileyok:


packages/bsky/src/api/app/bsky/notification/listNotifications.ts
```diff
@@ -195,6 +195,20 @@ const noBlockOrMutesOrNeedsReview = (
...
+    if (
+      item.reason === 'reply' ||
+      item.reason === 'quote' ||
+      item.reason === 'mention'
+    ) {
+      const post = hydration.posts?.get(item.uri)
+      if (post) {
+        post.tags.forEach((tag) => {
+          if (ctx.cfg.threadTagsHide.has(tag)) {
+            return !!hydration.profileViewers?.get(did)?.following
```
ghost: This `return` is for the `forEach` callback, but it won't return on the main flow, which I think is what we want here, right?

I think you want something like:

```ts
        const hide = !hydration.profileViewers?.get(did)?.following && post.tags.some((tag) => ctx.cfg.threadTagsHide.has(tag))
        if (hide) return false
```

haileyok: oop yea. this isn't an array so can't do `some`, but fixed the problem

PR #4142, opened 2025-08-27
revokeAccountCredentials lexicon

DavidBuchanan314:
In temp namespace, for use between entryway and ozone.

### DIFF
--- a/.changeset/five-lions-sing.md
+++ b/.changeset/five-lions-sing.md
@@ -1,0 +1,8 @@
+---
+"@atproto/ozone": patch
+"@atproto/bsky": patch
+"@atproto/api": patch
+"@atproto/pds": patch
+---
+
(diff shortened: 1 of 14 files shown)


.changeset/five-lions-sing.md
```diff
@@ -0,0 +1,8 @@
+---
+"@atproto/ozone": patch
+"@atproto/bsky": patch
+"@atproto/api": patch
+"@atproto/pds": patch
```
devinivy, line 5: This is fine, but just a headsup that it's often alright to only mark `@atproto/api` on lexicon updates, if the idea is to add the lexicon to the client.  I think about it as _which changelogs do I want this line to appear in?_  (And side note—changesets also bump dependencies transitively, so these will organically get a patch version by depending on `@atproto/api`.  In the changelog it will look like a "Dependency update".)

devinivy (approved):

PR #4144, opened 2025-08-27
Support multiple redirect URIs for @atproto/oauth-client-browser

ThisIsMissEm:
I think this might be the appropriate fix here, but I'm not sure if it really is, or if it might break things. This continues on from #4139 but for the browser package.

Whilst working on this, eslint and types in vscode were all manner of weird, so I'm not as confident on this change: https://bsky.app/profile/thisismissem.social/post/3lxfetdclwc2f

cc @matthieusieben

### DIFF
--- a/packages/oauth/oauth-client-browser/src/browser-oauth-client.ts
+++ b/packages/oauth/oauth-client-browser/src/browser-oauth-client.ts
@@ -368,10 +368,10 @@ export class BrowserOAuthClient extends OAuthClient implements Disposable {
 
   async signInCallback() {
     const params = this.readCallbackParams()
-    const redirectUri = this.readRedirectUrl()
+    const redirectUrl = this.readRedirectUrl()
 
     // Not a (valid) OAuth redirect
-    if (!params || !redirectUri) return null
+    if (!params || !redirectUrl) return null
 
     // Replace the current history entry without the params (this will prevent
     // the following code to run again if the user refreshes the page)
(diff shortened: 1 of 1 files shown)


matthieusieben:
Not sure what's the best way to contribute to a branch from a forked repo. Here is a fix for the error you're having

packages/oauth/oauth-client-browser/src/browser-oauth-client.ts
```diff
@@ -344,23 +360,18 @@ export class BrowserOAuthClient extends OAuthClient implements Disposable {
...
     return params
   }
 
   async signInCallback() {
     const params = this.readCallbackParams()
+    const redirectUrl = this.readRedirectUrl()
 
     // Not a (valid) OAuth redirect
-    if (!params) return null
+    if (!params || !redirectUrl) return null
```
matthieusieben: Note we could have the `redirectUrl` fallback to `undefined` (instead of `null`), and not check it here, which would result in the default behavior being used (first item from the list)

matthieusieben: Like so https://github.com/bluesky-social/atproto/pull/4147/commits/a6dd3730c08b91263a22e1c4323b0c32e6bc490d

ThisIsMissEm: Have applied that change, and made `readRedirectUri` return a `OAuthRedirectUri` to make the types happy (even though it's just a string..)

matthieusieben:
Merged through the other PR. Thank you again for taking care of this.

PR #4149, opened 2025-08-28
Allow encoding `scope` claims of oauth access token JWT

matthieusieben:
- Adds `onCreateToken` and `onDecodeToken` hooks to intercept access token JWT claims encoding/decoding
- Adds redis based `SimpleStore` implmenetation
- Adds `com.atproto.temp.decodeScope` lexicon
- Adds decoding, in PDS, of encoded scopes when used with entryway

### DIFF
--- a/packages/pds/src/context.ts
+++ b/packages/pds/src/context.ts
@@ -424,7 +424,7 @@ export class AppContext {
         dpopSecret: secrets.dpopSecret,
         redis: redisScratch,
         onDecodeToken: scopeRefGetter
-          ? async ({ payload, dpopProof, tokenType }) => {
+          ? async ({ payload, dpopProof }) => {
               // @TODO drop this once oauth provider no longer accepts DPoP proof with
               // query or fragment in "htu" claim.
               if (dpopProof?.htu.match(/[?#]/)) {


packages/oauth/oauth-provider/src/oauth-provider.ts
```diff
@@ -1075,38 +1077,25 @@ export class OAuthProvider extends OAuthVerifier {
     }
   }
 
-  protected override async verifyToken(
+  protected override async decodeToken(
```
matthieusieben, line 1080: See changes in `super` (`OAuthVerifier`)

packages/oauth/oauth-provider/src/oauth-verifier.ts
```diff
@@ -134,22 +145,62 @@ export class OAuthVerifier {
...
+        )
+      }
+
+      // Unexpected DPoP proof received for a Bearer token
+      if (dpopProof) {
+        throw new InvalidTokenError(
+          'Bearer',
+          `DPoP proof not expected for Bearer token type`,
+        )
+      }
```
matthieusieben: Would it make sense to actually remove this check ? (there is no harm in providing a DPoP proof)

packages/internal/simple-store-redis/src/index.ts
```diff
@@ -0,0 +1,73 @@
...
+    if (options.ttl != null && options.ttl <= 0) {
+      throw new TypeError(`ttl must be greater than 0`)
+    }
+  }
+
+  protected encodeKey(key: K): string {
+    return `${this.options.keyPrefix}${key satisfies string}`
+  }
+
+  async get(key: K, _options?: GetOptions): Promise<V | undefined> {
```
devinivy: Reading this, a natural question is "why is `_options` ignored?"  I got to the bottom of it, but it may be worth a short comment.

packages/oauth/oauth-provider/src/oauth-verifier.ts
```diff
@@ -118,38 +130,73 @@ export class OAuthVerifier {
...
-  ): Promise<VerifyTokenClaimsResult> {
+  ): Promise<SignedTokenPayload> {
     if (!isSignedJwt(token)) {
       throw new InvalidTokenError(tokenType, `Malformed token`)
     }
 
     const { payload } = await this.signer
-      .verifyAccessToken(token)
+      .verifyAccessToken(token, {
+        requiredClaims: ['iss', 'jti', 'iat', 'exp', 'sub'],
```
devinivy: It would be great if there were a way for readers to easily verify this list is accurate.  E.g. referencing `tokenManager.buildAccessToken()` implementation is probably sufficient.

packages/oauth/oauth-provider/src/token/verify-token-claims.ts
```diff
@@ -25,77 +11,35 @@ export type VerifyTokenClaimsOptions = {
...
+        : options.audience.includes(aud))
+    if (!hasMatch) {
       throw new InvalidTokenError(tokenType, `Invalid audience`)
     }
   }
 
   if (options?.scope) {
     const scopes = tokenClaims.scope?.split(' ')
-    if (!scopes || !options.scope.some((v) => scopes.includes(v))) {
+    if (!scopes || !options.scope.some(includedIn, scopes)) {
```
devinivy: I think that the way this is written is not going to be familiar for many readers.  For that reason I'd recommend leaving it as it was but no biggie either way.

packages/oauth/oauth-provider/src/token/token-claims.ts
```diff
@@ -0,0 +1,14 @@
...
+export type TokenClaims = {
+  jti: TokenId
+  sub: string
+  iat: number
+  exp: number
+  cnf?: { jkt: string }
+  aud?: string | [string, ...string[]]
+  scope?: OAuthScope
+  client_id?: ClientId
+}
```
devinivy: Should `iss` be included here, consistent with the `requiredClaims` list in `decodeToken()`?  I assume there is a reason it's not here, but it's not jumping out at me.

packages/oauth/oauth-provider/src/token/token-manager.ts
```diff
@@ -55,29 +51,45 @@ export class TokenManager {
...
+      scope,
+      // https://datatracker.ietf.org/doc/html/rfc8693#section-4.3
+      client_id: client.id,
+
+      ...(parameters.dpop_jkt && {
+        cnf: { jkt: parameters.dpop_jkt },
+      }),
 
       ...(this.accessTokenMode === AccessTokenMode.stateless && {
         aud: account.aud,
```
devinivy: It looks like `aud` is already included above now.
```suggestion
```

packages/oauth/oauth-provider/src/token/token-manager.ts
```diff
@@ -55,29 +51,45 @@ export class TokenManager {
...
-        scope,
+        scope: scope ?? parameters.scope,
```
devinivy: I believe `scope` is always a string (so never nullish) and also included above.  Can we nix it?
```suggestion
```

packages/oauth/oauth-provider/src/token/token-manager.ts
```diff
@@ -365,40 +380,31 @@ export class TokenManager {
...
-      // These are not stored in the JWT access token in "light" access token
-      // mode. See `buildAccessToken`.
-      aud: account.aud,
-      // Note we fallback to parameters.scope for sessions created before
-      // TokenData.scope was introduced.
-      scope: data.scope ?? parameters.scope,
-      client_id: data.clientId,
+    try {
+      // Fool proof, make sure that the database & token payload are consistent
+      if (tokenPayload.cnf?.jkt !== data.parameters.dpop_jkt) {
```
devinivy: I assume this will compare properly when both should be missing?  E.g. wont get a `undefined !== null` or something along those lines.

packages/oauth/oauth-provider/src/token/token-manager.ts
```diff
@@ -365,40 +380,31 @@ export class TokenManager {
...
+        exp: dateToEpoch(data.expiresAt),
+        aud: account.aud,
+        scope: data.scope ?? data.parameters.scope,
+        // https://datatracker.ietf.org/doc/html/rfc8693#section-4.3
+        client_id: data.clientId,
+      }
+    } catch (err) {
+      await this.deleteToken(tokenId)
+      throw err
     }
```
devinivy: This kind of broad cleanup "delete on error" stands out to me a little bit, as I've seen similar patterns allow for one user to delete another user's resources by triggering an error in a certain code path.  Programmer errors resulting in an unexpected thrown error would have a similar effect.

packages/pds/src/auth-verifier.ts
```diff
@@ -383,28 +382,13 @@ export class AuthVerifier {
           throw err
         })
 
-      // @TODO drop this once oauth provider no longer accepts DPoP proof with
-      // query or fragment in "htu" claim.
```
devinivy, line 387: Could you just remind me the story here / why it's safe to drop this now?

packages/pds/src/context.ts
```diff
@@ -410,13 +411,33 @@ export class AppContext {
...
+              if (dpopProof?.htu.match(/[?#]/)) {
+                oauthLogger.info(
+                  { htu: dpopProof.htu, client_id: payload.client_id },
+                  'DPoP proof "htu" contains query or fragment',
+                )
+              }
+
+              if (!isEncodedScope(payload.scope)) return
+              const scope = await scopeDecoder.get(payload.scope)
+              return { ...payload, scope }
```
devinivy: Just confirming: the full scope string must be `enc:<reference>`, i.e. can't appear alongside other individual scopes 👍

packages/pds/src/account-manager/scope-decoder.ts
```diff
@@ -0,0 +1,46 @@
+import Redis from 'ioredis'
+import { Agent } from '@atproto/api'
+import { DAY } from '@atproto/common'
+import { OAuthScope } from '@atproto/oauth-provider'
+import { CachedGetter } from '@atproto-labs/simple-store'
+import { SimpleStoreMemory } from '@atproto-labs/simple-store-memory'
+import { SimpleStoreRedis } from '@atproto-labs/simple-store-redis'
+
+type EncodedScope = `enc:${string}`
```
devinivy: This is a bit of a nit and I am not sure whether it would be worth changing the terminology everywhere, but I think of this as "referencing" and "dereferencing" the scope rather than "encoding" and "decoding" it.  I wonder if @DavidBuchanan314 may have thoughts on the terminology too.

DavidBuchanan314: I must admit, the encode/decode terminology threw me a little, it took me a while to figure out which feature was being implemented here. But I can't think of anything much better!

One thought is "compress"/"decompress". Also a little misleading since we're not using compression per se, but the underlying motivation is in reducing size.

Specifically for this string prefix, I think `ref:` would work best regardless of terminology elsewhere - "enc" makes me think of encryption.

packages/oauth/oauth-provider/src/token/token-manager.ts
matthieusieben: They are both `string | undefined`. Added a comment.

packages/pds/src/context.ts
matthieusieben: yep

matthieusieben: Adapted to support multiple values

.changeset/brown-boxes-bow.md
```diff
@@ -0,0 +1,5 @@
+---
+"@atproto/oauth-scopes": patch
+---
+
+Add scope normzlization utility
```
devinivy:
```suggestion
Add scope normalization utility
```

packages/pds/src/context.ts
```diff
@@ -433,9 +436,27 @@ export class AppContext {
...
+                  // re-build the scope as the user re-authenticates. This
+                  // should never happen though.
+                  throw InvalidTokenError.from(cause, tokenType)
+                }
+
+                throw new UpstreamFailureError(
+                  'Failed to fetch token permissions',
+                  undefined,
+                  { cause },
+                )
```
devinivy: This is a good reminder—we should consider some retry logic when fetching the scope, if it's not already in there.

PR #4156, opened 2025-08-30
Fix URL regex in api/rich-text

### LINKED ISSUE #4120: Link recognition in api/richtext is broken
**Describe the bug**

The [RegEx to identify URIs in the RichText class](https://github.com/bluesky-social/atproto/blob/331a356ce27ff1d0b24747b0c16f3b54b07a0a12/packages/api/src/rich-text/util.ts#L3) is incorrect. For example, `ko-fi.com` and `日本語.jp` are not considered links. Besides that, the class also rejects `HTTPS://google.com` as a link because of the [logic in this line](https://github.com/bluesky-social/atproto/blob/331a356ce27ff1d0b24747b0c16f3b54b07a0a12/packages/api/src/rich-text/detection.ts#L45).

**To Reproduce**

Add the following strings to `inputs` in  `detectFacets` test in `/packages/api/tests`:
```ts
'HTTPS://google.com',
'ko-fi.com',
'日本語.jp'
```
And the following arrays to `outputs` in the same location:
```ts
[['HTTPS://google.com', 'HTTPS://google.com']],
[['ko-fi.com', 'https://ko-fi.com']],
[['日本語.jp', 'https://日本語.jp']],
```

**Expected behavior**

The test for `detectFacets` in `rich-text-detection`should pass.

**Details**

- Operating system: Linux Mint 22.1
- Node version: v22.9.0

**Additional context**

A RegEx that accepts both `ko-fi.com` and `日本語.jp` should look like that (this passes the unit tests provided above except the case-sensitive one):

```ts
export const URL_REGEX =
  /(^|\s|\()((https?:\/\/\S+)|((?<domain>[a-z0-9\u00a1-\uffff-]*(\.[a-z0-9\u00a1-\uffff]+)+)\S*))/gim
```

`\u00a1-\uffff` covers almost all [unicode symbols](https://symbl.cc/en/unicode-table/), with exception for characters like `/`, `:`, `(`, for example. 

Before refactoring the RegEx, I think is important to understand what the internet specifications and standards says and this [stackoverflow answer](https://stackoverflow.com/a/36667242) is a great start.

In the [bluesky's mobile post composer](https://github.com/bluesky-social/social-app/blob/cced762a7fb7a2729b63922abc34ae5406a58bce/src/view/com/composer/text-input/text-input-util.ts#L88) a much more complex RegEx is used that comes from this **well-reviewed** gist: https://gist.github.com/dperini/729294

I have created [an issue](https://github.com/bluesky-social/social-app/issues/8864) in the bluesky's social-app repo to standardize any link recognition. The plan is to always use the `atproto/api` lib for that. So is important that the link recognition from this repo works well first.

Ah, an attempt to fix the case-sensitive "http" was made [here](https://github.com/bluesky-social/atproto/pull/3002) but, as explained by one of the comments, is incomplete.

arturo32:
This PR fixes  #4120 where the [RegEx in api/rich-text](https://github.com/bluesky-social/atproto/blob/331a356ce27ff1d0b24747b0c16f3b54b07a0a12/packages/api/src/rich-text/util.ts#L3) does not follow any standard. The current RegEx does not validate URIs such as `ko-fi.com` or `日本語.jp`.

The new RegEx is based on the [WHATWG Living Standard](https://url.spec.whatwg.org/) by using an altered version of a [well-reviewed gist by Diego Perini](https://gist.github.com/dperini/729294 ).

## The new RegEx

The new RegEx is considerably longer than the old one. This is mainly due to the complex IPv4 validation. It is also important to notice that the new RegEx does not cover (yet) IPv6.

If the complexity of the new RegEx makes this PR hard to review, the IPv4 validation can be relaxed (allowing any pattern in the format  4 groups of digits that can have between 1 and 3 digits each [ `(\d{1,3}\.){3}\d{1,3}` ]).

Here is the original RegEx from Perini:
```js
/^(?:(?:(?:https?|ftp):)?\/\/)(?:\S+(?::\S*)?@)?(?:(?!(?:10|127)(?:\.\d{1,3}){3})(?!(?:169\.254|192\.168)(?:\.\d{1,3}){2})(?!172\.(?:1[6-9]|2\d|3[0-1])(?:\.\d{1,3}){2})(?:[1-9]\d?|1\d\d|2[01]\d|22[0-3])(?:\.(?:1?\d{1,2}|2[0-4]\d|25[0-5])){2}(?:\.(?:[1-9]\d?|1\d\d|2[0-4]\d|25[0-4]))|(?:(?:[a-z0-9\u00a1-\uffff][a-z0-9\u00a1-\uffff_-]{0,62})?[a-z0-9\u00a1-\uffff]\.)+(?:[a-z\u00a1-\uffff]{2,}\.?))(?::\d{2,5})?(?:[/?#]\S*)?$/i
```

And here is the one that I used:
```js
/(?:^|\s|\()(?<uri>(?<protocol>https?:\/\/)?(?<domain>(?!(?:10|127)(?:\.\d{1,3}){3})(?!(?:169\.254|192\.168)(?:\.\d{1,3}){2})(?!172\.(?:1[6-9]|2\d|3[0-1])(?:\.\d{1,3}){2})(?:[1-9]\d?|1\d\d|2[01]\d|22[0-3])(?:\.(?:1?\d{1,2}|2[0-4]\d|25[0-5])){2}\.(?:1\d\d|2[0-4]\d|25[0-4]|[1-9]\d?)|(?:(?:[a-z0-9\u00a1-\uffff][a-z0-9\u00a1-\uffff_-]*)?[a-z0-9\u00a1-\uffff]\.)+(?<tld>[a-z\u00a1-\uffff]{2,})\.?)(?::\d{2,5})?(?:[/?#]\S*)?)/gim
```

**Main differences between Perini's and my RegEx**:

1. Removal of the required begin of line (`^`) and end of line (`$`) tokens;
2. Removal of 64 characters/bytes limit to the domain;
3. Only `http` and `https` protocols are accepted, if a protocol is present;
4. Keeping of `(?:^|\s|\()` from the old RegEx at the beginning;
5. Extensive use of named capture groups;
6. Change from `(?:[1-9]\d?|1\d\d|2[0-4]\d|25[0-4])` to `(?:1\d\d|2[0-4]\d|25[0-4]|[1-9]\d?))`, putting `[1-9]\d?` in the last position (necessary after the removal of `$` ).

## The new logic

The added named groups and the new logic in the `detectFacets` function solves the problem of not accepting URLs in the upper-case format. It also helps with readability of the code.

The validation of a URI by its TLD part is made only when the TLD exists, i.e., it is not an IP.

## Tests

This fix is compatible with all previous tests that existed and it also add more tests in [here](https://github.com/arturo32/atproto/blob/dafe564e5c9291565e487da65a5110e7830b8bb3/packages/api/tests/rich-text-detection.test.ts#L18):

```js
'HTTPS://google.com',
'https://google.COM',
'ko-fi.com',
'日本語.jp',
'GOOGLE.com',
'https://34.64.0.52',
'198.185.159.145',
'invalid IPs: http://127.0.0.1 https://255.255.255.255 https://0.0.0.0 https://169.254.1.1 https://1.1.1.011',
'invalid URIs: https://google.a https://localhost',
```

The last two are considered only text.

## Notes

This PR, if accepted, should close https://github.com/bluesky-social/atproto/pull/3002 as it solves the case-sensitive problem.

### DIFF
--- a/packages/api/src/rich-text/util.ts
+++ b/packages/api/src/rich-text/util.ts
@@ -1,6 +1,8 @@
 export const MENTION_REGEX = /(^|\s|\()(@)([a-zA-Z0-9.-]+)(\b)/g
+// inspired by https://gist.github.com/dperini/729294 (2018/09/12 version)
+// gist credit: Diego Perini
 export const URL_REGEX =
-  /(^|\s|\()((https?:\/\/[\S]+)|((?<domain>[a-z][a-z0-9]*(\.[a-z0-9]+)+)[\S]*))/gim
+  /(?:^|\s|\()(?<uri>(?<protocol>https?:\/\/)?(?<domain>(?!(?:10|127)(?:\.\d{1,3}){3})(?!(?:169\.254|192\.168)(?:\.\d{1,3}){2})(?!172\.(?:1[6-9]|2\d|3[0-1])(?:\.\d{1,3}){2})(?:[1-9]\d?|1\d\d|2[01]\d|22[0-3])(?:\.(?:1?\d{1,2}|2[0-4]\d|25[0-5])){2}\.(?:1\d\d|2[0-4]\d|25[0-4]|[1-9]\d?)|(?:(?:[a-z0-9\u00a1-\uffff][a-z0-9\u00a1-\uffff_-]*)?[a-z0-9\u00a1-\uffff]\.)+(?<tld>[a-z\u00a1-\uffff]{2,})\.?)(?::\d{2,5})?(?:[/?#]\S*)?)/gim
 export const TRAILING_PUNCTUATION_REGEX = /\p{P}+$/gu
 
 /**
--- a/packages/api/src/rich-text/util.ts
+++ b/packages/api/src/rich-text/util.ts
@@ -1,8 +1,5 @@
 export const MENTION_REGEX = /(^|\s|\()(@)([a-zA-Z0-9.-]+)(\b)/g
-// inspired by https://gist.github.com/dperini/729294 (2018/09/12 version)
-// gist credit: Diego Perini
-export const URL_REGEX =
-  /(?:^|\s|\()(?<uri>(?<protocol>https?:\/\/)?(?<domain>(?!(?:10|127)(?:\.\d{1,3}){3})(?!(?:169\.254|192\.168)(?:\.\d{1,3}){2})(?!172\.(?:1[6-9]|2\d|3[0-1])(?:\.\d{1,3}){2})(?:[1-9]\d?|1\d\d|2[01]\d|22[0-3])(?:\.(?:1?\d{1,2}|2[0-4]\d|25[0-5])){2}\.(?:1\d\d|2[0-4]\d|25[0-4]|[1-9]\d?)|(?:(?:[a-z0-9\u00a1-\uffff][a-z0-9\u00a1-\uffff_-]*)?[a-z0-9\u00a1-\uffff]\.)+(?<tld>[a-z\u00a1-\uffff]{2,})\.?)(?::\d{2,5})?(?:[/?#]\S*)?)/gim
+
 export const TRAILING_PUNCTUATION_REGEX = /\p{P}+$/gu
 
 /**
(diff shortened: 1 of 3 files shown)


packages/api/src/rich-text/util.ts
```diff
@@ -1,6 +1,8 @@
 export const MENTION_REGEX = /(^|\s|\()(@)([a-zA-Z0-9.-]+)(\b)/g
+// inspired by https://gist.github.com/dperini/729294 (2018/09/12 version)
+// gist credit: Diego Perini
```
matthieusieben: Note that simply referencing this does not satisfy the license terms.

> The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.

arturo32: Omg, you are right. I will include this text this afternoon.

arturo32: I also added a comment of the same length of the regex to visualize better the parts of it:

```js
export const URL_REGEX =
  /(?:^|\s|\()(?<uri>(?<protocol>https?:\/\/)?(?<domain>(?!(?:10|127)(?:\.\d{1,3}){3})(?!(?:169\.254|192\.168)(?:\.\d{1,3}){2})(?!172\.(?:1[6-9]|2\d|3[0-1])(?:\.\d{1,3}){2})(?:[1-9]\d?|1\d\d|2[01]\d|22[0-3])(?:\.(?:1?\d{1,2}|2[0-4]\d|25[0-5])){2}\.(?:1\d\d|2[0-4]\d|25[0-4]|[1-9]\d?)|(?:(?:[a-z0-9\u00a1-\uffff][a-z0-9\u00a1-\uffff_-]*)?[a-z0-9\u00a1-\uffff]\.)+(?<tld>[a-z\u00a1-\uffff]{2,}))(?::\d{2,5})?(?:[/?#]\S*)?)/gim
//-(-prefix--)(uri---(-------protocol-------)-(domain---(not-private-and-loopback-ips)(---not-system-and-class-c-private-ips--)(----------not-class-b-private-ips-----------)(----------ip-1st-oct------------)(----------ip-2nd-and-3rd-oct---------)--(-----------ip-4th-oct------------)-(--------------------------------dns-domain---------------------------------)-(-------------tld------------))(---port---)-(---path---)-)
```

arturo32:
Hey, @matthieusieben, is there something I could do to make the review of this PR faster? Like,

1. Is performance a concern? I could do some tests with thousands of posts with links on them (in regex101 the old regex **seems** to have a mean of 1ms while the new regex takes 3ms);
2. Is the copyright notice an impediment? I could remove the IP identification part, making it different enough from the original code so that a copyright notice is not needed anymore (I would keep the "inspired by" old comment);
3. Should I propose a library instead? Is reliability in this long regex a concern?

There are more than [5 issues in social-app](https://github.com/bluesky-social/social-app/issues/8864) that are related to URLs, so the acceptance of non-latin characters, dashes and non-case-sensitive matching that this PR provides are things that I consider important.

nicknacnic:
Hit this in the wild today — posted from a registered IDN domain, даркнетиан.com (xn--80aakfpkvdvx.com), and the naked domain rendered as plain text with no facet. So #4120 isn't theoretical; real IDN domains just don't linkify.

Two things worth flagging for review:

1. Matching the domain is only half of it. Once the regex catches 日本語.jp, the facet uri is still the raw Unicode string — that should probably be IDNA/punycode-encoded (https://xn--…) so the link feature is a valid, resolvable URI. Otherwise you detect it and emit a uri some clients will choke on.
2. There's a homograph stance to reconcile. Over in social-app #7308, IDN handles only display as Unicode if they pass a TR-39 single-script check, explicitly to limit spoofing. This PR would auto-link any Unicode domain with no such guard — same risk, opposite posture. Might be worth running naked IDN links through the same single-script check so the two paths agree, and so this doesn't stall on a security objection.

PR #4169, opened 2025-09-03
Various perf fixes

matthieusieben:


packages/aws/src/s3.ts
```diff
@@ -53,70 +61,70 @@ export class S3BlobStore implements BlobStore {
...
+    // which will not cause the pending request to be actually destroyed, and
+    // can lead any further request (made through the same HTTP Agent) to hang
+    // indefinitely. Note that older version of AWS's SDK also presented this
+    // issue for *every* request made through the client. So make sure to use
+    // recent enough version (>= 3.879.0).
+
+    // The main drawback of using "PutObjectCommand" vs "Upload" is that
+    // "PutObjectCommand" does not support concurrent chunked uploads. This
+    // should not be an issue in our case though as the chunk size is 5MB, which
+    // is smaller than most blobs.
```
matthieusieben:
```suggestion
    // is larger than most blobs.
```

.changeset/famous-mice-pay.md
```diff
@@ -0,0 +1,5 @@
+---
+"@atproto/aws": patch
+---
+
+BlobStore's `deleteMany` now supports any number of input (and with process deletes by chunks internally)
```
matthieusieben:
```suggestion
`S3BlobStore`'s `deleteMany` now supports any number of input (and will process deletes by chunks internally)
```

.changeset/silver-squids-listen.md
```diff
@@ -0,0 +1,5 @@
+---
+"@atproto/pds": patch
+---
+
+Fully read CAR stream outside of transaction in `importRepo`
```
matthieusieben:
```suggestion
Fully read CAR stream before starting transaction in `importRepo`, resulting in faster transactions
```

.changeset/stale-rocks-press.md
```diff
@@ -0,0 +1,5 @@
+---
+"@atproto/aws": patch
+---
+
+Set a 5 seconds timeout on every BlobStore requests
```
matthieusieben:
```suggestion
Set a 5 seconds timeout on every `S3BlobStore` requests
```

.changeset/dry-cows-roll.md
```diff
@@ -0,0 +1,5 @@
+---
+"@atproto/repo": patch
+---
+
+Add iterator support to the `BlockMap` class
```
devinivy, line 5: Huh!  I'm curious, how did this tie in?

.changeset/chilled-shirts-ring.md
```diff
@@ -0,0 +1,5 @@
+---
+"@atproto/aws": patch
+---
+
+Optimistically attempt to move files before checking for their excistance, resulting in faster `makePermanent` calls
```
devinivy:
```suggestion
Optimistically attempt to move files before checking for their existence, resulting in faster `makePermanent` calls
```

packages/aws/src/s3.ts
```diff
@@ -53,70 +61,70 @@ export class S3BlobStore implements BlobStore {
...
+    // which will not cause the pending request to be actually destroyed, and
+    // can lead any further request (made through the same HTTP Agent) to hang
+    // indefinitely. Note that older version of AWS's SDK also presented this
+    // issue for *every* request made through the client. So make sure to use
+    // recent enough version (>= 3.879.0).
+
+    // The main drawback of using "PutObjectCommand" vs "Upload" is that
+    // "PutObjectCommand" does not support concurrent chunked uploads. This
+    // should not be an issue in our case though as the chunk size is 5MB, which
+    // is larger than most blobs.
```
DavidBuchanan314: Does this mean we need to buffer the whole blob in memory? Most blobs are small but they can be up to 100MB I think (for videos), which would get a bit hairy if many uploads are happening concurrently.

.changeset/stale-rocks-press.md
```diff
@@ -0,0 +1,5 @@
+---
+"@atproto/aws": patch
+---
+
+Set a 5 seconds timeout on every `S3BlobStore` requests
```
devinivy: We should scrutinize whether we like this timeout, as sometimes are are working with somewhat large files (e.g. videos).

packages/aws/src/s3.ts
DavidBuchanan314: I feel like we're between a rock and a hard place here, but my understanding of the Upload bug is that it will only hang the current S3 connection. If an uploadBlob triggers the hang, it is outside of the db transaction so it is at worst a resource leak, and should not affect e.g. subsequent createRecord request (which will create a fresh S3 connection).

Unless the S3 library is doing some connection-reuse magic internally... I'll take a look

.changeset/dry-cows-roll.md
matthieusieben: Was looking for ways to optimize the operations performed during transactions.

packages/aws/src/s3.ts
matthieusieben: This comment is partly based on previous version of the SDK and might be slightly nacurate.

.changeset/stale-rocks-press.md
matthieusieben: File uploads are still using a larger (configurable) timer (which defaults to 10 seconds)

matthieusieben: This only applies to other requests (like move / delete / etc.)

matthieusieben:
```suggestion
Set a 5 seconds timeout on every `S3BlobStore` requests (does not apply to file uploads)
```

packages/aws/src/s3.ts
matthieusieben: My comment is also based on observation. We know `Upload` can trigger hangs. We have transactions that are in a dead-lock but during which we do not make any call to `Upload` => `Upload` must be causing *other* http request to hang.

packages/pds/src/actor-store/blob/transactor.ts
```diff
@@ -113,38 +115,68 @@ export class BlobTransactor extends BlobReader {
...
+        concurrency: 20,
+        // The blob store should already limit the time of every operation. We
+        // add a timeout here as an extra precaution.
+        timeout: 60 * SECOND,
+        throwOnTimeout: true,
+      })
+
+      // Will reject as soon as any task fails, causing the "finally" block
+      // below to run, aborting every other pending tasks.
+      await queue.addAll(tasks, { priority: 1 })
```
devinivy: I don't think there's any significance to the priority here, so we can probably omit this.
```suggestion
      await queue.addAll(tasks)
```

packages/pds/src/actor-store/blob/transactor.ts
```diff
@@ -177,54 +205,84 @@ export class BlobTransactor extends BlobReader {
...
-          ? w.blobs
-          : [],
-      )
-      .flat()
-      .map((b) => b.cid.toString())
     const cidsToKeep = [
-      ...newBlobCids,
+      ...writes
+        .filter((w) => isUpdate(w) || isCreate(w))
+        .flatMap((w) => w.blobs.map((b) => b.cid.toString())),
```
devinivy: I found the `newBlobCids` variable to be a useful way to self-document what's going on:
```ts
    const cidsToKeep = [
      ...newBlobCids,
      ...duplicateCids.map((row) => row.blobCid),
    ]
```

"We're keeping new CIDs plus any that had duplicates."

packages/aws/src/s3.ts
```diff
@@ -1,33 +1,41 @@
...
-  'apiVersion'
->
+export type S3Config = {
+  bucket: string
+  requestTimeoutMs?: number
+  uploadTimeoutMs?: number
+} & Omit<S3ClientConfig, 'apiVersion' | 'requestHandler'>
 
 // @NOTE we use Upload rather than client.putObject because stream
 // length is not known in advance. See also aws/aws-sdk-js-v3#2348.
```
DavidBuchanan314: This comment is outdated now

DavidBuchanan314 (approved):
I'm quite confident that these changes will fix the locking issue, primarily through not using Upload. (as confident as we can be about a bug we can't reproduce locally!)

As I mentioned earlier, I'm happy to see some of those `Promise.all()`s go away.

matthieusieben:
See also https://github.com/aws/aws-sdk-js-v3/pull/7322

DavidBuchanan314 (approved):

PR #4178, opened 2025-09-08
Various improvements

matthieusieben:


packages/pds/src/actor-store/blob/transactor.ts
```diff
@@ -241,56 +224,113 @@ export class BlobTransactor extends BlobReader {
...
+    // blobs to make permanent before actually making any of them permanent.
 
-          throw err
-        })
+    // If any upload task fails (e.g. due to a timeout), we might still end-up
+    // in a situation where some blobs are made permanent while the transaction
+    // will be aborted. This should be fixed by moving the code that makes blobs
+    // permanent outside of the transaction (e.g. in the background queue). This
+    // is fine for now since this can't be triggered by a user (only by failing
+    // uploads, which is a server issue).
```
DavidBuchanan314: Yup, this aligns with my understanding :+1:

packages/pds/src/actor-store/blob/transactor.ts
```diff
@@ -241,56 +224,113 @@ export class BlobTransactor extends BlobReader {
...
+        )
+
+        throw err
+      }
+    })
+
+    // Limit the number of parallel requests made to the BlobStore by using a
+    // queue with concurrency management.
+    const queue = new PQueue({
+      concurrency: 20,
```
DavidBuchanan314, line 328: Was there any particular reasoning behind the choice of 20, as opposed to 50 as in blobStore deleteMany?

packages/pds/src/actor-store/blob/transactor.ts
```diff
@@ -241,56 +224,113 @@ export class BlobTransactor extends BlobReader {
...
+
+    // @NOTE it is less than ideal to perform i/o operations during a
+    // transaction. Especially since there have been instances of the actor-db
+    // being locked, requiring to kick the processes. The better solution would
+    // be to update the blob state in the database (e.g. "makeItPermanent") and
+    // to process those updates outside of the current transaction.
+    const updatedCids = await this.makeBlobsPermanent(blobsToMakePermanent)
+
+    // @TODO Used to rely on the (non partial !) "blob_tempkey_idx" index. That
+    // index can now be removed.
```
DavidBuchanan314: Neat!

packages/pds/src/actor-store/blob/transactor.ts
matthieusieben: `deleteMany()` uses chunks of 500, which was how it always has been. Not sure where that number comes from.

20 here is another arbitrary value that was introduced by https://github.com/bluesky-social/atproto/pull/4169 here:
https://github.com/bluesky-social/atproto/blob/055a413fba4fab510ec899377154f1204ab12099/packages/pds/src/actor-store/blob/transactor.ts#L139

PR #4183, opened 2025-09-08
Update to node.js 22 (with typescript)

ThisIsMissEm:
I was looking at the code going "why are we still kind of on node.js 18 (which is deprecated)?", and then I noticed some stuff is actually running under node.js 20.

So I just went ahead and moved everything up to Node.js 24, and resolved as many of the type issues as I could. Unfortunately it meant a fair few `as Uint8Array<ArrayBuffer>` in places, because many were using `Uint8Array` by itself, which meant dealing with `ArrayBufferLike` which isn't super compatible in all the dependencies which expect just an `ArrayBuffer`, so there's a bunch of upstream dependencies that could move to accepting `ArrayBufferLike` from the later node.js types.

I'm not sure exactly what is going on in `packages/oauth/oauth-provider-frontend/src/routeTree.gen.ts` though.

For `req.ip` there is a special typescript way to extend Express's `req` type, but I can't remember what it is.

I think `pnpm upgrade @types/node@*` may have triggered an update to all package versions that were upgradable, not just the types.

### DIFF
--- a/services/pds/Dockerfile
+++ b/services/pds/Dockerfile
@@ -1,6 +1,8 @@
-# @NOTE just a temp fix: alpine3.19 breaks sharp install, see nodejs/docker-node#2009
-# see additional reference to this image further down.
-FROM node:22.19-alpine3.21 AS build
+FROM node:22.19-alpine3.22 AS build
+
+# @NOTE just a temp fix: 22.19-alpine3.22 doesn't have python by default,
+# see: https://github.com/nodejs/docker-node/issues/2009
+RUN apk add --update --no-cache python3 build-base gcc && ln -sf /usr/bin/python3 /usr/bin/python
 
 RUN corepack enable
 
@@ -57,7 +59,7 @@ RUN pnpm install --prod --shamefully-hoist --frozen-lockfile --prefer-offline > 
 WORKDIR services/pds
 
 # Uses assets from build stage to reduce build size
-FROM node:22.19-alpine3.21
+FROM node:22.19-alpine3.22
 
 RUN apk add --update dumb-init
 
--- a/packages/dev-infra/docker-compose.yaml
+++ b/packages/dev-infra/docker-compose.yaml
@@ -1,4 +1,3 @@
-version: '3.8'
 services:
   # An ephermerally-stored postgres database for single-use test runs
   db_test: &db_test
--- a/services/bsky/Dockerfile
+++ b/services/bsky/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:22.19-alpine3.21 as build
+FROM node:22.19-alpine3.22 AS build
 
 RUN corepack enable
 
@@ -42,7 +42,7 @@ RUN pnpm install --prod --shamefully-hoist --frozen-lockfile --prefer-offline > 
 WORKDIR services/bsky
 
 # Uses assets from build stage to reduce build size
-FROM node:22.19-alpine3.21
+FROM node:22.19-alpine3.22
 
 # dumb-init is used to handle signals properly.
 # runit is installed so it can be (optionally) used for logging via svlogd.
--- a/services/bsync/Dockerfile
+++ b/services/bsync/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:22.19-alpine3.21 as build
+FROM node:22.19-alpine3.22 AS build
 
 RUN corepack enable
 
@@ -30,7 +30,7 @@ RUN pnpm install --prod --shamefully-hoist --frozen-lockfile --prefer-offline > 
 WORKDIR services/bsync
 
 # Uses assets from build stage to reduce build size
-FROM node:22.19-alpine3.21
+FROM node:22.19-alpine3.22
 
 RUN apk add --update dumb-init
 
--- a/services/ozone/Dockerfile
+++ b/services/ozone/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:22.19-alpine3.21 as build
+FROM node:22.19-alpine3.22 AS build
 
 RUN corepack enable
 
@@ -37,7 +37,7 @@ RUN pnpm install --prod --shamefully-hoist --frozen-lockfile --prefer-offline > 
 WORKDIR services/ozone
 
 # Uses assets from build stage to reduce build size
-FROM node:22.19-alpine3.21
+FROM node:22.19-alpine3.22
 
 RUN apk add --update dumb-init
 


packages/oauth/oauth-client-browser/src/browser-runtime-implementation.ts
```diff
@@ -11,6 +11,7 @@ import {
  */
 const nativeRequestLock: undefined | RuntimeLock = navigator.locks?.request
   ? <T>(name: string, fn: () => T | PromiseLike<T>): Promise<T> =>
+      // @ts-expect-error
```
ThisIsMissEm: I couldn't figure out the error here, it's basically complaining the `T` can be something different from `Promise<T>` (i.e., the `T`'s aren't compatible), but I have no idea how that can be the case.

packages/oauth/jwk-jose/src/jose-key.ts
```diff
@@ -192,7 +196,7 @@ export class JoseKey<J extends Jwk = Jwk> extends Key<J> {
 
     if (typeof input === 'object') {
       // Jwk
-      if ('kty' in input || 'alg' in input) {
+      if (isJwk(input)) {
```
ThisIsMissEm, line 199: Needed to use type-narrowing via `is`

package.json
```diff
@@ -6,9 +6,9 @@
   "license": "MIT",
   "private": true,
   "engines": {
-    "node": ">=18.7.0"
+    "node": ">=22.0.0"
```
ThisIsMissEm:
```suggestion
    "node": ">=24.0.0"
```

.python-version
```diff
@@ -0,0 +1 @@
+3.13
```
ThisIsMissEm: This is likely necessary for anyone using `pyenv` to manage python, as previously python 3.11 was required but now python 3.13 can be used.

ThisIsMissEm:
Sweet! Got a green build. There's probably some cache weirdness going on which is why I needed to pin sharp at the version I did.

services/pds/Dockerfile
```diff
@@ -1,6 +1,8 @@
-# @NOTE just a temp fix: alpine3.19 breaks sharp install, see nodejs/docker-node#2009
-# see additional reference to this image further down.
-FROM node:20.11-alpine3.18 AS build
+FROM node:22.19-alpine3.22 AS build
+
+# @NOTE just a temp fix: 22.19-alpine3.22 doesn't have python by default,
+# see: https://github.com/nodejs/docker-node/issues/2009
+RUN apk add --update --no-cache python3 build-base gcc && ln -sf /usr/bin/python3 /usr/bin/python
```
ThisIsMissEm, line 5: This isn't the prettiest workaround, but it does work.

ThisIsMissEm:
I've now been able to build the `services/pds` image, but I haven't been able to test it because I don't have a good `.env` file handy to make it work (I'm guessing I'd need to use https://github.com/bluesky-social/pds and swap out the pds version with my local version somehow)

package.json
```diff
@@ -59,7 +59,22 @@
...
+      "@datadog/pprof",
+      "@swc/core",
+      "better-sqlite3",
+      "cbor-extract",
+      "dd-trace",
+      "esbuild",
+      "protobufjs",
+      "puppeteer",
+      "sharp"
+    ]
```
ThisIsMissEm, line 77: I'm not a huge fan of needing this, but various versions were being _very_ angry due to prebuilt versions not being available for the node.js version, especially for better-sqlite3 and sharp.

package.json
```diff
@@ -38,7 +38,7 @@
     "@swc/core": "^1.3.42",
     "@swc/jest": "^0.2.24",
     "@types/jest": "^28.1.4",
-    "@types/node": "^18.19.67",
+    "@types/node": "^24.3.1",
```
ThisIsMissEm, line 41: Should probably drop this back to 22 since that's what I ended up using.

.python-version
```diff
@@ -0,0 +1 @@
+3.11
```
matthieusieben, line 1: hu ?

packages/bsky/src/data-plane/server/routes/records.ts
```diff
@@ -56,7 +56,10 @@ export const getRecords =
       const indexedAt = row?.indexedAt
         ? Timestamp.fromDate(new Date(row?.indexedAt))
         : undefined
-      const recordBytes = ui8.fromString(json, 'utf8')
+      const recordBytes = ui8.fromString(
+        json,
+        'utf8',
+      ) as Uint8Array<ArrayBuffer>
```
matthieusieben, line 62: We can probably replace lots of these with `Buffer.from`

packages/oauth/jwk-jose/src/jose-key.ts
```diff
@@ -232,7 +236,7 @@ export class JoseKey<J extends Jwk = Jwk> extends Key<J> {
   }
 
   static async fromJWK(
-    input: string | Record<string, unknown>,
+    input: string | Jwk | JWK,
```
matthieusieben, line 239: What motivated this ?

Since there is a parser inside here (and since it already is pretty permissive because of the `string` type), I don't think we need to put the responsibility on the caller to provide a strict type.

matthieusieben:
> Unfortunately it meant a fair few as Uint8Array<ArrayBuffer> in places, because many were using Uint8Array by itself, which meant dealing with ArrayBufferLike which isn't super compatible in all the dependencies which expect just an ArrayBuffer, so there's a bunch of upstream dependencies that could move to accepting ArrayBufferLike from the later node.js types.

Wouldn't it be possible to leave the internal types as-is, and typecase `ArrayBufferLike` as `ArrayBuffer` whenever interacting with those external libs? This means that as soon as they upgrade, we can simply remove the type cast, instead of carrying the `Uint8Array<ArrayBuffer>` as legacy forever.

.python-version
matthieusieben, line 1: I understand we need `python` through the docker build but I am not sure I understand how this particular file gets used ?

matthieusieben:
Looks like lots the issues come from the fact that TS was upgraded to 5.9, which should be a separate work

.python-version
ThisIsMissEm: This makes things play nicely when using pyenv (like nvm for python)

https://github.com/pyenv/pyenv

Basically if you only have python >= 3.12 in your $PATH then node-gyp will fail to compile because distutils was removed between 3.11 and 3.12: https://docs.python.org/3/whatsnew/3.12.html

packages/oauth/jwk-jose/src/jose-key.ts
ThisIsMissEm: We'd probably at minimum need the type narrowing via that isJwk check I added, since the parser needs a Jwk in one code path and Record<string, unknown> doesn't conform to Jwk's types

though only a typescript 5.9 issue

PR #4198, opened 2025-09-11
Update node.js to 22, pnpm to 10

ThisIsMissEm:
I'm still adding in the pyenv file for python 3.11, since that is still required by better-sqlite3 and sharp at the current versions.

It looks like we do need to approve builds for both sharp and better-sqlite3.

Even though better-sqlite3 @ 10.0.0 should support node.js 22, the prebuilt binary is only available via install script, which for some reason always triggers a fresh build even though the appropriate prebuilt binary should be available: https://github.com/WiseLibs/better-sqlite3/releases/tag/v10.0.0

That said, installing `npm:@atproto/pds@0.3.0-entryway.3` for `@atproto/pds-entryway` which is used in the `entryway.test.ts` causes `better-sqlite3@9.4.0` to be installed which **does not** support node.js 22, this could be causing other weird issues where the wrong version of a dependency is being loaded or something.

To rectify the issue with the entryway for testing the PDS code, I've opened or prepared the following PRs:
- #4255 
- https://github.com/ThisIsMissEm/bluesky-social-atproto/pull/1 — this is a minimal enough Entryway implementation built off the current PDS implementation.


matthieusieben:
`handle.invalid` snapshot failures are due to flaky tests. Restarting failed tests usually helps make those test pass.

package.json
```diff
@@ -59,7 +59,24 @@
...
+      "@datadog/native-iast-taint-tracking",
+      "@datadog/native-metrics",
+      "@datadog/pprof",
+      "@swc/core",
+      "cbor-extract",
+      "dd-trace",
+      "esbuild",
+      "protobufjs",
+      "puppeteer"
+    ],
```
ThisIsMissEm, line 75: these may need to be approved, but for now everything works without them being approved from what I can tell.

services/pds/Dockerfile
```diff
@@ -1,6 +1,8 @@
-# @NOTE just a temp fix: alpine3.19 breaks sharp install, see nodejs/docker-node#2009
-# see additional reference to this image further down.
-FROM node:20.11-alpine3.18 AS build
+FROM node:22.19-alpine3.22 AS build
+
+# @NOTE just a temp fix: 22.19-alpine3.22 doesn't have python by default,
+# see: https://github.com/nodejs/docker-node/issues/2009
+RUN apk add --update --no-cache python3 build-base gcc && ln -sf /usr/bin/python3 /usr/bin/python
```
ThisIsMissEm, line 5: Weirdly enough this installs python 3.12 which iirc has issues compiling the older versions of node.js addons, but for some reason in this image it seems to work fine. It may be because the installation is able to use a prebuilt binary for this architecture, so doesn't need to compile from scratch (which I guess is not the case on my host machine, an M3 Max Mac)

I haven't been able to verify this though, as I can't get pnpm to give me proper debug logs for installs.

aspiers:
I see that the handle is supposed to be created in `packages/dev-env/src/service-profile-ozone.ts` which is invoked from `TestNetwork` in `packages/dev-env/src/network.ts`.

I guess the point of the tests is to mock moderation actions and test that those admin actions are proxied so that the events appear as created by a catch-all moderation authority identity (`mod-authority.test` in this fake test environment) rather than identifying individual admins?  IIUC, then there's something going wrong relating to that handle.

But given that the ids in `packages/pds/tests/proxied/__snapshots__/admin.test.ts.snap` never go above 7, the fact that the tests are showing much higher ids suggests that either there's some pollution in a db somewhere, or something else is erroneously creating more ids than expected, perhaps due to some non-deterministic ordering.  I do feel like I'm scrabbling around in the dark a bit though.

PR #4202, opened 2025-09-14
Adjust uploadBlob rate limit: reduce window to 1 hour, increase limit to 2500 requests

afbase:
Updated the rate limiting configuration for the com.atproto.repo.uploadBlob endpoint to allow an improved migration experience.

- 2,500 uploads per 1-hour period per IP
- 2,500 uploads per hour burst capacity
- Resets every hour instead of daily

This change provides:
- Better user experience: Users can upload more blobs without hitting daily limits
- Faster recovery: Rate limits reset hourly instead of waiting up to 24 hours
- Maintained protection: Still prevents sustained abuse (2,500/hour is reasonable for legitimate use)

### DIFF
--- a/packages/pds/src/api/com/atproto/repo/uploadBlob.ts
+++ b/packages/pds/src/api/com/atproto/repo/uploadBlob.ts
@@ -1,4 +1,4 @@
-import { DAY } from '@atproto/common'
+import { HOUR } from '@atproto/common'
 import { UpstreamTimeoutError, parseReqEncoding } from '@atproto/xrpc-server'
 import { BlobMetadata } from '../../../../actor-store/blob/transactor'
 import { AppContext } from '../../../../context'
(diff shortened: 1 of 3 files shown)


afbase:
1. I have since revised the PR so that the repo.uploadBlob's rate limit now references a configuration defined by either environment variables or the original default values (1000, DAY).
2. I changed it to be environment variable configurable as PDS Owners could choose to have different rate limit preferences for that xrpc call.

erikologic:
Can we merge this pls!?

matthieusieben:
@copilot resolve the merge conflicts in this pull request

matthieusieben:
humm looks like the "Fix with Copilot" does not work when the PR is from a fork. Could you fix the PR and create a `pnpm changeset` please?

PR #2398, opened 2024-04-10
React Native OAuth Client

haileyok:
This will be ugly for a bit

- Working through needed changes
- Working out whether existing crypto implementation will suffice or if a small native module will be required
- Includes all of the base OAuth branch, and since it isn't inside of this repo I can't set that branch as the base. Oh well!

PR #4220, opened 2025-09-23
`oauth-client-expo`

matthieusieben:


packages/oauth/jwk/src/key.ts
```diff
@@ -8,7 +8,7 @@ import { cachedGetter } from './util.js'
 const jwkSchemaReadonly = jwkSchema.readonly()
 
 export abstract class Key<J extends Jwk = Jwk> {
-  constructor(protected readonly jwk: Readonly<J>) {
+  constructor(readonly jwk: Readonly<J>) {
```
matthieusieben, line 11: This needs to be public in `ExpoKey`. We could expose it there (through another prop) or make it public here

package.json
```diff
@@ -48,6 +48,7 @@
     "eslint-plugin-import": "^2.31.0",
     "eslint-plugin-n": "^17.15.0",
     "eslint-plugin-prettier": "^5.1.3",
+    "expo-module-scripts": "^5.0.7",
```
matthieusieben: Needed at root for `tsconfig` files

packages/oauth/oauth-client-browser/src/browser-oauth-client.ts
```diff
@@ -232,12 +232,10 @@ export class BrowserOAuthClient extends OAuthClient implements Disposable {
...
 
-  signIn(
+  async signIn(
     input: string,
-    options: AuthorizeOptions & { display: 'popup' },
-  ): Promise<OAuthSession>
-  signIn(input: string, options?: AuthorizeOptions): Promise<never>
-  async signIn(input: string, options?: AuthorizeOptions) {
+    options?: AuthorizeOptions,
+  ): Promise<OAuthSession> {
```
matthieusieben, line 244: Overload causes `Promise<OAuthSession>` not assignable to `Promise<never>`

ThisIsMissEm:
Might be worth linking these pull requests: https://github.com/bluesky-social/atproto/pull/2398

packages/oauth/oauth-client-expo/android/src/main/java/expo/modules/atprotooauthclient/Jose.kt
```diff
@@ -0,0 +1,116 @@
...
+
+    options.typ?.let {
+      if (jwt.header.type.toString() != it) {
+        throw InvalidPayloadException("typ mismatch")
+      }
+    }
+
+    val claims = jwt.jwtClaimsSet
+
+    options.requiredClaims?.let { requiredClaims ->
```
haileyok, line 71: i am reasonably confident in these checks, but worth having your eyes on them all to make sure it lines up with what we are doing elsewhere

packages/oauth/oauth-client-expo/ios/Jose.swift
```diff
@@ -0,0 +1,137 @@
...
+      protectedHeader["typ"] = header.typ
+    }
+    if header.cty != nil {
+      protectedHeader["cty"] = header.cty
+    }
+    if header.crit != nil {
+      protectedHeader["crit"] = header.crit
+    }
+
+    if let typ = options.typ {
```
haileyok, line 70: same here

haileyok:
once we merge/publish we'll want to make sure that installing this in an outside project works for ios (particulary the cocoapods part). should be fine, but i've never tried to do this with pnpm. worst case, we can get an assist from the expo folks

haileyok (approved):
putting my "this looks good from reading" and my "i wrote this and used it on my machine" stamps on this, assuming that we are confident in the jwt validation logic. thanks for porting this over, awesome stuff

packages/oauth/oauth-client-expo/android/src/main/java/expo/modules/atprotooauthclient/Jose.kt
matthieusieben: This seems about right. TBH I think we should move this code to the JS. We might do that in another pass though.

PR #4221, opened 2025-09-23
Add email based otp to @atproto/pds

ThisIsMissEm:
This pull request adds support for email-based OTP on sign-in for `@atproto/pds`, additionally, on sign-up, it sends the confirm email email.  The email in the 2FA error is obfuscated.

We may need some way to disable this for Bluesky hosted PDSes, though I'm not 100% sure what that would look like at the moment (suggestions welcome from the Bluesky team). Perhaps it's disabled based on the presence of an entryway?

TBD: Should AppPasswords require email-based OTP? I don't think so, but need confirmation

In `@atproto/oauth-provider-frontend` I have:
- Fixed a bug where upon incorrect OTP, the code field was not reset for the subsequent login attempts.
- Added display of the 2FA hint message

@bluesky.team: It could be advantageous to add [mailpit](https://mailpit.axllent.org/) to the development infrastructure. This would allow actually receiving emails and also writing tests that include email-based flows (and remove the need for overwriting the mailer with `mockMailer` / `MailCatcher`).

There's a screen recording here: https://bsky.app/profile/thisismissem.social/post/3lzjp27tfzk2s

Screenshot of the change to `@atproto/oauth-provider-frontend`:
Screenshot 2025-10-06 at 21 58 54


packages/pds/src/mailer/templates/confirm-signin.hbs
```diff
@@ -0,0 +1,152 @@
...
+                    >To confirm the sign in for your account, enter the code
+                      below in the app.</p><code
+                      style="display:block;padding:16px;border-radius:8px;border-width:1px;border-style:solid;background-color:hsl(211, 20%, 95.3%);border-color:hsl(211, 20%, 85.89999999999999%);font-size:14px;letter-spacing:0.25px;font-family:monospace;text-transform:uppercase"
+                    >{{token}}</code>
+                    <p
+                      style="font-size:14px;line-height:1.4;margin:0px 0px;letter-spacing:0.25px;color:hsl(211, 20%, 53%);font-family:-apple-system, BlinkMacSystemFont, &#x27;Roboto&#x27;, &#x27;Oxygen&#x27;, &#x27;Ubuntu&#x27;, &#x27;Cantarell&#x27;, &#x27;Fira Sans&#x27;, &#x27;Droid Sans&#x27;, &#x27;Helvetica Neue&#x27;, sans-serif;padding-top:12px"
+                    >If this wasn&#x27;t you, we recommend taking steps to
+                      protect your account by changing your password.</p>
```
ThisIsMissEm, line 76: It feels like this should link somewhere, but I'm not sure to where?

packages/oauth/oauth-provider/src/errors/invalid-credentials-error.ts
```diff
@@ -0,0 +1,20 @@
...
+ * {@link https://datatracker.ietf.org/doc/html/rfc6749#section-5.2 | RFC6749 - Issuing an Access Token }
+ *
+ * The provided authorization grant (e.g., authorization code, resource owner
+ * credentials) or refresh token is invalid, expired, revoked, does not match
+ * the redirection URI used in the authorization request, or was issued to
+ * another client.
+ */
+export class InvalidCredentialsError extends OAuthError {
+  constructor(cause?: unknown) {
+    super('invalid_request', 'Invalid identifier or password', 400, cause)
```
ThisIsMissEm, line 14: Should this be a 401?
```suggestion
    super('invalid_request', 'Invalid identifier or password', 401, cause)
```

packages/oauth/oauth-provider/src/errors/invalid-credentials-error.ts
```diff
@@ -0,0 +1,20 @@
+import { OAuthError } from './oauth-error.js'
+
+/**
+ * @see
+ * {@link https://datatracker.ietf.org/doc/html/rfc6749#section-5.2 | RFC6749 - Issuing an Access Token }
+ *
+ * The provided authorization grant (e.g., authorization code, resource owner
+ * credentials) or refresh token is invalid, expired, revoked, does not match
+ * the redirection URI used in the authorization request, or was issued to
+ * another client.
```
ThisIsMissEm: This comment is wrong, need to think of a better comment.

packages/pds/tests/oauth.test.ts
```diff
@@ -319,6 +359,70 @@ describe('oauth', () => {
     // TODO: Find out why we can't use "using" here
     await page[Symbol.asyncDispose]()
   })
+
+  it('Allows to sign-in through OAuth when Email OTP is required', async () => {
```
ThisIsMissEm, line 390: This currently tests the happy-path, i.e., entering the correct Email OTP.

Should I add entering an incorrect OTP? It'd make the test longer and more complex, but would cover all the cases here.

packages/pds/src/account-manager/oauth-store.ts
```diff
@@ -181,6 +181,13 @@ export class OAuthStore
           const account = await this.accountManager.getAccount(did)
           if (!account) throw new Error('Account not found')
 
+          if (account.email && !account.emailConfirmedAt) {
+            await this.accountManager.sendConfirmEmail({
+              did,
+              email: account.email,
+            })
+          }
+
```
ThisIsMissEm, line 190: The rationale for adding this here is that accounts created via XRPC calls likely will want more fine-grained control over when the email confirmation is sent, where as via open registration, you probably want to always send confirmation.

packages/pds/src/mailer/templates/confirm-signin.hbs
ThisIsMissEm: There is the `/.well-known/change-password` endpoint on the PDS, but I'm not sure if it's really the right URL to use here? Might depend on if there's an entryway or not?

packages/pds/src/account-manager/account-manager.ts
```diff
@@ -387,11 +396,29 @@ export class AccountManager {
...
+        }
+      }
+
+      if (user.email && user.emailConfirmedAt) {
+        if (!emailOtp) {
+          await this.sendConfirmSignin({ did: user.did, email: user.email })
+
+          throw new SecondAuthenticationFactorRequiredError(
+            'emailOtp',
+            user.email,
```
ThisIsMissEm: The email here needs masking, I know the entryway code has something for this (but it's closed-source), so it might be better to reuse that than me reinventing the logic here and missing some edge-case.

So in the response we wouldn't leak the users' email address, so `thisismissem@noreply.users.github.com` would become like `th**********@no******************.com` or something.

packages/oauth/oauth-provider/src/errors/invalid-credentials-error.ts
ThisIsMissEm:
```suggestion
 * @see
 * {@link https://github.com/bluesky-social/atproto/blob/main/packages/oauth/oauth-provider-frontend/src/routes/account/_minimalLayout/sign-in.tsx#L123}
 *
 * The credentials supplied are invalid to complete authentication. This occurs when either
 * the identifier (email or handle), password, or email-based OTP code are incorrect. In the
 * case of invalid email-based OTP code, we do not inform the requester what actually
 * caused the error, as to prevent revealing information.
```

packages/pds/src/account-manager/account-manager.ts
```diff
@@ -387,11 +396,29 @@ export class AccountManager {
...
+          throw new InvalidCredentialsError()
         }
         appPassword = await this.verifyAppPassword(user.did, password)
         if (appPassword === null) {
-          throw new AuthRequiredError('Invalid identifier or password')
+          throw new InvalidCredentialsError()
+        }
+      }
+
+      if (user.email && user.emailConfirmedAt) {
```
ThisIsMissEm, line 407: I think this would be the change to ensure that we only enforce email OTP for regular passwords but not app passwords, which in my testing appears to be the way this is implemented by Bluesky PBC in the entryway.

```suggestion
      if (validAccountPass && user.email && user.emailConfirmedAt) {
```

packages/pds/src/account-manager/account-manager.ts
ThisIsMissEm: I was looking for `obfuscateEmail` from `@atproto/common`

packages/pds/src/mailer/templates/confirm-signin.hbs
ThisIsMissEm: Probably can't use that, as PDSes behind an entryway don't have that endpoint, e.g., https://puffball.us-east.host.bsky.network/.well-known/change-password

Instead for those the URL is: https://bsky.social/.well-known/change-password

ThisIsMissEm:
@devinivy with the spike in compromised accounts lately, it'd be very good to lamd 2FA in self-hosted PDSes without using pds gatekeeper.

More than happy to rebase this PR and make adjustments to ensure it works with the entryway setup too

ThisIsMissEm:
> Is this ever going to get merged, or is this a dead PR?

@spookalucca I can now say this PR is dead, but it lives on as #5207 😁

PR #4223, opened 2025-09-23
Async CAR fixes

dholms:
Parsing large CAR files was being done synchronously and could jam up the entire thread.

There's two fixes here:
- If a serialized CAR file is passed in (rather than a streaming CAR file), we don't try to force it into a streaming interface but instead just keep an index on the byte array
- as we iterate over blocks, we yield to the event loop every 

A quick benchmark shows that for reading a car ~100MB CAR file consisting of 10k 10kb blocks:
- the previous implementation took 653 ms and never yielded to the event loop
- this implementation took 602 ms and yielded to the event loop 193 times

### DIFF
--- a/.github/workflows/build-and-push-pds-ghcr.yaml
+++ b/.github/workflows/build-and-push-pds-ghcr.yaml
@@ -3,7 +3,6 @@ on:
   push:
     branches:
       - main
-      - async-car-fixes
 env:
   REGISTRY: ghcr.io
   USERNAME: ${{ github.actor }}


packages/repo/src/car.ts
```diff
@@ -143,18 +152,26 @@ const readCarBlocksIter = (
...
       if (blockSize === null) {
         break
       }
       const blockBytes = await reader.read(blockSize)
       const cid = parseCidFromBytes(blockBytes.subarray(0, 36))
       const bytes = blockBytes.subarray(36)
       yield { cid, bytes }
+
+      // yield to the event loop every 25 blocks
+      // in the case the incoming CAR is synchrnous, this can end up jamming up the thread
```
dholms:
```suggestion
      // in the case the incoming CAR is synchronous, this can end up jamming up the thread
```

packages/repo/src/car.ts
```diff
@@ -143,18 +152,26 @@ const readCarBlocksIter = (
...
       const blockBytes = await reader.read(blockSize)
       const cid = parseCidFromBytes(blockBytes.subarray(0, 36))
       const bytes = blockBytes.subarray(36)
       yield { cid, bytes }
+
+      // yield to the event loop every 25 blocks
+      // in the case the incoming CAR is synchronous, this can end up jamming up the thread
+      blocks++
+      if (blocks % 25 === 0) {
+        await new Promise((resolve) => setImmediate(resolve))
```
devinivy: Can pull this right out of node if you want.
```ts
import { setImmediate } from 'node:timers/promises';
// ...
await setImmediate()
```

dholms: oh man that's nice

darthmaim: You can even do this, which has the same effect, but might make it even more clear what the intent is:

```ts
import { scheduler } from 'node:timers/promises';
// ...
await scheduler.yield()
```

devinivy: Good call out, though I expect we'll want to hold off while it remains experimental.

PR #4230, opened 2025-09-25
PDS: support websocket permessage-deflate

devinivy:
This adds support for opt-in permessage-deflate on the PDS via `PDS_WS_PERMESSAGE_DEFLATE`.  Also adds ability to configure ws server options on xrpc-server.

### DIFF
--- a/packages/xrpc-server/src/server.ts
+++ b/packages/xrpc-server/src/server.ts
@@ -386,9 +386,11 @@ export class Server {
     const authVerifier = this.createAuthVerifier(cfg)
 
     const { handler } = cfg
+    const { ws } = this.options
     this.subscriptions.set(
       nsid,
       new XrpcStreamServer({
+        ...ws,
         noServer: true,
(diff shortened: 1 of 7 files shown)


packages/xrpc-server/src/server.ts
```diff
@@ -386,9 +386,11 @@ export class Server {
     const authVerifier = this.createAuthVerifier(cfg)
 
     const { handler } = cfg
+    const { ws } = this.options
     this.subscriptions.set(
       nsid,
       new XrpcStreamServer({
```
ThisIsMissEm, line 392: would it be worth changing the typing of the argument here? It's currently:
```ts
  constructor(opts: ServerOptions & { handler: Handler }) {
```

https://github.com/bluesky-social/atproto/blob/b5afb723be392d236799bbcb6a55956bd12316ba/packages/xrpc-server/src/stream/server.ts#L9

PR #4241, opened 2025-09-30
Ozone delayed action

foysalit:
This PR adds a new module to ozone for scheduling account takedowns. When takedowns are scheduled or a scheduled takedown is cancelled, 2 new and different events are logged in the mod event stream for the subject.

### DIFF
--- a/packages/ozone/src/api/moderation/cancelScheduledActions.ts
+++ b/packages/ozone/src/api/moderation/cancelScheduledActions.ts
@@ -2,6 +2,7 @@ import { AuthRequiredError } from '@atproto/xrpc-server'
 import { AppContext } from '../../context'
 import { Server } from '../../lexicon'
 import { subjectFromInput } from '../../mod-service/subject'
+import { ScheduledTakedownTag } from './util'
 
 export default function (server: Server, ctx: AppContext) {
   server.tools.ozone.moderation.cancelScheduledActions({
@@ -9,7 +10,7 @@ export default function (server: Server, ctx: AppContext) {
     handler: async ({ input, auth }) => {
       const access = auth.credentials
       const db = ctx.db
-      const { subjects } = input.body
+      const { subjects, comment } = input.body
 
       if (!access.isModerator) {
         throw new AuthRequiredError(
@@ -32,6 +33,20 @@ export default function (server: Server, ctx: AppContext) {
           await modService.logEvent({
             event: {
               $type: 'tools.ozone.moderation.defs#cancelScheduledTakedownEvent',
+              comment,
+            },
+            subject: subjectFromInput({
+              did: subject,
+              $type: 'com.atproto.admin.defs#repoRef',
+            }),
+            createdBy,
+            createdAt: now,
+          })
+          await modService.logEvent({
+            event: {
+              $type: 'tools.ozone.moderation.defs#modEventTag',
+              remove: [ScheduledTakedownTag],
+              add: [],
             },
             subject: subjectFromInput({
               did: subject,
--- a/packages/ozone/src/api/moderation/scheduleAction.ts
+++ b/packages/ozone/src/api/moderation/scheduleAction.ts
@@ -5,6 +5,7 @@ import { Server } from '../../lexicon'
 import { subjectFromInput } from '../../mod-service/subject'
 import { ExecutionSchedule } from '../../scheduled-action/types'
 import { getScheduledActionType } from '../util'
+import { ScheduledTakedownTag } from './util'
 
 export default function (server: Server, ctx: AppContext) {
   server.tools.ozone.moderation.scheduleAction({
@@ -80,6 +81,19 @@ export default function (server: Server, ctx: AppContext) {
                 createdBy: actualCreatedBy,
                 createdAt: now,
                 modTool,
+              })
+              await modService.logEvent({
+                event: {
+                  $type: 'tools.ozone.moderation.defs#modEventTag',
+                  add: [ScheduledTakedownTag],
+                  remove: [],
+                },
+                subject: subjectFromInput({
+                  did: subject,
+                  $type: 'com.atproto.admin.defs#repoRef',
+                }),
+                createdBy,
+                createdAt: now,
               })
             }
             succeeded.push(subject)
--- a/packages/ozone/src/api/moderation/util.ts
+++ b/packages/ozone/src/api/moderation/util.ts
@@ -92,3 +92,5 @@ export const assertProtectedTagAction = ({
     }
   })
 }
+
+export const ScheduledTakedownTag = 'scheduled-takedown'
--- a/lexicons/tools/ozone/moderation/cancelScheduledActions.json
+++ b/lexicons/tools/ozone/moderation/cancelScheduledActions.json
@@ -19,6 +19,10 @@
                 "format": "did"
               },
               "description": "Array of DID subjects to cancel scheduled actions for"
+            },
+            "comment": {
+              "type": "string",
+              "description": "Optional comment describing the reason for cancellation"
             }
           }
         }
--- a/packages/api/src/client/lexicons.ts
+++ b/packages/api/src/client/lexicons.ts
@@ -14309,6 +14309,11 @@ export const schemaDict = {
                 description:
                   'Array of DID subjects to cancel scheduled actions for',
               },
+              comment: {
+                type: 'string',
+                description:
+                  'Optional comment describing the reason for cancellation',
+              },
             },
           },
         },
--- a/packages/api/src/client/types/tools/ozone/moderation/cancelScheduledActions.ts
+++ b/packages/api/src/client/types/tools/ozone/moderation/cancelScheduledActions.ts
@@ -20,6 +20,8 @@ export type QueryParams = {}
 export interface InputSchema {
   /** Array of DID subjects to cancel scheduled actions for */
   subjects: string[]
+  /** Optional comment describing the reason for cancellation */
+  comment?: string
 }
 
 export type OutputSchema = CancellationResults
--- a/packages/ozone/src/lexicon/lexicons.ts
+++ b/packages/ozone/src/lexicon/lexicons.ts
@@ -14309,6 +14309,11 @@ export const schemaDict = {
                 description:
                   'Array of DID subjects to cancel scheduled actions for',
               },
+              comment: {
+                type: 'string',
+                description:
+                  'Optional comment describing the reason for cancellation',
+              },
             },
           },
         },
--- a/packages/ozone/src/lexicon/types/tools/ozone/moderation/cancelScheduledActions.ts
+++ b/packages/ozone/src/lexicon/types/tools/ozone/moderation/cancelScheduledActions.ts
@@ -19,6 +19,8 @@ export type QueryParams = {}
 export interface InputSchema {
   /** Array of DID subjects to cancel scheduled actions for */
   subjects: string[]
+  /** Optional comment describing the reason for cancellation */
+  comment?: string
 }
 
 export type OutputSchema = CancellationResults
--- a/packages/pds/src/lexicon/lexicons.ts
+++ b/packages/pds/src/lexicon/lexicons.ts
@@ -14309,6 +14309,11 @@ export const schemaDict = {
                 description:
                   'Array of DID subjects to cancel scheduled actions for',
               },
+              comment: {
+                type: 'string',
+                description:
+                  'Optional comment describing the reason for cancellation',
+              },
             },
           },
         },
--- a/packages/pds/src/lexicon/types/tools/ozone/moderation/cancelScheduledActions.ts
+++ b/packages/pds/src/lexicon/types/tools/ozone/moderation/cancelScheduledActions.ts
@@ -19,6 +19,8 @@ export type QueryParams = {}
 export interface InputSchema {
   /** Array of DID subjects to cancel scheduled actions for */
   subjects: string[]
+  /** Optional comment describing the reason for cancellation */
+  comment?: string
 }
 
 export type OutputSchema = CancellationResults


packages/ozone/src/api/moderation/cancelScheduledActions.ts
```diff
@@ -0,0 +1,57 @@
...
+
+      const result = await db.transaction(async (tx) => {
+        const scheduledActionService = ctx.scheduledActionService(tx)
+        const modService = ctx.modService(tx)
+
+        const cancellations =
+          await scheduledActionService.cancelScheduledActions(subjects)
+
+        for (const subject of cancellations.succeeded) {
+          await modService.logEvent({
```
haileyok, line 33: just a question since i've been thinking about how to get kafka in here. is `logEvent` how we handle event emissions or is that something else?

haileyok (approved):
this seems pretty solid to me

lexicons/tools/ozone/moderation/listScheduledActions.json
```diff
@@ -0,0 +1,76 @@
...
+            "startTime": {
+              "type": "string",
+              "format": "datetime",
+              "description": "Filter actions scheduled to execute after this time"
+            },
+            "endTime": {
+              "type": "string",
+              "format": "datetime",
+              "description": "Filter actions scheduled to execute before this time"
+            },
```
devinivy: Just a small consistency thing—for times and time ranges I've seen us use the "at", "before", and "after" suffixes, but I haven't seen "time" used as a suffix.  E.g. `queryEvents` uses `createdAfter` and `createdBefore`.  For consistency should we consider something like `startsAfter` and `endsBefore`?

packages/ozone/src/api/moderation/cancelScheduledActions.ts
devinivy: Two things need to happen for every event (typically dispatched via `tools.ozone.moderation.emitEvent`):
1. They need to take effect, e.g. a label event would end up creating a label on the subject.
2. They need to be logged, e.g. for use in `tools.ozone.moderation.queryEvents`.

The `logEvent()` method here just does no. 2: it logs the event but has no effect.

packages/ozone/src/api/moderation/scheduleAction.ts
```diff
@@ -0,0 +1,118 @@
...
+          'Must specify createdBy when using admin auth',
+        )
+      }
+
+      const actionType = getScheduledActionType(
+        action.$type?.split('#')[1] || '',
+      )
+
+      const succeeded: string[] = []
+      const failed: { subject: string; error: string; errorCode?: string }[] =
```
devinivy: This type could also be imported from the lexicons directory.

packages/ozone/src/api/moderation/cancelScheduledActions.ts
foysalit: yeah logEvent is what registers the event in ozone db so if we were to fan it out to somewhere else, this may be the right place. however,  we probably dont want to bake that into the code here and rather build a separate service to consume the event stream and push it from there?

packages/ozone/src/api/moderation/scheduleAction.ts
```diff
@@ -0,0 +1,118 @@
...
+            executeAfter: new Date(scheduling.executeAfter!),
+            executeUntil: scheduling.executeUntil
+              ? new Date(scheduling.executeUntil)
+              : undefined,
+          }
+
+      const eventData = { ...action, modTool }
+      const actualCreatedBy =
+        access.type === 'admin_token'
+          ? createdBy || ctx.cfg.service.did
```
devinivy: I think `|| ctx.cfg.service.did` is a dead codepath due to the validation above.  Maybe just worth noting/reflecting that if it's not easy to eliminate the codepath.

packages/ozone/src/api/moderation/util.ts
```diff
@@ -0,0 +1,94 @@
+import { InvalidRequestError } from '@atproto/xrpc-server'
```
devinivy, line 1: Didn't review this thoroughly, at a glance seems it's identical to the code moved from `packages/ozone/src/api/moderation/emitEvent.ts`.  If there are any specific changes to review just call them out and I'll take a closer look.

packages/ozone/src/api/util.ts
```diff
@@ -211,3 +213,34 @@ const safelinkPatterns = new Set(['domain', 'url'])
...
+  | 'cancelled'
+  | 'failed'
+
+const scheduledActionTypes = new Set(['takedown'])
+const scheduledActionStatuses = new Set([
+  'pending',
+  'executed',
+  'cancelled',
+  'failed',
+])
```
devinivy, line 246: Totally up to you, but could do something like this to minimize the duplication if you like:
```ts
type MembersOfSet<T extends Set<unknown>> = T extends Set<infer U> ? U : never

const scheduledActionTypes = new Set([
  'pending',
  'executed',
  'cancelled',
  'failed',
] as const)

type ScheduledActionTypes = MembersOfSet<typeof scheduledActionTypes>
```

packages/ozone/src/db/migrations/20250923T000000000Z-scheduled-actions.ts
```diff
@@ -0,0 +1,62 @@
...
+  await db.schema
+    .createTable('scheduled_action')
+    .addColumn('id', 'bigserial', (col) => col.primaryKey())
+    .addColumn('action', 'varchar', (col) => col.notNull())
+    .addColumn('eventData', 'jsonb')
+    .addColumn('did', 'varchar', (col) => col.notNull())
+    .addColumn('executeAt', 'timestamptz')
+    .addColumn('executeAfter', 'timestamptz')
+    .addColumn('executeUntil', 'timestamptz')
+    .addColumn('randomizeExecution', 'boolean', (col) => col.defaultTo(false))
```
devinivy: I forget, do you still need to use `notNull()` to ensure non-nullable even when using `defaultTo()`?  I think you might.

devinivy: If so, there are a few more in here.

packages/ozone/src/db/migrations/20250923T000000000Z-scheduled-actions.ts
```diff
@@ -0,0 +1,62 @@
...
+    .addColumn('executionEventId', 'bigint')
+    .execute()
+
+  // Unique constraint to prevent multiple pending actions for the same subject
+  await db.schema
+    .createIndex('scheduled_action_unique_pending_subject')
+    .unique()
+    .on('scheduled_action')
+    .columns(['did', 'action', 'status'])
+    .execute()
```
devinivy: I think this may prevent multiple actions of the same type being marked as `'executed'`.  I think the intention is just to avoid multiple `'pending'`, which might be something that either needs to be remodeled or enforcement moved up into the application.

packages/ozone/src/scheduled-action/service.ts
```diff
@@ -0,0 +1,276 @@
...
+          failed.push({
+            did,
+            error: 'No pending scheduled actions found for subject',
+            errorCode: 'NoPendingActions',
+          })
+        }
+      } catch (error) {
+        failed.push({
+          did,
+          error: error instanceof Error ? error.message : 'Unknown error',
```
devinivy: For unexpected operational errors it's typically best practice not to expose them through the API since they can leak sensitive details.

packages/ozone/src/scheduled-action/service.ts
```diff
@@ -0,0 +1,276 @@
...
+    actionId: number,
+    executionEventId: number,
+  ): Promise<void> {
+    await this.db.db
+      .updateTable('scheduled_action')
+      .set({
+        status: 'executed',
+        lastExecutedAt: new Date().toISOString(),
+        executionEventId,
+        updatedAt: new Date().toISOString(),
```
devinivy: Would recommend using a single date so that these two times are logged identically.

packages/ozone/src/scheduled-action/service.ts
```diff
@@ -0,0 +1,276 @@
...
+    actionId: number,
+    failureReason: string,
+  ): Promise<void> {
+    await this.db.db
+      .updateTable('scheduled_action')
+      .set({
+        status: 'failed',
+        lastExecutedAt: new Date().toISOString(),
+        lastFailureReason: failureReason,
+        updatedAt: new Date().toISOString(),
```
devinivy: Would recommend using a single date so that these two times are logged identically.

packages/ozone/src/db/migrations/20250923T000000000Z-scheduled-actions.ts
```diff
@@ -0,0 +1,62 @@
...
+    .on('scheduled_action')
+    .column('did')
+    .execute()
+
+  // for querying actions by status
+  await db.schema
+    .createIndex('scheduled_action_status_idx')
+    .on('scheduled_action')
+    .column('status')
+    .execute()
```
devinivy: Just a note that this will only be useful for low-cardinality statuses, largely because the result sets are still paginated by another column.  So it will need to enumerate the full result set, then sort and paginate.

packages/ozone/src/daemon/scheduled-action-processor.ts
```diff
@@ -0,0 +1,243 @@
...
+
+  poll() {
+    if (this.destroyed) return
+    this.processingPromise = this.findAndExecuteScheduledActions()
+      .catch((err) =>
+        dbLogger.error({ err }, 'scheduled action processing errored'),
+      )
+      .finally(() => {
+        this.timer = setTimeout(() => this.poll(), getInterval())
+      })
```
devinivy, line 45: This will set the timer again even if `ScheduledActionProcessor` has been destroyed.

packages/ozone/src/daemon/scheduled-action-processor.ts
```diff
@@ -0,0 +1,243 @@
...
+    modTool,
+    moderationTxn,
+    settingService,
+  }: {
+    action: Selectable<ScheduledAction>
+    event: ModEventType
+    modTool: ModTool | undefined
+
+    moderationTxn: ModerationService
+    settingService: SettingService
```
devinivy, line 156: Making queries outside of the transaction during the transaction can create deadlocks with the connection pool—I'd recommend ensuring `settingService` is also bound to this transaction.

packages/ozone/src/api/moderation/util.ts
foysalit: yeah this is basically the same code from that file, putting it here to indicate it's shared utility. nothing to review here IMO

packages/ozone/src/api/moderation/scheduleAction.ts
foysalit: right yeah, we do validate this above so the default is not necessary here.

packages/ozone/src/api/util.ts
foysalit: we do have a getter up above which abstracts the typing so i'll leave it as is for now. neat tho!

packages/ozone/src/db/migrations/20250923T000000000Z-scheduled-actions.ts
foysalit: right yeah we have the validation in the application layer but want to have something in the db layer too. ended up using this 

```
  await sql`
    CREATE UNIQUE INDEX scheduled_action_unique_pending_subject
    ON scheduled_action (did, action)
    WHERE status = 'pending'
  `.execute(db)
```

packages/ozone/src/db/migrations/20250923T000000000Z-scheduled-actions.ts
foysalit: thanks. im gonna make the status property required on listScheduledAction endpoint and then combine the status and createdAt columns to create a compound index.

packages/ozone/src/db/migrations/20250923T000000000Z-scheduled-actions.ts
foysalit: i think we need both!

packages/ozone/src/daemon/scheduled-action-processor.ts
foysalit: hmm... i copied this from the event-reverser worker https://github.com/bluesky-social/atproto/blob/main/packages/ozone/src/daemon/event-reverser.ts so didn't pay too much attention.
even if it does, wouldn't it just poll 1 extra time and then end it because of the `if (this.destroyed) return` check?

packages/ozone/src/db/migrations/20250923T000000000Z-scheduled-actions.ts
```diff
@@ -0,0 +1,56 @@
...
+export async function up(db: Kysely<unknown>): Promise<void> {
+  await db.schema
+    .createTable('scheduled_action')
+    .addColumn('id', 'bigserial', (col) => col.primaryKey())
+    .addColumn('action', 'varchar', (col) => col.notNull())
+    .addColumn('eventData', 'jsonb')
+    .addColumn('did', 'varchar', (col) => col.notNull())
+    .addColumn('executeAt', 'varchar')
+    .addColumn('executeAfter', 'varchar')
+    .addColumn('executeUntil', 'varchar')
```
ThisIsMissEm, line 12: The database here is postgresql, wouldn't it make sense to use postgresql native date/time types? (`varchar` would make sense for sqlite)

foysalit: this is mostly just following precedent and yeah, native timestamps would work just fine but we aren't missing anything, feature-wise, by using varchar, at least for our usecases.

PR #4249, opened 2025-10-03
Add missing `#reasonCivicOther` report reason

estrattonbailey:
Looks like this slipped through #3881

Related: https://github.com/bluesky-social/atproto/pull/4251


estrattonbailey:
Hold off on this for now, couple other changes incoming

PR #4251, opened 2025-10-03
Match report reason mapping to RFC-0009

estrattonbailey:
In [RFC-0009](https://github.com/bluesky-social/proposals/tree/main/0009-mod-report-granularity#backwards-compatibility), we recommended mapping `#reasonRuleOther` to `#reasonOther`. In order for this to function correctly, I think we should mirror this relationship in the opposite direction.

This way, if a labeler only supports `#reasonOther`, the user is still able to submit reports to that labeler using `#reasonOther`.

### DIFF
--- a/.changeset/young-cherries-worry.md
+++ b/.changeset/young-cherries-worry.md
@@ -1,0 +1,5 @@
+---
+"@atproto/ozone": patch
+---
+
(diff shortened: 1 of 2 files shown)


estrattonbailey:
Superseded by #4262

estrattonbailey:
Superseded by https://github.com/bluesky-social/atproto/pull/4262

PR #4255, opened 2025-10-06
Feat: asymmetrical keys for PDS JWT signing

ThisIsMissEm:
I believe this is the code that's required for enabling asymmetrical keys for the JWTs within the PDS codebase.

Took me a while to work this out, but I think this would enable the work to have an entryway in the codebase, replacing the `@atproto/pds-entryway` package (which is really just `@atproto/pds@0.3.0-entryway.3`)

We'd obviously still need the code for an entryway itself in the future, but I'm trying to lay the ground work for that.


ThisIsMissEm:
The tests are being flaky here, but not because of my changes.

packages/pds/src/config/env.ts
```diff
@@ -128,6 +128,9 @@ export const readEnv = (): ServerEnvironment => {
     // secrets
     dpopSecret: envStr('PDS_DPOP_SECRET'),
     jwtSecret: envStr('PDS_JWT_SECRET'),
+    jwtSigningKeyK256PrivateKeyHex: envStr(
+      'PDS_JWT_SIGNING_KEY_K256_PRIVATE_KEY_HEX',
```
ThisIsMissEm, line 132: There's perhaps a need for extra documentation around this, basically you'd only set this when building any entryway or when using an entryway — otherwise the OAuthProvider would be verifying the JWT coming in with the private key, when it should probably use the public key (which is what `entrywayJwtVerifyKeyK256PublicKeyHex` gives you)

packages/pds/tests/jwt-signing-key.test.ts
```diff
@@ -0,0 +1,94 @@
...
+import { TestPds, TestPlc, mockNetworkUtilities } from '@atproto/dev-env'
+import { createPublicKeyObject } from '../'
+
+const getPublicHex = (key: Secp256k1Keypair) => {
+  return key.publicKeyStr('hex')
+}
+
+const getPrivateHex = async (key: Secp256k1Keypair) => {
+  return ui8.toString(await key.export(), 'hex')
+}
```
ThisIsMissEm, line 15: It may be an idea to have a package somewhere for these methods, since similar code is repeated all over the place:
- https://github.com/bluesky-social/atproto/blob/30f9b6690e0e2c5810772e94e631322b9d89c65a/packages/dev-env/src/pds.ts#L22
- https://github.com/bluesky-social/atproto/blob/61dc0d60e19b88c6427a54c6d95a391b5f4da7bd/packages/pds/tests/entryway.test.ts#L221
- https://github.com/bluesky-social/atproto/blob/61dc0d60e19b88c6427a54c6d95a391b5f4da7bd/packages/bsky/tests/entryway-auth.test.ts#L13

There's probably more places, too.

ThisIsMissEm:
I think I'm going to close this, since this was really just a shortcut to having an open source entryway implementation, as it meant not reimplementing all the account management code from the PDS into another package within the repository. This isn't actually necessary for using an entryway, just for using the PDS implementation as a basis for an entryway, per https://github.com/ThisIsMissEm/bluesky-social-atproto/pull/1 

Was a fun learning experience. Maybe I'll build an entryway implementation in time using what I learned here.

PR #4261, opened 2025-10-07
:zap: Add better indexes for searching ozone events by comment keyword and batch id

foysalit:


packages/ozone/src/db/migrations/20251007T000000000Z-mod-event-batch-id-comment-idx.ts
```diff
@@ -0,0 +1,38 @@
...
+  )
+  // create the new optimized index for batch id
+  await sql`
+    CREATE INDEX "moderation_event_batch_created_id_desc_idx"
+    ON "moderation_event" ( ("modTool"->'meta'->>'batchId'), "createdAt" DESC, "id" DESC )
+    WHERE ("modTool"->'meta'->>'batchId') IS NOT NULL
+  `.execute(db)
+
+  // Add the trgm extension first
+  await sql`CREATE EXTENSION IF NOT EXISTS pg_trgm`.execute(db)
```
devinivy, line 25: Just a note that I think this _may_ require superuser access on the database.  So would have to check how it will play with the ozone distro and other deployments.

packages/ozone/src/db/migrations/20251007T000000000Z-mod-event-batch-id-comment-idx.ts
```diff
@@ -0,0 +1,38 @@
...
+  // Remove the old index for batch id that's not very optimized
+  await sql`DROP INDEX IF EXISTS "moderation_event_mod_tool_batch_id_idx"`.execute(
+    db,
+  )
+  // create the new optimized index for batch id
+  await sql`
+    CREATE INDEX "moderation_event_batch_created_id_desc_idx"
+    ON "moderation_event" ( ("modTool"->'meta'->>'batchId'), "createdAt" DESC, "id" DESC )
+    WHERE ("modTool"->'meta'->>'batchId') IS NOT NULL
+  `.execute(db)
```
devinivy: Probably no big deal, but this will create a period where there's no index coverage at all.  If that's an issue it could be addressed by creating the new index before dropping the old one.

packages/ozone/src/db/migrations/20251007T000000000Z-mod-event-batch-id-comment-idx.ts
```diff
@@ -0,0 +1,38 @@
...
+    db,
+  )
+  await sql`
+    CREATE INDEX "moderation_event_mod_tool_batch_id_idx"
+    ON "moderation_event" ( ("modTool" -> 'meta' ->> 'batchId') )
+    WHERE "modTool" #> '{meta,batchId}' IS NOT NULL
+  `.execute(db)
+
+  await sql`DROP INDEX IF EXISTS "moderation_event_comment_trigram"`.execute(db)
+  await sql`DROP IF EXISTS pg_trgm CASCADE`.execute(db)
```
devinivy:
```suggestion
  await sql`DROP EXTENSION IF EXISTS pg_trgm CASCADE`.execute(db)
```

devinivy (approved):

PR #4262, opened 2025-10-07
Finalize report reasons

estrattonbailey:
This PR introduces more changes in addition to those outlined in [RFC0009](https://github.com/bluesky-social/proposals/tree/main/0009-mod-report-granularity#backwards-compatibility). An addendum will be added to that RFC, and an updated will be posted to the [discussion thread](https://github.com/bluesky-social/atproto/discussions/3880).

This marks the final version of these report reason lexicons. We will be implementing them in the Bluesky app shortly hereafter.


foysalit (approved):
looks good!

PR #4265, opened 2025-10-08
Add account data in pwd reset hooks

matthieusieben:


packages/pds/src/account-manager/helpers/email-token.ts
```diff
@@ -71,7 +71,7 @@ export const assertValidTokenAndFindDid = async (
 ): Promise<string> => {
   const res = await db.db
     .selectFrom('email_token')
-    .selectAll()
+    .select(['did', 'requestedAt'])
```
matthieusieben, line 74: Unrelated optimization

PR #4273, opened 2025-10-11
update links to ietf docs

aaronpk:
* Update links for Client ID Metadata Documents from drafts.aaronpk.com to ietf.org
* Update links for Protected Resource Metadata from -05 to RFC9728


bnewbold (approved):
thanks aaron!

PR #4279, opened 2025-10-17
:sparkles: Add account level strike count in ozone

foysalit:
This PR introduces a numeric strike count for accounts based on infraction of policy and severity level. Policies and severity levels are introduced as instance-wide ozone options/settings which is not tightly coupled with event level association of policy, severity level and strike count, which by design, allows moderators/ozone client to flexibly apply strikes to accounts through actions.

### DIFF
--- a/.changeset/three-baboons-knock.md
+++ b/.changeset/three-baboons-knock.md
@@ -1,0 +1,8 @@
+---
+"@atproto/dev-env": patch
+"@atproto/ozone": patch
+"@atproto/api": patch
+"@atproto/pds": patch
+---
+
+Add strike system to ozone


lexicons/tools/ozone/moderation/defs.json
```diff
@@ -595,6 +637,25 @@
...
+        },
+        "policies": {
+          "type": "array",
+          "maxLength": 5,
+          "items": { "type": "string" },
+          "description": "Names/Keywords of the policies that necessitated the email."
+        },
+        "severityLevel": {
+          "type": "string",
+          "description": "Severity level of the violation. normally 'sev-1' that adds strike on repeat offense"
```
haileyok:
```suggestion
          "description": "Severity level of the violation. Normally 'sev-1' that adds strike on repeat offense"
```

lexicons/tools/ozone/moderation/defs.json
```diff
@@ -595,6 +637,25 @@
...
+        },
+        "strikeCount": {
+          "type": "integer",
+          "description": "Number of strikes to assign to the user for this violation. Normally 0 as indicator of warning and only added as strike on repeat offense."
```
haileyok:
```suggestion
          "description": "Number of strikes to assign to the user for this violation. Normally 0 as an indicator of a warning and only added as a strike on a repeat offense."
```

packages/ozone/src/db/migrations/20251008T120000000Z-add-strike-system.ts
```diff
@@ -0,0 +1,87 @@
...
+  // This supports fast look up for background job that aggregates strike data per subjectDid
+  await db.schema
+    .createIndex('moderation_event_subject_did_strike_count_idx')
+    .on('moderation_event')
+    .columns(['subjectDid', 'strikeCount'])
+    .execute()
+
+  // This supports fast lookup in the background job that needs to find strikes that have expired
+  await sql`
+    CREATE INDEX moderation_event_strike_expires_at_strike_countidx
```
haileyok: `count_idx`?

packages/ozone/src/mod-service/index.ts
```diff
@@ -609,6 +650,19 @@ export class ModerationService {
       subject.blobCids,
     )
 
+    // Process strikes if strikeCount is set - update the account_strike table
+    if (modEvent.strikeCount !== null && modEvent.strikeCount !== undefined) {
```
haileyok: just a small nit, this can be `modEvent.strikeCount != null`

foysalit: yh looks like this can be cleaned up even more

packages/ozone/src/db/migrations/20251008T120000000Z-add-strike-system.ts
```diff
@@ -0,0 +1,87 @@
...
+    WHERE "strikeExpiresAt" IS NOT NULL AND "strikeCount" IS NOT NULL
+  `.execute(db)
+
+  // for sorting and filtering by active strike count
+  await db.schema
+    .createIndex('account_strike_active_count_idx')
+    .on('account_strike')
+    .column('activeStrikeCount')
+    .execute()
+}
```
haileyok, line 59: if we are removing the `minStrikes` option from the query (which i think we are but i could have misread that) i might suggest we make this a partial index with like
```
  .where('activeStrikeCount', '>', 0)
```

foysalit: we did remove the `minStrikeCount` param but only for queryEvents endpoint, not for the queryStatuses endpoint. i dont think the partial index helps here since we will only have rows in this table when there is a strike count to be added. granted, the `activeStrikeCount` may go to zero but we would still want those to be indexed for sorting purposes.

PR #4282, opened 2025-10-17
atproto/lexicon:Allow unions as definitions

wdantuma:
This PR adds union as a type allowed in definitions so it behaves the same as the indigo implementation does ( goat )

### DIFF
--- a/packages/lexicon/tests/general.test.ts
+++ b/packages/lexicon/tests/general.test.ts
@@ -478,10 +478,10 @@ describe('Record validation', () => {
       defs: {
         main: {
           type: 'union',
-          refs: ['object']
+          refs: ['type1']
         },
         type1: {
-          object: 'object',
+          type: 'object',
           properties: {
             string: {
               type: 'string'


wdantuma:
I understand that is a problem, and disallowing unions for main would not be a problem at all, actually I will update the PR for it, but not being able to reuse a predefined union is ( for me ) a big issue for lexicon reuse, I ( and I guess more people )  have a lot of use cases with large unions (more than 10 types), and always needing to copy their definitions in types that use the union would be very developer unfriendly.

wdantuma:
Added test to check if union as main definition is still invalid

wdantuma:
@bnewbold, @devinivy Is this a correct fix for this issue ?

bnewbold:
We discussed this a bunch and we are pretty sure we want to disallow unions as definitions.

I have a long write-up here: https://github.com/bluesky-social/atproto/discussions/4343

Please reply over there to consolidate discussion.

wdantuma:
Closing this one awaiting the discussion in #4343

PR #4289, opened 2025-10-21
OAuth client validation improvements

matthieusieben:


packages/oauth/oauth-types/src/uri.ts
```diff
@@ -167,12 +159,41 @@ export const privateUseUriSchema = dangerousUriSchema.superRefine(
...
+    // > In addition to the collision-resistant properties, requiring a URI
+    // > scheme based on a domain name that is under the control of the app can
+    // > help to prove ownership in the event of a dispute where two apps claim
+    // > the same private-use URI scheme (where one app is acting maliciously).
+    //
+    // We can't check for ownership here (as there is no concept of proven
+    // ownership in a generic validation logic), but we can check that the
+    // domain is not local.
+
+    const urlDomain = url.protocol.slice(0, -1).split('.').reverse().join('.')
```
ThisIsMissEm: Should this be url.hostname here?

matthieusieben: We're parsing a private-use URI scheme here (`<scheme>:/<path>`) and making sure that the reversed `<scheme>` part is indeed a (non local) domain name.

This means that we need to process (reverse) the url's protocol.

ThisIsMissEm: Ah, okay; in context of the diff and the comment, it gave me a moment of pause. Maybe explicitly give an example in the comment? Might help make it much clearer

packages/oauth/oauth-types/src/uri.ts
```diff
@@ -167,12 +159,41 @@ export const privateUseUriSchema = dangerousUriSchema.superRefine(
...
+    // https://datatracker.ietf.org/doc/html/rfc8252#section-8.4
+    //
+    // > In addition to the collision-resistant properties, requiring a URI
+    // > scheme based on a domain name that is under the control of the app can
+    // > help to prove ownership in the event of a dispute where two apps claim
+    // > the same private-use URI scheme (where one app is acting maliciously).
+    //
+    // We can't check for ownership here (as there is no concept of proven
+    // ownership in a generic validation logic), but we can check that the
+    // domain is not local.
```
devinivy: I think it helps to connect these ideas— local domains are relevant because they can't be controlled/owned by the app.  Section 7.1:
> When choosing a URI scheme to associate with the app, apps MUST use a
   URI scheme based on a domain name under their control, expressed in
   reverse order...

And then our spec:
> Any custom scheme must match the `client_id` hostname in reverse-domain order.

I assume this latter requirement is enforced in the auth server provider.

PR #4290, opened 2025-10-21
Sync tool library

dholms:
Client lib for https://github.com/bluesky-social/indigo/pull/1170

Tap is an all-in-one AT sync tool for backfilling and consuming the firehose.

More details on usage in the included README.

### DIFF
--- a/packages/tap/README.md
+++ b/packages/tap/README.md
@@ -39,6 +39,9 @@ const channel = tap.channel(indexer)
 channel.start()
 
 await tap.addRepos(['did:plc:ewvi7nxzyoun6zhxrhs64oiz'])
+
+// On shutdown
+await channel.destroy()
 ```
 
 ## Running Tap


packages/nexus/src/channel.ts
```diff
@@ -0,0 +1,118 @@
...
+  bufferedAcks: number[] = []
+
+  constructor(url: string, options: NexusChannelOptions = {}) {
+    super()
+    this.url = url
+    this.reconnect = options.reconnect ?? true
+    this.reconnectDelay = options.reconnectDelay ?? 1000
+  }
+
+  ackEvent = (id: number): Promise<boolean> => {
```
devinivy: Any special reason this isn't a typical class method?

devinivy: I see that this method never resolves `false`, but what conditions would cause it in theory?

devinivy: Ah, I see the usage where it's relied on now.  It can mess with inheritance (e.g. no `super.ackEvent` on an extending class), but not a major deal and could always be changed later.

packages/nexus/src/channel.ts
```diff
@@ -0,0 +1,118 @@
...
+        this.ws.send(JSON.stringify({ id }), (err) => {
+          if (err) {
+            reject(err)
+          } else {
+            resolve(true)
+          }
+        })
+      } else {
+        this.bufferedAcks.push(id)
+        resolve(true)
```
devinivy: I think there's a case to be made that this should resolve only once the buffered ack is processed.  I don't think it would be a very big change if we prefer that.

packages/nexus/src/channel.ts
```diff
@@ -0,0 +1,118 @@
...
+    }
+  }
+
+  async connect(): Promise<void> {
+    this.ws = new WebSocket(this.url)
+
+    this.ws.on('open', () => {
+      this.emit('open')
+      this.flushBufferedAcks().catch((err) => {
+        this.emit('error', new Error(`Failed to flush buffered acks: ${err}`))
```
devinivy: This is a nice builtin way to handle that (similar below):
```suggestion
        this.emit('error', new Error('Failed to flush buffered acks', { cause: err }))
```

packages/tap/src/client.ts
```diff
@@ -0,0 +1,50 @@
...
+export class Nexus {
+  url: string
+
+  constructor(url: string) {
+    this.url = url
+  }
+
+  connect(opts?: NexusChannelOptions) {
+    const url = new URL(this.url)
+    url.protocol = url.protocol === 'https:' ? 'wss:' : 'ws:'
```
devinivy, line 38: If the protocol is already `wss:` we'd probably want to preserve that too.

packages/tap/src/client.ts
```diff
@@ -0,0 +1,50 @@
...
+    const response = await fetch(`${this.url}/add-repos`, {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+      },
+      body: JSON.stringify({ dids }),
+    })
+
+    if (!response.ok) {
+      throw new Error(`Failed to add repos: ${response.statusText}`)
```
devinivy, line 55: If the response body isn't drained here it will leak resources in the node environment.  A simple `await response.arrayBuffer()` would do, for example.  (Similar for other `fetch()` calls too.)

packages/nexus/src/client.ts
```diff
@@ -0,0 +1,50 @@
...
+  connect(opts?: NexusChannelOptions) {
+    const url = new URL(this.url)
+    url.protocol = url.protocol === 'https:' ? 'wss:' : 'ws:'
+    url.pathname = '/listen'
+    const channel = new NexusChannel(url.toString(), opts)
+    channel.connect()
+    return channel
+  }
+
+  async addRepos(dids: string[]) {
```
devinivy: I think this will return an `any` type.  It can also be nice for public interfaces to have their full type signatures to avoid unintended breaking changes.

packages/nexus/src/events.ts
```diff
@@ -0,0 +1,124 @@
...
+  user: userEventDataSchema,
+})
+
+export const nexusEvt = z.union([recordEvt, userEvt])
+
+export interface NexusEvent {
+  id: number
+  did: string
+  isRecordEvt(): this is RecordEvent
+  isUserEvt(): this is UserEvent
```
devinivy: Just a style thing, but I'd advocate lightly for `isRecordEvent` and `isUserEvent`.

packages/tap/tsconfig.build.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "../../tsconfig/node.json",
+  "compilerOptions": {
+    "rootDir": "./src",
+    "outDir": "./dist"
+  },
```
matthieusieben, line 7: As this is a new project I suggest we build it using the safer `noImplicitAny` option (I have a PR in limbo to enable it on the entire repo)

```suggestion
  "compilerOptions": {
    "noImplicitAny": true,
    "rootDir": "./src",
    "outDir": "./dist"
  },
```

packages/tap/src/client.ts
dholms: I'm actually gonna restrict the constructor so it only takes http/https

packages/tap/src/channel.ts
```diff
@@ -0,0 +1,141 @@
...
+  private destroyDefer: Deferrable
+
+  private bufferedAcks: BufferedAck[] = []
+
+  constructor(
+    url: string,
+    handler: NexusHandler,
+    wsOpts: NexusWebsocketOptions = {},
+  ) {
+    this.abortController = new AbortController()
```
matthieusieben: Nit: You could init this at the declaration to make things less verbose:

```ts
private readonly abortController = new AbortController()
```

packages/tap/src/client.ts
```diff
@@ -0,0 +1,51 @@
...
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+      },
+      body: JSON.stringify({ dids }),
+    })
+    await response.arrayBuffer()
+
+    if (!response.ok) {
+      await response.arrayBuffer()
```
matthieusieben: Body was already consumed above

```suggestion
```

packages/tap/src/client.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  async removeRepos(dids: string[]): Promise<void> {
+    const response = await fetch(`${this.url}/remove-repos`, {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+      },
+      body: JSON.stringify({ dids }),
+    })
+    await response.arrayBuffer()
```
matthieusieben: Nit: A more resource efficient way to drain would avoid putting it memory (though in practice it doesn't probably matter)

```ts
    await response.body?.cancel()
```

https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream/cancel

> Cancel is used when you've completely finished with the stream and don't need any more data from it, even if there are chunks enqueued waiting to be read.

packages/tap/src/channel.ts
```diff
@@ -0,0 +1,141 @@
...
+      const data = chunk.toString()
+      evt = parseNexusEvent(JSON.parse(data))
+    } catch (err) {
+      this.handler.onError(new Error('Failed to parse message', { cause: err }))
+      return
+    }
+
+    try {
+      await this.handler.onEvent(evt, { signal: this.abortController.signal })
+      await this.ackEvent(evt.id)
```
matthieusieben: Would it make sense for the event handler to actually perform the ack ?

```suggestion
      await this.handler.onEvent(evt, {
        signal: this.abortController.signal,
        ack: async () => this.ackEvent(evt.id)
      })
```

packages/tap/src/channel.ts
```diff
@@ -0,0 +1,141 @@
...
+        return
+      }
+    }
+  }
+
+  async start() {
+    try {
+      for await (const chunk of this.ws) {
+        await this.processWsEvent(chunk)
+      }
```
matthieusieben, line 112: We should probably handle back pressure here, and limit the amount of parallel processing. Otherwise the process running here will spawn thousands of event processing microtasks which would end-up competing for resources (eg. db access, etc.)

packages/tap/src/channel.ts
dholms: Actually I think this is right. `TapHandler` implementation should be in charge of it. So for instance the `SimpleIndexer` calls this, but it doesn't pass it all the way to the dev. unless the handler implementation makes that explicit decision

PR #4294, opened 2025-10-23
Export record types using the `Main` identifier

matthieusieben:


packages/lex-cli/src/codegen/lex-gen.ts
```diff
@@ -520,19 +520,45 @@ export function genRecord(
...
 
-  //= export function isRecord(v: unknown): v is Record {...}
-  genObjHelpers(file, lexUri, 'Record', {
+  //= export function is{X}(v: unknown): v is {X} {...}
+  genObjHelpers(file, lexUri, ifaceName, {
     requireTypeProperty: true,
   })
+
+  // For convenience, we re-export the type and the type guard under the generic
+  // names "Record", "isRecord" and "validateRecord".
```
DavidBuchanan314, line 540: Is this really for convenience, or more for backward compatibility?

matthieusieben: both

.changeset/slimy-berries-yell.md
```diff
@@ -0,0 +1,5 @@
+---
+"@atproto/lex-cli": patch
+---
+
+Export record types using the `Main` identifier
```
devinivy, line 5: We probably also want a patch changeset for the @atproto/api package, as I believe codegen ends up being part of its public interface.

PR #4295, opened 2025-10-23
Use fetch redirect: 'manual' and handle redirects manually for environments that disallow redirect: 'error'

### LINKED ISSUE #3292: OAuth client in CloudFlare environments

### LINKED ISSUE #4052: `@atproto/oauth-client-node` doesn't work on Cloudflare Workers
**Describe the bug**

`@atproto-labs/fetch` breaks in Cloudflare Workers because of `redirect: error`

On Cloudflare, the Request class does not support the following options that are used in AT Protocol packages:
1. `redirect: error`
2. `cache: no-cache`

**To Reproduce**

1. Create a Cloudflare Worker project with wrangler
2. Implement a basic DID resolver using `@atproto-labs/did-resolver`

**Expected behavior**

You should see an error like this:
```
✘ DidError: Invalid redirect value, must be one of "follow" or "manual" ("error" won't be implemented since it does not make sense at the edge; use "manual" and check the response status code).

      at DidError.from
  (.../@atproto+did@0.1.5/node_modules/@atproto/did/src/did-error.ts:41:12)
      at DidResolverCommon.resolve
  (.../@atproto-labs+did-resolver@0.2.0/node_modules/@atproto-labs/did-resolver/src/did-resolver-base.ts:67:22)
  {
    did: 'did:plc:62ubtodv46bn25wzhj65lc75',
    code: 'did-unknown-error',
    status: 400,
    [cause]: TypeError: Invalid redirect value, must be one of "follow" or "manual" ("error" won't
  be implemented since it does not make sense at the edge; use "manual" and check the response
  status code).
```

**Details**

- Operating system: N/A
- Node version: N/A
- Cloudflare Compatibility Date: "2025-07-15"
- Cloudflare Compatibility Flags: [ "nodejs_compat" ]

**Additional context**

tomo-x7:
# Summary
Replace uses of fetch option redirect: 'error' with redirect: 'manual' and add unified handling for redirect responses. This change ensures the SDK works in environments that reject redirect: 'error' (for example, Cloudflare Workers) by detecting and treating redirects explicitly.

# What changed
- All fetch calls outside the oauth-provider now use redirect: 'manual' instead of redirect: 'error'.
- For each modified fetch call, added redirect checks that work across runtimes and throw context-appropriate errors. Specifically:
  - In browser / worker runtimes: redirect responses appear as response.type === 'opaqueredirect' (status may be 0 and headers are not exposed).
  - In Node runtimes: redirect responses appear as 3xx status codes with a Location header.
# Rationale
Some edge runtimes (notably Cloudflare Workers) do not accept redirect: 'error' and will throw a TypeError. Switching to redirect: 'manual' and detecting redirects allows the SDK to behave predictably across browser, Cloudflare Workers, and Node environments.

fix #3292
fix #4052

### DIFF
--- a/.changeset/hot-parents-shake.md
+++ b/.changeset/hot-parents-shake.md
@@ -1,0 +1,10 @@
+---
+"@atproto/identity": patch
+"@atproto-labs/did-resolver": patch
+"@atproto-labs/fetch": patch
+"@atproto-labs/handle-resolver": patch
+"@atproto/oauth-client": patch
+"@atproto/oauth-client-browser": patch
+---
+
+replace redirect:'error' with 'manual'


tomo-x7:
The test is failing, but I can't reproduce it locally. I made sure to fix it carefully to avoid any impact, but is this test failure caused by my fix?

PR #4296, opened 2025-10-23
dev-env: fix flaky handle resolutions in tests

devinivy:
By placing network mocks before data is flowing, we should be able to avoid invalid handles appearing in the tests due to missed handle resolutions.


ThisIsMissEm:
Oooh, this makes total sense now that I look at it! Thanks for catching & fixing this, it's easily been my biggest annoyance when trying to contribute.

PR #4297, opened 2025-10-23
[TS-160] Add `debug` field to `PostView` and `ProfileView*`s

estrattonbailey:
Adds a flexible `debug` field to `PostView`s and `ProfileView*`s. This field is an untyped dictionary, which can serve as a dumping ground for any metadata that might be useful to the development of applications. It's intended to be used as a reference, not used directly by clients, since the fields here are unstable and will change without notice.

### DIFF
--- a/packages/bsky/src/views/index.ts
+++ b/packages/bsky/src/views/index.ts
@@ -367,9 +367,7 @@ export class Views {
       createdAt: actor.createdAt?.toISOString(),
       verification: this.verification(did, state),
       status: this.status(did, state),
-      debug: state.ctx?.includeDebugField
-        ? this.debugField([actor.debug])
-        : undefined,
+      debug: state.ctx?.includeDebugField ? actor.debug : undefined,
     }
   }
 
@@ -949,20 +947,8 @@ export class Views {
         ? this.threadgate(threadgateUri, state)
         : undefined,
       debug: state.ctx?.includeDebugField
-        ? this.debugField([{ post: post.debug }, { author: author.debug }])
+        ? { post: post.debug, author: author.debug }
         : undefined,
-    }
-  }
-
-  debugField<D = { [key: string]: unknown } | undefined>(debugs: D[]) {
-    const merged: { [key: string]: unknown } = {}
-    try {
-      return debugs.reduce(
-        (acc, debug) => ({ ...acc, ...(debug || {}) }),
-        merged,
-      )
-    } catch {
-      return merged
     }
   }
 


packages/bsky/src/hydration/actor.ts
```diff
@@ -304,6 +341,12 @@ export class ActorHydrator {
         blockingByList: parseString(rels.blockingByList),
         following: parseString(rels.following),
         followedBy: parseString(rels.followedBy),
+        extra: {
+          isDebugFieldAllowed: isDebugFieldAllowed(
+            this.config.debugFieldAllowedDids,
+            viewer,
+          ),
+        },
```
devinivy: I see that `isDebugFieldAllowed(debugFieldAllowedDids, viewer)` is applied both down in hydration and up in the views.  Do we need to do it in both places?  I feel like just handling it in the views may be sufficient.

My recommendation would be to calculate `isDebugFieldAllowed(debugFieldAllowedDids, viewer)` once and tack it on `HydrationCtx`.  Then it can be applied in the views via `state.ctx.includeDebugField`, as `state.ctx` is available in every view method.

packages/bsky/src/hydration/util.ts
```diff
@@ -164,3 +164,23 @@ export const isActivitySubscriptionEnabled = ({
   post: boolean
   reply: boolean
 }): boolean => post || reply
+
+export function isDebugFieldAllowed(
+  allowedDids: Set<string>,
+  did?: string | null,
```
devinivy: I think we want to ensure this arg is always passed, even if it's `undefined`.
```suggestion
  did: string | null | undefined,
```

packages/bsky/src/hydration/util.ts
```diff
@@ -164,3 +164,23 @@ export const isActivitySubscriptionEnabled = ({
...
+  if (!did) return false
+  return allowedDids.has(did)
+}
+
+export type DebugFieldObject = { [key: string]: unknown } | undefined
+
+export function mergeDebugFieldObjects(debugs: DebugFieldObject[]) {
+  const merged: DebugFieldObject = {}
+  for (const debug of debugs) {
+    if (!!debug && typeof debug === 'object') {
```
devinivy: I think type narrowing has our back here.
```suggestion
    if (debug) {
```

packages/bsky/tests/views/posts-debug.test.ts
```diff
@@ -0,0 +1,56 @@
...
+import { SeedClient, TestNetwork, basicSeed } from '@atproto/dev-env'
+import { ids } from '../../src/lexicon/lexicons'
+
+jest.mock('../../dist/hydration/util.js', () => {
+  const originalModule = jest.requireActual('../../src/hydration/util.ts')
+  return {
+    ...originalModule,
+    isDebugFieldAllowed: jest.fn(() => true),
+  }
+})
```
devinivy: If we can use the config and avoid jest module mocks that would rule.  I'd probably even prefer doing something a little dirty like reaching into `network.bsky.ctx.cfg` to add alice's DID.

packages/bsky/tests/views/profile-debug.test.ts
```diff
@@ -0,0 +1,49 @@
...
+})
+
+describe('profile views w/ debug field', () => {
+  let network: TestNetwork
+  let agent: AtpAgent
+  let sc: SeedClient
+
+  beforeAll(async () => {
+    network = await TestNetwork.create({
+      dbPostgresSchema: 'bsky_views_profile',
```
devinivy:
```suggestion
      dbPostgresSchema: 'bsky_views_profile_debug',
```

packages/bsky/tests/views/posts-debug.test.ts
estrattonbailey: Ah, hadn't considered just mutating the config, that should work

packages/bsky/src/hydration/actor.ts
estrattonbailey: Was thinking we should guard both, but now that I'm looking I don't really see any usages of these hydrators that _don't_ end up being processed by a view, aside from like `getRecord` which only returns the record to the client. If that's safe to do that sounds like a great idea!

matthieusieben:
Hey @estrattonbailey we updated `lex-cli` on main so please make sure to merge `main` and re-run `codegen` 🙏

packages/bsky/tests/views/posts-debug.test.ts
estrattonbailey: Niiiiice

packages/bsky/src/hydration/actor.ts
estrattonbailey: Like a charm, good call

packages/dev-env/src/bsky.ts
```diff
@@ -85,6 +85,7 @@ export class TestBsky {
       disableSsrfProtection: true,
       threadTagsBumpDown: new Set(),
       threadTagsHide: new Set(),
+      debugFieldAllowedDids: new Set(),
```
rafaeleyng, line 88: In dev mode, it could be useful to have a way to enable the `debug` field.

Maybe we do something like accept `true` or a list of DIDs, and then `true` means enable it for all?

Then there's also no way to mess this up in config, since `true` would only be a valid value if passed from dev-env, and when reading from env vars it wouldn't accept `true` (it would interpret it as one of the accepted DID values).

packages/bsky/src/views/index.ts
```diff
@@ -945,6 +948,21 @@ export class Views {
       threadgate: !post.record.reply // only hydrate gate on root post
         ? this.threadgate(threadgateUri, state)
         : undefined,
+      debug: state.ctx?.includeDebugField
+        ? this.debugField([{ post: post.debug }, { author: author.debug }])
+        : undefined,
```
rafaeleyng, line 951: I would prefer:
```ts
      debug: state.ctx?.includeDebugField
        ? { post: post.debug, author: author.debug }
        : undefined,
```

Then we don't need `debugField` nor any extra complexity.

packages/dev-env/src/bsky.ts
estrattonbailey: Certainly not against this, but I'm not sure we have much need for it in `dev-env` and would rather not have to document a carve-out for this. For testing, mutating the config directly works just fine. We could def add this later if we ever need it though.

packages/bsky/src/views/index.ts
estrattonbailey: Yeah good call. This handling evolved a bit, `debugField` and the `try/catch` isn't really needed anymore.

rafaeleyng (approved):

PR #4303, opened 2025-10-24
Update com.atproto.label defintions to give clearer descriptions

ThisIsMissEm:
The `val` property appears a few times, but from what I can tell, this should be the `identifier` in the `labelValueDefinition`, but may be an arbitrary string, if no definition exists?

### DIFF
--- a/lexicons/com/atproto/label/defs.json
+++ b/lexicons/com/atproto/label/defs.json
@@ -29,6 +29,7 @@
         "val": {
           "type": "string",
           "maxLength": 128,
+          "maxGraphemes": 128,
           "description": "The identifier for this label, see labelValueDefintion's identifier property."
         },
         "neg": {
@@ -71,6 +72,7 @@
         "val": {
           "type": "string",
           "maxLength": 128,
+          "maxGraphemes": 128,
           "description": "The identifier for this label, see labelValueDefintion's identifier property."
         }
       }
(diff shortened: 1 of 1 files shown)


lexicons/com/atproto/label/defs.json
```diff
@@ -29,7 +29,7 @@
         "val": {
           "type": "string",
           "maxLength": 128,
```
ThisIsMissEm, line 31: This max-length does not match the `maxLength` of `labelValueDefinition`'s property `identifier`, which is `maxLength: 100, maxGraphemes: 100`

lexicons/com/atproto/label/defs.json
```diff
@@ -71,7 +71,7 @@
...
```
ThisIsMissEm, line 74: This max-length does not match the `maxLength` of `labelValueDefinition`'s property `identifier`, which is `maxLength: 100, maxGraphemes: 100`

ThisIsMissEm:
@bnewbold @haileyok my current thinking here is that the `selfLabel` property should be for "labels that users of this labeler service can use to label their own content/accounts", and should be a strict subset of those defined in `labelValueDefinition` — i.e., giving discover over how people can moderate themselves, rather than just leaving it to a labeler service to label them.

This would also reduce the need for the "like this reply to have this label applied to your account" pattern that we currently see.

PR #4327, opened 2025-11-06
[APP-1484] Add thread/search exploration code and feature gate

estrattonbailey:
As we [wrote recently](https://bsky.social/about/blog/10-31-2025-building-healthier-social-media-update), we're in the process of evaluating and improving the systems we use to surface relevant content to users.

This PR integrates some new mechanisms for sorting/filtering post and replies in threads and search. It's feature-gated to Bluesky team members while we assess the effectiveness of new tooling. It should match existing handling pretty closely already, and but we'll be continually finessing it from here.


rafaeleyng (approved):
lgtm

PR #4330, opened 2025-11-08
explicit-slurs: remove stray match for unrelated word

### LINKED ISSUE #4288: explicitSlurRegexes identifies "faq" as a prohibited slur
**Describe the bug**

One of the regexes in `explicitSlurRegexes` incorrectly matches on the lowercase string `faq`. (It doesn't match the uppercase string `FAQ`.) This prevents creating a post with the hashtag `#faq`, among other similar problems.

Image

This was brought to my attention by [this blog post](https://alecmuffett.com/article/117533), which isn't by me, but I noticed no one had opened an issue.

The regex in question is the [third in the array](https://github.com/bluesky-social/atproto/blob/main/packages/pds/src/handle/explicit-slurs.ts#L6), which appears to be intending to match a similar-sounding actual slur with a different last letter. Specifically, the character group consisting mostly of `g` and characters which visually resemble a `g` also contains a `q` and several characters which resemble `q`s.

**To Reproduce**

1. Create a post containing a hashtag such as `#faq`

**Expected behavior**

Nothing is flagged.

Since "faq" is short for "frequently asked questions", and since Bluesky hashtags are by convention lowercase, many users would expect to be able to create a post with the hashtag `#faq`. (In fact, this was discovered by someone linking to a blog post which contained a list of frequently asked questions.)

**Details**

- Operating system: N/A
- Node version: N/A

**Additional context**

mistydemeo:
This regular expression is intended to match a slur ending with the letter "g", but it also matches the lowercase string `faq`, preventing people from creating posts with the hashtag `#faq`. Since that's an acronym for "frequently asked questions", and because hashtags are written in lowercase by convention, this actually does come up reasonably often - see [this blog post](https://alecmuffett.com/article/117533).

These regular expressions were originally written for another project, rather than bluesky, and I can understand that in that context matching "q" might have been appropriate because it can be somewhat visually similar to the other word in some fonts. However, those words don't look alike in the fonts used by bluesky.

Fixes #4288.


packages/pds/src/handle/explicit-slurs.ts
```diff
@@ -3,7 +3,7 @@
 const explicitSlurRegexes = [
```
bnewbold, line 4: could you add a comment under the "regexes taken from" line clarifying that there has been a modification? rg:

```
// these regexes were originally copied from: https://github.com/Blank-Cheque/Slurs
// small modifications have been made
```

bnewbold (commented):
generally looks good to me, left a comment about adding a comment.

we need a "changeset" generated for the "handle" package before merging. IIRC you run `pnpm changeset` from the top of the repo. this can be a patch-level change.

mistydemeo:
Done. I've added the comment and added the changeset.

bnewbold:
for what it is worth, I went to update the Go equivalent of this routine, which was based on the same regex set, and found that we had previously updated it to require two chars, not one (eg, `{2}` not `{1,2}`) for that specific character, while still matching "q" not just "g". I think either approach works and the two implementations don't need to be an exact match.

PR #3654, opened 2025-03-20
oauth-provider-ui: Add Japanese translation

dolciss:
Thanks for making the strings easier to translate on #2945.
I translated it into Japanese, please merge it if there is no problem.

PR #4331, opened 2025-11-08
OAuth: Add spanish (es) translation

tdelgado00:
Hi!

I've added Spanish translations for the oauth-provider and oauth-provider-frontend packages.

The changes follow the same approach as PR #3654.

As I understand from issue #7685, OAuth translations are not yet on Crowdin, so I hope submitting this here is correct.

Thanks!

P.S. This is for use in the Cabildo Abierto app (https://github.com/cabildo-abierto/cabildo-abierto).


matthieusieben:
Thank you for your help! Note that many changes where made since you last updated this. Please open a new PR with the new strings if you also want those translated.

PR #4333, opened 2025-11-09
bsky: improve error messaging for DPoP bound access tokens

ThisIsMissEm:
See the changeset for the details, I discovered whilst helping someone in the AT Protocol Touchers discord that the bsky.social server will accept DPoP bound access tokens sent as `Bearer <dpop-access-token>` instead of `DPoP <dpop-access-token>` (a fairly easy developer mistake), but the server misdirects the developer with the error "Bad token scope", because the `scope` on a DPoP bound access token is the actual oauth scopes like `atproto transition:generic`, when the bsky.social server expects `com.atproto.access` or similar.

This change just makes that error a little bit more visible to a developer. cc @matthieusieben

### DIFF
--- a/packages/bsky/src/auth-verifier.ts
+++ b/packages/bsky/src/auth-verifier.ts
@@ -234,7 +234,15 @@ export class AuthVerifier {
         )
       })
 
-    const { sub, aud, scope } = res.payload
+    const { sub, aud, scope, cnf } = res.payload
+    if (typeof cnf !== 'undefined') {
+      // Proof-of-Possession (PoP) tokens are not allowed here
+      // https://www.rfc-editor.org/rfc/rfc7800.html
+      throw new AuthRequiredError(
+        'Malformed token: DPoP not supported',
+        'InvalidToken',
+      )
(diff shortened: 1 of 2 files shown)


packages/bsky/src/auth-verifier.ts
```diff
@@ -234,7 +234,15 @@ export class AuthVerifier {
         )
       })
 
-    const { sub, aud, scope } = res.payload
+    const { sub, aud, scope, cnf } = res.payload
+    if (typeof cnf !== 'undefined') {
+      // Proof-of-Possession (PoP) tokens are not allowed here
+      // https://www.rfc-editor.org/rfc/rfc7800.html
+      throw new AuthRequiredError(
+        'Malformed token: DPoP not supported',
```
ThisIsMissEm, line 242: In the PDS we just give `Malformed token`, but here I think it's probably wise to be explicit and say "DPoP isn't supported", since getting here in the first place would be already a developer error (using `Bearer <token>` with a DPoP bound access token.

PR #4334, opened 2025-11-10
Avoid too much drift in generated TIDs

matthieusieben:
In the current implementation, every time more than one TID is generated within a single millisecond, a counter is increased and used to in create a global counter. That global counter does not get reset afterwards. This means that for every ms during which more than one TID is generated, the TID generator clock drifts for 1/1000 of a second.

If two TIDs are generated during the same second, once per minute, this results in a "drift" of about 1.44 second per day. Since servers are typically running for weeks, this can amount to a non negligeable amount of drift.

This change fixes that by:
1) Resetting the count when more than one second after the previously generated TID
2) Avoids drifting too much under heavy load, by re-using unused TIDs from the elapsed second
3) Throws an error if the load caused the drift to reach 1 minute (because the load caused the generation of 1_000_000 TIDs per second, sustained for 1 minute)

### DIFF
--- a/packages/common-web/src/tid.ts
+++ b/packages/common-web/src/tid.ts
@@ -3,8 +3,50 @@ import { s32decode, s32encode } from './util'
 const TID_LEN = 13
 
 let lastTimestamp = 0
-let timestampCount = 0
-let clockid: number | null = null
+let usCounter = 0
+
+// javascript does not have microsecond precision. instead, we append a counter
+// to the timestamp to indicate if multiple timestamps were created within the
+// same millisecond take max of current time & last timestamp to prevent tids
+// moving backwards if system clock drifts backwards
+function timeUs(): number {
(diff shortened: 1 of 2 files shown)


packages/common-web/src/tid.ts
```diff
@@ -22,20 +64,8 @@ export class TID {
   }
 
   static next(prev?: TID): TID {
-    // javascript does not have microsecond precision
-    // instead, we append a counter to the timestamp to indicate if multiple timestamps were created within the same millisecond
-    // take max of current time & last timestamp to prevent tids moving backwards if system clock drifts backwards
-    const time = Math.max(Date.now(), lastTimestamp)
-    if (time === lastTimestamp) {
-      timestampCount++
```
matthieusieben, line 30: We never reset this value, causing it to drift further every time there is high demand, without ever resetting when demand gets lower.

PR #4340, opened 2025-11-11
:sparkles: Update ozone action scheduler to queue emailing

foysalit:
This PR introduces a few new properties to ozone's action scheduling endpoint. This allows moderation clients to send in email subject and content to be sent out after successful takedowns to the owners of the accounts.

### DIFF
--- a/.changeset/mighty-pugs-admire.md
+++ b/.changeset/mighty-pugs-admire.md
@@ -1,0 +1,7 @@
+---
+"@atproto/ozone": patch
+"@atproto/api": patch
+"@atproto/pds": patch
+---
+
+Add optional email data to scheduled action api in ozone


packages/ozone/src/api/moderation/emitEvent.ts
```diff
@@ -165,13 +166,20 @@ const handleModerationEvent = async ({
...
+        ctx.modService(db).sendEmail({
+          subject: subjectLine,
+          content,
+          recipientDid: subject.did,
+        }),
+      )
+      event.isDelivered = true
+    } catch (err) {
+      httpLogger.error({ err, event }, 'failed to send mod event email')
+      event.isDelivered = false
```
ThisIsMissEm: To match the comment on line 169 it'd be:
```suggestion
      event.isDelivered = false
      httpLogger.error({ err, event }, 'failed to send mod event email')
```

packages/ozone/src/daemon/scheduled-action-processor.ts
```diff
@@ -92,6 +97,16 @@ export class ScheduledActionProcessor {
                 acknowledgeAccountSubjects:
                   eventData.acknowledgeAccountSubjects,
                 policies: eventData.policies,
+                severityLevel: eventData.severityLevel,
+                strikeCount: eventData.strikeCount,
+              }
+              const emailData = action.eventData as {
+                emailSubject?: string
+                emailContent?: string
+              }
```
ThisIsMissEm: Is this `as` just because the types for the lexicon for `ModEventTakedown` aren't updated yet? Or maybe you could reuse `eventData` here instead of `action.eventData` which is already cast to a `ModEventTakedown` type.

packages/ozone/src/daemon/scheduled-action-processor.ts
```diff
@@ -142,12 +158,14 @@ export class ScheduledActionProcessor {
...
 
   async performTakedown({
+    email,
     action,
     event,
     modTool,
     moderationTxn,
     settingService,
   }: {
+    email: { subject: string; content: string }
```
ThisIsMissEm, line 167: You could change this to:
```suggestion
    email: { subject: string; content: string } | null
```

And then in the earlier part of the code only set email if there's content, that'd remove the duplicate "do we have content" check on line 213. Extracting `{ subject: string; content: string }` into a type would also help with ensuring the `email` in `executeScheduledAction` aligns with the expected type here.

Another option is:
```suggestion
    email?: { subject: string; content: string }
```

packages/ozone/src/daemon/scheduled-action-processor.ts
```diff
@@ -192,6 +210,36 @@ export class ScheduledActionProcessor {
...
+      try {
+        await retryHttp(() =>
+          moderationTxn.sendEmail({
+            ...email,
+            recipientDid: action.did,
+          }),
+        )
+        isDelivered = true
+      } catch (err) {
+        dbLogger.error(
```
ThisIsMissEm, line 223: this probably isn't the right logger to use here? You may want a new subsystem logger for `ozone:daemon`, since this error isn't related to a database query.

packages/ozone/src/daemon/scheduled-action-processor.ts
```diff
@@ -192,6 +210,36 @@ export class ScheduledActionProcessor {
...
+          { err, did: action.did },
+          'failed to send takedown email',
+        )
+      }
+      await moderationTxn.logEvent({
+        event: {
+          content: email.content,
+          subjectLine: email.subject,
+          $type: 'tools.ozone.moderation.defs#modEventEmail',
+          comment: 'Communication attached to scheduled action',
```
ThisIsMissEm: Maybe vary the comment based on `isDelivered`? Then you'd be able to clearly see from the comment alone if the email communication was sent correctly or not from the scheduled action.

packages/ozone/tests/scheduled-action-processor.test.ts
```diff
@@ -74,7 +81,14 @@ describe('scheduled action processor', () => {
       const testSubject = sc.dids.alice
 
       const pastTime = new Date(Date.now() - 1000).toISOString()
-      await scheduleTestAction(testSubject, { executeAt: pastTime })
+      await scheduleTestAction(
```
ThisIsMissEm, line 84: I think this change would leave the non-email code path untested? May be worth having an explicit test to ensure scheduled actions that don't have email data don't trigger an email send?

Also, I made this comment in https://github.com/bluesky-social/atproto/pull/4221 but adding in mailpit to the dev-infra would allow you to actually test that the email was indeed sent. There's also some code in that PR for asserting sent emails with the current infrastructure in a clean-ish way.

packages/ozone/src/api/moderation/emitEvent.ts
foysalit: nice catch!

packages/ozone/src/daemon/scheduled-action-processor.ts
foysalit: yh reusing makes sense 👍🏽

packages/ozone/src/daemon/scheduled-action-processor.ts
foysalit: yeah we may have interchangeably used them in some places but makes sense to keep the context right.

packages/ozone/tests/scheduled-action-processor.test.ts
foysalit: other tests in this suite do not pass email data, we were just not explicitly validating the lack of email event. i've added that now 👍🏽

PR #4344, opened 2025-11-14
❇️ Ozone takedown target service

foysalit:
This PR adds an additional field in the takedown event which allows ozone to specify on which services a takedown should be applied.
It also adds `targetServices` as a property of policy so that from the client side, ozone can determine which services a takedown will be applied based on policy selection. This is intentionally designed to be loosely coupled so that ozone client can override the selection when applying takedowns.


shyamsoumya1991-dotcom:
Very nice thanks 🙏👍

PR #4348, opened 2025-11-14
WebSocket Library

dholms:
Pulls the `WebSocketKeepAlive` code out of `xrpc-server` into its own package so that we can re-use it in https://github.com/bluesky-social/atproto/pull/4290

Arguably this package should contain utilities for working with XRPC subscriptions, but I decided to hold off on moving that code over as we're actively working on refactoring lexicon/xrpc stuff more generally


packages/websocket/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "@atproto/websocket",
```
matthieusieben: nit: would call this something more specific like `websocket-server` or `keepalive-ws-server`

packages/ws-client/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "@atproto/websocket",
+  "version": "0.0.1",
```
matthieusieben, line 3: I might be wrong but I think that even new packages need a changeset to be released by our ci.

matthieusieben: (i think you have to create them manually as `pnpm changeset` won't pickup on new packages)

packages/websocket/package.json
dholms: yeah it's actually a client lib, but i'm down with `@atproto/ws-client`

PR #4349, opened 2025-11-16
Initial Permission Sets for `app.bsky.*` and `chat.bsky.*`

bnewbold:
See discussion for the naming and definition of these permission sets here: https://github.com/bluesky-social/atproto/discussions/4437

The protocol specs for the permission system and `permission-set` Lexicons was recently published: https://atproto.com/specs/permission

The `tools.ozone.*` permission set will require getting a compatible endpoint for mod reporting set up under that namespace.


lexicons/app/bsky/authDeletePosts.json
```diff
@@ -0,0 +1,25 @@
+{
+  "lexicon": 1,
+  "id": "app.bsky.authDeletePosts",
```
dholms: if we're taking one or the other, i lean towards `authManagePosts` than `authDeletePosts`

lexicons/app/bsky/authManageNotifs.json
```diff
@@ -0,0 +1,32 @@
+{
+  "lexicon": 1,
+  "id": "app.bsky.authManageNotifs",
```
dholms: nit: but we use the fully spelled out "notification" in all the other nsids

lexicons/app/bsky/authDeletePosts.json
bnewbold: I dug a bit more and I don't think either "manage posts" or "delete posts" are quite right, so I yanked this out for now.

- I don't think "manage posts" will work for many use-cases: actual social manager tools will want social graph, reposts, etc as well
- many of the "cleanup old posts" tools also want to hit likes and reposts

I'll open a follow-up PR with a "cleanup interactions" or similar permission set.

bnewbold: here is that PR: https://github.com/bluesky-social/atproto/pull/4453

lexicons/app/bsky/authManageFeedDeclaration.json
```diff
@@ -0,0 +1,21 @@
+{
+  "lexicon": 1,
+  "id": "app.bsky.authManageFeedDeclarations",
```
timburks, line 3: Nit: This id doesn't match the file name. I haven't seen a formal rule requiring this, but for every other lexicon in this repo, the filename can be directly derived from the ID (replace dots with spaces, append ".json", etc).

bnewbold: good catch! opened a PR to fix this: https://github.com/bluesky-social/atproto/pull/4568

Issue #4341, opened 2025-11-12
`app.bsky.feed.getPosts` returns 500 when batch includes posts from suspended accounts


ceciliaccwei:
**Describe the bug**

When using the `app.bsky.feed.getPosts` endpoint to batch fetch posts by URI, some requests fail with a `500 Internal Server Error` if **any** URI in the batch belongs to a **suspended account**. This behavior prevents partial success and disrupts downstream processing.

**To Reproduce**

1. Call `app.bsky.feed.getPosts` with a mix of valid and suspended post URIs.
     * GET `/xrpc/app.bsky.feed.getPosts?uris=at://did:plc:z3q6qapmm2h5kr65cnuc7ran/app.bsky.feed.post/32p3abzmaox17&uris=at://did:plc:erd6wu7fh27yobl7swebf4gr/app.bsky.feed.post/3m52gf5ham22i` 
3. Observe that the API returns `500 Internal Server Error`.
4. Removing the suspended URIs `at://did:plc:z3q6qapmm2h5kr65cnuc7ran/app.bsky.feed.post/32p3abzmaox17` allows the same request to succeed.

**Expected behavior**

The endpoint should **soft-fail** suspended or invalid posts — skipping over those entries and returning only valid posts — instead of failing the entire batch.

**Details**

- Operating system:
- Node version:

**Additional context**

**Example problematic URIs**

```
at://did:plc:z3q6qapmm2h5kr65cnuc7ran/app.bsky.feed.post/32p3abzmaox17  
at://did:plc:zn5mkvjauntoaefjnruoy4pl/app.bsky.feed.post/71ubrri8sddmn  
at://did:plc:qu5n5yybueiftsjartwxfrao/app.bsky.feed.post/yvcqhge6yq34g  
at://did:plc:4bhwiw67bjmkkvmo7hp464gn/app.bsky.feed.post/wt38gt22zh3qe  
```

Using the `app.bsky.actor.getProfile` endpoint, we confirmed these accounts are **suspended**.
When any of their posts are included in a batch request, the entire call fails with a 500.

rafaeleyng:
@ceciliaccwei This should be fixed, see:

```
> curl -s 'https://api.bsky.app/xrpc/app.bsky.feed.getPosts?uris=at://did:plc:z3q6qapmm2h5kr65cnuc7ran/app.bsky.feed.post/32p3abzmaox17'
{"posts":[]}%
```

Please close this if it working as you expected. ty

PR #4359, opened 2025-11-20
Adds tests for takendown cases in getPosts

rafaeleyng:
I wrote these tests trying to reproduce the bug in https://github.com/bluesky-social/atproto/issues/4341.

Turns out it seems we don't have that bug (see my first reply there). But there might be something to it, I just don't know what yet, and don't have time to investigate rn.


packages/bsky/tests/views/posts.test.ts
```diff
@@ -49,6 +72,70 @@ describe('pds posts views', () => {
...
+      {
+        headers: await network.serviceHeaders(
+          sc.dids.alice,
+          ids.AppBskyFeedGetPosts,
+        ),
+      },
+    )
+
+    expect(posts.data.posts.length).toBe(uris.length - 1)
+    expect(forSnapshot(posts.data.posts)).toMatchSnapshot()
```
dholms: no biggie but i think i would write these tests more explicitly by checking to see if the URI was returned.

ie `expect(posts.data.posts.map(p => p.uri).includes(TEST_URI)).toBe(false)` or similar

rafaeleyng: Done in 14e3e9fbe

PR #4406, opened 2025-12-02
Improve validation of DID Documents

matthieusieben:


.changeset/easy-aliens-stare.md
```diff
@@ -0,0 +1,5 @@
+---
+'@atproto/common-web': minor
+---
+
+Allow `serviceEndpoint` to be an array.
```
ThisIsMissEm, line 5: This is permissible? Is that just due to JSON-LD-isms or is there an actual use case for this? I feel like many are probably assuming a string value for serviceEndpoint (because a DID Document looks like JSON to the uninitiated)

.changeset/fuzzy-cows-enjoy.md
```diff
@@ -0,0 +1,5 @@
+---
+'@atproto/did': minor
+---
+
+Disallow (invalid) use of URI as `verificationMethod` array item.
```
ThisIsMissEm, line 5: Why would this be invalid? It could allow a key server separate from the DID Document?

.changeset/sharp-pugs-grow.md
```diff
@@ -0,0 +1,5 @@
+---
+'@atproto/did': patch
+---
+
+Allow spexcifying start and end indices when validating DID with `assertDid`
```
ThisIsMissEm: You had a typo here on specifying

packages/did/src/did-document.ts
```diff
@@ -111,13 +139,19 @@ export const didDocumentSchema = z.object({
...
   id: didSchema,
-  controller: didControllerSchema.optional(),
-  alsoKnownAs: z.array(rfc3968UriSchema).optional(),
+  controller: z.union([didSchema, z.array(didSchema)]).optional(),
+  alsoKnownAs: z.array(rfc3986UriSchema).optional(),
   service: z.array(didServiceSchema).optional(),
-  authentication: z.array(didAuthenticationSchema).optional(),
-  verificationMethod: z
-    .array(z.union([didVerificationMethodSchema, didRelativeUriSchema]))
+  authentication: z
```
ThisIsMissEm: Would we also consider supporting the other verification relationships that DID Documents support? Key Agreement could for instance help Germ or similar MLS/E2EE apps

packages/did/src/did-document.ts
```diff
@@ -83,18 +90,39 @@ const didServiceEndpointSchema = z.union([
...
+   * > conforming producer MUST NOT produce multiple service entries with the same
+   * > id. A conforming consumer MUST produce an error if it detects multiple
+   * > service entries with the same id.
+   *
+   * @see {@link https://www.w3.org/TR/did-1.0/#services}
+   * @note
+   * {@link https://www.rfc-editor.org/rfc/rfc3986#section-4.4 RFC3986 - 4.4. Same-Document Reference }
+   * allows relative URIs
+   */
+  id: z.union([rfc3986SameDocumentRefSchema, didUrlWithFragmentSchema]),
```
ThisIsMissEm: Is there a refinement on ID to prevent successful validation of non-unique service IDs?

ThisIsMissEm: Looks like maybe around line 186

.changeset/easy-aliens-stare.md
matthieusieben: This is just me reading the spec:

> The value of the serviceEndpoint property MUST be a string, a map, or a
> set composed of one or more strings and/or maps. All string values MUST
> be valid URIs conforming to [RFC3986] and normalized according to the
> Normalization and Comparison rules in RFC3986 and to any normalization
> rules in its applicable URI scheme specification.

.changeset/fuzzy-cows-enjoy.md
matthieusieben: Again did spec only allows Json objects in `verificationMethod` items:

> The verificationMethod property is OPTIONAL. If present, the value MUST be a [set](https://infra.spec.whatwg.org/#ordered-set) of [verification methods](https://www.w3.org/TR/did-1.0/#dfn-verification-method), where each [verification method](https://www.w3.org/TR/did-1.0/#dfn-verification-method) is expressed using a [map](https://infra.spec.whatwg.org/#ordered-map).

packages/did/src/did-document.ts
matthieusieben: Sure why not

.changeset/easy-aliens-stare.md
ThisIsMissEm: Yeah, I've read that too, I am concerned that if someone were to use this, it'd be breaking a lot of code. e.g., `getPdsEndpoint` in `@atproto/common-web` would break, afaict.

.changeset/fuzzy-cows-enjoy.md
ThisIsMissEm: I'm thinking of https://www.w3.org/TR/did-1.0/#referring-to-verification-methods here, where the array member is like `"did:example:123456789abcdefghi#keys-1",`

ThisIsMissEm: > However, if the value is a URL [string](https://infra.spec.whatwg.org/#string), the [verification method](https://www.w3.org/TR/did-1.0/#dfn-verification-method) has been included by reference and its properties will need to be retrieved from elsewhere in the [DID document](https://www.w3.org/TR/did-1.0/#dfn-did-documents) or from another [DID document](https://www.w3.org/TR/did-1.0/#dfn-did-documents). This is done by dereferencing the URL and searching the resulting [resource](https://www.w3.org/TR/did-1.0/#dfn-resources) for a [verification method](https://www.w3.org/TR/did-1.0/#dfn-verification-method) [map](https://infra.spec.whatwg.org/#ordered-map) with an id property whose value matches the URL.

This sounds like it would be valid to have a verificationMethod that is a URI.

matthieusieben: Referring to them is allowed. Its only the "verificationMethod" field that must be an array of json objects and must not contain strings, which is what was fixed.

ThisIsMissEm: Ohhh, I think I misunderstood the changelog entry then. Perhaps add a link to https://www.w3.org/TR/did-1.0/#verification-methods to it?

We can mark this as resolved.

.changeset/eighty-clowns-arrive.md
```diff
@@ -0,0 +1,5 @@
+---
+'@atproto/did': minor
+---
+
+Enforce `service` ids to be references to items of the current document (ie. values that start with `#` or `<document_id>#`)
```
matthieusieben: @copilot remove this file

PR #4408, opened 2025-12-02
Move PDS code over to `@atproto/lex`

matthieusieben:
Note, while reviewing, change the base branch to `msi/codegen-at-build` (https://github.com/bluesky-social/atproto/pull/4410) to hide codegen from the diff. The base branch is set to `main` to allow CI to run.

### DIFF
--- a/packages/bsky/src/hydration/hydrator.ts
+++ b/packages/bsky/src/hydration/hydrator.ts
@@ -15,9 +15,9 @@ import { ParsedLabelers } from '../util'
 import { uriToDid, uriToDid as didFromUri } from '../util/uris'
 import {
   ProfileRecord,
-  isListRule,
-  isRecordEmbed,
-  isRecordWithMedia,
+  isListRuleType,
+  isRecordEmbedType,
+  isRecordWithMediaType,
 } from '../views/types.js'
 import {
   ActivitySubscriptionStates,
@@ -1461,9 +1461,9 @@ const nestedRecordUrisFromPosts = (
 const nestedRecordUris = (post: Post['record']): AtUriString[] => {
   const uris: AtUriString[] = []
   if (!post?.embed) return uris
-  if (isRecordEmbed(post.embed)) {
+  if (isRecordEmbedType(post.embed)) {
     uris.push(post.embed.record.uri)
-  } else if (isRecordWithMedia(post.embed)) {
+  } else if (isRecordWithMediaType(post.embed)) {
     uris.push(post.embed.record.record.uri)
   }
   return uris
@@ -1472,7 +1472,7 @@ const nestedRecordUris = (post: Post['record']): AtUriString[] => {
 const getListUrisFromThreadgates = (gates: Threadgates): AtUriString[] => {
   const uris: AtUriString[] = []
   for (const gate of gates.values()) {
-    const listRules = gate?.record.allow?.filter(isListRule) ?? []
+    const listRules = gate?.record.allow?.filter(isListRuleType) ?? []
     for (const rule of listRules) {
       uris.push(rule.list)
     }
--- a/packages/bsky/src/util/uris.ts
+++ b/packages/bsky/src/util/uris.ts
@@ -1,6 +1,6 @@
 import { AtUri, DidString } from '@atproto/syntax'
 import { app } from '../lexicons/index.js'
-import { StrongRef, parseStrongRef } from '../views/types.js'
+import { StrongRef, validateStrongRef } from '../views/types.js'
 
 /**
  * Convert a post URI to a threadgate URI. If the URI is not a valid
@@ -32,13 +32,7 @@ export function uriToDid(uri: string): DidString {
 }
 
 // @TODO temp fix for proliferation of invalid pinned post values
-export function safePinnedPost(value: unknown) {
-  if (!value || typeof value !== 'object') {
-    return
-  }
-  const validated = parseStrongRef(value)
-  if (!validated.success) {
-    return
-  }
-  return validated.value as StrongRef
+export function safePinnedPost(value: unknown): StrongRef | undefined {
+  const validated = validateStrongRef(value)
+  return validated.success ? validated.value : undefined
 }
--- a/packages/bsky/src/views/index.ts
+++ b/packages/bsky/src/views/index.ts
@@ -1,5 +1,5 @@
 import { HOUR, MINUTE, mapDefined } from '@atproto/common'
-import { $Typed, Un$Typed, UriString } from '@atproto/lex'
+import { $Typed, Un$Typed, Unknown$TypedObject, UriString } from '@atproto/lex'
 import {
   AtUri,
   AtUriString,
@@ -92,17 +92,17 @@ import {
   VerificationView,
   VideoEmbed,
   VideoEmbedView,
-  isExternalEmbed,
-  isImagesEmbed,
-  isLabelerRecord,
-  isListRule,
-  isPostRecord,
-  isPostView,
-  isProfileRecord,
-  isRecordEmbed,
-  isRecordWithMedia,
-  isSelfLabels,
-  isVideoEmbed,
+  isExternalEmbedType,
+  isImagesEmbedType,
+  isLabelerRecordType,
+  isListRuleType,
+  isPostRecordType,
+  isPostViewType,
+  isProfileRecordType,
+  isRecordEmbedType,
+  isRecordWithMediaType,
+  isSelfLabelsType,
+  isVideoEmbedType,
 } from './types.js'
 import {
   VideoUriBuilder,
@@ -598,9 +598,10 @@ export class Views {
       cid,
       record: record,
       status: record.status,
-      embed: isExternalEmbed(record.embed)
-        ? this.externalEmbed(did, record.embed)
-        : undefined,
+      embed:
+        record.embed && isExternalEmbedType(record.embed)
+          ? this.externalEmbed(did, record.embed)
+          : undefined,
       expiresAt,
       isActive,
     }
@@ -770,15 +771,19 @@ export class Views {
 
     // Only these have a "labels" property:
     if (
-      !isPostRecord(record) &&
-      !isProfileRecord(record) &&
-      !isLabelerRecord(record)
+      !isPostRecordType(record) &&
+      !isProfileRecordType(record) &&
+      !isLabelerRecordType(record)
     ) {
       return []
     }
 
     // Ignore if no labels defines
-    if (!isSelfLabels(record.labels) || !record.labels.values.length) {
+    if (
+      !record.labels ||
+      !isSelfLabelsType(record.labels) ||
+      !record.labels.values.length
+    ) {
       return []
     }
 
@@ -938,7 +943,7 @@ export class Views {
       cid: gate.cid,
       record: gate.record,
       lists: mapDefined(gate.record.allow ?? [], (rule) => {
-        if (!isListRule(rule)) return
+        if (!isListRuleType(rule)) return
         return this.listBasic(rule.list, state)
       }),
     }
@@ -1041,24 +1046,31 @@ export class Views {
       const childBlocks = state.postBlocks?.get(uri)
       const parentBlocks = state.postBlocks?.get(parent.uri)
       // if child blocks parent, block parent
-      if (isPostView(parent) && childBlocks?.parent) {
+      if (isPostViewType(parent) && childBlocks?.parent) {
         parent = this.blockedPost(parent.uri, parent.author.did, state)
       }
       // if child or parent blocks root, block root
-      if (isPostView(root) && (childBlocks?.root || parentBlocks?.root)) {
+      if (isPostViewType(root) && (childBlocks?.root || parentBlocks?.root)) {
         root = this.blockedPost(root.uri, root.author.did, state)
       }
     }
     let grandparentAuthor: ProfileViewBasic | undefined
-    if (
-      isPostView(parent) &&
-      isPostRecord(parent.record) &&
-      parent.record.reply
-    ) {
-      grandparentAuthor = this.profileBasic(
-        creatorFromUri(parent.record.reply.parent.uri),
-        state,
-      )
+    if (isPostViewType(parent)) {
+      // @NOTE The "parent.record" property is of type "unknown" in the lexicon
+      // schema, which means that it is typed as LexMap here. In order to avoid
+      // (expensive) validation using "isPostRecord(parent.record)", we only
+      // check that the "$type" property is a post record type, then use a
+      // try/catch to "validate" the post uri.
+      if (isPostRecordType(parent.record) && parent.record.reply != null) {
+        const uri = (parent.record.reply as any).parent?.uri
+        if (typeof uri === 'string') {
+          try {
+            grandparentAuthor = this.profileBasic(creatorFromUri(uri), state)
+          } catch {
+            // ignore (just as if validation had failed)
+          }
+        }
+      }
     }
     return {
       root,
@@ -2005,7 +2017,7 @@ export class Views {
 
     const mutedByViewer = this.viewerMuteExists(authorDid, state)
     const isPushPin =
-      isPostRecord(post.record) && post.record.text.trim() === '📌'
+      isPostRecordType(post.record) && post.record.text.trim() === '📌'
 
     return {
       isOther: hiddenByTag || hiddenByThreadgate || mutedByViewer || isPushPin,
@@ -2040,19 +2052,19 @@ export class Views {
 
   embed(
     postUri: AtUriString,
-    embed: Embed | { $type: string },
+    embed: $Typed<Embed> | Unknown$TypedObject,
     state: HydrationState,
     depth: number,
   ): $Typed<EmbedView> | undefined {
-    if (isImagesEmbed(embed)) {
+    if (isImagesEmbedType(embed)) {
       return this.imagesEmbed(creatorFromUri(postUri), embed)
-    } else if (isVideoEmbed(embed)) {
+    } else if (isVideoEmbedType(embed)) {
       return this.videoEmbed(creatorFromUri(postUri), embed)
-    } else if (isExternalEmbed(embed)) {
+    } else if (isExternalEmbedType(embed)) {
       return this.externalEmbed(creatorFromUri(postUri), embed)
-    } else if (isRecordEmbed(embed)) {
+    } else if (isRecordEmbedType(embed)) {
       return this.recordEmbed(postUri, embed, state, depth)
-    } else if (isRecordWithMedia(embed)) {
+    } else if (isRecordWithMediaType(embed)) {
       return this.recordWithMediaEmbed(postUri, embed, state, depth)
     } else {
       return undefined
@@ -2265,11 +2277,11 @@ export class Views {
       | $Typed<ImagesEmbedView>
       | $Typed<VideoEmbedView>
       | $Typed<ExternalEmbedView>
-    if (isImagesEmbed(embed.media)) {
+    if (isImagesEmbedType(embed.media)) {
       mediaEmbed = this.imagesEmbed(creator, embed.media)
-    } else if (isVideoEmbed(embed.media)) {
+    } else if (isVideoEmbedType(embed.media)) {
       mediaEmbed = this.videoEmbed(creator, embed.media)
-    } else if (isExternalEmbed(embed.media)) {
+    } else if (isExternalEmbedType(embed.media)) {
       mediaEmbed = this.externalEmbed(creator, embed.media)
     } else {
       return
--- a/packages/bsky/src/views/util.ts
+++ b/packages/bsky/src/views/util.ts
@@ -4,12 +4,12 @@ import {
   GateRecord,
   PostRecord,
   PostgateRecord,
-  isFollowerRule,
-  isFollowingRule,
-  isListRule,
-  isMentionFacet,
-  isMentionRule,
-  isPostgateDisableRule,
+  isFollowerRuleType,
+  isFollowingRuleType,
+  isListRuleType,
+  isMentionFacetType,
+  isMentionRuleType,
+  isPostgateDisableRuleType,
 } from './types.js'
 
 export const parseThreadGate = (
@@ -26,16 +26,16 @@ export const parseThreadGate = (
     return { canReply: true }
   }
 
-  const allowMentions = gate.allow.some(isMentionRule)
-  const allowFollower = gate.allow.some(isFollowerRule)
-  const allowFollowing = gate.allow.some(isFollowingRule)
-  const allowListUris = gate.allow.filter(isListRule).map((item) => item.list)
+  const allowMentions = gate.allow.some(isMentionRuleType)
+  const allowFollower = gate.allow.some(isFollowerRuleType)
+  const allowFollowing = gate.allow.some(isFollowingRuleType)
+  const allowListUris = gate.allow.filter(isListRuleType).map((i) => i.list)
 
   // check mentions first since it's quick and synchronous
   if (allowMentions) {
     const isMentioned = rootPost?.facets?.some((facet) => {
       return facet.features.some(
-        (item) => isMentionFacet(item) && item.did === replierDid,
+        (item) => isMentionFacetType(item) && item.did === replierDid,
       )
     })
     if (isMentioned) {
@@ -80,7 +80,7 @@ export const parsePostgate = ({
     return { embeddingRules: { canEmbed: true } }
   }
 
-  const disabled = gate.embeddingRules.some(isPostgateDisableRule)
+  const disabled = gate.embeddingRules.some(isPostgateDisableRuleType)
   if (disabled) {
     return { embeddingRules: { canEmbed: false } }
   }
--- a/packages/bsky/src/api/app/bsky/notification/listNotifications.ts
+++ b/packages/bsky/src/api/app/bsky/notification/listNotifications.ts
@@ -15,7 +15,7 @@ import {
 import { Notification } from '../../../../proto/bsky_pb'
 import { uriToDid as didFromUri } from '../../../../util/uris'
 import { Views } from '../../../../views'
-import { isPostRecord } from '../../../../views/types'
+import { isPostRecordType } from '../../../../views/types'
 import { resHeaders } from '../../../util'
 
 export default function (server: Server, ctx: AppContext) {
@@ -181,7 +181,7 @@ const noBlockOrMutesOrNeedsFiltering = (
     if (item.reason === 'reply') {
       const post = hydration.posts?.get(uri)
       if (post) {
-        const rootPostUri = isPostRecord(post.record)
+        const rootPostUri = isPostRecordType(post.record)
           ? post.record.reply?.root.uri
           : undefined
         const isRootPostByViewer =
--- a/packages/bsky/src/views/types.ts
+++ b/packages/bsky/src/views/types.ts
@@ -3,7 +3,7 @@ import { app, chat, com } from '../lexicons/index.js'
 // app.bsky.actor
 
 export type ProfileRecord = app.bsky.actor.profile.Main
-export const isProfileRecord = app.bsky.actor.profile.$matches
+export const isProfileRecordType = app.bsky.actor.profile.$isTypeOf
 
 export type ProfileViewer = app.bsky.actor.defs.ViewerState
 export type KnownFollowers = app.bsky.actor.defs.KnownFollowers
@@ -24,19 +24,19 @@ export type BookmarkView = app.bsky.bookmark.defs.BookmarkView
 
 // app.bsky.embed
 
-export const isImagesEmbed = app.bsky.embed.images.$matches
+export const isImagesEmbedType = app.bsky.embed.images.$isTypeOf
 export type ImagesEmbed = app.bsky.embed.images.Main
 export type ImagesEmbedView = app.bsky.embed.images.View
 
-export const isVideoEmbed = app.bsky.embed.video.$matches
+export const isVideoEmbedType = app.bsky.embed.video.$isTypeOf
 export type VideoEmbed = app.bsky.embed.video.Main
 export type VideoEmbedView = app.bsky.embed.video.View
 
-export const isExternalEmbed = app.bsky.embed.external.$matches
+export const isExternalEmbedType = app.bsky.embed.external.$isTypeOf
 export type ExternalEmbed = app.bsky.embed.external.Main
 export type ExternalEmbedView = app.bsky.embed.external.View
 
-export const isRecordEmbed = app.bsky.embed.record.$matches
+export const isRecordEmbedType = app.bsky.embed.record.$isTypeOf
 export type RecordEmbed = app.bsky.embed.record.Main
 export type RecordEmbedView = app.bsky.embed.record.View
 export type EmbedBlocked = app.bsky.embed.record.ViewBlocked
@@ -44,7 +44,7 @@ export type EmbedDetached = app.bsky.embed.record.ViewDetached
 export type EmbedNotFound = app.bsky.embed.record.ViewNotFound
 export type PostEmbedView = app.bsky.embed.record.ViewRecord
 
-export const isRecordWithMedia = app.bsky.embed.recordWithMedia.$matches
+export const isRecordWithMediaType = app.bsky.embed.recordWithMedia.$isTypeOf
 export type RecordWithMedia = app.bsky.embed.recordWithMedia.Main
 export type RecordWithMediaView = app.bsky.embed.recordWithMedia.View
 export type RecordWithMediaEmbedView = app.bsky.embed.recordWithMedia.View
@@ -78,10 +78,11 @@ export type LikeRecord = app.bsky.feed.like.Main
 export type RepostRecord = app.bsky.feed.repost.Main
 
 export type PostRecord = app.bsky.feed.post.Main
-export const isPostRecord = app.bsky.feed.post.$matches
+export const isPostRecordType = app.bsky.feed.post.$isTypeOf
 
 export type PostgateRecord = app.bsky.feed.postgate.Main
-export const isPostgateDisableRule = app.bsky.feed.postgate.disableRule.$matches
+export const isPostgateDisableRuleType =
+  app.bsky.feed.postgate.disableRule.$isTypeOf
 
 export type FeedViewPost = app.bsky.feed.defs.FeedViewPost
 export type ReasonPin = app.bsky.feed.defs.ReasonPin
@@ -94,13 +95,15 @@ export type BlockedPost = app.bsky.feed.defs.BlockedPost
 export type GeneratorView = app.bsky.feed.defs.GeneratorView
 export type NotFoundPost = app.bsky.feed.defs.NotFoundPost
 export type PostView = app.bsky.feed.defs.PostView
-export const isPostView = app.bsky.feed.defs.postView.$matches
+export const isPostViewType = app.bsky.feed.defs.postView.$isTypeOf
 
 export type GateRecord = app.bsky.feed.threadgate.Main
-export const isFollowerRule = app.bsky.feed.threadgate.followerRule.$matches
-export const isFollowingRule = app.bsky.feed.threadgate.followingRule.$matches
-export const isListRule = app.bsky.feed.threadgate.listRule.$matches
-export const isMentionRule = app.bsky.feed.threadgate.mentionRule.$matches
+export const isFollowerRuleType =
+  app.bsky.feed.threadgate.followerRule.$isTypeOf
+export const isFollowingRuleType =
+  app.bsky.feed.threadgate.followingRule.$isTypeOf
+export const isListRuleType = app.bsky.feed.threadgate.listRule.$isTypeOf
+export const isMentionRuleType = app.bsky.feed.threadgate.mentionRule.$isTypeOf
 
 export type FeedGenRecord = app.bsky.feed.generator.Main
 
@@ -121,7 +124,7 @@ export type VerificationRecord = app.bsky.graph.verification.Main
 // app.bsky.labeler
 
 export type LabelerRecord = app.bsky.labeler.service.Main
-export const isLabelerRecord = app.bsky.labeler.service.$matches
+export const isLabelerRecordType = app.bsky.labeler.service.$isTypeOf
 export type LabelerView = app.bsky.labeler.defs.LabelerView
 export type LabelerViewDetailed = app.bsky.labeler.defs.LabelerViewDetailed
 
@@ -137,7 +140,7 @@ export type NotificationView =
 
 // app.bsky.richtext
 
-export const isMentionFacet = app.bsky.richtext.facet.mention.$matches
+export const isMentionFacetType = app.bsky.richtext.facet.mention.$isTypeOf
 
 // app.bsky.unspecced
 
@@ -159,12 +162,12 @@ export type ChatDeclarationRecord = chat.bsky.actor.declaration.Main
 // com.atproto.label
 
 export type Label = com.atproto.label.defs.Label
-export const isSelfLabels = com.atproto.label.defs.selfLabels.$matches
+export const isSelfLabelsType = com.atproto.label.defs.selfLabels.$isTypeOf
 
 // com.atproto.label
 
 export type StrongRef = com.atproto.repo.strongRef.Main
-export const parseStrongRef = com.atproto.repo.strongRef.$safeParse
+export const validateStrongRef = com.atproto.repo.strongRef.$safeValidate
 
 // com.germnetwork.declaration
 


packages/bsky/src/api/app/bsky/actor/getSuggestions.ts
```diff
@@ -82,7 +83,7 @@ const skeleton = async (input: {
       cursor: params.cursor,
       limit: params.limit,
     })
-    let dids = suggestions.dids
+    let dids = suggestions.dids.filter(isDidString)
```
matthieusieben, line 90: Allows to change the type without casting

packages/bsky/src/api/app/bsky/contact/importContacts.ts
```diff
@@ -38,10 +38,7 @@ export default function (server: Server, ctx: AppContext) {
         viewer,
       })
 
-      const result = await importContacts(
-        { ...input.body, hydrateCtx: hydrateCtx.copy({ viewer }) },
```
matthieusieben, line 42: There was no point in making a copy with the same viewer as the original.

.github/workflows/repo.yaml
```diff
@@ -24,20 +24,6 @@ jobs:
...
-          # Note: Keep here any package that has codegen output that is not checked in
-          path: |
-            packages/bsky/src/lexicon
-            packages/pds/src/lexicon
-          enableCrossOsArchive: true
-
-      - name: Perform code generation
-        if: steps.codegen-cache.outputs.cache-hit != 'true'
-        run: pnpm run codegen
-
```
matthieusieben: Remove caching introduced by intermediary PR

packages/bsky/tests/views/__snapshots__/posts.test.ts.snap
```diff
@@ -111,7 +111,7 @@ Object {
         },
         "video": Object {
           "$type": "blob",
-          "mimeType": "image/mp4",
+          "mimeType": "video/mp4",
```
matthieusieben, line 114: invalid mime was caught by type checking.

packages/bsky/tests/stash.test.ts
```diff
@@ -1,25 +1,39 @@
...
+      cid: FAKE_CID_1,
+    },
+  }
+  const validPayload1: app.bsky.bookmark.defs.Bookmark = {
+    subject: {
+      uri: 'at://did:plc:example/app.bsky.feed.post/2',
+      cid: FAKE_CID_2,
+    },
+  }
+  const invalidPayload: LexMap = {
```
matthieusieben: This is probably due to a merge conflict that was not properly resolved. I need to double check this.

packages/sync/src/firehose/index.ts
```diff
@@ -234,30 +233,35 @@ export const parseCommitAuthenticated = async (
...
-    }
-  })
+  const verifiedOps: com.atproto.sync.subscribeRepos.RepoOp[] = ops.filter(
+    (op) => {
+      const verifiedCid = verifiedCids[op.path]
+      if (op.action === 'delete') {
+        return verifiedCid === null
+      } else {
+        return (
+          op.cid != null && verifiedCid != null && verifiedCid?.equals(op.cid)
```
devinivy:
```suggestion
          op.cid != null && verifiedCid != null && verifiedCid.equals(op.cid)
```

packages/repo/bench/repo.bench.ts
```diff
@@ -1,48 +0,0 @@
-import { TID } from '@atproto/common'
```
devinivy, line 1: Why are the repo and mst benches removed?

packages/repo/src/mst/mst.ts
```diff
@@ -1,12 +1,12 @@
-import { CID } from 'multiformats'
 import { z } from 'zod'
-import { cidForCbor, dataToCborBlock, schema as common } from '@atproto/common'
+import { cidForLex, encode } from '@atproto/lex-cbor'
```
devinivy, line 2: I'd recommend we don't adopt lex-cbor in the repo until it is properly tested.  I believe this should also not be necessary to get the PDS converted to the new lex tooling.

packages/repo/src/sync/provider.ts
```diff
@@ -1,25 +1,30 @@
...
 
-async function* iterateFullRepo(storage: RepoStorage, commitCid: CID) {
-  const commit = await storage.readObjAndBytes(commitCid, def.commit)
-  yield { cid: commitCid, bytes: commit.bytes }
+async function* iterateFullRepo(
+  storage: RepoStorage,
+  commitCid: Cid,
+): AsyncGenerator<CarBlock> {
+  const cid = commitCid
+  if (!cid) throw new TypeError('Invalid CID format')
```
devinivy: When would this be falsy?

packages/repo/src/car.ts
```diff
@@ -163,15 +158,15 @@ async function* readCarBlocksIterGenerator(
...
+      const cid = decodeCid(blockBytes.subarray(0, 36))
       const bytes = blockBytes.subarray(36)
       yield { cid, bytes }
 
       // yield to the event loop every 25 blocks
       // in the case the incoming CAR is synchronous, this can end up jamming up the thread
       blocks++
       if (blocks % 25 === 0) {
-        await setImmediate()
+        await new Promise<void>((resolve) => setImmediate(resolve))
```
devinivy: What is the purpose of this change?

packages/repo/src/car.ts
```diff
@@ -262,7 +259,7 @@ class BufferedReader implements BytesReader {
         this.isDone = true
         return
       }
-      this.buffer = ui8.concat([this.buffer, next.value])
+      this.buffer = Buffer.concat([this.buffer, next.value])
```
devinivy: `Uint8Array` and `Buffer` have some subtle differences don't they, e.g. `slice()` behavior and probably some others?  I feel like this could end up being a little trap—perhaps we should convert back into a `Uint8Array` to avoid the possible footgun.

packages/repo/src/util.ts
```diff
@@ -66,9 +67,10 @@ export const ensureCreates = (
 }
 
 export const parseDataKey = (key: string): RecordPath => {
-  const parts = key.split('/')
-  if (parts.length !== 2) throw new Error(`Invalid record key: ${key}`)
-  return { collection: parts[0], rkey: parts[1] }
+  const { length, 0: collection, 1: rkey } = key.split('/')
+  if (length !== 2) throw new Error(`Invalid record key: ${key}`)
+  ensureValidNsid(collection)
```
devinivy: Is there a reason we're validating nsids here but not rkeys?  I would probably expect both or neither (and it looks like to date this codepath does not perform validation).

packages/dev-env/src/service-profile.ts
```diff
@@ -16,25 +17,29 @@ export class ServiceProfile {
...
+    protected agent: AtpAgent,
     protected userDetails: ServiceUserDetails,
   ) {}
 
-  get did() {
-    return this.client.assertDid
+  get did(): DidString {
+    const { assertDid } = this.agent
+    if (!isDidString(assertDid)) {
+      throw new Error('Agent is not authenticated')
```
devinivy: I wonder if this may be a confusing error if it occurs.

packages/pds/src/account-manager/helpers/account.ts
```diff
@@ -55,9 +61,9 @@ export const getAccount = async (
   const found = await selectAccountQB(db, flags)
     .where((qb) => {
       if (handleOrDid.startsWith('did:')) {
-        return qb.where('actor.did', '=', handleOrDid)
+        return qb.where('actor.did', '=', handleOrDid as DidString)
       } else {
-        return qb.where('actor.handle', '=', handleOrDid)
+        return qb.where('actor.handle', '=', handleOrDid as HandleString)
       }
```
devinivy: I think that kysely may have type helpers that can be used in schemas to distinguish types for where conditions vs. reading and writing.

devinivy: (It is certainly a bit inconvenient to have to use `as` in all these cases.)

packages/api/src/agent.ts
```diff
@@ -219,8 +219,9 @@ export class Agent extends XrpcClient {
   /**
    * Get the authenticated user's DID, or throw an error if not authenticated.
    */
-  get assertDid(): string {
+  get assertDid(): DidString {
```
devinivy, line 225: I would expect the types of `this.did` and `this.assertDid` to be consistent with each other, both using `string` or both using `DidString`.

packages/dev-env/src/service-profile.ts
matthieusieben: This is no longer needed because `assertDid` is already returning a `DidString`

packages/pds/src/account-manager/helpers/account.ts
matthieusieben: This requires to define the column as `ColumnType<string, DidString, DidString>`, where the first argument is the "select" type. Changing the type like so means every `select()` will no longer be typed as `DidString`, and that type casting will be needed over there.

packages/repo/src/car.ts
matthieusieben: The goal is to make this package able to run in more environments than just NodeJS by removing dependencies on `node:` imports.

The only missing peace to make this package fully runtime agnostic is to replace the `Readable` (from `node:stream`) with `AsyncIterable` (which would be another breaking change)

packages/repo/src/mst/mst.ts
matthieusieben: `lex-cbor` is actually pretty thoroughly tested. Code coverage is almost 100%, and tests contain both unit tests as well as integration tests based on our test vectors, in addition to tests ported from `@ipld/dag-cbor`.

Also, the `lex-cbor` is already used as underlying CBOR encoding/decoding library in the entire `atproto` repo.

matthieusieben: The reason it *is* actually needed is because `@ipld/dag-cbor` relies on `multiformats` for the `CID` implementation, which is not 100% compatible with the `Cid` interface from `@atproto/lex-data` & `@atproto/lex` (`CID` can be assigned to `Cid` but not the other way around)

packages/repo/src/sync/provider.ts
matthieusieben: must be a left over.

packages/repo/src/util.ts
matthieusieben: Added `rkey` validation

packages/repo/bench/repo.bench.ts
matthieusieben: Because they both are broken on `main`, which seems to indicate that they are not very useful. I can restore them if you think they are valuable still.

Capture d’écran 2026-03-06 à 12 30 22
Capture d’écran 2026-03-06 à 12 30 09

packages/pds/src/account-manager/helpers/account.ts
matthieusieben: I fixed this by typing `handleOrDid` as `AtIdentityString` instead (which is the actual right solution here)

packages/api/src/atp-agent.ts
```diff
@@ -88,10 +88,6 @@ export class AtpAgent extends Agent {
     return this.sessionManager.hasSession
   }
 
-  get did() {
```
matthieusieben, line 91: Inherited from parent

packages/repo/src/car.ts
devinivy: What environment do you have in mind that supports `setImmediate()` but not `node:timers`?  My impression is that compat has gotten quite good on the server side around the node builtin modules but maybe there's a gap.

packages/repo/src/util.ts
devinivy: Are we sure we want to be performing validation in this codepath?  I expected there may have been a reason we weren't doing it before.

packages/repo/src/car.ts
matthieusieben: I was thinking Browsers. I didn't realize `setImmediate` was [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/Window/setImmediate) in browsers, so this would need to use `setTimeout`.

This is more a general pattern to avoid relying on NodeJS specific features when possible.

packages/repo/src/util.ts
matthieusieben: I added them because consumer code needs `collection` to be `NsidString`, and `rkey` to be `RecordKeyString`. Like here:

https://github.com/bluesky-social/atproto/blob/e6ed423aeaab24d46dd9c37f71a4b5ac6c6e7d74/packages/pds/src/scripts/sequencer-recovery/recoverer.ts#L241

Also it makes things more consistent: AT Protocol describes `collection` to be an NSID and also describes the format of rkey. I am not sure what allowing anything else here would help us with.

That being said, I was able to revert the changes here and move the assertion to the consumer code paths.

packages/repo/src/mst/mst.ts
devinivy: I didn't catch that we ported the @ipld/dag-cbor tests over, that should be good 👍 Last time I was in there there were only tests of the happy path, and all the subtle bugs happen elsewhere.

packages/repo/src/util.ts
devinivy: Just double-checking!  For example, perhaps it's a hot codepath and is only used on trusted inputs; or maybe a bad collection/rkey needs to be detected and handled specially by some calling code.  This codepath could have been validating but it wasn't, so I just want to double-check that making it stricter wont have any unintended consequences.  I didn't mean to imply it was a bad change, just that it's worth scrutinizing.

packages/bsky/src/hydration/util.ts
```diff
@@ -155,10 +157,10 @@ export const urisByCollection = <T extends string>(
   return result
 }
 
-export const split = <T>(
-  items: T[],
+export function split<T>(
+  items: Iterable<T>,
   predicate: (item: T) => boolean,
-): [T[], T[]] => {
+): [T[], T[]] {
```
devinivy: What is the story with these changes?

devinivy: At a glance these look like arbitrary changes to appview helpers (though maybe I'm off there!), and it takes `split` out of the arrow style used in this file.  If I am understanding the changes right and they are unrelated to the @atproto/lex rollout I'd recommend we just leave these as they were.

packages/bsky/src/views/util.ts
```diff
@@ -31,9 +29,7 @@ export const parseThreadGate = (
   const allowMentions = gate.allow.some(isMentionRule)
   const allowFollower = gate.allow.some(isFollowerRule)
   const allowFollowing = gate.allow.some(isFollowingRule)
-  const allowListUris = gate.allow
-    .map((item) => (isListRule(item) ? item.list : undefined))
-    .filter((v) => v != null)
+  const allowListUris = gate.allow?.filter(isListRule).map((item) => item.list)
```
devinivy: I believe `gate.allow` should exist here, no need for the `?`.

packages/bsky/src/stash.ts
```diff
@@ -13,7 +13,9 @@ export const Namespaces = {
     app.bsky.notification.defs.subjectActivitySubscription,
   AppBskyUnspeccedDefsAgeAssuranceEvent:
     app.bsky.unspecced.defs.ageAssuranceEvent,
-} satisfies Record<string, TypedObjectSchema>
+  AppBskyActorDefsProfileAssociatedChat:
+    app.bsky.actor.defs.profileAssociatedChat,
```
devinivy: Was this added to fix a type error?

packages/bsky/tests/stash.test.ts
```diff
@@ -1,37 +1,29 @@
...
 describe('private data', () => {
   let network: TestNetwork
   let stashClient: StashClient
   let db: Database
 
   const actorDid = 'did:plc:example'
-  // Using bookmark namespace which is available in Namespaces
-  const namespace = Namespaces.AppBskyBookmarkDefsBookmark
+  // This lexicon has nothing special other than being simple, convenient to use in a test.
+  const namespace = Namespaces.AppBskyActorDefsProfileAssociatedChat
```
devinivy: I don't think that we want to place `AppBskyActorDefsProfileAssociatedChat` in the user's stash, so I feel like we may want to keep it off of `Namespaces`.  What was the impetus for this change?

packages/bsky/src/hydration/util.ts
matthieusieben: Yeah this was a transient change. will revert it back 👍

packages/repo/src/car.ts
```diff
@@ -162,11 +162,15 @@ async function* readCarBlocksIterGenerator(
...
+      // yield to the event loop every 25 blocks in the case the incoming CAR is
+      // synchronous, this can end up jamming up the thread
       blocks++
       if (blocks % 25 === 0) {
-        await new Promise<void>((resolve) => setImmediate(resolve))
+        // @NOTE using `setTimeout` instead of `setImmediate` to remove
+        // dependency on runtime specific functions, allowing to (eventually)
+        // ensure that this package can be used in any JS environment (including
+        // browsers).
+        await new Promise<void>((resolve) => setTimeout(resolve))
```
devinivy: I think repeated calls to `setTimeout()` like this might end up triggering browser clamping to 4ms, which we wouldn't want.  And the perf impact in nodejs isn't obvious to me, but I expect it could be significant.  In my view if we want this to work in a browser setting we need a proper ponyfill for `setImmediate()`.

devinivy: This also isn't super related to moving the PDS to @atproto/lex, so it may be smart for us to punt on this if it's not super simple.

packages/bsky/tests/stash.test.ts
matthieusieben: I don't know. This is just  a port from main...

packages/pds/src/actor-store/blob/transactor.ts
```diff
@@ -340,46 +341,24 @@ async function mimeTypeFromStream(
...
-    throwInvalid(
+    throw new InvalidRequestError(
       `Referenced Mimetype does not match stored blob. Expected: ${found.mimeType}, Got: ${blob.mimeType}`,
       'InvalidMimeType',
     )
   }
-  if (
-    blob.constraints.accept &&
-    !acceptedMime(blob.mimeType, blob.constraints.accept)
-  ) {
```
devinivy, line 379: I assume this logic now occurs during lex validation against the blob ref, is that right?  So now here we just need to ensure the blob ref matches the uploaded blob exactly.

packages/pds/src/api/app/bsky/feed/getPostThread.ts
```diff
@@ -204,12 +206,15 @@ const readAfterWriteNotFound = async (
...
     try {
       assert(ctx.bskyAppView)
-      const parentsRes =
-        await ctx.bskyAppView.agent.app.bsky.feed.getPostThread(
-          { uri: highestParent, parentHeight: params.parentHeight, depth: 0 },
-          await ctx.appviewAuthHeaders(requester, ids.AppBskyFeedGetPostThread),
-        )
-      thread.parent = parentsRes.data.thread
+      const parentsRes = await ctx.bskyAppView.client.call(
+        app.bsky.feed.getPostThread.main,
```
devinivy: Is the `.main` part necessary here?
```suggestion
        app.bsky.feed.getPostThread,
```

packages/pds/src/api/com/atproto/identity/requestPlcOperationSignature.ts
```diff
@@ -14,34 +15,39 @@ export default function (server: Server, ctx: AppContext) {
...
+    handler: entrywayClient
+      ? async ({ auth, req }) => {
+          const { headers } = await ctx.entrywayAuthHeaders(
             req,
             auth.credentials.did,
-            ids.ComAtprotoIdentityRequestPlcOperationSignature,
-          ),
-        )
-        return
-      }
```
devinivy, line 28: Is there a reason we are changing the way these entryway codepaths have been written?  I feel like the readability suffers a bit moving to ternaries, and I can't think of any practical downsides to the way it has been handled to date.

packages/repo/src/util.ts
matthieusieben: I just reverted as it is safer. I am not familiar enough to be 100% confident of the side effect of this. In particular if there are some repos out there where things are in a funky state, we probably don't want to cause server errors.

packages/pds/src/api/com/atproto/identity/requestPlcOperationSignature.ts
matthieusieben: I changed to use an if-else

packages/bsky/src/stash.ts
matthieusieben: This is needed to properly migrate the tests:

[packages/bsky/tests/stash.test.ts](https://github.com/bluesky-social/atproto/pull/4408/files/7047f14ee862b5c50b66f2decea729d690533908#diff-312807450e0bbec413d4d0f80914f69fc7ca1f92708ed8a016ac3250cc847d36)

Capture d’écran 2026-03-10 à 19 24 32

packages/pds/src/actor-store/blob/transactor.ts
matthieusieben: Yep that's right 👍

packages/bsky/tests/stash.test.ts
matthieusieben: https://github.com/bluesky-social/atproto/blob/3dc37915436dec7e18c7dc9dcf01b72cad53fdbe/packages/bsky/tests/stash.test.ts#L15-L22

devinivy: Got it!

packages/bsky/src/stash.ts
matthieusieben: Ok I guess I was confused. I changed the way this works and `AppBskyActorDefsProfileAssociatedChat` was removed from `Namespaces`.

packages/bsky/tests/stash.test.ts
matthieusieben: I removed `AppBskyActorDefsProfileAssociatedChat` from `Namespaces`. Now this test was updated to use `app.bsky.actor.defs.profileAssociatedChat` directly.

matthieusieben: https://github.com/bluesky-social/atproto/pull/4408/commits/f182acdbf8c0fd47bb381e4778c49880f7abdc70

packages/bsky/src/stash.ts
matthieusieben: https://github.com/bluesky-social/atproto/pull/4408/commits/f182acdbf8c0fd47bb381e4778c49880f7abdc70

packages/bsky/tests/_util.ts
```diff
@@ -39,17 +28,19 @@ export function assertIsThreadViewPost(
 
 // Swap out identifiers and dates with stable
 // values for the purpose of snapshot testing
-export const forSnapshot = (obj: unknown) => {
+export const forSnapshot = (obj: any) => {
```
devinivy: `unknown` seemed good!  Do we need `any`?

packages/bsky/src/api/app/bsky/feed/getFeed.ts
```diff
@@ -205,69 +202,58 @@ const skeletonFromFeedGen = async (
...
-          'feed provided an invalid response',
-          'InvalidFeedResponse',
-        )
-      }
+    if (result.error === 'UpstreamFailure') {
+      throw new UpstreamFailureError(
+        'feed provided an invalid response',
+        'InvalidFeedResponse',
+      )
     }
```
devinivy: Could you help me understand how each of these cases w/ the upstream get handled here?
 - Network error
 - Response validation error
 - 500
 - 502

Here's what it seems like from the code:
- Network error → `500 / InternalServerError` → `502 / "feed unavailable"`
- Response validation error → `502 / UpstreamFailure` → `502 / InvalidFeedResponse`
- 500 → `500 / InternalServerError` → `502 / "feed unavailable"`
- 502 → `502 / UpstreamFailure` → `502 / InvalidFeedResponse`

If that's right, these are the parts that surprise me:
- Network error → `500 / InternalServerError`
  - Would expect a network error to be mapped into a non-http error or something more like a 503/504.
- `502 / UpstreamFailure` → `502 / InvalidFeedResponse`
  - Would expect the end result to be `502 / "feed unavailable"`.

Also as a general principle we should probably aim to preserve the existing behavior as much as possible.  And please forgive me if I've just misunderstood these changes 😆

packages/bsky/src/api/app/bsky/feed/getPostThread.ts
```diff
@@ -105,10 +103,10 @@ const presentation = (
 ) => {
   const { ctx, params, skeleton, hydration } = inputs
   const thread = ctx.views.thread(skeleton, hydration, {
-    height: params.parentHeight,
+    height: params.parentHeight!,
     depth: getDepth(ctx, skeleton.anchor, params),
   })
-  if (isNotFoundPost(thread)) {
+  if (app.bsky.feed.defs.notFoundPost.$matches(thread)) {
```
devinivy: This is full validation, right?  I expect at this point we can just discriminate.

packages/bsky/src/api/app/bsky/feed/getFeed.ts
```diff
@@ -205,69 +202,58 @@ const skeletonFromFeedGen = async (
...
-    throw err
+    throw result.reason
   }
 
-  const { feed: feedSkele, ...skele } = skeleton
+  const { feed: feedSkele, cursor, ...skele } = result.body
   const feedItems = feedSkele.slice(0, params.limit).map((item) => ({
     post: { uri: item.post },
-    repost: isSkeletonReasonRepost(item.reason)
+    repost: app.bsky.feed.defs.skeletonReasonRepost.$matches(item.reason)
```
devinivy: I believe we can discriminate here rather than validate.

packages/bsky/src/api/app/bsky/feed/searchPosts.ts
```diff
@@ -82,10 +85,11 @@ const skeleton = async (inputs: SkeletonFnInput<Context, Params>) => {
...
         url: params.url,
         viewer: params.hydrateCtx.viewer ?? undefined,
-      })
+      },
+    )
     return {
-      posts: res.posts.map(({ uri }) => uri),
-      cursor: parseString(res.cursor),
+      posts: res.posts.map(({ uri }) => uri as AtUriString),
+      cursor: res.cursor,
```
devinivy: Why is `parseString()` no longer needed?

packages/bsky/src/api/app/bsky/graph/getListsWithMembership.ts
```diff
@@ -64,7 +63,11 @@ const skeleton = async (
     cursor: params.cursor,
     limit: params.limit,
   })
-  return { actorDid, listUris, cursor: cursor || undefined }
+  return {
+    actorDid,
+    listUris: listUris as AtUriString[],
+    cursor: cursor || undefined,
```
devinivy: In `getLists` a `parseString()` was added here.  We should probably keep these consistent with each other (and maybe leave them both as-is to avoid mixing in subtle behavioral changes to this already-extensive work?).

packages/bsky/src/api/app/bsky/graph/getRelationships.ts
```diff
@@ -14,36 +22,35 @@ export default function (server: Server, ctx: AppContext) {
...
+            })
       })
       return {
         encoding: 'application/json',
         body: {
           actor,
           relationships,
         },
-      }
+      } satisfies app.bsky.graph.getRelationships.$Output
```
devinivy: Does this have any effect?

packages/bsky/src/api/app/bsky/graph/getRelationships.ts
```diff
@@ -1,10 +1,18 @@
...
 export default function (server: Server, ctx: AppContext) {
-  server.app.bsky.graph.getRelationships({
+  server.add(app.bsky.graph.getRelationships, {
     handler: async ({ params }) => {
       const { actor, others = [] } = params
+
+      // @TODO Should this be validated at the schema level?
+      if (!isDidString(actor) || !others.every(isDidString)) {
+        throw new InvalidRequestError('Invalid DID format')
+      }
```
devinivy: These don't have to be DIDs, they can also be handles.

packages/bsky/src/api/app/bsky/unspecced/getSuggestedStarterPacks.ts
```diff
@@ -53,21 +53,21 @@ export default function (server: Server, ctx: AppContext) {
...
+      {
+        limit: params.limit,
+        viewer: params.hydrateCtx.viewer ?? undefined,
+      },
+      {
+        headers: params.headers,
+      },
+    )
   } else {
     throw new InternalServerError('Topics agent not available')
```
devinivy: In the other endpoint this was turned into a 501.  I'm not sure we should make additional behavioral changes as part of this work since it is already extensive, but if we're going to we should probably keep everything consistent.

packages/bsky/src/api/app/bsky/unspecced/getTrendingTopics.ts
```diff
@@ -59,44 +58,37 @@ const skeleton = async (input: SkeletonFnInput<Context, Params>) => {
         headers: params.headers,
       },
     )
-    return res.data
   } else {
     throw new InternalServerError('Topics agent not available')
```
devinivy: Here's another one 👍

packages/bsky/src/api/app/bsky/unspecced/getTrends.ts
```diff
@@ -61,7 +60,6 @@ const skeleton = async (input: SkeletonFnInput<Context, Params>) => {
         headers: params.headers,
       },
     )
-    return res.data
   } else {
     throw new InternalServerError('Topics agent not available')
```
devinivy: Here is another.  This is the last one I'll comment on 😆

packages/bsky/src/api/app/bsky/unspecced/initAgeAssurance.ts
```diff
@@ -4,18 +4,19 @@ import { isDisposableEmail } from 'disposable-email-domains-js'
...
   InvalidRequestError,
   MethodNotImplementedError,
+  Server,
 } from '@atproto/xrpc-server'
 import { AppContext } from '../../../../context'
 import { KwsExternalPayloadError } from '../../../../kws'
-import { Server } from '../../../../lexicon'
-import { InputSchema } from '../../../../lexicon/types/app/bsky/unspecced/initAgeAssurance'
+import { app } from '../../../../lexicons/index.js'
+type InputSchema = app.bsky.unspecced.initAgeAssurance.$InputBody
```
devinivy: Maybe we should move the `type` beneath the imports, rather than mix it in.

packages/bsky/src/api/com/atproto/admin/updateSubjectStatus.ts
```diff
@@ -21,7 +20,7 @@ export default function (server: Server, ctx: AppContext) {
       const now = new Date()
       const { subject, takedown } = input.body
       if (takedown) {
-        if (isRepoRef(subject)) {
+        if (com.atproto.admin.defs.repoRef.$matches(subject)) {
```
devinivy: Can we discriminate these instead of re-validate them?

packages/bsky/src/api/com/atproto/identity/resolveHandle.ts
```diff
@@ -1,11 +1,11 @@
...
-import { InvalidRequestError } from '@atproto/xrpc-server'
+import { normalizeHandle } from '@atproto/syntax'
+import { InvalidRequestError, Server } from '@atproto/xrpc-server'
 import { AppContext } from '../../../../context'
-import { Server } from '../../../../lexicon'
+import { com } from '../../../../lexicons/index.js'
 
 export default function (server: Server, ctx: AppContext) {
-  server.com.atproto.identity.resolveHandle(async ({ req, params }) => {
-    const handle = ident.normalizeHandle(params.handle || req.hostname)
```
devinivy, line 8: 😆 wow, weird that this was in there!

packages/bsky/src/data-plane/server/indexing/plugins/germ-declaration.ts
```diff
@@ -41,23 +39,16 @@ const notifsForDelete = () => {
...
+    schema: com.germnetwork.declaration.main,
     insertFn,
     findDuplicate,
     deleteFn,
     notifsForInsert,
     notifsForDelete,
   })
-  // @TODO use lexicon validation
-  processor.assertValidRecord = () => null
-  return processor
```
devinivy, line 60: Hmm, not sure I understand why this was here originally.

packages/bsky/src/hydration/actor.ts
```diff
@@ -129,306 +147,376 @@ export class ActorHydrator {
...
-      if (
-        !actor.exists ||
-        (isNoHosted && !includeTakedowns) ||
-        !!actor.tombstonedAt
-      ) {
-        return acc.set(did, null)
-      }
+    const map: Actors = new HydrationMap()
+
+    if (dids.length) {
```
devinivy: Can we leave the `!dids.length` case as it was, handled up front?

packages/bsky/src/hydration/actor.ts
```diff
@@ -129,306 +147,376 @@ export class ActorHydrator {
...
-    return uris.reduce((acc, uri, i) => {
-      const record = parseRecord<ChatDeclarationRecord>(
-        res.records[i],
-        includeTakedowns,
-      )
-      return acc.set(uri, record ?? null)
-    }, new HydrationMap<ChatDeclaration>())
+    const map: ChatDeclarations = new HydrationMap()
+
+    if (uris.length) {
```
devinivy: Can we keep the special empty cases at the top of the functions as they were?  I'd love to avoid misc stylistic changes as part of this already quite large batch of work.  (I wont add comments to the others 👍 )

packages/bsky/src/hydration/actor.ts
```diff
@@ -129,306 +147,376 @@ export class ActorHydrator {
...
+            { actorDid: viewer, subjectDids: dids },
+            { signal: AbortSignal.timeout(100) },
+          )
 
-    const activitySubscription = (val: ActivitySubscription | undefined) => {
-      if (!val) return undefined
+        for (let i = 0; i < dids.length; i++) {
+          const did = dids[i]
+          const state = {
+            post: subscriptions[i].post != null,
```
devinivy: The code used to be defensive around a missing `subscriptions[i]`, I feel like we should probably preserve that since it was there explicitly in the types.

packages/bsky/src/hydration/hydrator.ts
```diff
@@ -1434,38 +1444,41 @@ const nestedRecordUrisFromPosts = (
...
+    if (gate?.record.allow) {
+      for (const rule of gate.record.allow) {
+        if (isListRule(rule)) {
+          yield rule.list
+        }
+      }
     }
   }
-  return uris
 }
```
devinivy: Could we set this back?  The existing code seems to be just fine, probably clearer for many readers, and it's not really related to @atproto/lex adoption.  There are already plenty of code changes here to contend with!

packages/bsky/src/api/app/bsky/feed/getPostThread.ts
matthieusieben: Eye of the tiger

packages/bsky/src/api/app/bsky/feed/searchPosts.ts
matthieusieben: This is a mistake.

packages/bsky/src/api/app/bsky/graph/getRelationships.ts
matthieusieben: I was helpful to determine why the typing was wrong. We no longer need it though.

packages/bsky/src/hydration/actor.ts
matthieusieben: sure

packages/bsky/src/api/app/bsky/graph/getRelationships.ts
matthieusieben: This is awkward because the code downstream of this only supports DIDs.

matthieusieben: I've added support for handles in `getProfileViewerStatesNaive`

packages/bsky/src/api/com/atproto/admin/updateSubjectStatus.ts
matthieusieben: indeed.

packages/bsky/src/api/com/atproto/identity/resolveHandle.ts
matthieusieben: This actually shows the usefulness of strongly typed strings.

packages/bsky/src/hydration/actor.ts
matthieusieben: I kinda assumed that we can trust the protobuf codegen here, which does not have null/undefined as possible data. I did restore the previous behavior.

packages/bsky/src/api/app/bsky/feed/getFeed.ts
matthieusieben: ### `result.error === 'InternalServerError'`

This covers both internal errors as well as `fetch()` errors. Since values are validated, this is _most likely_ a `fetch()` error.

This case corresponds to the previous `err.status === ResponseType.Unknown` case (either a  `fetchHandler` error, `response.arrayBuffer()` error, or an internal error when processing the response):

https://github.com/bluesky-social/atproto/blob/ce0fa90bedca5af2aa4be5ef0e6b5877fdcd6e6c/packages/xrpc/src/xrpc-client.ts#L121

As you can see ^ network errors are currently converted into 502.

We have [changes](https://github.com/bluesky-social/atproto/blob/5f73f8ed724aea74210201a658fb7a06bd8b1adf/packages/lex/lex-client/src/xrpc.ts#L185), on `main`, that allow to specifically target `fetch()` errors. We could leverage those to return 503/504 here. But this would actually be a change in behavior. 

Also, i'd prefer not to merge `main` into this branch to keep things simple to review.

### `result.error === 'UpstreamFailure'`

This means that the request could be made, but something went wrong with the processing of the response (5xx, payload encoding / reading, payload validation). I've added a new check for validation errors in particular, allowing to distinguish invalid 2xx from other processing errors.

This case corresponds to the previous `err.status === ResponseType.InvalidResponse`.

### `throw result.reason` (fallback)

This typically corresponds to the server answering with a properly formatted XRPC error (`XrpcResponseError`) that is not declared in the schema's `errors`.

These will be forwarded "as-is" down stream.

### Note

All `XrpcError` from `@atproto/lex-client` have a `toDownstreamError()` method  which allows them to be safely propagated to the `XrpcServer` catch block. The code here is meant to be consistent with previous behavior rather than allowing the "logical" thing (503/504) to happen.

matthieusieben: Also note `UpstreamFailure` does **not** mean 502. It means that we failed to process the response from the upstream server, including because it answered with and "Invalid Response".

devinivy: Got it, I think in the old system it was specifically a 502.

matthieusieben: `XrpcUpstreamError` (which uses `error` = `UpstreamFailure`) can mean many things:
- failed to read the response bytes fully
- failed to parse declared encoding (invalid json, invalid UTF8 string, etc.)
- 4xx or 5xx with a payload that is not a valid XRPC error JSON payload (`{ error: string; message?: string }`)
- 2xx with an encoding that does not match the one defined in the schema
- 2xx with payload that does not match the schema (has its dedicated class)

> [!NOTE]
>
> We could review the way this error is represented (eg. change the `error`) as `@atproto/lex` is still in beta.

packages/bsky/src/api/app/bsky/graph/getRelationships.ts
```diff
@@ -1,11 +1,14 @@
...
 export default function (server: Server, ctx: AppContext) {
-  server.app.bsky.graph.getRelationships({
+  server.add(app.bsky.graph.getRelationships, {
     handler: async ({ params }) => {
-      const { actor, others = [] } = params
-      if (others.length < 1) {
+      const { others = [] } = params
+
+      const [actor] = await ctx.hydrator.actor.getDids([params.actor])
+      if (!actor || others.length < 1) {
```
devinivy, line 11: This change violates the invariant that the indexes of `relationships` match the indexes of `others`.

If we want we can return the `ActorNotFound` error in that case, though.

packages/bsky/src/hydration/actor.ts
```diff
@@ -266,169 +299,260 @@ export class ActorHydrator {
...
     includeTakedowns = false,
   ): Promise<ChatDeclarations> {
-    if (!uris.length) return new HydrationMap<ChatDeclaration>()
+    const map: ChatDeclarations = new HydrationMap()
+    if (!uris.length) return map
+
     const res = await this.dataplane.getActorChatDeclarationRecords({ uris })
-    return uris.reduce((acc, uri, i) => {
+    for (let i = 0; i < uris.length; i++) {
+      const uri = uris[i]
```
devinivy, line 318: I like this style too, but it's not clear to me why all these instances are changed as part of the work to adopt @atproto/lex.  Both are acceptable and the codebase has a bit of an existing style to it.

packages/bsky/src/hydration/label.ts
```diff
@@ -120,61 +124,80 @@ export class LabelHydrator {
...
+    }
+
+    return map
   }
 }
 
-const labelerDidToUri = (did: string): string => {
-  return AtUri.make(did, ids.AppBskyLabelerService, 'self').toString()
+const labelerDidToUri = (did: DidString): AtUriString => {
+  return AtUri.make(did, 'app.bsky.labeler.service', 'self').toString()
```
devinivy: We could use the lex `app.bsky.labeler.service.$type` here.

packages/bsky/src/hydration/util.ts
```diff
@@ -52,77 +58,96 @@ export const mergeNestedMaps = <V, M extends HydrationMap<HydrationMap<V>>>(
...
+
+  const schema = KNOWN_RECORD_TYPES.get(value.$type)
+  if (!schema) {
     return false
   }
+
+  return schema.matches(value)
 }
 
-// @NOTE not parsed into lex format, so will not match lexicon record types on CID and blob values.
```
devinivy, line 99: I believe this was intentional due to the inherent overhead of parsing CIDs, which was quite measurable for us.  Perhaps some more context in the commit where it was introduced: https://github.com/bluesky-social/atproto/pull/1924/changes/ae0180aa379a3855dcd7246c051a65b4c2771689

packages/bsky/src/util/uris.ts
```diff
@@ -23,22 +19,22 @@ export function postUriToThreadgateUri(postUri: string) {
...
+  if (urip.collection === 'app.bsky.feed.post') {
+    urip.collection = 'app.bsky.feed.postgate'
   }
   return urip.toString()
 }
 
-export function uriToDid(uri: string) {
-  return new AtUri(uri).hostname
+export function uriToDid(uri: string): DidString {
+  return new AtUri(uri).did
```
devinivy: I believe this will lead to loads of DID validations that aren't necessary / desirable.

packages/bsky/src/util/uris.ts
```diff
@@ -1,18 +1,14 @@
...
 /**
  * Convert a post URI to a threadgate URI. If the URI is not a valid
  * post URI, return URI unchanged. Threadgate lookups will then fail.
  */
 export function postUriToThreadgateUri(postUri: string) {
   const urip = new AtUri(postUri)
-  if (urip.collection === ids.AppBskyFeedPost) {
-    urip.collection = ids.AppBskyFeedThreadgate
+  if (urip.collection === 'app.bsky.feed.post') {
+    urip.collection = 'app.bsky.feed.threadgate'
```
devinivy: Can we use the lexicon `$type`?

packages/bsky/src/util/uris.ts
```diff
@@ -23,22 +19,22 @@ export function postUriToThreadgateUri(postUri: string) {
  */
 export function postUriToPostgateUri(postUri: string) {
   const urip = new AtUri(postUri)
-  if (urip.collection === ids.AppBskyFeedPost) {
-    urip.collection = ids.AppBskyFeedPostgate
+  if (urip.collection === 'app.bsky.feed.post') {
+    urip.collection = 'app.bsky.feed.postgate'
```
devinivy: Can we use the lexicon `$type`?

packages/bsky/src/views/index.ts
```diff
@@ -798,7 +809,7 @@ export class Views {
     const viewer = state.labelerViewers?.get(did)
     const aggs = state.labelerAggs?.get(did)
 
-    const uri = AtUri.make(did, ids.AppBskyLabelerService, 'self').toString()
+    const uri = AtUri.make(did, 'app.bsky.labeler.service', 'self').toString()
```
devinivy: Can we use the lexicon `$nsid`?

packages/bsky/src/views/index.ts
```diff
@@ -936,14 +947,14 @@ export class Views {
...
-    uri: string,
+    uri: AtUriString,
     state: HydrationState,
     depth = 0,
   ): Un$Typed<PostView> | undefined {
     const post = state.posts?.get(uri)
     if (!post) return
     const parsedUri = new AtUri(uri)
-    const authorDid = parsedUri.hostname
+    const authorDid = parsedUri.did
```
devinivy, line 960: I believe this will perform DID validation, which shouldn't be necessary at this point.

devinivy: One tradeoff of these getters w/ validation is that the performance cost is somewhat invisible.  There's a lot of usage of these throughout and I wont comment on all of them— I am not sure exactly what to do.

Maybe we need a `.did` and `.didSafe` on `AtUri`, where at least cursory validation is done for `.did` when the URI is parsed.  Or of course we could use `.hostname as DidString` to assert our knowledge of the URI format (which in most cases here we know is enforced by the dataplane).

packages/bsky/src/views/index.ts
```diff
@@ -1086,7 +1099,7 @@ export class Views {
     }
   }
 
-  notFoundPost(uri: string): $Typed<NotFoundPost> {
+  notFoundPost(uri: AtUriString): $Typed<NotFoundPost> {
     return {
       $type: 'app.bsky.feed.defs#notFoundPost',
```
devinivy: I feel like the type builders would be great here.

packages/bsky/src/views/index.ts
```diff
@@ -1130,13 +1143,15 @@ export class Views {
     if (!bookmark) return
 
     const atUri = new AtUri(bookmark.subjectUri)
-    if (atUri.collection !== ids.AppBskyFeedPost) return
+    if (atUri.collection !== 'app.bsky.feed.post') return
```
devinivy: Can we use the `$nsid`?

packages/bsky/src/views/index.ts
```diff
@@ -2223,7 +2241,7 @@ export class Views {
       return this.embedDetached(uri)
     }
 
-    if (parsedUri.collection === ids.AppBskyFeedPost) {
+    if (parsedUri.collection === 'app.bsky.feed.post') {
```
devinivy: This and several below could use the `$nsid`.

packages/bsky/src/views/index.ts
```diff
@@ -2395,55 +2416,55 @@ export class Views {
...
-    if (uri.collection === ids.AppBskyFeedPost) {
-      recordInfo = state.posts?.get(notif.uri)
-    } else if (uri.collection === ids.AppBskyFeedLike) {
-      recordInfo = state.likes?.get(notif.uri)
-    } else if (uri.collection === ids.AppBskyFeedRepost) {
-      recordInfo = state.reposts?.get(notif.uri)
-    } else if (uri.collection === ids.AppBskyGraphFollow) {
-      recordInfo = state.follows?.get(notif.uri)
-    } else if (uri.collection === ids.AppBskyGraphVerification) {
+    if (uri.collection === 'app.bsky.feed.post') {
```
devinivy: I think we can use `$nsid` here and below.

packages/bsky/src/views/util.ts
```diff
@@ -58,18 +56,11 @@ type ParsedThreadGate = {
...
   allowFollowing?: boolean
-  allowListUris?: string[]
+  allowListUris?: AtUriString[]
 }
 
-export const cidFromBlobJson = (json: BlobRef) => {
-  if (json instanceof BlobRef) {
-    return json.ref.toString()
-  }
-  // @NOTE below handles the fact that parseRecordBytes() produces raw json rather than lexicon values
```
devinivy, line 68: I commented elsewhere, but just a note that this was intentional due to the overhead we experienced parsing CIDs throughout the read paths.  Though worth noting the overhead may be much less now, and I would much rather simply use CIDs if it doesn't create a meaningful perf regression.

packages/bsky/src/config.ts
```diff
@@ -567,7 +568,7 @@ export class ServerConfig {
   }
 
   get labelsFromIssuerDids() {
-    return this.cfg.labelsFromIssuerDids ?? []
+    return this.cfg.labelsFromIssuerDids?.filter(isDidString) ?? []
```
devinivy: Just a headsup that this may be a hot path.  Could we pre-validate or cast them?

packages/repo/src/car.ts
```diff
@@ -34,16 +26,19 @@ export async function* writeCarStream(
...
   blocks: BlockMap,
-): Promise<Uint8Array> => {
-  const carStream = blocksToCarStream(root, blocks)
-  return streamToBuffer(carStream)
+): Promise<Uint8Array> {
+  const chunks: Uint8Array[] = []
+  for await (const chunk of blocksToCarStream(root, blocks)) {
+    chunks.push(chunk)
+  }
+  return Buffer.concat(chunks)
```
devinivy: There are some pitfalls when you think you have a `Uint8Array` but it is node's `Buffer`.  I was also under the impression you were hoping to get away from nodejs-specific APIs, which weren't used within `streamToBuffer`.

packages/pds/src/api/app/bsky/feed/getPostThread.ts
```diff
@@ -88,14 +86,12 @@ export default function (server: Server, ctx: AppContext) {
...
-  original: OutputSchema,
-  local: LocalRecords,
-): Promise<OutputSchema> => {
+const getPostThreadMunge: MungeFn<
+  app.bsky.feed.getPostThread.$OutputBody
+> = async (localViewer, original, local) => {
   // @TODO if is NotFoundPost, handle similarly to error
   // @NOTE not necessary right now as we never return those for the requested uri
-  if (!isThreadViewPost(original.thread)) {
+  if (!app.bsky.feed.defs.threadViewPost.$matches(original.thread)) {
```
devinivy: I think we may be able to discriminate here rather than revalidate.

packages/pds/src/api/app/bsky/feed/getPostThread.ts
```diff
@@ -152,7 +151,7 @@ const insertIntoThreadReplies = async (
   if (!view.replies) return view
   const replies = await Promise.all(
     view.replies.map(async (reply) =>
-      isThreadViewPost(reply)
+      app.bsky.feed.defs.threadViewPost.$matches(reply)
```
devinivy: I think we may be able to discriminate here as well.  (I wont mark any more in here if there are any.)

packages/pds/src/api/com/atproto/repo/describeRepo.ts
```diff
@@ -1,39 +1,43 @@
...
-import { Server } from '../../../../lexicon'
+import { com } from '../../../../lexicons/index.js'
 import { assertRepoAvailability } from '../sync/util'
 
 export default function (server: Server, ctx: AppContext) {
-  server.com.atproto.repo.describeRepo(async ({ params }) => {
-    const { repo } = params
+  server.add(
+    com.atproto.repo.describeRepo,
+    async ({ params }): Promise<com.atproto.repo.describeRepo.$Output> => {
```
devinivy, line 11: I expect this explicit type isn't necessary.

packages/pds/src/api/com/atproto/repo/importRepo.ts
```diff
@@ -90,35 +89,11 @@ export default function (server: Server, ctx: AppContext) {
               rev,
               now,
             )
-            const recordBlobs = findBlobRefs(parsedRecord)
+            const recordBlobs = Array.from(enumBlobRefs(parsedRecord))
```
devinivy: I believe this will drop legacy blob refs during import, which would impact a fair number of repositories including my own!  Can we pass the legacy flag here?

packages/pds/src/api/com/atproto/repo/listMissingBlobs.ts
```diff
@@ -1,23 +1,25 @@
...
     auth: ctx.authVerifier.authorization({
       authorize: () => {
         // always allow
       },
     }),
     handler: async ({ auth, params }) => {
       const { did } = auth.credentials
-      const { limit, cursor } = params
+      // @TODO infer that "limit" cannot be undefined thanks to "default" in schema
+      const { limit = 500, cursor } = params
```
devinivy: This is no longer a limitation.
```suggestion
      const { limit, cursor } = params
```

packages/pds/src/api/com/atproto/repo/listRecords.ts
```diff
@@ -1,36 +1,39 @@
...
 import { AppContext } from '../../../../context'
-import { Server } from '../../../../lexicon'
+import { com } from '../../../../lexicons/index.js'
 
 export default function (server: Server, ctx: AppContext) {
-  server.com.atproto.repo.listRecords(async ({ params }) => {
-    const { repo, collection, limit = 50, cursor, reverse = false } = params
+  server.add(
+    com.atproto.repo.listRecords,
+    async ({ params }): Promise<com.atproto.repo.listRecords.$Output> => {
```
devinivy: I think we can remove the explicit type here.

packages/pds/src/api/com/atproto/repo/putRecord.ts
```diff
@@ -81,20 +81,16 @@ export default function (server: Server, ctx: AppContext) {
...
       const { commit, write } = await ctx.actorStore.transact(
         did,
         async (actorTxn) => {
           const current = await actorTxn.record.getRecord(uri, null, true)
           const isUpdate = current !== null
 
-          // @TODO temporaray hack for legacy blob refs in profiles - remove after migrating legacy blobs
-          if (isUpdate && collection === ids.AppBskyActorProfile) {
-            await updateProfileLegacyBlobRef(actorTxn, record)
-          }
```
devinivy, line 97: To my knowledge legacy blobs were never migrated (imagine all the broken strong refs), and removing this may have an impact for reactivating users.

packages/pds/src/api/com/atproto/server/createAppPassword.ts
```diff
@@ -16,31 +16,32 @@ export default function (server: Server, ctx: AppContext) {
...
-            await ctx.entrywayAuthHeaders(
-              req,
-              auth.credentials.did,
-              ids.ComAtprotoServerCreateAppPassword,
-            ),
-          ),
-        )
-      }
+    handler: entrywayClient
+      ? async ({ input: { body }, auth, req }) => {
```
devinivy: I think we've moved away from the ternary assignment of `handler` elsewhere and will want to adopt here too.

packages/pds/src/api/com/atproto/server/createInviteCode.ts
```diff
@@ -1,30 +1,31 @@
...
     auth: ctx.authVerifier.adminToken,
-    handler: async ({ input }) => {
-      if (ctx.cfg.entryway) {
-        throw new InvalidRequestError(
-          'Account invites are managed by the entryway service',
-        )
-      }
-      const { useCount, forAccount = 'admin' } = input.body
+    handler: ctx.cfg.entryway
+      ? () => {
```
devinivy: I think we've moved away from this pattern elsewhere.

packages/pds/src/api/com/atproto/server/createInviteCodes.ts
```diff
@@ -1,34 +1,35 @@
...
-    handler: async ({ input }) => {
-      if (ctx.cfg.entryway) {
-        throw new InvalidRequestError(
-          'Account invites are managed by the entryway service',
-        )
-      }
-
-      const { codeCount, useCount } = input.body
+    handler: ctx.cfg.entryway
+      ? () => {
```
devinivy: I think we've moved away from this pattern over the course of review, wont comment on any more of these 👍

packages/pds/src/api/com/atproto/server/createSession.ts
```diff
@@ -22,54 +23,61 @@ export default function (server: Server, ctx: AppContext) {
...
+          const { headers } = ctx.entrywayPassthruHeaders(req)
+          return entrywayClient.xrpc(com.atproto.server.createSession, {
+            validateResponse: false, // ignore invalid upstream responses
+            headers,
+            body,
+          })
+        }
+      : async ({
+          input: { body },
+        }): Promise<com.atproto.server.createSession.$Output> => {
```
devinivy: Can likely get rid of the explicit type.

packages/pds/src/api/com/atproto/sync/listBlobs.ts
```diff
@@ -1,27 +1,28 @@
...
+  server.add(com.atproto.sync.listBlobs, {
     auth: ctx.authVerifier.authorizationOrAdminTokenOptional({
       additional: [AuthScope.Takendown],
       authorize: () => {
         // always allow
       },
     }),
     handler: async ({ params, auth }) => {
-      const { did, since, limit, cursor } = params
+      const { did, since, limit = 500, cursor } = params
```
devinivy: I think we can remove the redundant default from here now.
```suggestion
      const { did, since, limit, cursor } = params
```

packages/pds/tests/proxied/read-after-write.test.ts
```diff
@@ -263,16 +264,17 @@ describe('proxy read after write', () => {
...
-    assert(isThreadViewPost(replies[0]))
+    assert(AppBskyFeedDefs.isThreadViewPost(replies[0]))
     expect(replies[0].post.uri).toEqual(replyRes.uri)
     const embed = replies[0].post.embed
-    assert(isRecordEmbedView(embed))
-    assert('uri' in embed.record) // @TODO: assert based in "$type"
-    expect(embed.record.uri).toEqual(sc.posts[alice][0].ref.uriStr)
+    assert(AppBskyEmbedRecord.isView(embed))
+    const { record } = embed
+    assert('uri' in record) // @TODO: assert based in "$type"
```
devinivy, line 276: Perhaps this todo is resolvable now?

packages/pds/tests/crud.test.ts
```diff
@@ -397,205 +395,209 @@ describe('crud operations', () => {
...
       expect(profile.value).toEqual({
-        $type: ids.AppBskyActorProfile,
+        $type: 'app.bsky.actor.profile',
         displayName: 'Robert',
         description: 'Dog lover',
       })
     })
 
-    // @TODO remove after migrating legacy blobs
-    it('updates a legacy blob ref when updating profile', async () => {
```
devinivy, line 534: This migration hasn't happened, my best guess is that we want to preserve this behavior.

packages/pds/tests/preferences.test.ts
```diff
@@ -55,7 +55,10 @@ describe('user preferences', () => {
   it('only gets preferences in app.bsky namespace.', async () => {
     await network.pds.ctx.actorStore.transact(sc.dids.alice, (store) =>
       store.pref.putPreferences(
-        [{ $type: 'com.atproto.server.defs#unknown' }],
+        [
+          // @ts-expect-error unspecced pref
+          { $type: 'com.atproto.server.defs#unknown' },
```
devinivy: I expect there's a more well-trodden path for using the "open" part of open unions.  Does it just need to be annotated with `as Unknown$Type`?

packages/bsky/src/hydration/graph.ts
```diff
@@ -172,51 +194,78 @@ export class GraphHydrator {
...
-    if (!uris.length) return new HydrationMap<Block>()
-    const res = await this.dataplane.getBlockRecords({ uris })
-    return uris.reduce((acc, uri, i) => {
-      const record = parseRecord<BlockRecord>(res.records[i], includeTakedowns)
-      return acc.set(uri, record ?? null)
-    }, new HydrationMap<Block>())
+  ): Promise<HydrationMap<AtUriString, Block>> {
+    const map = new HydrationMap<AtUriString, Block>()
+
+    if (uris.length) {
```
devinivy: There are some methods in here with these stylistic changes still kicking around.

packages/bsky/src/hydration/actor.ts
```diff
@@ -266,169 +299,260 @@ export class ActorHydrator {
...
-    if (!dids.length) return new HydrationMap<ProfileViewerState>()
+    const map: ProfileViewerStates = new HydrationMap()
+    if (!actors.length) return map
+
+    // @NOTE we are not using getDidsDefined here because we want to preserve
+    // the order of the input and be able to handle cases where some of the
+    // inputs don't resolve to a DID.
+    const actorDids = await this.getDids(actors).catch(() => {
+      // Failed to resolve handles. Ignore handles from input.
+      return actors.map((actor) => (isDidIdentifier(actor) ? actor : undefined))
```
devinivy: Does this case occur?  I would have expected failures of individual items not to propagate here, instead return an empty result.

packages/bsky/src/hydration/actor.ts
```diff
@@ -266,169 +299,260 @@ export class ActorHydrator {
...
+    return map
   }
 
   // "naive" because this method does not verify the existence of the list itself
   // a later check in the main hydrator will remove list uris that have been deleted or
   // repurposed to "curate lists"
   async getProfileViewerStatesNaive(
-    dids: string[],
-    viewer: string,
+    actors: AtIdentifierString[],
```
devinivy, line 391: There are enough changes moving from `dids` to `actors` that we should probably make sure it gets some test coverage.  Implementation issues can lead to pretty confusing behavior, unfortunately.

It's also fine if we want to treat handles passed into that one endpoint where this was relevant (I forget which one) as though they are not resolvable—since that was the existing behavior which doesn't need to be fixed in this work—and leave a TODO.

packages/bsky/src/hydration/actor.ts
matthieusieben: The reason it is like this is because as some intermediary state of this PR it made more sense; and most code paths were moved to this syntax (so most places are consistent with this new style now). I didn't see the need to revert them back to the old style but I can if you don't feel comfortable with these changes.

packages/bsky/src/hydration/actor.ts
matthieusieben: this is mostly being "safe". I'll revert this bo be more consistent with the previous behavior (ie. not resolving DIDs)

packages/bsky/src/hydration/actor.ts
matthieusieben: I'll remove the `getDids` call and leave a TODO.

packages/bsky/src/util/uris.ts
matthieusieben: Alright will use a simple assert and type cast + comment. I can also use a simple type cast if you think that's enough.

packages/pds/src/api/com/atproto/repo/describeRepo.ts
matthieusieben: It is needed because of the `ts-expect-error` bellow

packages/pds/src/api/com/atproto/repo/importRepo.ts
matthieusieben: done

packages/pds/src/api/com/atproto/server/createSession.ts
matthieusieben: no, because of the ts-expect-error

packages/pds/tests/proxied/read-after-write.test.ts
matthieusieben: Not really, because the test are still relying on `@atproto/api`

packages/repo/src/car.ts
matthieusieben: That's right ! i'll fix this

packages/bsky/src/util/uris.ts
devinivy: I think a type cast here with a comment that `uri` is known to be in the DID (rather than handle) form would be alright!  The reason it's alright is usually that we know the URIs that the dataplane hands back are always in DID form.

packages/repo/bench/repo.bench.ts
devinivy: That's fine, I suppose we can always pull them out of history if we need to bring them back.

packages/bsky/src/hydration/actor.ts
devinivy: Most code paths in the bsky package or most code paths in the broader atproto repo?

I don't think we need to revert them at this point, but in general I feel we should minimize these kinds of sweeping stylistic changes, particularly in (very) large chunks of work.

packages/bsky/src/hydration/util.ts
matthieusieben: This is awkward. If we don't parse CIDs, the `isValidRecord` **will** fail for any record type we use that relies on CIDs. Note that we should be able to move away from multiformats 9 (thanks to the `Cid` interface), which should solve the performance issue!

Should we merge this as-is, and work on using another `Cid` implementation as a separate track?

packages/bsky/src/views/util.ts
matthieusieben: I'll leave this comment open until we resolve this in https://github.com/bluesky-social/atproto/pull/4408#discussion_r2926035045

packages/pds/src/api/com/atproto/repo/putRecord.ts
matthieusieben: Alright will revert this.

packages/pds/tests/crud.test.ts
matthieusieben: will revert this

packages/bsky/src/views/index.ts
matthieusieben: I was able to make the `.did` getter much more efficient

packages/bsky/src/hydration/util.ts
devinivy: Yeah let's move forward, but it is something to track during the rollout of this version of the appview.  If it causes any issues we will probably need to resolve them in relatively short order.

PR #4415, opened 2025-12-04
Experiment: wire-through germ chat declaration in bsky appview backend

bnewbold:
This wires up bsky appview support for hydrating through Germ chat declaration context in profile views. The goal is to support showing Germ chat support status. This all works somewhat similar to the `chat.bsky.*` declaration record, and the bluesky account status (aka, "live streaming") feature.

### DIFF
--- a/.changeset/fruity-frogs-boil.md
+++ b/.changeset/fruity-frogs-boil.md
@@ -1,0 +1,6 @@
+---
+'@atproto/bsky': patch
+'@atproto/api': patch
+---
+
+support for Germ Networks chat declaration records


packages/bsky/src/hydration/actor.ts
```diff
@@ -5,6 +5,7 @@ import { Record as ProfileRecord } from '../lexicon/types/app/bsky/actor/profile
 import { Record as StatusRecord } from '../lexicon/types/app/bsky/actor/status'
 import { Record as NotificationDeclarationRecord } from '../lexicon/types/app/bsky/notification/declaration'
 import { Record as ChatDeclarationRecord } from '../lexicon/types/chat/bsky/actor/declaration'
+import { Record as GermDeclarationRecord, MessageMe as GermDeclarationMessageMe } from '../lexicon/types/com/germnetwork/declaration'
```
bnewbold: note to self: maybe don't need the `MessageMe` import any more?

packages/bsky/src/hydration/actor.ts
```diff
@@ -268,11 +277,27 @@ export class ActorHydrator {
...
+      return acc.set(uri, record ?? null)
+    }, new HydrationMap<GermDeclaration>())
+  }
+
   async getNotificationDeclarations(
     uris: string[],
     includeTakedowns = false,
   ): Promise<NotificationDeclarations> {
     if (!uris.length) return new HydrationMap<NotificationDeclaration>()
+    // TODO: should this be getNotificationDeclarationRecords?
```
bnewbold: @rafaeleyng came across this while reading the code

packages/bsky/src/views/index.ts
```diff
@@ -366,6 +369,7 @@ export class Views {
           ? { allowIncoming: actor.allowIncomingChatsFrom }
           : undefined,
         activitySubscription: this.profileAssociatedActivitySubscription(actor),
+        // NOTE: not hydrating germ on "basic"
```
bnewbold: My thinking was that the "basic" profileView response doesn't really need this particular extra metadata.

But it is probably more correct (eg, schema compliant) and expected to hydrate the germ context through in both cases?

packages/bsky/tests/views/__snapshots__/profile.test.ts.snap
```diff
@@ -529,6 +529,8 @@ Object {
 }
 `;
 
+exports[`pds profile views germ returns germ record if it does exist 1`] = `undefined`;
```
bnewbold: this is wrong/failing

packages/lex/lex-client/scripts/lex-build.mjs
```diff
@@ -19,6 +19,7 @@ Promise.all([
       'com.atproto.repo.putRecord',
       'com.atproto.repo.listRecords',
       'com.atproto.repo.uploadBlob',
+      'com.germnetwork.declaration',
```
bnewbold: I'm not sure if this is necessary, now that `app.bsky.actor.profile` isn't directly referencing `com.germnetwork.declaration`

packages/bsky/src/hydration/actor.ts
rafaeleyng: Yes, I'll prepare a PR. Thanks for that

rafaeleyng: We're handling it on https://github.com/bluesky-social/atproto/pull/4444

lexicons/com/germnetwork/declaration.json
```diff
@@ -0,0 +1,45 @@
...
+      "type": "object",
+      "required": ["showButtonTo", "messageMeUrl"],
+      "properties": {
+        "messageMeUrl": {
+          "type": "string",
+          "format": "uri"
+        },
+        "showButtonTo": {
+          "type": "string",
+          "knownValues": ["usersIFollow", "everyone"]
```
dholms, line 40: i'm inclined to use similar values to chat declaration: at least `following` as opposed to `usersIFollow`

lexicons/app/bsky/actor/defs.json
```diff
@@ -149,7 +149,8 @@
         "activitySubscription": {
           "type": "ref",
           "ref": "#profileAssociatedActivitySubscription"
-        }
+        },
+        "germ": { "type": "ref", "ref": "#profileAssociatedGerm" }
```
dholms, line 153: i kinda wonder if there's a more generic way to do this. I can just see these views accumulating a ton of "associateds" over time

 idk don't want to over-design this tho

packages/bsky/src/data-plane/server/indexing/plugins/germ-declaration.ts
```diff
@@ -0,0 +1,63 @@
+import { CID } from 'multiformats/cid'
+import { AtUri } from '@atproto/syntax'
+import * as lex from '../../../../lexicon/lexicons'
+import { BackgroundQueue } from '../../background'
+import { Database } from '../../db'
+import { DatabaseSchema } from '../../db/database-schema'
+import { RecordProcessor } from '../processor'
+
+// @NOTE this indexer is a placeholder to ensure it gets indexed in the generic records table
```
dholms, line 9: yeah this is fine for now. we should develop a pattern if we start doing this more

packages/bsky/proto/bsky.proto
```diff
@@ -487,6 +495,7 @@ message ActorInfo {
   Record status_record = 17;
   string allow_activity_subscriptions_from = 18;
   optional AgeAssuranceStatus age_assurance_status = 19;
+  Record germ_record = 20;
```
dholms: how big are the continuity proofs in the record? & do we want to be passing that whole thing from the dataplane to the application if it doesn't go in the view?

packages/bsky/src/views/index.ts
dholms: we already have it hydrated right (because it comes in with actor info)? and it's pretty lightweight in the view so i think i lean towards just adding it

packages/bsky/proto/bsky.proto
germ-mark: It’s a placeholder; the intent is 2 signatures (from predecessor and successor keys). Modal length of the list should be one.

The main data in the record for app view consumption is the message me object.

lexicons/com/germnetwork/declaration.json
bnewbold: I'm open to that; we should coordinate with Germ team on any final schema tweaks (they designed and are authoritative, we've iterated a bit together already).

lexicons/app/bsky/actor/defs.json
bnewbold: The "associated" object does work as a somewhat generic place to plop these in. I could see there being more/better abstraction around this, but don't want to hold up this particular experiment/collab for that.

packages/bsky/src/data-plane/server/indexing/plugins/germ-declaration.ts
bnewbold: This was copied verbatim from `chat-declaration.ts`, but yeah agree.

packages/bsky/proto/bsky.proto
bnewbold: They are some extra bytes on the dataplane interface. Some more thinking:

- having the full/raw record in the dataplane interface means we don't need to make future changes in the dataplane backend if the record schema changes. I wanted to follow the precedent of other declaration records getting passed through and not have much "business logic" in the dataplane backend
- not all accounts will have this account configured, at least to start
- if it is a perf/resource problem in prod it should not be too hard to update and only pass the messageMe info (aka, this isn't a "one way" decision, and doesn't impact external APIs)
- not all accounts will have these records to start (though there will probably be bias where big/popular/active accounts are more likely to have them)
- this will burn some extra cache bytes; but only proportional to the number accounts with the record
- we have really high throughput between appview TS code and dataplane
- bytes on the wire to clients is more important to me, and we do keep it small in that context

bnewbold:
Tests have been fixed up, and I've done some basic manual testing with our other dataplane implementation: germ declarations come back over the gRPC interface and get correctly parsed and transformed for the API response. Deleting the declaration results in nothing coming back.

There are some small-ish style questions under discussion here, and this probably needs review from an app team person, but I think is mostly ready to go from a backend perspective.

lexicons/app/bsky/actor/defs.json
estrattonbailey: Counter point: once we release this, I bet we'll see more calls to integrate other platforms and services. If it's simply a URL/permissions for now, seems reasonable to make it generic.

bnewbold: I'd like to come up with generic ways to link accounts (in the abstract, eg as data even if it isn't the product), and give folks more control over buttons/links on their profiles (in the product).

I see this effort with Germ as a prototype to figure out what the requirements and UX considerations will be around that. Eg, I think experience here will inform the design of a more powerful/generic system.

matthieusieben: Is there a way to use a (list?) of open unions for this ?

```suggestion
        "externalProfiles": {
          "type": "array",
          "item": {
            "type": "union",
            "refs": ["#externaProfileGerm"]
          }
        }
```

bnewbold: Is that better? The "list of open union" pattern is used for the current preferences API. I feel like that has not been ergonomic to work with, and incurs (small) `$type` discriminator overhead.

I think we should not be afraid about adding non-required fields which might be deprecated later (for a more generic account linkage system). Especially in API responses (versus record schemas).

matthieusieben: I personally like it better as it is more generic and future proof. It also allows to describe more than one item which might be desirable is some (yet unforeseen) cases.

lexicons/app/bsky/actor/defs.json
```diff
@@ -162,6 +163,20 @@
...
+      "type": "object",
+      "required": ["showButtonTo", "messageMeUrl"],
+      "properties": {
+        "messageMeUrl": {
+          "type": "string",
+          "format": "uri"
+        },
+        "showButtonTo": {
+          "type": "string",
+          "knownValues": ["usersIFollow", "everyone"]
```
ThisIsMissEm, line 176:
```suggestion
          "knownValues": ["mutuals", "following", "followers", "everyone"]
```

But maybe this should be a list with multiple known values?

lexicons/app/bsky/actor/defs.json
estrattonbailey: (finally returning to +1 this) I +1 this, array of open union is great imo

lexicons/app/bsky/actor/defs.json
bnewbold: I'd defer to Germ folks on this, but my personal inclination would be to keep it a single string value for simplicity and to avoid an iteration on the dataplane gRPC interface right now.

lexicons/app/bsky/actor/defs.json
bnewbold: Chatted with this a bit with @dholms and we think this is a reasonable suggestion but not totally settled either way, and not a blocking issue.

PR #4421, opened 2025-12-05
repo: mst key example fixes

bnewbold:
This is the pair of the indigo PR: https://github.com/bluesky-social/indigo/pull/1222

From MST height example fixes: https://github.com/bluesky-social/atproto-interop-tests/pull/11

I updated the interop test fixtures in the go code, by updating the keys, and assuming the go code is correct and taking the generated before/after commits and diff blocks.

But then when I pull things over in this repo, I get errors. So either I messed something up copy/pasting, or there is an actual implementation difference.

@dholms: if you could confirm that the new keys actually have the expected heights, and then figure out why these diff blocks don't match? It's possible the go code is outputting extra blocks or something.


dholms:
verified that keys all have correct height

PR #4429, opened 2025-12-09
[APP-1667] Handle Rolodex service errors, translate to `XRPCError`

estrattonbailey:
This PR maps `ConnectError`s received from the Rolodex service to known `XRPCError` classes, as defined in the lexicons. The error `name` strings match those returned from Rolodex, and we will need to keep these in sync in the future.

### DIFF
--- a/lexicons/app/bsky/contact/getMatches.json
+++ b/lexicons/app/bsky/contact/getMatches.json
@@ -42,10 +42,7 @@
           "name": "INVALID_LIMIT"
         },
         {
-          "name": "INVALID_INCOMING_CURSOR"
-        },
-        {
-          "name": "INVALID_OUTGOING_CURSOR"
+          "name": "INVALID_CURSOR"
         },
         {
           "name": "INTERNAL_ERROR"
--- a/lexicons/app/bsky/contact/importContacts.json
+++ b/lexicons/app/bsky/contact/importContacts.json
@@ -52,7 +52,7 @@
           "name": "INVALID_CONTACTS"
         },
         {
-          "name": "INVALID_CONTACTS_COUNT"
+          "name": "TOO_MANY_CONTACTS"
         },
         {
           "name": "INVALID_TOKEN"
--- a/packages/api/src/client/lexicons.ts
+++ b/packages/api/src/client/lexicons.ts
@@ -1908,10 +1908,7 @@ export const schemaDict = {
             name: 'INVALID_LIMIT',
           },
           {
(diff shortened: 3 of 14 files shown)


lexicons/app/bsky/contact/getMatches.json
```diff
@@ -36,8 +36,19 @@
...
+          "name": "INVALID_DID"
+        },
+        {
+          "name": "INVALID_LIMIT"
+        },
+        {
+          "name": "INVALID_INCOMING_CURSOR"
+        },
+        {
+          "name": "INVALID_OUTGOING_CURSOR"
```
rafaeleyng: Same as I commented on rolodex, this one might not make sense.

lexicons/app/bsky/contact/getMatches.json
```diff
@@ -36,8 +36,19 @@
       },
       "errors": [
         {
-          "name": "TODO",
-          "description": "TODO"
+          "name": "INVALID_DID"
```
rafaeleyng, line 39: In practice, this error shouldn't happen, as the appview should never send to rolodex an invalid DID as the user session, but that's fine to keep here.

Or we could do a per-route way of checking this. Because it will be repeated for many of these endpoints, and it doesn't necessarily make sense for any of them.

lexicons/app/bsky/contact/importContacts.json
```diff
@@ -46,8 +46,19 @@
...
         {
-          "name": "TODO",
-          "description": "TODO"
+          "name": "INVALID_DID"
+        },
+        {
+          "name": "INVALID_CONTACTS"
+        },
+        {
+          "name": "INVALID_CONTACTS_COUNT"
```
rafaeleyng: In practice, this means "too many contacts". The name isn't great, I did it because before you did the 2 error types, I was using the name (invalid vs failure) to differentiate between internal and public errors. Maybe now that we don't have to rely on the name you could rename this to something clearer here and in Rolodex?

E.g., `TOO_MANY_CONTACTS`?

rafaeleyng (commented):
Looks good!
I suggested some changes. LMK, and I can stamp as-is if you disagree with them.

lexicons/app/bsky/contact/getMatches.json
estrattonbailey: I think it's OK to include. The appview is the client in this case I guess, and it shouldn't happen, but if it does it tells us why.

Not sure what you mean by per-route 🤔

PR #4434, opened 2025-12-09
Sitemap Handlers

jcalabro:
Linear: https://linear.app/blueskyweb/project/sitemaps-0c9e8bd80a0b

Implements sitemaps basically as a pass-through of the gzipped xml coming from the data plane.

These sitemaps only contain certain users, not all.

Running the appview locally against the canary data plane unit, I get:

```
$ curl -s http://localhost:2584/sitemap/users.xml.gz --output - | gunzip -d | head
<?xml version="1.0" encoding="UTF-8"?>
<sitemapindex xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
  <sitemap>
    <loc>https://bsky.app/sitemap/users/2025-12-07/1.xml.gz</loc>
  </sitemap>
  <sitemap>
    <loc>https://bsky.app/sitemap/users/2025-12-06/1.xml.gz</loc>
  </sitemap>
  <sitemap>
    <loc>https://bsky.app/sitemap/users/2025-12-05/1.xml.gz</loc>

$ curl -s http://localhost:2584/sitemap/users/2022-11-17/1.xml.gz --output - | gunzip -d
<?xml version="1.0" encoding="UTF-8"?>
<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
  <url>
    <loc>https://bsky.app/profile/jay.bsky.team</loc>
    <changefreq>daily</changefreq>
  </url>
  <url>
    <loc>https://bsky.app/profile/dholms.xyz</loc>
    <changefreq>daily</changefreq>
  </url>
  <url>
    <loc>https://bsky.app/profile/why.bsky.team</loc>
    <changefreq>daily</changefreq>
  </url>
  <url>
    <loc>https://bsky.app/profile/divy.zone</loc>
    <changefreq>daily</changefreq>
  </url>
  <url>
    <loc>https://bsky.app/profile/pfrazee.com</loc>
    <changefreq>daily</changefreq>
  </url>
</urlset>
```

### DIFF
--- a/packages/bsky/src/api/sitemap.ts
+++ b/packages/bsky/src/api/sitemap.ts
@@ -1,6 +1,7 @@
-import { Timestamp } from '@bufbuild/protobuf'
-import express, { RequestHandler, Router } from 'express'
 import { Readable } from 'node:stream'
+import { Timestamp } from '@bufbuild/protobuf'
+import { Code, ConnectError } from '@connectrpc/connect'
+import express, { RequestHandler, Router } from 'express'
 import { AppContext } from '../context'
 import { httpLogger as log } from '../logger'
 import { SitemapPageType } from '../proto/bsky_pb'
@@ -8,7 +9,10 @@ import { SitemapPageType } from '../proto/bsky_pb'
 export const createRouter = (ctx: AppContext): Router => {
   const router = Router()
   router.get('/external/sitemap/users.xml.gz', userIndexHandler(ctx))
-  router.get('/external/sitemap/users/:date/:bucket.xml.gz', userPageHandler(ctx))
+  router.get(
+    '/external/sitemap/users/:date/:bucket.xml.gz',
+    userPageHandler(ctx),
+  )
   return router
 }
 
@@ -63,6 +67,9 @@ const userPageHandler =
       res.set('Content-Encoding', 'gzip')
       Readable.from(Buffer.from(result.sitemap)).pipe(res)
     } catch (err) {
+      if (err instanceof ConnectError && err.code === Code.NotFound) {
+        return res.status(404).send('Sitemap page not found')
+      }
       log.error({ err }, 'failed to get sitemap page')
       return res.status(500).send('Internal Server Error')
     }
--- a/packages/bsky/src/data-plane/server/routes/sitemap.ts
+++ b/packages/bsky/src/data-plane/server/routes/sitemap.ts
@@ -1,6 +1,7 @@
 import { gzipSync } from 'node:zlib'
-import { ServiceImpl } from '@connectrpc/connect'
+import { Code, ConnectError, ServiceImpl } from '@connectrpc/connect'
 import { Service } from '../../../proto/bsky_connect'
+import { GetSitemapPageRequest } from '../../../proto/bsky_pb'
 
 const MOCK_SITEMAP_INDEX = `<?xml version="1.0" encoding="UTF-8"?>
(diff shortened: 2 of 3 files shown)


packages/bsky/src/index.ts
```diff
@@ -225,6 +225,11 @@ export class BskyAppView {
...
     app.use(blobResolver.createMiddleware(ctx))
     app.use(imageServer.createMiddleware(ctx, { prefix: '/img/' }))
+
+    if (config.dataplaneUrls.length > 0 || config.dataplaneUrlsEtcdKeyPrefix) {
+      app.use(sitemap.createRouter(ctx))
+    }
+
     app.use(server.xrpc.router)
     app.use(error.handler)
     app.use('/external', external.createRouter(ctx))
```
rafaeleyng, line 235: I had this pattern of doing `/external` for where 3rd-parties will call us. Might want to put sitemaps under this, but not a big deal either way.

packages/bsky/src/api/sitemap.ts
```diff
@@ -0,0 +1,68 @@
...
+  router.get('/sitemap/users.xml.gz', indexHandler(ctx))
+  router.get('/sitemap/users/:date/:bucket.xml.gz', pageHandler(ctx))
+  return router
+}
+
+const indexHandler =
+  (ctx: AppContext): RequestHandler =>
+  async (_req: express.Request, res: express.Response) => {
+    try {
+      const result = await ctx.dataplane.getSitemapIndex({
```
rafaeleyng, line 23: No need to change anything here now, but I wonder if we could just stream this result to the response we send. This is usually possible with Node streams, but not sure about the gRPC interface. Would prevent us from loading the whole files in-memory.

packages/bsky/src/api/sitemap.ts
```diff
@@ -0,0 +1,68 @@
...
+import { SitemapPageType } from '../proto/bsky_pb'
+
+export const createRouter = (ctx: AppContext): Router => {
+  const router = Router()
+  router.get('/sitemap/users.xml.gz', indexHandler(ctx))
+  router.get('/sitemap/users/:date/:bucket.xml.gz', pageHandler(ctx))
+  return router
+}
+
+const indexHandler =
```
rafaeleyng: If we add a new index for posts, we'll want to rename the handlers to e.g. `postsIndexHandler`, but this is fine for now.

packages/bsky/src/data-plane/server/routes/sitemap.ts
```diff
@@ -0,0 +1,30 @@
...
+  </url>
+</urlset>`
+
+export default (): Partial<ServiceImpl<typeof Service>> => ({
+  async getSitemapIndex() {
+    return {
+      sitemap: gzipSync(Buffer.from(MOCK_SITEMAP_INDEX)),
+    }
+  },
+  async getSitemapPage() {
```
rafaeleyng: Could make this check if the expected date and bucket match `2025-01-01/1`, then fail if not. Then you could add a test for the not-found case. Feel free to ignore.

packages/bsky/src/api/sitemap.ts
jcalabro: We can use a slightly more efficient copy function in the HTTP handler, but yeah since it's not a full streaming [gRPC streaming RPC](https://grpc.io/docs/what-is-grpc/core-concepts/#server-streaming-rpc), we won't get all the benefit (this is still slightly more efficient I think though)

PR #4440, opened 2025-12-10
add recId to SuggestedUserSkeleton

iwsmith:


.changeset/sharp-pillows-wink.md
```diff
@@ -0,0 +1,6 @@
+---
+'@atproto/api': patch
+'@atproto/bsky': patch
+---
+
+Add recID field to getSuggestedUsers
```
estrattonbailey: Small typo, and might as well tick these since it's translate to markdown
```suggestion
Add `recId` field to `getSuggestedUsers` and `getSuggestedUsersSkeleton`
```

lexicons/app/bsky/unspecced/getSuggestedUsersSkeleton.json
```diff
@@ -37,6 +37,10 @@
                 "type": "string",
                 "format": "did"
               }
+            },
+            "recId": {
+              "type": "integer",
+              "description": "Snowflake for this recommendation, use when submitting recommendation events."
```
estrattonbailey, line 43: Looks like this skeleton method is only used in one place: [getSuggestedUsers](https://github.com/bluesky-social/atproto/blob/d55b4b7b8a2e7974163208f6ab81ee902bbed666/packages/bsky/src/api/app/bsky/unspecced/getSuggestedUsers.ts#L61). So we'll need to add this lex def to that endpoint as well, and pipe the property through to the API response object.

jcalabro (approved):
@iwsmith @mozzius @estrattonbailey good to merge?

estrattonbailey (approved):
That should do it! Thanks @cuducos 👍

PR #4458, opened 2025-12-15
new `tools.ozone.report.createReport` endpoint (same as `com.atproto.moderation.createReport`)

bnewbold:
In some sense this has been planned for a while: moderation reporting schemas are pretty specific to Ozone, and are not really defined as part of the protocol itself.

The current motivation around this is to be able to define a permission-set under the `tools.ozone.*` namespace giving clients permission to submit moderation reports. We'd particularly like to document this as part of app development in general.

This is just the lexicon; will need to run codegen and add changesets if this PR is approved. Then we can get this implemented in Ozone as an alternative endpoint.

Some small decisions:

- I had this lexicon schema refer to `tools.ozone.report.defs#reasonType`. it might be marginally easier for code updates if it references `com.atproto.moderation.defs#reasonType` instead? it doesn't matter either way for OAuth permissions; it just feels more consistent to do it this way
- because of the above I copied some old `com.atproto.moderation.defs` reason defs over. this is just a `knownValues` string, but it maybe helps for type generation and eases transition? this might not be needed
- I could have also copied over `#subjectType` (another string with `knownValues`), and updates some other ozone schemas to ref the more local copy. That one should maybe go under `tools.ozone.moderation.defs` (not `.report.`), and doesn't impact OAuth stuff, so I didn't bother

### DIFF
--- a/packages/pds/src/pipethrough.ts
+++ b/packages/pds/src/pipethrough.ts
@@ -593,7 +593,6 @@ const defaultService = (
         serviceInfo: ctx.cfg.modService,
       }
     case ids.ComAtprotoModerationCreateReport:
-    case ids.ToolsOzoneReportCreateReport:
       return {
         serviceId: 'atproto_labeler',
         serviceInfo: ctx.cfg.reportService,
--- a/packages/ozone/tests/report-reason.test.ts
+++ b/packages/ozone/tests/report-reason.test.ts
@@ -82,7 +82,10 @@ describe('report reason', () => {
           },
           {
             encoding: 'application/json',
-            headers: sc.getHeaders(sc.dids.alice),
+            headers: {
+              ...sc.getHeaders(sc.dids.alice),
+              'atproto-proxy': `${network.ozone.ctx.cfg.service.did}#atproto_labeler`,
+            },
           },
         ),
       ])


foysalit:
any objection on using `tools.ozone.moderation.createReport` instead of `tools.ozone.report.createReport`? makes sense to me to use the existing namespace.

bnewbold:
@foysalit I don't have a strong objection.

I used `tools.ozone.report.*` because it already exits and has the `tools.ozone.report.defs#reasonType` definition under it, which is relevant to `createReport`. If that had been put under `tools.ozone.moderation.defs` then there would definitely be a clean mapping from the old com.atproto namespace, but give that it exists it feels a bit weird to split basic reporting functionality across two namespaces.

foysalit:
yeah that makes sense. the endpoint is working now. i'll merge once you're back @bnewbold

packages/pds/src/pipethrough.ts
```diff
@@ -593,6 +593,7 @@ const defaultService = (
         serviceInfo: ctx.cfg.modService,
       }
     case ids.ComAtprotoModerationCreateReport:
+    case ids.ToolsOzoneReportCreateReport:
```
devinivy: I am of two minds on this one.  On one hand it's nice to be in parity with `com.atproto.moderation.createReport`, on the other hand this is deprecated functionality that we may want to encourage folks to move away from, particularly as they adopt the new endpoint.

packages/pds/src/api/com/atproto/moderation/createReport.ts
```diff
@@ -38,4 +38,37 @@ export default function (server: Server, ctx: AppContext) {
       }
     },
   })
+
+  server.tools.ozone.report.createReport({
```
devinivy, line 42: I assume we need these endpoints here only in order to preserve the ability to create a report after having been takendown?  Otherwise I would hope to just use service proxying.  In general we're looking to minimize lexicon/app-specific behaviors in the PDS, so that's the hat I have on.

lexicons/com/atproto/moderation/createReport.json
```diff
@@ -4,7 +4,7 @@
   "defs": {
     "main": {
       "type": "procedure",
-      "description": "Submit a moderation report regarding an atproto account or record. Implemented by moderation services (with PDS proxying), and requires auth.",
+      "description": "Submit a moderation report regarding an atproto account or record. Implemented by moderation services (with PDS proxying), and requires auth. Compatible with and replaced by the tools.ozone.report.createReport NSID.",
```
devinivy: I wonder if we want to express that `tools.ozone.report.createReport` is compatible with `com.atproto.moderation.createReport`, i.e. as a superset, but not the other way around.  That allows `tools.ozone.report.createReport` to continue to adopt functionality that doesn't necessarily make its way back into `com.atproto`.

lexicons/tools/ozone/report/createReport.json
```diff
@@ -0,0 +1,90 @@
...
+              "maxLength": 20000,
+              "description": "Additional context about the content and violation."
+            },
+            "subject": {
+              "type": "union",
+              "refs": [
+                "com.atproto.admin.defs#repoRef",
+                "com.atproto.repo.strongRef"
+              ]
+            },
```
devinivy, line 31: May not really be our job to fix here, but I always found this to be an awkward union.  It would be neat if it could be something like `["com.atproto.repo.repoRef", "com.atproto.repo.strongRef"]`.

devinivy: (Though we need to continue to support `com.atproto.admin.defs#repoRef` regardless, in order to maintain compat with `com.atproto.moderation.createReport`, so I think we can ignore this for now.)

lexicons/com/atproto/moderation/createReport.json
bnewbold: I switches to the term "supercede" in both.

lexicons/tools/ozone/report/createReport.json
bnewbold: agree in general, but don't think this PR is the place for it.

packages/pds/src/pipethrough.ts
bnewbold: yeah, I think we should not add additional passthroughs, and this is a good reminder to make a plan for deprecating and removing them overall.

packages/pds/src/api/com/atproto/moderation/createReport.ts
foysalit: yeah i dont think we can handle reports from takendown accounts without this?

packages/pds/src/pipethrough.ts
foysalit: hmm... i get what you're saying but without this the requests would never reach ozone, right?

PR #4461, opened 2025-12-16
Support initiating user registration via OAuth flow with prompt=create

ThisIsMissEm:
Recently there's been a lot of discussion about how to do user registration from applications (clients) against given PDS services. This PR implements a standardised mechanism for initiating user registration via the OIDC `prompt` parameter (we already had support for other modes).

Specification: https://openid.net/specs/openid-connect-prompt-create-1_0.html

This PR adds:
- support for exposing the `prompt` values supported via `prompt_values_supported` in Authorization Server Metadata, per the above specification.
- support in the oauth-provider-ui to allow `prompt=create` to show the signup screen
- support in oauth-provider for showing the signup screen for `prompt=create`

This allows clients to initiate user registration with the target PDS service, where the user will immediately be dropped into the user registration flow, instead of the "create account or login" or "select account or create account" flows.

I haven't yet added test coverage for this PR, as I wanted to get the Bluesky Team's thoughts on it first.

> [!NOTE]
> Something that we may want is an explicit environment variable in the PDS implementation to control whether or not user registration should be supported. Currently it's inferred by `PDS_SERVICE_HANDLE_DOMAINS` which may not be desirable in all use-cases.

You can see a demo here: https://bsky.app/profile/thisismissem.social/post/3ma4setzr6s2t

### DIFF
--- a/packages/pds/tests/oauth.test.ts
+++ b/packages/pds/tests/oauth.test.ts
@@ -134,7 +134,8 @@ describe('oauth', () => {
     await browser?.close()
   })
 
-  it('Allows to sign-up trough OAuth', async () => {
+  // This uses prompt=create under the hood:
+  it('Allows to sign-up through OAuth', async () => {
     const page = await PageHelper.from(browser)
 
     await page.goto(appUrl)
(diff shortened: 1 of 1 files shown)


packages/oauth/oauth-provider/src/metadata/build-metadata.ts
```diff
@@ -84,6 +85,15 @@ export function buildMetadata(
       // 'wap', LoL
     ],
 
+    // https://openid.net/specs/openid-connect-prompt-create-1_0.html
+    prompt_values_supported: [
+      'none',
+      'login',
+      'consent',
+      'select_account',
+      'create',
```
ThisIsMissEm, line 94: We could optionally include this if the PDS does not support user registration (see the note in the pull request description)

matthieusieben:
We can prefix that env variable with PDS_OAUTH for now so it's clear only oauth signups are disabled

ThisIsMissEm:
I'm going to suggest we leave completely disabling user registration out of this pull request, and instead implement in it a follow up pull request once we decide on the exact concept we want for this. Does that sound good @matthieusieben ?

The Bluesky PBC PDSes are also unaffected by this change since they are not authorization servers. The entryway may need some minor adaptation for this to work (I'm not sure)

PR #4482, opened 2025-12-23
feat(doc): add @avivkeller to disclosure credits

avivkeller:
cc @pfrazee


pfrazee:
Thank you so much for your contribution, @avivkeller. I couldn't be happier to add you to this.

PR #4483, opened 2025-12-23
Tap: Lex Indexer

dholms:
Adds a new `LexIndexer` class that is fully typed and does validation through the `@atproto/lex` library.

### DIFF
--- a/packages/tap/src/lex-indexer.ts
+++ b/packages/tap/src/lex-indexer.ts
@@ -17,11 +17,7 @@ export type UpdateEvent<R> = BaseRecordEvent & {
   cid: string
 }
 
-export type PutEvent<R> = BaseRecordEvent & {
-  action: 'create' | 'update'
-  record: R
-  cid: string
-}
+export type PutEvent<R> = CreateEvent<R> | UpdateEvent<R>
 
 export type DeleteEvent = BaseRecordEvent & {
   action: 'delete'
@@ -65,14 +61,14 @@ export type RecordHandler<R> =
   | PutHandler<R>
   | DeleteHandler
 
-interface RegisteredHandler<R> {
+interface RegisteredHandler {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  handler: RecordHandler<any>
   schema: RecordSchema
-  handler: RecordHandler<R>
 }
 
 export class LexIndexer implements TapHandler {
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  private handlers = new Map<string, RegisteredHandler<any>>()
+  private handlers = new Map<string, RegisteredHandler>()
   private otherHandler: UntypedHandler | undefined
   private identityHandler: IdentityHandler | undefined
   private errorHandler: ErrorHandler | undefined


packages/tap/src/lex-indexer.ts
```diff
@@ -0,0 +1,182 @@
...
+    const registered = this.handlers.get(key)
+
+    if (!registered) {
+      await this.otherHandler?.(evt, opts)
+      return
+    }
+
+    if (action === 'create' || action === 'update') {
+      const result = registered.schema.safeParse(evt.record)
+      if (!result.success) {
```
ThisIsMissEm: I think you forgot to use result here?

ThisIsMissEm: Or, rather, the type inference for the `result.data` isn't used, because the `as UntypedHandler` is casting the handler to ignore the type of value. I'm wondering whether it'd actually be better to keep separate maps for create/update/data which can then give you the correctly typed handler.

packages/tap/src/lex-indexer.ts
```diff
@@ -0,0 +1,182 @@
+import { Infer, Namespace, RecordSchema, getMain } from '@atproto/lex'
```
matthieusieben: I just renamed `Namespace` to `Main` (for consistency with `getMain`):

https://github.com/bluesky-social/atproto/commit/9af7a2d12240e91248610ce4fe7d93387733c59c#diff-44b24f67e08736656e21b9fb3d240fe2d91ac5ccbfef79a4baa0663a6d8bfe07

Maybe that was a mistake? wdyt the best name for this would be?

packages/tap/src/lex-indexer.ts
```diff
@@ -0,0 +1,182 @@
...
+  put<const T extends RecordSchema>(
+    ns: Namespace<T>,
+    handler: PutHandler<Infer<T>>,
+  ): this {
+    this.register('create', ns, handler)
+    return this.register('update', ns, handler)
+  }
+
+  other(fn: UntypedHandler): this {
+    this.otherHandler = fn
```
matthieusieben, line 127: Should this throw if a handle is already set? (For consistency with `register()`)
```suggestion
    if (this.otherHandler) {
      throw new Error(`Handler already registered for "other"`)
    }
    this.otherHandler = fn
```

(same for other methods)

packages/tap/src/lex-indexer.ts
```diff
@@ -0,0 +1,182 @@
...
+    const key = this.handlerKey(collection, action)
+    const registered = this.handlers.get(key)
+
+    if (!registered) {
+      await this.otherHandler?.(evt, opts)
+      return
+    }
+
+    if (action === 'create' || action === 'update') {
+      const result = registered.schema.safeParse(evt.record)
```
matthieusieben: `registered.schema.safeParse` will apply defaults and return a potentially different result than the input.

Either use 
```ts
if (!registered.schema.matches(evt.record)) {
  throw Error(..)
}
```

or, if you want the original error:

```ts
const result = registered.schema.safeParse(evt.record, { allowTransform: false })
```

packages/tap/src/lex-indexer.ts
dholms: good call 👍

packages/tap/src/lex-indexer.ts
dholms: Hmmm I think `Main` is fine. It's an awkward type to name

packages/tap/src/lex-indexer.ts
dholms: Yeah I actually do think it's better to keep separate maps for create/update/delete. Unfortunately I don't think that solves the problem of using the type from `result.data` for the same reason why we need to type the events with `<any>` in the `Map` type

dholms: Actually I tried splitting it up into three maps and it complicated the code a good deal for no real benefit in the types. So I think I'll leave it as is

PR #4496, opened 2026-01-01
Fix queryLabels cursor pagination by adding missing ORDER BY

### LINKED ISSUE #4328: queryLabels not behaving consistently, and therefore giving different outputs.
**Describe the bug**

I am trying to gather labels from one labeller related to an account using the queryLabels endpoint from the labeller. When I try to run queryLabels on the default Bluesky Moderator (https://mod.bsky.app/), I get different cursors for the same query. 

**To Reproduce**

1. Run the following multiple times: https://mod.bsky.app/xrpc/com.atproto.label.queryLabels?uriPatterns=at://did:plc:dafghqyxkv7cmx4vlrvlaorc*&cursor=0

You will see that the cursor value change and the values returned in the label field changes as well. This makes it not possible to find all entries corresponding to the URI ("at://did:plc:dafghqyxkv7cmx4vlrvlaorc*" in this case)

**Expected behavior**

**Details**

I would expect to get back the same values in the label field, and the same cursor. 

**Additional context**

I think the problem could be that the cursor returned is "const resCursor = res.at(-1)?.id.toString(10)" found here:  https://github.com/bluesky-social/atproto/blob/92702322c357c377607d81e80906fe46ac4e1069/packages/ozone/src/api/label/queryLabels.ts#L50

This corresponds to the last entry returned. But the database currently I think does not necessarily return labels in a specified sorted order. Hence, the cursor values are different, as are the number of items returned in labels.

heathdutton:
Fixes #4328

The `queryLabels` endpoint uses cursor-based pagination with `where('id', '>', cursorId)`, but was missing an `ORDER BY` clause. This caused inconsistent results across queries and unreliable pagination.

Added `.orderBy('id', 'asc')` to ensure deterministic ordering.

### DIFF
--- a/packages/ozone/src/api/label/queryLabels.ts
+++ b/packages/ozone/src/api/label/queryLabels.ts
@@ -6,7 +6,11 @@ import { Server } from '../../lexicon'
 export default function (server: Server, ctx: AppContext) {
   server.com.atproto.label.queryLabels(async ({ params }) => {
     const { uriPatterns, sources, limit, cursor } = params
-    let builder = ctx.db.db.selectFrom('label').selectAll().limit(limit)
+    let builder = ctx.db.db
+      .selectFrom('label')
+      .selectAll()
+      .orderBy('id', 'asc')
+      .limit(limit)
     // if includes '*', then we don't need a where clause
     if (!uriPatterns.includes('*')) {
       builder = builder.where((qb) => {


matthieusieben:
Looks good. Could you add a patch changeset by running `pnpm changeset` ? thanks.

PR #4501, opened 2026-01-04
Small @atproto/lex improvements

matthieusieben:


matthieusieben:
@ThisIsMissEm For now, I'll keep it as is as it would require a more complex rework. If there is high demand for it, we can revisit this decision.

PR #4522, opened 2026-01-08
Perf: Optimize listRecords to avoid un-indexed db queries

DavidBuchanan314:
`listRecords` on a very large collection is very slow (easily ~tens of seconds, for example: https://pdsls.dev/at://did:plc:ifeda2auw5iicswhcdtgelai/app.bsky.feed.like ). There is no actor store index for `record(collection, rkey)`, only `record(collection)` on its own, and so it has to do a full scan and sort.

This is easily fixable by adding a new index on (collection, rkey), which I implemented here: https://github.com/bluesky-social/atproto/commit/3a6508adf7850b21dc45767d84dd93d8e03b2dc1

However, this involves a migration on the actor store dbs, which has never happened in prod before and might take time for large repos. Maybe it'd be fine, but on further thought I realised that the indexes we need are already here.

The primary key for the records table is `uri`, which is effectively (did, collection, rkey). And since did is constant... it's equivalent to (collection, rkey) - we just have to modify our queries so that the db can make use of this fact, by constructing appropriate URIs - which is what this PR implements.

There was already good test coverage of the listRepos endpoint.

### DIFF
--- a/packages/pds/src/actor-store/record/reader.ts
+++ b/packages/pds/src/actor-store/record/reader.ts
@@ -12,6 +12,7 @@ import { Record as PostRecord } from '../../lexicon/types/app/bsky/feed/post'
 import { StatusAttr } from '../../lexicon/types/com/atproto/admin/defs'
 import { LocalRecords } from '../../read-after-write/types'
 import { ActorDb, Backlink } from '../db'
+import { makeAtUriStringWithoutValidation } from './util'
 
 export type RecordDescript = {
   uri: string
@@ -97,12 +98,12 @@ export class RecordReader {
       .where(
         'record.uri',
         '>=',
-        AtUri.make(this.did, collection, '').toString(),
+        makeAtUriStringWithoutValidation(this.did, collection, ''),
       )
       .where(
         'record.uri',
         '<',
-        AtUri.make(this.did, collection, '\x7f').toString(),
+        makeAtUriStringWithoutValidation(this.did, collection, '\x7f'),
       ) // together, these two .where() clauses are equivalent to .where('record.collection', '=', collection), but index-friendly
       .if(!includeSoftDeleted, (qb) =>
         qb.where(notSoftDeletedClause(ref('record'))),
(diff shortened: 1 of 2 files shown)


DavidBuchanan314:
(The test failure above is unrelated to my changes)

I had claude write a quick benchmark script: <https://gist.github.com/DavidBuchanan314/bb46937e16122544a77f9c8d9a0ed569>

After populating the db with 1m records, *without* the fix it takes about 2500ms for each listRecords query, and after the fix <10ms.

packages/pds/src/actor-store/record/reader.ts
```diff
@@ -83,31 +86,51 @@ export class RecordReader {
...
-      .where('record.collection', '=', collection)
+      .where(
+        'record.uri',
+        '>=',
+        AtUri.make(this.did, collection, '').toString(),
+      )
+      .where(
+        'record.uri',
+        '<',
+        AtUri.make(this.did, collection, '\x7f').toString(),
```
matthieusieben: I feel a little uneasy with using `AtUri.make` like this. Ideally, `AtUri` should be stricter in the way it validates its input (just as `URL` does not allow creating a "partial" url).

The reason I mention this is that there are other parts of the code (e.g. data parsing and validation) that use `AtUri` fields assuming they are valid (ie. without checking that they actually are).

Sadly, we can't just make `AtUri` stricter because of some edge cases like this one. I think, as a long term strategy, that we should steer away from abusing these common helpers for edge cases, and instead use dedicated helpers instead.

DavidBuchanan314: Good point, and looking closer `AtUri.make(this.did, collection, '')` doesn't actually do what I intended, because `.make()` skips the trailing slash if the rkey is falsey (which empty string is).

DavidBuchanan314: I added a dedicated helper, although I'm a little nervous that if AtUri.make() ever changes encoding logic (e.g. percent-encoding for dids), we'd get breakage. Although we'd get breakage in any case, because `getRecord` involves building URIs for lookups too (it was perhaps a mistake to have URI strings in the database in the first place).

DavidBuchanan314: The "ultimate fix" might be to do a (nontrivial!) db migration that gets rid of the uri column entirely, replacing it with (collection, rkey) as composite primary key.

PR #4526, opened 2026-01-08
Adds GrowthBook feature flag for suggested users from Discover

cuducos:
This PR adds a feature flag for routing `app.bsky.unspecced.getSuggestedUsersSkeleton` to Discover (instead of Hot Topics):

1. Refactor feature gate to work with GrowthBook instead of StatSig
2. Drop `SearchFilteringExploration` (will not be migrated to GrowthBook and not needed anymore)
3. Drop StatSig dependency

Settings as draft because:

- [x] Could not confirm our URL to access the GrowthBook
- [x] Need to make sure the GrowthBook API key is present in our vault

Follow-up PRs planned:

- Migrate `ThreadsV2ReplyRankingExploration` to GrowthBook (aligned with @estrattonbailey)
- Rename `skeletonFromDiscover` to `skeleton` (and drop resulting dead code, ie current `skeleton` and `skeletonFromTopics`)

### DIFF
--- a/.changeset/funky-camels-hunt.md
+++ b/.changeset/funky-camels-hunt.md
@@ -1,0 +1,5 @@
+---
+'@atproto/bsky': patch
+---
+
+Swaps StatSig for GrowthBook in AppView


packages/bsky/src/feature-gates.ts
```diff
@@ -1,9 +1,19 @@
-import Statsig, { StatsigUser } from 'statsig-node'
+import { GrowthBookClient } from '@growthbook/growthbook'
 import { featureGatesLogger } from './logger'
 
+// TODO: double-check that (this URL is mostly a placeholder by now)
+const growthBookUrl = 'https://growthbook.prod.k1.bsky.dev'
```
jcalabro: This should come as an env var rather than being hard-coded

packages/bsky/src/api/app/bsky/feed/searchPosts.ts
```diff
@@ -137,18 +122,8 @@ const noBlocksOrTagged = (inputs: RulesFnInput<Context, Params, Skeleton>) => {
     // Cases to never show.
     if (ctx.views.viewerBlockExists(creator, hydration)) return false
 
-    let tagged = false
-    if (
-      params.hydrateCtx.featureGates.get(
-        FeatureGateID.SearchFilteringExploration,
-      )
-    ) {
-      tagged = post.tags.has(ctx.cfg.visibilityTagHide)
```
estrattonbailey, line 146: Oh sorry this wasn't clear. This experiment was ramped to 100%, so we should use this truthy codepath instead.

packages/bsky/src/feature-gates.ts
```diff
@@ -12,76 +22,63 @@ export enum FeatureGateID {
...
   }
 
-  user({ did }: { did?: string }): StatsigUser | undefined {
-    return did
-      ? {
-          userID: did,
-        }
-      : undefined
+  contextForDid(did: string): Context {
+    return { attributes: { did: did } }
```
estrattonbailey: Totally fine with this since we've never added additional attributes thus far, but if we want to add more attributes in the future using data from specific endpoints/methods, we'll want to expand on this def to allow more properties. Just noting the intent of the original method, nbd rn 👍

cuducos: Yes. I sort of thought about that and given (a) avoidance of premature optimization and (b) my lack of context of feature gate use cases, I decided to start with this simple method. But yep, we need to revist that as our use-cases demands more complexity : )

packages/bsky/src/api/app/bsky/feed/searchPosts.ts
cuducos: Ah, got it! I'm gonna revert & invert that. No worries.

packages/bsky/src/feature-gates.ts
cuducos: Good call, gonna update it!

cuducos:
I think it should be good to go. Since configs (URL and API key) are optional, it should just ignore feature flags by now.

packages/bsky/src/feature-gates.ts
```diff
@@ -1,9 +1,17 @@
...
+import { GrowthBookClient } from '@growthbook/growthbook'
 import { featureGatesLogger } from './logger'
 
 export type Config = {
+  apiUrl?: string
   apiKey?: string
-  env?: 'development' | 'staging' | 'production' | string
+}
+
+interface Attrs {
```
rafaeleyng: I think we have a slight preference for type aliases over interfaces, so `type Attrs = { // ...`

packages/bsky/src/api/app/bsky/feed/searchPosts.ts
rafaeleyng: The implementation seems to match what @estrattonbailey asked for, so I'm guessing this was already addressed?

packages/bsky/src/feature-gates.ts
```diff
@@ -12,76 +20,63 @@ export enum FeatureGateID {
...
+      featureGatesLogger.error({ err }, 'Failed to initialize GrowthBook')
       this.ready = false
     }
   }
 
   destroy() {
     if (this.ready) {
       this.ready = false
-      this.statsig.shutdown()
+      this.client = undefined
```
rafaeleyng: If there is no equivalent to `shutdown()` in growthbook, we possibly don't need this `destroy` method at all.

packages/bsky/src/feature-gates.ts
```diff
@@ -12,76 +20,63 @@ export enum FeatureGateID {
...
 
-  check(gate: FeatureGateID, user?: StatsigUser): boolean {
-    if (!this.ready) return false
-    if (!user) return false
-    return this.statsig.checkGateSync(user, gate)
+  check(gate: FeatureGateID, ctx: Context): boolean {
+    if (!this.ready || !this.client) return false
+    if (!ctx.attributes || !ctx.attributes.did) return false
+
+    // TODO: migrate from StatSig to GrowthBook
```
rafaeleyng: Is this TODO supposed to be addressed still on your PR? If no, I think a better comment would be e.g. `Remove feature flag, this experiment was ramped up to 100%.`.

cuducos: Nope : ) As in the PR description:

> Follow-up PRs planned:
>
>  * Migrate ThreadsV2ReplyRankingExploration to GrowthBook (aligned with @estrattonbailey)

packages/bsky/src/api/app/bsky/feed/searchPosts.ts
estrattonbailey: Oh shoot, I'm really sorry, I confused this with another experiment. This was ramped only for the bsky team as part of another experiment. The correct handling was as you had it before 🤦  it should use `ctx.cfg.searchTagsHide` instead.

packages/bsky/src/feature-gates.ts
```diff
@@ -22,67 +30,48 @@ export enum FeatureGateID {
...
-         */
-        await this.statsig.initialize(this.config.apiKey, {
-          localMode: this.config.env === 'test',
-          environment: {
-            tier: this.config.env || 'development',
-          },
+        this.client = new GrowthBookClient({
+          apiHost: this.config.apiUrl,
+          clientKey: this.config.apiKey,
         })
```
estrattonbailey, line 64: Need a `await this.client.init()` call here too I think

packages/bsky/src/feature-gates.ts
rafaeleyng: Could you clarify on the comment what exactly should be done? Like, should we remove these 2 lines?

cuducos: Sure thing, good call.
Adding here and pushing to this branch in the next few minutes.

packages/bsky/src/feature-gates.ts
cuducos: Good catch ✨ Added in a2ab9dd

packages/bsky/src/feature-gates.ts
```diff
@@ -22,67 +30,49 @@ export enum FeatureGateID {
...
+        await this.client.init()
```
estrattonbailey: Also just realized we're going to want to get refreshed gates periodically since these servers can run for a long time without restarting. [Some docs here](https://docs.growthbook.io/lib/node#refreshing-features), I think a poll would be fine, maybe every 5 minutes or so. StatSig appears to have defaulted to every 10 seconds, but that seems like overkill to me.

PR #4532, opened 2026-01-09
Replace `tap`'s event validation from "zod" to "@atproto/lex"

matthieusieben:


packages/tap/src/types.ts
```diff
@@ -89,21 +93,27 @@ export const parseTapEvent = (data: unknown): TapEvent => {
...
       collection: parsed.record.collection,
       rkey: parsed.record.rkey,
-      record: parsed.record.record,
+      record:
+        parsed.record.record === undefined
+          ? undefined
+          : (jsonToLex(parsed.record.record as JsonObject, {
+              // reject invalid blobs/cids in records
+              strict: true,
+            }) as LexMap),
```
matthieusieben: This part I am less sure of. I think we should expose "lex" data (with parsed CID and Uint8array) but I am not sure.

packages/tap/src/channel.ts
```diff
@@ -122,8 +123,11 @@ export class TapChannel {
   private async processWsEvent(chunk: Uint8Array) {
     let evt: TapEvent
     try {
-      const data = chunk.toString()
-      evt = parseTapEvent(JSON.parse(data))
+      const data = lexParse(chunk.toString(), {
+        // Reject invalid CIDs and blobs
+        strict: true,
+      })
```
matthieusieben, line 130: There are two breaking changes in here:
1) the `data` is no longer json structures only: it also contains `Cid` and `Uint8Arrays`
2) the `strict` parameter will reject any json that does not properly comply to the atproto data model (floats, invalid CID version or codec, invalid `$type`, etc.)

packages/tap/src/lex-indexer.ts
```diff
@@ -169,7 +173,7 @@ export class LexIndexer implements TapHandler {
     if (action === 'create' || action === 'update') {
       const match = registered.schema.safeValidate(evt.record)
       if (!match.success) {
-        const uriStr = AtUri.make(evt.did, evt.collection, evt.rkey).toString()
+        const uriStr: AtUriString = `at://${evt.did}/${evt.collection}/${evt.rkey}`
```
ThisIsMissEm, line 176: From real-world usage, it'd actually be very handy for the record event handlers to receive the AT URI, since we tend to need this for like the record primary key in the database.

matthieusieben: Do not hesitate to create an issue / PR ;-)

PR #4537, opened 2026-01-10
Add Korean translation

quiple:
Add Korean translation to `oauth-provider-frontend` and `oauth-provider-ui` packages.


quiple:
Updated for #4820.

matthieusieben:
Hey! This is great and definitely something we want to adopt! Thank you for contributing.

This is missing a few things before we can merge:
1) The .po file needs to be updated (merge `main`)
2) We need a changeset file (run `pnpm changeset` from the root)
3) You need to add a `ko` entry [here](https://github.com/bluesky-social/atproto/blob/9cc6b3369878fe46554727f2ae9342376f2e41b6/packages/oauth/oauth-provider-ui/src/locales/locales.ts#L3-L16)

matthieusieben:
Thank you for your help! Note that many changes where made since you last updated this. Please open a new PR with the new strings if you also want those translated.

PR #4538, opened 2026-01-10
Clarify applyWrites commit behaviour

DavidBuchanan314:
`applyWrites` was always strongly implied to produce a single commit (and this is what the reference implementation does), but it wasn't explicitly stated - now it is.


bnewbold:
I don't think this needs a changeset, but it maybe does need a codegen?

PR #4544, opened 2026-01-14
Pds: enable ratelimits by default

dholms:


bnewbold (approved):
:gem:

packages/pds/src/config/config.ts
```diff
@@ -224,15 +224,16 @@ export const envToCfg = (env: ServerEnvironment): ServerConfig => {
...
-      }
-    : { enabled: false }
+  const rateLimitsCfg: ServerConfig['rateLimits'] =
+    env.rateLimitsEnabled === false
+      ? { enabled: false }
+      : {
+          enabled: true,
+          bypassKey: env.rateLimitBypassKey,
+          bypassIps: env.rateLimitBypassIps?.map((ipOrCidr) =>
+            ipOrCidr.split('/')[0]?.trim(),
```
ThisIsMissEm, line 234: Wait, this isn't right. This would be taking just the IP component of a CIDR, so 74.125.227.1/30 would only trust 74.125.227.1, not 74.125.227.2 or 74.125.227.3 or 74.125.227.0

ThisIsMissEm (commented):
It'd also be an idea to modify the installer and upgrade scripts in the pds repo to set the appropriate env vars (even if they're commented out)

packages/pds/src/config/config.ts
DavidBuchanan314: The node [`net.BlockList`](https://nodejs.org/api/net.html#class-netblocklist) class has support for dealing with cidr ranges properly, and might be a good approach to fixing this (even though this is more of an allowlist than a blocklist).

bnewbold:
yeah it looks like tests are failing b/c rate-limits are now on by default

PR #4546, opened 2026-01-14
Refactor feed insertion to use original feed directly

erikologic:
Should provide a minimal performance gain


erikologic:
@ThisIsMissEm 
> Sounds like formatAndInsertPosts may be mutative, so maybe that's why the copy was happening?

Yeah I wondered, but the copy is A) shallow B) for an object that AFAIU is the AppView server response so its life should be bound to this XRPC execution and after that we should only care about the resulting `feed`.  

BTW if you want to protect against mutations you should do so in the writer and not the coordinator, or use languages/data structures that enforce immutability IMO.
Is hard to code when you have to worry of what could happen in nested code paths.

NABD anyway

PR #4556, opened 2026-01-16
Update pds service entrypoint's node version

devinivy:
Updates the pds's node version to v20.20 in services/.


ThisIsMissEm:
Is this dockerfile even used? Isn't this the one for the PDS releases? https://github.com/bluesky-social/pds/blob/main/Dockerfile

PR #4560, opened 2026-01-19
PDS: Stop probing image dimensions on blob upload

### LINKED ISSUE #3151: Can't upload text/html blob with contents that contain "<svg" + any character

DavidBuchanan314:
For blobs that look like images, the PDS attempts to parse out the image's pixel dimensions. The image's dimensions are saved to the actor db, but they are never used for anything. They aren't even returned by `BlobReader.getBlobMetadata()`.

So, this PR removes the image dimension detection, and removes the `sharp` library as a PDS dependency. Should fix #3151 

We should consider removing the width/height DB columns at some point in the future.

(MIME sniffing in general still occurs)

### DIFF
--- a/packages/pds/src/actor-store/db/schema/blob.ts
+++ b/packages/pds/src/actor-store/db/schema/blob.ts
@@ -3,8 +3,8 @@ export interface Blob {
   mimeType: string
   size: number
   tempKey: string | null
-  width: number | null
-  height: number | null
+  // width: number | null  // @TODO: actually drop these columns from the db
+  // height: number | null
   createdAt: string
   takedownRef: string | null
 }


packages/pds/tests/file-uploads.test.ts
```diff
@@ -248,8 +244,6 @@ describe('file uploads', () => {
       .executeTakeFirst()
 
     expect(found?.mimeType).toBe('image/png')
-    expect(found?.width).toBe(554)
-    expect(found?.height).toBe(532)
```
DavidBuchanan314, line 252: note, these tests were only able to inspect the dimensions because they're reading directly from the db

packages/pds/package.json
```diff
@@ -74,7 +74,6 @@
     "p-queue": "^6.6.2",
     "pino": "^8.21.0",
     "pino-http": "^8.2.1",
-    "sharp": "^0.33.5",
```
dholms, line 77: 🙌

ThisIsMissEm:
This should also remove a blocker for the #3750 PRs, as outdated Sharp being incompatible with newer node.js versions was a pain point iirc.

(now there's just a matter of the PDS entryway which relies on a very old better-sqlite3 install)

PR #4569, opened 2026-01-20
Fix oauth response when using `prompt=select_account` with no sessions available

### LINKED ISSUE #4566: Prompt value of `select_account` does not return an error if no accounts are signed in.
**Describe the bug**

According to [OpenID Connect Core 1.0](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest) when using the `select_account` prompt parameter, if no sessions currently exist, then the authorization server should return the error `account_selection_required`

> If it cannot obtain an account selection choice made by the End-User, it MUST return an error, typically account_selection_required. 

**To Reproduce**

1. Start an authorization request with the `prompt` parameter value of `select_account` with an Authorization Server where no sessions exist.
2. Observe that you see the login form, not a list of active sessions to select from.

**Expected behavior**

If I'm reading the OpenID Connect Core 1.0 specification correctly, then the error `account_selection_required` MUST be returned.

**Additional context**

To be honest, I would actually expect the request to downgrade to the `login` prompt parameter value, however, the specification does not mention that.

cc @matthieusieben

matthieusieben:
The implementation now correctly returns the `account_selection_required` error when the `prompt` parameter is set to `select_account` and no active sessions are found. This change aligns with the OpenID Connect Core 1.0 specification, ensuring proper error handling in scenarios where account selection is required but unavailable.

Fixes #4566

### DIFF
--- a/packages/oauth/oauth-provider/src/oauth-provider.ts
+++ b/packages/oauth/oauth-provider/src/oauth-provider.ts
@@ -711,7 +711,9 @@ export class OAuthProvider extends OAuthVerifier {
           consentRequired: session.consentRequired,
 
           selected:
-            parameters.prompt == null || parameters.prompt === 'consent'
+            parameters.prompt == null ||
+            parameters.prompt === 'login' ||
+            parameters.prompt === 'consent'
               ? matchesHint.call(parameters, session)
               : false,
         })),


ThisIsMissEm:
So I was actually right in my reading of OpenID Connect Core 1.0? tbh, I think downgrading a select_account to a login or consent flkw is actually the correct user experience, even though the spec says otherwise

ThisIsMissEm:
Okay, I guess this means that on servers that don't offer `prompt=create` I definitely can't reliably offer registration.

PR #4601, opened 2026-02-02
Add docs in lex SDK packages

matthieusieben:


packages/lex/lex-data/src/cid.ts
```diff
@@ -4,31 +4,65 @@ import { sha256, sha512 } from 'multiformats/hashes/sha2'
...
-export type DAG_CBOR_MULTICODEC = typeof DAG_CBOR_MULTICODEC
+/**
+ * Codec code that indicates the CID references a CBOR-encoded data structure.
+ *
+ * Used when encoding structured data in ATProto repositories.
+ *
+ * @see {@link https://dasl.ing/cid.html Content IDs (DASL)}
+ */
+export const CBOR_DATA_CODEC = 0x71
+export type CBOR_DATA_CODEC = typeof CBOR_DATA_CODEC
```
devinivy, line 15: It looks like there are some breaking changes in this package that require a changeset.  But in a sea of documentation updates this is a bit hard to catch!  In would recommend splitting these kinds of changes out or flagging them clearly in PR description/comments.

packages/lex/lex-data/src/cid.ts
```diff
@@ -4,31 +4,65 @@ import { sha256, sha512 } from 'multiformats/hashes/sha2'
...
 
+/**
+ * Compares two {@link Multihash} for equality.
+ *
+ * @param a - First {@link Multihash}
+ * @param b - Second {@link Multihash}
+ * @returns `true` if both multihashes have the same code and digest
+ */
 export function multihashEquals(a: Multihash, b: Multihash): boolean {
+  if (a === b) return true
```
devinivy, line 65: It is fairly difficult to identify and review implementation changes sprinkled throughout this kind of PR.

packages/lex/lex-json/src/bytes.ts
```diff
@@ -15,12 +41,33 @@ export function parseLexBytes(input?: {
   }
 
   if (typeof input.$bytes !== 'string') {
-    throw new TypeError('$bytes must be a base64-encoded string')
+    return undefined
   }
```
devinivy, line 45: This is a subtle change in behavior.  Having `$bytes` set to a value other than a string is invalid in the lex data model, so it's not obvious at a glance for me what we'd want to happen here—sort of depends where `parseLexBytes()` sits in the whole lex stack.  I think changes like this are best broken out and ideally backed by a test.

packages/lex/lex-schema/src/schema/custom.ts
```diff
@@ -6,18 +6,46 @@ import {
...
+ *
+ * @example
+ * ```ts
+ * const schema = new CustomSchema(
+ *   (input): input is Date => input instanceof Date,
+ *   'Expected a Date instance'
+ * )
+ * ```
+ */
+export class CustomSchema<out TValue = unknown> extends Schema<TValue> {
```
devinivy, line 48: I swear I've never seen `out` before 😆

packages/lex/lex-server/README.md
```diff
@@ -18,7 +18,7 @@ npm install @atproto/lex-server
 
 **What is this?**
 
-Building AT Protocol servers requires handling XRPC requests, validating inputs against Lexicon schemas, managing authentication, and supporting real-time subscriptions. `@atproto/lex-server` automates this by:
+Building ATProto servers requires handling XRPC requests, validating inputs against Lexicon schemas, managing authentication, and supporting real-time subscriptions. `@atproto/lex-server` automates this by:
```
devinivy: Should be "AT Protocol" or "atproto", but not "ATProto".
```suggestion
Building atproto servers requires handling XRPC requests, validating inputs against Lexicon schemas, managing authentication, and supporting real-time subscriptions. `@atproto/lex-server` automates this by:
```

packages/lex/lex/README.md
```diff
@@ -1485,7 +1184,7 @@ export const post: Action<PostInput, { uri: string; cid: string }> = async (
 
 // Usage
 await client.call(post, {
-  text: 'Hello, AT Protocol!',
+  text: 'Hello, ATProto!',
```
devinivy:
```suggestion
  text: 'Hello, atproto!',
```

devinivy: I see many other places use `ATProto` so `atproto` would look inconsistent.  I think we should just leaves these as `AT Protocol` for now since it is correct, and over time switch `ATProto` to `AT Protocol` or `atproto`.

packages/lex/lex-json/src/bytes.ts
matthieusieben: Added a test and changelog. The main motivation here is that we don't want `jsonToLex` to throw in non-strict mode

PR #4602, opened 2026-02-03
PDS: fix rpc audience check for service proxying

devinivy:
Fixes an issue where the rpc scope audience checks weren't properly taking into account the service id when service proxying.

### DIFF
--- a/packages/pds/src/pipethrough.ts
+++ b/packages/pds/src/pipethrough.ts
@@ -56,7 +56,8 @@ export const proxyHandler = (ctx: AppContext): CatchallHandler => {
         throw new InvalidRequestError('Bad token method', 'InvalidToken')
       }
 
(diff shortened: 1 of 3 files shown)


packages/pds/src/pipethrough.ts
```diff
@@ -56,7 +56,8 @@ export const proxyHandler = (ctx: AppContext): CatchallHandler => {
         throw new InvalidRequestError('Bad token method', 'InvalidToken')
       }
 
-      const { url: origin, did: aud } = await parseProxyInfo(ctx, req, lxm)
+      const proxy = await parseProxyInfo(ctx, req, lxm)
+      const aud = serviceAud(proxy)
```
devinivy: This is where the issue was: `aud` was a bare DID without a service id, causing it to fail the scope checks.

matthieusieben (approved):
neat

PR #4605, opened 2026-02-03
Fix scope `aud` param

matthieusieben:
POC for fixing request proxying with oauth credentials. On hold as there is an issue with the OAuth spec of the `rpc:` scope that needs to be tackled first.


packages/oauth/oauth-scopes/src/index.ts
```diff
@@ -11,3 +11,11 @@ export * from './scopes/identity-permission.js'
 export * from './scopes/include-scope.js'
 export * from './scopes/repo-permission.js'
 export * from './scopes/rpc-permission.js'
+
+// Legacy
```
matthieusieben, line 15: For backwards compat

PR #4606, opened 2026-02-03
Throw more detailed error upon CSRF login issue

matthieusieben:
This PR essentially attempts to fix an OAuth login issue in iOS.

For some reason, `Set-Cookie` headers seem to be ignored when the authorization flow is open in a `ASWebAuthenticationSession` web view.

This attempts to fix that by detecting the ability to set cookies using a self redirect (on iOS devices only), and displays a warning to the user if the test fails. This warning replaces a behavior where the user would attempt to submit the login form, and was getting a "The data you submitted is invalid" error.

This PR also sets looser security properties if the oauth provider is running from localhost (dev & test mode), allowing to use Safari in dev.

### DIFF
--- a/packages/oauth/oauth-provider/src/lib/write-form-redirect.ts
+++ b/packages/oauth/oauth-provider/src/lib/write-form-redirect.ts
@@ -11,10 +11,6 @@ export type WriteFormRedirectOptions = SecurityHeadersOptions
 const SCRIPT = js`
 const form = document.forms[0];
 
-setTimeout(() => {
-  form.submit();
-}, 1);
-
 let canSubmit = true;
 
 form.addEventListener('submit', (event) => {
@@ -25,9 +21,9 @@ form.addEventListener('submit', (event) => {
   }
 });
 
-form.addEventListener('error', (event) => {
-  canSubmit = true;
-});
+setTimeout(() => {
+  form.submit();
+}, 1);
 `
 
 // @NOTE If translations and design are needed, consider replacing this with a
--- a/packages/oauth/oauth-provider/src/oauth-provider.ts
+++ b/packages/oauth/oauth-provider/src/oauth-provider.ts
@@ -308,7 +308,7 @@ export class OAuthProvider extends OAuthVerifier {
       ...rest,
       cookie: {
         ...rest.cookie,
-        // secure defaults to true in DeviceManager. For the oauth routes to
+        // "secure" defaults to "true" in DeviceManager. For the oauth routes to
         // work from localhost on Safari, we need to explicitly set secure to
         // false for localhost usage. This is not really an issue with Chrome
         // and Firefox, but Safari enforces it strictly.
--- a/packages/oauth/oauth-provider/src/router/assets/assets.ts
+++ b/packages/oauth/oauth-provider/src/router/assets/assets.ts
@@ -13,8 +13,9 @@ import {
   CrossOriginEmbedderPolicy,
   SecurityHeadersOptions,
 } from '../../lib/http/security-headers.js'
+import { mergeDefaults } from '../../lib/util/object.js'
 import { Simplify } from '../../lib/util/type.js'
-import { writeHtml } from '../../lib/write-html.js'
+import { WriteHtmlOptions, writeHtml } from '../../lib/write-html.js'
 import { parseAssetsManifest } from './assets-manifest.js'
 import { setupCsrfToken } from './csrf.js'
 
@@ -71,7 +72,7 @@ export type SendWebAppOptions = SecurityHeadersOptions & WriteResponseOptions
 export function sendWebAppFactory<P extends keyof HydrationData>(
   page: P,
   customization: Customization,
-  defaults?: SendWebAppOptions,
+  defaults: SendWebAppOptions = {},
 ) {
   // Pre-computed options:
   const customizationData = buildCustomizationData(customization)
@@ -97,18 +98,20 @@ export function sendWebAppFactory<P extends keyof HydrationData>(
       __customizationData: customizationData,
     })
 
-    return writeHtml(res, {
-      ...defaults,
-      ...options,
-      bodyAttrs: {
-        class: 'bg-white text-slate-900 dark:bg-slate-900 dark:text-slate-100',
-      },
-      csp: options?.csp ? mergeCsp(csp, options.csp) : csp,
-      coep: options?.coep ?? CrossOriginEmbedderPolicy.credentialless,
-      meta: [{ name: 'robots', content: 'noindex' }],
-      body: html`<div id="root"></div>`,
-      scripts: [script, ...scripts],
-      styles: [...styles, customizationCss],
-    })
+    return writeHtml(
+      res,
+      mergeDefaults<WriteHtmlOptions>(defaults, options, {
+        bodyAttrs: {
+          class:
+            'bg-white text-slate-900 dark:bg-slate-900 dark:text-slate-100',
+        },
+        csp: options?.csp ? mergeCsp(csp, options.csp) : csp,
+        coep: options?.coep ?? CrossOriginEmbedderPolicy.credentialless,
+        meta: [{ name: 'robots', content: 'noindex' }],
+        body: html`<div id="root"></div>`,
+        scripts: [script, ...scripts],
+        styles: [...styles, customizationCss],
+      }),
+    )
   }
 }
--- a/packages/oauth/oauth-provider/src/router/create-authorization-page-middleware.ts
+++ b/packages/oauth/oauth-provider/src/router/create-authorization-page-middleware.ts
@@ -23,15 +23,11 @@ import type { Awaitable } from '../lib/util/type.js'
 import { writeFormRedirect } from '../lib/write-form-redirect.js'
 import type { OAuthProvider } from '../oauth-provider.js'
 import { parseRequestUri, requestUriSchema } from '../request/request-uri.js'
-import { AuthorizationResultRedirect } from '../result/authorization-result-redirect.js'
 import { sendAuthorizePageFactory } from './assets/send-authorization-page.js'
 import { sendCookieErrorPageFactory } from './assets/send-cookie-error-page.js'
 import { sendErrorPageFactory } from './assets/send-error-page.js'
 import {
-  OAuthRedirectOptions,
-  buildRedirectMode,
-  buildRedirectParams,
-  buildRedirectUri,
+  sendAuthorizationResultRedirect,
   sendRedirect,
 } from './assets/send-redirect.js'
 import { parseRedirectUrl } from './create-api-middleware.js'
@@ -199,7 +195,7 @@ export function createAuthorizationPageMiddleware<
       const result = await server.authorize(query, device)
 
       if ('redirect' in result) {
-        return sendAuthorizeRedirect(res, result)
+        return sendAuthorizationResultRedirect(res, result, securityOptions)
       } else {
         return sendAuthorizePage(req, res, result)
       }
@@ -240,42 +236,27 @@ export function createAuthorizationPageMiddleware<
       try {
         await handler.call(this, req, res)
       } catch (err) {
-        return errorRouteHandler(err, req, res)
+        onError?.(req, res, err, `Authorization Request Error`)
+
+        if (!res.headersSent) {
+          if (err instanceof AuthorizationError) {
+            return sendAuthorizationResultRedirect(
+              res,
+              {
+                issuer: server.issuer,
+                parameters: err.parameters,
+                redirect: err.toJSON(),
+              },
+              securityOptions,
+            )
+          } else {
+            return sendErrorPage(req, res, err)
+          }
+        } else if (!res.destroyed) {
+          res.end()
+        }
       }
     }
-  }
-
-  async function errorRouteHandler(err: unknown, req: Req, res: Res) {
-    onError?.(req, res, err, `Authorization Request Error`)
-
-    if (!res.headersSent) {
-      if (err instanceof AuthorizationError) {
-        return sendAuthorizeRedirect(res, {
-          issuer: server.issuer,
-          parameters: err.parameters,
-          redirect: err.toJSON(),
-        })
-      } else {
-        return sendErrorPage(req, res, err)
-      }
-    } else if (!res.destroyed) {
-      res.end()
-    }
-  }
-
-  function sendAuthorizeRedirect(
-    res: ServerResponse,
-    result: AuthorizationResultRedirect,
-  ) {
-    const { issuer, parameters, redirect } = result
-
-    const redirectOptions: OAuthRedirectOptions = {
-      redirectUri: buildRedirectUri(parameters),
-      mode: buildRedirectMode(parameters),
-      params: buildRedirectParams(issuer, parameters, redirect),
-    }
-
-    return sendRedirect(res, redirectOptions, securityOptions)
   }
 }
 
--- a/packages/oauth/oauth-provider/src/lib/util/object.ts
+++ b/packages/oauth/oauth-provider/src/lib/util/object.ts
@@ -1,0 +1,23 @@
+export function mergeDefaults<T extends object>(
+  defaults: T,
+  ...overrides: (T | undefined)[]
+): T {
+  // @NOTE Not using the spread operator here because TS allows "undefined"
+  // values to be spread, which can lead to defaults being overwritten with
+  // "undefined". This function ensures that only defined values in "options"
+  // will overwrite the corresponding values in "defaults".
+  if (!overrides.length) return defaults
+  if (!overrides.some(Boolean)) return defaults
+  const result: T = { ...defaults } as T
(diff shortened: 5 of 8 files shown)


packages/oauth/oauth-provider-ui/src/lib/json-client.ts
```diff
@@ -42,52 +42,57 @@ export class JsonClient<
         : undefined,
     options?: Options,
   ): Promise<Endpoints[Path]['output']> {
-    const url = new URL(`${this.baseUrl}${path}`)
-    if (method === 'GET') {
-      if (input) {
-        for (const [key, value] of Object.entries(input)) {
-          url.searchParams.set(key, value)
+    try {
```
matthieusieben, line 45: Only diff here is `try/catch` wrapping

packages/oauth/oauth-provider-ui/src/views/error/cookie-error-view.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { Override } from '../../lib/util.ts'
+
+export type CookieErrorViewProps = Override<
+  LayoutWelcomeProps,
+  {
+    continueUrl: string
+  }
+>
+
+export const CookieErrorView = memo(function CookieErrorView({
```
matthieusieben, line 17: New page displayed to iOS users if the cookie test failed.

packages/oauth/oauth-provider/src/device/device-manager.ts
```diff
@@ -204,44 +205,18 @@ export class DeviceManager {
...
-    // the "/oauth/authorize" path to the "/" path. This will only work if the
-    // user visits the "/oauth/authorize" path first.
-
-    const device =
-      this.parseCookie(cookies, `dev-id`, deviceIdSchema) ||
-      this.parseCookie(cookies, 'device-id', deviceIdSchema)
-    const session =
-      this.parseCookie(cookies, `ses-id`, sessionIdSchema) ||
-      this.parseCookie(cookies, 'session-id', sessionIdSchema)
+    const device = this.parseCookie(cookies, `dev-id`, deviceIdSchema)
```
matthieusieben, line 211: Removal of legacy behavior

packages/oauth/oauth-provider/src/lib/http/request.ts
```diff
@@ -139,8 +147,8 @@ export function parseHttpCookies(
   req: IncomingMessage & { cookies?: any },
 ): Record<string, undefined | string> {
   req.cookies ??= req.headers['cookie']
-    ? parseCookie(req.headers['cookie'])
-    : Object.create(null)
+    ? { __proto__: null, ...parseCookie(req.headers['cookie']) }
```
matthieusieben, line 150: Allows to safely access parsed `cookies` with an indexed signature

packages/oauth/oauth-provider/src/lib/http/response.ts
```diff
@@ -81,18 +77,6 @@ export function staticJsonMiddleware(
   }
 }
 
-export type WriteHtmlOptions = WriteResponseOptions & SecurityHeadersOptions
-
-export function writeHtml(
```
matthieusieben, line 86: Moved in `src/lib/write-html.ts`

packages/oauth/oauth-provider/src/lib/http/security-headers.ts
```diff
@@ -50,37 +50,49 @@ export type SecurityHeadersOptions = {
...
+
+export function setSecurityHeaders(
+  res: ServerResponse,
+  options: SecurityHeadersOptions,
+): void {
+  for (const [header, value] of buildSecurityHeaders(options)) {
+    // Only set the header if it is not already set
+    if (!res.hasHeader(header)) {
+      res.setHeader(header, value)
+    }
```
matthieusieben, line 91: Avoid overriding already set headers.

packages/oauth/oauth-provider/src/lib/write-form-redirect.ts
```diff
@@ -0,0 +1,61 @@
+import type { ServerResponse } from 'node:http'
```
matthieusieben, line 1: Copy-paste from `send-redirect` into a shared location (for re-use from `/autorize` endpoint)

packages/oauth/oauth-provider/src/lib/write-form-redirect.ts
```diff
@@ -0,0 +1,61 @@
...
+`
+
+// @NOTE If translations and design are needed, consider replacing this with a
+// proper web app page.
+
+export function writeFormRedirect(
+  res: ServerResponse,
+  uri: string,
+  params: Iterable<[string, string]>,
+  options: WriteFormRedirectOptions,
```
matthieusieben: This is new though, and allows the caller to set `SecurityHeadersOptions`

packages/oauth/oauth-provider/src/lib/write-html.ts
```diff
@@ -1,38 +1,50 @@
...
   // (e.g. 'frame-ancestors'). Therefore, it's better to set the CSP through the
   // HTTP header.
-  const csp = mergeCsp(DEFAULT_CSP, inputCsp, {
-    'base-uri': options.base?.origin as undefined | `https://${string}`,
-    'script-src': options.scripts?.map(assetToCsp).filter((v) => v != null),
-    'style-src': options.styles?.map(assetToCsp).filter((v) => v != null),
-  })
+  const csp = mergeCsp(
+    {
+      'upgrade-insecure-requests': options.hsts !== false,
```
matthieusieben, line 35: This was causing Safari to always load assets (js, css files) using HTTPS, breaking local dev.

packages/oauth/oauth-provider/src/oauth-provider.ts
```diff
@@ -308,7 +304,13 @@ export class OAuthProvider extends OAuthVerifier {
     this.metadata = buildMetadata(this.issuer, this.keyset, metadata)
     this.customization = customizationSchema.parse(rest)
 
-    this.deviceManager = new DeviceManager(deviceStore, deviceManagerOptions)
+    this.deviceManager = new DeviceManager(deviceStore, {
+      ...rest,
+      cookie: {
+        ...rest.cookie,
+        secure: !this.issuer.startsWith('http:'),
```
matthieusieben, line 315: Do not set the `Secure` flag for local dev. Unlike Chrome&FF, Safari does not ignore this on localhost.

packages/oauth/oauth-provider/src/request/request-manager.ts
```diff
@@ -316,7 +316,7 @@ export class RequestManager {
     return parameters
   }
 
-  async get(requestUri: RequestUri, deviceId: DeviceId, clientId?: ClientId) {
+  async get(requestUri: RequestUri, deviceId?: DeviceId, clientId?: ClientId) {
```
matthieusieben, line 319: Allows fetching the request-uri details, even if there is no DeviceID (which is the case when cookies are broken)

packages/oauth/oauth-provider/src/request/request-uri.ts
```diff
@@ -25,3 +27,16 @@ export function decodeRequestUri(requestUri: RequestUri): RequestId {
   const requestIdEnc = requestUri.slice(REQUEST_URI_PREFIX.length)
   return decodeURIComponent(requestIdEnc) as RequestId
 }
+
+export function parseRequestUri(
```
matthieusieben, line 31: Minor code move (into shared location)

packages/oauth/oauth-provider/src/router/assets/assets.ts
```diff
@@ -46,9 +59,56 @@ export const SPA_CSP: CspConfig = {
...
+const HCAPTCHA_CSP: CspConfig = {
   'script-src': ['https://hcaptcha.com', 'https://*.hcaptcha.com'],
   'frame-src': ['https://hcaptcha.com', 'https://*.hcaptcha.com'],
   'style-src': ['https://hcaptcha.com', 'https://*.hcaptcha.com'],
   'connect-src': ['https://hcaptcha.com', 'https://*.hcaptcha.com'],
 }
+
+export type SendWebAppOptions = SecurityHeadersOptions & WriteResponseOptions
+
+export function sendWebAppFactory<P extends keyof HydrationData>(
```
matthieusieben, line 72: Refactor of the web-app setup logic to allow centralized configuration of `SecurityHeadersOptions`.

This also allowed to remove a lot of code duplication when setting up page handlers (see [assets/send-account-page.ts](https://github.com/bluesky-social/atproto/pull/4606/files#diff-b203cca2bafab31ef1c30fc36c2c651742ad7c22fb2c1e565e65b36f1caea746), [assets/send-authorization-page.ts](https://github.com/bluesky-social/atproto/pull/4606/files#diff-644bab78bb2171ba78294d005863ce7ee35abbc76c3f8d39a9b2888eb23ed515) etc.)

packages/oauth/oauth-provider/src/router/create-authorization-page-middleware.ts
```diff
@@ -53,56 +73,129 @@ export function createAuthorizationPageMiddleware<
...
+                )
+                const parameters = await server.decodeJAR(client, query)
+                throw new AuthorizationError(parameters, message)
+              } else {
+                throw new AuthorizationError(query, message)
+              }
+            }
           }
         }
+      }
```
matthieusieben, line 190: This is the "big" change

packages/oauth/oauth-provider-ui/src/lib/json-client.ts
devinivy: So the `console.warn()` below is intentional, is that right?

packages/oauth/oauth-provider-ui/src/views/error/cookie-error-view.tsx
```diff
@@ -0,0 +1,50 @@
...
+  const { t } = useLingui()
+
+  const url = useMemo(() => new URL(continueUrl), [continueUrl])
+
+  return (
+    <LayoutTitlePage {...props} title={title ?? t`Cookie Error`}>
+      <FormCard
+        action={url.origin}
+        method="GET"
+        actions={<Button type="submit" color="primary">{t`Continue`}</Button>}
```
devinivy, line 34: In your mind, what are the tradeoffs to forcing the user to interact by clicking this button versus just "continuing" for them automatically, at least the first time.

packages/oauth/oauth-provider-ui/cookie-error-page.html
```diff
@@ -0,0 +1,116 @@
...
+
+      // Provide a value here to test the "sing-in only" flow
+      const loginHint = undefined // 'alice.test'
+
+      // Use empty array to disable the "sing-up" flow, use a single value to
+      // disable the domain selector.
+      const availableUserDomains = ['.bsky.social', '.bsky.team']
+
+      // Use non empty string to enable hCaptcha during "sing-up" flow
+      const hcaptchaSiteKey = undefined
```
devinivy:
```suggestion
      // Provide a value here to test the "sign-in only" flow
      const loginHint = undefined // 'alice.test'

      // Use empty array to disable the "sign-up" flow, use a single value to
      // disable the domain selector.
      const availableUserDomains = ['.bsky.social', '.bsky.team']

      // Use non empty string to enable hCaptcha during "sign-up" flow
      const hcaptchaSiteKey = undefined
```

packages/oauth/oauth-provider/src/device/device-manager.ts
devinivy: Is the thought that these should all be expired at this point?

packages/oauth/oauth-provider/src/lib/http/request.ts
devinivy: The cookie package doesn't do this?  Huh!

packages/oauth/oauth-provider/src/router/create-authorization-page-middleware.ts
```diff
@@ -53,56 +73,130 @@ export function createAuthorizationPageMiddleware<
       res.setHeader('Cache-Control', 'no-store')
       res.setHeader('Pragma', 'no-cache')
 
-      validateFetchSite(req, ['cross-site', 'none'])
+      validateFetchSite(req, ['same-origin', 'cross-site', 'none'])
```
devinivy, line 77: Is this needed during a "continue" from the error cookie page?

packages/oauth/oauth-provider/src/oauth-provider.ts
devinivy: This would make a good code comment.

packages/oauth/oauth-provider/src/lib/http/security-headers.ts
```diff
@@ -50,37 +50,49 @@ export type SecurityHeadersOptions = {
...
-): void {
+export function* buildSecurityHeaders({
+  csp = { 'default-src': ["'none'"] },
+  coep = CrossOriginEmbedderPolicy.requireCorp,
+  corp = CrossOriginResourcePolicy.sameOrigin,
+  coop = CrossOriginOpenerPolicy.sameOrigin,
+  // @NOTE Setting HSTS headers on a localhost endpoint will basically break
+  // localhost on Safari. In an attempt to avoid being bitten by this, we
+  // disable HSTS in development by default.
+  hsts = process.env.NODE_ENV === 'development' ? false : { maxAge: 63072000 },
```
devinivy: Sprinkling dev mode checks around has a certain convenience to it, but I think it makes it difficult to track the overall behavior of the software in different environments.  There is already precedence for it in `oauth-provider` so I think this is alright, but it is something I hope we can use sparingly or chat about if we want to learn hard into / use more in other packages.

packages/oauth/oauth-provider/src/device/device-manager.ts
matthieusieben: Yep

packages/oauth/oauth-provider/src/lib/http/request.ts
matthieusieben: Nope, I checked.

packages/oauth/oauth-provider/src/lib/write-form-redirect.ts
```diff
@@ -0,0 +1,60 @@
+import type { ServerResponse } from 'node:http'
+import { html, js } from './html/index.js'
+import { setCookie } from './http/request.js'
+import { SecurityHeadersOptions } from './http/security-headers.js'
+import { writeHtml } from './write-html.js'
+
+export type WriteFormRedirectOptions = SecurityHeadersOptions
+
+const SCRIPT = js`
+const form = document.forms[0];
```
devinivy, line 12: This script does look a bit more sophisticated than the previous one.  Is the idea to prevent multiple submissions, since the user can also interact with it?

packages/oauth/oauth-provider-ui/src/views/error/cookie-error-view.tsx
matthieusieben: It does continue automatically the first time! The message is only displayed of that fails. This happens though another piece of code (sendFormRedirect iirc), which is a much lighter (and faster) redirect that does not require loading the wepapp

packages/oauth/oauth-provider-ui/src/lib/json-client.ts
matthieusieben: Yep.

matthieusieben: Yep.

packages/oauth/oauth-provider/src/router/create-authorization-page-middleware.ts
matthieusieben: Yep. The goal of that check was initially to prevent any abuse through getBlob, but same-origin is now required to fir redirecting to self.... This might introduce an issue with blobs, which we could fix by setting the content-disposition header on getBlob. 

Once we do that though, this change opens up the possibility to completely remove this check, allowing appviews working on the same domain to work with oauth

packages/oauth/oauth-provider/src/lib/http/security-headers.ts
matthieusieben: I could actually revert that change as it is a leftover from earlier tests. I was able to "hoist" the HSTS option for all the relevant use cases.

packages/oauth/oauth-provider/src/lib/write-form-redirect.ts
matthieusieben: Yeah, the idea is to make it more robust, in case the user navigates "back" to that page and starts hammering the button

packages/oauth/oauth-provider-ui/src/views/error/cookie-error-view.tsx
devinivy: Oh yeah I saw that later! 👍

PR #4608, opened 2026-02-04
Prevent verification of handle.invalid

ThisIsMissEm:
I think this change should prevent this:
Screenshot 2026-02-04 at 02 21 39

Though, it may actually be worth that endpoint doing a completely fresh resolution for the subject, just to really assert that the handle is definitely valid at that given point in time and that the handle value matches the supplied did.

The account in question above was: https://bsky.app/profile/did:plc:bo44s3mpcgypge54ao5i43w2
Although, I don't see a verification record from bsky.app — I'm guessing it was deleted by something?

cc @foysalit

### DIFF
--- a/.changeset/blue-baboons-peel.md
+++ b/.changeset/blue-baboons-peel.md
@@ -1,0 +1,5 @@
+---
+'@atproto/ozone': patch
+---
+
+Prevent tools.ozone.verification.grantVerification from verifying handle.invalid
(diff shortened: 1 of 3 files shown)


ThisIsMissEm:
Apparently the event the bot received in order to post that message was: https://tangled.org/strings/did:plc:tas6hj2xjrqben5653v5kohk/3mdyqouulyt22

So I believe this change would probably fix this issue, or at least be the start of a fix.

PR #4609, opened 2026-02-04
Fix `method.resolveNoCheck is not a function` 

matthieusieben:
Performing service proxying with `did:toString:foo#bar` results in `method.resolveNoCheck is not a function` because `toString` is a property of the `Object` prototype in JS.

### DIFF
--- a/packages/pds/tests/proxied/proxy-header.test.ts
+++ b/packages/pds/tests/proxied/proxy-header.test.ts
@@ -56,6 +56,18 @@ describe('proxy header', () => {
       parseProxyHeader(network.pds.ctx, ` ${proxyServer.did}#atproto_test`),
     ).rejects.toThrow('proxy header cannot contain spaces')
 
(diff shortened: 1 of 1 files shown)


packages/identity/src/did/did-resolver.ts
```diff
@@ -5,24 +5,28 @@ import { DidPlcResolver } from './plc-resolver'
...
-    const split = did.split(':')
-    if (split[0] !== 'did') {
+    if (!did.startsWith('did:')) {
+      throw new PoorlyFormattedDidError(did)
+    }
+    const methodSepIdx = did.indexOf(':', 4)
+    if (methodSepIdx === -1) {
       throw new PoorlyFormattedDidError(did)
     }
-    const method = this.methods[split[1]]
```
matthieusieben, line 25: `method` could be a method from the `Object.prototype` if `split[1]` is a property name from `Object.prototype` (e.g. `toString`)

PR #4612, opened 2026-02-05
feat(oauth): add domain_handle oauth parameter for sign up

mmattbtw:
(authored with codex)

this makes it so applications with a PDS with multiple available user domains can choose which domain handle to use as default at sign up.

not too sure about putting it in `packages/oauth/oauth-types/src/oauth-authorization-request-parameters.ts` with the oauth spec stuff, but can pull that out if needed. figured I'd go ahead and make the PR to get opinions.

granted this is a problem and solution to something that [only I](https://bsky.app/profile/mmatt.net/post/3me34qpvotc2e) may care about, so if this is out of scope just let me know :^)

### DIFF
--- a/packages/oauth/oauth-provider-ui/src/authorization-page.tsx
+++ b/packages/oauth/oauth-provider-ui/src/authorization-page.tsx
@@ -22,9 +22,13 @@ if (
   url.pathname === '/oauth/authorize' &&
   !url.searchParams.has('request_uri')
 ) {
+  const domainHandle = url.searchParams.get('domain_handle')
   url.search = ''
   url.searchParams.set('client_id', authorizeData.clientId)
   url.searchParams.set('request_uri', authorizeData.requestUri)
+  if (domainHandle) {
+    url.searchParams.set('domain_handle', domainHandle)
+  }
   window.history.replaceState(history.state, '', url.pathname + url.search)
 }
 
--- a/packages/oauth/oauth-provider-ui/src/views/authorize/authorize-view.tsx
+++ b/packages/oauth/oauth-provider-ui/src/views/authorize/authorize-view.tsx
@@ -144,6 +144,7 @@ export function AuthorizeView({
       <SignUpView
         {...props}
         customizationData={customizationData}
+        domainHandle={authorizeData.domainHandle}
         onValidateNewHandle={doValidateNewHandle}
         onBack={showHome}
         onDone={doSignUp}
--- a/packages/oauth/oauth-provider/src/router/assets/send-authorization-page.ts
+++ b/packages/oauth/oauth-provider/src/router/assets/send-authorization-page.ts
@@ -33,6 +33,11 @@ export function sendAuthorizePageFactory(customization: Customization) {
   ): Promise<void> {
     await setupCsrfToken(req, res)
 
+    const domainHandle = getDomainHandle(
+      data.parameters.domain_handle,
+      customizationData.availableUserDomains,
+    )
+
     const script = declareHydrationData<HydrationData['authorization-page']>({
       __customizationData: customizationData,
       __authorizeData: {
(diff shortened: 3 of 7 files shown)


packages/oauth/oauth-types/src/oauth-authorization-request-parameters.ts
```diff
@@ -86,6 +86,12 @@ export const oauthAuthorizationRequestParametersSchema = z.object({
   // How the AS should prompt the user for authorization:
   prompt: oauthPromptModeSchema.optional(),
 
+  // atproto extension: preferred handle domain for account creation
+  domain_handle: z
+    .string()
+    .regex(/^\.[a-z0-9-]+(\.[a-z0-9-]+)*$/)
```
LemmaEOF, line 92: This allows illegal domains - as far back as [RFC 952](https://datatracker.ietf.org/doc/html/rfc952), domain segments must start with `a-z` and may not end with `-`.

packages/oauth/oauth-types/src/oauth-authorization-request-parameters.ts
```diff
@@ -86,6 +86,12 @@ export const oauthAuthorizationRequestParametersSchema = z.object({
   // How the AS should prompt the user for authorization:
   prompt: oauthPromptModeSchema.optional(),
 
+  // atproto extension: preferred handle domain for account creation
+  domain_handle: z
```
LemmaEOF, line 90: If the comment here says "handle domain", I think it would make sense to call this field `handle_domain` across the board - `domain_handle` is backwards and doesn't actually mean anything.

packages/oauth/oauth-provider/src/router/assets/send-authorization-page.ts
```diff
@@ -47,6 +52,7 @@ export function sendAuthorizePageFactory(customization: Customization) {
         uiLocales: data.parameters.ui_locales,
         loginHint: data.parameters.login_hint,
         promptMode: data.parameters.prompt,
+        domainHandle,
```
LemmaEOF, line 55: Should be `domainHandle: domainHandle` to keep consistency in formatting.

packages/oauth/oauth-provider-ui/src/authorization-page.tsx
```diff
@@ -22,9 +22,13 @@ if (
   url.pathname === '/oauth/authorize' &&
   !url.searchParams.has('request_uri')
 ) {
+  const domainHandle = url.searchParams.get('domain_handle')
   url.search = ''
   url.searchParams.set('client_id', authorizeData.clientId)
   url.searchParams.set('request_uri', authorizeData.requestUri)
+  if (domainHandle) {
+    url.searchParams.set('domain_handle', domainHandle)
```
LemmaEOF, line 30: I might just be reading the code wrong, but this feels redundant to me. Save the `domain_handle` from `searchParams` to a const, then...re-apply it almost immediately? Does blanking out `url.search` have a side effect of emptying out params and I don't remember? It feels like this data should be coming from somewhere other than the window href...

packages/oauth/oauth-provider-ui/src/views/authorize/authorize-view.tsx
```diff
@@ -144,6 +144,7 @@ export function AuthorizeView({
       <SignUpView
         {...props}
         customizationData={customizationData}
+        domainHandle={authorizeData.domainHandle}
```
LemmaEOF, line 147: Does `domainHandle` get added to the `authorzeData`? If so, why does the change in `authorization-page.tsx` grab it from the window href instead of the `authorizeData`?

mmattbtw:
overall i think adding a new parameter probably isn't the best idea. imo [using login_hint during signup](https://bsky.app/profile/mmatt.net/post/3me3hsfji222r) to pre fill the handle is a better idea, will probably just open an issue or discussion instead

PR #4616, opened 2026-02-05
Force `getBlob` to trigger a browser download

matthieusieben:
https://github.com/bluesky-social/atproto/discussions/4618

### DIFF
--- a/packages/pds/src/api/com/atproto/sync/getBlob.ts
+++ b/packages/pds/src/api/com/atproto/sync/getBlob.ts
@@ -35,8 +35,23 @@ export default function (server: Server, ctx: AppContext) {
         throw new InvalidRequestError('Blob not found')
       }
       res.setHeader('content-length', found.size)
+
+      // Important Security headers
+
+      // This prevents the browser from trying to guess the content type
+      // and potentially loading the blob as executable code, or rendering it
+      // in some other unsafe way.
       res.setHeader('x-content-type-options', 'nosniff')
+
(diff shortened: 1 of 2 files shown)


packages/pds/src/api/com/atproto/sync/getBlob.ts
```diff
@@ -35,8 +35,23 @@ export default function (server: Server, ctx: AppContext) {
...
+      // This prevents the browser from trying to guess the content type
+      // and potentially loading the blob as executable code, or rendering it
+      // in some other unsafe way.
       res.setHeader('x-content-type-options', 'nosniff')
+
+      // This forces the browser to download the blob instead of trying to
+      // render it when visiting the URL. This is important to prevent XSS
+      // attacks if the blob happens to be HTML. Even if JS is disabled via the
+      // CSP header below, a blob could still contain malicious HTML links.
+      res.setHeader('content-disposition', `attachment; filename="${cid}"`)
```
DavidBuchanan314, line 50: fwiw I do "cid.bin" in millipds https://github.com/DavidBuchanan314/millipds/blob/faf0c36ba65183eba47c1d4f50d084aa11f4ce95/src/millipds/atproto_sync.py#L49 (I don't have strong feelings either way, although I wonder if we should standardise it at all?)

devinivy: I don't think this needs to be standardized, and both options seem cool to me.

PR #4630, opened 2026-02-10
Create index for fetching mod events by creator

gcwill70:
## Summary
Fetching moderation events that a user created can take a while. This PR creates an index on that column and adds a test

## Changes
* Add `moderation_event` index for queries that search for events by creator
* Add `queryEvents` test that mirrors runtime scenario

## How to test

`pnpm test`

or

`pnpm test tests/moderation-events*`

### DIFF
--- a/packages/ozone/src/db/migrations/20260210T154806448Z-mod-event-created-by-indexes.ts
+++ b/packages/ozone/src/db/migrations/20260210T154806448Z-mod-event-created-by-indexes.ts
@@ -10,7 +10,7 @@ export async function up(db: Kysely<unknown>): Promise<void> {
   await db.schema
     .createIndex('moderation_event_created_by_idx')
     .on('moderation_event')
-    .columns(['createdBy', 'createdAt'])
+    .columns(['createdBy', 'createdAt', 'id'])
     .execute()
 }
 


packages/ozone/src/db/migrations/20260210T154806448Z-mod-event-created-by-indexes.ts
```diff
@@ -0,0 +1,13 @@
+import { Kysely } from 'kysely'
+
+export async function up(db: Kysely<unknown>): Promise<void> {
+  await db.schema
+    .createIndex('moderation_event_created_by_idx')
+    .on('moderation_event')
+    .column('createdBy')
+    .execute()
```
ThisIsMissEm, line 14: I'm not sure if kysely does this automatically or not, but make sure this is creating the index concurrently, otherwise this could take out a large ozone server for a while whilst the migration runs

packages/ozone/tests/moderation-events.test.ts
```diff
@@ -465,6 +465,42 @@ describe('moderation-events', () => {
         "only bob's account reports are returned, no events have a URI even though the subjectType is record",
       )
     })
+
+    it('queries events by creator', async () => {
+      const now = new Date()
+      const startDate = new Date(now.getTime() - 24 * 60 * 60 * 1000) // 1 day ago
```
ThisIsMissEm, line 471: Use process.hrtime for this sort of benchmarking — though it's not really a "test" as such

packages/ozone/src/db/migrations/20260210T154806448Z-mod-event-created-by-indexes.ts
gcwill70: Good point! I found another migration in this package that has a comment above, so I added it to this transaction. I assume this is the accepted practice for this package?

Let me know if I should restructure the query itself. It may not be possible to use concurrently in the migration file itself due to an [issue](https://github.com/kysely-org/kysely/issues/352) where kysely doesn't allow non-transactional migrations.

foysalit: I believe the kysely version we are using does not support creating index concurrently.

packages/ozone/tests/moderation-events.test.ts
gcwill70: This test is meant to be functional. Wanted to add this scenario since it's used in prod.

packages/ozone/src/db/migrations/20260210T154806448Z-mod-event-created-by-indexes.ts
ThisIsMissEm: You may want to bypass the index creation this way in production and instead create it concurrently directly with the database, since I think this will be a large database in the bluesky development?

gcwill70: Great point! But we manually run each migration in prod so it is okay to leave as is. Just need a commend to remind whomever is doing the migration in prod.

PR #4634, opened 2026-02-11
Fix invalid handle showing verification

rafaeleyng:
Context:

1. We were returning `handle.invalid` as the handle for a specific profile.
2. While that was going on, someone from the bsky team verified that profile, and then `handle.invalid` got into the verification record.
3. The profile was showing as verified while having the invalid handle badge.

This prevents showing any verification while the handle is invalid (failing bidirectional resolution).

### DIFF
--- a/packages/bsky/src/views/index.ts
+++ b/packages/bsky/src/views/index.ts
@@ -503,7 +503,10 @@ export class Views {
   ): VerificationState | undefined {
     const actor = state.actors?.get(did)
     if (!actor) return
-    if (actor.handle === INVALID_HANDLE) return
+
+    // Currently, the handle comes as "handle.invalid" from the production dataplane.
+    // But the contract allows for empty handle, so we cover both cases.
+    if (!actor.handle || actor.handle === INVALID_HANDLE) return
 
     const isImpersonation = state.labels?.get(did)?.isImpersonation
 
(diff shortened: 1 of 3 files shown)


packages/bsky/src/views/index.ts
```diff
@@ -503,6 +503,7 @@ export class Views {
   ): VerificationState | undefined {
     const actor = state.actors?.get(did)
     if (!actor) return
+    if (actor.handle === INVALID_HANDLE) return
```
devinivy: I see that `actor.handle` is optional, so the question I have here is whether we expect `actor.handle` to be missing here or be equal to `INVALID_HANDLE`.  The tests set things up such that it would be equal to `INVALID_HANDLE`, but I am not sure if that is how it is intended to work.

packages/dev-env/src/seed/verifications.ts
```diff
@@ -108,6 +114,38 @@ export default async (sc: SeedClient) => {
...
+  await createLabel(sc, labelerDid, {
+    src: labelerDid,
+    uri: sc.dids.verifier3,
+    cid: '',
+    val: 'impersonation',
+  })
+
+  await sc.network.bsky.db.db
+    .updateTable('actor')
+    .set({ handle: INVALID_HANDLE })
```
devinivy, line 146: Followup on the previous comment.  This column is nullable—is a missing/bad handle here intended to be stored as `null` or `INVALID_HANDLE`?

packages/bsky/src/views/index.ts
rafaeleyng: We discussed this, the current implementation seems correct but we also want to handle explicitly the empty handle case. Done in 11866af20.

packages/dev-env/src/seed/verifications.ts
rafaeleyng: Done in 11866af20.

devinivy (approved):

PR #4638, opened 2026-02-13
Include knownValues as suggestions into generated types

gaearon:
This makes sure `knownValues` for strings are offered as suggestions in IDEs while unknown values are still allowed.

Screenshot 2026-02-13 at 14 26 49

Note: the actual code was written by Claude. I'm not familiar enough to judge whether the approach is right.


matthieusieben:
I've added tests here https://github.com/bluesky-social/atproto/pull/4639

gaearon:
Possible fix: https://github.com/bluesky-social/atproto/pull/4651

PR #4642, opened 2026-02-16
OAuth client improvements

matthieusieben:


.changeset/full-seals-bathe.md
```diff
@@ -0,0 +1,5 @@
+---
+'@atproto/oauth-client': patch
+---
+
+Delete any pre-existing OAuth session when a new one is created (for a given `sub`)
```
devinivy, line 5: Was this causing a bug, or no bug but was just unnecessarily holding onto expired state?

.changeset/jolly-books-fix.md
```diff
@@ -0,0 +1,5 @@
+---
+'@atproto/oauth-client-browser': patch
+---
+
+Properly await disposal of underlying indexed db when disposing a `BrowserOAuthClient`
```
devinivy, line 5: Did this fix a bug?

.changeset/neat-humans-fix.md
```diff
@@ -0,0 +1,5 @@
+---
+'@atproto/oauth-client': minor
+---
+
+Remove dependency on `EventTarget` (missing in some environements)
```
devinivy:
```suggestion
Remove dependency on `EventTarget` (missing in some environments)
```

.changeset/jolly-books-fix.md
matthieusieben: This is more of a correctness issue. Previously, the *async* disposal of the database was run from a *sync* process, preventing errors from propagating. I've changed the disposal of OAuthBrowser to be async to properly reflect that the underlying resournces also need async disposal.

.changeset/full-seals-bathe.md
matthieusieben: This could potentially cause the client to override a previous session with a new one, forgetting about the previous session. These session are a vector of attack as a part of the protection mechanisms is the fact that a session unexpectedly becomes inactive (because an attacker stole it).

PR #4655, opened 2026-02-18
PDS Actor Store Migrations

DavidBuchanan314:
TODO:

- [x] Auto-migrate actor stores on open
- [x] Synchronize migration state with account db
- [x] Proactive startup-blocking migrate-all loop
- [x] Option to background the migrator loop (non-blocking)
- [x] Handle "stuck" migrations gracefully (consider "in progress" migrations to retryable after some time interval)
- [x] Admin endpoint to get migration status
- [x] Limit migration concurrency
- [x] Figure out how to handle/log migration errors
- [x] Figure out how to test it (on a prod-scale PDS)
- [x] Check that it interacts correctly with deleted/takendown/suspended accounts etc.
- [ ] ~~Think about down-migrations???~~
- [ ] ~~Option to disable migrations entirely?~~

See https://github.com/bluesky-social/atproto/pull/4655/changes#diff-c82b06cb10839b27266be7df6a402de9e0b6fe791a54b29d95878e10dc2d034d for docs of this feature. Note that we want to set `PDS_ACTOR_STORE_MIGRATE_IN_BACKGROUND` to true when deploying on prod PDSes.


lexicons/com/atproto/unspecced/getActorStoreMigrationStatus.json
```diff
@@ -0,0 +1,27 @@
+{
+  "lexicon": 1,
+  "id": "com.atproto.unspecced.getActorStoreMigrationStatus",
```
DavidBuchanan314: I'm on the fence as to whether this should be in `com.atproto.admin` or `com.atproto.unspecced`. It's an admin interface but it's an implementation detail of the PDS and not part of atproto proper.

packages/pds/src/account-manager/helpers/actor-store-migration.ts
```diff
@@ -0,0 +1,86 @@
...
+        const actorDb = await this.actorStore.openDb(claimed.did)
+        actorDb.close()
+      } catch (e) {
+        logger.error(
+          { did: claimed.did, err: e },
+          'Failed to migrate actor store',
+        )
+        if (this.throwOnError) {
+          throw e
+        }
```
DavidBuchanan314, line 62: If `throwOnError` is false (as is the case when it's backgrounded), if the the migration fails (corrupted db?) then we could end up in a tight loop reattempting the same migration on the same db. Could fix this by ordering on storeMigratedAt so we reattempt the least recent db (w/ appropriate index)

packages/pds/src/actor-store/actor-store.ts
```diff
@@ -66,6 +74,70 @@ export class ActorStore {
...
+  private async ensureMigrated(
+    did: string,
+    db: ActorDb,
+    skipConcurrencyLimit?: boolean,
+  ): Promise<void> {
+    const actor = await this.accountDb.db
+      .selectFrom('actor')
+      .select('storeSchemaVersion')
+      .where('did', '=', did)
+      .executeTakeFirst()
```
DavidBuchanan314: We hit the account db with a query on every single actor db open. Can we avoid this? Does it matter?

DavidBuchanan314: Maybe a cache of already-updated dbs.

DavidBuchanan314: Thinking about this again I should probably check the actor db itself

DavidBuchanan314:
I'll add another commit shortly that switches to the `internal.pds.` namespace to see how that goes

packages/pds/src/account-manager/helpers/actor-store-migration.ts
```diff
@@ -0,0 +1,143 @@
...
+        await wait(1000)
+        continue
+      }
+      try {
+        const actorDb = await this.actorStore.openDb(claimed.did, {
+          migrateOnOpen: false,
+        })
+        // Calling migrateToLatestOrThrow expliticly lets us skip the concurrency limits and the early-exit if the db is already migrated
+        // The latter is important in the case where the actor store and account db are out-of-sync.
+        await getMigrator(actorDb).migrateToLatestOrThrow()
```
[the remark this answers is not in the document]
DavidBuchanan314: This is by design

packages/pds/src/actor-store/actor-store.ts
```diff
@@ -66,6 +78,68 @@ export class ActorStore {
...
+        'too many concurrent actor store migrations',
+      )
+    }
+
+    await this.accountDb.db
+      .updateTable('actor')
+      .set({ storeIsMigrating: 1, storeMigratedAt: new Date().toISOString() })
+      .where('did', '=', did as DidString)
+      .where('storeIsMigrating', '=', 0) // don't bump storeMigratedAt if there's a concurrent migration
+      .execute()
```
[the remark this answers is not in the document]
DavidBuchanan314: This is intentional and fine - it is safe if two migrations are attempted concurrently (they are transactional so one will migrate and the other will be a nop), and the concurrency limit is not a *hard* limit.

packages/pds/src/actor-store/actor-store.ts
```diff
@@ -66,6 +78,68 @@ export class ActorStore {
...
+        })
+        .where('did', '=', did as DidString)
+        .execute()
+    } catch (err) {
+      await this.accountDb.db
+        .updateTable('actor')
+        .set({ storeIsMigrating: 0 })
+        .where('did', '=', did as DidString)
+        .execute()
+      throw err
```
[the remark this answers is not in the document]
DavidBuchanan314: I don't think this is a real problem.

packages/pds/src/db/db.ts
```diff
@@ -7,13 +7,31 @@ import {
...
 import { retrySqlite } from './util'
 
+// Kysely 0.22.0's SqliteAdapter incorrectly reports supportsTransactionalDdl = false.
+// SQLite fully supports transactional DDL. Without this fix, the migrator runs migrations
+// outside of a transaction, causing race conditions when multiple migrations start concurrently.
+// packages/pds/tests/concurrent-migration.test.ts tests for this scenario.
+// This is relevant for both account db and actor store db migrations.
+class TransactionalSqliteAdapter extends SqliteAdapter {
+  get supportsTransactionalDdl() {
+    return true
```
DavidBuchanan314, line 25: I should probably PR this to upstream kysely

packages/pds/src/actor-store/db/migrations/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+Migrations must be written with this in mind, and ideally new schemas should never break queries from earlier versions.
+
+If breaking backwards compatibility is unavoidable, explicit schema version checks should be used, and those checks should be deployed prior to the migration itself.
+
+## Env Vars
+
+`PDS_ACTOR_STORE_MIGRATE_IN_BACKGROUND` (bool, default false)
+
+`PDS_ACTOR_STORE_MAX_CONCURRENT_MIGRATIONS` (int, default 1000)
```
dholms: My immediate reaction is this is quite high for a default, especially while running in production & handling normal load. My instinct would be more like 1-10.

packages/pds/src/actor-store/actor-store.ts
```diff
@@ -13,14 +16,16 @@ import { ActorStoreReader } from './actor-store-reader'
...
+import { ActorDb, getDb, getMigrationLevel, getMigrator } from './db'
+import { getLatestStoreSchemaVersion } from './db/migrations'
 
 export class ActorStore {
   reservedKeyDir: string
 
   constructor(
     public cfg: ActorStoreConfig,
     public resources: ActorStoreResources,
+    public accountDb: AccountDb,
```
dholms: It's kinda a shame to have this be a dependency for the `ActorStore` class...

Is there a way we could factor it out? I think it might be comfier if we defined a lightweight interface around the AccountManager class with just the methods needed here and then passed it in in `ActorStoreResources`. Does that make sense?

packages/pds/src/actor-store/actor-store.ts
```diff
@@ -58,6 +67,9 @@ export class ActorStore {
       await retrySqlite(() =>
         db.db.selectFrom('repo_root').selectAll().execute(),
```
dholms: Could we get migration level here instead?

We're already just doing a "dummy read" from the DB to make sure we can access it fine. Might as well read migration level?

packages/pds/src/actor-store/actor-store.ts
```diff
@@ -66,6 +78,68 @@ export class ActorStore {
...
+      lastMigration !== null &&
+      lastMigration >= getLatestStoreSchemaVersion()
+    ) {
+      return
+    }
+
+    // We need to do a migration
+
+    if (
+      (await countInProgressMigrations(this.accountDb)) >=
```
dholms: I think I'd move these functions to the `AccountManager` class and then give a lightweight interface for them

packages/pds/src/actor-store/actor-store.ts
```diff
@@ -66,6 +78,68 @@ export class ActorStore {
...
+      this.cfg.maxConcurrentMigrations
+    ) {
+      throw new InternalServerError(
+        'too many concurrent actor store migrations',
+      )
+    }
+
+    await this.accountDb.db
+      .updateTable('actor')
```
dholms: same with these

packages/pds/src/actor-store/actor-store.ts
```diff
@@ -66,6 +78,68 @@ export class ActorStore {
...
+      .set({ storeIsMigrating: 1, storeMigratedAt: new Date().toISOString() })
+      .where('did', '=', did as DidString)
+      .where('storeIsMigrating', '=', 0) // don't bump storeMigratedAt if there's a concurrent migration
+      .execute()
+
+    try {
+      await getMigrator(db).migrateToLatestOrThrow()
+
+      // NOTE: If the process dies here (after migration but before the account db has been updated),
```
dholms: do we have a way to get them back in sync? even if manual?

lexicons/internal/pds/getActorStoreMigrationStatus.json
```diff
@@ -0,0 +1,43 @@
+{
+  "lexicon": 1,
+  "id": "internal.pds.getActorStoreMigrationStatus",
```
dholms, line 3: ooh la la internal 💅

packages/pds/src/context.ts
```diff
@@ -276,6 +275,26 @@ export class AppContext {
...
+      accountManager.db,
+    )
+
+    const actorStoreMigrator = new ActorStoreMigrator(
+      accountManager.db,
+      actorStore,
+      // fail loudly if it's foregrounded
+      cfg.actorStore.migrateInBackground ? false : true,
+    )
+    actorStoreMigrator.start()
```
dholms: Hmmm was about to say, we should run this in the service lifecycle (in `start` in `pds/src/index.ts`), but I see we have some other db migrations in here as well...

Still the fact that this one may be running in the background the whole time & we stop it in service teardown makes me think we should move at least this one to the `start` method

packages/pds/src/account-manager/db/migrations/008-actor-db-migrator.ts
```diff
@@ -0,0 +1,45 @@
+import { Kysely, sql } from 'kysely'
+
+export async function up(db: Kysely<unknown>): Promise<void> {
+  await db.schema
+    .alterTable('actor')
+    .addColumn(
+      'storeSchemaVersion',
+      'varchar',
+      (col) => col.notNull().defaultTo('001'), // current latest schema version for actor store
```
dholms, line 9: i wonder how long this will take to migrate on some of our PDSes since it has to update ~500k rows to do the table migration. wonder if it makes more sense to just own up to being nullable for now

packages/pds/src/account-manager/db/migrations/008-actor-db-migrator.ts
```diff
@@ -0,0 +1,45 @@
...
+    .addColumn('storeIsMigrating', 'integer', (col) =>
+      col.notNull().defaultTo(0),
+    )
+    .execute()
+  await db.schema
+    .alterTable('actor')
+    .addColumn('storeMigratedAt', 'varchar')
+    .execute()
+
+  // efficiently get unmigrated actors, oldest-first
```
dholms, line 23: hmmm these indexes almost make me want to just make this a separate table...

packages/pds/src/account-manager/helpers/actor-store-migration.ts
```diff
@@ -0,0 +1,151 @@
+import { sql } from 'kysely'
+import { wait } from '@atproto/common'
+import type { ActorStore } from '../../actor-store/actor-store'
+import { getMigrator } from '../../actor-store/db'
+import { getLatestStoreSchemaVersion } from '../../actor-store/db/migrations'
+import { actorStoreMigrationLogger as logger } from '../../logger'
+import type { AccountDb } from '../db'
+
+export const allActorStoresMigrated = async (
```
dholms: as i mentioned elsewehre, I'd just put all these on the `AccountManager` class

packages/pds/src/account-manager/db/migrations/008-actor-db-migrator.ts
DavidBuchanan314: I believe the combo of notNull + default does not require a full rewrite

PR #4658, opened 2026-02-18
Pass through `x-bsky-*` headers to appview

estrattonbailey:


packages/pds/src/pipethrough.ts
```diff
@@ -70,11 +70,20 @@ export const proxyHandler = (ctx: AppContext): CatchallHandler => {
...
 
+      const bskyHeaders = Object.entries(req.headers)
+        .filter(([key]) => key.startsWith('x-bsky-'))
+        .reduce(
+          (acc, [key, value]) => ({
+            ...acc,
+            [key]: value,
+          }),
+          {} as Record<string, string | string[] | undefined>,
+        )
```
matthieusieben, line 81: Inline functions are generally pretty bad for GC ([more objects created means more objects in old space](https://youtube.com/watch?v=_OnTUIYxGRs&t=603)). Since this is a hot path, I would recommend using a loop:

```ts
const headers: IncomingHttpHeaders = {
  ...
}

// Headers pass through
for (const name in req.headers) {
  if (!(name in headers) && isProxiedHeader(name)) {
    headers[name] = req.headers[name]
  }
}

function isProxiedHeader(name: string) {
  // @TODO adapt based on protocol discussion
  return name.startsWith('x-bsky-')
}
```

estrattonbailey:
@matthieusieben yep you're right, I just pushed this up as a talking point, should have marked as a draft. We've got a thread going in Slack, will tag you!

PR #4663, opened 2026-02-19
Ozone User Assignments

gcwill70:
Allows for assigning users to reports and queues. (Does not validate report and queue IDs yet.)

# Changes
* Create `moderator_assignment` db table to store ozone assignments to queues and reports
* Create endpoints for managing assignments
* Create assignment service for handling report and queue assignments

# How to test

`pnpm --filter ozone test -- tests/report-assignment.test.ts tests/queue-assignment.test.ts`

### DIFF
--- a/packages/ozone/src/assignment/index.ts
+++ b/packages/ozone/src/assignment/index.ts
@@ -5,6 +5,7 @@ import { Database } from '../db'
 import { EndAtIdKeyset, paginate } from '../db/pagination'
 import { ModeratorAssignment } from '../db/schema/moderator_assignment'
 import { QueueServiceCreator } from '../queue/service'
+import { ReportQueue } from '../db/schema/report_queue'
 
 export interface AssignmentServiceOpts {
   queueDurationMs: number
@@ -49,12 +50,14 @@ type AssignmentRowWithQueue = Selectable<ModeratorAssignment> & {
   queueCreatedAt: string | null
   queueUpdatedAt: string | null
   queueEnabled: boolean | null
+  queueDeletedAt: string | null
 }
 
 export class AssignmentService {
   constructor(
     public db: Database,
     public opts: AssignmentServiceOpts,
+    public queueServiceCreator: QueueServiceCreator,
   ) {}
 
   async getQueueAssignments(input: GetQueueAssignmentsInput): Promise<{
@@ -81,6 +84,7 @@ export class AssignmentService {
         'report_queue.createdAt as queueCreatedAt',
         'report_queue.updatedAt as queueUpdatedAt',
         'report_queue.enabled as queueEnabled',
+        'report_queue.deletedAt as queueDeletedAt',
       ])
       .where('reportId', 'is', null)
       .where('queueId', 'is not', null)
@@ -114,7 +118,7 @@ export class AssignmentService {
     const results = await paginatedQuery.execute()
 
     return {
-      assignments: results.map((row) => this.viewQueue(row)),
+      assignments: results.map((row) => this.viewQueueAssignment(row)),
       cursor: keyset.packFromResult(results),
     }
   }
@@ -143,6 +147,7 @@ export class AssignmentService {
         'report_queue.createdAt as queueCreatedAt',
         'report_queue.updatedAt as queueUpdatedAt',
         'report_queue.enabled as queueEnabled',
+        'report_queue.deletedAt as queueDeletedAt',
       ])
       .where('reportId', 'is not', null)
 
@@ -178,7 +183,7 @@ export class AssignmentService {
     const results = await paginatedQuery.execute()
 
     return {
-      assignments: results.map((row) => this.viewReport(row)),
+      assignments: results.map((row) => this.viewReportAssignment(row)),
       cursor: keyset.packFromResult(results),
     }
   }
@@ -246,11 +251,12 @@ export class AssignmentService {
         'report_queue.createdAt as queueCreatedAt',
         'report_queue.updatedAt as queueUpdatedAt',
         'report_queue.enabled as queueEnabled',
+        'report_queue.deletedAt as queueDeletedAt',
       ])
       .where('moderator_assignment.id', '=', result.id)
       .executeTakeFirstOrThrow()
 
-    return this.viewQueue(row)
+    return this.viewQueueAssignment(row)
   }
 
   async assignReport(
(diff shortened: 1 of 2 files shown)


lexicons/tools/ozone/queue/assign.json
```diff
@@ -0,0 +1,38 @@
+{
+  "lexicon": 1,
+  "id": "tools.ozone.queue.assign",
```
foysalit: we have had some conversations around using verbs for lexicon procedures/queries and while there are still some lexicons that use this style, we decided not to do it this way and instead, always combine verb and subject. so instead of `tools.ozone.queue.assign` i'd recommend something like `tools.ozone.queue.assignModerator` or `tools.ozone.queue.assignMember`

lexicons/tools/ozone/queue/assignModerator.json
```diff
@@ -0,0 +1,38 @@
+{
+  "lexicon": 1,
+  "id": "tools.ozone.queue.assign",
+  "defs": {
+    "main": {
+      "type": "procedure",
+      "description": "Assign a user to a queue.",
```
foysalit, line 7: may be add some clarification here that it's not always for "hard assignment" but more like a `view` unless `assign: true`? which makes me think, if we should try to come up with a better name than `tools.ozone.queue.assignModerator` so that it's more generic? like `tools.ozone.queue.registerView`? happy to hear alternatives though.

foysalit: on 2nd thought, may be we don't need a "view history" for queues? I think we only care about who is assigned to a queue and not who is viewing so may be we can keep it as `tools.ozone.queue.assignModerator` and get rid of the assign property?

lexicons/tools/ozone/queue/getAssignments.json
```diff
@@ -0,0 +1,55 @@
...
+            "items": {
+              "type": "string",
+              "format": "did"
+            },
+            "description": "If specified, returns assignments for these moderators only."
+          },
+          "subject": {
+            "type": "string",
+            "format": "uri",
+            "description": "If specified as a DID, returns assignments for all records and the DID. If specified as an AT-URI, returns assignments for that URI only."
```
foysalit: I wrote the specs so it's totally my fault but this line doesn't explain very well what is the relationship of the subject to the queue. is it gonna return queues that the subject belongs to?

lexicons/tools/ozone/report/claimReport.json
```diff
@@ -0,0 +1,44 @@
...
+          "type": "object",
+          "required": ["reportId"],
+          "properties": {
+            "reportId": {
+              "type": "integer",
+              "description": "The ID of the report to claim."
+            },
+            "queueId": {
+              "type": "integer",
+              "description": "Optional queue ID to associate the claim with."
```
foysalit: this will default to the queueId that the report belongs to, right? may be worth specifying here?

lexicons/tools/ozone/report/defs.json
```diff
@@ -218,6 +218,33 @@
     "reasonSelfHarmOther": {
       "type": "token",
       "description": "Other dangerous content"
+    },
+    "assignmentView": {
```
foysalit, line 222: wonder if the `assignment` term here is a bit of a mismatch with the `claimReport` language? am i overthinking this?

packages/ozone/src/api/queue/assignModerator.ts
```diff
@@ -0,0 +1,87 @@
+import { ForbiddenError, InvalidRequestError } from '@atproto/xrpc-server'
+import { AppContext } from '../../context'
+import { Server } from '../../lexicon'
+
+const ASSIGNMENT_DURATION_MS = 5 * 60 * 1000 // 5 minutes
```
foysalit: should we move this into env var so that folks hosting their own instance can configure without changing code?

packages/ozone/src/api/queue/assignModerator.ts
```diff
@@ -0,0 +1,87 @@
...
+      }
+
+      // access
+      if (did !== authDid && !auth.credentials.isAdmin) {
+        throw new ForbiddenError('Cannot assign others')
+      }
+
+      const result = await ctx.db.transaction(async (dbTxn) => {
+        // Check for an existing active assignment for this moderator + queue
+        const existing = await dbTxn.db
```
foysalit: this is prolly fine but traditionally, we have tried to keep direct queries out of the handlers and abstract them away in service files. if you dont need a full-fledged class for this but having them in small utility functions may be enough.

packages/ozone/src/api/queue/getAssignments.ts
```diff
@@ -0,0 +1,52 @@
...
+      }
+
+      if (dids?.length) {
+        query = query.where('did', 'in', dids)
+      }
+
+      if (subject) {
+        if (subject.startsWith('at://')) {
+          // AT-URI: return assignments for that specific URI's report
+          query = query.where('reportId', 'is not', null)
```
foysalit: hmm... not sure if i understand this one. if subject is at-uri, we are returning everything that doesn't have a reportId? that doesn't sound right?

packages/ozone/src/db/migrations/20260219T165302248Z-moderator-assignment.ts
```diff
@@ -0,0 +1,26 @@
+import { Kysely, sql } from 'kysely'
+
+export async function up(db: Kysely<unknown>): Promise<void> {
+  await db.schema
```
foysalit, line 4: we probably need some very specific indices here for faster lookups. luckily, we have very few variations of queries on this table.

lexicons/tools/ozone/queue/assign.json
gcwill70: Sounds good! Will rename to `tools.ozone.queue.assignModerator`

lexicons/tools/ozone/queue/getAssignments.json
gcwill70: This was a parameter that I needed to take a second look at. I didn't use it while testing the core requirements

But yes right now it will return the users's assignments if passed in as a DID. For AT-URI, I had it return only report assignments. You would use it in conjunction with `dids` to get a users specific reports they have claimed.

Kind of a confusing design currently. I think we should list out all the use cases for this endpoint so we can clear up the parameters we need.

lexicons/tools/ozone/queue/assignModerator.json
gcwill70: Removed the assign parameter. Will just let queue assignments expire naturally

lexicons/tools/ozone/report/defs.json
gcwill70: Renamed claimReport to assignModerator

lexicons/tools/ozone/queue/defs.json
```diff
@@ -107,6 +107,30 @@
...
+      "required": ["id", "did", "queueId", "startAt", "endAt"],
+      "properties": {
+        "id": {
+          "type": "integer"
+        },
+        "did": {
+          "type": "string",
+          "format": "did"
+        },
+        "queueId": {
```
foysalit: should we replace the queueId with a queueView here which includes the full queue config?

lexicons/tools/ozone/report/defs.json
```diff
@@ -218,6 +218,33 @@
...
+      "required": ["id", "did", "reportId", "startAt", "endAt"],
+      "properties": {
+        "id": {
+          "type": "integer"
+        },
+        "did": {
+          "type": "string",
+          "format": "did"
+        },
+        "queueId": {
```
foysalit: similarly, wonder if we should send back a full queueView here instead of queueId alone.

packages/ozone/src/api/queue/getAssignments.ts
```diff
@@ -0,0 +1,19 @@
...
+  server.tools.ozone.queue.getAssignments({
+    auth: ctx.authVerifier.modOrAdminToken,
+    handler: async ({ params }) => {
+      const assignments =
+        await ctx.assignmentService.getQueueAssignments(params)
+
+      return {
+        encoding: 'application/json',
+        body: {
+          assignments,
```
foysalit: do we want to get a cursor here too? would rather have it in place and not need it immediately than having to add it later down the road, wdyt?

packages/ozone/src/db/migrations/20260219T165302248Z-moderator-assignment.ts
```diff
@@ -0,0 +1,22 @@
...
+    // assignee
+    .addColumn('did', 'text', (col) => col.notNull())
+    // assigned over
+    .addColumn('reportId', 'integer')
+    .addColumn('queueId', 'integer')
+    // validity
+    .addColumn('startAt', 'timestamptz', (col) =>
+      col.defaultTo(sql`now()`).notNull(),
+    )
+    .addColumn('endAt', 'timestamptz', (col) => col.notNull())
```
foysalit: traditionally, we've always opted for varchar instead of timestamp columns and used date.toISOString() for dates. i haven't found any issues with it so far so may be we can stick to that pattern?

gcwill70: fixed!

packages/ozone/src/api/queue/getAssignments.ts
gcwill70: done!

lexicons/tools/ozone/report/defs.json
```diff
@@ -218,6 +218,34 @@
...
+      "required": ["id", "did", "reportId", "startAt", "endAt"],
+      "properties": {
+        "id": {
+          "type": "integer"
+        },
+        "did": {
+          "type": "string",
+          "format": "did"
+        },
+        "queueView": {
```
foysalit: I feel like the prop name should just be `queue`?

lexicons/tools/ozone/queue/defs.json
```diff
@@ -109,6 +109,31 @@
...
+      "required": ["id", "did", "queueView", "startAt", "endAt"],
+      "properties": {
+        "id": {
+          "type": "integer"
+        },
+        "did": {
+          "type": "string",
+          "format": "did"
+        },
+        "queueView": {
```
foysalit: weird that i don't actually see the definition for queueView anywhere?

packages/ozone/src/api/queue/assignModerator.ts
```diff
@@ -0,0 +1,41 @@
+import { ForbiddenError, InvalidRequestError } from '@atproto/xrpc-server'
+import { AppContext } from '../../context'
+import { Server } from '../../lexicon'
+
+export default function (server: Server, ctx: AppContext) {
+  server.tools.ozone.queue.assignModerator({
+    auth: ctx.authVerifier.modOrAdminToken,
+    handler: async ({ input, auth }) => {
+      const queueId = input.body.queueId
+      const authDid =
```
foysalit: we've been doing this in a few places now, should we just get a util function for this?

packages/ozone/src/api/queue/assignModerator.ts
```diff
@@ -0,0 +1,41 @@
...
+        throw new InvalidRequestError('DID is required')
+      }
+
+      // RBAC
+      if (!auth.credentials.isModerator) {
+        throw new ForbiddenError('Unauthorized')
+      }
+
+      // RuBAC
+      if (did !== authDid && !auth.credentials.isAdmin) {
```
foysalit, line 25: nice!

packages/ozone/src/assignment/index.ts
```diff
@@ -0,0 +1,394 @@
...
+
+  viewQueue(row: AssignmentRowWithQueue): ToolsOzoneQueueDefs.AssignmentView {
+    if (row.queueId === null || row.queueName === null) {
+      throw new Error('Queue data missing for queue assignment')
+    }
+
+    return {
+      id: row.id,
+      did: row.did,
+      queueView: {
```
foysalit: I expect this to be necessary in a couple places so probably worth getting a helper method to take queue and shape the queueView for it.

lexicons/tools/ozone/queue/defs.json
gcwill70: The intent was to reference queueView defined on line 5 in this file. It appears this is working properly in the generated code...

lexicons/tools/ozone/queue/assignModerator.json
```diff
@@ -0,0 +1,34 @@
...
+          "type": "object",
+          "required": ["queueId"],
+          "properties": {
+            "queueId": {
+              "type": "integer",
+              "description": "The ID of the queue to assign the user to."
+            },
+            "did": {
+              "type": "string",
+              "description": "DID to be assigned. Assigns to whomever sent the request if not provided."
```
pfrazee, line 20: What happens if this isn't provided? If the goal is to assign nobody, it might be smart to require the did and treat an empty string as "no assignment"

lexicons/tools/ozone/report/assignModerator.json
```diff
@@ -0,0 +1,45 @@
...
+              "description": "The ID of the report to assign."
+            },
+            "queueId": {
+              "type": "integer",
+              "description": "Optional queue ID to associate the assignment with. If not provided and the report has been assigned on a queue before, it will stay on that queue."
+            },
+            "assign": {
+              "type": "boolean",
+              "default": true,
+              "description": "Whether to assign or un-assign the report to the user."
```
pfrazee, line 25: I'd suggest a separate method for unassigning moderators. This is a little clunky

lexicons/tools/ozone/queue/assignModerator.json
gcwill70: We added the default behavior as a convenience but requiring some intentionality here makes sense to me., Going to require this field in another branch

PR #4681, opened 2026-02-23
❇️ Action reports via events

foysalit:
With the new report based queue system in ozone, we need a way to `action` reports and this PR implements action association via existing events.

### DIFF
--- a/lexicons/tools/ozone/moderation/defs.json
+++ b/lexicons/tools/ozone/moderation/defs.json
@@ -241,11 +241,7 @@
           "ref": "com.atproto.moderation.defs#subjectType"
         },
         "subject": {
-          "type": "union",
-          "refs": [
-            "com.atproto.admin.defs#repoRef",
-            "com.atproto.repo.strongRef"
-          ]
+          "type": "string"
         },
         "status": {
           "type": "ref",
--- a/packages/api/src/client/lexicons.ts
+++ b/packages/api/src/client/lexicons.ts
@@ -16250,11 +16250,7 @@ export const schemaDict = {
             ref: 'lex:com.atproto.moderation.defs#subjectType',
           },
           subject: {
-            type: 'union',
-            refs: [
-              'lex:com.atproto.admin.defs#repoRef',
-              'lex:com.atproto.repo.strongRef',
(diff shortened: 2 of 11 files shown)


lexicons/tools/ozone/moderation/emitEvent.json
```diff
@@ -85,6 +90,34 @@
...
+          },
+          "description": "Target reports matching these report types on the subject (fully qualified NSIDs)"
+        },
+        "all": {
+          "type": "boolean",
+          "description": "Target ALL reports on the subject"
+        },
+        "note": {
+          "type": "string",
+          "description": "Note to send to reporter(s) when actioning their report"
```
gcwill70, line 118: Possible requirement for down the line is some templates for these notes e.g. 'Your report was viewed and it was determined that no action is necessary.'

packages/ozone/src/api/moderation/emitEvent.ts
```diff
@@ -224,6 +225,21 @@ const handleModerationEvent = async ({
...
+    // Validate reportAction if provided (actual processing happens after event is logged)
+    if (input.body.reportAction) {
+      // Validate that at least one targeting criteria is provided
+      const { reportAction } = input.body
+      if (
+        !reportAction.ids?.length &&
+        !reportAction.types?.length &&
+        !reportAction.all
+      ) {
+        throw new InvalidRequestError(
```
gcwill70, line 237: Is there lexicon syntax for generating client only logic like this? Would be nice to have codegen generate some validator that can be used on the client and server...

lexicons/tools/ozone/moderation/emitEvent.json
foysalit: yep! would probably move this to the client though? since we've been generating automated email templates on the client for the most part. could also do it via options api if the templates are not that many.

packages/ozone/src/api/moderation/emitEvent.ts
foysalit: not that i'm aware of. we can make certain properties required but there's no way of requiring `at least one of the properties`

PR #4686, opened 2026-02-25
❇️ Ozone queue management

foysalit:
This PR adds CRUD operations on queue entity for ozone. The new queue management ensures basic filter based unique queue creation based on report properties.

### DIFF
--- a/packages/ozone/src/api/queue/createQueue.ts
+++ b/packages/ozone/src/api/queue/createQueue.ts
@@ -1,6 +1,8 @@
-import { AuthRequiredError } from '@atproto/xrpc-server'
+import { AuthRequiredError, InvalidRequestError } from '@atproto/xrpc-server'
 import { AppContext } from '../../context'
 import { Server } from '../../lexicon'
+
+const VALID_SUBJECT_TYPES = ['account', 'record', 'message']
 
 export default function (server: Server, ctx: AppContext) {
   server.tools.ozone.queue.createQueue({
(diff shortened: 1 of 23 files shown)


packages/ozone/src/api/queue/listQueues.ts
```diff
@@ -0,0 +1,37 @@
+import { AppContext } from '../../context'
+import { Server } from '../../lexicon'
+
+export default function (server: Server, ctx: AppContext) {
+  server.tools.ozone.queue.listQueues({
+    auth: ctx.authVerifier.modOrAdminToken,
+    handler: async ({ params }) => {
+      const {
+        limit = 50,
```
gcwill70, line 9: I meant to ask about validation. I'm assuming lexicon defaults and range limits aren't enforced within lexicon's generated code? Is data shape the only validation that is enforced?

packages/ozone/src/queue/service.ts
```diff
@@ -0,0 +1,190 @@
...
+    return await this.db.db
+      .updateTable('report_queue')
+      .set({ ...updates, updatedAt: now })
+      .where('id', '=', id)
+      .returningAll()
+      .executeTakeFirstOrThrow()
+  }
+
+  async delete(id: number): Promise<void> {
+    await this.db.db.deleteFrom('report_queue').where('id', '=', id).execute()
```
gcwill70: We should consider soft deletes. I could see mods wanting to un-do a delete if the overflow queue starts filling up too fast. A "last enabled" relative timestamp could also be helpful context when deciding to un-delete a queue or not.

packages/ozone/src/api/queue/createQueue.ts
```diff
@@ -0,0 +1,41 @@
...
+  server.tools.ozone.queue.createQueue({
+    auth: ctx.authVerifier.modOrAdminToken,
+    handler: async ({ input, auth }) => {
+      const access = auth.credentials
+
+      if (!access.isModerator) {
+        throw new AuthRequiredError('Must be a moderator to create a queue')
+      }
+
+      const { name, subjectTypes, collection, reportTypes } = input.body
```
gcwill70, line 17: Wondering if validation is needed here (e.g. `knownValues` for `subjectTypes`, `reportTypes` max length, and `collection` being a valid NSID).

Just wanting to make sure these fields are validated since each queue adds overhead to every report processed right?

gcwill70 (changes requested):
Looks great! Not all comments are important but brought up a couple points to consider

packages/ozone/src/queue/service.ts
foysalit: yh good idea but i want deletes to be irreversible. we can always recreate a queue with same params and reprocess reports. checkout the queue router implementation PR which uses cursor based report processing for routing reports.

packages/ozone/src/api/queue/listQueues.ts
foysalit: all available validation tooling available in lexicons is is enforced. there are just some validation approaches that we don't support... yet through lexicons.

packages/ozone/src/api/queue/createQueue.ts
foysalit: good call. yeah we stopped validating reportTypes  a while ago but for subjectTypes, it makes sense to have some implementation level validation for now and not hoist it up to lexicon level. for collection, it feels too broad to set up validation logic.

PR #4687, opened 2026-02-26
❇️ Ozone queue routing

foysalit:
This PR adds a background job that periodically routes incoming reports into preconfigured queues based on the report's properties such as report type, subject type, subject collection etc.

### DIFF
--- a/packages/ozone/src/api/queue/createQueue.ts
+++ b/packages/ozone/src/api/queue/createQueue.ts
@@ -1,6 +1,8 @@
-import { AuthRequiredError } from '@atproto/xrpc-server'
+import { AuthRequiredError, InvalidRequestError } from '@atproto/xrpc-server'
 import { AppContext } from '../../context'
 import { Server } from '../../lexicon'
+
+const VALID_SUBJECT_TYPES = ['account', 'record', 'message']
 
 export default function (server: Server, ctx: AppContext) {
   server.tools.ozone.queue.createQueue({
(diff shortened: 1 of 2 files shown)


packages/ozone/src/daemon/strike-expiry-processor.ts
gcwill70, line 1: These kind of updates are satisfying 😌

packages/ozone/src/db/migrations/20260225T000000000Z-add-report-queue-table.ts
```diff
@@ -17,19 +17,24 @@ export async function up(db: Kysely<unknown>): Promise<void> {
...
-  // Leading enabled column also satisfies standalone enabled-only filters,
-  // making a separate idx_queue_enabled redundant.
-  await db.schema
-    .createIndex('idx_queue_created_at')
-    .on('report_queue')
-    .columns(['enabled', 'createdAt', 'id'])
-    .execute()
+  // Partial unique index on name — only enforces uniqueness for non-deleted queues,
+  // so a soft-deleted queue's name can be reused by a new queue.
+  await sql`CREATE UNIQUE INDEX idx_queue_name_unique ON report_queue (name) WHERE "deletedAt" IS NULL`.execute(
```
gcwill70, line 25: Nice! Clean way to get conditional uniqueness

packages/ozone/src/queue/service.ts
```diff
@@ -89,6 +92,7 @@ export class QueueService {
       .selectFrom('report_queue')
       .selectAll()
       .where('id', '=', id)
+      .where('deletedAt', 'is', null)
```
gcwill70, line 95: Could add an optional `includeDeleted` param to get methods in this service. Would add some intentionality that would be clearer to callers.

packages/ozone/src/queue/service.ts
```diff
@@ -106,7 +110,30 @@ export class QueueService {
   }
 
   async delete(id: number): Promise<void> {
-    await this.db.db.deleteFrom('report_queue').where('id', '=', id).execute()
+    const now = new Date().toISOString()
+    await this.db.db
+      .updateTable('report_queue')
+      .set({ deletedAt: now })
```
gcwill70, line 116: Probably should add `updatedAt` to this update

packages/ozone/src/queue/service.ts
foysalit: as of now, we don't really need to ever query deleted stuff. we will need them soon but not for individual queue queries and I'll tackle it there.

packages/ozone/src/queue/service.ts
foysalit: hmm... idk, deleted itself is a timestamp, feels a bit redundant to update the updatedAt timestamp too.

PR #4688, opened 2026-02-26
Lex SDK error handling improvements

matthieusieben:
This PR extract changes made in https://github.com/bluesky-social/atproto/pull/4408 related to error management in Lex SDK.

The aim is to better align the way errors are treated through the new SDK with the way they work in `@atproto/xrpc` and `@atproto/xrpc-server`.

### DIFF
--- a/packages/lex/lex-server/README.md
+++ b/packages/lex/lex-server/README.md
@@ -558,7 +558,6 @@ router.add(schema, async ({ params }) => {
     {
       status: 201,
       headers: {
-        'Content-Type': 'application/json',
         'X-Custom-Header': 'value',
       },
     },


.changeset/happy-houses-start.md
```diff
@@ -0,0 +1,6 @@
+---
+'@atproto/lex-client': patch
+'@atproto/lex-server': patch
+---
+
+Update error management to be more aligned with the way errors work in `@atproto/xrpc` and `@atproto/xrpc-server`
```
devinivy, line 6: Just double-checking that this is indeed non-breaking.

packages/lex/lex-client/src/errors.ts
```diff
@@ -393,3 +458,39 @@ export function asXrpcFailure<M extends Procedure | Query>(
...
+    for (const name of connection.split(',')) {
+      result.delete(name.trim())
+    }
+  }
+
+  // These are not actually hop-by-hop headers, but we remove them because they
+  // pertain to the transport between us and the upstream server, and should not
+  // be forwarded to downstream clients.
+  result.delete('content-length')
+  result.delete('content-encoding')
```
devinivy: These can transfer just fine if you don't touch the bytes, right?  I assume `fetch()` or similar is forcing us to apply decompression before passing along the result, and that is at least partly why we don't want to include these.

matthieusieben: Thing is, since the payload is parsed (jsonParsed), stored into an error, and then re-encoded from the error into a payload, so we don't keep the raw original payload here.

matthieusieben: will update the comment

.changeset/happy-houses-start.md
matthieusieben: we are still on 0.0.x versioning

packages/lex/lex-server/src/errors.ts
```diff
@@ -60,62 +114,13 @@ export class LexServerAuthError<
     readonly wwwAuthenticate: WWWAuthenticate = {},
```
devinivy, line 114: I don't think it's something to address as part of this work, but seeing this for the first time did catch my attention a little because I believe there can be multiple challenges with the same scheme.  And it doesn't look like that can be modeled with `WWWAuthenticate` at the moment.

packages/lex/lex-server/src/lex-server.ts
```diff
@@ -934,23 +971,52 @@ export class LexRouter {
     request: Request,
     connection?: ConnectionInfo,
   ): Promise<Response> => {
-    const nsid = extractMethodNsid(request)
+    const { pathname } = new URL(request.url)
+
+    if (!pathname.startsWith('/xrpc/')) {
```
devinivy: The required `/xrpc` base path should probably be pulled out into a constant somewhere.

packages/lex/lex-server/src/lex-router.ts
```diff
@@ -934,23 +971,52 @@ export class LexRouter {
...
     connection?: ConnectionInfo,
   ): Promise<Response> => {
-    const nsid = extractMethodNsid(request)
+    const { pathname } = new URL(request.url)
+
+    if (!pathname.startsWith('/xrpc/')) {
+      // Handle non XRPC paths
+      const { fallback } = this.options
+      if (fallback) return fallback(request, connection)
+      return new Response('Not Found', { status: 404 })
```
devinivy, line 990: Is there a way in node to combine lex server and, say, an express server or raw http server?  Ideally there is a way to use these together so that xrpc can be incorporated into existing software.

packages/lex/lex-server/src/lex-server.ts
```diff
@@ -934,23 +971,52 @@ export class LexRouter {
...
+    const nsid = pathname.slice(6)
 
-    if (!nsid || !isNsidString(nsid)) {
+    const atprotoProxy = request.headers.get('atproto-proxy')
+    if (atprotoProxy == null) {
+      const handler = (this.handlers as Map<string, FetchHandler>).get(nsid)
+      if (handler) return handler(request, connection)
+    }
+
+    if (!isNsidString(nsid)) {
```
devinivy: Is it possible to maintain our `/xrpc/_health` endpoints?

packages/lex/lex-server/src/lex-server.ts
```diff
@@ -934,23 +971,52 @@ export class LexRouter {
...
+          {
+            error: 'InvalidRequest',
+            message: `Invalid atproto-proxy header value: ${atprotoProxy}`,
+          },
+          { status: 400 },
+        )
+      }
+
+      // @TODO actually implement service proxying logic here. The reason it was
+      // not done already if because we want to perform all the heavy lifting
```
devinivy:
```suggestion
      // not done already is because we want to perform all the heavy lifting
```

packages/lex/lex-server/src/lex-server.ts
```diff
@@ -934,23 +971,52 @@ export class LexRouter {
...
+    // @NOTE we don't validate the NSID format *before* checking for the
+    // handler because the presence of a handler is sufficient to ensure that
+    // the NSID is valid. If there is no handler, *then* we can check if the
+    // NSID is invalid in order to return a more specific error message.
+    const nsid = pathname.slice(6)
 
-    if (!nsid || !isNsidString(nsid)) {
+    const atprotoProxy = request.headers.get('atproto-proxy')
+    if (atprotoProxy == null) {
+      const handler = (this.handlers as Map<string, FetchHandler>).get(nsid)
```
devinivy: I believe there is implied case-sensitivity here.  This will be a change, and I think we're still nailing down final loose ends around the spec.  Let's make sure we assess impact of switching services to case-sensitive xrpc routing before shipping it.

packages/lex/lex-server/src/lex-server.ts
```diff
@@ -1052,9 +1126,35 @@ function encodeMessageFrame(method: Subscription, value: LexValue): Uint8Array {
...
+  did: DidString
+  serviceId: string
+}
+
+function parseAtprotoProxyHeader(value: string): ServiceProxyInfo | null {
+  // /!\ Hot path
+
+  // Basic sanity checks
+  if (!value.startsWith('did:')) return null
+  if (value.includes(' ')) return null
```
devinivy: Is there something specific this check is detecting?

packages/lex/lex-server/src/lex-router.ts
matthieusieben: This package also exports utilities to be used from nodejs. See the `"./nodejs"` export form package.json. An `LexRouter` can be both used as a specific route of an existing express/raw http server, and can be extended through the `fallback` to behave like an express/raw http server middleware.

packages/lex/lex-server/src/lex-server.ts
matthieusieben: Not really. Will remove.

packages/lex/lex-server/src/lex-server.ts
matthieusieben: Added normalization code to normalize the domain part of the NSID

packages/lex/lex-server/src/lex-server.ts
matthieusieben: Added support for it

packages/lex/lex-server/src/lex-router.ts
matthieusieben: https://github.com/bluesky-social/atproto/blob/d5948e8cd33a78a674c4b071fb09d5cf5d5c4807/packages/lex/lex-server/README.md#L522-L527

PR #4689, opened 2026-02-26
Increase type strictness in `@atproto/syntax`

matthieusieben:
This PR extract changes made in https://github.com/bluesky-social/atproto/pull/4408 related to `@atproto/common`, `@atproto/common-web` and  `@atproto/syntax`.

The main goal here is to bring more type strictness into the utilities exposed by these libraries.

### DIFF
--- a/packages/lexicon-resolver/tests/lexicon.test.ts
+++ b/packages/lexicon-resolver/tests/lexicon.test.ts
@@ -157,7 +157,7 @@ describe('Lexicon resolution', () => {
         name: 'LexiconResolutionError',
         message: 'Invalid Lexicon document (example.alice.baddoc)',
         cause: expect.objectContaining({
-          name: 'ValidationError',
+          name: 'LexValidationError',
         }),
       }),
     )


packages/syntax/src/datetime.ts
```diff
@@ -1,61 +1,96 @@
...
   if (input.length > 64) {
-    throw new InvalidDatetimeError('datetime is too long (64 chars max)')
+    return failure('datetime is too long (64 chars max)')
   }
   if (input.endsWith('-00:00')) {
-    throw new InvalidDatetimeError(
-      'datetime can not use "-00:00" for UTC timezone',
-    )
+    return failure('datetime can not use "-00:00" for UTC timezone')
   }
```
matthieusieben, line 266: Perform cheaper text based validation before performing more expensive date parsing

packages/syntax/src/datetime.ts
```diff
@@ -1,61 +1,96 @@
...
-  if (input.startsWith('000')) {
-    throw new InvalidDatetimeError('datetime so close to year zero not allowed')
+  if (!DATETIME_REGEX.test(input)) {
+    return failure("datetime didn't validate via regex")
   }
+
+  // must parse as ISO 8601; this also verifies semantics like month is not 13 or 00
+  const date = new Date(input)
+
+  return validateDatetimeDate(date)
```
matthieusieben: Move validation of valid date properties in separate function that can be re-used from `normalizeDatetime` without needing to stringify->re-parse date values.

packages/syntax/src/datetime.ts
```diff
@@ -1,61 +1,96 @@
...
 declare global {
   interface Date {
-    toISOString(): `${string}-${string}-${string}T${string}:${string}:${string}Z`
+    toISOString(): StandardDatetimeString
   }
 }
 
-/* Validates datetime string against atproto Lexicon 'datetime' format.
- * Syntax is described at: https://atproto.com/specs/lexicon#datetime
+const DATETIME_REGEX =
```
matthieusieben: Note the regex has changed. It no longer allows 61+ minute/second, 32+ day, 13+ month values.

.changeset/all-mirrors-brush.md
```diff
@@ -0,0 +1,5 @@
+---
+'@atproto/syntax': patch
+---
+
+Add type safe accessors to read and write `AtUri`'s `did`, `dollection` and `rkey` properties
```
devinivy:
```suggestion
Add type safe accessors to read and write `AtUri`'s `did`, `collection` and `rkey` properties
```

.changeset/eager-pugs-mix.md
```diff
@@ -0,0 +1,5 @@
+---
+'@atproto/common': patch
+---
+
+Remove un-necessary validation of already properly formatted ISO date string
```
devinivy:
```suggestion
Remove unnecessary validation of already properly formatted ISO date string
```

packages/syntax/src/aturi.ts
```diff
@@ -80,6 +99,18 @@ export class AtUri {
...
+  }
+
+  get rkeySafe(): RecordKeyString {
+    const { rkey } = this
+    if (!rkey) throw new TypeError('No rkey in AT URI')
+    ensureValidRecordKey(rkey)
+    return rkey
+  }
+
+  set rkeySafe(v: RecordKeyString) {
```
devinivy: Interesting.  It's actually not all that safe, since the type `RecordKeyString` isn't able to capture the requirements of an rkey.  If these were tagged types it would be a different story.  Same question applies to the other instances of this e.g. `collectionSafe` even though with NSIDs the story is less bad.  I don't mind the string format types—they do help catch issues and provide some nice self-documentation—but I also don't want to pretend they get us farther than they do.

packages/syntax/src/did.ts
```diff
@@ -65,7 +66,11 @@ export function ensureValidDidRegex<I extends string>(
 }
 
 export function isValidDid<I extends string>(input: I): input is I & DidString {
-  return input.length <= 2048 && DID_REGEX.test(input)
+  return (
+    input.length >= MIN_DID_LENGTH &&
```
devinivy: This looks like an optimization so that you don't land in the `DID_REGEX` when the length is too short, but if the vast majority of the time the input passes this check then it will just be extra work.  It is not clear that this will play out favorably, I would be inclined to leave it out to keep things simple.

packages/syntax/tests/aturi.test.ts
```diff
@@ -522,5 +522,42 @@ describe('AtUri validation', () => {
...
+    const urip = new AtUri('at://foo.com')
+    expect(urip.rkey).toBe('')
+    urip.rkey = 'valid_rkey-123'
+    expect(urip.rkey).toBe('valid_rkey-123')
+    expect(urip.rkeySafe).toBe('valid_rkey-123')
+    urip.rkeySafe = 'other_valid_rkey-456'
+    expect(urip.rkey).toBe('other_valid_rkey-456')
+    expect(urip.rkeySafe).toBe('other_valid_rkey-456')
+    expect(() => (urip.rkey = 'not a valid rkey')).toThrow()
+    urip.rkeySafe = 'not a valid rkey' // Checking is disabled
```
devinivy: This isn't a valid rkey because it contains whitespace, and there's no runtime or compile type error, which sort of highlights the limitations at play with trying to use the string format types for this purpose.

I think this at minimum calls for clearer documentation.  Or arguably we should just call these setters "unsafe" instead of "safe" to indicate that you should use it only if you've already pre-validated and know what you're doing.

packages/syntax/tests/datetime.test.ts
```diff
@@ -9,67 +8,97 @@ import {
...
-    ensureValidDatetime(h)
-    normalizeDatetime(h)
-    normalizeDatetimeAlways(h)
-  }
-  const expectInvalid = (h: string) => {
-    expect(() => ensureValidDatetime(h)).toThrow(InvalidDatetimeError)
-  }
-
-  it('conforms to interop valid datetimes', () => {
-    const lineReader = readline.createInterface({
```
devinivy, line 23: Woah!  I didn't know this was in here, funky.

packages/syntax/src/datetime.ts
devinivy: We should describe what checks from the spec are enforced by this regex.  Previously it was noted to validate the RFC3339 requirement.

I am also personally curious where these specific requirements come from!  For example, RFC3339 permits leap seconds of value `60` so I know it is not from there.

Lastly, I see that it does not validate certain checks of a similar class, e.g. permits the day `00`.  Overall I feel a bit uncertain about which code is responsible for making certain checks, and how those map onto the spec.

packages/syntax/src/datetime.ts
```diff
@@ -1,61 +1,100 @@
...
+ * datetime string that can be used where {@link DatetimeString} is expected.
  */
-export function isValidDatetime<I extends string>(
-  input: I,
-): input is I & DatetimeString {
-  try {
-    ensureValidDatetime(input)
-  } catch (err) {
-    return false
+function validateDatetimeDate(date: Date): Result<Date> {
```
devinivy: I like the idea of this function helper since it adapts the lexicon spec to `Date`.

packages/common-web/src/async.ts
```diff
@@ -43,17 +43,20 @@ export const readFromGenerator = async <T>(
...
 }
 
-export type Deferrable = {
-  resolve: () => void
-  complete: Promise<void>
+export type Deferrable<T = void> = {
+  resolve: (value: T | PromiseLike<T>) => void
+  reject: (reason?: unknown) => void
+  complete: Promise<T>
 }
```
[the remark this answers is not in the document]
matthieusieben: Typescript actually allows calling `function foo(arg: void) {}` with no args.

[Proof](https://www.typescriptlang.org/play/?#code/GYVwdgxgLglg9mABMOcAUBDATgcwFyIBucMAJgJSIDeAvgFB0rrl1A)

packages/syntax/tests/datetime.test.ts
```diff
@@ -106,14 +135,52 @@ describe('normalization', () => {
...
+    }
+  })
+
+  describe('fails on interop (invalid syntax)', () => {
+    for (const dt of interopInvalidSyntax) {
+      test(dt, () => {
+        // @NOTE we can't test the returned value as some will normalize while others won't.
+        expect(() => normalizeDatetimeAlways(dt)).not.toThrow()
+      })
+    }
```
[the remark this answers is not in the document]
matthieusieben: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/bluesky-social/atproto/pull/4689#discussion_r2861241700)

packages/syntax/src/datetime.ts
```diff
@@ -1,61 +1,100 @@
...
+  const fullYear = date.getUTCFullYear()
+  if (Number.isNaN(fullYear)) {
+    return failure('datetime did not parse as ISO 8601')
+  }
+  // Ensure that the ISO string representation does not start with ±YYYYYY
+  if (fullYear < 0) {
+    return failure('datetime normalized to a negative time')
+  }
+  if (fullYear > 9999) {
+    return failure('datetime year is too far in the future')
```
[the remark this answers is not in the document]
matthieusieben: @copilot open a new pull request to apply changes based on https://github.com/bluesky-social/atproto/pull/4689#discussion_r2861241779:

- Add a test for boundary
- DO NOT change the error message

packages/syntax/src/datetime.ts
```diff
@@ -1,61 +1,100 @@
-/** An ISO 8601 formatted datetime string (YYYY-MM-DDTHH:mm:ss.sssZ) */
-export type DatetimeString =
-  `${string}-${string}-${string}T${string}:${string}:${string}${'Z' | `+${string}` | `-${string}`}`
+/**
+ * A strictly formatted UTC datetime string (`YYYY-MM-DDTHH:mm:ss.sssZ` or `±YYYYYY-MM-DDTHH:mm:ss.sssZ`) as returned by {@link Date.toISOString}.
+ * @see {@link https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString}
+ */
+export type StandardDatetimeString =
+  `${string}-${string}-${string}T${string}:${string}:${string}Z`
```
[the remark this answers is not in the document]
matthieusieben: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/bluesky-social/atproto/pull/4689#discussion_r2861241801):

Update the definition of `StandardDatetimeString` to be `${string}-${string}-${string}T${string}:${string}:${string}.${string}Z`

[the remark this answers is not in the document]
matthieusieben: On second thought the type is correct as it is. The comment should be relaxed.

packages/common/src/dates.ts
```diff
@@ -8,7 +6,9 @@ export function toSimplifiedISOSafe(dateStr: string) {
     return new Date(0).toISOString()
   }
   const iso = date.toISOString()
-  if (!isValidISODateString(iso)) {
+  // Date.toISOString() always returns YYYY-MM-DDTHH:mm:ss.sssZ or ±YYYYYY-MM-DDTHH:mm:ss.sssZ
+  // developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString
+  if (iso.startsWith('-') || iso.startsWith('+')) {
```
devinivy, line 15: It seems plausible that there _could_ be invalid dates of the form `YYYY-MM-DDTHH:mm:ss.sssZ`—are we quite sure these never occur?

packages/syntax/src/datetime.ts
```diff
@@ -1,124 +1,303 @@
...
+  assertAtprotoDate(date)
+  return date
+}
+
+export function isAtprotoDate(date: Date): date is AtprotoDate {
+  return parseDate(date).success
+}
+
+declare global {
+  // Overload the global Date constructor to allow creating AtprotoDate objects
```
devinivy: "Overload the global Date constructor" certainly gets the blood pumping!  In my view projects should at least opt into this rather than it occur automatically.

matthieusieben: 1) There is no easy & convenient way to opt-into it with TS
2) We only overload `new Date()`, `new Date(0)` and `new Date(value as ISODatetimeString)`
3) The overload only affects the string value returned by `date.toISOString()` in a way that is compatible with `string`
4) The goal is to strike a balance between correctness (esp. when processing non-validated dates) and being able to use `new Date().toISOString()` (instead of `asAtprotoDte(new Date()).toISOString()`) when generating data that should contain a timestamp.

All that being said, I am open to remove the overload here and have users only rely on `asAtprotoDate()`

packages/common/src/dates.ts
matthieusieben: I don's see how `Date.toISOString()` would return invalid `YYYY-MM-DDTHH:mm:ss.sssZ` values.

In addition to this, the reported goal of this function is to "preserve temporal sort", which would be the case of any `YYYY-MM-DDTHH:mm:ss.sssZ` generated by Javascript.

Do you have any idea of (non `NaN`) `x` value for which `new Date(x).toISOString()` would an invalid `YYYY-MM-DDTHH:mm:ss.sssZ` value ?

packages/syntax/tests/aturi.test.ts
matthieusieben: Alright reworked the whole thing to work better. Created a `unsafelySetRkey` and `unsafelySetCollection` for unsafe setting. using the setter now is always safe and `rkeySafe` / `collectionSafe` only have a getter.

packages/syntax/src/aturi.ts
matthieusieben: Alright reworked the whole thing to work better. Created a `unsafelySetRkey` and `unsafelySetCollection` for unsafe setting. using the setter now is always safe and `rkeySafe` / `collectionSafe` only have a getter.

packages/syntax/src/datetime.ts
matthieusieben: I removed the global overload. We can think of ways to add them later if needed.

PR #4691, opened 2026-02-26
Add Lex SDK support to `@atproto/xrpc-server`

matthieusieben:
This PR extract changes made in https://github.com/bluesky-social/atproto/pull/4408 related to XRPC server.

In particular, it adds the ability to use `xrpc-server` with schema defined through `@atproto/lex` in addition to being usable with `@atproto/lexicon`.

### DIFF
--- a/packages/lex/lex-data/src/lex-error.ts
+++ b/packages/lex/lex-data/src/lex-error.ts
@@ -46,7 +46,7 @@ export type LexErrorData<N extends LexErrorCode = LexErrorCode> = {
  * @typeParam N - The specific error code type
  */
 export class LexError<N extends LexErrorCode = LexErrorCode> extends Error {
-  name = this.constructor.name
+  name = 'LexError'
 
   /**
    * @param error - The error code identifying the type of error, typically used in XRPC error payloads
--- a/packages/lex/lex-schema/src/util/lazy-property.ts
+++ b/packages/lex/lex-schema/src/util/lazy-property.ts
@@ -2,7 +2,7 @@
 export function lazyProperty<
   O extends object,
   const K extends keyof O,
-  const V,
+  const V extends O[K],
 >(obj: O, key: K, value: V): V {
   Object.defineProperty(obj, key, {
     value,
--- a/packages/lex/lex-server/src/lex-router.ts
+++ b/packages/lex/lex-server/src/lex-router.ts
@@ -1034,24 +1034,6 @@ export class LexRouter {
       // @NOTE see ./service-auth.ts for potential common code (did resolver, etc.)
     }
 
-    if (atprotoProxy != null) {
-      const proxyInfo = parseAtprotoProxyHeader(atprotoProxy)
-      if (!proxyInfo) {
-        return Response.json(
-          {
-            error: 'InvalidRequest',
-            message: `Invalid atproto-proxy header value: ${atprotoProxy}`,
-          },
-          { status: 400 },
-        )
-      }
-
-      // @TODO actually implement service proxying logic here. The reason it was
-      // not done already if because we want to perform all the heavy lifting
-      // here, while still allowing the possibility to override the endpoint
-      // resolution, etc.
-    }
-
     return Response.json(
       {
         error: 'MethodNotImplemented',


packages/xrpc-server/src/types.ts
```diff
@@ -233,6 +299,21 @@ export type StreamConfigOrHandler<
...
+    output != null &&
+    'body' in output && // body is non optional (contrary to what type inference may suggest)
+    'encoding' in output &&
+    // Allows using objects that extends HandlerSuccess with a "status" field as
+    // output, as long as the status is < 400, in order to avoid being confused
+    // with ErrorResult objects.
+    (!('status' in output) ||
+      output.status == null ||
+      Number(output.status) < 400)
+  )
```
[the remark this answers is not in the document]
matthieusieben: `output != null` is literally the first check there.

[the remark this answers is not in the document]
matthieusieben: Also the `Output` interface ensures it is an object.

packages/xrpc-server/tests/_util.ts
```diff
@@ -53,3 +62,36 @@ export function basicAuthHeaders(creds: {
       Buffer.from(`${creds.username}:${creds.password}`).toString('base64'),
   }
 }
+
+export async function buildMethodLexicons(
```
devinivy, line 73: I think it would be useful to leave a comment about the difference between `buildMethodLexicons` and `buildAddLexicons` to help explain why both are tested throughout the whole test suite.

packages/xrpc-server/src/util.ts
```diff
@@ -251,12 +440,20 @@ function trimString(str: string): string {
...
   if (!normalized) return false
 
   const allowed = parseDefEncoding(output)
-  return allowed.includes(ENCODING_ANY) || allowed.includes(normalized)
+  if (!allowed.length) return false
+
+  return (
+    allowed.includes(ENCODING_ANY) ||
+    allowed.includes(normalized) ||
+    allowed.includes(`${normalized.split('/', 1)[0]}/*`)
```
devinivy: This is pretty dense, I'd advocate to break this up and/or leave a comment.

packages/xrpc-server/src/server.ts
```diff
@@ -114,13 +130,142 @@ export class Server {
...
+          schema,
+          config as LexMethodConfig<l.Query, A>,
+        )
+      case 'subscription':
+        return this.addSubscriptionSchema(
+          schema,
+          config as LexSubscriptionConfig<l.Subscription, A>,
+        )
+      default:
+        throw new TypeError(`Unsupported schema`)
```
devinivy: Including the namespace and type in this error would probably be useful.

packages/xrpc-server/src/server.ts
```diff
@@ -381,12 +505,24 @@ export class Server {
...
+    cfg: StreamConfig<A, Params>,
   ) {
-    const paramsVerifier = this.createParamsVerifier(nsid, def)
-    const authVerifier = this.createAuthVerifier(cfg)
+    this.addSubscriptionInternal(
+      nsid,
+      this.createLexiconParamsVerifier(nsid, def),
+      this.createAuthVerifier(cfg),
+      // @NOTE outgoing messages are not validated against the lexicon schema
+      // (unlike the handlers for schema based subscriptions)
```
devinivy: It's not clear to me how this differs from a "schema-based subscription".

devinivy: I expect we're talking about the new vs. existing interfaces for adding lexicon methods to the server—if so "schema-based" may just be a little ambiguous.

packages/xrpc-server/src/server.ts
```diff
@@ -395,46 +531,98 @@ export class Server {
...
+  private createLexiconInputVerifier<I extends Input = Input>(
+    nsid: string,
+    def: LexXrpcQuery | LexXrpcProcedure,
+    opts?: RouteOptions,
+  ): InputVerifierInternal<I> {
+    return createLexiconInputVerifier(
+      nsid,
+      def,
+      {
+        blobLimit: opts?.blobLimit ?? this.options.payload?.blobLimit,
```
devinivy, line 585: Yay are the per-route limits new?

packages/xrpc-server/src/util.ts
```diff
@@ -81,49 +113,91 @@ export function decodeQueryParam(
...
+  def: LexXrpcQuery | LexXrpcProcedure | LexXrpcSubscription,
+  lexicons: Lexicons,
+): ParamsVerifierInternal<P> {
+  return (req) => {
+    try {
+      const queryParams = getQueryParams(req)
+      const params = decodeQueryParams(def, queryParams)
+      return lexicons.assertValidXrpcParams(nsid, params) as P
+    } catch (e) {
+      throw new InvalidRequestError(String(e))
```
devinivy, line 166: This stands out to me.  What if `e` is an unexpected error or is itself an xrpc error?  I think maybe we want this behavior only if `e` is a validation error thrown by `assertValidXrpcParams()`?

packages/xrpc-server/src/util.ts
```diff
@@ -81,49 +113,91 @@ export function decodeQueryParam(
...
+  M extends l.Procedure | l.Query | l.Subscription,
+>(ns: l.Main<M>): ParamsVerifierInternal<LexMethodParams<M>> {
+  const schema = l.getMain(ns)
+  return (req) => {
+    try {
+      return schema.parameters.fromURLSearchParams(
+        getSearchParams(req.url) ?? new URLSearchParams(),
+      ) as LexMethodParams<M>
+    } catch (err) {
+      throw new InvalidRequestError(String(err))
```
devinivy: Same as above, this error wrapping stands out to me and I am not sure we want this exact behavior.

packages/xrpc-server/src/util.ts
```diff
@@ -172,54 +248,167 @@ export function createInputVerifier(
...
-      }
+    try {
+      const result = lexicons.assertValidXrpcOutput(nsid, handlerOutput.body)
+      // @TODO Since the output verifier is typically enabled in dev/tests and
+      // disabled in production, we don't want to assign the (altered) output
+      // back to the handlerOutput object, as this would cause different
+      // behaviors between environments. Instead, we should compare the result
+      // with the original output and throw if they differ (indicating that the
+      // output was mutated during validation, e.g. due to default values).
+      result
```
devinivy: What does this line do?

packages/xrpc-server/src/util.ts
```diff
@@ -172,54 +248,167 @@ export function createInputVerifier(
...
-      )
+  }
+}
+
+export function createSchemaOutputVerifier<M extends l.Procedure | l.Query>(
+  ns: l.Main<M>,
+): OutputVerifierInternal<LexMethodOutput<M>> {
+  const outputSchema = l.getMain(ns).output
+  return (handlerOutput) => {
+    if (!outputSchema.matchesEncoding(handlerOutput?.encoding)) {
```
devinivy, line 406: As long as we retain an escape hatch to return other response encodings, e.g. to return html from an xrpc endpoint if specifically asked for via content negotiation.  I think the "passthrough" result covers this case though.

packages/xrpc-server/src/server.ts
matthieusieben: I don't think so:
https://github.com/bluesky-social/atproto/blob/52834aba182da8df3611fd9dff924e6c6a3973a7/packages/xrpc-server/src/server.ts#L277-L288

packages/xrpc-server/src/util.ts
matthieusieben: You are right. I restored the old behavior.

packages/xrpc-server/src/util.ts
matthieusieben: Added an `instanceof` check

packages/xrpc-server/src/util.ts
matthieusieben: prevent the "unused variable" lint error 🤪

matthieusieben: Used an `eslint-disable-next-line` comment instead

packages/xrpc-server/src/util.ts
matthieusieben: the passthrough result does cover this. I'll add a comment

PR #4697, opened 2026-02-27
Update `@atprto/lexicon-resolver` to use Lex SDK

matthieusieben:
This PR extract changes made in https://github.com/bluesky-social/atproto/pull/4408 related to error management in Lex SDK.

This updates `@atprto/lexicon-resolver` to use Lex SDK instead of `@atproto/lex-cli` as codegen.


devinivy (commented):
My understanding is that this package was deprecated and replaced by @atproto/lex-resolver, and no longer used anywhere.  In my view any efforts towards this package should probably go towards porting the tests into @atproto/lex-resolver instead!

PR #4701, opened 2026-03-01
fix: preserve service fragment in pipethrough audience for scope checks

grishaLR:
## overview

if this PR is applied `parseProxyInfo` and `parseProxyHeader` now return the full service ID (`did:web:api.bsky.app#bsky_appview`) instead of the bare DID. This preserves the fragment for fine-grained oauth scoping.

notes: 
- `stripFragment` was added in order to preserve the way JWTs are passed to receiving services (bare DIDs expected)
- `createReport` need to strip the `fragment` since is also passes `parseProxyInfo`'s `aud` to `serviceAuthHeaders` 

## background

this PR came about when attempting to create fine grained scopes in favor of `transition:generic` in a [client](https://protoimsg.app) i'm working on. i noticed 403s coming back for scopes i requested and dug in to find the discrepancy in the `serviceInfo` return in the `pipethrough.ts` file. 

note: this same convention can be found in `computeProxyTo`  and where my inspiration came from 

## how to test

- `pnpm --filter @atproto/pds test -- --testPathPattern proxy-header`
  - **`parses proxy header`** — verifies `parseProxyHeader` returns `did` with fragment
  - **`proxies requests based on header`** — verifies the JWT `aud` is the bare DID (no fragment)

## screengrabs

- Verified locally against a self-hosted PDS: `getMutes`, on  other proxied appview calls return 200 with `include:` scopes (previously 403)

### proto IM PDS (with fix):
getMinutes-pds-cahnge

### Bluesky PDS:
getMutes-bsky

### DIFF
--- a/.changeset/fix-pipethrough-scope-audience.md
+++ b/.changeset/fix-pipethrough-scope-audience.md
@@ -1,0 +1,5 @@
+---
+'@atproto/pds': patch
+---
+
+Fix scope check 403s for OAuth tokens using `include:` permission-set scopes with `inheritAud`. The pipethrough now preserves the service fragment (e.g. `#bsky_appview`) in the audience for scope matching, while stripping it from the JWT audience sent to the receiving service.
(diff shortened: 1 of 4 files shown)


grishaLR:
closing PR in favor of broader proposal around proxying bare DIDs vs w/service-ids and aud refs that the team is working on. 

thanks for the work you all put it <3

PR #4707, opened 2026-03-03
Ozone assignment refactors

gcwill70:
# Changes

* Check that queues and reports exist when assigning
* Require DID when assigning to queues
* Create report unassign endpoint
* Only enforce queue collection match when matching a record report
* add description to queues

### DIFF
--- a/packages/api/src/client/lexicons.ts
+++ b/packages/api/src/client/lexicons.ts
@@ -19019,6 +19019,10 @@ export const schemaDict = {
                 maxLength: 25,
                 description: 'Report reason types (fully qualified NSIDs)',
               },
+              description: {
+                type: 'string',
+                description: 'Optional description of the queue',
+              },
             },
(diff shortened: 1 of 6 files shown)


lexicons/tools/ozone/queue/assignModerator.json
```diff
@@ -9,15 +9,15 @@
         "encoding": "application/json",
         "schema": {
           "type": "object",
-          "required": ["queueId"],
+          "required": ["queueId", "did"],
```
foysalit, line 12: sorry, we may have talked about this but why can't we just derive it from the caller if did is not passed?

gcwill70: Thought it would be nice to make this endpoint more intentional.

Could revert this later if the defaulting behavior adds some desired convenience?

PR #4709, opened 2026-03-03
:sparkles: Add purge event for age assurance data for compliance

foysalit:
For compliance reasons, we need to be able to purge all AA related data upon user request. This PR introduces a purge event for AA data which deletes the user generated AA events from ozone db. The intent is for other microservices to do the same upon receiving this new purge event.

### DIFF
--- a/lexicons/tools/ozone/moderation/emitEvent.json
+++ b/lexicons/tools/ozone/moderation/emitEvent.json
@@ -35,6 +35,7 @@
                 "tools.ozone.moderation.defs#modEventPriorityScore",
                 "tools.ozone.moderation.defs#ageAssuranceEvent",
                 "tools.ozone.moderation.defs#ageAssuranceOverrideEvent",
+                "tools.ozone.moderation.defs#ageAssurancePurgeEvent",
                 "tools.ozone.moderation.defs#revokeAccountCredentialsEvent",
                 "tools.ozone.moderation.defs#scheduleTakedownEvent",
                 "tools.ozone.moderation.defs#cancelScheduledTakedownEvent"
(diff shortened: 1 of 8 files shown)


packages/ozone/src/mod-service/index.ts
```diff
@@ -681,6 +686,15 @@ export class ModerationService {
     return { event: modEvent, subjectStatus }
   }
 
+  async purgeAgeAssuranceEvents(subjectDid: string) {
+    this.db.assertTransaction()
+    await this.db.db
+      .deleteFrom('moderation_event')
+      .where('subjectDid', '=', subjectDid)
+      .where('action', '=', 'tools.ozone.moderation.defs#ageAssuranceEvent')
```
gcwill70, line 694: should we also delete `tools.ozone.moderation.defs#ageAssuranceOverrideEvent`?

packages/ozone/src/mod-service/index.ts
```diff
@@ -665,6 +666,10 @@ export class ModerationService {
       subject.blobCids,
     )
 
+    if (isAgeAssurancePurgeEvent(event)) {
+      await this.purgeAgeAssuranceEvents(subjectInfo.subjectDid)
```
gcwill70, line 670: Should the purge happen after the moderation_event insert in case it fails?

packages/ozone/src/mod-service/index.ts
foysalit: the override event is not user generated and does not contain any user info so i'm inclined to keep those.

packages/ozone/src/mod-service/index.ts
foysalit: I believe that's what we're doing here, the event insert happens above this line.

gcwill70 (commented):
Left some comments to consider

PR #4711, opened 2026-03-04
Restore previous resumeSession behavior

estrattonbailey:
In #4470 we removed two things the app relies upon when calling `resumeSession`:
- if the `accessJwt` is expired, we attempt to refresh the session, which will in turn call `persistSession('update', newSession)`
- even if the session isn't expired and hasn't changed, we call `persistSession('update', session)` with the current session

This PR attemps to restore this handling.

### DIFF
--- a/.changeset/stale-states-taste.md
+++ b/.changeset/stale-states-taste.md
@@ -1,0 +1,5 @@
+---
+'@atproto/api': patch
+---
+
+Restore `resumeSession` behavior lost in #4470, ensuring `persistSession` is called


packages/api/src/atp-agent.ts
```diff
@@ -376,43 +376,70 @@ export class CredentialSession implements SessionManager {
...
+
+        this._updateApiEndpoint(res.data.didDoc)
+        this.persistSession?.('update', session)
+
+        return res
+      }
+    } catch (err) {
+      if (
+        err instanceof XRPCError &&
+        (err.status === 401 ||
```
estrattonbailey, line 434: Haven't confirmed if `getSession` will return 401s as well

PR #4712, opened 2026-03-04
bsky: remove format from img urls by default

devinivy:
We're removing the image format from image URLs by default so that we can make changes in the future without busting the full CDN cache.  A new config is introduced to control the rollout.  Also defaulting to webp in the dev version of the image server instead of jpeg.


packages/bsky/src/image/util.ts
```diff
@@ -35,6 +35,58 @@ export const formatsToMimes = new Map<keyof FormatEnum, ImageMime>([
...
 ])
 
 export type SharpInfo = OutputInfo & { format: keyof FormatEnum }
+
+/**
+ * Values 0 through 16.
+ * 0: never include format in image URL.
+ * 16: always include format in image URL.
+ */
+export type IncludeFormatLevel =
```
rafaeleyng, line 48: Feel free to ignore, but could comment/TODO what is temporary so if we forget to delete it right after, the intent it is clearer for the future.

devinivy: Good call out, I'll make sure to track this and own the cleanup.

PR #4713, opened 2026-03-04
Make sure to always trigger `refreshSession` when resuming sessions

matthieusieben:
This change attempts to fix a regression introduced by https://github.com/bluesky-social/atproto/pull/4470

It takes a different approach than #4711 by always calling `refreshSession()`, which should ensure that `persistSession('update', session)` gets called as part of the resume process.

This also fixes a potential issue where a long lasting refresh operation (e.g. due to network timeouts) could override the concurrency guard (`refreshSessionPromise`) set by another operation.

### DIFF
--- a/packages/api/src/atp-agent.ts
+++ b/packages/api/src/atp-agent.ts
@@ -159,7 +159,7 @@ export class AtpAgent extends Agent {
 export class CredentialSession implements SessionManager {
   public pdsUrl?: URL // The PDS URL, driven by the did doc
   public session?: AtpSessionData
-  public refreshSessionPromise: Promise<void> | undefined
+  public refreshSessionPromise?: Promise<ComAtprotoServerRefreshSession.Response>
 
   /**
    * Private {@link ComAtprotoServerNS} used to perform session management API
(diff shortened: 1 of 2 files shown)


packages/api/src/atp-agent.ts
```diff
@@ -372,66 +372,52 @@ export class CredentialSession implements SessionManager {
...
-
-      // Another concurrent operation may have replaced the session while we
-      // were waiting for the refresh to complete.
-      if (session.did !== this.session?.did) {
-        throw new Error('DID mismatch on resumeSession')
-      }
-
-      return this.server.getSession(undefined, {
-        headers: { authorization: `Bearer ${this.session.accessJwt}` },
-      })
```
matthieusieben, line 395: We replace the old behavior that consisted of only performing a `getSession` if the current session is the same as the one being restored.

Now, `resumeSession` will always trigger a `refreshSession`, ensuring `persistSession` is called.

packages/api/src/atp-agent.ts
```diff
@@ -372,66 +372,52 @@ export class CredentialSession implements SessionManager {
...
-      .then(
-        (): void => {},
-        (): void => {},
-      )
-      .finally(() => {
+    if (this.refreshSessionPromise) return this.refreshSessionPromise
+
+    const promise = this._refreshSessionInner().finally(() => {
+      if (this.refreshSessionPromise === promise) {
         this.refreshSessionPromise = undefined
```
matthieusieben, line 414: This line could cause the `this.refreshSessionPromise` to be cleared when the `this.refreshSessionPromise` was actually not "owned" by this execution path:

1) call `resumeSession(someSession)` => this call hangs for netwrok reasons
2) call `login()` => this call will clear the pending `this.refreshSessionPromise` and (potentially) create a new one
3) `resumeSession()` from step 1) resolves before the new `this.refreshSessionPromise` resolves, causing it to be cleared.
4) calling `this.refreshSession()` at this moment could cause a concurrent refresh (since the guard was cleared).

This is fixed by ensuring that the promise if the right one before clearing ` this.refreshSessionPromise`

estrattonbailey: Makes sense 👍

PR #4721, opened 2026-03-05
use feature gate to remove image format from URL

rafaeleyng:
Build on top of https://github.com/bluesky-social/atproto/pull/4712, doing the rollout using a feature gate instead of static config.

### DIFF
--- a/packages/bsky/tests/_util.ts
+++ b/packages/bsky/tests/_util.ts
@@ -80,14 +80,16 @@ export const forSnapshot = (obj: unknown) => {
     if (str.match(/^\d+__bafy/)) {
       return constantKeysetCursor
     }
-    if (str.match(/\/img\/[^/]+\/.+\/did:plc:[^/]+\/[^/]+(?:@[\w]+)?$/)) {
-      // Match image urls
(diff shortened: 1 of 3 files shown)


packages/bsky/src/image/uri.ts
```diff
@@ -22,34 +20,47 @@ export class ImageUriBuilder {
...
     const options = presets[id]
     if (!options) {
       throw new Error(`Unrecognized requested common uri type: ${id}`)
     }
+
+    // TODO: Remove after image migration. It is not ideal to check feature gates outside of handlers with the current setup..
+    const map = featureGatesClient?.checkGates(
+      ['image:remove_format_from_url'],
+      {
+        viewer: did,
```
rafaeleyng, line 42: This is a workaround; the `did` here is not really the viewer, but the owner of the image being viewed.

Nonetheless, it is a useful way of bucketing the images by their owner's DID.

rafaeleyng: The tests are failing, and can possibly be related to this. I'm trying to investigate.

PR #4722, opened 2026-03-05
Ozone bulk events

gcwill70:
# Changes

* Add `tools.ozone.moderation.emitEvents`
* Only support a subset of events
* Create tests for each supported event and failure cases

# How to test

`pnpm --filter ozone test -- tests/moderation-events.test.ts`

### DIFF
--- a/lexicons/tools/ozone/moderation/emitEvents.json
+++ b/lexicons/tools/ozone/moderation/emitEvents.json
@@ -4,7 +4,7 @@
   "defs": {
     "main": {
       "type": "procedure",
-      "description": "Take a moderation action on multiple subjects at once. Each subject is processed independently; partial failures are reported rather than causing a full rollback. This batch endpoint only supports a subset of moderation events.",
+      "description": "Take a moderation action on multiple subjects at once. Failures for a individual subjects are listed in failedEvents. Only a subset of moderation events is supported.",
       "input": {
         "encoding": "application/json",
         "schema": {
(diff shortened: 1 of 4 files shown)


packages/ozone/tests/moderation-events.test.ts
```diff
@@ -646,4 +649,500 @@ describe('moderation-events', () => {
...
+      })
+
+      expect(result.events).toHaveLength(0)
+      expect(result.failedEvents).toHaveLength(2)
+      expect(result.failedEvents[0].error).toContain(
+        'Email can only be sent to a repo subject',
+      )
+    })
+
+    it.skip('bulk revokeAccountCredentials events on repo subjects', async () => {
```
gcwill70, line 1070: @foysalit Could not get this test to pass. Wondering if this service isn't set up in the test netowork?

packages/ozone/tests/moderation-events.test.ts
```diff
@@ -646,4 +649,484 @@ describe('moderation-events', () => {
...
+        },
+        subjects: subjects.map((s) => ({ subject: s })),
+        createdBy: 'did:example:admin',
+      })
+
+      expect(result.events).toHaveLength(2)
+      expect(result.failedEvents).toHaveLength(0)
+    })
+
+    it('works with mixed subject types', async () => {
```
gcwill70, line 775: @foysalit Should we only allow one subject type here? Might reduce complexity with error handling (e.g. request tries to takedown an account and one of their records at the same time)

PR #4727, opened 2026-03-06
:recycle: Refactor report view lexicons and introduce single report query

foysalit:
This PR moves `tools.ozone.moderation.queryReports` to `tools.ozone.report.queryReports` and introduces a new `tools.ozone.moderation.getReport` query endpoint to fetch single report by id.
We also discovered that we need to hydrate report assignment in report view so that has been added too.

### DIFF
--- a/packages/ozone/src/daemon/queue-router.ts
+++ b/packages/ozone/src/daemon/queue-router.ts
@@ -155,7 +155,11 @@ function findMatchingQueue(
   for (const queue of queues) {
     const subjectTypeMatch = queue.subjectTypes.includes(subjectType)
     const collectionMatch =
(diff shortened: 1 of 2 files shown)


lexicons/tools/ozone/report/defs.json
```diff
@@ -219,6 +219,131 @@
       "type": "token",
       "description": "Other dangerous content"
     },
+    "reportAssignment": {
```
gcwill70, line 222: Perhaps we can combine this with assignmentView later so that only one view is needed? We would probably want to make profile optional in a combined view though so that the assignment service doesnt have to enrich profiles for every assignment

PR #4728, opened 2026-03-06
pds & lexicon: explicit error when record exists during creation

devinivy:
When a record already exists during record creation (i.e. `com.atproto.repo.createRecord`) the PDS currently serves a 500.  With this change the PDS will serve an XRPC error newly specified by the `com.atproto.repo.createRecord` and `com.atproto.repo.applyWrites` lexicons.


dholms (approved):
nice looks good 👍

PR #4730, opened 2026-03-06
Disable all metrics from GB

estrattonbailey:


rafaeleyng:
Won't be needed, it wasn't the cause of the issue. Closing for now.

PR #4737, opened 2026-03-10
:sparkles: Enrich moderator team member for assignments

foysalit:
Instead of attaching profile views for assigned moderators, we are now returning `tools.ozone.team.defs#member` object which contains the profile entry inside.
we are also introducing the concept of a `temporary assignment/view` record for reports and a `permanent` assignment.

### DIFF
--- a/packages/ozone/src/assignment/index.ts
+++ b/packages/ozone/src/assignment/index.ts
@@ -363,6 +363,7 @@ export class AssignmentService {
           return dbTxn.db
             .updateTable('moderator_assignment')
             .set({
+              did,
               endAt: null,
               queueId: queueId ?? activeExisting.queueId ?? null,
             })


packages/ozone/src/assignment/index.ts
```diff
@@ -291,11 +326,75 @@ export class AssignmentService {
...
+          .where('reportId', '=', reportId)
+          .where('endAt', '>', now.toISOString())
+          .executeTakeFirst()
+
+        if (activeExisting) {
+          return dbTxn.db
+            .updateTable('moderator_assignment')
+            .set({
+              endAt: null,
+              queueId: queueId ?? activeExisting.queueId ?? null,
```
gcwill70, line 368: For this upgrade logic, should we assign the `did` here as well? This means a permanent assignment to a `did` will "takeover" an existing temporary assignment

packages/ozone/src/api/report/getReport.ts
```diff
@@ -22,12 +22,14 @@ export default function (server: Server, ctx: AppContext) {
       }
 
       const queueService = ctx.queueService(db)
+      const teamService = ctx.teamService(db)
       const hydrated = await hydrateReportInfo(
         [report],
         modService.views,
         (dids) => getPdsAccountInfos(ctx, dids),
-        (reportIds) => getActiveReportAssignments(db, reportIds),
+        (reportIds) => getPermanentReportAssignments(db, reportIds),
```
gcwill70, line 30: Should we show all assignments when hydrating this view? Or is it intended that another endpoint (e.g. `tools.ozone.report.getAssignments`) would have to be used to get temporary assignments?

foysalit: yeah the intention is to easily identify reports that are assigned already, for "views" we don't need that in this response.

packages/ozone/src/assignment/index.ts
foysalit: oh right yeah that sounds correct!

PR #4738, opened 2026-03-10
Ozone report rerouting

gcwill70:
# Changes
* Move queue-router logic into QueueService to be reusable
* Create `tools.ozone.queue.routeReports` endpoint for batch re-routing of reports

### DIFF
--- a/packages/ozone/src/mod-service/report.ts
+++ b/packages/ozone/src/mod-service/report.ts
@@ -34,15 +34,18 @@ export type QueryReportsResult = {
   reports: ReportWithEvent[]
   cursor: string | undefined
 }
+function reportQuery(db: Database) {
+  return db.db
+    .selectFrom('report as r')
+    .innerJoin('moderation_event as me', 'me.id', 'r.eventId')
+    .where('me.action', '=', 'tools.ozone.moderation.defs#modEventReport')
+}
 
 export async function queryReports(
   db: Database,
   params: QueryParams,
 ): Promise<QueryReportsResult> {
-  let builder = db.db
-    .selectFrom('report as r')
-    .innerJoin('moderation_event as me', 'me.id', 'r.eventId')
-    .where('me.action', '=', 'tools.ozone.moderation.defs#modEventReport')
+  let builder = reportQuery(db)
 
   if (params.queueId !== undefined) {
     builder = builder.where('r.queueId', '=', params.queueId)
@@ -170,10 +173,7 @@ export async function getReportById(
   db: Database,
   id: number,
 ): Promise<ReportWithEvent | undefined> {
-  return db.db
-    .selectFrom('report as r')
-    .innerJoin('moderation_event as me', 'me.id', 'r.eventId')
-    .where('me.action', '=', 'tools.ozone.moderation.defs#modEventReport')
+  return reportQuery(db)
     .where('r.id', '=', id)
     .selectAll('r')
     .select([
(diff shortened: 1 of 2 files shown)


lexicons/tools/ozone/queue/routeReports.json
```diff
@@ -0,0 +1,50 @@
+{
+  "lexicon": 1,
+  "id": "tools.ozone.queue.routeReports",
+  "defs": {
+    "main": {
+      "type": "procedure",
+      "description": "Re-route reports within an ID range to matching queues based on current queue configuration.",
```
foysalit: nitpick but since we're calling the endpoint routeReports, feels a bit odd describing it as Re-route reports.

packages/ozone/src/api/queue/routeReports.ts
```diff
@@ -0,0 +1,44 @@
...
+const MAX_REPORTS = 5000
+
+export default function (server: Server, ctx: AppContext) {
+  server.tools.ozone.queue.routeReports({
+    auth: ctx.authVerifier.modOrAdminToken,
+    handler: async ({ input, auth }) => {
+      const access = auth.credentials
+      const { startReportId, endReportId } = input.body
+
+      if (!access.isAdmin) {
```
foysalit: i think we can bring this down to moderator since it's mostly a harmless action.

packages/ozone/src/api/report/getLatestReport.ts
```diff
@@ -0,0 +1,46 @@
...
+  server.tools.ozone.report.getLatestReport({
+    auth: ctx.authVerifier.modOrAdminToken,
+    handler: async ({ auth, req }) => {
+      const db = ctx.db
+      const modService = ctx.modService(db)
+      const labelers = ctx.reqLabelers(req)
+
+      const report = await getLatestReport(db)
+      if (!report) {
+        throw new InvalidRequestError('No reports found')
```
foysalit: should it be `NotFoundError` instead?

packages/ozone/src/daemon/queue-router.ts
```diff
@@ -57,116 +54,30 @@ export class QueueRouter {
...
-      .select(['r.id', 'me.subjectUri', 'me.subjectMessageId', 'me.meta'])
-      .orderBy('r.id', 'asc')
-      .limit(BATCH_SIZE)
-
-    if (lastId !== null) {
-      query = query.where('r.id', '>', lastId)
-    }
-
-    const reports = await query.execute()
+    const result = await queueService.assignReportBatch({
```
foysalit, line 59: so satisfying!

packages/ozone/src/mod-service/report.ts
```diff
@@ -187,6 +187,27 @@ export async function getReportById(
     .executeTakeFirst()
 }
 
+export async function getLatestReport(
+  db: Database,
+): Promise<ReportWithEvent | undefined> {
+  return db.db
+    .selectFrom('report as r')
```
foysalit: this part of the querybuilder is being repeated a couple times i believe? wonder if we should think about refactoring it?

lexicons/tools/ozone/queue/routeReports.json
gcwill70: Good point. I agree

packages/ozone/src/daemon/queue-router.ts
gcwill70: Yeah I really wanted to re-use the same service logic in the daemon and routeReports 😄

packages/ozone/src/mod-service/report.ts
gcwill70: Created a default query in that service

packages/ozone/src/api/report/getLatestReport.ts
gcwill70: Made a custom error for this

PR #4741, opened 2026-03-11
Request (and parse) CBOR responses when method expects structured data

matthieusieben:
This is a PoC that allows allows CBOR to be used as response for methods that expect structured data.

### DIFF
--- a/packages/lex/lex-client/src/util.ts
+++ b/packages/lex/lex-client/src/util.ts
@@ -8,7 +8,7 @@ import {
 
 export const CONTENT_TYPE_BINARY = 'application/octet-stream'
 export const CONTENT_TYPE_JSON = 'application/json'
-export const CONTENT_TYPE_CBOR = 'application/cbor'
+export const CONTENT_TYPE_CBOR = 'application/dag+cbor'
 
 /**
  * The body type of an XRPC response, inferred from the method's output schema.


packages/lex/lex-client/src/xrpc.ts
```diff
@@ -213,7 +215,10 @@ function xrpcRequestInit<T extends Procedure | Query>(
   const headers = buildAtprotoHeaders(options)
 
   // Tell the server what type of response we're expecting
-  if (schema.output.encoding) {
+  if (schema.output.encoding === CONTENT_TYPE_JSON) {
+    // If the schema describes structured data, tell the server we prefer cbor
+    headers.set('accept', `${CONTENT_TYPE_CBOR},${CONTENT_TYPE_JSON};q=0.9`)
```
devinivy, line 220: I know this is just a PoC, but 2c here— I may favor a canonical client to only negotiate cbor when the caller opts in.  We advertise the protocol is "just json", and if folks open their network tab only to find binary responses it may be a bit confusing.

Separately— I would also be really interested to see cbor vs. json benchmarks, json is pretty damn efficient in javascript environments and both cbor and json should take compression quite nicely for large payloads.

packages/lex/lex-client/src/util.ts
```diff
@@ -6,6 +6,10 @@ import {
   Query,
 } from '@atproto/lex-schema'
 
+export const CONTENT_TYPE_BINARY = 'application/octet-stream'
+export const CONTENT_TYPE_JSON = 'application/json'
+export const CONTENT_TYPE_CBOR = 'application/cbor'
```
devinivy: If the client receives CBOR that isn't DAG-CBOR, I expect it will be unhappy.

PR #4742, opened 2026-03-12
Efficient deterministic CAR serialization

DavidBuchanan314:
Note: Depends on #4655 (PDS Actor Store Migrations)

Before merging this, a version of the PDS should be deployed that refuses to perform repo updates if the actor store schema is "from the future" (to avoid old containers from naively corrupting repos midway through a multi-container rolling deploy).

### How it works

A naive implementation of deterministic `getRepo` would walk the MST and emit blocks as it visits them. In practice this is pretty slow, because of all the random accesses (requiring many small db queries) and MST node parsing.

This is avoided by maintaining a new actor store db table, `preorder_map`, which maps "(lpath, depth)" tuples to CIDs, for all blocks in the repo.

For MST nodes:

- "lpath" is the record path ("nsid/rkey") immediately preceding this node (for the root node it is empty string)
- "depth" is `129-(mst key height)`  (where 129 is the hypothetical max height of an mst node plus 1)

For records:

- "lpath" is just the record’s path
- "depth" is fixed at 0 (so you can tell if an index entry is for a record based on this)

The trick here is that as long as `preorder_map` is kept up to date, a linear scan of the table produces the required block order for sync1.1 deterministic CAR files.


packages/pds/src/actor-store/repo/sql-repo-reader.ts
```diff
@@ -81,7 +81,48 @@ export class SqlRepoReader extends ReadableBlockstore {
     if (!root) {
       throw new RepoRootNotFoundError()
     }
```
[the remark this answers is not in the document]
DavidBuchanan314: Copilot is right, but this is a pre-existing issue and should probably be addressed separately

packages/pds/src/actor-store/repo/sql-repo-transactor.ts
```diff
@@ -74,6 +74,32 @@ export class SqlRepoTransactor extends SqlRepoReader implements RepoStorage {
...
+      }
+    }
+    for (const batch of chunkArray(deletes, 50)) {
+      for (const del of batch) {
+        await this.db.db
+          .deleteFrom('preorder_map')
+          .where('lpath', '=', del.lpath)
+          .where('depth', '=', del.depth)
+          .execute()
+      }
```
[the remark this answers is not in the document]
DavidBuchanan314: idk if copilot's suggestion here is any better in practice, on sqlite

packages/pds/src/actor-store/repo/sql-repo-reader.ts
```diff
@@ -81,7 +81,48 @@ export class SqlRepoReader extends ReadableBlockstore {
...
-    return writeCarStream(root, this.iterateCarBlocks(since))
+    if (since) {
+      return writeCarStream(root, this.iterateCarBlocks(since))
+    }
+    return writeCarStream(root, this.iterateCarBlocksPreorder(root))
+  }
```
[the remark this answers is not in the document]
DavidBuchanan314: I think this is the behaviour we want, especially for the golang clients out there...

packages/pds/src/scripts/sequencer-recovery/recoverer.ts
```diff
@@ -162,6 +162,7 @@ const processRepoCreation = async (
     newBlocks: blocks,
     relevantBlocks: new BlockMap(),
     removedCids: new CidSet(),
+    preorderOps: [],
```
DavidBuchanan314, line 171: Is this valid? I don't really grok what recoverer is actually doing here.

packages/repo/src/mst/diff.ts
```diff
@@ -1,37 +1,85 @@
...
+  }
+}
+
+export type MstDiffOpts = {
+  trackPreorder?: boolean
+}
+
+export const nullDiff = async (
+  tree: MST,
+  trackPreorder: boolean = false,
```
DavidBuchanan314: I made this optional and default-false for back-compat, and to avoid the overhead when it isn't needed. But I *think* we pretty much always do use/need it, within this codebase. Might be worth considering making it mandatory to get rid of all the optionality.

DavidBuchanan314: Oops I meant to comment on the `mstDiff` method, but same idea

packages/pds/src/actor-store/repo/sql-repo-transactor.ts
```diff
@@ -74,6 +74,35 @@ export class SqlRepoTransactor extends SqlRepoReader implements RepoStorage {
     await this.updateRoot(commit.cid, commit.rev, isCreate)
     await this.putMany(commit.newBlocks, commit.rev)
     await this.deleteMany(commit.removedCids.toList())
+    if (!commit.preorderOps) {
```
matthieusieben: I would throw earlier in the function to avoid unnecessary work

packages/repo/src/mst/diff.ts
```diff
@@ -65,18 +117,22 @@ export const mstDiff = async (
     if (leftWalker.layer() > rightWalker.layer()) {
       if (left.isLeaf()) {
         await diff.nodeAdd(right)
+        await diff.preorderInsert(right, rightWalker.lastLeafKey)
```
matthieusieben: Looks like most of the time we do both preorderAdd and nodeInsert. It might be worth using a single method that performs both under the hood.

Another pattern that might also work would have two classes: the current one (without tracking) and a sub-class that adds the tracking.this way we can use them both interchangeably (ie. The usage is unchanged), and we dont perform un-nesaccary operations when tracking is actually disabled.

If that method could avoid a bunch of await, that'd probably have a big perf impact (there are *lots* of await in here)

DavidBuchanan314: I should probably set up a benchmark. There's a lot of `await node.getLayer()` and that could probably be made synchronous if nodes knew their own height contextually (they almost do already but there are some edge cases...)

packages/pds/src/api/com/atproto/repo/importRepo.ts
```diff
@@ -82,6 +82,12 @@ export default function (server: Server, ctx: AppContext) {
               )
             }
 
+            if (typeof parsedRecord['$type'] !== 'string') {
+              throw new InvalidRequestError(
```
matthieusieben, line 87: Should we also check for empty string? Also, maybe we'd could just skip those records to avoid a car not being usable anywhere

matthieusieben: Or maybe even check for valid nsids?

packages/pds/src/scripts/sequencer-recovery/recoverer.ts
dholms: This is playing back the sequencer db over the actor stores. This gets played for the frist event in a repos history. Looks like `preorderOps` should be derived from the `writes` parameter. If there's a simple function to translate between the two, could use that.

Butttt, I don't think I see any code path where the first event would have any writes. So we can probably leave this as an empty array. But then I would check the length of writes & throw an error if > 0

packages/repo/src/mst/diff.ts
dholms: A lot of the sync/repo stuff in the atproto repo is more general, while indigo is more specifically tailored for sync1.1.

I think we may want to take a broader pass at revamping some of the repo stuff to be more tailored to how the repo is used. but for now I like leaving this as is.

dholms: could we use the `MstDiffOpts` here as well?

packages/pds/src/actor-store/repo/sql-repo-reader.ts
```diff
@@ -81,7 +81,52 @@ export class SqlRepoReader extends ReadableBlockstore {
...
+        `Missing commit block bytes for root CID ${root.toString()}`,
+      )
+    }
+    yield { cid: root, bytes: commitBytes }
+
+    // Yield remaining blocks in preorder traversal order
+    const { ref } = this.db.db.dynamic
+    let lastLpath = ''
+    let lastDepth = 0
+    for (;;) {
```
dholms: I've never seen this in typescript lol I honestly didn't even know it was valid syntax 😆 Can we just use a while loop?

packages/pds/src/actor-store/db/migrations/002-preorder-map.ts
```diff
@@ -0,0 +1,140 @@
...
+  for (const entry of entries) {
+    if (entry.isLeaf()) {
+      await inserter.add({
+        lpath: entry.key,
+        depth: 0,
+        cid: entry.value.toString(),
+      })
+      currentLpath = entry.key
+    } else {
+      const childLayer = (await entry.getLayer()) ?? layer - 1
```
dholms: Why compute childLayer by calling getLayer(), subtracting one, and passing it in as a param. Instead of just calculating the layer in the next function call?

packages/repo/src/mst/diff.ts
dholms: Actually after reading more, I think I changed my mind. Tracking this doesn't affect anything else in the diffing code. And if we're basically always doing it, I think we just track by default. It should help keep us from shooting ourselves in the foot

packages/repo/src/car.ts
```diff
@@ -23,14 +25,60 @@ export async function* writeCarStream(
       roots: root ? [root] : [],
     }),
   )
-  yield new Uint8Array(varint.encode(header.byteLength))
-  yield header
+
+  let buf = new Uint8Array(CAR_BUFFER_SIZE)
```
dholms: Okay I see what's going on here. Took me a second to get it. Would you be up for either adding a comment explaining this or wrapping the buf/pos/flush/write stuff in a helper class?

packages/repo/src/mst/diff.ts
```diff
@@ -1,37 +1,85 @@
...
-  const diff = new DataDiff()
+  const diff = new DataDiff({ trackPreorder })
 
   const leftWalker = new MstWalker(prev)
   const rightWalker = new MstWalker(curr)
   while (!leftWalker.status.done || !rightWalker.status.done) {
     // if one walker is finished, continue walking the other & logging all nodes
     if (leftWalker.status.done && !rightWalker.status.done) {
       await diff.nodeAdd(rightWalker.status.curr)
+      await diff.preorderInsert(
```
dholms: I wonder if for each of these, it would just maek more sense to bundle it with the other method being called. For isntance `nodeAdd` on `DataDiff` always does the logic for `preorderInsert`. `leafUpdate` does a `preorderDelete` + a `preorderInsert`.

packages/repo/src/mst/diff.ts
DavidBuchanan314: Aside from efficiency, one concretely different behaviour between the two versions is that `emitLpathFixups` can end up visiting nodes that otherwise would not have been visited. I wonder if this could break some assumptions somewhere?

packages/pds/src/actor-store/repo/sql-repo-reader.ts
DavidBuchanan314: C programmers used to argue that `for (;;)` was faster but it hasn't been true for a long time, I'll get rid of it heh

packages/repo/src/car.ts
```diff
@@ -23,14 +25,60 @@ export async function* writeCarStream(
...
+        return combined
+      }
+      buf.set(data, 0)
+      pos = data.byteLength
+      return out
+    }
+    buf.set(data, pos)
+    pos += data.byteLength
+    return null
+  }
```
matthieusieben: This looks like it's copying data quite a lot. What is the goal of using internal buffering here? Do we really need to aggregate into 16k chunks ? If so, I would keep the current implementation here and implement chunking in a wrapper:

```ts
async function* chunkStream(it: AsyncIterator<Uint8Array>, chunkSize = 0x4000): AsyncGenerator<Uint8Array, void, unknown> {
  // TODO chunking here
}

// use as

const chunked = chunkStream(writeCarStream(root, blobks))
```

DavidBuchanan314: My rationale for it is here https://github.com/bluesky-social/atproto/pull/4742#issuecomment-4056497479 - but that's a good suggestion, I'll do that

packages/repo/src/mst/walker.ts
```diff
@@ -28,28 +29,25 @@ export class MstWalker {
...
   layer(): number {
     if (this.status.done) {
       throw new Error('Walk is done')
     }
-    if (this.status.walking) {
-      return this.status.walking.layer ?? 0
-    }
-    // if curr is the root of the tree, add 1
-    if (this.status.curr.isTree()) {
-      return (this.status.curr.layer ?? 0) + 1
```
DavidBuchanan314, line 39: There was a subtle bug here. `curr.layer` may be null (the value is lazy-loaded), in which case it returns the wrong result. Getting the true value would require `getLayer()` which is async. So my fix is to pass the root layer in during construction and track it explicitly while walking.

DavidBuchanan314: (until now this wasn't an issue, iiuc, because .layer() was always called at a point where it was not-null)

DavidBuchanan314: @dholms could you help me understand why the `+ 1` was there in the first place? I changed the behaviour in beccd51c46baf0a6aa82e6d47cc96ddae5ee6db4 to effectively not do that because it simplifies things, and it doesn't seem to have broken anything, but I want to make sure I'm not missing something important

packages/repo/src/repo.ts
```diff
@@ -204,35 +203,6 @@ export class Repo extends ReadableRepo {
...
-      newBlocks,
-      relevantBlocks: newBlocks,
-      removedCids: new CidSet([this.cid]),
-    }
-  }
-
-  async resignCommit(rev: string, keypair: crypto.Keypair) {
-    const formatted = await this.formatResignCommit(rev, keypair)
-    return this.applyCommit(formatted)
-  }
```
DavidBuchanan314, line 233: This is dead code and I'm pretty sure we no longer have any need for it?

packages/pds/src/api/com/atproto/sync/getRepo.ts
```diff
@@ -25,8 +25,13 @@ export default function (server: Server, ctx: AppContext) {
 
       const carStream = await getCarStream(ctx, did, since)
 
+      // `atproto-car-block-order` acts as a hint for consumers - in the future it can be implicit/assumed
       return {
         encoding: 'application/vnd.ipld.car' as const,
+        headers:
+          since === undefined
+            ? { 'atproto-car-block-order': 'preorder-deterministic' }
```
DavidBuchanan314: I don't thiiiiink we need to bikeshed this (I'm thinking it's more of a hint than something that needs to be part of the spec proper)

DavidBuchanan314:
There is a potential read-consistency issue here, both in the original getRepo impl and the new deterministic one - if the repo is updated midway through iteration, an inconsistent snapshot will be returned.

We're not making the issue any worse than it was before, so I don't think we need to address it immediately, but we should probably make a decision between:

- Hold open a transaction for the whole `getRepo` - which would allow slow clients to grow the WAL arbitrarily and hold connections open for a long time. (could be mitigated with timeouts, but overlapping slow-reads on concurrent connections could still let the WAL grow arbitrarily - but maybe it's fine? "arbitrarily" is still bounded by the number of writes, which are rate limited)

- Detect updates (i.e. if mst root changed) and error out partway through. This would make very rapidly-updated repos hard to read.

- Leave it as is? (let clients do their own consistency check, if they care to)

packages/repo/src/mst/walker.ts
dholms: I truly have no idea 😅 
The logic also doesn't match up with the comment. If this doesn't break any of the MST diff tests then I think it's probably Okay ™️

packages/pds/src/api/com/atproto/sync/getRepo.ts
dholms: I'm also inclined to avoid bikeshedding. Buttt I would just say "preorder" instead of "preorder-deterministic" (I know @bnewbold is not a fan of the "deterministic" verbiage & I think I tend to agree that it's the wrong thing to highlight

.changeset/wicked-corners-cough.md
```diff
@@ -0,0 +1,5 @@
+---
+'@atproto/repo': patch
+---
+
+Optionally track block order metadata during MST diffs
```
DavidBuchanan314, line 5: Oop I need to update this, it's no longer optional

dholms (approved):
This looks good to me!

On your last comment around read-consistency: I'm inclined to just land this for now without addressing it (since we're at least not making the problem any _worse_

I'm not sure what the best fix will be, I don't think we can just error out because high-churn repos will also likely be quite large and will be the most likely to error. So we'll probably end up needing to do something smart on the read side

PR #4744, opened 2026-03-13
Improve correctness of DID Identifier string syntax

matthieusieben:
This change attempts to make DID validation aligned with, [AT protocol - DID Identifier Syntax](https://atproto.com/specs/did#did-identifier-syntax), [DID-core](https://www.w3.org/TR/did-core/#did-syntax) and [rfc3986 - section 2.1](https://www.rfc-editor.org/rfc/rfc3986#section-2.1) by ensuring that every `%` is followed by 2 uppercase hexadecimal digits, and by allowing digits in the did method.

In addition to these changes, `@atproto/syntax` exposes new type casting and type assertion utilities (similar to those for `DatetimeString`) that should make the use of `@atproto/lex` more convenient.

> [!NOTE]
> 
> This PR also includes updated interop files from https://github.com/bluesky-social/atproto-interop-tests/pull/12 . That PR (and the related spec changes) still need to be adopted before this PR can be merged

This PR also improves the output of tests allowing to see which DID failed to pass

Capture d’écran 2026-03-13 à 15 07 36


packages/syntax/tests/did.test.ts
```diff
@@ -1,100 +1,122 @@
...
+  'did:m123:val',
 
-    expectValid('did:m:v')
-    expectValid('did:method::::val')
-    expectValid('did:method:-')
-    expectValid('did:method:-:_:.:%ab')
-    expectValid('did:method:.')
-    expectValid('did:method:_')
-    expectValid('did:method::.')
+  'did:method:' + 'v'.repeat(2048),
```
matthieusieben: this was lowered from 8500

interop-test-files/syntax/did_syntax_invalid.txt
```diff
@@ -8,9 +8,18 @@ did:methodval)
 did.method.val
 did:method:val:
 did:method:val%
+did:method:val%bb
+did:method:-:_:.:%ab
```
matthieusieben, line 15: Lowecase HEXDIG are not valid

packages/syntax/tests/did.test.ts
```diff
@@ -1,100 +1,56 @@
...
-    expectInvalid('did:method:' + 'v'.repeat(8500))
+const ROOT_DIR = `${__dirname}/../../../`
 
-    expectValid('did:m:v')
-    expectValid('did:method::::val')
-    expectValid('did:method:-')
-    expectValid('did:method:-:_:.:%ab')
-    expectValid('did:method:.')
-    expectValid('did:method:_')
-    expectValid('did:method::.')
```
matthieusieben, line 50: All these are redundant with interop files

packages/syntax/tests/datetime.test.ts
```diff
@@ -196,10 +196,3 @@ describe(normalizeDatetimeAlways, () => {
     }
   })
 })
-
-function readLines(filePath: string): string[] {
```
matthieusieben, line 250: moved to shared location (for did tests)

packages/syntax/tests/interop-files/did_syntax_invalid.txt
```diff
@@ -1 +0,0 @@
-../../../../interop-test-files/syntax/did_syntax_invalid.txt
```
matthieusieben, line 1: these symlinks are pointless. Replaced with direct imports of the "source" interop file path

bnewbold: I left a comment about this on the datetime PR: I think the relative path reference may cause problems when testing npm packages in isolation (eg, when not running the tests in this git repo). I'm not sure how important that is. For the Go implementation, I simply copy the files in to the individual packages (aka, get rid of the top-level `interop-test-files` directory in this repo). Defer to you though.

bnewbold (commented):
the general code changes here seem fine.

I'll reply over on the interop test PR about the actual syntax strictness.

PR #4748, opened 2026-03-13
feat: change the uploadBlob endpoint to use both the ip and did

fatfingers23:
Since the uploadBlob endpoint is an authenticated endpoint we have access to the user's did and can use that as a key along with the ip for rate limits. This would help with backends that do large amounts of blob uploads on behalf of users, like PDS migration tools to help limit them being blocked globally for all users by ip to a single PDS

There is some concerns that an attacker could just rotate through did's to attack the server, but ideally registration is locked down by invite codes or a captcha to hopefully make it a bit harder for an attacker to leverage that.

### DIFF
--- a/packages/pds/src/api/com/atproto/repo/uploadBlob.ts
+++ b/packages/pds/src/api/com/atproto/repo/uploadBlob.ts
@@ -14,6 +14,7 @@ export default function (server: Server, ctx: AppContext) {
       },
     }),
     rateLimit: {
+      calcKey: ({ auth, req }) => `${auth.credentials.did}-${req.ip}`,
       durationMs: DAY,
       points: 1000,
     },


matteomarjanovic:
Has there been any progress for this?

I think having a daily limit by IP is too strict, especially for apps where users share pictures or similar.

An app that has 2000 users from the same pds allows only to 1000 of them to share a picture every day. This sounds odd to me.

Using both IP and did sounds like a reasonable solution to me.

lefelys:
Are there any plans to address this?
Apps built as confidential clients (following official atproto tutorials) can't use features that require blob uploads

If changing rate limit key is not an option, maybe consider uploads to PDS via presigned URLs?

ThisIsMissEm:
I just opened this which is upstreamed from Eurosky: https://github.com/bluesky-social/atproto/pull/5121

I've also mentioned this to various members on the Bluesky team privately.

fatfingers23:
@ThisIsMissEm i like that pr and agree with it. You're probably thinking the same thing. But I think both should be merged. The did and ip combo will be nice since not every PDS will set those new env variables. Especially for app developers who have no idea what PDS the user may be on, and who I think have been hitting this issue lately with their projects.

matteomarjanovic:
Any chance this gets picked back up? It's blocking for my app, [Currents](https://currents.is/). All users' uploads share one appview IP, so the IP-only uploadBlob limit is a shared chokepoint: a few bulk imports recently saturated it and 429'd unrelated users' saves. IP + DID keying would fix it.

PR #4752, opened 2026-03-16
❇️ Report activity log

foysalit:
This PR adds a dedicated table to keep report activities such as state change or internal note and meta data etc. Logging certain activities allow changing status of reports which is the primary way for escalating/acknowledging reports.

### DIFF
--- a/packages/ozone/src/mod-service/report.ts
+++ b/packages/ozone/src/mod-service/report.ts
@@ -3,22 +3,11 @@ import { AtUri } from '@atproto/syntax'
 import { Database } from '../db'
 import { Report } from '../db/schema/report'
 import { QueryParams } from '../lexicon/types/tools/ozone/report/queryReports'
-
-export function getReportStatusForEventType(eventType: string): string | null {
-  // Returns the status that reports should transition to for a given event type
-  // Returns null if the event type doesn't affect report status
-  switch (eventType) {
-    case 'tools.ozone.moderation.defs#modEventAcknowledge':
-    case 'tools.ozone.moderation.defs#modEventTakedown':
-    case 'tools.ozone.moderation.defs#modEventLabel':
-    case 'tools.ozone.moderation.defs#modEventComment':
-      return 'closed'
-    case 'tools.ozone.moderation.defs#modEventEscalate':
-      return 'escalated'
-    default:
-      return null
-  }
-}
+import {
+  AlreadyInTargetState,
+  InvalidStateTransition,
+  handleReportUpdate,
+} from '../report/handle-report-update'
 
 export type ReportWithEvent = Omit<Report, 'id'> & {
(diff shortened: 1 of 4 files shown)


packages/ozone/src/queue/service.ts
```diff
@@ -259,7 +259,13 @@ export class QueueService {
...
       .innerJoin('moderation_event as me', 'me.id', 'r.eventId')
-      .select(['r.id', 'me.subjectUri', 'me.subjectMessageId', 'me.meta'])
+      .select([
+        'r.id',
+        'r.status',
+        'me.subjectUri',
+        'me.subjectMessageId',
+        'me.meta',
+      ])
       .orderBy('r.id', 'asc')
```
gcwill70, line 271: I think this was missed when this logic was written initially but we should filter on status in the base query.

`.where('r.status', '!=', 'closed')`

packages/ozone/src/queue/service.ts
```diff
@@ -318,29 +328,77 @@ export class QueueService {
...
 
       if (report.id > maxReportId) maxReportId = report.id
     }
 
+    // Bulk UPDATE matched reports — one query per distinct queue, combines
+    // queueId + queuedAt + status into a single UPDATE
+    for (const [queueId, group] of matchedByQueue) {
+      await this.db.db
+        .updateTable('report')
+        .set({ queueId, queuedAt: now, status: 'queued', updatedAt: now })
```
gcwill70: We aren't checking state transition logic here. Looks like we should only update the status to `queued` if the current status is `open` to avoid overriding the status of escalated or assigned reports

packages/ozone/src/mod-service/report.ts
```diff
@@ -362,22 +371,41 @@ export async function processReportAction(
...
-      .where('id', '=', reportId)
-      .execute()
-  }
+  // Bulk UPDATE all matched reports in a single query
+  await db.db
+    .updateTable('report')
+    .set({
+      actionEventIds: sql`COALESCE("actionEventIds", '[]'::jsonb) || ${JSON.stringify(eventId)}::jsonb`,
+      actionNote: reportAction.note ?? null,
+      status: newStatus,
```
gcwill70: Looks like this isn't checking the state transition logic either

packages/ozone/src/queue/service.ts
foysalit: yeah good catch!

packages/ozone/src/mod-service/report.ts
foysalit: yep yep!

packages/ozone/src/queue/service.ts
foysalit: nice catch!

PR #4753, opened 2026-03-16
Avoid Go zero dates (year 0001) in bsky views

rafaeleyng:
This issue of zero dates has multiple facets to it:
1. Our Go dataplane returns them, where ideally it wouldn't.
2. Our new lex cli tooling doesn't accept dates with year < 10, which is something we'll likely change soon to support dates down to year 1.
3. In views where dates are optional, it is better to return `undefined` when we don't have a date than to make up a date for it (zero date, epoch, etc).

This PR attempts to fix `3.`.

### DIFF
--- a/packages/bsky/src/hydration/actor.ts
+++ b/packages/bsky/src/hydration/actor.ts
@@ -216,7 +216,9 @@ export class ActorHydrator {
               uri: `at://${actorDid}/app.bsky.graph.verification/${verificationMeta.rkey}`,
               handle: verificationMeta.handle,
               displayName: verificationMeta.displayName,
-              createdAt: verificationMeta.sortedAt.toDate().toISOString(),
+              createdAt:
+                parseDate(verificationMeta.sortedAt.toDate())?.toISOString() ??
+                new Date(0).toISOString(),
             }
           }
           // Filter out the verification meta that doesn't contain all info.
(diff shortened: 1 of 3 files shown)


packages/bsky/src/hydration/util.ts
```diff
@@ -121,6 +121,13 @@ export const parseCid = (cidStr: string | undefined): CID | undefined => {
   }
 }
 
+export const parseDate = (date: Date | undefined): Date | undefined => {
+  if (!date) return undefined
+  // Check for year 1 (0001-01-01 00:00:00 UTC) which is -62135596800000ms from epoch
```
matthieusieben: I would add that this is the "GO zero value generated by the dataplane"

.changeset/old-hats-type.md
```diff
@@ -0,0 +1,2 @@
+---
+---
```
matthieusieben: We probably should have a version bump here.

packages/bsky/src/hydration/util.ts
devinivy: Is it a go thing specifically, or a grpc/protobufs thing?

rafaeleyng: I believe it is a Go thing because those match the zero-values for dates, but not 100% sure.

f88da8bcd

.changeset/old-hats-type.md
rafaeleyng: f88da8bcd

devinivy (approved):
See my above comment.  We can always tidy this up at any time though, shouldn't be a correctness issue.  Might be nice to have a test that touches on this functionality.  In any case, looks good.

PR #4755, opened 2026-03-16
bsky: remove feature gate for including format in image URLs

devinivy:
Finalizes work to remove image format from CDN URLs.  Now the appview will not include the format in the URL, which will offer more agility in the future to update or negotiate image formats.

Refs #4712 #4723 #4721 #4726

### DIFF
--- a/.changeset/twenty-news-divide.md
+++ b/.changeset/twenty-news-divide.md
@@ -2,4 +2,4 @@
 '@atproto/bsky': patch
 ---
 
-Remove feature gate for including formate in image URLs.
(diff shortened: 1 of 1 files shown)


.changeset/twenty-news-divide.md
```diff
@@ -0,0 +1,5 @@
+---
+'@atproto/bsky': patch
+---
+
+Remove feature gate for including formate in image URLs.
```
devinivy:
```suggestion
Remove feature gate for including format in image URLs.
```

PR #4757, opened 2026-03-17
pds: use entryway mock server for tests

devinivy:
This replaces a stale dependency on legacy entryway code with a mock entryway server for the @atproto/pds package tests.  As a result we also no longer reference a problematic, old version of sharp.  Should help to unstick future nodejs upgrades.

Probably want to hold off on merging until #4408 lands to avoid unnecessary conflicts.

### DIFF
--- a/packages/pds/tests/entryway-mock.ts
+++ b/packages/pds/tests/entryway-mock.ts
@@ -10,13 +10,15 @@ import { AtpAgent } from '@atproto/api'
 import { getVerificationMaterial } from '@atproto/common'
 import { Secp256k1Keypair, randomStr } from '@atproto/crypto'
 import { IdResolver, getDidKeyFromMultibase } from '@atproto/identity'
+import { DidString, HandleString } from '@atproto/syntax'
 import {
   AuthRequiredError,
+  createServer,
   parseReqNsid,
   verifyJwt as verifyServiceJwt,
 } from '@atproto/xrpc-server'
 import { bearerTokenFromReq, createPublicKeyObject } from '../src/auth-verifier'
-import { createServer } from '../src/lexicon'
+import { com } from '../src/lexicons/index.js'
 
 interface Account {
   did: string
@@ -163,7 +165,7 @@ export class MockEntryway {
 
     const server = createServer()
 
-    server.com.atproto.server.createAccount({
+    server.add(com.atproto.server.createAccount, {
       handler: async ({ input }) => {
         const { email, handle } = input.body
 
@@ -222,7 +224,7 @@ export class MockEntryway {
         return {
           encoding: 'application/json' as const,
           body: {
-            did: plcCreate.did,
+            did: plcCreate.did as DidString,
             handle,
             accessJwt,
             refreshJwt,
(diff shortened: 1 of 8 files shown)


ThisIsMissEm:
Heeey, yay! Someone did this! I know I'd mentioned doing it but not having the time to yet

packages/pds/tests/entryway-mock.ts
```diff
@@ -0,0 +1,304 @@
...
+    const idResolver = new IdResolver({ plcUrl: opts.plcUrl })
+
+    const accounts = new Map<string, Account>()
+
+    const getSigningKey = async (
+      iss: string,
+      forceRefresh: boolean,
+    ): Promise<string> => {
+      const [did, serviceId] = iss.split('#')
+      assert(!serviceId, 'no service id expected in iss claim')
```
matthieusieben: Nit: why not throwing an `AuthRequiredError` ?

packages/pds/tests/entryway-mock.ts
```diff
@@ -0,0 +1,304 @@
...
+
+    // Auth: verify user access token (typ: 'at+jwt') signed by entryway
+    const accessAuth = async ({
+      req,
+    }: {
+      req: http.IncomingMessage
+    }): Promise<AccessAuthResult> => {
+      try {
+        const token = bearerToken(req)
+        const { payload } = await jose.jwtVerify(token, jwtPublicKey)
```
matthieusieben: Function comment says we check `typ: 'at+jwt'` but `jwtVerify` has no `{ typ: 'at+jwt' }` option argument.

Might be worth making the expected behavior explicit (e.g. `{ typ: undefined }` with a comment)

packages/pds/tests/entryway-mock.ts
```diff
@@ -0,0 +1,304 @@
...
+    }): Promise<ServiceAuthResult> => {
+      try {
+        const token = bearerToken(req)
+        const nsid = parseReqNsid(req)
+        const payload = await verifyServiceJwt(
+          token,
+          opts.serviceDid,
+          nsid,
+          getSigningKey,
+        )
```
matthieusieben, line 157: Here too, we should explicitly check that the `typ` is !== 'at+jwt'

packages/pds/tests/entryway-mock.ts
```diff
@@ -0,0 +1,304 @@
...
+        if (!account) {
+          throw new Error(
+            `Could not find account for DID: ${auth.credentials.did}`,
+          )
+        }
+        return {
+          encoding: 'application/json' as const,
+          body: {
+            did: account.did as DidString,
+            handle: account.handle as HandleString,
```
matthieusieben: Nit: Wouldn't it be better to type `did: DidString` and `handle: HandleString` in `Account` ?

matthieusieben (approved):
Looking good. I expected this would be way more involved than that!

packages/pds/tests/entryway-mock.ts
matthieusieben: I see now that we are checking the `typ` before calling this but I still think it would be valuable to have it here (e.g. to allow the method to be used on its own)

PR #4758, opened 2026-03-17
`datetime` validation consistency

matthieusieben:
This change performs the following:

- change `@atproto/lexicon`'s `datetime` format validation to be using `@atproto/syntax` (unifying the code base around a single datetime validation implementation)
- allows `@atproto/syntax` to accept **any** RFC 3339, ISO 8601, and WHATWG HTML compliant date time string, including date times that are less than 24H before the start of year 0 or less than 24H after the end of year 9999, and, in particular datetimes before year 10.


packages/syntax/src/datetime.ts
```diff
@@ -264,29 +319,59 @@ function parseString(input: unknown): Result<AtprotoDate> {
...
-  if (!DATETIME_REGEX.test(input)) {
+  const matches = input.match(DATETIME_REGEX)
+  if (!matches) {
     return failure(
       "datetime is not in a valid format (must match RFC 3339 & ISO 8601 with 'Z' or ±hh:mm timezone)",
     )
   }
 
-  // must parse as ISO 8601; this also verifies semantics like leap seconds and
-  // correct number of days in month, which the regex does not check for
```
matthieusieben, line 307: This comment was inaccurate:
- JS does not support leap seconds (eg. `new Date("2012-06-30T23:59:60+00:00") // "Invalid Date"`)
- JS allows invalid month/day combinations (eg. `new Date("2012-02-31")`), which are not allowed by `iso-datestring-validator`

interop-test-files/syntax/datetime_syntax_invalid.txt
```diff
@@ -63,6 +63,5 @@
 1985-04-12T23:20:50.123+
 1985-04-12T23:20:50.123-
 
-# ISO-8601, but normalizes to a negative time
-0000-01-01T00:00:00+01:00
```
matthieusieben, line 67: This value is actually not defined in interop tests

bnewbold: to clarify: this value is not in the "syntax invalid" interop tests (which this file should match), but it is in the "parse invalid" tests

packages/syntax/tests/datetime.test.ts
```diff
@@ -3,19 +3,21 @@ import { describe, expect, it, test } from 'vitest'
...
   isValidDatetime,
   normalizeDatetime,
   normalizeDatetimeAlways,
-} from '../src'
+  toDatetimeString,
+} from '../src/datetime.js'
 
 const interopValid = readLines(
-  `${__dirname}/interop-files/datetime_syntax_valid.txt`,
+  `${__dirname}/../../../interop-test-files/syntax/datetime_syntax_valid.txt`,
```
bnewbold, line 14: IIRC, doing references outside the package directory like this made tests fail when folks try to run the tests from the npm package. I can't remember the exact details, but I vaguely remember this being a problem a long time ago.

My proposed change would be to remove the symlinks and move the interop test files to live right here in the package; and remove the `interop-test-files` directory at the top of this repo.

(I'll defer to you on this, fine with me if you go ahead with this PR as it is, just wanted to share that history)

packages/syntax/tests/datetime.test.ts
```diff
@@ -43,10 +45,146 @@ describe(ensureValidDatetime, () => {
...
+
+  it('rejects non-string input', () => {
+    expect(() => ensureValidDatetime(123)).toThrow(InvalidDatetimeError)
+    expect(() => ensureValidDatetime(null)).toThrow(InvalidDatetimeError)
+    expect(() => ensureValidDatetime(undefined)).toThrow(InvalidDatetimeError)
+  })
+
+  it('rejects strings that are too long', () => {
+    expect(() =>
+      ensureValidDatetime('2024-01-15T12:30:00.' + '0'.repeat(50) + 'Z'),
```
bnewbold, line 98: would be good to get this in the interop tests file

bnewbold (commented):
this mostly looks like a great cleanup!

changing the behaviors around the "zero datetime" to be more flexible about things which are negative when normalized to UTC feels like a breaking change. for example, the indigo implementation strictly checks this by comparing in UTC timezone. discussed as well as review feedback over on the interop tests PR: https://github.com/bluesky-social/atproto-interop-tests/pull/13

jacoblapenna:
Hi! I've been building spec-conformance tests for a Python AT Protocol project, cross-checking against `@atproto/syntax` and the interop files, and this PR touches exactly the edges where I've found ecosystem divergence. Two vector suggestions and one question, while these files are already open:

**1. Consider adding a bare year-0000 valid vector.** The [Lexicon spec's valid examples](https://atproto.com/specs/lexicon#datetime) list `0000-01-01T00:00:00.000Z` under "preferred", and [current `datetime.ts` accepts it](https://github.com/bluesky-social/atproto/blob/20c5cc187aab538435c669c6e19a2d2f658af5f8/packages/syntax/src/datetime.ts#L362-L366) (only `fullYear < 0` fails) — but [`datetime_syntax_valid.txt`](https://github.com/bluesky-social/atproto/blob/20c5cc187aab538435c669c6e19a2d2f658af5f8/interop-test-files/syntax/datetime_syntax_valid.txt) never exercises it. The gap has real cost: at least [one downstream SDK](https://github.com/MarshalX/atproto/issues/697) rejects it undetected (Python's stdlib `datetime` can't represent year 0, so without a vector the divergence is invisible to its green CI). Since this PR re-draws the year-0 boundary anyway, locking the accepted side with a vector seems timely.

**2. The removed `0000-01-01T00:00:00+01:00` might belong in the valid file rather than deleted.** If the new behavior accepts datetimes within 24h before year 0000, that exact string becomes the ideal positive test of the new boundary. Deleting it leaves the new rule with no vector exercising it in either direction.

**3. Question: should the 64-char cap be reflected in the spec text?** [`parseString` rejects inputs over 64 chars](https://github.com/bluesky-social/atproto/blob/20c5cc187aab538435c669c6e19a2d2f658af5f8/packages/syntax/src/datetime.ts#L329-L330), but the Lexicon spec doesn't mention any length limit for `datetime`. Implementations following the spec text alone accept longer strings and silently diverge from the network. Since this PR is about validation consistency, it may be the right moment to either document the cap in the spec (and add a boundary vector) or reconsider it.

Happy to send the vectors as a PR against this branch or `main`, whichever is less churn for you. Thanks for the cleanup, this area was subtle to pin down for an outsider.

PR #4760, opened 2026-03-17
Fix `normalizeDatetime` inconsistencies

matthieusieben:
Currently, `syntax` tests can fail depending on the timezone of the machine running them. This is due to the fact that parsing a date without timezone information "as is" will result in the timezone of the machine being used.

The fix consists of attempting more consistent parsing strategies first (appending "Z" or " UTC" to the input) before falling back to parsing "as is", which can yield different results depending on the local timezone.

Capture d’écran 2026-03-17 à 14 31 44


packages/syntax/src/datetime.ts
```diff
@@ -163,19 +163,51 @@ export function toDatetimeString(date: Date): DatetimeString {
...
-  if (isAtprotoDate(date)) {
-    return date.toISOString()
-  }
+  if (
+    // Explicit timezone offset
+    /[+-]\d\d:?\d\d/.test(dtStr) ||
+    // 'Z' timezone designator
+    /\dZ\b/.test(dtStr) ||
+    // Timezone abbreviation (eg. "EST", "PST", "UTC", "GMT", etc)
+    /\b[A-Z]{3,4}\b/.test(dtStr)
```
devinivy, line 182: These will match anywhere in the string, is that intentional?

matthieusieben: yes that is intentional to match values generated with `new Date().toString()` like:

```
Tue Mar 17 2026 16:38:44 PST (Pacific Standard Time)
```

packages/syntax/src/datetime.ts
```diff
@@ -163,19 +163,51 @@ export function toDatetimeString(date: Date): DatetimeString {
...
+
+    const dateZ = new Date(`${dtStr}Z`)
+    if (isAtprotoDate(dateZ)) {
+      return dateZ.toISOString()
+    }
+
+    const dateUTC = new Date(`${dtStr} UTC`)
+    if (isAtprotoDate(dateUTC)) {
+      return dateUTC.toISOString()
+    }
```
devinivy, line 206: I don't think we need worry about these kinds of dates.   I believe the primary use case here is for common variants of ISO-ish strings.  (It's kind of unfortunate that `Date` even opens the door to parsing arbitrary date strings.)

matthieusieben: The only goal here is to avoid any kind of inconsistencies across systems that would be caused by processing the date "as is", which is what was causing the tests to fail.

matthieusieben: My point is that it doesn't hurt to have this, even if we can't assess its actual usefulness.

devinivy: If the main issue is these kinds of dates which we aren't so concerned about, an alternative would be to set `TZ` to standardize the environment in which the tests run.

packages/syntax/src/datetime.ts
devinivy: Ah got it!  That is hard to see.  I think if that's the case we should leave a comment around it.

I still do have an itch about explicitly supporting these non-standard date strings, hmmm.

packages/syntax/src/datetime.ts
matthieusieben: Apparently this is [not a silver bullet](https://github.com/nodejs/node-v0.x-archive/issues/3286).

This also won't prevent the code from running incoherently at runtime (eg if the server is setup with a different locale than utc)

packages/syntax/src/datetime.ts
matthieusieben: It is not so much about supporting them. It more a heuristic to avoid using the `Z` suffix strategy with the dates we don't support. The heuristic that was there before ("does not end with a letter") was kind of flimsy.

packages/syntax/src/datetime.ts
matthieusieben: What I mean here is that if we wanted to play with `TZ`, I would argue to run the tests with different values of `TZ`, and make sure that they all succeed, instead of forcing `TZ` to a value we control and have a flaky behavior in production, because the `TZ` there is not what we expected.

PR #4761, opened 2026-03-17
Allow processing of invalid lex data

matthieusieben:
This change adds an option, to the lex SDK `Client`, allowing the processing of JSON (response) payloads that are not conformant to the Lexicon Data Model (ie. invalid floats, $bytes and $link objects).

### DIFF
--- a/packages/lex/lex-client/src/client.ts
+++ b/packages/lex/lex-client/src/client.ts
@@ -748,7 +748,13 @@ export class Client implements Agent {
         : T extends Query
           ? XrpcRequestParams<T>
           : never,
-    options?: CallOptions,
+    options?: T extends Action
+      ? CallOptions
+      : T extends Procedure
+        ? Omit<XrpcOptions<T>, 'body'>
+        : T extends Query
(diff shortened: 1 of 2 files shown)


packages/lex/lex-client/src/xrpc.test.ts
```diff
@@ -583,6 +583,111 @@ describe(xrpc, () => {
       expect(response.body).toEqual({ value: 'hello' })
     })
   })
+
+  describe('allowInvalidLexData', () => {
```
devinivy: It would be nice to see the difference between `validateResponse: true` vs. `false` when used in combination with `allowInvalidLexData: true`.

devinivy: I suppose `allowInvalidLexData: false` and `validateResponse: false` is also an interesting combination to crystallize the behavior through tests as well.

matthieusieben: @copilot create a PR based on these comments.

PR #4762, opened 2026-03-17
improve checking for zero dates and test

rafaeleyng:


packages/bsky/tests/views/profile.test.ts
```diff
@@ -609,6 +610,40 @@ describe('pds profile views', () => {
     })
   })
 
+  it.only('filters out Go zero-value dates from dataplane', async () => {
```
devinivy:
```suggestion
  it('filters out Go zero-value dates from dataplane', async () => {
```

packages/bsky/src/hydration/actor.ts
```diff
@@ -252,13 +252,13 @@ export class ActorHydrator {
         profile: profile?.record,
         profileCid: profile?.cid,
         profileTakedownRef: profile?.takedownRef,
-        sortedAt: parseDate(profile?.sortedAt),
-        indexedAt: parseDate(profile?.indexedAt),
+        sortedAt: profile?.sortedAt,
+        indexedAt: profile?.indexedAt,
```
rafaeleyng, line 256: `sortedAt` and `indexedAt` don't need `parseDate` because they are reading from a record, which already used `parseDate` inside `parseRecord`.

packages/bsky/tests/views/profile.test.ts
rafaeleyng: Just testing you. You did well.

185b87875fe3e99491a4f8211c762624e881fbdd

PR #4774, opened 2026-03-19
Fix infinitely deep declaration

matthieusieben:
Fixes a regression introduced in https://github.com/bluesky-social/atproto/pull/4761 which manifested itself in real projects

### DIFF
--- a/.changeset/slow-bobcats-laugh.md
+++ b/.changeset/slow-bobcats-laugh.md
@@ -1,5 +1,0 @@
----
-'@atproto/lex-schema': patch
----
-
-Improve return type of `Payload.matchesEncoding()`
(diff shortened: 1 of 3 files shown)


43081j:
@matthieusieben this seems to have introduced a breaking change in the types 🤔 

```ts
await client.delete(dev.npmx.feed.like, {
  rkey
});
```

this no longer compiles since the options now require `body` to be set. but that doesn't seem right for a deletion.

should the delete options also `Omit` `'body'`?

the client always sets the `body` itself internally:

https://github.com/bluesky-social/atproto/blob/3b9af1bdb622607467d39407d10d57f67c5e1687/packages/lex/lex-client/src/client.ts#L603-L609

PR #4788, opened 2026-03-23
Allow payload from XRPC endpoints with no `output` in their schema

matthieusieben:
The goal is to relax processing, by `lex-client` of XRPC responses when no `output` encoding is specified. If nothing is specified, anything is allowed.

### DIFF
--- a/packages/lex/lex-client/src/response.ts
+++ b/packages/lex/lex-client/src/response.ts
@@ -215,7 +215,10 @@ export class XrpcResponse<M extends Procedure | Query>
       parse:
         method.output.schema || method.output.encoding === CONTENT_TYPE_JSON
           ? { strict: options?.strictResponseProcessing ?? true }
-          : false,
+          : // If there is no declared output encoding, we'll parse the output (in loose mode)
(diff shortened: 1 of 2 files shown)


packages/lex/lex-client/src/response.ts
```diff
@@ -217,18 +219,11 @@ export class XrpcResponse<M extends Procedure | Query>
...
-          payload,
-          `Expected response with no body, got "${payload.encoding}"`,
-        )
-      }
-    } else {
-      // Schema expects a payload
+    if (method.output.encoding != null) {
+      // If the schema specifies an output encoding, verify that the response
+      // properly matches the expected output format (encoding and schema). If
+      // no output encoding is specified, any payload could be returned.
```
devinivy: Honestly, I thought that `encoding` was required if there was an `output`.  So I expected this case to arise of there is no `output`.

matthieusieben: in `lex-schema`'s methods, `method.output` is always present, but `encoding` might not be (if the json schema does not have an `output`)

PR #4789, opened 2026-03-23
Ozone report stats

gcwill70:
# Changes

* Create report_stat table
  * Holds aggregate, per queue, per moderator, and per report category stats
  * Calculates stats each day. Stats for current day are re-computed every 15 minutes and finalized the next day.
* Create report stats service to refresh stats
* Create stats-computer daemon that runs every 15 minutes


packages/ozone/src/db/migrations/20260318T152058935Z-add-report-stat.ts
```diff
@@ -0,0 +1,36 @@
...
+  await db.schema
+    .createIndex('idx_report_stat_lookup')
+    .on('report_stat')
+    .columns(['mode', 'timeframe', 'queueId', 'computedAt'])
+    .execute()
+}
+
+export async function down(db: Kysely<unknown>): Promise<void> {
+  await db.schema.dropIndex('idx_report_stat_lookup').ifExists().execute()
+  await db.schema.dropIndex('idx_report_stat_computed_at').ifExists().execute()
```
foysalit: i dont see this index anywhere?

packages/ozone/src/report/stats.ts
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    const stats = await this.computeGroup(group)
+    const result = await this.updateGroup(group, stats)
+    return result
+  }
+
+  private isGroupFresh(stats: Selectable<ReportStat>): boolean {
+    const ttl =
+      stats.mode === 'live'
```
foysalit: should probably hoist these up as constants

packages/ozone/src/report/stats.ts
```diff
@@ -0,0 +1,225 @@
...
+  private async getLatestStats(
+    group: ReportStatGroup,
+  ): Promise<Selectable<ReportStat> | undefined> {
+    let qb = this.db.db
+      .selectFrom('report_stat')
+      .selectAll()
+      .where('mode', '=', group.mode)
+      .where('timeframe', '=', group.timeframe)
+      .orderBy('computedAt', 'desc')
+    qb = qb.where('queueId', '=', group.queueId)
```
foysalit: probably doesn't need to be a separate assignment and can be chained on to the initial builder.

foysalit (commented):
looking good. having trouble to wrap my head around the 'fixed' stats though. can you describe how that's supposed to work or are we leaving that implementation for later?

packages/ozone/src/report/stats.ts
```diff
@@ -0,0 +1,690 @@
...
+        ? REPORT_STAT_LIVE_TTL
+        : stats.timeframe === 'day'
+          ? 24 * HOUR
+          : 7 * 24 * HOUR
+    const expiresAt = Date.now() - ttl
+    const computedAt = new Date(stats.computedAt).getTime()
+    return computedAt > expiresAt
+  }
+
+  private async updateGroup(
```
foysalit: we usually run the daemon in one container but in the future, with more and more load moving to the daemon, we might want to start thinking about concurrent containers. in such setup, i think this would cause some issue where we are deleting and inserting without transaction. besides concurrent containers, requests arriving in between delete and insert query could be an issue. 

I would wrap this in a transaction or maybe do an upsert?

packages/ozone/src/report/stats.ts
```diff
@@ -0,0 +1,690 @@
...
+    return {
+      inboundCount,
+      pendingCount,
+      actionedCount,
+      escalatedCount,
+      actionRate,
+      avgHandlingTimeSec,
+    }
+  }
+  private async computeQueue(group: ReportStatGroup): Promise<QueueStatistics> {
```
foysalit: seeing a lot of similarities in the query generated by this and computeAggregate and computeReportType methods, we could make it a bit more DRY

packages/ozone/src/api/queue/listQueues.ts
```diff
@@ -25,10 +25,14 @@ export default function (server: Server, ctx: AppContext) {
         reportTypes,
       })
 
+      const queues = await Promise.all(
+        result.queues.map((q) => queueService.viewWithStats(q)),
```
foysalit: hmmm... even though these are single rows being read per query, still feels a bit risky to fire off many queries when we can do a WHERE IN query instead?

packages/ozone/src/db/schema/report_stat.ts
```diff
@@ -0,0 +1,32 @@
...
+  reportTypes: string[] | null // NULL = aggregate across all report types
+  moderatorDid: string | null // NULL = aggregate, non-null = per-moderator
+
+  // stats
+  inboundCount: number | null // Reports received in the last 24 hours
+  pendingCount: number | null // Reports with status = 'open' (all time)
+  actionedCount: number | null // Reports with status = 'closed' in last 24h
+  escalatedCount: number | null // Reports with status = 'escalated' in last 24h
+  actionRate: number | null // actionedCount / inboundCount * 100
+  avgHandlingTimeSec: number | null // Average time from open/assigned to closed, in milliseconds
```
foysalit: nit the comment sounds wrong? should be seconds, no?

packages/ozone/src/report/stats.ts
```diff
@@ -0,0 +1,690 @@
...
+        'tools.ozone.team.defs#roleAdmin',
+        'tools.ozone.team.defs#roleModerator',
+        'tools.ozone.team.defs#roleTriage',
+      ])
+      .execute()
+    // live
+    /// aggregate
+    groups.push({
+      mode: 'live',
+      timeframe: 'day',
```
foysalit: i may be missing something but it looks like we want to support day and week timeframes but in the compute part, we are only creating groups for day timeframe?

packages/ozone/src/report/stats.ts
```diff
@@ -0,0 +1,690 @@
...
+
+    const createdAt = await this.db.db
+      .selectFrom('report')
+      .select([
+        sql<number>`count(*)`.as('inboundCount'),
+        sql<number>`count(*) filter (where "status" = 'closed' and "updatedAt" > ${cutoff})`.as(
+          'actionedCount',
+        ),
+        sql<number>`count(*) filter (where "status" = 'escalated' and "updatedAt" > ${cutoff})`.as(
+          'escalatedCount',
```
foysalit, line 358: i think we mixed up the term `escalatedCount` and `escalatedPendingCount`?

foysalit (approved):
looking good! mostly concerned about the number of queries being fired off since we compute quite frequently. i think i would want to trigger recompute less frequently and added some grouping to queries to reduce query count.
also used claude to write some more documentation.

packages/ozone/src/api/queue/listQueues.ts
gcwill70: Good catch! Your update looks good

packages/ozone/src/report/stats.ts
gcwill70: I thought I'd start simple with daily only and add more timeframes after user testing and ensuring the solution scales properly

PR #4805, opened 2026-03-25
Replace DataDog with OTEL in the PDS

matthieusieben:
## Replace dd-trace with vendor-neutral OpenTelemetry (PDS)

Replaces the Datadog-specific `dd-trace` tracer in the PDS image with the OpenTelemetry NodeSDK, configured via conventional `OTEL_*` environment variables (or `OTEL_CONFIG_FILE`). The deployment side — fanning the OTLP stream out to both the Datadog and Grafana agents — can be performed at the infra level.

### Telemetry

- New `@atproto/pds/telemetry` entry point (loaded via `node --import`) replaces `services/pds/tracer.ts`. It registers a curated set of instrumentations (http, express, undici, aws-sdk, ioredis, pino, better-sqlite3, node runtime) instead of `auto-instrumentations-node`, and honors `OTEL_SDK_DISABLED`.
- Incoming/outgoing XRPC spans get a normalized `http.route` (`/xrpc/<nsid>`) via request hooks, preserving the old dd-trace "resource name" behavior (until dedicated Xrpc instrumentations exist).
- New `pds.account.creations` counter as a first app-level metric.
- The PDS image is labeled `social.bsky.pds.telemetry=otel` so deployment tooling can distinguish otel-based images from dd-trace ones.
- New `pds` bin script (`packages/pds/bin/pds.mjs`) to run a PDS from the published package.

### Docker builds

- Dockerfiles (pds, bsky, bsync, ozone) now copy only each service's actual transitive workspace dependencies and build via a single `--filter '<service>...'` pass; `.dockerignore` tightened accordingly. We remove build concurrency as it lead to inconsistent builds (in particular in CI).

### Notes

- The other services (bsky, bsync, ozone) still use `dd-trace`; only the PDS is migrated here.


packages/pds/src/index.ts
```diff
@@ -62,14 +69,82 @@ export class PDS {
   public app: express.Application
   public server?: http.Server
   private terminator?: httpTerminator.HttpTerminator
-  private dbStatsInterval?: NodeJS.Timeout
-  private sequencerStatsInterval?: NodeJS.Timeout
```
matthieusieben, line 65: dead code

packages/pds/src/scripts/rotate-keys.ts
```diff
@@ -131,16 +129,5 @@ const updatePlcSigningKey = async (ctx: RotateKeysContext, did: DidString) => {
     // already up to date
     return
   }
-  if (ctx.entrywayAdminAgent) {
```
matthieusieben, line 134: `entrywayAdminAgent` is never set

devinivy: Wow, nice find 🙈

packages/pds/src/metrics.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export const oauthAuthorizationCounter = meter.createCounter<{
+  clientId: string
+}>('oauth.authorization', {
+  description: 'Increased when an OAuth authorization is granted on this PDS',
+  valueType: ValueType.INT,
+})
+
+export const oauthTokenRefreshedCounter = meter.createCounter<{
+  clientId: string
```
devinivy: Any concern with the cardinality of `clientId` here?

packages/pds/src/telemetry.ts
```diff
@@ -0,0 +1,248 @@
...
+
+        // set the xrpc.method attribute for both incoming and outgoing requests
+        if (nsid) {
+          span.updateName(`${method} /xrpc/${nsid}`)
+          span.setAttribute(ATTR_XRPC_METHOD, nsid)
+        }
+      },
+    }),
+    new ExpressInstrumentation({
+      ignoreLayersType: [ExpressLayerType.MIDDLEWARE],
```
devinivy, line 116: I forget exactly what counts a middleware in our current setup.  E.g. does this ignore auth?

packages/pds/src/telemetry.ts
```diff
@@ -0,0 +1,248 @@
...
+        const nsid = extractNormalizedXrpcNsid(request.path)
+        if (nsid) {
+          span.setAttribute(ATTR_XRPC_METHOD, nsid)
+        }
+      },
+    }),
+    new AwsInstrumentation(),
+    new IORedisInstrumentation(),
+    new BetterSqlite3Instrumentation(),
+    new PinoInstrumentation(),
```
devinivy: Is this definitely how we want to ship logs, including for ourselves in prod?  I thought we had systems based on stdout/stderr already working.  Does stdout/stderr still get produced work normally?  I am pretty curious in what the overhead looks like too.

devinivy: While I am not sure about shipping logs this way, I do feel good about log correlation.

services/bsky/Dockerfile
```diff
@@ -7,25 +7,46 @@ WORKDIR /app
...
-# Move files into the image and install
 COPY ./pnpm-lock.yaml ./
 COPY ./pnpm-workspace.yaml ./
 COPY ./tsconfig ./tsconfig
 COPY ./lexicons ./lexicons
-COPY ./packages/internal ./packages/internal
-COPY ./packages/lex ./packages/lex
+
+# bsync is not a runtime dep but bsky's codegen:buf reads ../bsync/proto at build time
+COPY ./packages/bsync ./packages/bsync
```
devinivy: Should we only pull over `packages/bsync/proto` as we do in pds?

services/ozone/Dockerfile
```diff
@@ -7,17 +7,16 @@ WORKDIR /app
...
-# Move files into the image and install
 COPY ./pnpm-lock.yaml ./
 COPY ./pnpm-workspace.yaml ./
 COPY ./tsconfig ./tsconfig
 COPY ./lexicons ./lexicons
-COPY ./packages/internal ./packages/internal
-COPY ./packages/lex ./packages/lex
+
+# bsync is not a runtime dep but bsky's codegen:buf reads ../bsync/proto at build time
+COPY ./packages/bsync ./packages/bsync
```
devinivy: Should we only pull over `packages/bsync/proto` as we do in pds?

package.json
```diff
@@ -30,8 +30,8 @@
     "verify:style": "pnpm run style",
     "verify:lint": "pnpm lint",
     "format": "pnpm lint:fix && pnpm style:fix",
-    "codegen": "pnpm run build:tooling && pnpm run --parallel --workspace-concurrency Infinity '/^(codegen:.+)$/'",
-    "prebuild": "pnpm run build:tooling && pnpm run --parallel --workspace-concurrency Infinity 'prebuild'",
+    "codegen": "pnpm run build:tooling --force && pnpm run --parallel --workspace-concurrency Infinity '/^(codegen:.+)$/'",
+    "prebuild": "pnpm run build:tooling --force && pnpm run --parallel --workspace-concurrency Infinity 'prebuild'",
```
devinivy, line 35: Just double-checking that you intend to keep these `--force`s around.

package.json
```diff
@@ -48,7 +48,9 @@
     "version-packages": "changeset version && git add ."
   },
   "devDependencies": {
+    "@atproto/bsky": "workspace:^",
     "@atproto/dev-env": "workspace:^",
+    "@atproto/pds": "workspace:^",
```
devinivy, line 54: What is the purpose of these additions?

services/pds/Dockerfile
```diff
@@ -83,8 +98,11 @@ ENV UV_USE_IO_URING=0
 
 # https://github.com/nodejs/docker-node/blob/master/docs/BestPractices.md#non-root-user
 USER node
-CMD ["node", "--heapsnapshot-signal=SIGUSR2", "--enable-source-maps", "--import=./tracer.ts", "index.ts"]
+CMD ["node", "--heapsnapshot-signal=SIGUSR2", "--enable-source-maps", "--import=@atproto/pds/telemetry", "-e", "import '@atproto/pds/bin'"]
```
devinivy: This is clever but also has some considerations:
 - in the pds distro we still will need to find a way to wire up the `version` differently.
 - anyone overriding the default command in their docker compose may find breakage.
 - using `-e` instead of running a script for this is not something I've personally seen before, maybe a little atypical.

Could we keep the `index.ts` entrypoint and just make it a one-liner?  I think it deals with at least two of these neatly.

package.json
matthieusieben: yeah. It doesn't hurt and was cause of trouble.

package.json
matthieusieben: It avoids the circular dependency between pds and bsky. They only depend on each other for tests, which is arguably a `dev` feature of the mono repo.

services/pds/Dockerfile
matthieusieben: I've restored `index.ts`. The `version` can still be wired:

1) Either by keeping its current code
2) Or by using the new `PDS.run()` static method:

```ts
import { PDS, readEnv } from '@atproto/pds'
import pkg from "@atproto/pds/package.json" with { type: "json" };

const DISTRO_VER = "0.4";

void PDS.run({
  env: readEnvWithVersion(),
  onCreated: (pds) => {
    pds.app.get("/tls-check", (req, res) => {
      checkHandleRoute(pds, req, res);
    })
  }
})

function readEnvWithVersion() {
  const env = readEnv()
  env.version ||= ver(DISTRO_VER, pkg.version);
  return env
}

async function checkHandleRoute(pds: PDS, req: Request, res: Response) {
  // ...
}

// e.g. ver('0.4', '0.5.1') -> '0.4.5001'
function ver(base: `${string}.${string}`, pkgver: string) {
  // ...
}
```

packages/pds/src/telemetry.ts
matthieusieben: auth is not a middleware in our current setup (we need to instrument xrpc-server for that). Middlewares here are only the `app.use` middleware (cors, etc.)

packages/pds/src/metrics.ts
matthieusieben: I've replaced these with logs 👍

packages/pds/src/telemetry.ts
matthieusieben: I've changed the way logging works. We now log directly to OTEL events we want to track, and we disable the pino -> OTEL log sending.

packages/lex/lex-builder/src/lex-def-builder.ts
```diff
@@ -103,6 +94,28 @@ export class LexDefBuilder {
...
+
+    this.file.addTypeAlias({
+      name,
+      type: `typeof ${name}`,
+      docs: docs ? [docs] : undefined,
+    })
+
+    this.file.addExportDeclaration({
+      namedExports: [{ name }],
+    })
```
matthieusieben, line 116: To the reader; This results in:

```ts
const $nsid = "app.example.foo"
type $nsid = typeof $nsid
export { $nsid }
```

which allows `$foo` to be used as both a type and value:

```ts
function doStuff(string: app.example.foo.$nsid | app.example.bar.$nsid) {

}

doStuff(app.example.foo.$nsid) // ok
doStuff(app.example.bar.$nsid) // ok
```

Before this, we needed to do:

```ts
function doStuff(
  string:
    | typeof app.example.foo.$nsid
    | typeof app.example.bar.$nsid
) {

}

```

devinivy (approved):

PR #4806, opened 2026-03-25
Improve `AtUriString` type an utilities

matthieusieben:
The current `isValidAtUri` is not spec compliant. So much so that "invalid" interop tests are disabled. This change addresses that by introducing a `strict` mode for validation (allows "legacy" behavior by using string: false).

In addition to this, the following issues are also tackled:

The `AtUri` class adds a trailing slash when stringifying an `AtUri` with no collection/rkey component. Trailing slashed are not allowed in the at-uri format. This PR fixes that.

There are currently two implementation: `ensureValidAtUri` and `ensureValidAtUriRegex`. They essentially differ in terms of the level of details they provide. They also have incoherent behaviors: one enforces the length restriction to the full input while the other applies if only to the part before the fragment. This PR changes that by unifying them into a single implementation that allows opting into detailed errors.

`AtUriString` is typed as a template literal but is missing the fragment part.

The RegExp validation was non optimal as it requires the regepx engine to backtrack depending on the authority type (did vs handle). This had no benefit since the authority part was validated (though `ensureValidHandleRegex` and `ensureValidDidRegex`) after regexp matching. This change removes this un-necessary complexity in the regexp.

In addition to this, the regexp was missing non-capturing group indicators (`(?:...)`) causing un-necessary strings to be extracted into memory by the regex engine.

The `isValidAtUri` function uses try/catch to detect invalid format, which is hurts performances, especially in worst case scenarios. Similarly, validation uses nested try/catch to check the validity of `authority`. The first try/catch would always throw if the URI's authority is a DID. Benchmarks on my machine have showed a 400% efficiency improve of this function by using `isAtIdentifierString` instead of the nested try/catch logic.

The exposed interface of `AtUriString` related utilities was also updated to match the utilities of other types. A dedicated error class was created in addition to utilities matching those related to `DatetimeString` and `AtIdentifierString`. These utilites are also re-exported as-is from `@atproto/lex`.


packages/syntax/src/aturi_validation.ts
```diff
@@ -1,152 +1,311 @@
...
-  if (fragmentIndex !== -1) {
-    if (input.charCodeAt(fragmentIndex + 1) !== 47) {
-      throw new Error('ATURI fragment must be non-empty and start with slash')
-    }
-    if (input.includes('#', fragmentIndex + 1)) {
-      throw new Error('ATURI can have at most one "#", separating fragment out')
-    }
+  options?: Omit<ParseUriStringOptions, 'detailed'>,
+): input is I & AtUriString {
+  return parseUriString(input, options).success
```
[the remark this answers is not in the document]
matthieusieben: We don't really need this since 1) `AtUri` supports trailing slashes 2) `${string}` matches `${string}/`

packages/syntax/src/lib/result.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export type Success<T> = { success: true; value: T }
+export function success<T>(value: T): Success<T> {
+  return { success: true, value }
+}
+
+export type Failure = { success: false; message: string }
+export function failure(message: string): Failure {
+  return { success: false, message }
+}
```
matthieusieben, line 11: Factorized from `nsid.ts`

packages/syntax/src/aturi.ts
```diff
@@ -133,19 +133,27 @@ export class AtUri {
...
-      path = `/${path}`
+    let pathname = this.pathname
+    if (pathname && !pathname.startsWith('/')) {
+      pathname = `/${pathname}`
     }
+    while (pathname.endsWith('/')) {
+      pathname = pathname.slice(0, -1)
+    }
+    // @TODO query strings are actually not allowed by the AT URI specification,
+    // but we support them for legacy reasons. This should be removed.
```
matthieusieben:
```suggestion
```

packages/syntax/src/aturi_validation.ts
```diff
@@ -1,152 +1,319 @@
...
-  } catch {
-    try {
-      ensureValidDidRegex(groups.authority)
-    } catch {
-      throw new Error('ATURI authority must be a valid handle or DID')
+  if (options?.strict !== false) {
+    if (groups.trailingSlash != null) {
+      return failure('ATURI can not have a trailing slash')
+    }
+
```
matthieusieben, line 276: Should we also forbid query string in strict mode? (since we don't use them)

packages/syntax/src/aturi_validation.ts
```diff
@@ -1,152 +1,319 @@
...
-    authorityEnd === -1 ? uri.slice(5) : uri.slice(5, authorityEnd)
-  try {
-    ensureValidAtIdentifier(authority)
-  } catch (cause) {
-    throw new Error('ATURI authority must be a valid handle or DID', { cause })
+  const invalidChar = input.match(INVALID_CHAR_REGEXP)
+  if (invalidChar) {
+    return failure(
+      `ATURI contains invalid character "${invalidChar[0]}" at position ${invalidChar.index}`,
+    )
```
matthieusieben:
```suggestion
    return failure('Disallowed characters in ATURI (ASCII)')
```

packages/syntax/src/aturi_validation.ts
bnewbold: I'd say yes, forbid query string in strict mode

packages/lex/lex-schema/src/core/string-format.ts
```diff
@@ -72,20 +74,22 @@ export function isDatetimeStringLoose<I>(
...
  */
-export const isAtUriString: CheckFn<AtUriString> = isValidAtUri
-export type {
-  /**
-   * An AT URI string pointing to a resource in the AT Protocol network.
-   *
-   * @example `"at://did:plc:1234.../app.bsky.feed.post/3k2..."`
-   */
-  AtUriString,
+export function isAtUriStringLoose<I>(input: I): input is I & AtUriString {
```
bnewbold: if we were doing these from scratch, i'd recommend "lenient" over "loose". but given there are already some "loose" methods, being consistent is probably most important.

packages/syntax/src/aturi_validation.ts
```diff
@@ -1,152 +1,317 @@
...
+ */
+export function isValidAtUri<I>(input: I): input is I & AtUriString {
+  return isAtUriString(input, { strict: false })
+}
 
-  if (!uri.startsWith('at://')) {
-    throw new Error('ATURI must start with "at://"')
+export class InvalidAtUriError extends Error {}
+
+export type ParseUriStringOptions = {
```
bnewbold: should this be `ParseAtUriStringOptions` to avoid ambiguity with `format: uri` parsing at some future point?

bnewbold (commented):
This PR seems like a solid improvement on the status quo, and I don't really see anything here I would block moving ahead with.

I do think the AT-URI spec is more flexible than needed right now, and we could maybe simplify the implementation here if we made the spec more conservative. For example, we just don't use query string syntax anywhere in the protocol ecosystem, and we could just reserve that for future use in the spec. Same with fragment syntax (unless folks are using the fragment syntax somewhere i'm not aware of). We could probably more clearly rule out percent-encoding in almost all situations (or maybe all situations, to keep it simple?).

The above is most relevant in lexicon validation of fields in records. Eg, for the `parent` field of a bsky post, we really wouldn't want query string or fragment syntax in there.

As a cross-reference, the go (indigo) syntax package simply forbids query params and fragments.

packages/syntax/src/aturi_validation.ts
matthieusieben: yep

PR #4809, opened 2026-03-26
Create new endpoints for suggested users

ds-boyce:
Feature gated so we can test them.

### DIFF
--- a/lexicons/app/bsky/unspecced/getSuggestedUsersForDiscover.json
+++ b/lexicons/app/bsky/unspecced/getSuggestedUsersForDiscover.json
@@ -4,7 +4,7 @@
   "defs": {
     "main": {
       "type": "query",
-      "description": "Get a list of suggested users for Discover",
+      "description": "Get a list of suggested users for the Discover page",
       "parameters": {
         "type": "params",
         "properties": {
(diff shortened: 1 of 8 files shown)


packages/bsky/src/api/app/bsky/unspecced/getSuggestedUsersForExplore.ts
```diff
@@ -0,0 +1,178 @@
...
+      )
+      return {
+        encoding: 'application/json',
+        body: result,
+      }
+    },
+  })
+}
+
+// TODO: rename to `skeleton` once we can fully migrate to Discover
```
rafaeleyng: This comment would be better suited on the other endpoint, given that this is the one that will go away.

Also, it is not migrating to Discover, but to "the new endpoint".

ds-boyce: Whoops, copypasta.

ds-boyce:
Thanks, @iwsmith! I removed `category` from `getSuggestedUsersForDiscover`. Can you take another look?

lexicons/app/bsky/unspecced/getSuggestedUsersForDiscover.json
```diff
@@ -0,0 +1,41 @@
+{
+  "lexicon": 1,
+  "id": "app.bsky.unspecced.getSuggestedUsersForDiscover",
+  "defs": {
+    "main": {
+      "type": "query",
+      "description": "Get a list of suggested users for Discover",
```
rafaeleyng:
```suggestion
      "description": "Get a list of suggested users for the Discover page",
```
to avoid confusion with the system we internally call `discover`, which incidentally is the one that fulfills all of these requests.

Could do the same for all 3 cases.

PR #4821, opened 2026-04-02
Workaround missing dependency when installing npm@latest

ds-boyce:
This is causing deployments to fail, e.g., https://github.com/bluesky-social/atproto/actions/runs/23920326164/attempts/1

See issues linked in comments for more context:
* https://github.com/npm/cli/issues/9151
* https://github.com/npm/cli/pull/9152

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -26,6 +26,10 @@ jobs:
           cache: pnpm
           node-version-file: '.nvmrc'
           registry-url: 'https://registry.npmjs.org'
+      # Temporary workaround for an issue with Node.js v22
+      # https://github.com/npm/cli/issues/9151
+      # https://github.com/npm/cli/pull/9152
+      - run: npm install --global npm@11.11.1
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yaml
```diff
@@ -26,6 +26,10 @@ jobs:
           cache: pnpm
           node-version-file: '.nvmrc'
           registry-url: 'https://registry.npmjs.org'
+      # Temporary workaround for an issue with Node.js v22
+      # https://github.com/npm/cli/issues/9151
+      # https://github.com/npm/cli/pull/9152
+      - run: npm install --global npm@11.11.1
       - run: npm install --global npm@latest
```
estrattonbailey, line 33: Do we need to rm this line?

ds-boyce: It can stay if we want to be on latest. 11.11.1 installs the missing dependency, which enables latest to install successfully. But we could just pin 11.11.1 until the bug is fixed.

PR #4827, opened 2026-04-05
[APP-1983] New gallery embed type

estrattonbailey:
The purpose of this embed type is to give us room to expand on the embeddable media in a post. We want to increase the number of images allowed on a post, and eventually allow a mix of media types e.g. images and video.

For now, we will begin with supporting only images, and likely limit the total number of images to 10.


mfnboer:
Shouldn't there be an update to drafts as well? Otherwise a post with a gallery embed cannot be saved as draft.

blackmichael:
@mfnboer great call out, I just realized this myself and came here to say the same

packages/bsky/src/views/index.ts
```diff
@@ -2133,6 +2141,44 @@ export class Views {
     })
   }
 
+  galleryEmbed(did: DidString, embed: GalleryEmbed): $Typed<GalleryEmbedView> {
+    const items = embed.items.flatMap((item) => {
+      const view = this.galleryItemView(did, item)
+      return view ? [view] : []
+    })
```
estrattonbailey: We should probably slice the array here to 10 items to enforce the soft limit

estrattonbailey (commented):
Looks great! One last tweak I think

.changeset/few-squids-hammer.md
```diff
@@ -0,0 +1,5 @@
+---
+'@atproto/bsky': patch
```
estrattonbailey, line 2: Oh, also need to add `api` here too so we get new types

estrattonbailey (commented):
Nice one! ✅

PR #4829, opened 2026-04-06
Remove impossible RecordNotFound error

uniphil:
`com.atproto.sync.getRecord` must prove the non-existence of a record when it's not found, which requires returning a CAR slice, so it cannot return an XRPC-shaped error body, so `RecordNotFound` can't be a compliant response. I think.

the description of this endpoint itself says

> Get data blocks needed to prove the existence **or non-existence** of record in the current version of repo. Does not require auth.

(emphasis added)


bnewbold (approved):
good catch!

PR #4847, opened 2026-04-12
lex: more flexible 'aud' for getServiceAuth

bnewbold:
Updates `getServiceAuth` to support single-string `aud` syntax.

Background/motivation is described in https://github.com/bluesky-social/proposals/tree/main/0013-service-auth-refs; this is a revised/alternative proposal, which matches the spec updates proposed in https://github.com/bluesky-social/atproto-website/pull/616

Before merging:
- [ ] run codegen
- [ ] fix any build errors (eg, in PDS/entryway implementation)
- [ ] create a changeset

### DIFF
--- a/lexicons/com/atproto/server/getServiceAuth.json
+++ b/lexicons/com/atproto/server/getServiceAuth.json
@@ -4,15 +4,14 @@
   "defs": {
     "main": {
       "type": "query",
-      "description": "Get a signed token on behalf of the requesting DID for the requested service.",
+      "description": "Get a signed token on behalf of the requesting DID for the requested service, which can be used to make an authenticated XRPC HTTP API request. Implemented by PDS, requires authentication, and constrained by session permissions.",
       "parameters": {
         "type": "params",
         "required": ["aud"],
         "properties": {
           "aud": {
             "type": "string",
-            "format": "did",
-            "description": "The DID of the service that the token will be used to authenticate with"
+            "description": "The DID (and optional service ID as a fragment) of the recipient service for this request (the \"audience\")."
           },
           "exp": {
             "type": "integer",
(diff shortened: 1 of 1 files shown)


lexicons/com/atproto/server/getServiceAuth.json
```diff
@@ -4,15 +4,14 @@
...
+      "description": "Get a signed token on behalf of the requesting DID for the requested service, which can be used to make an authenticated XRPC HTTP API request. Implemented by PDS, requires authentication, and constrained by session permissions.",
       "parameters": {
         "type": "params",
         "required": ["aud"],
         "properties": {
           "aud": {
             "type": "string",
-            "format": "did",
-            "description": "The DID of the service that the token will be used to authenticate with"
+            "description": "The DID (and optional service ID as a fragment) of the recipient service for this request (the \"audience\")."
```
matthieusieben:
```suggestion
            "description": "The DID and service ID as a fragment (or optionally a standalone DID) of the recipient service for this request (the \"audience\")."
```

matthieusieben (approved):
Looks good. I left a suggestion as we may want to emphasis on the use of fragments, since they work better with oauth scopes. There is no oauth scope that allows to describe "no fragment". Enabling empty fragments will require to request the `aud=*` permission, which is not desirable. Another option would be to prevent the creation of service auth tokens without fragment altogehter (I think this is currently the case), which would naturally "push" towards fragments everywhere.

lexicons/com/atproto/server/getServiceAuth.json
bnewbold: I updated this description to cover these points

bnewbold:
Something else that I noticed when going over this is that the `lxm` field is currently optional. I added to the description that it is "effectively required". I'm not sure if we would ever allow that to not be set?

Maybe we should break the evolution rules and make it required? But don't need to in this PR if it would slow things down.

PR #4852, opened 2026-04-13
Data Plane Caller Interceptor

jcalabro:
Similar to the changes we've made in Discover and elsewhere, this enables the AppView to self-identify when calling the data plane so we can track it nicely in metrics.

This would have been useful in debugging last weeks' major incident.

Note that this was written by claude; I'm not super familiar with this code base, but a quick interceptor that sets this header seems reasonable?

### DIFF
--- a/.changeset/easy-insects-dress.md
+++ b/.changeset/easy-insects-dress.md
@@ -1,0 +1,5 @@
+---
+'@atproto/bsky': patch
+---
+
+Add interceptor to dataplane GRPC client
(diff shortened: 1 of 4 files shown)


rafaeleyng (approved):

rafaeleyng (commented):
To your question, I don't know the exact way to add an interceptor on this lib, but what you did here seems to do it just about right, so don't see any issues.

You could generate a small test that spins up a small server and just checks that header.

PR #4854, opened 2026-04-15
group chat lexicons

rafaeleyng:
We're adding support for group chats 🐴🐴🐴

These might still change a bit, even after we merge this. Note how I've left a big ol' warning everywhere (for the lack of a proper way to flag unstable lexicons). Once I remove that, they should be stable.


mozzius:
@claude please review

mozzius (approved):
🐴

PR #4853, opened 2026-04-14
Enforce Max Depth on lexToIpld

jcalabro:
I'm unsure that this is actual root cause fix of this issue. Definitely want review from @matthieusieben before this goes out

This was suggested by claude for fixing this user's issues https://bsky.app/profile/did:plc:rhzwcsk5kcoidziamyopsjez

as suggested by this log line:

> "payload":{"error":"InternalServerError","message":"Internal Server Error"},"name":"xrpc-server","msg":"unhandled exception in xrpc method app.bsky.feed.getAuthorFeed"}

Screenshot 2026-04-14 at 19 35 07

It seems to be caused by this repo?

```
$ atp repo export did:plc:b2wskpxho4srwilqyssafllz
wrote did:plc:b2wskpxho4srwilqyssafllz.car (5375403 bytes)

$ atp repo inspect did\:plc\:b2wskpxho4srwilqyssafllz.car
DID:  did:plc:b2wskpxho4srwilqyssafllz
Rev:  3mckxtl4azw22
Version:  3
Records:  9417

  app.bsky.actor.profile                   1
  app.bsky.actor.status                    1
  app.bsky.feed.like                       36
  app.bsky.feed.post                       9366
  app.bsky.feed.repost                     1
  app.bsky.graph.block                     1
  app.bsky.graph.follow                    1
  com.example.blob                         8
  com.example.htmlblob                     2

$ atp resolve did:plc:b2wskpxho4srwilqyssafllz
DID:  did:plc:b2wskpxho4srwilqyssafllz
Handle:  test142342.bsky.social
PDS:  https://inkcap.us-east.host.bsky.network
Signing:  did:key:zQ3shVja3jn4uK62iaLrhBeESaqeQCSr2BGfFGrXXfJ1Ndr5J  (K-256)
```

PR #4856, opened 2026-04-15
Support serialization of deeply nested JSON structures

matthieusieben:
Currently, lex data manipulation rely on `JSON.parse` and `JSON.stringify` for json serialization. The issue with this is that `JSON.parse` will allow pretty big/complex data structures (e.g. unlimited nested `[]`), while `JSON.stringify` will throw a `RangeError` when it reaches a depth of ~1300.

This is bad for several reasons:

- Servers are able receive and process payloads (both from downstream clients and from upstream services), but end-up returning a 500 as serialization (to downstream clients) fails.
- There is no check whatsoever on the max complexity of incoming data (besides total HTTP request body length). Because of `JSON.parse`, most endpoint will happily accept data with excessively high nesting depth.
- The depth of data for which `JSON.stringify` will start throwing depends on the current execution stack. This makes the server behavior inconsistent (some data might cause some code path to throw while other don't).

In AT Protocol, most data is constrained in terms of size. Data is always serialized as JSON and/or CBOR when in transit or stored. This ensures that:
1) The size of the data is not unlimited (generally bound to whatever the PDS & appview accept as max size for incoming data).
2) No circular references can occur.

This PR improves the status quo by making explicit limits in term of size of the data we accept. It uses a `strict` to allow more lenient limits in read path, while imposing stricter limits on write path.

This PR will make sure that any lexicon data that is accepted as input can safely be returned as output using `lexStringify`, by supporting highly nested data where `JSON.stringify` fails.

The limits introduced are:
- maxDepth: the maximum nesting depth that will be allowed
- maxArrayLength: the max number of items an array can have
- maxObjectEntries: the max number of entries an objects can have
- maxNestingFactor: the max total number of object and arrays in the data (at any level)

> [!NOTE]
> The current limits where arbitrarily set and are subject to change. Input is welcome in that regard.

Note that CBOR serialization/deserialization can still cause recursion `RangeError` because of its recursive nature. I believe that we should rely on a CBOR implementation that allows to explicitly set these limits. This is left as another work stream.

Similarly, we should specify what to do when importing data as raw bytes (e.g. through `importRepo`) that have a nesting level that is too deep (reject those specific records, truncate the records, reject the whole import, etc. ?)


jcalabro (approved):
Thanks for picking this up! In reference to #4853

DavidBuchanan314:
This is definitely an improvement but it's worth noting that clients might start crashing when processing very deep objects, instead. Do we have depth limits in lex validation? (Might make sense to consider a too-deep object to be invalid)

packages/lex/lex-json/src/lex-json.ts
```diff
@@ -1,126 +1,19 @@
...
- * const json = lexStringify({
- *   ref: someCid,
- *   data: new Uint8Array([72, 101, 108, 108, 111])
- * })
- * // json is '{"ref":{"$link":"bafyrei..."},"data":{"$bytes":"SGVsbG8="}}'
- * ```
+ * @see {@link LexParseOptions}
+ * @see {@link JsonTransformOptions}
  */
-export function lexStringify(input: LexValue): string {
```
matthieusieben, line 38: move in another file

packages/lex/lex-json/src/lex-json.ts
```diff
@@ -248,105 +89,9 @@ function jsonObjectToLexMap(
...
-      copy[key] = jsonValue
-    }
-  }
-  return copy ?? (input as JsonObject)
-}
-
-/**
- * @internal
- */
-export function parseSpecialJsonObject(
```
matthieusieben, line 316: moved to another file

packages/lex/lex-data/src/constants.ts
```diff
@@ -0,0 +1,82 @@
...
+ * record, including both those with and without Lexicon-specified string
+ * lengths. Measured as bytes (UTF-8 encoded). Should try to keep these to tens
+ * of KBytes at most. For an upper-bound limit (MAX_RECORD_STRING_LEN),
+ * reasonable to just rely on the overall CBOR record size limit. Notably, some
+ * early implementations had a 8 KByte (8192 bytes) limit.
+ *
+ * @see
+ * {@link https://atproto.com/guides/data-validation#recommended-data-limits}
+ */
+export const MAX_RECORD_STRING_LEN = MAX_JSON_RECORD_SIZE
```
DavidBuchanan314: nit: would using `MAX_CBOR_RECORD_SIZE` here match the comment/spec more directly?

packages/lex/lex-data/src/constants.ts
```diff
@@ -0,0 +1,82 @@
...
+
+/**
+ * **Container nesting depth**: for example, how many layers of map inside an
+ * array inside an array, etc. If your CBOR or JSON parsing library supports a
+ * limit, the default is probably fine.
+ *
+ * @see
+ * {@link https://atproto.com/guides/data-validation#recommended-data-limits}
+ */
+export const MAX_CBOR_NESTED_LEVELS = 32
```
DavidBuchanan314: This *is* the ~~spec's~~ guide's recommendation but it seems very conservative to me, I think we could safely go up to at least 256 or so. If we do, might be worth updating the guide likewise.

packages/xrpc-server/tests/bodies.test.ts
```diff
@@ -570,6 +614,27 @@ for (const buildServer of [buildMethodLexicons, buildAddLexicons]) {
         message: 'Request encoding (Content-Type) required but not provided',
       })
     })
+
+    it.only('handles deeply nested output', async () => {
```
devinivy:
```suggestion
    it('handles deeply nested output', async () => {
```

packages/tap/src/channel.ts
```diff
@@ -110,6 +110,7 @@ export class TapChannel implements AsyncDisposable {
     this.abortController.signal.throwIfAborted()
     try {
       for await (const chunk of this.ws) {
+        // @NOTE types are wrong. chunks are actually string (at least in tests)
```
devinivy, line 113: Curious!  Good find.

packages/lex/lex-data/src/constants.ts
matthieusieben: Note that in practice, this will only be applied to new records (`createRecord`). There is another limit (5_000) for read paths.

matthieusieben: I did update to `256`

packages/lex/lex-json/src/index.test.ts
```diff
@@ -0,0 +1,521 @@
+import { describe, expect, test } from 'vitest'
```
matthieusieben: All tests here where moved from `lex-parse.test.ts` (no new test were added)

packages/lex/lex-json/src/json-to-lex.ts
```diff
@@ -0,0 +1,101 @@
+import {
```
matthieusieben: Was moved from `lex-json.ts`

packages/lex/lex-json/src/lex-parse.test.ts
```diff
@@ -0,0 +1,377 @@
+import { assert, describe, expect, it } from 'vitest'
```
matthieusieben, line 1: Was moved from `lex-json.test.ts`

packages/lex/lex-json/src/lex-parse.ts
```diff
@@ -0,0 +1,94 @@
+import { LexValue, utf8FromBytes } from '@atproto/lex-data'
```
matthieusieben, line 1: Was moved from `lex-json.ts`

matthieusieben: Only noticeable change is the options definition

packages/lex/lex-client/src/xrpc.test.ts
```diff
@@ -1168,7 +1168,6 @@ describe(xrpcSafe, () => {
       expect(result).toBeInstanceOf(XrpcInvalidResponseError)
       expect(result.message).toMatch('Unable to parse response payload')
       assert(result.cause instanceof TypeError)
-      expect(result.cause.message).toBe('Invalid blob object')
```
devinivy, line 1171: There are a few assertions like this in this file that were removed.  What is the purpose of the removal?

packages/lex/lex-data/src/utf8-to-base64.test.ts
```diff
@@ -1,35 +1,22 @@
...
+      'Hello, World!',
+      '¡Hola, Mundo!',
+      'こんにちは世界',
+      '😀👩‍💻🌍',
+      '',
+      '𓀀𓁐𓂀𓃰𓄿𓅱𓆑𓇋𓈖𓉔𓊃𓋴𓌳𓍿𓎛𓏏',
+    ])('%s', (text) => {
+      const expected = Buffer.from(text, 'utf8')
+        .toString(encoding)
+        .replaceAll('=', '')
```
devinivy: In this previous version of the tests this replacement did not occur in `encodes utf8 string to base64url`.  I expect this is fine because `=` will not occur in `base64url`, but I think it merits a comment.

devinivy (commented):
Left a few notes, mostly superficial things.  I see copilot also had some notes, and ours ended-up overlapping a little.

I think we should talk about @DavidBuchanan314's point.  It would be great not to have to sidestep `JSON.stringify()` and `JSON.parse()` to get a working protocol implementation in JS.  And I think the protocol implementation guidelines are intentionally giving us leeway to be able to use our runtime's builtins if we use them with care and are willing to work with a relatively minor tradeoff.  I think one of @DavidBuchanan314's points is that there's an interoperability issue at play here that gets highlighted when moving away from builtins too.

packages/lex/lex-client/src/xrpc.test.ts
matthieusieben: Because the error is now thrown from another place, that uses a different (but similar) error message. I don't think we need to make the `cause` part of the API contract so there is no point in checking the actual message (since we already check that the `cause` is an error).

matthieusieben:
See https://github.com/bluesky-social/atproto/pull/4923 for a version that does not implement a custom json stringifyer.

packages/lex/lex-json/src/deep-transform.ts
```diff
@@ -0,0 +1,326 @@
...
+      throw new TypeError(`Invalid ${typeof input} input`)
+  }
+
+  // "input" is an object (or array)
+
+  // Keep a reference to the initial nested level at which we started processing
+  // this input, so that we can restore it after processing this input and its
+  // children. This is more efficient than "context.initialNestedLevel++" and
+  // "context.initialNestedLevel--" around the recursive calls as it saves the
+  // JS engine from having to read the value before updating it.
```
devinivy, line 139: I am surprised that this would be measurable.

packages/lex/lex-json/src/json-stringify-deep.ts
```diff
@@ -0,0 +1,182 @@
+import { JsonValue } from './json.js'
```
devinivy, line 1: I am going to skip review here— I think we should land #4923 in this PR.

PR #4857, opened 2026-04-15
OAuth Hooks Enhancements

DavidBuchanan314:
- Add new `onSignInFailed` hook
- OAuth stores can optionally throw `InvalidCredentialsError` in place of `InvalidRequestError`, which allows the `onSignInFailed` hook to distinguish failures due to invalid identifier vs invalid password.
- Plumb `clientId` through to existing sign-in hooks.
- Update PDS OAuth store to throw `InvalidCredentialsError` when appropriate (Entryway will need similar changes)
- Test that we're not introducing any enumeration attack vectors, at the HTTP boundary.

These changes will enable improved entryway analytics.


packages/oauth/oauth-provider/src/request/request-manager.ts
```diff
@@ -316,6 +316,18 @@ export class RequestManager {
...
+  /**
+   * Reads the {@link ClientId} associated with a request URI without any of
+   * the validation or side-effects performed by {@link RequestManager.get}
+   *
+   * Returns `null` when no such request exists.
+   */
+  async peekClientId(requestUri: RequestUri): Promise<ClientId | null> {
+    const requestId = decodeRequestUri(requestUri)
+    const data = await this.store.readRequest(requestId)
+    return data?.clientId ?? null
```
matthieusieben: No need to coalesce to null here since the result will be coalesced to undefined


```suggestion
    return data?.clientId
```

matthieusieben (approved):
Super clean and we'll documented . Looks great

packages/oauth/oauth-provider/src/request/request-manager.ts
matthieusieben: Very much a nit. feel free to ignore

PR #4865, opened 2026-04-16
Add user-agent header to service proxied requests

matthieusieben:


bnewbold:
Because we are proxying, I always want there to be a proxy-specific header we could use. `Via` is the closest, but has a hostname or "pseudoname", not software type. Might still be worth including `Via` with the HTTP version and `atproto-pds` or something like that?

I'm still not sure if putting PDS implementation in the `User-Agent` header is the idiomatic HTTP thing to do, but i'm not directly opposed/blocking.

PR #4872, opened 2026-04-20
Fix/reset password

Spotchi:
This is a PR to indicate what is causing the issue for password reset on the PDS


matthieusieben:
Are you trying to add tests to the account manager ? I've been trying to this it [here](https://github.com/bluesky-social/atproto/pull/4844) in a way that does not need to rework the whole CI.

PR #4874, opened 2026-04-21
Add errors property to app.bsky.actor.getProfile

ThisIsMissEm:
This was causing `@atproto/lex` to fail parsing the response for when a profile is not found.

Arguably this should be a more explicit error, but that's what the bsky appview returns at the moment, e.g., https://public.api.bsky.app/xrpc/app.bsky.actor.getProfile?actor=did:plc:a2amvy3onlpobuzi3dspka7b

```
{"error":"InvalidRequest","message":"Profile not found"}
```

There's probably other cases where this occurs too.

cc @matthieusieben

### DIFF
--- a/lexicons/app/bsky/actor/getProfile.json
+++ b/lexicons/app/bsky/actor/getProfile.json
@@ -23,7 +23,7 @@
           "ref": "app.bsky.actor.defs#profileViewDetailed"
         }
       },
-      "errors": [{ "name": "InvalidRequest" }]
+      "errors": [{ "name": "InvalidRequest" }, { "name": "AccountDeactivated" }]
     }
   }
 }
--- a/lexicons/app/bsky/actor/getProfile.json
+++ b/lexicons/app/bsky/actor/getProfile.json
@@ -22,7 +22,8 @@
           "type": "ref",
           "ref": "app.bsky.actor.defs#profileViewDetailed"
         }
-      }
+      },
+      "errors": [{ "name": "InvalidRequest" }]
     }
   }
 }


ThisIsMissEm:
@bnewbold it may make sense to have a lint warning for queries and procedures not declaring `errors`, because almost every single query and procedure _can_ return some type of error.

lexicons/app/bsky/actor/getProfile.json
```diff
@@ -22,7 +22,8 @@
           "type": "ref",
           "ref": "app.bsky.actor.defs#profileViewDetailed"
         }
-      }
+      },
+      "errors": [{ "name": "InvalidRequest" }]
```
ThisIsMissEm:
```suggestion
      "errors": [{ "name": "InvalidRequest" }, { "name": "AccountDeactivated" }]
```

ThisIsMissEm:
Example of AccountDeactivated error: https://public.api.bsky.app/xrpc/app.bsky.actor.getProfile?actor=did:plc:kyqcmuwjmlljqtxbfitbqu2h

bnewbold:
hmmm!

I kind of think the client/SDK should be more resilient here. Eg, it should be possible to add new error types to a lexicon, and "old" client software should not crash/fail when encountering that. I think the defined error types should mostly be used for recovery code paths or provide better user experience.

Also think that we should probably provide generic errors at the protocol layer which can always be used. Eg, some errors that align with defined 4xx and 5xx errors (there is also a whole conversation/iteration to be done about how HTTP status codes are used in XRPC). Or things like `InvalidSyntax` (for invalid identifiers) or `InvalidData` (for requests which break general atproto data model). Maybe also these account status errors.

Basically I think lexicons should only need to declare special/unique errors that they return, not common/generic ones.

Having `getProfile` errors align with other similar endpoints would make sense right now though.

ThisIsMissEm:
Yeah, definitely agree that clients could be more resilient here, this specifically affects the `@atproto/lex` client on version 0.0.24 — so it may be fixed already in 0.0.25 (I haven't tested yet)

PR #4880, opened 2026-04-23
Fix `/.well-known/change-password` page

matthieusieben:
This fixes `/.well-known/change-password` no longer working. The reason is was no longer working is that https://github.com/bluesky-social/atproto/pull/4820 introduced a new router system that didn't have "unauthenticated" routes. Indeed, the `RouterProvider` was gated by the `AuthenticationProvider`.

This PR inverts the order of these two provider, creating two distinct sets of routes: authenticated and unauthenticated. The authenticated routes are still gated by the `AuthenticationProvider`, while unauthenticated routes can be accessed regardless of authentication status.


packages/oauth/oauth-provider-ui/src/pages/account/(authenticated)/route.tsx
```diff
@@ -0,0 +1,222 @@
...
+  // popup, to let the user manage their atmosphere account without leaving the
+  // app. In that case, we constrain the user to only use the account for which
+  // they opened the page, and we send a signal to the opener when they perform
+  // actions that signal the user is done with the page (like logging out, or
+  // explicitly "canceling" the sign-in).
+
+  // @NOTE The VERY EXPERIMENTAL API used here **WILL** change in the future as
+  // it gets specified (or not) in the AT Protocol specification. It MUST NOT be
+  // used in places where security is a concern and is ONLY there for testing
+  // and experimentation purposes. DO NOT USE.
```
devinivy, line 169: What API does this refer to specifically?

devinivy (approved):
I am not sure if I understand why grouping by auth/unauthed is necessary, versus making the choice on a page-by-page basis, but the work looks good to me.

packages/oauth/oauth-provider-ui/src/pages/account/(authenticated)/route.tsx
matthieusieben: A feature that allows to embed the account manager in a webview or browser popup.

```ts
const popup = window.open(`<user_pds>/account?login_hint=<user_handle>`, "_blank")

window.addEventListener('message', (event) => {
  if (event.origin === '<user_pds>') {
    popup.close()
    
    // refresh session data as the user might have updated
    // their email,handle,etc.
    session.refresh()
  }
})

```

matthieusieben:
Could have done it like that (and we still could). Since all the routes need auth, I figured it was simpler to just wrap everything into an auth guard.

packages/oauth/oauth-provider-ui/src/pages/account/(authenticated)/route.tsx
matthieusieben: The "API" is the communication primitives that can be used between the embedder (application webview or browser window) and the embedded (the account manager web page).

- DOM messages
- Redirection
- back channel ?
- other ?

PR #4892, opened 2026-04-28
Fix "impossible" comparison in serviceAuth

essential-randomness:
Fix `serviceAuth` from `@atproto/lex-server` comparing `typeof payload.lxm` (type) and `options.lxm` (actual string), which caused authenticated paths to always throw.


axfelix:
cc @matthieusieben

PR #4905, opened 2026-05-05
Add devEngines config for node and pnpm

ds-boyce:
Adds a `devEngines` config to ensure `pnpm` is used to install dependencies.

```sh
~/Projects/atproto d/package-engines
❯ yarn
This project is configured to use pnpm because /Users/dsboyce/Projects/atproto/package.json has a "packageManager" field
```

```sh
~/Projects/atproto d/package-engines
❯ npm i
npm warn Unknown project config "enable-pre-post-scripts". This will stop working in the next major version of npm. See `npm help npmrc` for supported config options.
npm error code EBADDEVENGINES
npm error EBADDEVENGINES The developer of this package has specified the following through devEngines
npm error EBADDEVENGINES Invalid devEngines.packageManager
npm error EBADDEVENGINES Invalid name "pnpm" does not match "npm" for "packageManager"
npm error EBADDEVENGINES {
npm error EBADDEVENGINES   current: { name: 'npm', version: '11.13.0' },
npm error EBADDEVENGINES   required: { name: 'pnpm', onFail: 'error' }
npm error EBADDEVENGINES }
```

Also adds an `executionEnv` config to automatically use the specified version of Node.js when running `pnpm`, although it won’t do anything until pnpm is updated to v9.6.0 or newer (currently on v8.15.9, latest is v11.0.6).

### DIFF
--- a/.github/workflows/repo.yaml
+++ b/.github/workflows/repo.yaml
@@ -16,6 +16,8 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: pnpm/action-setup@v4
+        with:
+          version: 8.15.9
       - uses: actions/setup-node@v6
         with:
           cache: pnpm
@@ -49,6 +51,8 @@ jobs:
         with:
           fetch-depth: 0 # needed for git diff against base branch
       - uses: pnpm/action-setup@v4
+        with:
+          version: 8.15.9
       - uses: actions/setup-node@v6
         with:
           cache: pnpm
@@ -81,6 +85,8 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: pnpm/action-setup@v4
+        with:
+          version: 8.15.9
       - uses: actions/setup-node@v6
         with:
           cache: pnpm
@@ -105,6 +111,8 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: pnpm/action-setup@v4
+        with:
+          version: 8.15.9
       - uses: actions/setup-node@v6
         with:
           cache: pnpm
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -8,7 +8,13 @@
   "engines": {
     "node": ">=18.7.0"
```
ds-boyce, line 9: @matthieusieben Should this be updated to match what’s in `.nvmrc`? That file specifies `22`.

.github/workflows/repo.yaml
```diff
@@ -16,6 +16,8 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: pnpm/action-setup@v4
+        with:
+          version: 8.15.9
```
devinivy: I feel like our package.json should be able to help us setup the package manager version, without having to be explicit about it everywhere.  Perhaps we keep both `packageManager` and `devEngines` for now?

devinivy: See also https://github.com/pnpm/action-setup/issues/150

package.json
```diff
@@ -8,7 +8,13 @@
...
   },
-  "packageManager": "pnpm@8.15.9+sha512.499434c9d8fdd1a2794ebf4552b3b25c0a633abcee5bb15e7b5de90f32f47b513aca98cd5cfd001c31f0db454bc3804edccd578501e4ca293a6816166bbd9f81",
```
devinivy, line 11: I think we may want to keep this for the time being, as some of our tooling (such as the pnpm github action) don't support `devEngines` yet.

package.json
devinivy: No, that's alright— this will all get synced-up in the upcoming move to node v22+.

devinivy (approved):

PR #4929, opened 2026-05-12
Upgrade packages to node v22+

devinivy:
This is the first in a suite of PRs to get us on node v22+, TypeScript 6, and with ESM-only packages.  Here's an overview of this PR:

  - Drop support for Node.js 18 and 20 across all packages. Node.js 22 is now the minimum supported version.
  - Docker images bumped to `node:24.15-alpine3.23`.
  - Upgrade `@types/node` from `^18` to `^22`.
  - Upgrade `better-sqlite3` from `^10` to `^12` (ships prebuilt binaries for Node 22/24, eliminates native compile dependency).
  - Upgrade `opentelemetry-plugin-better-sqlite3` to `^1.13` (widens peer range for better-sqlite3@12).
  - Simplify `@atproto-labs/fetch-node`: remove the `dangerouslyForceKeepAliveAgent` option and the Node <20 compatibility branch from `unicastFetchWrap` — on Node 22+ the dispatcher init is always supported. A runtime guard remains as a safety net.
  - Replace deprecated `corepack prepare --activate` with `corepack install` in Dockerfiles.
  - Add `"engines": { "node": ">=22" }` to every published package so consumers see warnings on unsupported versions.
  - Fix `WithImplicitCoercion` type in `@atproto/lex-data` — `@types/node@22` scoped this type inside `declare module "buffer"` rather than leaking it globally.

  ### Breaking changes

  - All packages now declare `engines.node >= 22`. Consumers on Node 18/20 will see npm/pnpm engine
  warnings.
  - `@atproto-labs/fetch-node`: `unicastFetchWrap` and `safeFetchWrap` no longer accept `dangerouslyForceKeepAliveAgent`.

### DIFF
--- a/.changeset/drop-node-18-20.md
+++ b/.changeset/drop-node-18-20.md
@@ -1,0 +1,65 @@
+---
+'@atproto/bsky': patch
+'@atproto/bsync': patch
+'@atproto/ozone': patch
+'@atproto/pds': patch
+'@atproto/api': minor
+'@atproto/aws': minor
+'@atproto/common': minor
+'@atproto/common-web': minor
+'@atproto/crypto': minor
+'@atproto/dev-env': minor
+'@atproto/did': minor
+'@atproto/identity': minor
+'@atproto/jwk': minor
+'@atproto/jwk-jose': minor
+'@atproto/jwk-webcrypto': minor
+'@atproto/lex': minor
+'@atproto/lex-builder': minor
+'@atproto/lex-cbor': minor
+'@atproto/lex-cli': minor
+'@atproto/lex-client': minor
+'@atproto/lex-data': minor
+'@atproto/lex-document': minor
+'@atproto/lex-installer': minor
+'@atproto/lex-json': minor
+'@atproto/lex-password-session': minor
+'@atproto/lex-resolver': minor
+'@atproto/lex-schema': minor
+'@atproto/lex-server': minor
+'@atproto/lexicon': minor
+'@atproto/lexicon-resolver': minor
+'@atproto/oauth-client': minor
+'@atproto/oauth-client-browser': minor
+'@atproto/oauth-client-browser-example': minor
+'@atproto/oauth-client-expo': minor
+'@atproto/oauth-client-node': minor
+'@atproto/oauth-provider': minor
+'@atproto/oauth-provider-api': minor
+'@atproto/oauth-provider-ui': minor
+'@atproto/oauth-scopes': minor
+'@atproto/oauth-types': minor
+'@atproto/repo': minor
+'@atproto/sync': minor
+'@atproto/syntax': minor
+'@atproto/tap': minor
+'@atproto/ws-client': minor
+'@atproto/xrpc': minor
+'@atproto/xrpc-server': minor
+'@atproto-labs/did-resolver': minor
+'@atproto-labs/fetch': minor
+'@atproto-labs/fetch-node': minor
+'@atproto-labs/handle-resolver': minor
+'@atproto-labs/handle-resolver-node': minor
+'@atproto-labs/identity-resolver': minor
+'@atproto-labs/pipe': minor
+'@atproto-labs/rollup-plugin-bundle-manifest': minor
+'@atproto-labs/simple-store': minor
+'@atproto-labs/simple-store-memory': minor
+'@atproto-labs/simple-store-redis': minor
+'@atproto-labs/xrpc-utils': minor
+---
+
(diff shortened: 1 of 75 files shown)


packages/internal/fetch-node/src/unicast.ts
```diff
@@ -15,181 +15,40 @@ const { IPv4, IPv6 } = ipaddr
...
-      const url = extractUrl(input)
-
-      if (url.hostname && isUnicastIp(url.hostname) === false) {
-        throw new FetchRequestError(
-          asRequest(input, init),
-          400,
-          'Hostname is a non-unicast address',
-        )
-      }
+  const dispatcher = new Agent({
```
matthieusieben, line 35: Since this is a security feature, we might want to keep a light check to ensure users don't unsafely use this:

```suggestion
  if (!SUPPORTS_REQUEST_INIT_DISPATCHER) {
    throw new Error('unicast protection unavailable on your platform. Update to Node22+')
  }

  const dispatcher = new Agent({
```

packages/lex/lex-data/src/lib/nodejs-buffer.ts
```diff
@@ -1,5 +1,10 @@
 type Encoding = 'utf8' | 'base64' | 'base64url'
 
+// Node's buffer module declares this type internally, but referencing it here
+// would couple this file to @types/node. Local copy keeps this module
```
matthieusieben, line 4: Nice catch. How did you even detect this ?!


We might want to also point at this config, which is why this is needed in the first place:

https://github.com/bluesky-social/atproto/blob/ad8f9bf63e8eb848ad7db3924b3102cdf6aa1068/tsconfig/isomorphic.json#L5-L12

.changeset/drop-node-18-20.md
```diff
@@ -0,0 +1,65 @@
...
+'@atproto-labs/identity-resolver': minor
+'@atproto-labs/pipe': minor
+'@atproto-labs/rollup-plugin-bundle-manifest': minor
+'@atproto-labs/simple-store': minor
+'@atproto-labs/simple-store-memory': minor
+'@atproto-labs/simple-store-redis': minor
+'@atproto-labs/xrpc-utils': minor
+---
+
+Drop support for Node.js 18 and 20. Node.js 22 is now the minimum supported version. Docker images now use `node:22-alpine3.23`.
```
matthieusieben: I like being able to see at a glance the breaking changes in a changelog. Emphasizing on the breaking nature of this change is particularly useful since we don't minor bump service packages.

```suggestion
**BREAKING:** Drop support for Node.js 18 and 20. Node.js 22 is now the minimum supported version. Docker images now use `node:22-alpine3.23`.
```

.changeset/drop-node-18-20.md
```diff
@@ -0,0 +1,65 @@
...
+
+`@atproto-labs/fetch-node`: `unicastFetchWrap` and `safeFetchWrap` no longer accept `dangerouslyForceKeepAliveAgent` — on Node.js 22+ the keep-alive dispatcher is always used via `new Request(input, { dispatcher })`, so the option was a no-op.
```
matthieusieben: I would split this in it's own changeset (otherwise this will appear in the CHANGELOG.md of every package)

matthieusieben (approved):
Looking good! Left a few small comments

.changeset/drop-node-18-20.md
matthieusieben: I would also add the `**BREAKING:** ` prefix.

packages/lex/lex-data/src/lib/nodejs-buffer.ts
devinivy: Done!  I believe it was an improvement to @types/node that turned this into a type error.  I think it's now an internal type you could technically import if you wanted to.

devinivy: (The type error would occur below at usage of `WithImplicitCoercion`)

packages/internal/fetch-node/src/unicast.ts
devinivy: Done! https://github.com/bluesky-social/atproto/pull/4929/changes/aaa880e791ef1e04855ed8c79e7b6fe8d3464fab

.changeset/drop-node-18-20.md
devinivy: Good call, done https://github.com/bluesky-social/atproto/pull/4929/changes/aaa880e791ef1e04855ed8c79e7b6fe8d3464fab

PR #4930, opened 2026-05-12
Upgrade packages to TypeScript 6

devinivy:
This is the first in a suite of PRs to get us on node v22+, TypeScript 6, and with ESM-only packages.  Waterfalls into #4929.  Here's an overview of this PR:

  - Bump `typescript` from `^5.8.3` to `^6.0.3` across all 44 workspace packages plus root.
  - Bump `@typescript-eslint/eslint-plugin` and `@typescript-eslint/parser` from `^7.4.0` to `^8.59.2` for TypeScript 6 compatibility.
  - Add `"ignoreDeprecations": "6.0"` to `tsconfig/base.json` to silence the `moduleResolution=node10` deprecation. The proper migration to `node16`/`bundler` resolution is deferred.
  - Address TypeScript 6's stricter `Uint8Array<ArrayBuffer>` typing:
    - Narrow return type of `ui8Concat` (`@atproto/lex-data`) to `Uint8Array<ArrayBuffer>`
    - Narrow interface types at producers where feasible (`repo/types`, `common-web/types`, `lex-client/types`, `oauth-client/runtime-implementation`, `bsync/db/schema`, `oauth-provider/dpop-nonce`)
    - Narrow `randomBytes` return type (`@atproto/crypto`)
    - Cast at consumer sites only where the producer is a third-party dep (`uint8arrays`, `@bufbuild/protobuf`) or DOM API (`crypto.subtle`)
    - Add `@note` to `BinaryBodyInit` explaining why `Uint8Array<ArrayBuffer>` is required
  - Fix `AbortSignal` return type on ozone `BackgroundQueue` / `PeriodicBackgroundTask` getters (TS6 + `@types/node@22` moved `AbortSignal` into an internal module that TS won't reference in emitted
   declarations).
  - Add `vite/client` types to `oauth-provider-ui` tsconfig for CSS import support.
  - Adapt eslint config for `@typescript-eslint` v8:
    - Keep `no-unused-expressions`, `no-require-imports`, `no-empty-object-type` enabled (fix violations inline)
    - Add `caughtErrorsIgnorePattern: "^_|^err$|^error$"` for catch params
    - Allow `no-require-imports` in `.cjs`/`.cts` files only
    - Remove stale `@typescript-eslint/ban-types` disable comments (rule removed in v8)
    - Fix bugs caught by `no-unused-expressions` (missing `return`, missing `()` on `.toBeTruthy`)
    - Prefix vars used only in type position with underscore in test files

  ### Breaking changes

  - Emitted `.d.ts` files now use TypeScript 6's stricter `Uint8Array<ArrayBuffer>` typing where Web/Node APIs require buffer-backed byte arrays. Consumers on older TypeScript should see no runtime impact, but may need to widen or cast in spots that previously relied on `Uint8Array` defaulting to `<ArrayBufferLike>`.


packages/lex/lex-cbor/tests/dag-cbor.test.ts
```diff
@@ -13,7 +13,8 @@ const bytes = {
     return data instanceof Uint8Array
   },
   toHex: (data: Uint8Array | ArrayBuffer): string => {
-    return Buffer.from(data).toString('hex')
+    const bytes = data instanceof Uint8Array ? data : new Uint8Array(data)
+    return Buffer.from(bytes).toString('hex')
```
matthieusieben, line 17: Nit (i realized after writing this that this is a test file)

`Buffer.from(bytes)` performs a copy. We can avoid this with:

```suggestion
    const bytes = data instanceof Uint8Array ? data : new Uint8Array(data)
    const buffer = bytes instanceof Buffer ? buffer: Buffer.from(bytes.buffer, bytes.byteOffset, bytes.byteLength)
    return buffer.toString('hex')
```

packages/lex/lex-client/src/types.ts
```diff
@@ -47,12 +47,15 @@ export type Service = `${DidString}#${DidServiceIdentifier}`
  * await client.xrpc(uploadMethod, { body: file })
  * ```
  */
+// Uint8Array is parameterized with ArrayBuffer (not ArrayBufferLike) because
+// fetch's BodyInit requires ArrayBuffer-backed views — SharedArrayBuffer is not
+// supported for network I/O.
```
matthieusieben: This could be a `@note` in the JSDoc

packages/oauth/jwk-jose/src/util.ts
```diff
@@ -1,4 +1,3 @@
-// eslint-disable-next-line @typescript-eslint/ban-types
```
matthieusieben, line 1: Why is this rule no longer needed ? It was enforcing the use of better patterns.

```suggestion
export type Simplify<T> = { [K in keyof T]: T[K] } & NonNullable<unknown>
```

> [!NOTE]
> I sometimes disable it when the patterns I use are well-known ones (like in this instance), or when `NonNullable<unknown>` is too verbose.
> 
> We _could_ also special case `{}` in the `@typescript-eslint/ban-types` config (if that is even possible)

packages/oauth/oauth-provider-ui/src/components/utils/admonition.tsx
```diff
@@ -17,8 +17,8 @@ import {
 import { Override } from '#/lib/util.ts'
 import { Button, ButtonProps } from '../forms/button.tsx'
 
-const VARIANTS = ['info', 'warning', 'error', 'success'] as const
-type Variant = (typeof VARIANTS)[number]
+const _VARIANTS = ['info', 'warning', 'error', 'success'] as const
+type Variant = (typeof _VARIANTS)[number]
```
matthieusieben: This might have been used as a runtime value at a point. It might be better to remove the runtime definition:

```suggestion
type Variant = 'info' | 'warning' | 'error' | 'success'
```

.eslintrc
```diff
@@ -75,22 +80,30 @@
     },
     {
       "files": ["*.js", "*.cjs"],
-      "rules": {
-        "@typescript-eslint/no-var-requires": "off"
-      }
+      "rules": {}
     },
```
matthieusieben: We might want to keep the explicit rule for `.cjs` files?

```suggestion
    {
      "files": ["*.cjs", "*.cts"],
      "rules": {
        "@typescript-eslint/no-var-requires": "off"
      }
    },
```

.eslintrc
```diff
@@ -51,9 +51,14 @@
...
         "argsIgnorePattern": "^_",
         "varsIgnorePattern": "^_",
-        "ignoreRestSiblings": true
+        "caughtErrorsIgnorePattern": "^_|^err|^error|^e$|^cause$",
+        "ignoreRestSiblings": true,
+        "reportUsedIgnorePattern": false
       }
     ],
+    "@typescript-eslint/no-unused-expressions": "off",
+    "@typescript-eslint/no-require-imports": "off",
```
matthieusieben: I would prefer keeping this rule if possible (with an exception for `.cjs` / `.cts` files, see below)

```suggestion
```

.eslintrc
```diff
@@ -51,9 +51,14 @@
...
+    "@typescript-eslint/no-empty-object-type": "off",
```
matthieusieben: I'd prefer keeping this rule as `{}` is misleading. I prefer the more explicit `NonNullable<unknown>`.

```suggestion
```

> [!NOTE]
> Note that I personally *do* sometimes use the `{}` type, but mostly in places where it is a common/well-known pattern (see other comment)

.eslintrc
```diff
@@ -51,9 +51,14 @@
       {
         "argsIgnorePattern": "^_",
         "varsIgnorePattern": "^_",
-        "ignoreRestSiblings": true
+        "caughtErrorsIgnorePattern": "^_|^err|^error|^e$|^cause$",
+        "ignoreRestSiblings": true,
+        "reportUsedIgnorePattern": false
       }
     ],
+    "@typescript-eslint/no-unused-expressions": "off",
```
matthieusieben: I'd prefer if we removed unused expression (since they are currently mostly used to build inferred types), or if we locally disabled this rule, rather than disabling it globally.

.eslintrc
```diff
@@ -51,9 +51,14 @@
       {
         "argsIgnorePattern": "^_",
         "varsIgnorePattern": "^_",
-        "ignoreRestSiblings": true
+        "caughtErrorsIgnorePattern": "^_|^err|^error|^e$|^cause$",
```
matthieusieben: - `^err` and `^error` are redundant. I'd prefer being more explicit (`^err$|^error$`)
- I don't like that we allow to silently ignore errors. I'd prefer that we use the `try {} catch {}` syntax to make it explicit, or if we allowed to explicitly ignore with the same logic as for vars (`_` prefix):

```suggestion
        "caughtErrorsIgnorePattern": "^_",
```

.eslintrc
```diff
@@ -75,22 +80,30 @@
...
+        "@typescript-eslint/no-unused-vars": [
+          "error",
+          {
+            "argsIgnorePattern": "^_",
+            "varsIgnorePattern": "^_",
+            "caughtErrorsIgnorePattern": "^_|^err|^error|^e$|^cause$",
+            "ignoreRestSiblings": true,
+            "reportUsedIgnorePattern": false
+          }
         ]
```
matthieusieben, line 98: See other comment

matthieusieben (approved):
Looking good. I'd rather that we don't relax linting rules too much if possible.

packages/lex/lex-cbor/tests/dag-cbor.test.ts
devinivy: Given that this is just a test suite (as you mentioned) I am inclined to keep the easier-to-verify implementation here.

packages/oauth/jwk-jose/src/util.ts
devinivy: This particular rule no longer exists in the updated version of typescript-eslint, was replaced with several more specific rules such as `no-empty-object-type`, `no-unsafe-function-type`, `no-wrapper-object-types`.  I'll put `no-empty-object-type` back in place (I think it was the only related rule that was disabled), though I kind of think in the long haul we should just stick with `NonNullable<unknown>` everywhere if we want this rule.

.eslintrc
devinivy: I'll bring it back, but I'd advocate we aim to be consistent about usage if we're going to include it.  For now I'll keep the line ignores so we don't need to get blocked on it 👍

.eslintrc
devinivy: > ^err and ^error are redundant. I'd prefer being more explicit (^err$|^error$)

Sounds good, will look at how far that gets us 👍 

>I don't like that we allow to silently ignore errors. I'd prefer that we use the try {} catch {} syntax to make it explicit, or if we allowed to explicitly ignore with the same logic as for vars (_ prefix)

I agree, but updating the style isn't a goal here so I am going to err on the side of compatibility with our existing lint config.

packages/repo/src/types.ts
```diff
@@ -18,23 +18,23 @@ const cidSchema = z.unknown().transform((input, ctx): Cid => {
   return z.NEVER
 })
 
-const unsignedCommit = z.object({
+const _unsignedCommit = z.object({
```
matthieusieben, line 21: Should we leave a TODO to replace with a type declaration? There is no point in keeping this runtime behavior here...

matthieusieben (approved):

PR #4931, opened 2026-05-12
Upgrade packages to ESM-only

devinivy:
_Coming soon_

Waterfalls into #4930


packages/api/tests/errors.test.ts
```diff
@@ -1,9 +1,10 @@
+// @NOTE must be imported from here to match dev-env's built error types
+import { ComAtprotoServerCreateAccount } from '@atproto/api'
```
matthieusieben, line 2: Could we `export * from '@atproto/api'` from `dev-env` ?

packages/api/tests/errors.test.ts
```diff
@@ -1,9 +1,10 @@
...
 import { TestNetworkNoAppView } from '@atproto/dev-env'
-import { AtpAgent, ComAtprotoServerCreateAccount } from '..'
 
 describe('errors', () => {
   let network: TestNetworkNoAppView
-  let client: AtpAgent
+  let client: ReturnType<TestNetworkNoAppView['pds']['getAgent']>
```
matthieusieben, line 7: We should be able to use `AtpAgent` from `@atproto/api` here (or dev-env if it re-exports `*` from `@atproto/api`)

packages/api/tests/moderation-mutewords.test.ts
```diff
@@ -1,5 +1,5 @@
-import { RichText, mock, moderatePost } from '../src/'
-import { matchMuteWords } from '../src/moderation/mutewords'
+import { RichText, mock, moderatePost } from '../src//index.js'
```
matthieusieben, line 1: We could probably clean these double slashes easily ?

packages/bsky/src/api/age-assurance/util.ts
```diff
@@ -19,7 +19,7 @@ export function computeAgeAssuranceAccessOrThrow(
     regionCode?: string
     verifiedMinimumAge: number
   },
-) {
+): any {
```
matthieusieben, line 22: Could we return an actual type instead? Or at least leave a TODO

matthieusieben:
I wish the path rewrites where in another PR, separate from this one (possibly a noop on `main`). This is pretty hard to review in GH.

packages/api/tsconfig.build.json
```diff
@@ -5,5 +5,5 @@
     "outDir": "./dist",
     "noUnusedLocals": false
   },
-  "include": ["./src"]
+  "include": ["./src", "../../tsconfig/cjs-default-exports.d.ts"]
```
matthieusieben, line 8: Removing `"../../tsconfig/cjs-default-exports.d.ts"` here does not seem to make a difference when building.

packages/api/tsconfig.tests.json
```diff
@@ -6,5 +6,5 @@
     "noEmit": true,
     "noUnusedLocals": false
   },
-  "include": ["./tests"]
+  "include": ["./tests", "../../tsconfig/cjs-default-exports.d.ts"]
```
matthieusieben, line 9: Removing "../../tsconfig/cjs-default-exports.d.ts" here does not seem to make a difference when building.

packages/aws/tsconfig.build.json
```diff
@@ -4,5 +4,5 @@
     "rootDir": "./src",
     "outDir": "./dist"
   },
-  "include": ["./src"]
+  "include": ["./src", "../../tsconfig/cjs-default-exports.d.ts"]
```
matthieusieben, line 7: Removing "../../tsconfig/cjs-default-exports.d.ts" here does not seem to make a difference when building.

matthieusieben:
Instead of `include: ["./folder", "../../tsconfig/cjs-default-exports.d.ts"]`, it would be more accurate / cleaner to use: 

```json
{
  "compilerOptions": {
    "types": ["../../tsconfig/cjs-default-exports.d.ts"]
  },
  // keep "include" unchanged
}
```

I didn't try this myself but we might even add these `"type"` imports directly from the base tsconfig files (instead of every project).

The biggest downside of this approach is that the modules defined in that `d.ts` file will appear as being globally available (when they aren't). Eslint should catch this though.

matthieusieben (commented):
Still reviewing but here is my first batch of comments.

devinivy:
> I wish the path rewrites where in another PR, separate from this one (possibly a noop on main). This is pretty hard to review in GH.

@matthieusieben That makes sense.  I started pulling these out this evening, but wasn't able to complete it.  I can continue on that tomorrow if it's hampering review.

matthieusieben:
Do we need `--experimental-strip-types` in places where we call `node file.ts`? like in:


```dockerfile
CMD ["node", "--heapsnapshot-signal=SIGUSR2", "--enable-source-maps", "api.ts"]
```

devinivy:
> Do we need --experimental-strip-types in places where we call node file.ts? like in:

The services are all on node v24, so flag not necessary :) We can also probably update some of the package bin scripts, but will leave that as followup.

devinivy:
Closing in favor of #4942 and #4943.

PR #4936, opened 2026-05-12
Downgrade devEngines onFail to warn until we can upgrade pnpm

ds-boyce:
It seems pnpm prior to v11 calls `npm` under the hood for some operations, so we need to downgrade from `"onFail": "error"` to `"onFail": "warn"` until we can upgrade pnpm.

Confirmed by running `pnpm info @atproto/aws` locally:

* Fails with v8.15.9
* Succeeds with v11.1.1

Related GitHub issue: https://github.com/pnpm/pnpm/issues/9924#issuecomment-4297229243

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
   "devEngines": {
     "packageManager": {
       "name": "pnpm",
-      "onFail": "error"
+      "onFail": "warn"
     }
   },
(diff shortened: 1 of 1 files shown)


ds-boyce:
#4928 is a better fix.

PR #4942, opened 2026-05-15
Add .js import extensions, upgrade test suites, update test snapshots

devinivy:
This is the third in a suite of PRs to get us on node v22+, TypeScript 6, and with ESM-only packages. Waterfalls into #4930. Here's an overview of this PR:

  > **Reviewing tip:** The bulk of the diff is two mechanical commits — `.js` import extensions (1650 files) and snapshot format updates (49 files). These are the first and third commit respectively.  To review the substantive changes, you can filter those out.

  - Add `.js` extensions to all ~1650 relative imports across the codebase, required for Node16 module resolution.
  - Upgrade `jest` from `^28.1.2` to `^30.0.0` across all packages. Rename `jest.config.js` to `jest.config.cjs` in preparation for `type: module`.
  - Migrate `packages/bsky` and `packages/lexicon-resolver` from Jest to Vitest. Add explicit vitest imports, convert `jest.fn()`/`jest.useFakeTimers()` to
  `vi.fn()`/`vi.useFakeTimers()`, and fix `vi.mock('node:dns/promises')` to provide a `default` export for default imports.
  - Update 49 snapshot files for the new Jest 30 / Vitest format (`Object {}` → `{}`, `Array []` → `[]`, test name `>` separator).
  - Replace deprecated `toThrowError()` calls with `toThrow()`.
  - Fix `crypto/tests/did.test.ts` — the assertion was a no-op (`.toBeTruthy` without `()`) and the comparison was incorrect (`parseDidKey` returns decompressed bytes). Now
  properly imports `decompressPubkey` and asserts equality.
  - Exclude `bsky/tests/views/age-assurance-v2.test.ts` from vitest — its `vi.mock()` needs to intercept CJS `require()` in dist, which only works once the package has
  `type: module`. Re-enabled on the ESM branch.
  - Jest configs stay in CJS mode (no `extensionsToTreatAsEsm`). The ESM jest configuration is deferred to the follow-up branch where `type: module` is set.


matthieusieben (commented):
LGTM based on the description (I did not actually review all the noise 🤪)

packages/api/jest.config.cjs
```diff
@@ -1,7 +1,14 @@
 /** @type {import('jest').Config} */
 module.exports = {
   displayName: 'API',
-  transform: { '^.+\\.ts$': '@swc/jest' },
+  transform: {
+    '^.+\\.(t|j)s$': [
+      '@swc/jest',
+      { jsc: { transform: {} }, module: { type: 'es6' } },
+    ],
+  },
```
matthieusieben: Why do we need to transform `.js` here ?

I suspect this will slow tests down.

Maybe we should have two transforms?

```js
  transform: {
    '^.+\\.ts$': [
      '@swc/jest',
      // Not sure what we need here in this case
      { jsc: { transform: {} }, module: { type: 'es6' } },
    ],
    '^.+\\.js$': [
      // Not sure what we need here ?
    ],
  },
```

Or we just keep this as it is until we ditch jest. No need to get hung up on this.

packages/bsky/tests/views/verification.test.ts
```diff
@@ -252,10 +253,10 @@ describe('verification views', () => {
...
 
         const urlPrefixes = getExpectedUrisPrefixes()
-
-        expect(urlPrefixes.length).toBe(
-          profile.verification?.verifications.length ?? 0,
-        )
+        profile.verification &&
+          expect(urlPrefixes.length).toBe(
+            profile.verification.verifications.length,
+          )
```
matthieusieben: doesn't this create a lint error? We probably need an `if()` here.

packages/bsky/package.json
```diff
@@ -20,9 +20,9 @@
     "prebuild": "pnpm run codegen",
     "build": "tsc --build tsconfig.build.json",
     "start": "node --enable-source-maps dist/bin.js",
-    "test": "../dev-infra/with-test-redis-and-db.sh jest",
+    "test": "../dev-infra/with-test-redis-and-db.sh vitest run",
```
matthieusieben, line 23: 🥳

packages/bsky/vitest.config.ts
```diff
@@ -0,0 +1,8 @@
+import { defineProject } from 'vitest/config'
+
+export default defineProject({
+  test: {
+    testTimeout: 60000,
+    setupFiles: ['../../jest.setup.ts'],
```
matthieusieben: Although `jest.setup.ts` has nothing `jest` specific in it, this line definitely grinds my gears (importing `jest` setup from `vitest` config).

Claude tells me we can just do:

```js
  setupFiles: ['dotenv/config'],
```

But I don't know if the difference in file location matters...

packages/bsky/src/data-plane/client/util.test.ts
```diff
@@ -1,14 +1,16 @@
...
 describe('callerInterceptor', () => {
   it('sets x-atlantis-caller header on the request', async () => {
     const interceptor = callerInterceptor('appview')
     const expectedResponse = { status: 'ok' }
-    const next = jest.fn().mockResolvedValue(expectedResponse)
+    const next = vi.fn<NextFn>().mockResolvedValue(expectedResponse)
 
     const req = { header: new Headers() }
-    const handler = interceptor(next)
+    const handler = interceptor(next as any)
```
matthieusieben: All tests in this file are doing essentially the same thing.

The "best" casting method would probably be to properly type the request/response objects like:

```ts
import {
  StreamRequest,
  StreamResponse,
  UnaryRequest,
  UnaryResponse,
} from '@connectrpc/connect'
import { callerInterceptor } from './util.js'

describe('callerInterceptor', () => {
  it('sets x-atlantis-caller header on the request', async () => {
    const fakeRequest = { header: new Headers({ 'x-other': 'value' }) } as
      | UnaryRequest
      | StreamRequest
    const fakeResponse = {} as UnaryResponse | StreamResponse
    const fakeHandler = jest.fn(async (_res: UnaryRequest | StreamRequest) => {
      return fakeResponse
    })

    const interceptor = callerInterceptor('appview')
    const next = interceptor(fakeHandler)
    const res = await next(fakeRequest)

    expect(fakeRequest.header.get('x-atlantis-caller')).toBe('appview')
    expect(fakeRequest.header.get('x-other')).toBe('value')
    expect(fakeHandler).toHaveBeenCalledWith(fakeRequest)
    expect(res).toBe(fakeResponse)
  })
})
```

(and remove all other tests)

matthieusieben (commented):
lg

packages/bsky/vitest.config.ts
devinivy: The issue with `dotenv/config` is just that it pulls from a different env file.  I will address by renaming to `test.setup.ts` so that it's not jest-specific.

PR #4943, opened 2026-05-15
Switch to ESM: type: module, Node16 resolution, CJS interop

devinivy:
This is the fourth and final in a suite of PRs to get us on node v22+, TypeScript 6, and with ESM-only packages. Waterfalls into #4942. Here's an overview of this PR:

  - Upgrade ESM-only dependencies: `multiformats` ^9 → ^13, `uint8arrays` 3 → ^5, `p-queue` ^6 → ^8, `jose` ^4 → ^5.
  - Set `"type": "module"` and add `"exports"` to all 60+ package.json files.
  - Switch tsconfig base to `module: Node16`, `moduleResolution: node16`, `target: ES2022`. Set `useDefineForClassFields: false`.
  - Enable ESM mode in jest configs (`extensionsToTreatAsEsm: ['.ts']`, SWC `module: { type: 'es6' }`). Add `NODE_OPTIONS=--experimental-vm-modules` to test scripts. Add
  `jest.varint-shim.cjs` for varint CJS interop.
  - Convert service entrypoints from `.js` to TypeScript (bsky, bsync, ozone, pds).
  - Fix CJS default import interop throughout: `await-lock`, `key-encoder`, `typed-emitter`, `iso-datestring-validator`, `file-type`, `@hapi/accept`, `pg`.
  - Replace `__dirname` with `import.meta.dirname`, `require.resolve` with `import.meta.resolve`.
  - Fix `EventEmitter` import (`node:stream` → `node:events`).
  - Bundle handlebars runtime into PDS email templates (`build.templates.cjs`).
  - Use `jest.unstable_mockModule` for ESM-compatible mocking in PDS and identity tests.
  - Re-enable `bsky/tests/views/age-assurance-v2.test.ts` (now works with `type: module`).
  - Add `tsconfig/cjs-default-exports.d.ts` for packages that import CJS modules with broken default exports under Node16 resolution.
  - Remove `multiformats` pnpm patch (no longer needed with native ESM).
  - Add explicit return type annotations for kysely query builders (TS2883).
  - Address PR #4931 feedback: use `AtpAgent` type in errors test, fix double-slash imports, add proper return type to `age-assurance/util.ts`, remove unnecessary
  `cjs-default-exports.d.ts` from api/aws tsconfigs.
  - Add changeset for ESM migration.


matthieusieben:
Something we might consider after all this is done is to remove the bundling of the `lex-data` package and replace it with a dependency. (we needed bundling to be able to rely on ESM dependencies in from CJS).

packages/crypto/tests/signatures.test.ts
```diff
@@ -21,7 +21,9 @@ describe('signatures', () => {
 
   beforeAll(() => {
     vectors = JSON.parse(
-      fs.readFileSync(`${__dirname}/signature-fixtures.json`).toString(),
+      fs
```
matthieusieben: In other places we do the following, that seems more "canonical" (maybe?):

```ts
const __dirname = path.dirname(fileURLToPath(import.meta.url))
```

packages/dev-env/bin.js
```diff
@@ -1,12 +1,9 @@
...
-/* eslint-disable @typescript-eslint/no-var-requires */
-
-'use strict'
 
 // @NOTE This file exists so that pnpm can properly link the "bin" scripts when
 // the monorepo is being setup (during initial "pnpm install"), but was never
 // built.
 
-require('./dist/bin.js')
+import('./dist/bin.js')
```
matthieusieben: This would be better written as


```suggestion
import './dist/bin.js'
```

matthieusieben: (or be awaited but I don't like top level await)

packages/lex-cli/src/codegen/lex-gen.ts
```diff
@@ -50,7 +50,7 @@ export function genCommonImports(file: SourceFile, baseNsid: string) {
       moduleSpecifier: `${baseNsid
         .split('.')
         .map((_str) => '..')
-        .join('/')}/lexicons`,
+        .join('/')}/lexicons.js`,
```
matthieusieben, line 53: This might deserve it own changeset

packages/lex-cli/bin.js
```diff
@@ -1,12 +1,9 @@
...
-/* eslint-disable @typescript-eslint/no-var-requires */
-
-'use strict'
 
 // @NOTE This file exists so that pnpm can properly link the "bin" scripts when
 // the monorepo is being setup (during initial "pnpm install"), but was never
 // built.
 
-require('./dist/index.js')
+import('./dist/index.js')
```
matthieusieben:
```suggestion
import './dist/index.js'
```

(won't comment further on this particular suggestion if other instances of this appear in the PR)

packages/lex/lex-builder/package.json
```diff
@@ -27,15 +27,13 @@
...
   "sideEffects": false,
-  "type": "commonjs",
-  "main": "./dist/index.js",
-  "types": "./dist/index.d.ts",
+  "type": "module",
   "exports": {
     ".": {
       "types": "./dist/index.d.ts",
       "browser": "./dist/index.js",
       "require": "./dist/index.js",
```
matthieusieben: I believe these are no longer needed:

```suggestion
```

packages/lex/lex-cbor/package.json
matthieusieben: We probably want to either remove the bundling (`vite.config.mts`), or replace `default` with `require`:

```suggestion
"require": "./dist/index.cjs",
"default": "./dist/index.mjs"
```

packages/lex/lex-document/src/index.ts
```diff
@@ -1,6 +1,3 @@
-import 'core-js/modules/esnext.symbol.async-dispose'
-import 'core-js/modules/esnext.symbol.dispose'
-
 export * from './lexicon-document.js'
```
matthieusieben, line 4: `Symbol.dispose` is not yet baseline. We should keep the polyfill.


https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/dispose

packages/lex/lex-data/src/core-js.d.ts
```diff
@@ -0,0 +1,2 @@
+declare module 'core-js/modules/es.uint8-array.from-base64.js'
+declare module 'core-js/modules/es.uint8-array.to-base64.js'
```
matthieusieben: We should use `@types/core-js` instead.

matthieusieben: On second look it seems that package does not declare the module we need.

matthieusieben: Note that the `core-js/module/...` should be re-written as e.g.:

```
import 'core-js/es/typed-array/from-base64'
```

packages/oauth/oauth-client-browser/src/core-js.d.ts
```diff
@@ -0,0 +1,2 @@
+declare module 'core-js/modules/esnext.symbol.async-dispose'
+declare module 'core-js/modules/esnext.symbol.dispose'
```
matthieusieben: Can we use `@types/core-js` instead?

packages/oauth/oauth-provider-ui/vite.config.mjs
```diff
@@ -13,6 +13,7 @@ export default defineConfig({
     alias: {
       '#': resolve(__dirname, './src'),
     },
+    conditions: ['import', 'module', 'browser', 'default'],
```
matthieusieben: Browser should probably be first in that list!

packages/lex-cli/src/codegen/lex-gen.ts
devinivy: Added 👍 Luckily I think this should be backwards-compatible in many environments.  We can add settings to control this if needed.

packages/lex/lex-cbor/package.json
```diff
@@ -29,31 +29,23 @@
...
     }
   },
   "dependencies": {
     "@atproto/lex-data": "workspace:^",
+    "cborg": "^4.5.8",
     "tslib": "^2.8.1"
   },
   "devDependencies": {
     "@atproto/lex-json": "workspace:^",
-    "cborg": "^4.5.8",
```
devinivy, line 50: Any idea why this was originally a dev dep @matthieusieben?  I assume it was likely unintentional, but making sure.

matthieusieben: this was dev dep because the whole thing was getting bundled. Should be a normal dep now indeed.

packages/pds/tests/handles.test.ts
```diff
@@ -23,12 +22,16 @@ jest.mock('node:dns/promises', () => {
   }
 })
 
+const { IdResolver } = await import('@atproto/identity')
+const { TestNetworkNoAppView } = await import('@atproto/dev-env')
+const { default: basicSeed } = await import('./seeds/basic.js')
```
matthieusieben, line 30: This is super hacky (and introduces top level await) for no obvious reason. A comment would be welcome.

matthieusieben: note that if the reason is that we need `jest.unstable_mockModule('node:dns/promises',` to run before, a cleaner solution would be to move the mock to its own file and import it first from here:

```ts
// @NOTE must be first
import "./_mock_node_dns_promise.js"

// other imports
```

.changeset/lex-cli-js-extensions.md
```diff
@@ -0,0 +1,6 @@
+---
+'@atproto/lex-cli': minor
+---
+
+**BREAKING:** Generated lexicon code now uses `.js` extensions on relative imports (e.g. `from '../lexicons.js'` instead of `from '../lexicons'`), required for Node16 module resolution under ESM. Consumers of `lex-cli` will need to use `moduleResolution: "node16"` (or equivalent) for the generated code to compile.
```
matthieusieben: > Consumers of `lex-cli` will need to use `moduleResolution: "node16"` (or equivalent) for the generated code to compile.

This is not true. Importing using the exact `.js` file path also works with older moduleResolution .

packages/lex/lex-document/src/index.ts
devinivy: Addressed.  Looks like this now:
```diff
-import 'core-js/modules/esnext.symbol.async-dispose'
-import 'core-js/modules/esnext.symbol.dispose'
+import 'core-js/es/symbol/async-dispose.js'
+import 'core-js/es/symbol/dispose.js'
```

packages/lex/lex-data/src/core-js.d.ts
devinivy: Addressed 👍 Looks like this now:
```ts
declare module 'core-js/es/typed-array/from-base64.js'
declare module 'core-js/es/typed-array/to-base64.js'
```

(The `.js` were indeed needed.)

packages/oauth/oauth-provider-ui/vite.config.mjs
devinivy: Addressed

packages/oauth/oauth-client-browser/src/core-js.d.ts
devinivy: I think we determined elsewhere @types/core-js wouldn't work for us since it is out of date (still for v2 of core-js).

packages/lex/lex-cbor/package.json
devinivy: Oh right, right.

.changeset/lex-cli-js-extensions.md
devinivy: I think I already fixed this.  It reads:
> Generated lexicon code now uses `.js` extensions on relative imports (e.g. `from '../lexicons.js'` instead of `from '../lexicons'`), enabling compatibility with Node16/NodeNext module resolution under ESM. Consumers using classic `node` or `bundler` resolution should be unaffected.

packages/pds/tests/handles.test.ts
devinivy: Taking another look.

packages/lex/lex-cbor/package.json
devinivy: Removed the bundling 👍

ThisIsMissEm:
For mocking, since these tests almost always run in the context of node.js, instead of:
> Use jest.unstable_mockModule for ESM-compatible mocking in PDS and identity tests.

It may be better to look at native node.js mocking functionality, which should work independently of the test runner: https://nodejs.org/learn/test-runner/mocking

packages/pds/tests/handles.test.ts
devinivy: It turns out that doesn't do the trick, but I am quite sure vitest addresses the issue at play here.  Added a comment.

antoinerousseau:
I'm still getting this warning with React Native with @atproto/common-web v0.5.1, any clue?

>  WARN  Attempted to import the module "node_modules/.pnpm/multiformats@13.4.2/node_modules/multiformats/dist/src/hashes/sha2-browser.js" which is not listed in the "exports" of "node_modules/.pnpm/multiformats@13.4.2/node_modules/multiformats" under the requested subpath "./dist/src/hashes/sha2-browser.js". Falling back to file-based resolution. Consider updating the call site or asking the package maintainer(s) to expose this API.

PR #4950, opened 2026-05-18
use dotenv/config from vitest file

matthieusieben:


packages/bsky/package.json
```diff
@@ -51,6 +51,7 @@
     "compression": "^1.7.4",
     "cors": "^2.8.5",
     "disposable-email-domains-js": "^1.5.0",
+    "dotenv": "^16.0.3",
```
devinivy, line 54: I think we probably want this added as a dev dep.

PR #4954, opened 2026-05-19
Improve support for `ArrayBufferLike`-backed `Uint8Array` data

matthieusieben:
This relaxes some of the constraints introduced by the recent TypeScript 6 migration (and release through #4952)


packages/oauth/oauth-client/src/runtime-implementation.ts
```diff
@@ -10,7 +10,7 @@ export type DigestAlgorithm = { name: 'sha256' | 'sha384' | 'sha512' }
 export type RuntimeDigest = (
   data: Uint8Array<ArrayBuffer>,
```
matthieusieben, line 11: The generic oauth client will provide a strongly typed `Uint8Array<ArrayBuffer>` to the "runtime implementation", while it does not require the implementation to return a strongly typed `Uint8Array`. This allows e.g. to compute the digest from a worker thread.

devinivy (approved):

PR #4959, opened 2026-05-19
Fix compilation of protobuf files

matthieusieben:
Builds on top of https://github.com/bluesky-social/atproto/pull/4960

This PR ensures that `codegen` always runs during the release process (`publish.yaml` workflow). This will make sure that released code always uses up-to-date generated code.


matthieusieben:
moved to
- https://github.com/bluesky-social/atproto/pull/4964
- https://github.com/bluesky-social/atproto/pull/4965

PR #4967, opened 2026-05-20
Update Typescript to v7

matthieusieben:
- Replace `typescript@^6.0.3` with `@typescript/native-preview@^7.0.0-beta` in the root.
  - Every package's `build` script becomes `tsgo --build tsconfig.build.json`. Per-package typescript devDeps are dropped — `tsgo` is hoisted at the root only (Remove TS6 dependency commit).
  - Trivial code adjustments scattered across `.ts` files are forced by stricter TS7 emit / module semantics (the bulk of the line count in the diff stat — they're all the same kind of change, no behavioral effect).


packages/oauth/oauth-provider/src/account/account-store.ts
```diff
@@ -228,22 +228,22 @@ export interface AccountStore {
...
   'authenticateAccount',
-  'setAuthorizedClient',
+  'createAccount',
   'getAccount',
-  'upsertDeviceAccount',
   'getDeviceAccount',
-  'removeDeviceAccount',
   'listDeviceAccounts',
-  'resetPasswordRequest',
+  'removeDeviceAccount',
```
devinivy, line 236: What is going on with the `buildInterfaceChecker` usage updates?

matthieusieben: `tsgo` and `tsc` have different internal ordering of object properties. `tsgo` uses alphabetical, while `tsc` uses definition order (I think?).

The "interface checker" utility here relies on that internally defined ordering when listing the needed properties to satisfy an interface.

devinivy (approved):
Looks good to me.  Only questions on my mind which can't be addressed directly through code review are:
- Are there any interop issues with the emitted output?  E.g. will the output be compatible with folks using our libraries under TS6?
- Sort of related— the `buildInterfaceChecker()` changes caught my attention.  I don't fully understand them, but they point to potential interop sensitivities.

matthieusieben:
The emitted output should be usable with TS version < 6.

`buildInterfaceChecker` is not part of the public interface so this should not be a issue. This is a compile time implementation detail.

Capture d’écran 2026-06-15 à 22 40 43

PR #4969, opened 2026-05-20
Log rate limiter error in the context of the HTTP `req`

matthieusieben:
Currently, any error occurring while computing the rate limit (e.g. Redis connection error) are logged using a global `logger`.

This PR changes that by using the `req.log` if it is a Pino logger, so that logs contain the request context (url, headers, etc.)

### DIFF
--- a/packages/xrpc-server/src/rate-limiter.ts
+++ b/packages/xrpc-server/src/rate-limiter.ts
@@ -1,4 +1,3 @@
-import { IncomingMessage, ServerResponse } from 'node:http'
 import {
   RateLimiterAbstract,
   RateLimiterMemory,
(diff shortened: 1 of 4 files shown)


packages/xrpc-server/src/rate-limiter.ts
```diff
@@ -263,86 +255,6 @@ export class CombinedRateLimiter<
...
-    if (resp instanceof RateLimitExceededError) return resp
-    if (lowest === null || resp.remainingPoints < lowest.remainingPoints) {
-      lowest = resp
-    }
-  }
-  return lowest
-}
-
-export type RouteRateLimiterOptions<
-  C extends RateLimiterContext = RateLimiterContext,
```
matthieusieben, line 281: This was moved mostly unchanged into `rate-limiter-http.ts` to increase separation of concerns in this file.

packages/xrpc-server/src/server.ts
```diff
@@ -806,3 +826,19 @@ const defaultPoints: CalcPointsFn = () => 1
...
  */
 const defaultKey: CalcKeyFn<HandlerContext> = ({ req }) => req.ip
+
+async function rateLimiterLoggerErrorHandler(
+  err: unknown,
+  ctx: HandlerContext,
+  { limiter: { keyPrefix, points, duration } }: RateLimiterErrorHandlerDetails,
+): Promise<null> {
+  const { req } = ctx
+  const logger = isPinoHttpRequest(req) ? req.log : log
```
matthieusieben, line 836: This is the heart of the change.

PR #4972, opened 2026-05-20
Fix lex SDL files being non tree-shakable

matthieusieben:
Due to the helpers exported by `lex` code generation, the code ends up being not as tree-shakable as expected. Indeed accessing an object's property cannot be marked as `/*#__PURE__*/` resulting in the object being kept around.

```ts
export const $lxm = main.nsid;
export const $params = main.parameters;
export const $input = main.input;
export const $output = main.output;
```

This PR adresses that in different ways:

- For XRPC methods (procedure, query, subscription), instead of exporting the input/output by reading them from the object, we create them as standalone exported variables, which are then used to build the method.
- "main" validator (TBD)

### DIFF
--- a/.changeset/grumpy-bags-thank.md
+++ b/.changeset/grumpy-bags-thank.md
@@ -1,0 +1,5 @@
+---
+'@atproto/lex-builder': patch
+---
+
+Fix XRPC methods being non tree-shakable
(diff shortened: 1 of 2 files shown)


matthieusieben:
## Method example

Before (notice hilighted text)
Capture d’écran 2026-05-20 à 15 49 58


After
Capture d’écran 2026-05-20 à 15 50 57

devinivy (approved):

PR #4988, opened 2026-05-21
email templates: apply branding env, optional bsky link removal

knotbin:
In pursuit of a PDS that is brand and app agnostic.
### Changes made to all email templates
- applies existing branding env vars to all email templates. 
  - previously hard-coded Bluesky branding is now just the default.

### Changes specific to the email confirmation template (not PLC operations, account deletion, etc.)
- adds a new `PDS_EMAIL_DISABLE_CONFIRMATION_LINK` env var that disables the bsky.app verification link in email confirmation emails. 
  - Enabling this _**does NOT**_ remove the ability to verify your email inside bsky.app. The confirmation code can still be used to confirm an email in bsky.app.
- adds the verification code to the email description to ensure the email is compatible with Apple's confirmation code AutoFill. 
  - This is specifically **_ONLY_** applied to the email confirmation template because all other emails with confirmation codes are destructive actions where users should read the email and fully understand the action


matthieusieben (approved):
This is looking great. I'll need a second pass before merging but I like it!

matthieusieben:
Merged trhough https://github.com/bluesky-social/atproto/pull/5009

PR #4989, opened 2026-05-21
PDS: bump request JSON size limit on record methods

dholms:
Records can be up to 1MB when CBOR-encoded. Currently the record methods on the PDS inherit the baseline json size limit of 150kb.


devinivy (approved):

uwx:
Shouldn't the limit be a little bit higher, since CBOR takes up less space than JSON?

PR #4992, opened 2026-05-21
Fix OAuth service-proxying audience for combined `did#serviceId` scopes

### LINKED ISSUE #4850: Bluesky permission sets require inconsistent aud parameter
**Describe the bug**

In trying to rewrite [sill.social](https://sill.social) to use granular permission sets, I've found that `app.bsky.authViewAll` does not allow access to `app.bsky.feed.getFeedGenerator`, `app.bsky.graph.getLists`, and others. Note that these are available, with `inheritAud: true`, [on the `app.bsky.authViewAll` permission set](https://lexicon.garden/lexicon/did:plc:4v4y5r3lwsbtmsxhile2ljac/app.bsky.authViewAll).

```
Could not find feed at://did:plc:z72i7hdynmk6r22z27h6tvur/app.bsky.feed.generator/whats-hot XRPCError: Missing required scope "rpc:app.bsky.feed.getFeedGenerator?aud=did:web:api.bsky.app"
```

My scopes are as follows:

```
  "scope": "atproto include:app.bsky.authViewAll?aud=did:web:api.bsky.app%23bsky_appview repo:community.lexicon.bookmarks.bookmark",
```

Other `get...` calls are working with these scopes, such as `app.bsky.actor.getProfile` and `app.bsky.feed.getTimeline`.

**To Reproduce**

I put together a minimal reproduction [here](https://tangled.org/tylerjfisher.com/repro-bluesky-aud-bug).

**Expected behavior**

I would expect all XRPC calls within the `authViewAll` permission set to work with the bsky appview specified.

**Details**

- Operating system: macOS 26.2
- Node version: 22.15.1

devinivy:
Implements the first chunk of phase 1 of the [Service Auth Audience Revised proposal](https://github.com/bluesky-social/proposals/blob/main/0014-service-auth-revised/README.md) and resolves #4850.

OAuth scopes for service-proxying use `rpc:<lxm>?aud=<did>#<serviceId>`, where `aud` has the combined form including a `serviceId`. The PDS scope check was passing the _bare DID_ `<did>` as `aud`, so no granted scope ever matched and proxied calls failed for OAuth callers.

This PR aligns the scope-check audience to combined `did#serviceId` form across the relevant PDS code paths. Note that **service-auth JWT audiences are unchanged** — they continue to use bare DID, matching what receivers expect today. This asymmetry is deliberate in phase 1 and documented inline.

As part of this phase the `com.atproto.server.getServiceAuth` lexicon was also loosened to permit `aud` with the combined `did#serviceId` form.  OAuth callers can now request tokens with combined `aud` in line with their `rpc:<lxm>?aud=<did>#<serviceId>` permissions.

A future config flag (phase 2 of [the proposal](https://github.com/bluesky-social/proposals/blob/main/0014-service-auth-revised/README.md)) will allow flipping the proxy-path JWT `aud` to the combined form.

Along the way `@atproto/did` gained a couple new types: `AtprotoDidRefAbsolute` (renamed from `AtprotoAudience`, deprecated alias retained) e.g. `'did:web:api.bsky.app#bsky_appview'`; and `DidRefRelative` e.g. `'#atproto'` with an optional id-narrowing parameter (sets up `kid`-based signing-key resolution for the next part of phase 1).  `@atproto/oauth-scopes` migrated to the new canonical names.

### DIFF
--- a/packages/did/src/did-ref.ts
+++ b/packages/did/src/did-ref.ts
@@ -15,6 +15,7 @@ export const isAtprotoDidRefAbsolute = (
   if (typeof value !== 'string') return false
   const hashIndex = value.indexOf('#')
   if (hashIndex === -1) return false
+  if (hashIndex === value.length - 1) return false
   if (value.indexOf('#', hashIndex + 1) !== -1) return false
   return (
     isFragment(value, hashIndex + 1) &&
--- a/packages/did/src/did-ref.ts
+++ b/packages/did/src/did-ref.ts
@@ -18,8 +18,7 @@ export const isAtprotoDidRefAbsolute = (
   if (hashIndex === value.length - 1) return false // empty fragment
   if (value.indexOf('#', hashIndex + 1) !== -1) return false // more than one '#'
   return (
-    isFragment(value, hashIndex + 1) &&
-    isAtprotoDid(value.slice(0, hashIndex))
+    isFragment(value, hashIndex + 1) && isAtprotoDid(value.slice(0, hashIndex))
   )
 }
 
@@ -57,6 +56,5 @@ export type AtprotoTokenAud = AtprotoDid | AtprotoDidRefAbsolute
 export type AtprotoAudience = AtprotoDidRefAbsolute
 
 /** @deprecated use {@link isAtprotoDidRefAbsolute}. */
-export const isAtprotoAudience: (
-  value: unknown,
-) => value is AtprotoAudience = isAtprotoDidRefAbsolute
+export const isAtprotoAudience: (value: unknown) => value is AtprotoAudience =
+  isAtprotoDidRefAbsolute
--- a/packages/did/src/did-ref.ts
+++ b/packages/did/src/did-ref.ts
@@ -14,9 +14,9 @@ export const isAtprotoDidRefAbsolute = (
 ): value is AtprotoDidRefAbsolute => {
   if (typeof value !== 'string') return false
   const hashIndex = value.indexOf('#')
-  if (hashIndex === -1) return false
-  if (hashIndex === value.length - 1) return false
-  if (value.indexOf('#', hashIndex + 1) !== -1) return false
+  if (hashIndex === -1) return false // no '#'
+  if (hashIndex === value.length - 1) return false // empty fragment
+  if (value.indexOf('#', hashIndex + 1) !== -1) return false // more than one '#'
   return (
     isFragment(value, hashIndex + 1) &&
     isAtprotoDid(value.slice(0, hashIndex))
@@ -34,11 +34,11 @@ export function isDidRefRelative<I extends string = string>(
   id?: I,
 ): value is DidRefRelative<I> {
   if (typeof value !== 'string') return false
-  if (value.charCodeAt(0) !== 35 /* '#' */) return false
-  if (value.length < 2) return false
-  if (value.indexOf('#', 1) !== -1) return false
+  if (value.charCodeAt(0) !== 35 /* '#' */) return false // doesn't start with '#'
+  if (value.length < 2) return false // empty fragment
+  if (value.indexOf('#', 1) !== -1) return false // more than one '#'
   if (!isFragment(value, 1)) return false
-  if (id !== undefined && value !== `#${id}`) return false
+  if (id !== undefined && value !== `#${id}`) return false // id mismatch
   return true
 }
 
--- a/packages/did/src/did-ref.ts
+++ b/packages/did/src/did-ref.ts
@@ -3,9 +3,6 @@ import { isFragment } from './lib/uri.js'
 
 /**
  * An atproto-constrained absolute DID reference: `${AtprotoDid}#${fragment}`.
- *
- * This is the same shape as the legacy `AtprotoAudience` type, kept under a
- * more accurate name now that the union is reused for non-audience contexts.
  */
 export type AtprotoDidRefAbsolute = `${AtprotoDid}#${string}`
 
@@ -40,21 +37,3 @@ export function isDidRefRelative<I extends string = string>(
   if (id !== undefined && value !== `#${id}`) return false // id mismatch
   return true
 }
-
-/**
- * The shape of `aud=...` values appearing in OAuth scopes — always combined.
- */
-export type AtprotoScopeAud = AtprotoDidRefAbsolute
-
-/**
- * The shape of `aud` claims in atproto service-auth JWTs — bare DID or
- * combined absolute reference. Phase 1 accepts both.
- */
-export type AtprotoTokenAud = AtprotoDid | AtprotoDidRefAbsolute
-
-/** @deprecated use {@link AtprotoScopeAud} or {@link AtprotoDidRefAbsolute}. */
-export type AtprotoAudience = AtprotoDidRefAbsolute
-
-/** @deprecated use {@link isAtprotoDidRefAbsolute}. */
-export const isAtprotoAudience: (value: unknown) => value is AtprotoAudience =
-  isAtprotoDidRefAbsolute
--- a/packages/did/src/did-ref.ts
+++ b/packages/did/src/did-ref.ts
@@ -1,27 +1,25 @@
-import { AtprotoDid, isAtprotoDid } from './atproto.js'
+import { Did, isDid } from './did.js'
 import { isFragment } from './lib/uri.js'
 
 /**
- * An atproto-constrained absolute DID reference: `${AtprotoDid}#${fragment}`.
+ * An absolute DID reference: `${Did}#${fragment}`.
+ *
+ * @see {@link https://www.w3.org/TR/did-core/#did-url-syntax}
  */
-export type AtprotoDidRefAbsolute = `${AtprotoDid}#${string}`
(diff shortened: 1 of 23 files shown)


packages/did/src/did-ref.ts
```diff
@@ -0,0 +1,62 @@
...
+  if (typeof value !== 'string') return false
+  const hashIndex = value.indexOf('#')
+  if (hashIndex === -1) return false // no '#'
+  if (hashIndex === value.length - 1) return false // empty fragment
+  if (value.indexOf('#', hashIndex + 1) !== -1) return false // more than one '#'
+  return (
+    isFragment(value, hashIndex + 1) &&
+    isAtprotoDid(value.slice(0, hashIndex))
+  )
+}
```
matthieusieben: nit: in it's current form, this package is atproto agnostic, except for the `atproto.ts` file.

I would recommend keeping it that way by keeping low level building blocks in here, and exporting `*Atproto*` definitions there.

Here is an illustration of what I mean.

```ts
// did-ref.ts


export function parseDidRefAbsolute(value: unknown): 
  | null
  | { did: Did; fragment: Fragment }
{
  const parsed = parseDidRef(value)
  if (!parsed) return null
  if (!parsed.did) return null
  return parsed
}

export type DidRefAbsolute<M extends string> = `${Did<M>}${Fragment}`
export function isDidRefAbsolute(value: unknown): value is DidRefAbsolute {
  return parseDidRefAbsolute(value) != null
}


export function parseDidRef(value: unknown):
  | null
  | { did?: Did; fragment: Fragment }
{
  // ...
}


export type DidRef<M extends string> = Fragment | DidRefAbsolute<M>
export function isDidRef(value: unkown): value is DidRef {
  return parseDidRef(value) != null
}

```

```ts
// atproto.ts

export function parseAtprotoDidRefAbsolute(value: unknown):
  | null
  | { did: AtprotoDid; fragment: Fragment }
{
  const parsed = parseDidRefAbsolute(value)
  if (!parsed) return null
  if (!isAtprotoDid(parsed.did)) return null
  return parsed
}

export type AtprotoDidRefAbsolute<M extends AtprotoIdentityDidMethods = AtprotoIdentityDidMethods> = DidRefAbsolute<M>
export function isAtprotoDidRefAbsolute(value: unknown): value is AtprotoDidRefAbsolute {
  return atprotoAtprotoDidRefAbsolute(value) != null
}
```

packages/did/src/did-ref.ts
```diff
@@ -0,0 +1,62 @@
...
+  if (value.indexOf('#', 1) !== -1) return false // more than one '#'
+  if (!isFragment(value, 1)) return false
+  if (id !== undefined && value !== `#${id}`) return false // id mismatch
+  return true
+}
+
+/**
+ * The shape of `aud=...` values appearing in OAuth scopes — always combined.
+ */
+export type AtprotoScopeAud = AtprotoDidRefAbsolute
```
matthieusieben: I feel like this might be a good opportunity to remove those OAuth related definitions from `@atproto/did`.

Since `@atproto/did` is being re-written to declare lower level utilities (`isDidRefRelative`, `isAtprotoDidRefAbsolute`), I think it'd make sense to remove non did specific definitions from this package, especially since it helps consolidate OAuth related concepts under `@atproto/oauth-scopes`.


Note: `@atproto/oauth-scopes` is the current only consumer of `AtprotoAudience`. I feel totally fine creating a breaking change in this package by removing that deprecated value. Especially since the migration path for consumer would just be to import from `@atproto/oauth-scopes` instead (it just needs a clear changeset). If we don't want breaking change, we can also export old names as deprecated aliases.

packages/did/src/did-ref.ts
```diff
@@ -0,0 +1,62 @@
...
+export type AtprotoDidRefAbsolute = `${AtprotoDid}#${string}`
+
+export const isAtprotoDidRefAbsolute = (
+  value: unknown,
+): value is AtprotoDidRefAbsolute => {
+  if (typeof value !== 'string') return false
+  const hashIndex = value.indexOf('#')
+  if (hashIndex === -1) return false // no '#'
+  if (hashIndex === value.length - 1) return false // empty fragment
+  if (value.indexOf('#', hashIndex + 1) !== -1) return false // more than one '#'
```
matthieusieben: Nitnit: ~I *think* that, under the hood, `includes` uses more preformant CPU instructions.~ 

Claude tells me they both use `memchr` for single char lookups.

```suggestion
  if (value.includes('#', hashIndex + 1)) return false // more than one '#'
```

packages/oauth/oauth-scopes/src/scopes/rpc-permission.ts
```diff
@@ -1,18 +1,21 @@
...
+  isAtprotoDidRefAbsolute,
+} from '@atproto/did'
 import { Nsid, isNsid } from '../lib/nsid.js'
 import { Parser } from '../lib/parser.js'
 import { ResourcePermission } from '../lib/resource-permission.js'
 import { ScopeStringSyntax } from '../lib/syntax-string.js'
 import { NeRoArray, ScopeSyntax, isScopeStringFor } from '../lib/syntax.js'
 
-export { type AtprotoAudience, type Nsid, isAtprotoAudience, isNsid }
+export { type AtprotoScopeAud, type Nsid, isAtprotoDidRefAbsolute, isNsid }
```
matthieusieben: I like the current pattern of using a type `X` and it's assertion function `isX`. It allows to code to be more symetric & readable.


Image

vs

Image

I find it much less load bearing when reading the code (we don't have to remember that, under the hook, `AtprotoScopeAud` is actually a `AtprotoDidRefAbsolute`)


In the same way we alias `AtprotoScopeAud` to `AtprotoDidRefAbsolute`, could we also alias the assertion functions ?

matthieusieben (approved):
This looks great!

Left a few comments about minor architectural changes.

packages/pds/src/api/app/bsky/actor/getPreferences.ts
```diff
@@ -13,8 +17,8 @@ export default function (server: Server, ctx: AppContext) {
       additional: [AuthScope.Takendown],
       authorize: (permissions, { req }) => {
         const lxm = app.bsky.actor.getPreferences.$lxm
-        const aud = computeProxyTo(ctx, req, lxm)
-        permissions.assertRpc({ aud, lxm })
+        const scopeAud = computeProxyTo(ctx, req, lxm)
```
dholms: Is there a reason we renamed this variable in a few parts?

I actually find it a bit misleading since this aud comes from the proxy header not from the scope. I'd be inclined to keep it just `aud` but if we want to be more specific then i think `proxyAud` is more accurate

packages/did/src/did-ref.ts
dholms: I actually kinda wonder, do we even need this type? Could we just use use `AtprotoDidRefAbsolute`? I kinda like talking about what the thing _is_ rather than what it's used for

dholms (approved):
beautiful thanks for nailing this down 💎 

just some small comments on naming convention for the new type but overall looking good

packages/did/src/did-ref.ts
devinivy: I feel you on that.  I think it makes the most sense as a sibling to `AtprotoTokenAud`, where we'd have to type it as `AtprotoDid | AtprotoDidRefAbsolute` everywhere.

I think `AtprotoTokenAud` and service auth helpers in general may need a home that doesn't exist yet— I don't think xrpc-server is the best place for that anymore even though it has been practical to put helpers there so far.

packages/pds/src/api/app/bsky/actor/getPreferences.ts
devinivy: My intention was to disambiguate `aud` when a token audience and scope audience are nearby each other in the code, since these two values differ.  That isn't the case here though, will remove!

packages/oauth/oauth-scopes/src/scopes/rpc-permission.ts
devinivy: I took Daniel's suggestion and removed `AtprotoScopeAud` entirely for now.  I found it was also a little ambiguous in name/purpose versus `AudParam`.  So now the symmetry is back, just without the higher level concept of `AtprotoScopeAud`.

packages/did/src/did-ref.ts
devinivy: I took both of your suggestions here.  While things shake out there is no `AtprotoTokenAud` nor `AtprotoScopeAud` anymore.  In the future I could imagine introducing `AtprotoTokenAud` in a package aimed specifically at implementing service auth.

packages/did/src/did-ref.ts
devinivy: I took this suggestion, but with a small shortcut by not introducing the parse helpers.  I found the tradeoff in sprawl vs. DRYness worth it at this time.

packages/did/src/did-ref.ts
devinivy: I took this suggestion because it is more readable.

matthieusieben (approved):
Super neat 💥

PR #4993, opened 2026-05-21
Add associatedProfiles to app.bsky.embed.getEmbedExternalView

blackmichael:


packages/bsky/src/hydration/hydrator.ts
```diff
@@ -810,20 +811,46 @@ export class Hydrator {
       ),
     ) as AtUriString[]
     if (!ssUris.length) return { ctx }
+    const ssDids = dedupeStrs(dids)
```
estrattonbailey: Could extract DIDs from these URIs, don't really need to pass them in from the skeleton

packages/bsky/src/hydration/hydrator.ts
```diff
@@ -810,20 +811,46 @@ export class Hydrator {
...
+    // is complete.
+    const knownDids = new Set<string>(ssDids)
+    const extraDids: DidString[] = []
+    for (const key of documents.keys()) {
+      const did = uriToDid(parseSiteStandardRecordKey(key).uri)
+      if (!knownDids.has(did)) {
+        knownDids.add(did)
+        extraDids.push(did)
+      }
+    }
```
estrattonbailey: Not a bad idea to cover both, but atm the dataplane will never return additional documents. Only publications are auto-resolved

estrattonbailey (commented):
Looking good! Now we gotta add the same logic to `hydratePosts`: extract DIDs from the SS URIs we find, hydrate profiles, maybe hydrate extra profiles if needed.

packages/bsky/src/hydration/hydrator.ts
blackmichael: I'll remove it to keep only what's needed

packages/bsky/src/hydration/hydrator.ts
```diff
@@ -657,9 +665,30 @@ export class Hydrator {
         labels,
       )
     }
+    // Defensive top-up: in the unlikely case the dataplane returned a
+    // publication owned by a DID not present in the post-author or
+    // pinned-ref sets, fetch its profile serially and fold it in.
+    const knownProfileDids = new Set<string>([
+      ...allPostUris.map(didFromUri),
+      ...ssRefDids,
+    ])
```
estrattonbailey: Nit: this dedupe is replicated above, could re-use

packages/bsky/src/hydration/hydrator.ts
```diff
@@ -657,9 +665,30 @@ export class Hydrator {
...
+    const extraSsDids: DidString[] = []
+    for (const key of siteStandardPublications.keys()) {
+      const did = uriToDid(parseSiteStandardRecordKey(key).uri)
```
estrattonbailey, line 676: Could loop over values and grab the URI from the `ref` instead of parsing those keys (sorry for weird key structure)

packages/bsky/src/api/app/bsky/embed/getEmbedExternalView.ts
```diff
@@ -105,6 +113,11 @@ const standardSitePresentation = (
       }),
     )
     associatedRecords.push(slot.info.record as LexMap)
+    const profile = ctx.views.profileBasic(
+      new AtUri(slot.ref.uri).did,
+      hydration,
+    )
+    if (profile) associatedProfiles.push(profile as ProfileViewBasic)
```
estrattonbailey: Shouldn't need this bc the `view` should come back on `overlay`

estrattonbailey (commented):
Couple small nits 🥳

packages/bsky/src/hydration/hydrator.ts
estrattonbailey: Oh wait not sorry nvm we don't have URI there

estrattonbailey (approved):
:shipit:

packages/bsky/src/views/index.ts
```diff
@@ -2154,6 +2162,7 @@ export class Views {
           : undefined,
         ...ssView,
         associatedRefs: embed.external.associatedRefs,
+        associatedProfiles,
```
estrattonbailey: Oh failing test: let's omit this if array is empty

packages/bsky/src/hydration/hydrator.ts
```diff
@@ -810,20 +838,40 @@ export class Hydrator {
       ),
     ) as AtUriString[]
     if (!ssUris.length) return { ctx }
+    const dids = dedupeStrs(ssUris.map((uri) => new AtUri(uri).did))
```
estrattonbailey: Nit: while we're at it, could use `uriToDid` here for parity

PR #5016, opened 2026-05-27
Fix app.bsky.actor.getProfile verifier data

blackmichael:
In my previous PR https://github.com/bluesky-social/atproto/pull/5015, I introduced `handle` and `displayName` for the verifier on `app.bsky.actor.getProfile` but I did not source that data from the correct hydrated actor. This PR fixes the bug by hydrating the issuer and fetching the issuer actor's data from the hydration state. Verified with updated tests.

### DIFF
--- a/packages/bsky/src/hydration/hydrator.ts
+++ b/packages/bsky/src/hydration/hydrator.ts
@@ -277,6 +277,30 @@ export class Hydrator {
       this.label.getLabelsForSubjects(labelSubjectsForDid(dids), ctx.labelers),
       this.hydrateProfileViewers(dids, ctx),
     ])
+    // Hydrate verification issuer actors so the verification view can expose
+    // the issuer's current handle and displayName. Skipped when no hydrated
+    // actor has any verifications, which is the common case.
+    const issuerDids: DidString[] = []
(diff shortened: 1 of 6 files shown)


packages/bsky/src/hydration/hydrator.ts
```diff
@@ -277,6 +277,30 @@ export class Hydrator {
...
+        if (actors.has(issuer) || issuerDidSet.has(issuer)) continue
+        issuerDidSet.add(issuer)
+        issuerDids.push(issuer)
+      }
+    }
+    if (issuerDids.length > 0) {
+      const issuerActors = await this.actor.getActors(issuerDids, {
+        includeTakedowns,
+      })
+      // Merge into actors without overwriting existing entries (the original
```
rafaeleyng, line 298: Good thinking

rafaeleyng (approved):
Makes sense!

rafaeleyng:
Maybe we should add "issuer" to the properties you added? `issuerDisplayName`, `issuerHandle`?

Or should it maybe be a profile view? Just thinking forward, in case we also need to return `subjectHandle` etc later.

PR #5035, opened 2026-06-02
bsky: enable search v2 behind feature gate

treethought:
This PR routes appview search requests to the new dataplane V2 API.

 The handlers keep their existing V1 skeleton and use a new V2 skeleton based on the `search:v2:enable` feature gate. Search lexicons are unchanged and provide the same response shape to clients.

Note that the current lexicon queries don't expose every filter V2 supports such as`exclude` filters, reply/thread filters, has_media / has_video, following, etc.

### DIFF
--- a/.changeset/crisp-breads-fail.md
+++ b/.changeset/crisp-breads-fail.md
@@ -1,0 +1,5 @@
+---
+'@atproto/bsky': patch
+---
+
+Route appview search through new V2 dataplane RPCs behind the `search:v2:enable` feature gate.
--- a/packages/bsky/proto/bsky.proto
+++ b/packages/bsky/proto/bsky.proto
@@ -1086,7 +1086,7 @@ message GetThreadResponse {
 }
 
 //
-// Search
+// Search V1
 //
 
 // - Return DIDs of actors matching term, paginated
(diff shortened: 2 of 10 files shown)


.changeset/crisp-breads-fail.md
```diff
@@ -0,0 +1,5 @@
+---
+'@atproto/bsky': minor
```
estrattonbailey: Kinda silly that we do this, but this would be the first minor version of this package lol. Let's stick with patches until we're ready to commit to more proper semver :P

packages/bsky/proto/bsky.proto
```diff
@@ -1128,6 +1128,163 @@ message SearchStarterPacksResponse {
...
+  SEARCH_SORT_ORDER_TOP = 2;
+}
+
+enum SearchQueryLanguage {
+  SEARCH_QUERY_LANGUAGE_UNSPECIFIED = 0;
+  SEARCH_QUERY_LANGUAGE_JA = 1; // Japanese (Kuromoji)
+  SEARCH_QUERY_LANGUAGE_ZH = 2; // Chinese (smartcn)
+  SEARCH_QUERY_LANGUAGE_KO = 3; // Korean (nori)
+  SEARCH_QUERY_LANGUAGE_TH = 4; // Thai
+}
```
estrattonbailey, line 1147: How do we plan to support these? Do we ideally want to pass them in from the client?

packages/bsky/src/api/app/bsky/feed/searchPosts.ts
```diff
@@ -107,6 +109,51 @@ const skeleton = async (
...
+  inputs: SkeletonFnInput<Context, Params>,
+): Promise<Skeleton> => {
+  const { ctx, params } = inputs
+  const parsedQuery = parsePostSearchQuery(params.q, {
+    author: params.author,
+  })
+
+  const res = await ctx.dataplane.searchPostsV2({
+    params: {
+      query: params.q,
```
estrattonbailey, line 121: Btw atm the client encodes all filters into the `q` param and the search service parses it. If/when we add an advanced search UI, we'll start using the other params. But pure text queries like `from:me <query>` will probably still be parsed on the backend. Just FYI in case that wasn't clear.

packages/bsky/proto/bsky.proto
treethought: yeah i think ideally we'd pass them from the client. But the dataplane also attempts to autodetect the query's language, so I think this can come later

packages/bsky/src/api/app/bsky/feed/searchPosts.ts
treethought: sounds good, we can parse those in the dataplane handlers

estrattonbailey:
@claude you silly goose, give this one a shake

estrattonbailey (approved):
Nice one!

PR #5037, opened 2026-06-03
Add `atUri()` helper to create AT-URI

matthieusieben:
Adds an `atUri()` helper to `@atproto/lex` for safely constructing `AtUriString` values, and migrates `bsky` / `pds` call-sites to use it.

The new helper provides a new type safe & runtime safe way of generating AT-URIs.

```ts
import { atUri } from '@atproto/lex'

// From a record schema — type-safe, validates literal/format constraints
atUri(did, app.bsky.feed.post, 'my-post')

// Literal rkeys (e.g. `self`) can be omitted
atUri(did, app.bsky.actor.profile)

// From raw strings — validated at runtime, branded at compile time
atUri(did, collection, rkey)

// Authority only
atUri(did)
```

Inputs are validated via `assertAtIdentifierString` / `assertStringFormat` so untrusted values can't produce malformed URIs.


packages/pds/src/api/com/atproto/repo/getRecord.ts
```diff
@@ -11,7 +11,8 @@ export default function (server: Server, ctx: AppContext) {
 
     // fetch from pds if available, if not then fetch from appview
     if (did) {
-      const uri = AtUri.make(did, collection, rkey)
+      // @NOTE each part of the URI was validated by the lexicon schema
+      const uri: AtUriString = `at://${did}/${collection}/${rkey}`
```
devinivy: I prefer to avoid manually concatenating together at-uris in application code, as it relies on the assumption that there is never any encoding, for example.  Unfortunately the type system itself wont be enough to give us the guarantees we need that it's a valid at-uri, so in my view we need to pass through a library/helper for robustness.

packages/lex/lex-schema/src/schema/record.ts
```diff
@@ -120,21 +168,13 @@ export class RecordSchema<
...
-          : never
+  RecordKeyValue<Key>
 
 export type RecordKeySchema<Key extends LexiconRecordKey> = Schema<
-  RecordKeySchemaOutput<Key>
+  RecordKeyValue<Key>
 >
 
-const keySchema = string({ minLength: 1 })
+const keySchema = string({ format: 'record-key' })
```
devinivy, line 130: Ah, nice.

packages/lex/lex-schema/src/schema/record.test.ts
```diff
@@ -279,12 +278,93 @@ describe('RecordSchema', () => {
...
+      })
+    })
+  })
+
   describe('key type: any', () => {
     const schema = record(
       'any',
       'app.bsky.feed.post',
       object({
-        $type: string(),
```
devinivy, line 287: Is this removed not because it's wrong, but because it's not what's being tested?

matthieusieben: Both, actually

devinivy (approved):

PR #5044, opened 2026-06-05
pds: proxy unregisterPush to notification service (mirrors registerPush)

### LINKED ISSUE #5043: `app.bsky.notification.unregisterPush` is not proxied to the notification service by PDS (unlike `registerPush`)
**Describe the bug**

`app.bsky.notification.unregisterPush` is not proxied to the notification service by PDS, unlike the analogous `registerPush` endpoint.

`registerPush` has a dedicated PDS handler that reads `serviceDid` from the request body, resolves the `bsky_notif` endpoint from the DID document, and proxies the request there with a service auth JWT. `unregisterPush` has no such handler, so requests fall through to the generic catchall pipethrough, which routes to the configured bskyAppView instead of the correct notification service — resulting in an error.

**To Reproduce**

1. Call `app.bsky.notification.unregisterPush` through PDS without the `atproto-proxy` header:

```
POST /xrpc/app.bsky.notification.unregisterPush
Authorization: Bearer <access_token>
Content-Type: application/json

{
  "serviceDid": "<notification-service-did>",
  "token": "...",
  "platform": "ios",
  "appId": "..."
}
```

2. Observe `InvalidRequestError: Invalid serviceDid.`

On a local dev server (where the bskyAppView is not configured with Courier), the error is:

```
MethodNotImplemented: This service is not configured to support push token registration.
```

**Expected behavior**

`unregisterPush` should behave identically to `registerPush` from the client's perspective: the client sends the request with `serviceDid` in the body, and the PDS resolves the correct notification service endpoint from the DID document and proxies the request transparently — no `atproto-proxy` header needed.

**Details**

- Operating system: macOS
- Node version: v22.16.0

**Additional context**

A workaround exists: include `atproto-proxy: <notification-service-did>#bsky_notif` in the request headers. However, this workaround has two problems:

1. **Not documented.** The [API reference for `unregisterPush`](https://docs.bsky.app/docs/api/app-bsky-notification-unregister-push) states that authenticated requests should be routed through the user's PDS, but says nothing about needing to set `atproto-proxy` manually. A client that follows the documentation has no way to discover this requirement.

2. **Redundant with the request body.** `serviceDid` is already a required field in the request body. Having to also encode the same DID in an `atproto-proxy` header (e.g. `<serviceDid>#bsky_notif`) is unexpected duplication that `registerPush` — the directly analogous endpoint — does not require.

nnabeyang:
`app.bsky.notification.registerPush` has a dedicated PDS handler that reads
`serviceDid` from the request body, resolves the `bsky_notif` endpoint from
the DID document, and proxies the request there with a service auth JWT.
`app.bsky.notification.unregisterPush` had no such handler, so requests fell
through to the catchall pipethrough and were routed to the bskyAppView instead
of the correct notification service — resulting in `InvalidRequestError: Invalid
serviceDid` for clients that did not manually set the `atproto-proxy` header.

This PR adds a dedicated `unregisterPush` handler that mirrors `registerPush`
exactly, so clients can call either endpoint the same way: send `serviceDid` in
the request body, and the PDS handles routing transparently.

**Before:** calling `unregisterPush` without `atproto-proxy` header fails with
`InvalidRequestError: Invalid serviceDid.`

**After:** `unregisterPush` is proxied to the notification service identified by
`serviceDid`, identical to how `registerPush` behaves.

Fixes #5043

### DIFF
--- a/.changeset/calm-bulls-cure.md
+++ b/.changeset/calm-bulls-cure.md
@@ -1,0 +1,5 @@
+---
+'@atproto/pds': patch
+---
+
+Proxy `app.bsky.notification.unregisterPush` to the notification service identified by `serviceDid`, matching the behavior of `registerPush`.
--- a/packages/pds/src/api/app/bsky/notification/index.ts
+++ b/packages/pds/src/api/app/bsky/notification/index.ts
@@ -1,7 +1,9 @@
 import { Server } from '@atproto/xrpc-server'
 import { AppContext } from '../../../../context.js'
 import registerPush from './registerPush.js'
+import unregisterPush from './unregisterPush.js'
 
 export default function (server: Server, ctx: AppContext) {
   registerPush(server, ctx)
+  unregisterPush(server, ctx)
 }
--- a/packages/pds/src/api/app/bsky/notification/unregisterPush.ts
+++ b/packages/pds/src/api/app/bsky/notification/unregisterPush.ts
@@ -1,0 +1,65 @@
+import { getNotif } from '@atproto/identity'
+import { xrpc } from '@atproto/lex'
+import { InvalidRequestError, Server } from '@atproto/xrpc-server'
+import { AppContext } from '../../../../context.js'
+import { app } from '../../../../lexicons/index.js'
+import { getDidDoc } from '../util/resolver.js'
+
+export default function (server: Server, ctx: AppContext) {
+  const { bskyAppView } = ctx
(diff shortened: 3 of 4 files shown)


devinivy (approved):

nnabeyang:
@devinivy Rebased onto latest `main` to resolve a merge conflict and force-pushed.

PR #5047, opened 2026-06-05
expand notification preferences lexicon with chat requests

rafaeleyng:


lexicons/app/bsky/notification/defs.json
```diff
@@ -10,7 +10,15 @@
       "type": "object",
       "required": ["include", "push"],
       "properties": {
-        "include": { "type": "string", "knownValues": ["all", "accepted"] },
+        "include": { "type": "string", "knownValues": ["all", "follows"] },
```
rafaeleyng, line 13: These known values changed, but this `chat` property wasn't used at all (it is not presented on our UI currently). So shouldn't break compatibility.

ds-boyce (approved):

rafaeleyng:

PR #5048, opened 2026-06-05
Add gallery embed images to og data-plane and dev-env

blackmichael:


packages/dev-env/src/mock/index.ts
blackmichael, line 1: Tangentially related - introduced this to make local development on https://github.com/bluesky-social/ozone/pull/512 easier

estrattonbailey (approved):
LGTM, also tested against the local app

PR #5050, opened 2026-06-05
bsky: consume expanded notification preferences

rafaeleyng:
Track tango proto changes (bluesky-social/tango#1034):
- rename ChatNotificationInclude.ACCEPTED -> FOLLOWS (value 2 kept)
- add ChatRequestNotificationPreference message + chat_request field

Wire chatRequest through the getPreferences/putPreferencesV2 lex<->proto converters and the dataplane mock, and update the chat include value from "accepted" to "follows" to match the app.bsky.notification lexicon.

### DIFF
--- a/packages/bsky/proto/bsky.proto
+++ b/packages/bsky/proto/bsky.proto
@@ -818,10 +818,15 @@ message NotificationPreference {
 enum ChatNotificationInclude {
   CHAT_NOTIFICATION_INCLUDE_UNSPECIFIED = 0;
   CHAT_NOTIFICATION_INCLUDE_ALL = 1;
-  CHAT_NOTIFICATION_INCLUDE_ACCEPTED = 2;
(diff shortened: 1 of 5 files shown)


packages/bsky/proto/bsky.proto
```diff
@@ -818,14 +818,19 @@ message NotificationPreference {
 enum ChatNotificationInclude {
   CHAT_NOTIFICATION_INCLUDE_UNSPECIFIED = 0;
   CHAT_NOTIFICATION_INCLUDE_ALL = 1;
-  CHAT_NOTIFICATION_INCLUDE_ACCEPTED = 2;
```
rafaeleyng, line 821: It's a bit weird to change a proto label, but it should be fine. First, because we never used it. Second, because it is just the label. What matters is the numeric tag.

rafaeleyng:

PR #5051, opened 2026-06-05
bsky: add internal.bsky.notification.getPreferences endpoint

rafaeleyng:
New internal (service-to-service) query that takes a list of DIDs and returns notification preferences for each account that has them stored.

- new `internal` lexicon namespace with getPreferences query; output is a non-positional array of { did, preferences } items, so accounts with no stored preferences are simply omitted
- handler uses ctx.authVerifier.role and fetches from the dataplane, reusing the existing protobufToLex converter (default-filled)
- tests covering found/omitted/empty cases and role-auth enforcement

### DIFF
--- a/.changeset/sour-lizards-repair.md
+++ b/.changeset/sour-lizards-repair.md
@@ -1,0 +1,5 @@
+---
+'@atproto/bsky': patch
+---
+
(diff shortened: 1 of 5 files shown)


rafaeleyng:

PR #5053, opened 2026-06-08
Account manager account deactivation

matthieusieben:


packages/oauth/oauth-provider-ui/src/components/forms/button-cooldown.tsx
```diff
@@ -50,66 +50,50 @@ export function ButtonCooldown({
   const percent = ((handler.total - handler.remaining) / handler.total) * 100
 
   return (
```
matthieusieben, line 60: the popover caused wiered behavior on hover (blur currently focussed input), so I simply removed it.

matthieusieben:
@claude review the whole diff with `main`, as if I just opened the PR.

.changeset/fine-dots-tap.md
```diff
@@ -0,0 +1,5 @@
+---
+'@atproto/pds': patch
+---
+
+`com.atproto.admin.updateSubjectStatus` will throw when trying to re-activate an account who's did does not resolve to the right PDS
```
devinivy:
```suggestion
`com.atproto.admin.updateSubjectStatus` will throw when trying to re-activate an account whose DID does not resolve to the right PDS
```

packages/oauth/oauth-provider-ui/src/components/delete-account-dialog.tsx
```diff
@@ -0,0 +1,219 @@
...
+        open={open}
+        onOpenChange={setOpen}
+        dismissable={dismissable}
+        title={<Trans>Are you really, really sure?</Trans>}
+        description={
+          handle ? (
+            <Trans>
+              This will irreversibly delete your Bluesky account{' '}
+              <Handle handle={handle} className="font-bold" /> and all
+              associated data. Note that this will affect any other AT Protocol
```
devinivy: We tend to keep AT Protocol out of content for end users, and instead use Atmosphere.
```suggestion
              associated data. Note that this will affect any other Atmosphere
```

packages/oauth/oauth-provider-ui/src/components/delete-account-dialog.tsx
```diff
@@ -0,0 +1,219 @@
...
+              services you use with this account.
+            </Trans>
+          ) : (
+            <Trans>
+              This will irreversibly delete your Bluesky account and all
+              associated data. Note that this will affect any other AT Protocol
```
devinivy:
```suggestion
              associated data. Note that this will affect any other Atmosphere
```

packages/oauth/oauth-provider-ui/src/components/reactivate-account-dialog.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+  return (
+    <DialogSimple
+      trigger={children}
+      title={<Trans>Reactivate account</Trans>}
+      description={
+        <Trans>
+          Your profile, posts, feeds, and lists will become visible again across
+          the Atmosphere network — that includes the Bluesky app and any other
+          AT Protocol app you use this account on.
```
devinivy:
```suggestion
          Atmosphere app you use with this account.
```

packages/oauth/oauth-provider-ui/src/contexts/authentication.tsx
```diff
@@ -82,17 +84,23 @@ export function AuthenticationProvider({
...
-          account.preferred_username === forcedIdentifier,
+          account.did === forcedIdentifier ||
+          account.handle === forcedIdentifier,
       )
-      return [false, matchingSessions[0] ?? null, matchingSessions]
+      const matchingSession = currentSession
+        ? matchingSessions.find(
+            ({ account }) => account.did === currentSession.account.did,
+          ) ?? null
+        : null
```
devinivy, line 95: I take it there's no more than one session per DID?

packages/oauth/oauth-provider/src/router/create-api-middleware.ts
```diff
@@ -159,46 +158,21 @@ export function createApiMiddleware<
...
           : await server.signer.createEphemeralToken({
-              sub: account.sub,
+              sub: account.did,
               deviceId,
               requestUri,
             })
 
-        if (requestUri) {
-          // Check if a consent is required for the client, but only if this
-          // call is made within the context of an oauth request.
```
devinivy, line 179: Where did the behavior from this block of code move?

packages/oauth/oauth-provider/src/token/token-store.ts
```diff
@@ -65,7 +65,7 @@ export interface TokenStore {
 
   findTokenByCode(code: Code): Awaitable<null | TokenInfo>
 
-  listAccountTokens(sub: Sub): Awaitable<TokenInfo[]>
+  listAccountTokens(did: Did): Awaitable<TokenInfo[]>
```
devinivy, line 68: Lots of these `sub -> did` changes mixed in to review as part of this work!  I think it may have made things a bit simpler to have these changes broken out.

packages/pds/src/account-manager/helpers/account-device.ts
```diff
@@ -52,9 +48,7 @@ export function selectQB(
         'device.ipAddress',
         'device.lastSeenAt',
       ])
-      .if(filter.sub != null, (qb) =>
-        qb.where('actor.did', '=', filter.sub! as DidString),
-      )
+      .if(filter.did != null, (qb) => qb.where('actor.did', '=', filter.did!))
```
devinivy: I think these `.if()` may need to be updated with the kysely upgrade having landed.

packages/pds/src/account-manager/helpers/token.ts
```diff
@@ -122,9 +123,9 @@ export const findByQB = (
     )
 }
 
-export const removeByDidQB = (db: AccountDb, did: string) =>
+export const removeByDid = async (db: AccountDb, did: Did) =>
   // uses "token_did_idx" index
-  db.db.deleteFrom('token').where('did', '=', did)
+  db.db.deleteFrom('token').where('did', '=', did).execute()
```
devinivy: Should this be executed with retry?

.changeset/fine-dots-tap.md
devinivy: It would be cool for the new `updateSubjectStatus` behaviors to be codified into some tests.  In order to get this landed, I'd be fine making that a small followup task if we're already confident in the behavior.

devinivy (approved):
Left a few notes, but looking solid!

packages/oauth/oauth-provider-ui/src/contexts/authentication.tsx
matthieusieben: yep

packages/pds/src/account-manager/helpers/token.ts
matthieusieben: added

packages/oauth/oauth-provider/src/router/create-api-middleware.ts
matthieusieben: The way consent works has changed, and was actually the cause of the "redirection stuck" issue.

The front-end does nothing with the `consentRequired`, so we don't compute that value here.

Consent required is only used backend side to determine if SSO can be used. Otherwise, a consent screen will always be shown. This is not really a change wrt. `main` but this branch cleansup a bunch of related stuff which is why we see this here.

packages/oauth/oauth-provider/src/token/token-store.ts
matthieusieben: I'm sorry about that. I started changing them after I started working. Changing the order of the change would have been equally hard to implement.

.changeset/fine-dots-tap.md
matthieusieben: added tests

PR #5067, opened 2026-06-10
Upgrade pnpm v8 to v11 and drop corepack

devinivy:
Upgrades pnpm from 8.15.9 to 11.5.2, one major at a time. All resolved dependency versions are preserved — only the lockfile format and pnpm configuration change.

The notable migrations along the way:

- The `pnpm.executionEnv.nodeVersion` field is dropped (pnpm 9 rejects a range, pnpm 11 ignores the field entirely); the Node requirement now lives in `devEngines.runtime`.
- pnpm 10 stopped running dependency build scripts by default, and pnpm 11 turned this into a hard error. Build-script approval now lives in `allowBuilds` in `pnpm-workspace.yaml`, with each dependency that has a build script explicitly allowed or denied. `overrides` and `patchedDependencies` moved there too, since pnpm 11 no longer reads the `pnpm` field in `package.json`.

This also moves off corepack, which is being unbundled from Node. The pnpm version is pinned in `devEngines.packageManager` (enforced by pnpm itself, so the per-Dockerfile version can't silently drift), CI bumps `pnpm/action-setup` to v6 so it reads that field, and the service Dockerfiles install pnpm with `npm install -g pnpm@11.5.2` instead of corepack.

Finally, CI now publishes with `pnpm publish -r --provenance` rather than `changeset publish`. The previous setup relaxed the pnpm `devEngines` check before publishing because changesets shells out to npm internally; this was originally worked around because pnpm 8 didn't support npm's OIDC-based registry auth. pnpm 11 publishes via OIDC natively, so npm is no longer in the publish path. (See #4928 #4938)


package.json
matthieusieben, line 33: Don't these also need the `--recursive`, `--stream` args to be before the command ?

devinivy: If I am understanding correctly these are before the command, which is `build`.

matthieusieben (approved):

PR #5070, opened 2026-06-10
Make account creation and deletion more resilient

matthieusieben:
This PR attempts to fix a situation in which an account would sill be referenced from the account db when their actor store (db & key) were removed from disk. This can happen either if the server crashes during an account deletion, or if any of the account deletion steps failed, or if the account creation fails in the middle.


matthieusieben:
Note that https://github.com/bluesky-social/atproto/pull/4014 should also help here;

matthieusieben:
@claude Please review this PR

packages/pds/src/account-manager/oauth-store.ts
```diff
@@ -172,39 +168,57 @@ export class OAuthStore
...
+              await this.sequencer.deleteAccount(did)
+              throw err
+            }
+          } catch (err) {
+            await this.accountManager.deleteAccount(did)
+            throw err
+          }
         } catch (err) {
-          this.accountManager.deleteAccount(did)
+          await this.plcClient.tombstone(did, signingKey)
```
dholms: this needs to be the rotationKey

packages/pds/src/api/com/atproto/server/createAccount.ts
```diff
@@ -47,68 +45,77 @@ export default function (server: Server, ctx: AppContext) {
...
+              },
+            }
+          } catch (err) {
+            await ctx.accountManager.deleteAccount(did)
+            throw err
+          }
+        } catch (err) {
+          // If we created the did just now, let's tombstone it
+          if (!ctx.entrywayClient && !input.body.did && plcOp) {
+            await ctx.plcClient.tombstone(did, signingKey)
```
dholms: same here!

dholms (approved):
overall looks right & more robust than what I had before! just the note on which key to use for tombstones

packages/pds/src/account-manager/account-manager.ts
```diff
@@ -221,33 +221,36 @@ export class AccountManager {
...
+    const passwordScrypt =
+      email && password ? await scrypt.genSaltAndHash(password) : undefined
 
     const now = currentDatetimeString()
-    await this.db.transaction(async (dbTxn) => {
+    return this.db.transaction(async (dbTxn) => {
       if (inviteCode) {
         await invite.ensureInviteIsAvailable(dbTxn, inviteCode)
       }
-      await Promise.all([
```
devinivy, line 233: Very happy to see this made serial and more deterministic 💆

packages/pds/src/api/com/atproto/admin/deleteAccount.ts
```diff
@@ -8,13 +7,14 @@ export default function (server: Server, ctx: AppContext) {
...
     handler: async ({ input }) => {
       const { did } = input.body
-      await ctx.actorStore.destroy(did)
+
+      // @NOTE Order matters here and is the reverse order of account creation.
+      // Putting the sequencer first allows for proper restoration of the
+      // account's state in case of outage recovery. We then "unlink" the
+      // account and finally remove the files from the file system.
+      await ctx.sequencer.deleteAccount(did)
       await ctx.accountManager.deleteAccount(did)
```
devinivy: I think there's a good case to be made that the account db should be the canonical state and come first. Perhaps specifically the relevant row in the actor or account table. The sequencer data is ephemeral, regularly trimmed, and backed-up with the same guarantees as the sequencer.

packages/pds/src/account-manager/oauth-store.ts
devinivy: This is a sensitive one— we need to be absolutely certain that this `did` is brand new, and not an existing one we control.  Particularly in this generic error condition that could be caused by a programmer error.  I think we should at minimum have a comment documenting what makes this safe.

packages/pds/src/sequencer/sequencer.ts
```diff
@@ -208,6 +209,42 @@ export class Sequencer extends (EventEmitter as new () => SequencerEmitter) {
...
+  }
+
+  async deleteAccount(did: DidString) {
+    const seq = await this.sequenceAccountEvt(did, AccountStatus.Deleted)
+    await this.db.executeWithRetry(
+      this.db.db
+        .deleteFrom('repo_seq')
+        .where('did', '=', did)
+        .where('seq', '!=', seq),
+    )
```
devinivy, line 238: Should we use the existing method `deleteAllForUser()`?

packages/pds/src/api/com/atproto/server/createAccount.ts
devinivy: Similar to my other note, I think we should at minimum document with a comment why this is definitely a new DID and not an existing one.  Acknowledging all the data used to compute a DID is public.

packages/pds/src/api/com/atproto/server/deleteAccount.ts
```diff
@@ -54,13 +53,14 @@ export default function (server: Server, ctx: AppContext) {
...
         token,
       )
-      await ctx.actorStore.destroy(did)
+
+      // @NOTE Order matters here and is the reverse order of account creation.
+      // Putting the sequencer first allows for proper restoration of the
+      // account's state in case of outage recovery. We then "unlink" the
+      // account and finally remove the files from the file system.
+      await ctx.sequencer.deleteAccount(did)
       await ctx.accountManager.deleteAccount(did)
```
devinivy: This is also relevant to my comment about recovery and canonical account state.

packages/pds/src/sequencer/sequencer.ts
matthieusieben: `deleteAllForUser` was actually not used anywere else to I removed it, keeping only this simpler implementation

packages/pds/src/api/com/atproto/admin/deleteAccount.ts
matthieusieben: I've put it first and made sure that the sequencer and actor store deletion both run

devinivy (approved):

PR #5073, opened 2026-06-10
Add ability to disable global rate limits at a route level

matthieusieben:


jcalabro (commented):
nice thanks!

packages/xrpc-server/src/server.ts
```diff
@@ -711,9 +713,6 @@ export class Server {
       }
     })
 
-    // If the route config contains an empty array, use global rate limiter.
-    if (!rateLimiters.length) return globalRateLimiter
```
matthieusieben, line 715: This is not needed as it is already performed by `HttpRateLimiter.from`

packages/xrpc-server/src/types.ts
```diff
@@ -157,6 +157,7 @@ export type SharedRateLimitOpts<C extends HandlerContext = HandlerContext> = {
 }
 
 export type RouteRateLimitOpts<C extends HandlerContext = HandlerContext> = {
+  ignoreGlobalLimits?: boolean
```
devinivy: This doesn't seem to be used anywhere.

packages/xrpc-server/src/types.ts
```diff
@@ -202,6 +203,7 @@ export type MethodConfig<
   handler: MethodHandler<A, P, I, O>
   auth?: MethodAuth<A, P>
   opts?: RouteOptions
+  ignoreGlobalRateLimits?: boolean
   rateLimit?: MethodRateLimit<A, P, I>
```
devinivy, line 206: It would be ideal if `ignoreGlobalRateLimits` could live somewhere in the `rateLimit` config.  Is there not really a place to fit it in there?

devinivy: Side thing—I noticed this wasn't added to `MethodConfigWithAuth` which felt inconsistent.  But is that type defunct anyway?

matthieusieben: I liked that better too, but `rateLimit` can be an array which means that one could write:

```ts
rateLimit: [{
  ignoreGlobalRateLimits: true,
  durationMs: 12
  points: 12
}, {
  ignoreGlobalRateLimits: false,
  durationMs: 12
  points: 12
}]
```

We could change that property in a backwards incompatible way if you think that's better ?

matthieusieben:
Note that this could also trivially achieved by specifying a custom `calcKey` function in the [global rate limit config](https://github.com/bluesky-social/atproto/blob/cc329bf55c658752675d561dba792fedbe9c2626/packages/pds/src/index.ts#L135-L139), which might be a better pattern:

```ts
              {
                name: 'global-ip',
                durationMs: 5 * MINUTE,
                points: 3000,
                calcKey: (ctx) => {
                  // "sync" requires more than 3k/5min. Rate limit fine tuned
                  // on those endpoints.
                  if (ctx.req.path === '/xrpc/com.atproto.sync.getRecord') {
                    return null
                  }

                  return ctx.req.ip
                },
              },
```

PR #5074, opened 2026-06-10
📊 Add prometheus to ozone

foysalit:
This PR adds `opt-in` prometheus metrics server wiring to both ozone's app server and daemon. It also adds a few preliminary metrics like 
1. histogram metrics for app's xrpc endpoints
2. histogram metrics for daemon's queue routing


packages/ozone/src/index.ts
```diff
@@ -141,4 +179,69 @@ export class OzoneService {
...
+
+  static create(
+    register: prometheus.Registry,
+    opts: MetricsServiceOpts = {},
+  ): MetricsService {
+    const app = express()
+
+    app.get('/metrics', async (_req, res) => {
+      res.set('Content-Type', register.contentType)
+      res.end(await register.metrics())
```
matthieusieben: Any error thrown here will crash the server

foysalit: Fixed: the handler now catches `register.metrics()` failures and returns HTTP 500, preventing Express 4 from surfacing the rejection as an unhandled process-level error.

PR #5082, opened 2026-06-11
Increase PDS com.atproto.sync.getRepo Rate Limits

jcalabro:
Previous work #5073 

This doubles the rate limit on `com.atproto.sync.getRepo` from 3k/5m to 6k/5m. Other routes are untouched.

I didn't add this as a tuning parameter for operators to set. Thoughts on allowing people to control this one endpoint via an env var? I was thinking that other operators may not want this change? This is a heavy-duty endpoint. If you think that's worthwhile, I can add that as a knob.

### DIFF
--- a/.changeset/strict-animals-make.md
+++ b/.changeset/strict-animals-make.md
@@ -1,5 +1,4 @@
 ---
-'@atproto/common': patch
 '@atproto/pds': patch
 ---
 
--- a/.changeset/strict-animals-make.md
+++ b/.changeset/strict-animals-make.md
@@ -1,0 +1,6 @@
+---
+'@atproto/common': patch
+'@atproto/pds': patch
+---
+
+Increase rate limit rules on com.atproto.sync.getRepo
--- a/packages/pds/src/index.ts
+++ b/packages/pds/src/index.ts
@@ -16,9 +16,7 @@ const { createHttpTerminator } = httpTerminator
 type HttpTerminator = ReturnType<typeof createHttpTerminator>
 import { DAY, SECOND } from '@atproto/common'
 import {
-  MemoryRateLimiter,
   MethodHandler,
-  RedisRateLimiter,
   ResponseType,
   XRPCError,
   createServer,
@@ -114,14 +112,7 @@ export class PDS {
 
         return XRPCError.fromError(err)
       },
-      rateLimits: rateLimits.enabled
-        ? {
-            ...buildRateLimitsConfig(rateLimits),
-            creator: ctx.redisScratch
-              ? (opts) => new RedisRateLimiter(ctx.redisScratch, opts)
-              : (opts) => new MemoryRateLimiter(opts),
-          }
-        : undefined,
+      rateLimits: buildRateLimitsConfig(rateLimits, ctx.redisScratch),
     })
 
     apiRoutes(server, ctx)
--- a/packages/pds/src/index.ts
+++ b/packages/pds/src/index.ts
@@ -14,7 +14,7 @@ import httpTerminator from 'http-terminator'
 // eslint-disable-next-line import/no-named-as-default-member
 const { createHttpTerminator } = httpTerminator
 type HttpTerminator = ReturnType<typeof createHttpTerminator>
-import { DAY, HOUR, MINUTE, SECOND } from '@atproto/common'
+import { DAY, SECOND } from '@atproto/common'
 import {
   MemoryRateLimiter,
   MethodHandler,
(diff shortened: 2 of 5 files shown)


packages/pds/src/index.ts
```diff
@@ -115,41 +116,10 @@ export class PDS {
...
-                durationMs: HOUR,
-                points: 5000, // creates=3, puts=2, deletes=1
-              },
-              {
-                name: 'repo-write-day',
-                durationMs: DAY,
-                points: 35000, // creates=3, puts=2, deletes=1
-              },
-            ],
           }
```
matthieusieben: This pattern could silently cause gnarly issues. Eg. if a `bypass` / `global` / `shared` property is added here, it will silently override the same property returned by `buildRateLimitsConfig()`. I recommend either moving the code back here or make `buildRateLimitsConfig()` build the entire config object:

```ts
      rateLimits: buildRateLimitsConfig(rateLimits, ctx.redisScratch),
```

.changeset/strict-animals-make.md
```diff
@@ -0,0 +1,6 @@
+---
+'@atproto/common': patch
```
matthieusieben:
```suggestion
```

packages/pds/src/rate-limits.ts
```diff
@@ -0,0 +1,58 @@
...
+          if (req.path === SYNC_GET_REPO_PATH) {
+            return null
+          }
+          return req.ip
+        },
+      },
+    ],
+    shared: [
+      {
+        name: SYNC_GET_REPO_IP_RATE_LIMIT,
```
matthieusieben: The current practice is to use strings here. I would either change every `name` to use constants or simply use a string so that we don't have two ways of doing things.

Are you using a "shared" limit because you foresee more endpoints needing to use the same limit? Otherwise I would just move this rule to that endpoint's implementation file.

packages/pds/src/rate-limits.ts
```diff
@@ -0,0 +1,58 @@
...
+      }
+      return false
+    },
+    global: [
+      {
+        name: 'global-ip',
+        durationMs: 5 * MINUTE,
+        points: 3000,
+        calcKey: ({ req }) => {
+          if (req.path === SYNC_GET_REPO_PATH) {
```
matthieusieben, line 39: It is probably worth calling out here why we need this particular exception here (in a comment)

jcalabro: What did you think of the idea of making it user configurable with an env var? Do you think other PDS hosters will be happy with this change?

.changeset/strict-animals-make.md
jcalabro: :+1: I had modified the common path at the time I made the patch, but reverted it. Thanks!

packages/pds/src/rate-limits.ts
matthieusieben: I think the reference implementation sets the bar for what actors in the network should do to ensure proper interop, so not making it configurable is fine. 

That being said, this highly depends on the number of users per pds, which can greatly vary from one deployment to another.

So yeah maybe using an env Var makes sense

jcalabro: To me it feels a little weird to expose a knob for a single endpoint. But it's a relatively heavy-duty endpoint, and rate limits matter a lot on this particular endpoint. :shrug: looking for a call here, I don't think it's my place to say

devinivy: Yeah, I think I'd rather not expose a knob just for this endpoint, and I'd also rather not expand the scope of this PR to other rate limiting config.  So I think we can land this as-is and we can consider separately how we'd want to expose rate limit config.

uniphil: for small operators trying to defend pds responsiveness for users over crawl traffic (or DOS), singling out this endpoint for a separate higher limit seems like potentially opposite of what would be helpful if resources are constrained?

i'm assuming it's the most expensive endpoint (and the most likely DOS target), at least for non-tiny repos (tiny repos are very over-represented on mushrooms vs indie).

any reason to give it a separate bucket instead of just bumping the ip limiter?

(i don't think ratelimits should be a very critical part of interop tbh, any pds outside bluesky can set ip limits like 1req/sec and backfill still finishes them before ~any mushroom)

matthieusieben (approved):

matthieusieben:
Let's merge like this for now, and move the discussions about allowing configuration elsewhere.

jcalabro:
Thanks for the review all! Excited about this!

PR #5083, opened 2026-06-11
ozone: add custom queues

gcwill70:
Allows routing of a report directly to a named queue via modTool.meta.queueId on the modEventReport, bypassing attribute matching. An unknown or disabled queue lands the report unrouted (queueId = -1) rather than falling back, so misconfigured rules surface visibly.


packages/ozone/bin/migration-create.ts
```diff
@@ -2,6 +2,9 @@
 
 import * as fs from 'node:fs/promises'
 import * as path from 'node:path'
+import { fileURLToPath } from 'node:url'
```
foysalit, line 5: not clear why this change is necessary?

foysalit (approved):
looks good. i like the `isAutomated` flag but wonder if we could avoid that entirely through some other mechanism. at some point, we brainstormed marking ozone members as `bot/automation` so that all actions taken by them can be flagged as automated across the board. doing this just for reports feels a bit awkward but i think extending won't be that difficult.

packages/ozone/bin/migration-create.ts
gcwill70: I discovered that this script broke after we migrated to ESM

You get this error below when trying to create a migration without this change.
```
const dir = path.join(__dirname, '..', 'src', 'db', 'migrations')
                        ^
ReferenceError: __dirname is not defined
```

gcwill70:
> looks good. i like the `isAutomated` flag but wonder if we could avoid that entirely through some other mechanism. at some point, we brainstormed marking ozone members as `bot/automation` so that all actions taken by them can be flagged as automated across the board. doing this just for reports feels a bit awkward but i think extending won't be that difficult.

That makes sense. I was following conventions of `reportActivityView` for this. I'll backlog a task to migrate away from this and allow ozone users to be marked as bots.

PR #5086, opened 2026-06-11
Upgrade kysely from 0.22 to 0.29

devinivy:
Upgrades `kysely` from `^0.22.0` to `^0.29.2` in the `bsync`, `bsky`, `pds`, and `ozone` packages, one commit per package.

Most of the diff is mechanical migration of APIs removed between 0.22 and 0.29:

  - `.if()` → `.$if()`
  - `orWhere` / `whereExists` / `whereNotExists` → expression builder forms (`eb.or([...])`, `eb.exists(...)`, `not(exists(...))`)
  - `Migrator` / `Migration` / `MigrationProvider` now import from `kysely/migration`
  - Raw SQL used in `where()` clauses is now explicitly typed (`sql<SqlBool>`), and bare generics like `RawBuilder` / `DynamicModule` take explicit type args
  - `alterColumn()` uses its callback form

A few smaller items along the way:

  - Removes `patches/kysely@0.22.0.patch` — it only re-exported some internal types which are no longer referenced
  - Bumps `@types/pg` to `^8.15.5` in the pg-backed packages (kysely 0.29's `PostgresPool` type requires `Pool.options`)
  - `set/service.ts` (ozone) uses `$narrowType` to keep the non-null `setSize` result type, since 0.25 made subquery selections nullable by default
  - One bsky test helper switches to `db.dynamic.table()`, since 0.28 dropped support for passing table-name unions to `selectFrom`

Worth noting for review: kysely 0.28 removed the `preventAwait` runtime guard, so awaiting a query builder without `.execute()` is now a silent no-op rather than a throw. No such call sites exist today (0.22 would have thrown at runtime), but nothing statically guards against new ones since the repo's eslint config doesn't use type-aware rules.

All four packages' test suites pass against 0.29.2.


devinivy:
@claude the `eb.or([])` does not throw at runtime on 0.29 as suggested in finding 1.  Instead it uses `1 = 0`, similar to how we previously guarded for that on 0.22 in the application code.

matthieusieben:
We probably should also address these hacks:

https://github.com/bluesky-social/atproto/blob/bf9c49f4bb18b3b7a3906e77e0446a9f551a173c/packages/ozone/src/db/migrations/20250211T132135150Z-moderation-event-message-partial-idx.ts#L12-L15

https://github.com/bluesky-social/atproto/blob/bf9c49f4bb18b3b7a3906e77e0446a9f551a173c/packages/pds/src/account-manager/db/migrations/004-oauth.ts#L20-L21


https://github.com/bluesky-social/atproto/blob/bf9c49f4bb18b3b7a3906e77e0446a9f551a173c/packages/pds/src/account-manager/db/migrations/004-oauth.ts#L91

https://github.com/bluesky-social/atproto/blob/bf9c49f4bb18b3b7a3906e77e0446a9f551a173c/packages/pds/src/account-manager/db/migrations/007-lexicon-failures-index.ts#L7

There are also a bunch of:

```ts
    .expression(sql`"suspendCount" ASC NULLS FIRST`)
```

which may now be supported by the query builder (?)


Searching for every instance of <code>.expression(sql`"</code> return most of the cases.

@claude open a PR which addresses this.

devinivy:
@matthieusieben I'd prefer to keep the upgrade as mechanical as possible and let service owners address these now that they have the new version available to them.

foysalit (approved):
Ozone changes look great!

matthieusieben (approved):

PR #5089, opened 2026-06-12
Fix stuck OAuth redirection

### LINKED ISSUE #5077: PDS gets stuck on "Login complete, You are being redirected..." after OAuth login flow.
[I don't have consistent repro but this happens all the time right now across many different accounts and client applications.](https://bsky.app/profile/iame.li/post/3mnxvivk2uk23)

Image

That's just a screenshot of the page, I'm not stuck there. ~~Maybe it only happens for self-hosters and not on bsky.social, can't be 100% sure.~~ [danilotakagi.bsky.social says happening on a mushroom too](https://bsky.app/profile/danilotakagi.bsky.social/post/3mnxx6rkjik2f).

Nothing in the console. It looks like it's just not attempting the `window.location.href = whatever` at all. Happens on both native and web.

matthieusieben:
Fixes https://github.com/bluesky-social/atproto/issues/5077

### DIFF
--- a/.changeset/ten-teams-hang.md
+++ b/.changeset/ten-teams-hang.md
@@ -2,4 +2,4 @@
 '@atproto/dev-env': patch
 ---
 
-Temprarily ignore error when creating mock post replies
+Temporarily ignore error when creating mock post replies
--- a/.changeset/ten-teams-hang.md
+++ b/.changeset/ten-teams-hang.md
@@ -1,0 +1,5 @@
+---
+'@atproto/dev-env': patch
+---
+
+Temprarily ignore error when creating mock post replies
--- a/packages/oauth/oauth-provider-ui/src/authorization-page.tsx
+++ b/packages/oauth/oauth-provider-ui/src/authorization-page.tsx
@@ -88,11 +88,12 @@ function App() {
       // browser may restore the page state.
       //
       // On a related note, client processing of the token response should be a
-      // one time operation (because of nonce invalidation). So we should ensure
-      // that the navigation event does not happen more than once.
+      // one time operation (because of nonce invalidation). So we should do our
+      // best to prevent the navigation from being interrupted, or happening
+      // more than once.
       //
-      // This gets tricky as users may have a bad network connection, for which we
-      // should do the best we can to help them complete the login process.
+      // This gets tricky as users may have a bad network connection, for which
+      // we should do the best we can to help them complete the login process.
       //
       // We do this by first attempting to automatically redirect the user:
       redirectTo(url)
@@ -101,6 +102,12 @@ function App() {
       // a long(ish) pause in between these actions to allow the browser to
       // perform the navigation.
       setRedirectUrl(url)
+
+      // If, despite our best efforts, the client backend receives multiple
+      // redirect requests, it should handle them gracefully: Either by
+      // recognizing that the login process has already been completed for that
+      // user (through the use of a cookie), or by revoking previous credentials
+      // and triggering a new login process.
 
       // Prevent react from rendering the "redirecting..." view while the
       // browser is navigating by delaying the state update.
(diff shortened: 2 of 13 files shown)


.changeset/ten-teams-hang.md
```diff
@@ -0,0 +1,5 @@
+---
+'@atproto/dev-env': patch
+---
+
+Temprarily ignore error when creating mock post replies
```
devinivy:
```suggestion
Temporarily ignore error when creating mock post replies
```

packages/oauth/oauth-provider-ui/src/locales/fr/messages.po
```diff
@@ -53,8 +53,8 @@ msgstr "<0>Qu'est-ce qu'un compte Atmosphère ?</0><1>Un <2>compte Atmosphère</
...
 
-#: src/components/forms/input-handle-custom-instructions.tsx:55
-#: src/components/forms/input-handle-custom-instructions.tsx:70
+#: src/components/forms/input-handle-custom-instructions.tsx:54
+#: src/components/forms/input-handle-custom-instructions.tsx:69
```
devinivy, line 57: Only three `.po` files changed, and I feel like I usually see a ton more.  So just double-checking that this looks right to you.

packages/oauth/oauth-provider-ui/src/authorization-page.tsx
```diff
@@ -63,22 +63,58 @@ function App() {
...
+      // redirect requests, it should handle them gracefully: Either by
+      // recognizing that the login process has already been completed for that
+      // user (through the use of a cookie), or by revoking previous credentials
+      // and triggering a new login process.
+
+      // Prevent react from rendering the "redirecting..." view while the
+      // browser is navigating by delaying the state update.
+      setTimeout(() => setIsDone(true), 250)
+    },
+    [setRedirectUrl, setIsDone, redirectTo],
```
devinivy: I don't this setters from `useState` need to be passed as dependencies to `useCallback` (last I knew they conventionally aren't).

packages/oauth/oauth-provider-ui/src/authorization-page.tsx
```diff
@@ -115,6 +151,16 @@ function App() {
...
         <LayoutTitle title={msg`Login complete`}>
           <Trans>You are being redirected...</Trans>
+          <br />
+          {redirectUrl && (
+            <ButtonCooldown
+              startWithCooldown
+              cooldown={10}
+              action={() => redirectTo(redirectUrl)}
+            >
+              <Trans>Click here if nothing happens</Trans>
```
devinivy: How would you feel about:
```suggestion
              <Trans>Click here if you are not automatically redirected</Trans>
```

packages/oauth/oauth-provider-ui/src/locales/fr/messages.po
matthieusieben: I removed the languages we don't use

packages/oauth/oauth-provider-ui/src/authorization-page.tsx
```diff
@@ -63,22 +63,58 @@ function App() {
...
+      // This gets tricky as users may have a bad network connection, for which
+      // we should do the best we can to help them complete the login process.
+      //
+      // We do this by first attempting to automatically redirect the user:
+      redirectTo(url)
+
+      // If that fails, we show a link that they can click to continue. There is
+      // a long(ish) pause in between these actions to allow the browser to
+      // perform the navigation.
+      setRedirectUrl(url)
```
ThisIsMissEm, line 107: The comment here could be a little clearer that it's referring not to something between `redirectTo` and `setRedirectUrl` but rather that `setRedirectUrl` alone does not show the "done" page, which is triggered by the `setIsDone`

I think we could probably remove `setIsDone / isDone` and just have an `await sleep(250)`  between these two calls.

ThisIsMissEm (commented):
This seems fine to me, one minor comment on the comment/flow but otherwise it looks like it'd work.

PR #5099, opened 2026-06-15
Msi/ts 7 optimization

matthieusieben:
# Streamline build & verify pipeline

Follow-up to #4967 (TypeScript 7 / `tsgo` migration). That PR was a like-for-like compiler swap; this one restructures the pipeline around it so we can actually cash in on `tsgo`'s parallelism.

## Motivation

The current build does two things that work against `tsgo`:

1. **TypeScript runs twice.** Once during the topological per-package `build`, then again during `verify:types` against the root `tsconfig.json`.
2. **No real parallelism.** Each package spawns its own `tsgo` process, gated on its dependencies' emitted output. The graph serializes work the compiler is perfectly capable of doing in parallel inside a single process.

The blocker for collapsing this was codegen: `lex-cli` and `@atproto/lex-builder` had to be **built** before any package could run its `prebuild` / codegen step, which forced a full topological build order up front.

## Approach

Isolate the codegen tooling into a single dedicated `build:tooling` step that compiles only `@atproto/lex-builder` + `@atproto/lex-cli` (and their composite references) in one `tsgo --build`. That subgraph is tiny compared to the whole monorepo, so codegen no longer waits on a full topological build — it can then run **fully in parallel**. After codegen, run `tsgo --build` once at the root over the entire project-references graph, and drop `verify:types` entirely.

New pipeline:

```jsonc
"build:tooling": "tsgo --build ./packages/lex/lex-builder/tsconfig.build.json ./packages/lex-cli/tsconfig.build.json",
"prebuild":      "pnpm run build:tooling && pnpm run --parallel --workspace-concurrency Infinity 'prebuild'",
"build":         "tsgo --build tsconfig.json",
"postbuild":     "pnpm run --parallel build:ui",
// verify:types removed; verify is now just style + lint
```

Order of operations: `build:tooling` → `prebuild` → `build` (one root `tsgo --build tsconfig.json`) → `postbuild` (Vite/UI bundlers). The single root `tsgo --build` shares the incremental cache that `verify:types` used to populate, so `verify:types` is deleted.

`prebuild` will avoid running codegen of the slow and "old" `@atproto/lex-cli` (still needed by `@atproto/ozone` and `@atproto/api`) if it was recently built. Codegen can still be forces with the `codegen` script.

### Project-references buildout
For one root `tsgo --build` to validate everything:
- `composite: true` set globally on the shared base config.
- Every package now ships a `tsconfig.build.json` (composite, with explicit `references: [...]` to its workspace deps) plus a `tsconfig.test.json` for the test sources. The root `tsconfig.json` is a project-graph aggregator only.
- The old `tsconfig.tests.json` files were renamed to `tsconfig.test.json` across the workspace.
- `allowImportingTsExtensions: true` added to `tsconfig/jest.tsconfig.json` + `tsconfig/vitest.tsconfig.json` so test sources wire into the graph.
- `.eslintrc`'s import resolver now points at the `tsconfig.build.json` / `tsconfig.test.json` globs.

### Shared tsconfig reorganization
- `tsconfig/*.json` renamed to `tsconfig/*.tsconfig.json` (e.g. `base.json` → `base.tsconfig.json`).
- `nodenext.json` folded into `node.tsconfig.json`, which now extends `@tsconfig/node22` and sets `module`/`target` (`ES2023`) directly. `node.json` removed.
- New `tsconfig/verbatim.tsconfig.json` (`verbatimModuleSyntax: true`) applied to the `@atproto/lex` family + `syntax` + `lexicon`, shrinking their runtime import graph (`import type` elision). Adoption elsewhere is deferred.
- `noImplicitAny` enabled in the base config.

### Node version
- `.nvmrc` → `24` (default for local dev, CI build, CI verify).
- CI **tests** are pinned to `22.x` in `.github/workflows/repo.yaml`, overriding `.nvmrc`, so the matrix continues to validate our `engines.node: >=22` floor.

### Packaging & cleanup
- Explicit `files` in every `package.json` so npm tarballs bundle only `dist` + `CHANGELOG.md` / `README.md` (when present).
- Dropped the `ts-node` dev dependency; `bin/migration-create.ts` scripts run under `tsgo` / Node's type-stripping instead.
- Root `package.json` gains `"type": "module"`.
- Assorted correctness fixes surfaced by the stricter graph: `oauth-client-browser-example` converted CJS → ESM, removed an invalid `require` export from `lex-server`, added missing `@types` deps.

## Additional changes

- **`verbatimModuleSyntax` is enforced** in the `lex` family / `syntax` / `lexicon` — new code there needs explicit `import type`.
- **`noImplicitAny` is now enforced** repo-wide.
- **Node 24 for build, Node 22 for tests.** Contributors staying on 24 satisfy `>=22` for both; CI handles the split.


tsconfig/expo.tsconfig.json
```diff
@@ -1,13 +1,12 @@
...
-  "extends": ["./browser.json"],
+  "extends": ["./browser.tsconfig.json"],
   "compilerOptions": {
     "jsx": "react-native",
     "target": "ES2023",
     "module": "ESNext",
     "moduleResolution": "Bundler",
     "moduleDetection": "force",
     "esModuleInterop": true,
-    "skipLibCheck": true,
```
matthieusieben, line 11: already set to `true` in `base.tsconfig.json`

.changeset/mighty-mice-dream.md
```diff
@@ -0,0 +1,2 @@
+---
+---
```
devinivy: I think can remove this.

.claude/settings.json
```diff
@@ -28,7 +28,6 @@
       "Bash(pnpm --filter @atproto/*:*)",
       "Bash(pnpm -r:*)",
       "Bash(pnpm -w exec vitest:*)",
-      "Bash(pnpm -w run verify:types)",
```
devinivy, line 31: Is there something unsafe about this command?

packages/common-web/src/async.ts
```diff
@@ -1,4 +1,4 @@
-import { aggregateErrors, bailableWait } from './util.js'
+import { aggregateErrors, bailableWait } from './util.ts'
```
devinivy: It is going to be confusing that some packages use `.ts` and others `.js`.  (It's confusing that `.js` is used at all to be honest 😆 but that is another story)

Is there any way to work around or mitigate this?

packages/common/tsconfig.build.json
```diff
@@ -1,8 +1,14 @@
...
+    "composite": true,
     "rootDir": "./src",
     "outDir": "./dist",
   },
   "include": ["./src"],
+  "references": [
+    { "path": "../common-web/tsconfig.build.json" },
+    { "path": "../lex/lex-cbor/tsconfig.build.json" },
+    { "path": "../lex/lex-data/tsconfig.build.json" },
+  ],
```
devinivy, line 8: How are the `references` intended to be managed?

packages/lex/lex-document/tsconfig.test.json
```diff
@@ -1,8 +1,8 @@
 {
-  "extends": "../../../tsconfig/vitest.json",
+  "extends": "../../../tsconfig/vitest.tsconfig.json",
   "include": ["./tests", "./src/**/*.test.ts", "./src/core-js.d.ts"],
   "compilerOptions": {
-    "noImplicitAny": true,
```
devinivy, line 5: Did this just move somewhere else?

packages/lex/lex-json/tsconfig.build.json
```diff
@@ -1,11 +1,11 @@
 {
-  "extends": ["../../../tsconfig/isomorphic.json"],
+  "extends": ["../../../tsconfig/isomorphic-exec.tsconfig.json"],
   "include": ["./src"],
   "exclude": ["**/*.bench.ts", "**/*.test.ts"],
   "compilerOptions": {
-    "noImplicitAny": true,
-    "importHelpers": true,
```
devinivy, line 7: Are these no longer needed or did they move?

packages/oauth/oauth-client-browser-example/server.cjs
```diff
@@ -4,7 +4,8 @@
 
 const { once } = require('node:events')
 const { createServer } = require('node:http')
-const files = require('./dist/files.json')
+// @ts-expect-error
```
devinivy: Could we add a comment next to the ignore?

packages/ozone/tests/moderation.test.ts
```diff
@@ -749,7 +749,6 @@ describe('moderation', () => {
       // In the actual app, this will be instantiated and run on server startup
       const reverser = new EventReverser(
         network.ozone.ctx.db,
-        // @ts-expect-error Error due to circular dependency with the dev-env package
```
devinivy, line 752: Interesting!  How did this issue get solved?  Is it from the `references` somehow?

.nvmrc
```diff
@@ -1 +1 @@
-22
+26
```
devinivy: Another option that I didn't see mentioned is to use node v24, whose latest versions also support `using`.  An upside I see there is that we'll be developing against the nodejs version used by the production services.

I'd personally prefer to wait to move the production services to node v26 until it is active LTS, in preference of stability.  I could imagine some special circumstance prompting us to move earlier, but I don't see one at the moment.

packages/common/tsconfig.build.json
devinivy: I think I understand now that they need to be managed manually.  That is a little bit rough, seems easy to miss and may not be very clear how to fix issues when they arise.  What is your take on it?

packages/ozone/tests/moderation.test.ts
matthieusieben: yes. TS now recognizes the project structure and resolves to source files instead of dist.

packages/lex/lex-json/tsconfig.build.json
matthieusieben: `noImplicitAny` is part of the "parent" config (extend). `importHelpers` no longer needed with target `ES2023`

packages/lex/lex-document/tsconfig.test.json
matthieusieben: In parent config

packages/common/tsconfig.build.json
matthieusieben: I don't have a bullet proof solution. It does require manual curation. An option would be a custom script that keeps these up-to-date.

packages/common-web/src/async.ts
matthieusieben: I want to configure ESLINT to force the right extension. Will do this as a separate PR. Alternatively, we could move everything to `.ts`.

.claude/settings.json
matthieusieben: This particular script no longer exists. (replaced with `pnpm build`)

.nvmrc
matthieusieben: 24 is fine by me.

matthieusieben: I think it works better with better-sqlite3 anyways.

package.json
```diff
@@ -51,8 +52,9 @@
     "@jest/globals": "^30.0.0",
     "@swc/core": "^1.3.42",
     "@swc/jest": "^0.2.24",
+    "@tsconfig/node22": "^22.0.5",
     "@types/jest": "^30.0.0",
-    "@types/node": "^22.0.0",
+    "@types/node": "^22.x",
```
devinivy: What does this change do?

devinivy (approved):
🤩

package.json
matthieusieben: I used this to update the dependency.

PR #5101, opened 2026-06-16
fix(oauth-client): don't rely on AbortSignal.timeout (React Native compat)

### LINKED ISSUE #4477: AbortSignal.timeout is not a function, got undefined
**Describe the bug**

Using a recent version of `@atproto/oauth-client` in Expo + React Native causes this error to be returned instead of a successful login flow.

**To Reproduce**

Follow the steps in this blog https://samuel.leaflet.pub/3m3wtnioeuk26 to get a local environment:
1. Clone the repo and run `npm install`.
2. Make the suggested changes to get a publicly-accessible OAuth client manifest.
3. Attempt to test it end-to-end. The titular error is shown in red text.

**Expected behavior**

I would be able to complete the login flow successfully as depicted in the blog.

**Details**

- Operating system: iOS 16 Pro simulator
- React Native version: 0.81.4
- Expo version: 54
- `@atproto/oauth-client` version: 0.5.8

**Additional context**

The code referencing `AbortSignal.timeout` was only [added in 0.5.7](https://github.com/bluesky-social/atproto/pull/4247) so this seems to be a recent break.

It doesn't look like [React Native implements `AbortSignal.timeout`](https://github.com/facebook/react-native/issues/42042). 

This is [fairly trivial to polyfill though](https://github.com/DyuldinKS/abort-signal-polyfill/blob/main/src/staticMethods/timeout.ts) and works with as little as this code added to `utils/oauthClient.ts`:

```
AbortSignal.timeout = function (ms) {
  const controller = new AbortController();

  setTimeout(() => {
    controller.abort(new Error(`signal timed out after ${ms} ms`));
  }, ms);

  return controller.signal;
};
AbortSignal.prototype.throwIfAborted = function () {
  const { aborted, reason = "Aborted" } = this;
  if (!aborted) return;
  throw reason;
};
```

Image

ryanda9910:
## Problem

Using a recent `@atproto/oauth-client` (since `0.5.7`) in React Native / Expo breaks the OAuth login flow with:

```
TypeError: AbortSignal.timeout is not a function, got undefined
```

`AbortSignal.timeout` is a static method that is **not implemented in every runtime this package targets**. The package's own description lists React-Native as a target, but React Native does not provide `AbortSignal.timeout` ([facebook/react-native#42042](https://github.com/facebook/react-native/issues/42042)), so calling it directly throws at runtime.

Fixes #4477

## Root cause

Two call sites in the cross-platform `oauth-client` package call the static method directly:

- `src/oauth-server-agent.ts` — `signal: AbortSignal.timeout(10e3)` (issuer verification)
- `src/session-getter.ts` — `const signal = AbortSignal.timeout(30e3)` (session refresh deadline)

This is a **platform-compatibility / capability-assumption failure mode**: the code assumes a host API exists without feature-detecting it. (The server-only packages — `bsync`, `bsky`, `aws` — also use `AbortSignal.timeout`, but those run on Node ≥22 where it exists, so they're intentionally out of scope here. One PR = one concern.)

The package already follows a "detect-and-polyfill host gaps" pattern for the same RN reason — `src/index.ts` imports `core-js/es/symbol/dispose.js` to provide `Symbol.dispose`. `core-js` does **not** ship an `AbortSignal.timeout` polyfill, so the equivalent fix is a small internal helper.

## Fix

Add a `timeoutSignal(ms)` helper next to `combineSignals` in `src/util.ts` (the module that already centralises `AbortSignal` logic). It uses the native `AbortSignal.timeout` when present and otherwise falls back to an `AbortController` + `setTimeout`, aborting with a `TimeoutError` `DOMException` to mirror native semantics (and degrading to a plain `Error` when `DOMException` is also missing). Both call sites now use it.

This is **defensive capability detection** — the smallest change that removes the hard dependency on a non-universal host API without altering behaviour where the native method exists.

## Test evidence

The package had no test runner; this adopts **vitest** (the repo standard going forward, per the testing skill) and adds `src/util.test.ts` covering both the native path and the fallback path (RN simulated by temporarily removing `AbortSignal.timeout`).

Reproduction in a simulated RN environment (`delete globalThis.AbortSignal.timeout`):

```
# before — direct AbortSignal.timeout(10e3)
REPRO: TypeError: AbortSignal.timeout is not a function

# after — via timeoutSignal() against the built dist
OK: no TypeError. signal.aborted = false
After timeout: aborted = true | reason name = TimeoutError
```

Suite + types + lint:

```
 Test Files  1 passed (1)
      Tests  6 passed (6)

tsc --build tsconfig.json   # EXIT 0
prettier --check            # All matched files use Prettier code style!
eslint                      # clean
```

A changeset (`@atproto/oauth-client`: patch) is included.

---

Notes for reviewers:
- No new dependency (`vitest` is already used across the workspace; no `core-js` polyfill exists for this API).
- Server-side `AbortSignal.timeout` usages are intentionally untouched (Node-only, separate concern).


matthieusieben (approved):
Looks great. Thanks!

ryanda9910:
Thanks for the review and the quick turnaround, @matthieusieben! Glad the capability-detection fallback fits. Happy to tweak if anything comes up.

PR #5102, opened 2026-06-16
fix(handle-resolver): don't swallow unexpected errors in WellKnownHandleResolver

### LINKED ISSUE #4215: WellKnownHandleResolver eats the error thrown by safeFetch's unicast protection
**Describe the bug**

I was attempting to use the `NodeOAuthClient` without a custom `handleResolver` or `idResolver`, and had configured my server to respond to both `/.well-known/atproto-did` and `/.well-known/did.json` correctly, however my oauth calls still failed at handle resolution.

It turned out that `safeFetch` was throwing on unicast error, since for my local environment, `handle.princess.works` resolves to `127.0.0.1`, however this unicast error is swallowed by `WellKnownHandleResolver` so you have no idea that that's the reason why it's failing.

Additionally, passing a custom `fetch` to `NodeOAuthProvider` still somehow trips the unicast detector, even with:

```ts
import { safeFetchWrap } from "@atproto-labs/fetch-node";

const safeFetch = safeFetchWrap({
  allowIpHost: true,
  allowPrivateIps: true,
  allowImplicitRedirect: false,
  ssrfProtection: false,
  fetch: fetch,
});

export const createClient = async (ctx: Omit<AppContext, "oauthClient">) => {
  const url = ctx.service.url;

  return new NodeOAuthClient({
    fetch: fetch
    clientMetadata: { /* ... */ }
  })
}
```

**Expected behavior**

The unicast safeFetch error should probably bubble up from `WellKnownHandleResolver`, such that you can understand _why_ it's failing.

I'm not sure why passing a custom fetch implementation doesn't solve the problem though.

**Details**

- Operating system: MacOS Sequoia
- Node version: v24.7.0

ryanda9910:
## Problem

`WellKnownHandleResolver.resolve()` is documented (via the `HandleResolver` interface) to return `null` *only* when the handle does not resolve, and to throw on unexpected errors. In practice its `catch` block turned **every** failure other than an abort into `null`:

```ts
} catch (err) {
  options?.signal?.throwIfAborted()
  return null
}
```

As reported in #4215, when the `fetch` passed to the resolver is `safeFetchWrap`'s SSRF-protected fetch (the default in `AtprotoHandleResolverNode`, and therefore in `NodeOAuthClient`), the unicast guard throws *before any network exchange* if a handle resolves to a non-unicast address (e.g. `127.0.0.1` in local dev). That error — "Hostname is a non-unicast address" — was silently discarded, so OAuth/handle resolution failed with no indication of *why*.

## Root cause

The error-handling boundary collapsed several distinct error classes into a single `null` sentinel. This is the classic *exception-swallowing* anti-pattern (CWE-390, "Detection of Error Condition Without Action"): a security-relevant exception from the SSRF/unicast guard was caught and dropped, masking a misconfiguration and violating the resolver's own documented contract that `null` "should only be returned if no unexpected behavior occurred."

## Fix

Distinguish *expected* from *unexpected* failures in the `catch` block:

- The WHATWG `fetch` standard rejects with a `TypeError` for genuine network failures (DNS failure, connection refused, blocked redirect). Those mean the well-known endpoint is simply unreachable, so they still map to `null` — including the existing `redirect: 'error'` behavior.
- Any other error (the SSRF/unicast `FetchRequestError`, a misbehaving custom `fetch`, a programming error) is unexpected and is now re-thrown, surfacing the real cause.

The change is dependency-free and isomorphic — it relies only on the standard `TypeError` semantics of `fetch`, so the browser-targeting `@atproto-labs/handle-resolver` package gains no coupling to `fetch-node`.

## Test evidence

The package had no test setup; this PR adopts Vitest for it (per the repo's testing guide) and adds `well-known-handler-resolver.test.ts` covering resolve / non-DID body / network-failure-is-null / **unexpected-error-re-throws** (regression for #4215) / abort.

Before the fix (regression test fails):

```
× re-throws unexpected errors instead of swallowing them
Tests  1 failed | 4 passed (5)
```

After the fix:

```
Test Files  1 passed (1)
     Tests  5 passed (5)
```

`tsc --build`, `eslint`, and `prettier --check` all pass on the changed files. A changeset is included (`@atproto-labs/handle-resolver`, patch).

Fixes #4215


matthieusieben (approved):
Looks good

ryanda9910:
> Looks good

thanks

matthieusieben:
Could you please merge main & make sure CI passes ?

ryanda9910:
> Could you please merge main & make sure CI passes ?

sure on it

ryanda9910:
Merged the latest `main` and resolved the conflict (it was the TypeScript 7 / `tsgo` migration in `package.json`). The resolver change itself is unchanged. Ready for CI whenever you can approve the run 🙏

ryanda9910:
Local test run on the merged branch (`handle-resolver` package):

handle-resolver tests

matthieusieben:
Hey! Thanks again for helping us improve the quality of our code 🙏
Looks like CI isn't passing, would you mind giving it another look ? Thanks

ryanda9910:
Sorted now. The frozen-lockfile check was failing because I added vitest as a devDependency without updating pnpm-lock.yaml, so I committed the regenerated lockfile. One of the test shards also hung on an unrelated flake, re-running it cleared that. All checks are green now.

packages/internal/handle-resolver/src/internal-resolvers/well-known-handler-resolver.test.ts
```diff
@@ -0,0 +1,74 @@
...
+  it('propagates an abort error', async () => {
+    const controller = new AbortController()
+    controller.abort()
+    const resolver = new WellKnownHandleResolver({
+      fetch: fetchRejecting(new TypeError('aborted')),
+    })
+    await expect(
+      resolver.resolve('alice.test', { signal: controller.signal }),
+    ).rejects.toThrow()
+  })
```
matthieusieben, line 73: This test does not clearly shows what is causing the error:
- is the `resolver.resolve` throwing because of the aborted signal?
- is the `resolver.resolve` throwing because of the fetch rejection?
- Should the `resolver.resolve` throw the signal reason or the fetch error when both occur ?

matthieusieben:
I'm starting to have second thoughts about this.

On one hand we *do* want unexpected errors to bubble up. But on the other end, a domain resolving to a private IP range is an *expected* behavior. One could even argue that `WellKnownHandleResolver.resolve` should explicitly not resolve (return `null`) any hostname that resolve to a private IP range (this is not possible in the browser though).

Also, this PR hinges off the fact that `FetchError` extends `Error`, but an argument could be made that `FetchError` should actually extend `TypeError` (to be consistent with `fetch`'s erroring).

Finally, when both DNS and HTTP resolution are performed together (which is always the case), and DNS completes first successfully (which is most likely to be the case since HTTP is way slower), then the following lines *will swallow* any error that may have been thrown by the HTTP based resolution:

https://github.com/bluesky-social/atproto/blob/7d50e8395de1529b62b2193e6b8d88c0bdae0db1/packages/internal/handle-resolver/src/atproto-handle-resolver.ts#L60-L64

In my view, the `WellKnownHandleResolver.resolve` method implementation is so simple that, under the hypothesis that the `fetch()` implementation very well tested (through JS engines), we may consider that no implementation error will ever get thrown here, or at least that it is fine if we actually swallow them.

matthieusieben:
The `onError` hook as presented in #4215 seems like a better approach to me right now.

ryanda9910:
Makes sense, I agree the onError hook is the better fit. You and [Rex0Lux](https://github.com/Rex0Lux)'s sare right that a handle resolving to a private IP is an expected non-resolution, so returning null is the correct contract and my re-throw changes behavior that should stay as is.

[Rex0Lux](https://github.com/Rex0Lux) 's already sketched the onError shape and offered to open a PR for it, so I do not want to step on that work. If they are still planning to take it, I am happy to close this in favor of their PR. If it is open for someone to pick up, I can rework this branch to the onError approach instead (keep returning null, add the optional onError(err, context) callback, document that it should not throw) and keep the regression test from #4215. Whichever you and Rex0Lux prefer.

Rex0Lux:
> Makes sense, I agree the onError hook is the better fit. You and [Rex0Lux](https://github.com/Rex0Lux)'s sare right that a handle resolving to a private IP is an expected non-resolution, so returning null is the correct contract and my re-throw changes behavior that should stay as is.
> 
> [Rex0Lux](https://github.com/Rex0Lux) 's already sketched the onError shape and offered to open a PR for it, so I do not want to step on that work. If they are still planning to take it, I am happy to close this in favor of their PR. If it is open for someone to pick up, I can rework this branch to the onError approach instead (keep returning null, add the optional onError(err, context) callback, document that it should not throw) and keep the regression test from #4215. Whichever you and Rex0Lux prefer.

Thanks, I appreciate that. I'll have it done later today if that's okay.

Issue #3504, opened 2025-02-06
Add `User-Agent` header when fetching image blobs


snarfed:
Hi all! We consistently get reports of blank images and videos in bsky.app, and the most common root cause is, when the appview (CDN?) fetches blobs, the blob host refuses to serve them. This can happen for a range of reasons, but a common one seems to be that the appview doesn't include a `User-Agent` header in the HTTP request. Does that sound right?

GoToSocial is one common host that hits this, eg https://bsky.app/profile/isafashion.dev/post/3lgxro334cc2p :

> It's GoToSocial... Bluesky doesn't send UA hdr in req and gets HTTP 418. Tks!
> 
> Code:
> // UserAgent returns a gin middleware which aborts requests with
> // empty user agent strings, returning code 418 - I'm a teapot.
> 
> Log:
> level=INFO latency="11.38µs" userAgent="" method=GET statusCode=418 path=...

Any chance you all could add that header? Even anything non-blank would probably help a lot.

Related background in https://github.com/bluesky-social/atproto/discussions/3451 . Feel free to move this to indigo or anywhere else if appropriate. Thanks in advance!

snarfed:
Oh thanks! I'm not sure if it's still an issue; maybe the appview first fetched the blobs in that example before your fix. I'll close this for now, I can reopen if I see it recur.

leo60228:
The same issue appears to occur for profile pictures. Should a separate issue be opened?

snarfed:
It should be the same issue for all blobs, profile pictures or otherwise. And it's odd that we do keep seeing a trickle of complaints about this, even though it was supposedly fixed.

@devinivy I wonder, any chance the `User-Agent` is getting dropped on redirects, if the initial `getBlob` request redirects?

P33ry:
> I’ve confirmed this on my instance. It doesn’t seem to be a redirect issue. My instance 302s initial media requests to pre-signed URLs; after I added a fallback User-Agent for requests missing the header, Bsky could fetch and display the media correctly.
> 
> I also monitored the UAs for all requests after posting a specific piece of media:
> 
>     1. I didn’t see any requests with a `BskyAppView` UA. This suggests the fix in [Appview: add user-agent for feeds and resolving blobs #3450](https://github.com/bluesky-social/atproto/pull/3450) might not be working for this case.
> 
>     2. Bsky still seems to be sending requests without a User-Agent header:
> 
> 
> ```
[...]

PR #5116, opened 2026-06-19
Update undici to 8.5.0

matthieusieben:
Note that we still depend on `undici@6.19.8` through `@expo/cli@54.0.10`. That package should be updated separately.


devinivy (approved):
Looks good!  I believe this also addresses https://github.com/bluesky-social/atproto/issues/3504.  The old version of undici had a bug where headers defined with a `Map` were not being passed through on redirect.

PR #5118, opened 2026-06-19
Add repost and quotepost-only mutes

estrattonbailey:


estrattonbailey:
Both 1 and 2 are fine, ignore those

packages/bsky/src/api/app/bsky/graph/getMutes.ts
```diff
@@ -42,13 +42,15 @@ const skeleton = async (
...
+  const { dids, mutes, cursor } = await ctx.hydrator.dataplane.getMutes({
     actorDid: params.hydrateCtx.viewer,
     cursor: params.cursor,
     limit: params.limit,
   })
   return {
-    mutedDids: dids as DidString[],
+    mutedDids: (mutes.length
+      ? mutes.map((mute) => mute.did)
+      : dids) as DidString[],
```
estrattonbailey: Muted booleans are returned on `#viewerState`

packages/bsky/src/api/app/bsky/graph/muteActor.ts
```diff
@@ -3,12 +3,21 @@ import type { AppContext } from '../../../../context.js'
 import { app } from '../../../../lexicons/index.js'
 import { MuteOperation_Type } from '../../../../proto/bsync_pb.js'
 
+// The validation gate for mute kinds: everything downstream (bsync, the
+// dataplane) passes kind values through without validating them.
+const knownMuteKinds = new Set(['reposts', 'quoteposts'])
```
estrattonbailey: Dataplane ignores unknown values. Strict on write, permissive in storage, strict on read.

lexicons/app/bsky/actor/defs.json
```diff
@@ -195,7 +195,18 @@
       "type": "object",
       "description": "Metadata about the requesting account's relationship with the subject account. Only has meaningful content for authed requests.",
       "properties": {
-        "muted": { "type": "boolean" },
+        "muted": {
+          "type": "boolean",
+          "description": "Whether the account is fully muted, directly or via a mutelist. False when the mute is scoped to specific kinds; see mutedReposts and mutedQuoteposts."
+        },
+        "mutedReposts": {
```
blackmichael: I wonder if this should be `mutedOnlyReposts` and `mutedOnlyQuoteposts`

packages/bsky/proto/bsky.proto
```diff
@@ -545,6 +548,8 @@ message Relationships {
   string blocking_by_list = 6;
   string following = 7;
   string followed_by = 8;
+  bool muted_reposts = 9;
```
blackmichael: Same naming question/nit here

packages/bsky/src/views/index.ts
```diff
@@ -454,6 +466,8 @@ export class Views {
     const mutedByList = this.mutedByList(viewer, state)
     return {
       muted: !!(viewer.muted || mutedByList),
+      mutedReposts: !!viewer.mutedReposts,
+      mutedQuoteposts: !!viewer.mutedQuoteposts,
```
blackmichael: There might be some weirdness here wrt the exclusivity between the mute fields.
> Viewer state can violate the documented mutual-exclusivity invariant when a scoped direct mute overlaps a mute-list mute. `muted` is promoted to true by `mutedByList`, while the direct scoped flags are still emitted as true, so clients can receive `{ muted: true, mutedReposts: true }` or `{ muted: true, mutedQuoteposts: true }` even though the new lexicon explicitly says scoped fields are exclusive with `muted`. Suppress scoped flags whenever the effective account mute is full.

lexicons/app/bsky/actor/defs.json
estrattonbailey: Yep I'll align this with tango

packages/bsky/src/views/index.ts
estrattonbailey: I went back and forth on this previously and decided to leave it. But especially with the new designs and intention behind this, I think this is right, will update descriptions and make these exclusive.

lexicons/app/bsky/graph/getMutes.json
```diff
@@ -4,7 +4,7 @@
   "defs": {
     "main": {
       "type": "query",
-      "description": "Enumerates accounts that the requesting account (actor) currently has muted. Requires auth.",
+      "description": "Enumerates accounts that the requesting account (actor) currently has fully muted. Mutes scoped to specific kinds of content (only reposts, only quote posts) are not included. Responses may contain more items than the requested limit. Requires auth.",
```
estrattonbailey, line 7: I decided to fill pages on the backend rather than forcing the frontend to over-paginate. We only return full mutes on this endpoint.

blackmichael (approved):
LGTM

lexicons/app/bsky/actor/defs.json
```diff
@@ -195,7 +195,18 @@
...
-        "muted": { "type": "boolean" },
+        "muted": {
+          "type": "boolean",
+          "description": "Whether the account is fully muted, directly or via a mutelist. False when the mute is scoped to specific kinds; see mutedOnlyReposts and mutedOnlyQuoteposts."
+        },
+        "mutedOnlyReposts": {
+          "type": "boolean",
+          "description": "Whether the account's reposts are muted. Scoped mutes are exclusive with muted: this can be true while muted is false. If muted is true, this will be false."
+        },
+        "mutedOnlyQuoteposts": {
```
mfnboer, line 206: Is it on correct that it is Quoteposts and not QuotePosts?

PR #5121, opened 2026-06-21
Add configurable blob rate limit separate from global limits

ThisIsMissEm:
In order to support large account migrations, we had this change in the Eurosky fork of the atproto repo. I think this actually should be sent upstream, since it's a genuinely useful configuration change.

### DIFF
--- a/.changeset/metal-taxes-follow.md
+++ b/.changeset/metal-taxes-follow.md
@@ -1,0 +1,5 @@
+---
+'@atproto/pds': patch
+---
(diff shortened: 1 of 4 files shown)


devinivy:
This is just one of many rate limits a deployment may want to adjust.  If we're going to dig into this I'd recommend we look at this problem in context of all the other rate limits that also exist, and ensure we're going down the path of having a consistent way to configure all the pds rate limits.

PR #5142, opened 2026-06-24
Allow lex-client requests to opt out of default proxy headers

### LINKED ISSUE #5110: @atproto/lex: no way to remove the default 'service' proxy header
**Describe the bug**

When making xrpc requests with @atproto/lex, it is impossible to remove the default `service` proxy from requests, even when explicitly passing `undefined` to the service field. This issue also impacts default labelers set at client construction.

I want to make a request to a `com.atproto` endpoint, but to do that I need to unset the atproto-proxy header to do it in a spec compliant way, but I cannot unset the header for just one request unless I never set a default at all.

**To Reproduce**

```typescript
const client = new Client(session, {
  service: "example.com#service",
});

const profileRecord = await client.xrpcSafe(getRecord, {
  service: undefined,
  labelers: undefined,
  params: {
    collection: profile.$nsid,
    repo: session.did,
    rkey: "self",
  },
});

// request made with service header retained.
```

**Expected behavior**

The service and labelers headers should be unset for that specific request. Maybe allow specifying `null` to remove them explicitly?

**Details**

@atproto/lex 0.1.3

sjh9714:
## Summary

Fixes #5110.

`Client` default `service` and `labelers` options were always re-applied by `fetchHandler`, so a single XRPC request could not suppress the default `atproto-proxy` or `atproto-accept-labelers` headers. This adds an explicit `null` opt-out for per-request `service` and `labelers` while preserving the existing `undefined` behavior of inheriting defaults.

## Changes

- Allow `service: null` and `labelers: null` in XRPC request options.
- Carry that opt-out through the client boundary without sending the internal marker on the wire.
- Add regression coverage that verifies defaults still apply normally and can be disabled per request.

## Testing

- `pnpm --dir packages/lex/lex-client test -- tests/client.test.ts`
- `pnpm --dir packages/lex/lex-client run build`
- `pnpm exec prettier --check packages/lex/lex-client/src/client.ts packages/lex/lex-client/src/util.ts packages/lex/lex-client/tests/client.test.ts`

Note: I used Codex while preparing this change, reviewed the final diff, and ran the listed checks locally.

### DIFF
--- a/packages/lex/lex-client/src/client.ts
+++ b/packages/lex/lex-client/src/client.ts
@@ -280,6 +280,8 @@ export type UploadBlobOptions = Omit<XrpcOptions<typeof uploadBlob>, 'body'>
 
 export type GetBlobOptions = Omit<XrpcOptions<typeof getBlob>, 'params'>
 
+const xrpcDefaultHeaderOptOut = '__atproto_client_default_header_opt_out__'
+
 /**
  * Type-safe options for {@link Client.create}, combining record options with key requirements.
  * @typeParam T - The record schema type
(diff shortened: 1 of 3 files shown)


sjh9714:
Closing this because #5185 supersedes it and its implementation is now on main. It covers the same per-request service: null and labelers: null opt-out behavior as part of the broader client option rework.

devinivy:
Good call @sjh9714, and thanks for proposing the original fix!

PR #5147, opened 2026-06-24
Allow same-site authorization requests

joefitter:
Allows the authorization page to accept `Sec-Fetch-Site: same-site`, covering PDS/app deployments split across subdomains on the same site.

Tested:
- `corepack pnpm --filter @atproto/oauth-provider... build`


erlend-sh:
This fully resolves what #4618 started.

PR #5148, opened 2026-06-25
Fix node-fetch on Node24 and older

matthieusieben:
Since the recent update of undici to version 8 (https://github.com/bluesky-social/atproto/pull/5116), `node-fetch` has been broken on NodeJS 22 and NodeJS 24.

This is due to the fact that the interface of `dispatcher` option that can be passed to `fetch()` has significantly changed between Undici v6 and v8.

This regression was not detected in dev or tests because we explicitly disable SSRF in those environments.

The heart of the solution consists of loading a version of `undici` that matches the version used by the node process, ensuring that 1) we benefit from security updates in undici 2) we use a dispatcher with the right interface with `fetch()`.

**Before** (notice the node version at the top of each square)

Capture d’écran 2026-06-25 à 12 30 27

**After**

Capture d’écran 2026-06-25 à 12 22 17

### DIFF
--- a/.changeset/breezy-badgers-own.md
+++ b/.changeset/breezy-badgers-own.md
@@ -1,0 +1,5 @@
+---
+'@atproto-labs/fetch': patch
+---
+
+Fix support of NodeJS 24+
--- a/.changeset/breezy-badgers-own.md
+++ b/.changeset/breezy-badgers-own.md
@@ -2,4 +2,4 @@
 '@atproto-labs/fetch': patch
 ---
 
-Fix support of NodeJS 24+
+Fix support of NodeJS 24 and before
--- a/.changeset/slick-roses-sell.md
+++ b/.changeset/slick-roses-sell.md
@@ -1,0 +1,5 @@
+---
+'@atproto-labs/fetch-node': patch
+---
+
+unicastFetchWrap now cancels request bodies when throwing an error (to avoid resource leak)
--- a/.changeset/slick-roses-sell.md
+++ b/.changeset/slick-roses-sell.md
@@ -2,4 +2,4 @@
 '@atproto-labs/fetch-node': patch
 ---
 
-unicastFetchWrap now cancels request bodies when throwing an error (to avoid resource leak)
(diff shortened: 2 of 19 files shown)


.changeset/breezy-badgers-own.md
```diff
@@ -0,0 +1,5 @@
+---
+'@atproto-labs/fetch': patch
+---
+
+Fix support of NodeJS 24+
```
matthieusieben:
```suggestion
Fix support of NodeJS 24 and before
```

.changeset/slick-roses-sell.md
```diff
@@ -0,0 +1,5 @@
+---
+'@atproto-labs/fetch-node': patch
+---
+
+unicastFetchWrap now cancels request bodies when throwing an error (to avoid resource leak)
```
matthieusieben:
```suggestion
`unicastFetchWrap` now cancels request bodies when throwing an error (to avoid resource leak)
```

matthieusieben:
@claude review the PR as a whole, ignoring previous intermediary comments.

packages/internal/fetch-node/src/unicast.ts
```diff
@@ -1,77 +1,115 @@
-import dns, { LookupAddress } from 'node:dns'
-import { LookupFunction } from 'node:net'
+import type { LookupAddress, LookupOptions } from 'node:dns'
+import { lookup } from 'node:dns'
+import type { LookupFunction } from 'node:net'
 import ipaddr from 'ipaddr.js'
-import { Agent } from 'undici'
+import { Agent as Undici6Agent } from 'undici_v6' // NodeJS 22
```
ThisIsMissEm, line 5: These might be better as async imports?

matthieusieben: Maybe, but I don't want to make the function async. We could instantiate the `dispatcher` lazily (when fetch is first invoked) but then any error would get thrown in the wrong place.

Since this is meant to run on NodeJS servers exclusively, this is probably fine...

PR #5149, opened 2026-06-25
Update `pino` logger to version 11

matthieusieben:
Pino 8 is meant to support version of NodeJS [up to 20](https://getpino.io/#/docs/lts). Since `@atproto` packages now require NodeJS 22+, this means Pino needs to be updated.

Note that the only breaking changes (from 8 to 9 to 10) are only dropping support of older NodeJS versions


.changeset/tired-facts-occur.md
```diff
@@ -0,0 +1,9 @@
+---
+'@atproto/common': patch
+'@atproto/bsync': patch
+'@atproto/ozone': patch
+'@atproto/bsky': patch
+'@atproto/pds': patch
+---
+
+Update `pino` logger to version 11
```
matthieusieben:
```suggestion
Update `pino` logger to version 10
```

Issue #972, opened 2023-05-04
`ensureValidHandle` allows valid handles that are not possible in DNS


april:
`ensureValidHandle` allows handles up to 253 characters in length:

https://github.com/bluesky-social/atproto/blob/main/packages/identifier/src/handle.ts#L32

However, this fails to account for the `_atproto` prefix, meaning that it will say that some handles are valid (and try to validate them @ 253 characters), but they're not valid entries in DNS (because the total length of the DNS record would be 262 characters..

```python
>>> name = "a" * 63 + '.' + "a" * 63 + "." + "a" * 63 + "." + "a" * 48 + ".april.social"
>>> name
'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.april.social'
>>> len(name)
253
```

```shell
$ curl 'https://bsky.social/xrpc/com.atproto.identity.updateHandle' \
  -H 'authority: bsky.social' \
  -H 'accept: */*' \
  -H 'accept-language: en-US,en;q=0.9' \
  -H 'authorization: Bearer ...' \
  -H 'content-type: application/json' \
  --data-raw '{"handle":"aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.april.social"}'
{"error":"InvalidRequest","message":"External handle did not resolve to DID"}
```

But you can't actually create the DNS record of:
[...]

PR #5153, opened 2026-06-25
Update AT Protocol handle & NSID length limits to match DNS constraints

matthieusieben:
This is a draft fix for https://github.com/bluesky-social/atproto/issues/972

Tests & interop files need to be updated.

### DIFF
--- a/.changeset/nine-candles-thank.md
+++ b/.changeset/nine-candles-thank.md
@@ -1,0 +1,6 @@
+---
+'@atproto/oauth-provider-ui': patch
+'@atproto/syntax': patch
+---
+
+Update AT Protocol handle & NSID length limits to match DNS constraints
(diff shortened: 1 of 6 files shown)


bnewbold:
I feel mixed about this. I sort of feel like the semantics of handles (and NSID authorities) should be "any valid DNS hostname", with the caveat that some valid handles can not be resolved using certain resolution mechanisms. As opposed to defining the syntax as "things which can be resolved using certain resolution mechanisms". With my underlying feeling being that the resolution mechanisms might evolve over time, but we really don't want the syntax rules to change over time (because they retroactively make old records valid/invalid under lexicon validation). 

A lot of this comes down to the context of when the syntax is enforced. If you are trying to register a handle in the PDS UI, it would make sense to throw an error early for handles which won't be resolvable via DNS TXT. But if you are validating an AT-URI string during lexicon validation when bulk indexing records, it feels weird to reject entire records because the collection NSID wouldn't be resolvable under the current resolution rules.

The most important thing is to be broadly consistent about this though. If you make this change in the TS SDK we should also roll it out in the Go SDK, update the specs, announce to the ecosystem (for other SDKs), and even update the current IETF draft text which talks about NSID syntax (though not handle syntax).

matthieusieben:
I totally agree. This was more of an exercise to see how it would turn up & to have a discussion.

We could definitely come up with an HTTP based resolution mechanism for lexicon documents should that become an actual issue.

PR #5162, opened 2026-06-27
Reject duplicate BCP 47 variant subtags and extension singletons

### LINKED ISSUE #5161: BCP47_REGEXP accepts duplicate variant subtags and duplicate extension singletons, violating RFC 5646 §4.1
**Describe the bug**

`@atproto/syntax`'s `BCP47_REGEXP` (in `packages/syntax/src/language.ts`) accepts language tags that repeat a variant subtag or an extension singleton subtag, e.g. `isValidLanguage("de-DE-1901-1901")`, `isValidLanguage("en-a-foo-a-bar")`, and the case-insensitive equivalents `isValidLanguage("en-rozaj-ROZAJ")` / `isValidLanguage("en-u-co-phonebk-U-ca-buddhist")` all return `true`. Per RFC 5646 §4.1, "The same variant subtag MUST NOT be used more than once within a language tag" and "The same singleton (or letter, including the letter 'x') MUST NOT be used more than once within a language tag", so all four should return `false`. A single regex cannot detect repeats structurally; the fix is to add a Set-based post-validation step after the existing regex match.

**To Reproduce**

1. Clone the repo and check out `main` at commit `f2a0efe11`.
2. Append the following to `packages/syntax/tests/language.test.ts` (the repo uses vitest):

   ```ts
   import { describe, expect, it } from 'vitest'
   import { isValidLanguage } from '../src/index.js'

   describe('RFC 5646 §4.1 MUST-NOT duplicate subtags', () => {
     it('rejects duplicate variant subtags', () => {
       expect(isValidLanguage('de-DE-1901-1901')).toBe(false)
     })

     it('rejects duplicate variant subtags (case-insensitive)', () => {
       expect(isValidLanguage('en-rozaj-ROZAJ')).toBe(false)
     })

     it('rejects duplicate extension singletons', () => {
       expect(isValidLanguage('en-a-foo-a-bar')).toBe(false)
     })

     it('rejects duplicate extension singletons (case-insensitive)', () => {
       expect(isValidLanguage('en-u-co-phonebk-U-ca-buddhist')).toBe(false)
     })

     it('still accepts distinct variants (regression guard)', () => {
       expect(isValidLanguage('sl-rozaj-biske')).toBe(true)
     })

     it('still accepts multiple distinct singletons (regression guard)', () => {
       expect(isValidLanguage('en-u-co-phonebk-t-en-US')).toBe(true)
     })
   })
   ```

3. Run `pnpm install && pnpm test` inside `packages/syntax`.
4. Observe that the first four cases fail (`isValidLanguage` returns `true` instead of `false`); the last two regression-guard cases pass.

**Expected behavior**

All four malformed tags should be rejected, and both regression-guard tags should remain accepted:

| Input                              | Current (`isValidLanguage`) | Expected per RFC 5646 §4.1 |
|------------------------------------|-----------------------------|-----------------------------|
| `"de-DE-1901-1901"`                | `true` ❌                   | `false` (duplicate variant) |
| `"en-rozaj-ROZAJ"`                 | `true` ❌                   | `false` (duplicate variant, case-insensitive) |
| `"en-a-foo-a-bar"`                 | `true` ❌                   | `false` (duplicate singleton `a`) |
| `"en-u-co-phonebk-U-ca-buddhist"`  | `true` ❌                   | `false` (duplicate singleton `u`, case-insensitive) |
| `"sl-rozaj-biske"`                 | `true`                      | `true` (distinct variants) |
| `"en-u-co-phonebk-t-en-US"`        | `true`                      | `true` (distinct singletons) |

**Details**

- Operating system: macOS 26.5.1 (Darwin 25.5.0)
- Node version: v22.16.0 (project's minimum is `>=22` per root `package.json`); the bug is regex/logic-level and reproduces identically on any V8 version.
- Verified against `bluesky-social/atproto` `main` at commit `f2a0efe11`, by evaluating the exported `BCP47_REGEXP` directly with `RegExp.prototype.test` and by running the reference fix below over the same inputs.

**Additional context**

### Spec citations

**RFC 5646 §4.1 (Choice of Language Tag):**

> The same variant subtag MUST NOT be used more than once within a language tag.

> The same singleton (or letter, including the letter 'x') MUST NOT be used more than once within a language tag.

**RFC 5646 §2.1.1 (Case Considerations):**

> At all times, language tags and their subtags ... are to be treated as case insensitive

The duplicate check therefore needs to be case-insensitive.

### Suggested fix

A single regular expression cannot detect a duplicate subtag structurally, because JavaScript's named captures retain only the last occurrence of a repeated group — `match.groups.variant` only sees the last variant, not the full list. The fix is to add a Set-based post-validation step after the existing regex match.

Reference implementation (drop-in replacement for `isValidLanguage`):

```js
function isValidLanguage(input) {
  const m = input.match(BCP47_REGEXP)
  if (!m?.groups) return false
  // Grandfathered tags and privateuse-only tags have no variants or extension singletons.
  if (m.groups.grandfathered) return true
  if (m.groups.privateUseB) return true

  const subtags = input.split('-')
  let i = 0

  // language (subtags[0])
  i++

  // extlang section: 0-3 subtags of 3 letters (only if language is 2-3 letters)
  if (subtags[0].length === 2 || subtags[0].length === 3) {
    let count = 0
    while (count < 3 && i < subtags.length && /^[A-Za-z]{3}$/.test(subtags[i])) {
      i++; count++
    }
  }

  // Optional script (4 letters)
  if (i < subtags.length && /^[A-Za-z]{4}$/.test(subtags[i])) i++
  // Optional region (2 letters or 3 digits)
  if (i < subtags.length && /^([A-Za-z]{2}|[0-9]{3})$/.test(subtags[i])) i++

  // 0+ variants — dedup case-insensitively per RFC 5646 §2.1.1
  const seenVariants = new Set()
  while (i < subtags.length && /^([A-Za-z0-9]{5,8}|[0-9][A-Za-z0-9]{3})$/.test(subtags[i])) {
    const key = subtags[i].toLowerCase()
    if (seenVariants.has(key)) return false
    seenVariants.add(key)
    i++
  }

  // 0+ extensions — singleton (1 char, excluding x/X) + 1..n extension subtags.
  // Dedup singletons case-insensitively.
  const seenSingletons = new Set()
  while (i < subtags.length && /^[0-9A-WYZa-wyz]$/.test(subtags[i])) {
    const singleton = subtags[i].toLowerCase()
    if (seenSingletons.has(singleton)) return false
    seenSingletons.add(singleton)
    i++
    while (i < subtags.length && /^[A-Za-z0-9]{2,8}$/.test(subtags[i])) i++
  }

  // Optional privateuse (x-...) — RFC 5646 §4.1 forbids reusing the singleton 'x' itself,
  // which is already covered by the singleton dedup above. No further check needed.
  return true
}
```

`parseLanguageString` should be patched the same way (return `null` instead of `false` on the duplicate paths) so the two functions remain consistent.

*Side-effect characterization:* evaluating this implementation against the 4 claim cases, 2 regression guards, all 11 existing cases in `packages/syntax/tests/language.test.ts`, and 5 invalid sanity cases (22 total) shows the four claim cases flip from `true` to `false` (the intended fix) and every other case is unchanged.

| Input                              | Baseline | Fix       | Per RFC 5646 |
|------------------------------------|----------|-----------|--------------|
| `"de-DE-1901-1901"`                | true     | **false** | false        |
| `"en-rozaj-ROZAJ"`                 | true     | **false** | false        |
| `"en-a-foo-a-bar"`                 | true     | **false** | false        |
| `"en-u-co-phonebk-U-ca-buddhist"`  | true     | **false** | false        |
| `"sl-rozaj-biske"`                 | true     | true      | true         |
| `"en-u-co-phonebk-t-en-US"`        | true     | true      | true         |
| (all 11 existing test cases + 5 sanity invalid cases) | unchanged | unchanged | — |

For reference, `swift-atproto` implements the same RFC 5646 §4.1 dedup at `Sources/SwiftAtproto/StringFormats/Language.swift:191-215` using `seenVariants` / `seenSingletons` sets keyed on lowercased subtags — the JS reference above mirrors that approach.

### Severity

Moderate. Real-world Bluesky data can plausibly contain malformed tags from third-party clients, conversion tools, or copy-paste errors. A reference implementation that accepts them lets such tags propagate through the PDS / Relay path into records, where strict consumers (e.g. `swift-atproto`) will reject them downstream. Catching them at the producer / validator boundary is the standard place to enforce RFC 5646 §4.1.

### Background

This regex was added in #1219 (commit `3da032487`, 2023-06-23) to implement post record language tagging. The accompanying tests covered well-formed tags only; no duplicate-subtag cases were exercised, so the gap was never visible. The regex was later moved verbatim to its current location in #4366 (commit `261968fd6`, 2025-11-24, "New TS SDK") without behavioral change.

The original code comment cites only RFC 5646 §2.1 (Syntax), not §4.1 (Choice of Language Tag). §4.1's "MUST NOT be used more than once" requirements are structurally outside what a single regular expression can express in this style of grammar transliteration, so they were silently dropped. The fix is a post-match validation step rather than a regex change.

Related: #5160 — uppercase `X-` privateuse subtag (same file, sibling RFC 5646 compliance gap).

nnabeyang:
This PR rejects duplicate variant subtags and duplicate extension singleton subtags in `isValidLanguage` / `parseLanguageString` via a Set-based post-validation check, per RFC 5646 §4.1.

Closes #5161.

### DIFF
--- a/.changeset/dedup-bcp47-subtags.md
+++ b/.changeset/dedup-bcp47-subtags.md
@@ -1,5 +1,5 @@
 ---
-'@atproto/syntax': patch
+'@atproto/syntax': minor
 ---
 
-Reject duplicate variant subtags and duplicate extension singleton subtags in `isValidLanguage` / `parseLanguageString`, per RFC 5646 §4.1.
+`parseLanguageString` now rejects BCP 47 tags with duplicate variant subtags or duplicate extension singleton subtags, per RFC 5646 §4.1. `isValidLanguage` continues to check only well-formed syntax (§2.1) and is unchanged.
--- a/.changeset/dedup-bcp47-subtags.md
+++ b/.changeset/dedup-bcp47-subtags.md
@@ -1,0 +1,5 @@
+---
+'@atproto/syntax': patch
+---
+
+Reject duplicate variant subtags and duplicate extension singleton subtags in `isValidLanguage` / `parseLanguageString`, per RFC 5646 §4.1.
--- a/.changeset/language-strict-lenient.md
+++ b/.changeset/language-strict-lenient.md
@@ -1,0 +1,7 @@
+---
+'@atproto/common': minor
+'@atproto/lex-schema': minor
+'@atproto/lex': minor
+---
+
+Language-tag string-format validation now has strict and lenient variants, matching the pattern already used for `at-uri` and `datetime`. `isLanguageString` (and the default `language` string-format verifier) is now strict and rejects BCP 47 tags that violate RFC 5646 §4.1 (e.g. duplicate variant subtags or duplicate extension singletons). Pass `{ strict: false }` to fall back to the lenient (syntax-only) verifier, `isLanguageStringLenient`.
--- a/packages/lex/lex-schema/src/core/string-format.ts
+++ b/packages/lex/lex-schema/src/core/string-format.ts
@@ -20,6 +20,7 @@ import {
   isValidRecordKey,
   isValidTid,
   isValidUri,
+  parseLanguageString,
 } from '@atproto/syntax'
 import type { CheckFn } from '../util/assertion-util.js'
 
@@ -126,10 +127,25 @@ export type {
 /**
  * Type guard that checks if a value is a valid BCP-47 language tag.
  *
+ * Strict: rejects tags whose syntax is well-formed but violate semantic
+ * constraints from RFC 5646 §4.1 (e.g. repeated variant subtags or repeated
+ * extension singletons). Use {@link isLanguageStringLenient} to accept those.
+ *
  * @param value - The value to check
  * @returns `true` if the value is a valid language string
  */
-export const isLanguageString = isValidLanguage as CheckFn<LanguageString>
+export const isLanguageString = ((v) =>
+  parseLanguageString(v) !== null) as CheckFn<LanguageString>
+
+/**
+ * Lenient version of {@link isLanguageString} that only checks well-formed
+ * BCP 47 syntax (RFC 5646 §2.1) and does not enforce semantic constraints
+ * from §4.1.
+ *
+ * @see {@link isLanguageString}
+ */
+export const isLanguageStringLenient =
+  isValidLanguage as CheckFn<LanguageString>
 /**
  * A BCP-47 language tag string.
  *
@@ -242,7 +258,7 @@ const stringFormatVerifiers: {
   datetime: [isDatetimeString, isDatetimeStringLenient],
   did: [isDidString],
   handle: [isHandleString],
-  language: [isLanguageString],
+  language: [isLanguageString, isLanguageStringLenient],
   nsid: [isNsidString],
   'record-key': [isRecordKeyString],
   tid: [isTidString],
--- a/packages/lex/lex-schema/src/schema/string.test.ts
+++ b/packages/lex/lex-schema/src/schema/string.test.ts
@@ -514,6 +514,34 @@ describe('StringSchema', () => {
       const result = schema.safeParse('not valid')
       expect(result.success).toBe(false)
     })
+
+    it('rejects tags with duplicate variant subtags (RFC 5646 §4.1)', () => {
+      const result = schema.safeParse('de-DE-1901-1901')
+      expect(result.success).toBe(false)
+    })
+
+    it('rejects tags with duplicate extension singletons (RFC 5646 §4.1)', () => {
+      const result = schema.safeParse('en-a-foo-a-bar')
+      expect(result.success).toBe(false)
+    })
+
+    describe('loose validation', () => {
+      it('accepts tags with duplicate variant subtags', () => {
+        const result = schema.safeParse('de-DE-1901-1901', { strict: false })
+        expect(result.success).toBe(true)
+      })
+
+      it('accepts tags with duplicate extension singletons', () => {
+        const result = schema.safeParse('en-a-foo-a-bar', { strict: false })
+        expect(result.success).toBe(true)
+      })
+
+      it('still rejects syntactically invalid strings', () => {
+        expect(schema.safeParse('not valid', { strict: false }).success).toBe(
+          false,
+        )
+      })
+    })
   })
 
   describe('format: tid', () => {
--- a/packages/syntax/src/language.ts
+++ b/packages/syntax/src/language.ts
@@ -12,8 +12,80 @@ export type LanguageTag = {
   privateUse?: string
 }
 
-export function parseLanguageString(input: string): LanguageTag | null {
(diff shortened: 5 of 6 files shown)


matthieusieben (approved):
Looks good. Thank you.

Could you please add the valid/invalid tests cases in a PR against:

https://github.com/bluesky-social/atproto-interop-tests/blob/main/syntax/language_syntax_valid.txt
https://github.com/bluesky-social/atproto-interop-tests/blob/main/syntax/language_syntax_invalid.txt

Could you also add these two interop tests as part of this PR in:

https://github.com/bluesky-social/atproto/tree/main/interop-test-files/syntax

(and update the tests to actually use these interop files).

Would you mind also adding the `X-` tests from https://github.com/bluesky-social/atproto/pull/5169 as part of this?

Thank you very much.

nnabeyang:
@matthieusieben Re-reading the Lexicon spec for the language format, I read it as saying tags that are "well-formed" but not "valid" (e.g. RFC 5646 §4.1's duplicate-subtag rule) should still be accepted, not rejected — this PR's check seems too strict. Should I rework it to only enforce well-formedness?

nnabeyang:
Opened bluesky-social/atproto-interop-tests#17 with the requested valid/invalid fixtures (duplicate variant/singleton subtags, and the uppercase X- case from #5169).

matthieusieben:
Humm you are right. The way we started dealing with these is to have two validation modes strict vs. lenient. Only a 2 string formats have this distinction currently (`at-uri` and `datetime`):

https://github.com/bluesky-social/atproto/blob/4f1e20387bb2a212e263590ac7a3458e8f939091/packages/lex/lex-schema/src/core/string-format.ts#L239-L249

This translates into 3 interop files like (see datetime for example):

- [format]_syntax_valid.txt
- [format]_syntax_invalid.txt
- [format]_parse_invalid.txt (Valid syntax, invalid value)

So the better call here might be to:

1) keep the previous behavior of `isValidLanguage` (regexp only)
2) update `parseLanguageString` to use your logic (parsing should fail for invalid values)
3) Update [string-format.ts](https://github.com/bluesky-social/atproto/blob/main/packages/lex/lex-schema/src/core/string-format.ts#L133):
   - Define `isLanguageStringLenient = isValidLanguage`
   - Define `isLanguageString` to be based on `parseLanguageString` (strict validator)
   - Update `stringFormatVerifiers` to use both strict/lenient versions
4) Create a minor version changeset (that explains that parsing & strict validation of languages now rejects invalid tags) for:
   - `@atproto/common`
   - `@atproto/lex-schema`
   - `@atproto/lex`

matthieusieben (approved):
Looking good

PR #5163, opened 2026-06-27
Propagate skeletonReasonPin from custom feed generators

### LINKED ISSUE #7597: Custom feeds do not show a pinned post as "pinned" in the app interface
### Steps to Reproduce

We host a lot of custom feeds, and one of our features is to allow feed owners to pin any post to the top of their feed, much like how you can do on your own profile. According to the [documentation here](https://docs.bsky.app/docs/api/app-bsky-feed-get-feed-skeleton) , we should be able to reply with a "reason" indicating that a post is pinned. 

We are returning this information (hopefully in the correct format) but our feeds do not display the "pinned" status in the app interface:

Image

I can't see any information about "reason" coming through the API connection from the Bluesky website, so not sure if we are sending the wrong thing back, or if it's just not implemented for custom feeds:

Screenshot taken from the [Birds! 🦉 feed](https://bsky.app/profile/daryllmarie.bsky.social/feed/aaagllxbcbsje):

Image

### Attachments

_No response_

### What platform(s) does this occur on?

Web (Desktop), Android

### Device Info

_No response_

### What version of the app are you using?

Build version: 1.96.6; Bundle info: 5090426 (prod); Bundle date: 25012002; Platform: web; Platform version: undefined

### Additional Information

_No response_

spacecowboy37:
When a feed generator returns `app.bsky.feed.defs#skeletonReasonPin` as the `reason` for a skeleton item, the AppView silently drops it. The `skeletonFromFeedGen` function in `getFeed.ts` only maps `skeletonReasonRepost` to a `FeedItem` field — `skeletonReasonPin` is never mapped to `authorPinned: true`, so `feedViewPost()` never emits `ReasonPin` in the hydrated response. The client already renders `ReasonPin` correctly when it receives it.

Closes bluesky-social/social-app#7597


mozzius (approved):
LGTM! Thanks for the fix, this was the originally intended behaviour

PR #5169, opened 2026-06-30
fix(syntax): accept uppercase X- private-use subtags in BCP 47 tags

### LINKED ISSUE #5160: BCP47_REGEXP rejects uppercase "X-" privateuse subtag, violating RFC 5646 §2.1.1
**Describe the bug**

`@atproto/syntax`'s `BCP47_REGEXP` (in `packages/syntax/src/language.ts`) rejects well-formed BCP 47 language tags that use an uppercase `X-` private-use subtag, e.g. `isValidLanguage("X-fr-CH")` and `isValidLanguage("de-X-foo")` both return `false`. Per RFC 5234 §2.3 and RFC 5646 §2.1.1, the ABNF literal `"x"` is case-insensitive and private-use subtags MUST be treated as case-insensitive, so both should be accepted. The cause is the two `x` literals inside `BCP47_REGEXP` and the absence of an `i` flag on the regex.

**To Reproduce**

1. Clone the repo and check out `main` at commit `f2a0efe11`.
2. Append the following to `packages/syntax/tests/language.test.ts` (the repo uses vitest):

   ```ts
   import { describe, expect, it } from 'vitest'
   import { isValidLanguage } from '../src/index.js'

   describe('RFC 5646 §2.1.1 case-insensitive private-use', () => {
     it('accepts uppercase X- privateuse-only tag', () => {
       // RFC 5234 §2.3: ABNF literal "x" is case-insensitive.
       // RFC 5646 §2.1.1: private-use subtags are case-insensitive.
       expect(isValidLanguage('X-fr-CH')).toBe(true)
     })

     it('accepts uppercase -X- privateuse suffix on langtag', () => {
       expect(isValidLanguage('de-X-foo')).toBe(true)
     })

     it('still accepts lowercase forms (regression guard)', () => {
       expect(isValidLanguage('x-fr-CH')).toBe(true)
       expect(isValidLanguage('de-x-foo')).toBe(true)
     })
   })
   ```

3. Run `pnpm install && pnpm test` inside `packages/syntax`.
4. Observe that the first two cases fail (`isValidLanguage` returns `false` instead of `true`); the third case passes.

**Expected behavior**

All four inputs should be accepted as well-formed per RFC 5646:

| Input         | Current (`isValidLanguage`) | Expected per RFC 5646 |
|---------------|-----------------------------|-----------------------|
| `"X-fr-CH"`   | `false` ❌                  | `true`                |
| `"de-X-foo"`  | `false` ❌                  | `true`                |
| `"x-fr-CH"`   | `true`                      | `true`                |
| `"de-x-foo"`  | `true`                      | `true`                |

**Details**

- Operating system: macOS 26.5.1 (Darwin 25.5.0)
- Node version: v22.16.0 (project's minimum is `>=22` per root `package.json`); the bug is regex-level and reproduces identically on any V8 version.
- Verified against `bluesky-social/atproto` `main` at commit `f2a0efe11`, by evaluating the exported `BCP47_REGEXP` directly with `RegExp.prototype.test`.

**Additional context**

### Spec citations

**RFC 5646 §2.1 (Syntax):**

```
privateuse    = "x" 1*("-" (1*8alphanum))
```

**RFC 5234 §2.3 (Terminal Values for ABNF):**

> ABNF strings are case insensitive and the character set for these strings is US-ASCII.
>
> Example: `rulename = "abc"` matches `"abc", "Abc", "aBc", "abC", "ABc", "aBC", "AbC", and "ABC"`.

(Compare with the `singleton` production, which uses hex notation `%x30-39 / %x41-57 / ...` to make case sensitivity explicit. The `privateuse` rule uses a quoted string, so case insensitivity is intended.)

**RFC 5646 §2.1.1 (Case Considerations):**

> At all times, language tags and their subtags, **including private use and extensions**, are to be treated as case insensitive: there exist conventions for the capitalization of some of the subtags, but these MUST NOT be taken to carry meaning.

### Suggested fix (two options)

Both options pass all current cases in `packages/syntax/tests/language.test.ts`. The choice is about how broadly to align with RFC 5646 §2.1.1.

**Option 1 — minimal: replace `x` literals with `[xX]`**

Change the two `x` literals inside `BCP47_REGEXP` (one in `privateUseA`, one in `privateUseB`):

```diff
- (-(?<privateUseA>x(-[A-Za-z0-9]{1,8})+))?
+ (-(?<privateUseA>[xX](-[A-Za-z0-9]{1,8})+))?
...
- (?<privateUseB>x(-[A-Za-z0-9]{1,8})+)
+ (?<privateUseB>[xX](-[A-Za-z0-9]{1,8})+)
```

*Pros:* smallest possible diff, no impact on any other subtag.

*Cons:* addresses only the `x` literal. (Currently `x` is the only case-insensitive ABNF literal in the regex, so this is not a real concern today.)

*Side-effect characterization:* evaluating the patched regex against the 11 cases in the existing test file plus the four reproducer cases shows exactly one behavior change vs. the baseline (the intended fix):

| Input            | Baseline | Option 1 | Per RFC 5646 |
|------------------|----------|----------|--------------|
| `"X-fr-CH"`      | false    | **true** | true         |
| `"de-X-foo"`     | false    | **true** | true         |
| (all 11 existing test cases) | unchanged | unchanged | — |

**Option 2 — spec-aligned: add the `i` flag to the whole regex**

```diff
- const BCP47_REGEXP = /^...$/
+ const BCP47_REGEXP = /^...$/i
```

*Pros:* conceptually aligned with RFC 5646 §2.1.1, which makes the entire tag case-insensitive. Future case-insensitive ABNF literals (should any be added) are covered automatically.

*Cons:* widens the change beyond `privateuse`. Specifically, the grandfathered alternation currently lists only lowercase forms (`en-GB-oed`, `i-klingon`, …), so mixed-case grandfathered tags are rejected today and accepted with `i`:

| Input              | Baseline | Option 2 | Per RFC 5646 §2.1.1 |
|--------------------|----------|----------|---------------------|
| `"I-Klingon"`      | false    | **true** | should be true      |
| `"EN-GB-OED"`      | false    | **true** | should be true      |
| `"SGN-be-NL"`      | false    | **true** | should be true      |
| `"X-fr-CH"`        | false    | **true** | true                |
| `"de-X-foo"`       | false    | **true** | true                |
| (all 11 existing test cases) | unchanged | unchanged | — |

So Option 2 *also* silently fixes mixed-case grandfathered tags; this is arguably more spec-correct but is a broader behavior change. None of the existing tests break either way.

Other subtags (language / script / region / extlang / variant / extension) already use `[A-Za-z]` / `[A-Za-z0-9]` character classes, so the `i` flag does not alter their case behavior. Extension singletons `[0-9A-WY-Za-wy-z]` still correctly exclude `x` / `X` because `[A-W]` with `i` does not include `X` (the range matches the case-fold members of A–W, not X).

I have no preference between the two; whichever fits the project's stance on RFC compliance is fine.

### Severity

Low. Real-world clients (including the Bluesky `social-app`) do not generate uppercase `X-…` private-use tags, so wire-format compatibility impact is minimal. Filing this for spec compliance and reference implementation correctness.

### Background

This regex was added in #1219 (commit `3da032487`, 2023-06-23) to implement post record language tagging. The tests added alongside it covered lowercase `x-fr-CH` only, so the uppercase `X-` case was never exercised. It was later moved verbatim to its current location in #4366 (commit `261968fd6`, 2025-11-24, "New TS SDK") without any behavioral change.

The original code comment cites only RFC 5646 §2.1 (Syntax), not §2.1.1 (Case Considerations) or RFC 5234 §2.3 (which establishes that ABNF quoted-string literals are case-insensitive). The lowercase `x` looks like a faithful but incomplete transliteration of the ABNF rather than an intentional restriction — consistent with the fact that the same regex carefully distinguishes case where ABNF uses explicit hex ranges (e.g. `[0-9A-WY-Za-wy-z]` for extension singletons, which deliberately exclude `x` / `X`).

ryanda9910:
## Problem

`@atproto/syntax`'s `isValidLanguage` and `parseLanguageString` reject well-formed BCP 47 language tags whose private-use subtag is written with an uppercase `X`:

| Input        | Before      | Expected |
|--------------|-------------|----------|
| `X-fr-CH`    | `false` ❌  | `true`   |
| `de-X-foo`   | `false` ❌  | `true`   |
| `x-fr-CH`    | `true`      | `true`   |
| `de-x-foo`   | `true`      | `true`   |

## Root cause

`BCP47_REGEXP` in `packages/syntax/src/language.ts` hard-codes a lowercase `x` literal in both the `privateUseA` and `privateUseB` groups, and the regex carries no `i` flag.

Per RFC 5646 §2.1.1, language tags and their subtags, including private use, are case-insensitive. RFC 5234 §2.3 also makes the ABNF quoted-string literal `"x"` in the `privateuse = "x" 1*("-" (1*8alphanum))` production case-insensitive. So an uppercase `X-` private-use subtag is well-formed and should be accepted.

The lowercase `x` looks like a faithful but incomplete transliteration of the ABNF: where the spec does intend case sensitivity it uses explicit hex ranges, e.g. the extension singleton `[0-9A-WY-Za-wy-z]`, which deliberately excludes `x`/`X`. Private use uses a quoted literal, so it is not case sensitive.

## Fix

Replace the two `x` literals with the character class `[xX]`:

```diff
- (-(?<privateUseA>x(-[A-Za-z0-9]{1,8})+))?)|(?<privateUseB>x(-[A-Za-z0-9]{1,8})+))$/
+ (-(?<privateUseA>[xX](-[A-Za-z0-9]{1,8})+))?)|(?<privateUseB>[xX](-[A-Za-z0-9]{1,8})+))$/
```

This is the minimal, one-concern change for the reported bug and it is purely additive:

- Every tag that was valid before stays valid (verified: all 868 existing `packages/syntax` tests pass).
- The extension-singleton class `[0-9A-WY-Za-wy-z]` is untouched, so `x`/`X` are still correctly excluded there.
- All other subtags already use `[A-Za-z]` / `[A-Za-z0-9]` classes, so case behavior elsewhere is unchanged.

The issue also proposed a broader Option 2 (adding the `i` flag), which would additionally accept mixed-case grandfathered tags such as `I-Klingon`. That is a wider behavior change than the reported bug, so I kept this PR to the scoped private-use fix. Happy to switch to the `i`-flag approach instead if the maintainers prefer the broader RFC §2.1.1 alignment.

## Testing

Added case-insensitive private-use assertions to `packages/syntax/tests/language.test.ts` for both `isValidLanguage` and `parseLanguageString`. These fail on the unpatched source and pass with the fix.

```
### BEFORE FIX (unpatched source)
 ❯ tests/language.test.ts (2 tests | 2 failed)
   × validates BCP 47
   × parses BCP 47
 AssertionError: expected false to deeply equal true
 AssertionError: expected null to deeply equal { privateUse: 'X-fr-CH' }

### AFTER FIX
 Test Files  9 passed (9)
      Tests  868 passed | 92 skipped (960)
```

`pnpm exec prettier --check` and `pnpm exec eslint` on both changed files report clean. A changeset is included (`@atproto/syntax`, patch).

## Severity

Low. Real-world clients do not emit uppercase `X-` private-use tags today, so wire-format impact is minimal. This is a spec-compliance and reference-implementation-correctness fix.

Fixes #5160

### DIFF
--- a/.changeset/bcp47-uppercase-privateuse.md
+++ b/.changeset/bcp47-uppercase-privateuse.md
@@ -1,0 +1,10 @@
+---
+'@atproto/syntax': patch
+---
+
+`isValidLanguage` and `parseLanguageString` now accept BCP 47 tags whose
+private-use subtag uses an uppercase `X` (for example `X-fr-CH` or `de-X-foo`).
+Per RFC 5646 §2.1.1 (and RFC 5234 §2.3, which makes ABNF quoted-string literals
+case-insensitive), private-use subtags are case-insensitive, so these tags are
+well-formed. Previously only the lowercase `x` form was matched. Lowercase tags
+behave exactly as before.
--- a/packages/syntax/src/language.ts
+++ b/packages/syntax/src/language.ts
@@ -1,5 +1,5 @@
 const BCP47_REGEXP =
-  /^((?<grandfathered>(en-GB-oed|i-ami|i-bnn|i-default|i-enochian|i-hak|i-klingon|i-lux|i-mingo|i-navajo|i-pwn|i-tao|i-tay|i-tsu|sgn-BE-FR|sgn-BE-NL|sgn-CH-DE)|(art-lojban|cel-gaulish|no-bok|no-nyn|zh-guoyu|zh-hakka|zh-min|zh-min-nan|zh-xiang))|((?<language>([A-Za-z]{2,3}(-(?<extlang>[A-Za-z]{3}(-[A-Za-z]{3}){0,2}))?)|[A-Za-z]{4}|[A-Za-z]{5,8})(-(?<script>[A-Za-z]{4}))?(-(?<region>[A-Za-z]{2}|[0-9]{3}))?(-(?<variant>[A-Za-z0-9]{5,8}|[0-9][A-Za-z0-9]{3}))*(-(?<extension>[0-9A-WY-Za-wy-z](-[A-Za-z0-9]{2,8})+))*(-(?<privateUseA>x(-[A-Za-z0-9]{1,8})+))?)|(?<privateUseB>x(-[A-Za-z0-9]{1,8})+))$/
+  /^((?<grandfathered>(en-GB-oed|i-ami|i-bnn|i-default|i-enochian|i-hak|i-klingon|i-lux|i-mingo|i-navajo|i-pwn|i-tao|i-tay|i-tsu|sgn-BE-FR|sgn-BE-NL|sgn-CH-DE)|(art-lojban|cel-gaulish|no-bok|no-nyn|zh-guoyu|zh-hakka|zh-min|zh-min-nan|zh-xiang))|((?<language>([A-Za-z]{2,3}(-(?<extlang>[A-Za-z]{3}(-[A-Za-z]{3}){0,2}))?)|[A-Za-z]{4}|[A-Za-z]{5,8})(-(?<script>[A-Za-z]{4}))?(-(?<region>[A-Za-z]{2}|[0-9]{3}))?(-(?<variant>[A-Za-z0-9]{5,8}|[0-9][A-Za-z0-9]{3}))*(-(?<extension>[0-9A-WY-Za-wy-z](-[A-Za-z0-9]{2,8})+))*(-(?<privateUseA>[xX](-[A-Za-z0-9]{1,8})+))?)|(?<privateUseB>[xX](-[A-Za-z0-9]{1,8})+))$/
 
 export type LanguageTag = {
   grandfathered?: string
--- a/packages/syntax/tests/language.test.ts
+++ b/packages/syntax/tests/language.test.ts
@@ -11,6 +11,9 @@ describe(isValidLanguage, () => {
     expect(isValidLanguage('sl-IT-nedis')).toEqual(true)
     expect(isValidLanguage('mn-Cyrl-MN')).toEqual(true)
     expect(isValidLanguage('x-fr-CH')).toEqual(true)
+    // private-use subtags are case-insensitive (RFC 5646 §2.1.1)
+    expect(isValidLanguage('X-fr-CH')).toEqual(true)
+    expect(isValidLanguage('de-X-foo')).toEqual(true)
     expect(
       isValidLanguage('en-GB-boont-r-extended-sequence-x-private'),
     ).toEqual(true)
@@ -56,6 +59,14 @@ describe(parseLanguageString, () => {
     })
     expect(parseLanguageString('x-fr-CH')).toEqual({
       privateUse: 'x-fr-CH',
+    })
+    // private-use subtags are case-insensitive (RFC 5646 §2.1.1)
+    expect(parseLanguageString('X-fr-CH')).toEqual({
+      privateUse: 'X-fr-CH',
+    })
+    expect(parseLanguageString('de-X-foo')).toEqual({
+      language: 'de',
+      privateUse: 'X-foo',
     })
     expect(
       parseLanguageString('en-GB-boont-r-extended-sequence-x-private'),


matthieusieben:
Thank you 🙏

PR #5181, opened 2026-07-01
pds: improved s3 timeout configs

blackmichael:
## Problem

`com.atproto.repo.uploadBlob` requests can hang until the reverse proxy times out (>90s) when an S3/R2 connection stalls (the known lib-storage hang documented in `s3.ts`, aws/aws-sdk-js-v3#6426).

Root cause: `S3BlobStore` conflates two distinct timeouts. `requestTimeoutMs` (the smithy **socket idle timeout**, i.e. the stall detector) defaults to `uploadTimeoutMs` (the **total upload budget**), and the PDS only exposes the latter (`PDS_BLOBSTORE_S3_UPLOAD_TIMEOUT_MS`). Production sets the upload budget to 300s to accommodate slow clients streaming large blobs — which silently sets the stall detector to 300s too. A stalled connection is therefore only reaped after 5 minutes, long after the proxy gives up at 90s.

The two knobs genuinely have opposite requirements: the upload budget must be long (the body stream is client-paced), while the stall detector can be short — `@aws-sdk/lib-storage` buffers each chunk in memory before dispatching it, so requests to S3 are never client-paced.

## Changes

**`@atproto/aws` (`s3.ts`)**
- `requestTimeoutMs` now defaults to `min(uploadTimeoutMs, 15s)` instead of `uploadTimeoutMs`. Stalled connections are reaped in ~15s and transparently retried by the SDK (`TimeoutError` is classified transient), within the upload budget.
- New `connectionTimeoutMs` option (default 5s) to bound the connect phase.
- `uploadBytes` also translates `TimeoutError` (stall, retries exhausted) into `'Blob upload timed out'`, like `AbortError`.

**`@atproto/pds`**
- New `PDS_BLOBSTORE_S3_REQUEST_TIMEOUT_MS` env var plumbed through to `S3BlobStore`.
- `uploadBlob`'s `throwAbortAsUpstreamError` now walks the error **cause chain** and matches both `AbortError` and `TimeoutError`. This fixes a latent regression: #4169 wrapped the abort error in a plain `Error`, which broke the 504 mapping introduced in #2429 — blob store timeouts have been surfacing as 500s since.

**Tests**
- Adopts vitest in `packages/aws` (previously no test suite) and adds `s3.test.ts`: a fake S3 server with stall/ok handlers per request, verifying (1) stall reaped at `requestTimeoutMs` + retry succeeds, (2) persistent stall → `'Blob upload timed out'` with `TimeoutError` cause well before `uploadTimeoutMs`, (3) total `uploadTimeoutMs` abort still enforced.

## Deployment notes

No config change needed: with `PDS_BLOBSTORE_S3_UPLOAD_TIMEOUT_MS=300000`, the stall detector drops from 300s to 15s on deploy. Worst case with the SDK's default 3 attempts is ~50s — under the 90s proxy limit — and most stalls become invisible retries.


packages/aws/src/s3.test.ts
```diff
@@ -0,0 +1,237 @@
...
+  requests: { method?: string; url?: string }[] = []
+  private sockets = new Set<import('node:net').Socket>()
+
+  constructor() {
+    this.server = http.createServer((req, res) => {
+      this.requests.push({ method: req.method, url: req.url })
+      const handler =
+        this.handlers.length > 1 ? this.handlers.shift() : this.handlers[0]
+      assert(handler, 'no handler configured for incoming request')
+      handler(req, res)
```
matthieusieben: The logic here is a bit hard to follow, and leaks data (requests/handlers) from one test to the next.

```suggestion
      this.handler(req, res)
```

I recommend defining an `handler(req, res) {}` method and override it per test (removing the `requests` and `handlers` properties):

```ts
  handler(
    req: http.IncomingMessage,
    res: http.ServerResponse
  ): void {
    // Default handler that can be overriden in tests
    res.writeHead(501).end()
  }
```


Then, in tests, use `vi.spyOn` to mock the `handler` implementation:

```ts
  // "using" will ensure proper cleanup of the custom handler
  let count = 0
  using handlerMock = vi.spyOn(server, 'hanlder')
    .mockImplementation((req, res) => {
      // Custom logic for this test
      count++

      if (count === 0) return stallHandler(req, res)
      return okHandler(req, res)
    })

  // Access the req/res objects outside the handler
  const [req0, res0] = getActorsSpy.mock.calls[0]
  const [req1, res1] = getActorsSpy.mock.calls[1]
```

You can update the implementation throughout the test:

```ts
  handlerMock.mockImplementation(stallHandler)
    // ...
    handlerMock.mockImplementation(okHandler)
```

matthieusieben (changes requested):
Some small nits in the tests but otherwise looks good! Thanks for tracking this down 🙏

matthieusieben (approved):
Great, thanks !

PR #5185, opened 2026-07-02
Rework `service` and `labellers` XRPC Client options

### LINKED ISSUE #5110: @atproto/lex: no way to remove the default 'service' proxy header

matthieusieben:
## Summary

This PR reworks how the `Client`'s `service` (`atproto-proxy` header) and `labelers` (`atproto-accept-labelers` header) options behave, and simplifies the `Client` / `Agent` relationship:

- `service` and `labelers` set on a `Client` are now _defaults_, merged with per-request options using conventional option-merging semantics: `undefined` (or omitted) inherits the client's default, a defined value replaces it, and `null` explicitly disables the corresponding header for that request.
- `Client` no longer implements the `Agent` interface. A client's configuration applies exclusively to requests initiated through that client; authentication is shared by reusing the underlying session or `client.agent`.
- The record helpers (`create()`, `get()`, `put()`, `delete()`, `list()`, `*Record()`, `applyWrites()`, `uploadBlob()`, `getBlob()`) always target the user's PDS: they default to `service: null` / `labelers: null`, regardless of the client's configuration.

## Motivation

The `Client` is designed to be the single object that SDK-style libraries build upon — app code and prebuilt `Action`s share one instance:

```ts
import { Client } from '@atproto/lex'
import { upsertProfile } from '@bsky/sdk'
import { app } from '@bsky/sdk/lexicons'

const client = new Client(session, {
  // ALL calls are proxied to the AppView by default
  service: 'did:web:api.bsky.app#bsky_appview',
})

// App code: proxied to the AppView (client default)
const { feed } = await client.call(app.bsky.feed.getTimeline)

// SDK action: composes com.atproto.repo.* calls that must reach the
// user's PDS directly
await client.call(upsertProfile, (profile) => {
  profile.displayName = 'matthieu rulez'
})
```

For `upsertProfile` to work, the action must send its `com.atproto.repo.*` requests **without** the `atproto-proxy` header — i.e. suppress the client's default `service` for those specific requests:

```ts
// Before — impossible: the client's default service was always re-applied,
// even when explicitly passing `service: undefined`.

// After — possible: `service: null` opts out of the client's default, so the
// request is sent without an `atproto-proxy` header (straight to the PDS).
const { body } = await client.xrpc(com.atproto.repo.getRecord, {
  service: null,
  params: { repo: client.assertDid, collection, rkey: 'self' },
})
```

The same opt-out applies to `labelers` (`atproto-accept-labelers` header).

### Why `Client` is no longer an `Agent`

Earlier iterations of this PR kept the ability to use a `Client` as the agent of another `Client`, and instead scoped `service`/`labelers` per client. https://github.com/bluesky-social/atproto/issues/5110 and discussion in this PR showed this composition to be confusing and mismatched with real usage (e.g. the browser example's [PdsClientProvider](https://github.com/bluesky-social/atproto/blob/55ef3aff486be060e4fcab6881fd8ef46e2a7cf6/packages/oauth/oauth-client-browser-example/src/providers/PdsClientProvider.tsx#L14-L18) / [BskyClientProvider](https://github.com/bluesky-social/atproto/blob/55ef3aff486be060e4fcab6881fd8ef46e2a7cf6/packages/oauth/oauth-client-browser-example/src/providers/BskyClientProvider.tsx#L78-L83) both build clients directly from the OAuth session): headers were applied at two different levels (options at request time, client headers in `fetchHandler`), which made the interplay between `service: null`, default headers, and composed clients hard to reason about — and impossible to make fully coherent (a per-request `service: null` could be silently undone by a base client's headers).

Removing `Client`'s `fetchHandler` makes the model simple:

- An `Agent` handles _authentication and transport_.
- A `Client` holds _request configuration_ (`service`, `labelers`, custom headers, validation options), applied in exactly one place, at request build time.
- Clients with different configurations share authentication by being built from the same session or from an existing client's `agent`:

```ts
const pdsClient = new Client(session)
const bskyClient = new Client(pdsClient.agent, {
  service: 'did:web:api.bsky.app#bsky_appview',
})
```

## Changes

- **`service` / `labelers` defaults with per-request override and `null` opt-out** (see above). Previously, per-request `labelers` were _appended_ to the client's, and a per-request `service` could not unset the client's.
- **`Client` no longer implements `Agent`** — its `fetchHandler` method is removed. `buildAgent()` now supports default `headers` in `AgentConfig` (and rejects ambiguous configs mixing `fetchHandler` with config-only options).
- **Record helpers always target the PDS** — they default to `service: null` / `labelers: null`, ignoring the client's instance-wide defaults (still overridable through their options).
- **`appLabelers` option** — static app labelers (`Client.configure()`) are still always applied with `;redact`, and can now be overridden per client or per request via the new `appLabelers` option (`null` disables).
- **Header interplay** — an `atproto-proxy` header provided through the `Client` constructor's `headers` option acts as a fallback for the `service` option (the two are equivalent at the client level); per-request `atproto-proxy` headers are overridden by the (always-defined) `service` option. `atproto-accept-labelers` headers are merged with the `labelers`/`appLabelers` options, unless `labelers: null` disables the header. For the bare `xrpc()` helper (no `Client`), headers pass through untouched unless the corresponding option is provided.

Supersedes #5142
Fixes #5110


packages/lex/lex-client/src/util.ts
```diff
@@ -94,19 +118,25 @@ export type XrpcRequestHeadersOptions = {
 export function buildXrpcRequestHeaders(
   options: XrpcRequestHeadersOptions,
 ): Headers {
-  const headers = new Headers(options?.headers)
+  const headers = new Headers(options.headers)
+
+  // Custom headers starting with "atproto-" are reserved and cannot be set
+  // directly.
+  // @NOTE Copying the keys before deleting, as deleting entries while
+  // iterating causes entries to be skipped.
```
devinivy: In my opinion this is a surprising behavior.  My inclination would be for headers that are manually set to override the library's options.

packages/lex/lex-client/tests/client.test.ts
```diff
@@ -59,6 +59,270 @@ describe('utils', () => {
...
+      const fetchHandler = vi.fn<FetchHandler>(async (_url, _init) => {
+        return Response.json({ preferences: [] })
+      })
+
+      const client = new Client(fetchHandler, {
+        headers: {
+          'atproto-proxy': 'did:plc:existing#service',
+          'atproto-accept-labelers': 'did:plc:existing',
+        },
+      })
```
devinivy: I see this usage as an escape hatch that is quite useful to have.  Setting the header is "more specific" than the library's options and would override them.  E.g. if the library is missing some future functionality with these headers, or if you're talking to a service with its own usage quirk, you can drop down to set the headers manually rather than use the library.

packages/lex/lex-client/src/util.ts
```diff
@@ -94,19 +118,25 @@ export type XrpcRequestHeadersOptions = {
...
-    headers.set(
-      'atproto-accept-labelers',
-      [...options.labelers, headers.get('atproto-accept-labelers')?.trim()]
-        .filter(Boolean)
-        .join(', '),
-    )
+  const labelers = new Set(options.labelers)
+  if (labelers.size > 0) {
+    headers.set('atproto-accept-labelers', Array.from(labelers).join(', '))
   }
```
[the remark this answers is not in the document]
devinivy: I wonder why this type checked, huh!

packages/lex/lex-client/tests/client.test.ts
matthieusieben: The issue with these headers is that they directly conflict with the `service` and `labelers` properties:

```ts
const client = new Client(fetchHandler, {
  headers: {
    'atproto-proxy': 'did:plc:existing#service', // conflicts with "service"
    'atproto-accept-labelers': 'did:plc:a', // conflicts with "labelers"
  },
  service: null,
  labelers: ['did:plc:b']
})
```

The current implementation strips every `atproto-*` prefixed header. I'm totally fine with allowing headers others than `atproto-proxy` and `atproto-accept-labelers` go through though.


Note that an escape hatch **is** available:

```ts
const client = new Client(async (path, init) => {
  const headers = new Headers(init?.headers)

  // Override headers, including "atproto-proxy" and "atproto-accept-labelers"

  headers.set('atproto-foo', 'bar')
  headers.set('atproto-proxy', 'did:plc:existing#service')
  headers.delete('atproto-accept-labelers')
  
  return fetchHandler(path, {...init, headers})
}, {
  service: null,
  labelers: ['did:plc:b']
})
```

packages/lex/lex-client/src/util.ts
matthieusieben: We can't do that. This is the core of the issue this PR is trying to adress:

We need to be able to use `{ service: null }` (or even `headers: { "atproto-proxy": undefined }`) to **force** the disabling of service proxying on a per request basis.

Since options are "compiled" into an `Headers` instance, that instance will **not** contain the signal "this request wants to **not** use service proxying".

Since that signal is missing, if we have a behavior where the global headers "override the library's option", then there is no way for the caller to disable the use of service proxy.

packages/lex/lex-client/src/util.ts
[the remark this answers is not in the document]
matthieusieben: Copilot is wrong (tested in NodeJS 24, GC and FF):

```js
> new Set(null) // Set(0) {}
```

packages/lex/lex-client/src/util.ts
devinivy: My understanding is that it needs to be disabled at certain call sites, and that would still work.  So you still get the main override behavior we're solving for here, which is a call site overriding the client default.

E.g. this would _not_ end up sending the atproto-accept-labelers header because the call site overrides the client options (whether they are set via header or config):
```ts
const client = new Client(fetchHandler, {
  headers: {
    'atproto-accept-labelers': 'did:plc:a;custom-server-flag',
  },
})

const { body } = await client.xrpc(com.atproto.repo.getRecord, {
  ...options,
  labelers: null,
  params: { repo: client.assertDid, collection, rkey: 'self' },
})
```

This seems like a coherent usage to me, and it's what I would expect to be able to do.

I would also be surprised if I called this and the headers were silently not being sent:
```ts
const { body } = await client.xrpc(com.atproto.repo.getRecord, {
  ...options,
  params: { repo: client.assertDid, collection, rkey: 'self' },
  headers: {
    'atproto-proxy': 'did:plc:a#svc',
  }
})
```

matthieusieben: The problem is that these things act at different levels internally:

By specifying `options.labelers: null` at a particular call site, we are easily able to not use the `this.labelers` (from the constructor).

This results internally into an `Headers` instance to be constructed, which does not contain the `atproto-accept-labelers` header.

However, `this.headers` (which comes from the constructor), is applied at a later stage; in the `fetchHandler(){}` implementation.

In there, we add every header that is defined in `this.header` but not in the current request headers.

And so the header gets added back.

Also what should happen when the user does:

```ts
const { body } = await client.xrpc(com.atproto.repo.getRecord, {
  ...options,
  params: { repo: client.assertDid, collection, rkey: 'self' },
  labelers: null,
  headers: {
    'atproto-accept-labelers': 'did:plc:a;custom-server-flag',
  }
})
```

matthieusieben:
Hey @devinivy I've completely reworked this PR. I believe that the confusion in the existing code mainly comes from the fact that `Client` implements `Agent`, and uses itself as agent when performing the inner `xrpc()` call. Because of this, the way headers are built and combined could be quite unintuitive.

> client -> xrpc() -> client (fetchHandler) -> agent (session)

The new approach I took here is to remove that loop in the call graph by removing the `fetchHandler` method from the client. In practice, it is not that useful and the use of a client as agent to another client made that call graph even more hard to reason about:

> client -> xrpc() -> client (fetchHandler) -> inner client (fetchHandler) -> agent (session)

The new logic is much more straight forward

> client -> xrpc() -> agent

It also allows to be way more explicit about the way options are defined and propagated:


```ts
const client = new Client('http://a', {
  headers: {
    'atproto-proxy': "a" // has no effect if "service" is also specified
  },
  service: "b" // defaults to headers.get('atproto-proxy')
})

await client.xrpc(m, {
  headers: {
    'atproto-proxy': "c" // has no effect if "service" is also specified
  },
  service: "d" // defaults to headers.get('atproto-proxy')
})

// atproto-proxy will be the first defined value of "d", "c", "b", "a"
```

@claude Please thourghougly review this PR as a whole ignoring previous comments.

devinivy (commented):
I really like where you took this!

Findings 2 and 3 from claude's review seem like they are worth a look before merge.  I thought you were clear about the behavior of `appLabelers` in finding 1, and I don't see that as an issue.

PR #5187, opened 2026-07-02
Atproto spaces

dholms:
⚠️ WIP ⚠️


packages/lex/lex-installer/src/lex-installer.ts
```diff
@@ -424,6 +424,11 @@ function* defRefs(
         yield* defRefs(permission)
       }
       return
+    case 'space':
+      for (const collection of def.collections) {
+        yield collection
+      }
```
matthieusieben, line 430:
```suggestion
      yield* def.collections
```

lexicons/com/atproto/space/defs.json
```diff
@@ -0,0 +1,37 @@
+{
+  "lexicon": 1,
+  "id": "com.atproto.space.defs",
+  "defs": {
+    "signedCommit": {
```
matthieusieben, line 5: Why not make this it's own lexicon document:

`$type: "com.atproto.space.signedCommit"`

instead of:

`$type: "com.atproto.space.defs#signedCommit"`

packages/oauth/oauth-provider/src/identity/identity-manager.ts
```diff
@@ -0,0 +1,80 @@
+import { IdResolver, getHandle } from '@atproto/identity'
```
matthieusieben: Note that I've been trying to avoid `@atproto/identity` because it does not support "safeFetch" and is suceptible to SSRF attacks.

matthieusieben: See `@atproto-labs/handle-resolver-node` and `@atproto-labs/identity-resolver` for alternatives that support custom (safe) fetch implementation

matthieusieben (commented):
To me, it feels quite unintuitive that a space has to be created with `com.atproto.`**`simplespace.createSpace`** or `app.example.`**`complexspace.newSpace`**, and then retrieved with `com.atproto.`**`space.getSpace`**.

It feels even more unintuitive, and unfortunate, that a user can make their community go from "member based" to "public" to "private", but can **never** in the future go from `simplespace` to `complexspace`.

I am strongly against this dual way of configuring space access (the "polity type", and then the "policy type"'s config). Even naming them is awkward: is "simplespace" a "config", a "policy", a "policy type" ?. I think we should have only one way of configuring access:

- Either have multiple policy types that have a single behavior each ("simplespace", "publicspace", "privatespace", "managedspace", and more custom 3rd party policies in the future)
- Or a single policy type that has a configurable behavior (ie  merge all the "simplespace" stuff under "com.atproto.space")

If we do the former (which I believe is the right call), spaces would **all** be created with `com.atproto.space.createSpace`. That endpoint would have a required `policy` property, that can be any string token (e.g. `com.atproto.space.defs#publicspace`).

Then, based on that `policy` value, more endpoints become available to configure the space:

- `simplespace` enables the member list management endpoints (`addMember`, etc.)
- `managedspace` has endpoints to configure the managing app (like `getConfig`, `setManagingApp`, etc.)
- `publicspace` and `privatespace` might not really need any config

We could codify this under a new lexicon type but I don't think we need to go that way. A simple convention around "policy X enables endpoints Y" (which can be setup as each TOKEN's description) is enough IMO.

lexicons/com/atproto/simplespace/createSpace.json
```diff
@@ -0,0 +1,74 @@
...
+  "id": "com.atproto.simplespace.createSpace",
+  "defs": {
+    "main": {
+      "type": "procedure",
+      "description": "Create a new space managed by the simplespace implementation. The space is anchored on the authenticated user's DID, who becomes the space owner. Requires auth, implemented by PDS.",
+      "input": {
+        "encoding": "application/json",
+        "schema": {
+          "type": "object",
+          "required": ["type", "policy", "appAccess"],
```
wdantuma, line 12: I think the properties policy and appAccess should be optional ? the [proposal](https://github.com/bluesky-social/proposals/tree/main/0016-permissioned-data#configuration) allready defines defaults for them

bnewbold:
Noticed that this branch includes a new `space-ref` lexicon string format type, which isn't mentioned in the proposal doc.

PR #5189, opened 2026-07-04
fix(syntax): reject uppercase and bare 4-letter primary language subtags

### LINKED ISSUE #5184: Fix language string format validation
https://github.com/bluesky-social/atproto/pull/5183 adds tests for the `language` format against the https://github.com/bluesky-social/atproto-interop-tests/ suite.

While doing that, we realized that `JA` and `jaja`, although invalid, are allowed by the TypeScript implementation.

We should fix this.

ryanda9910:
## Problem

`@atproto/syntax`'s BCP 47 language validation accepts two tags that the [atproto interop test suite](https://github.com/bluesky-social/atproto-interop-tests/blob/main/syntax/language_syntax_invalid.txt) marks as invalid:

- `JA` — an uppercase primary language subtag
- `jaja` — a bare four-letter run

Both `isValidLanguage('JA')` and `isValidLanguage('jaja')` currently return `true`.

## Root cause

The `BCP47_REGEXP` `language` alternative is:

```
(?<language>([A-Za-z]{2,3}(-(?<extlang>[A-Za-z]{3}(-[A-Za-z]{3}){0,2}))?)|[A-Za-z]{4}|[A-Za-z]{5,8})
```

Two things make it too permissive relative to the interop suite:

1. `[A-Za-z]{2,3}` allows an **uppercase** primary language subtag, so `JA` matches. Per [RFC 5646 §2.1.1](https://www.rfc-editor.org/rfc/rfc5646.html#section-2.1.1) the primary language subtag is a lowercase two/three-letter code.
2. The `[A-Za-z]{4}` (reserved) and `[A-Za-z]{5,8}` (registered) bare alternatives let a plain four-letter string like `jaja` validate as a language.

## Fix

Restrict the `language` alternative to a lowercase two/three-letter code with a lowercase `extlang`, and drop the bare 4-letter / 5-8-letter alternatives:

```
(?<language>[a-z]{2,3}(-(?<extlang>[a-z]{3}(-[a-z]{3}){0,2}))?)
```

Script (`[A-Za-z]{4}`) and region (`[A-Za-z]{2}|[0-9]{3}`) casing is left untouched, so titlecase scripts (`Latn`) and uppercase regions (`BR`, `CH`) still validate. The two now-rejected cases are enabled in the vendored interop file, which previously kept them commented out with the note *"part of the invalid interop test, but current tests allow them."*

## Test evidence

Before (against `main`):

```
"JA"   -> true
"jaja" -> true
```

After — full `@atproto/syntax` suite is green, including the newly enabled interop cases:

```
✓ tests/language.test.ts (32 tests)
...
 Test Files  9 passed (9)
      Tests  1088 passed | 92 skipped (1180)
```

Explicit unit assertions for `JA` and `jaja` were added to `language.test.ts` for both `isValidLanguage` and `parseLanguageString`. Verified no regression against every case in `language_syntax_valid.txt` (`en-US-boont`, `zh-Hant`, `qaa-Qaaa-QM-x-southern`, `X-fr-CH`, grandfathered tags, etc.) and against downstream `@atproto/lex-schema` (1819 tests pass; the 2 pre-existing `RFC 5646 §4.1` duplicate-subtag failures there exist on `main` unchanged and are unrelated to this change).

Fixes #5184

### DIFF
--- a/.changeset/lang-lowercase-primary.md
+++ b/.changeset/lang-lowercase-primary.md
@@ -1,0 +1,9 @@
+---
+'@atproto/syntax': patch
+---
+
+`isValidLanguage` and `parseLanguageString` now reject language tags whose
+primary language subtag is uppercase (for example `JA`) or is a bare four-letter
+run (for example `jaja`). The primary language subtag must be a lowercase
+two-or-three-letter code per RFC 5646 §2.1.1, matching the atproto interop test
+suite. Well-formed lowercase tags are unaffected.
--- a/.changeset/lang-lowercase-primary.md
+++ b/.changeset/lang-lowercase-primary.md
@@ -2,8 +2,10 @@
 '@atproto/syntax': patch
 ---
 
-`isValidLanguage` and `parseLanguageString` now reject language tags whose
-primary language subtag is uppercase (for example `JA`) or is a bare four-letter
-run (for example `jaja`). The primary language subtag must be a lowercase
-two-or-three-letter code per RFC 5646 §2.1.1, matching the atproto interop test
-suite. Well-formed lowercase tags are unaffected.
+`parseLanguageString` (strict validation) now rejects language tags whose
+primary language subtag is uppercase (for example `JA`) or is a bare
+four-letter run (for example `jaja`). The primary language subtag must be a
+lowercase two-or-three-letter code per RFC 5646 §2.1.1, matching the atproto
+interop test suite. `isValidLanguage` is unchanged and stays permissive of
+these legacy forms (well-formed syntax only), so existing records are not
+retroactively invalidated. Well-formed lowercase tags are unaffected.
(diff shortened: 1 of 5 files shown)


matthieusieben:
Also fixed here: https://github.com/bluesky-social/atproto/pull/5191

packages/syntax/src/language.ts
```diff
@@ -1,5 +1,5 @@
 const BCP47_REGEXP =
-  /^((?<grandfathered>(en-GB-oed|i-ami|i-bnn|i-default|i-enochian|i-hak|i-klingon|i-lux|i-mingo|i-navajo|i-pwn|i-tao|i-tay|i-tsu|sgn-BE-FR|sgn-BE-NL|sgn-CH-DE)|(art-lojban|cel-gaulish|no-bok|no-nyn|zh-guoyu|zh-hakka|zh-min|zh-min-nan|zh-xiang))|((?<language>([A-Za-z]{2,3}(-(?<extlang>[A-Za-z]{3}(-[A-Za-z]{3}){0,2}))?)|[A-Za-z]{4}|[A-Za-z]{5,8})(-(?<script>[A-Za-z]{4}))?(-(?<region>[A-Za-z]{2}|[0-9]{3}))?(-(?<variant>[A-Za-z0-9]{5,8}|[0-9][A-Za-z0-9]{3}))*(-(?<extension>[0-9A-WY-Za-wy-z](-[A-Za-z0-9]{2,8})+))*(-(?<privateUseA>[xX](-[A-Za-z0-9]{1,8})+))?)|(?<privateUseB>[xX](-[A-Za-z0-9]{1,8})+))$/
+  /^((?<grandfathered>(en-GB-oed|i-ami|i-bnn|i-default|i-enochian|i-hak|i-klingon|i-lux|i-mingo|i-navajo|i-pwn|i-tao|i-tay|i-tsu|sgn-BE-FR|sgn-BE-NL|sgn-CH-DE)|(art-lojban|cel-gaulish|no-bok|no-nyn|zh-guoyu|zh-hakka|zh-min|zh-min-nan|zh-xiang))|((?<language>[a-z]{2,3}(-(?<extlang>[a-z]{3}(-[a-z]{3}){0,2}))?)(-(?<script>[A-Za-z]{4}))?(-(?<region>[A-Za-z]{2}|[0-9]{3}))?(-(?<variant>[A-Za-z0-9]{5,8}|[0-9][A-Za-z0-9]{3}))*(-(?<extension>[0-9A-WY-Za-wy-z](-[A-Za-z0-9]{2,8})+))*(-(?<privateUseA>[xX](-[A-Za-z0-9]{1,8})+))?)|(?<privateUseB>[xX](-[A-Za-z0-9]{1,8})+))$/
```
matthieusieben: The tests do not seem to explicitly cover the change in `extlang`. Why was that change required ? https://github.com/bluesky-social/atproto/pull/5191 keeps this part as is.

.changeset/lang-lowercase-primary.md
```diff
@@ -0,0 +1,9 @@
+---
```
matthieusieben, line 1: This PR actually introduces a pretty "bad" breaking change (requires minor bump in `@atproto/lex`, `@atproto/lex-schema`, `@atproto/syntax`). A lot of records might become invalid because of this change.

My recommendation is to only reject these values from `parseLanguageString` which is used in "strict" validation, while `isValidLanguage` should explicitly keep allowing these legacy values.

packages/syntax/src/language.ts
ryanda9910: You're right, the extlang change wasn't needed. I reverted the whole regex to the upstream version, so extlang is untouched now, matching #5191. The primary-subtag rule is enforced outside the regex instead, as a small post-match guard in parseLanguageString.

.changeset/lang-lowercase-primary.md
ryanda9910: Good call, thanks. I've reworked this to avoid the breaking change. BCP47_REGEXP is back to the upstream original, so isValidLanguage stays permissive of these legacy forms (JA, jaja) and no existing records are invalidated. The strict rejection now lives only in parseLanguageString as a small post-match guard on the primary subtag, so it's a @atproto/syntax patch rather than a minor bump across lex/lex-schema. I also moved the JA/jaja interop cases from language_syntax_invalid.txt into language_parse_invalid.txt to match the lenient/strict split.

ryanda9910:
Noted, thanks for the pointer. #5191 and this now take the same non-breaking direction on the syntax side (regex untouched, strict rejection scoped to the strict-parse path). Happy to defer to #5191 if you prefer it, or keep this one focused on the parseLanguageString behavior. Whichever is easier to land.

matthieusieben:
We'll keep this for now. Following up in https://github.com/bluesky-social/atproto-interop-tests/issues/18

matthieusieben:
Thanks for your contribution

PR #5191, opened 2026-07-04
Add lex-schema language strictness tests

arjxnt:
## Summary

Adds focused `@atproto/lex-schema` regression coverage for the language strictness boundary introduced by #5189.

- strict language validation rejects legacy primary subtags such as `JA` and `jaja`
- loose validation continues to accept those legacy forms for compatibility
- removes the broader runtime validation changes from this PR, because #5189 already fixed `parseLanguageString` while keeping `isValidLanguage` lenient

Follow-up to #5184 and #5189.

## Validation

- `pnpm --filter @atproto/syntax build`
- `pnpm --dir packages/lex/lex-schema exec vitest run src/schema/string.test.ts`
- `pnpm exec prettier --check packages/lex/lex-schema/src/schema/string.test.ts`
- `git diff --check`

### DIFF
--- a/.changeset/swift-languages-validate.md
+++ b/.changeset/swift-languages-validate.md
@@ -1,0 +1,8 @@
+---
+"@atproto/syntax": patch
+"@atproto/lex": patch
+"@atproto/lex-schema": patch
+"@atproto/lexicon": patch
+---
+
+Tighten language string validation to reject uppercase and four-letter primary language subtags.
--- a/packages/lex/lex/tests/string.test.ts
+++ b/packages/lex/lex/tests/string.test.ts
@@ -455,5 +455,17 @@ describe('com.example.language', () => {
         language: 'not-a-language-',
       }),
     ).toThrow('Invalid language (got "not-a-language-") at $.language')
+    expect(() =>
+      com.example.language.$parse({
+        $type: 'com.example.language',
+        language: 'JA',
+      }),
+    ).toThrow('Invalid language (got "JA") at $.language')
+    expect(() =>
+      com.example.language.$parse({
+        $type: 'com.example.language',
+        language: 'jaja',
+      }),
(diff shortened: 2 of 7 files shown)


matthieusieben:
Also fixed in https://github.com/bluesky-social/atproto/pull/5189

Could you please explain how this approach compares to that other approach ?

packages/lex/lex/tests/string.test.ts
```diff
@@ -455,5 +455,17 @@ describe('com.example.language', () => {
...
+        $type: 'com.example.language',
+        language: 'JA',
+      }),
+    ).toThrow('Invalid language (got "JA") at $.language')
+    expect(() =>
+      com.example.language.$parse({
+        $type: 'com.example.language',
+        language: 'jaja',
+      }),
+    ).toThrow('Invalid language (got "jaja") at $.language')
```
matthieusieben, line 469: This is a pretty "bad" regression. A lot of records might become invalid because of this change, and we cannot afford this.

My recommendation is to only reject these values from `parseLanguageString` which is used in "strict" validation, while `isValidLanguage` should explicitly keep allowing these legacy values.

This test should be updated to actually allow legacy values. Another test in `lex-schema` should address the strictness difference.

.changeset/swift-languages-validate.md
```diff
@@ -0,0 +1,8 @@
+---
+"@atproto/syntax": patch
+"@atproto/lex": patch
+"@atproto/lex-schema": patch
+"@atproto/lexicon": patch
```
matthieusieben, line 5: This is a breaking change that requires a `minor` bump.

`@atproto/lexicon` should not be affected by this (see my other comment about strictness).

matthieusieben:
Fixed in https://github.com/bluesky-social/atproto/pull/5189
Follow up in https://github.com/bluesky-social/atproto-interop-tests/issues/18 regarding strictness

PR #5192, opened 2026-07-04
Add permission-set description option

### LINKED ISSUE #5168: @atproto/lexicon: lexPermissionSet schema accepts an undocumented "description" field
**Describe the bug**

The `lexPermissionSet` Zod schema in `packages/lexicon/src/types.ts` declares an optional `description` field that is not part of the Lexicon specification for `permission-set`, is not used by any lexicon JSON in this repository, and is not modeled by the newer `@atproto/lex-schema` SDK. As a result, an implementer reading `@atproto/lexicon` as a reference may incorrectly conclude that `description` is a valid top-level field on a permission-set definition.

**To Reproduce**

1. View the `lexPermissionSet` Zod schema in the legacy `@atproto/lexicon` package: https://github.com/bluesky-social/atproto/blob/79f29409c0dd7d1ab4029095bb3c5ff69183f85d/packages/lexicon/src/types.ts#L225-L233 — observe `description: z.string().optional()` on line 227.
2. Compare with the spec at https://atproto.com/specs/lexicon#permission-set — only `title`, `title:lang`, `detail`, `detail:lang`, and `permissions` are listed as permission-set fields.
3. Compare with `PermissionSetOptions` in the new `@atproto/lex-schema` SDK: https://github.com/bluesky-social/atproto/blob/79f29409c0dd7d1ab4029095bb3c5ff69183f85d/packages/lex/lex-schema/src/schema/permission-set.ts#L12-L17 — only `title`, `title:lang`, `detail`, `detail:lang` are defined.
4. Run `grep -l '"description"' lexicons/app/bsky/auth*.json lexicons/chat/bsky/authFullChatClient.json` — no top-level `description` is used on any `"type": "permission-set"` definition.

**Expected behavior**

`lexPermissionSet` should match the spec and the new `@atproto/lex-schema` model. Either the `description` field should be removed from the schema, or the spec and the new SDK should be updated to include it. The three sources of truth (spec, legacy `@atproto/lexicon`, new `@atproto/lex-schema`) should agree.

**Details**

- Operating system: not applicable (static schema discrepancy)
- Node version: not applicable

**Additional context**

The `description` field has been present on `lexPermissionSet` since the introduction of `permission-set` in commit https://github.com/bluesky-social/atproto/commit/f9dc9aa4c9eaf2f82d140fbf011a9015e7f1a00d ("Permission set (#4108)", 2025-08-29). All 10 lexicon JSON files that currently use `"type": "permission-set"` are under `lexicons/app/bsky/` (9 files) and `lexicons/chat/bsky/` (1 file), and none use a top-level `description`. The newer `@atproto/lex-schema` SDK omits the field, which suggests it has already been treated as vestigial in the modern stack. Since `@atproto/lexicon` is being phased out in favor of `@atproto/lex`, removing the field from the legacy schema is a low-risk one-line change but would prevent confusion for downstream SDK authors reading this codebase as a reference.

arjxnt:
## Summary
Adds `description` to `@atproto/lex-schema` permission set options so it matches the legacy `@atproto/lexicon` permission-set schema and maintainer feedback.

- adds `description?: string` to `PermissionSetOptions`
- documents it as developer-facing permission set documentation
- adds regression coverage for all-options and description-only cases
- adds a patch changeset for `@atproto/lex-schema`

## Root cause
The legacy lexicon schema already models `description`, but the newer lex-schema permission-set options did not, so the two permission-set models disagreed.

Fixes #5168.

## Validation
- `pnpm exec prettier --check .changeset/quick-permissions-shave.md packages/lex/lex-schema/src/schema/permission-set.ts packages/lex/lex-schema/src/schema/permission-set.test.ts`
- `pnpm --dir packages/lex/lex-schema exec vitest run src/schema/permission-set.test.ts`
- `pnpm --filter @atproto/lex-schema build`
- `pnpm changeset status --since=origin/main`
- `git diff --check`

### DIFF
--- a/.changeset/quick-permissions-shave.md
+++ b/.changeset/quick-permissions-shave.md
@@ -1,0 +1,5 @@
+---
+"@atproto/lex-schema": patch
+---
+
+Add `description` to permission-set options for developer-facing documentation.
--- a/packages/lex/lex-schema/src/schema/permission-set.test.ts
+++ b/packages/lex/lex-schema/src/schema/permission-set.test.ts
@@ -11,6 +11,7 @@ describe('PermissionSet', () => {
         permission('app.bsky.feed.post:write', {}),
       ] as const
       const options = {
+        description: 'Developer-facing permission set documentation',
         title: 'Post Management',
         detail: 'Allows reading and writing posts',
       }
@@ -21,6 +22,9 @@ describe('PermissionSet', () => {
       expect(perms.nsid).toBe(nsid)
       expect(perms.permissions).toBe(permissions)
       expect(perms.options).toBe(options)
+      expect(perms.options.description).toBe(
+        'Developer-facing permission set documentation',
+      )
     })
 
     it('creates a PermissionSet instance with minimal options', () => {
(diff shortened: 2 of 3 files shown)


matthieusieben (changes requested):
I think that the better fix here would be to add a `description` field in both places. `description` are developer focused "documentation" for the lexicon document, and it can be useful in case of permission set.

matthieusieben (approved):

PR #5193, opened 2026-07-05
fix(handle-resolver): surface swallowed resolution errors via an onError hook

Rex0Lux:
## Summary

`WellKnownHandleResolver.resolve()` catches every non-abort failure and returns `null`, which is indistinguishable from "this handle has no `.well-known/atproto-did` endpoint." As reported in #4215, that means a `safeFetch` unicast/SSRF block (and any 5xx, redirect, or network error) disappears silently, leaving no way to understand *why* resolution failed.

This adds an opt-in `onError` observability hook to `ResolveHandleOptions`. The resolver's return contract is unchanged - it still returns `null` on failure - but callers can now observe the underlying cause. This implements the `onError`-callback direction discussed on #5102 rather than re-throwing, so existing call sites keep working without changes.

## Changes

- Add optional `onError?: (err, context) => void` to `ResolveHandleOptions`, plus an exported `HandleResolverErrorContext` type. The public return contract is unchanged.
- Route non-2xx responses in `WellKnownHandleResolver` through the same error path (`fetch()` does not throw on 4xx/5xx), so operational failures - not just transport-level errors - reach `onError`.
- Scope the context `resolver` field to `'well-known'`: it is the only strategy that swallows errors, so it is the only one whose failures would otherwise be invisible. The DNS and XRPC resolvers already throw on unexpected errors, so their failures surface to callers already.
- Guard the `onError` invocation in its own `try/catch`, so a throwing callback cannot change the resolver's return value (keeping the hook purely observational).
- Add vitest coverage for the success / non-DID / non-2xx / network-error / abort paths, the callback contract (fires once, with the well-known context, forwarding the underlying cause), the throwing-callback guarantee, and DNS-vs-well-known preference.

## Compatibility

- Additive and backward compatible: `onError` is optional and no existing caller passes it; adding an optional field cannot break interface implementers or option constructors.
- The only behavioral delta is that a non-2xx response whose body happens to be a valid DID now returns `null` instead of that DID - unreachable for spec-conformant endpoints, which serve the DID with `200`.
- Verified locally: `@atproto-labs/handle-resolver` builds, `vitest run` passes, and `pnpm install --frozen-lockfile` is consistent.

Changeset: `patch` bump for `@atproto-labs/handle-resolver`.

Addresses #4215.


packages/internal/handle-resolver/src/internal-resolvers/well-known-handler-resolver.ts
```diff
@@ -36,17 +37,42 @@ export class WellKnownHandleResolver implements HandleResolver {
...
+      // the request failed for a reason that prevents resolution (network
+      // error, SSRF block, 5xx, redirect, etc.) - we still return null per
+      // the resolver contract, but expose the cause via onError so callers
+      // can distinguish "no .well-known endpoint exists" from operational
+      // failures that warrant logging or end-user feedback.
       options?.signal?.throwIfAborted()
 
+      // The onError hook is purely observational: a throwing callback must not
+      // change the resolver's return contract, so isolate it in its own catch.
+      if (options?.onError) {
```
matthieusieben: I'm wondering if this would not be better implemented as an instance config (instead of a per-call option)

matthieusieben (commented):
Looking good. I'd like your opinion on making this an instance config instead of a per-call option, since I believe it'd make more sense.

packages/internal/handle-resolver/src/internal-resolvers/well-known-handler-resolver.ts
Rex0Lux: Good call, switched it to instance config.

`onError` is now a constructor option on `WellKnownHandleResolver`, and it flows through `AtprotoHandleResolver` and `AtprotoHandleResolverNode`. `ResolveHandleOptions` goes back to just `signal`/`noCache`.

This also fixes something the per-call version didn't: it actually reaches the case in #4215. A `NodeOAuthClient` user never calls `.resolve()` themselves, so there was no per-call seam to pass a handler through. With it on the instance they can do:

```ts
new NodeOAuthClient({
  handleResolver: new AtprotoHandleResolverNode({ onError: logHandleError }),
  // ...
})
```

and see the SSRF/5xx cause that was being swallowed.

Behavior is unchanged when `onError` isn't set (still returns null), a throwing handler is isolated so it can't affect the result, and non-2xx responses go through the same path. Happy to also add a passthrough on `NodeOAuthClientOptions` if you'd rather it be first-class on the client.

matthieusieben (approved):
Looks good thanks

matthieusieben:
Looks like we need to merge main 🙏

Rex0Lux:
> Looks like we need to merge main 🙏

No worries, glad it worked out.

packages/internal/handle-resolver/tests/well-known-handler-resolver.test.ts
```diff
@@ -0,0 +1,189 @@
+import { describe, expect, it, vi } from 'vitest'
+import { AtprotoHandleResolver, isResolvedHandle } from '../src/index.js'
+import { WellKnownHandleResolver } from '../src/internal-resolvers/well-known-handler-resolver.js'
+
+const DID = 'did:web:example.com'
+
+type MockResponse = { ok: boolean; status: number; body: string }
+
+function mockResponse({ ok, status, body }: MockResponse) {
```
matthieusieben: Is there any reason for not using an actual `new Response` instance?

matthieusieben (approved):
Looks good. Not sure I see the point of using a `mockResponse` rather than an actual `Response` instance, but I won't block on it.

packages/internal/handle-resolver/tests/well-known-handler-resolver.test.ts
Rex0Lux: No real reason, just habit. Switched them to actual Response instances (constructed per call, since a body can only be read once), which also drops the manual cast and derives `ok` from the status.

matthieusieben (approved):

PR #5200, opened 2026-07-07
Appview traces

estrattonbailey:
Replaces the Datadog-specific dd-trace tracer in the bsky image with the OpenTelemetry NodeSDK, following the same pattern as the PDS (#4805). See the docs file for all info.

Questions:
1. We match `peerService` names with those defined in `tango`. `atlantis` is clear, as is `rolodex`, but what do we call `bsync` (maybe `vortex`?) and `courier`?
2. A few span names like `rpc.system.name` and `rpc.response.status_code` are still RC stage, even though the other options are [marked as deprecated](https://github.com/open-telemetry/semantic-conventions/blob/v1.41.1/docs/registry/attributes/rpc.md#rpc-system). Not sure how much we care, seems a bit of a mess.
3. ⚠️ Deployment blocker: the appview compose template in bsky-infra (`pop/roles/appview/templates/compose.yaml.j2`) overrides the image CMD with its own `command:` line, which hardcodes `--import=./tracer.ts`. This PR deletes that file from the image, so rolling out an image built from this branch breaks the container at boot until the compose command is updated to `--import=@atproto/bsky/telemetry`. The `social.bsky.appview.telemetry=otel` image label exists to gate this — the compose change needs to land in lockstep with the image rollout. (The PDS rollout for #4805 presumably has the same coordination problem.)


packages/bsky/src/courier.ts
```diff
@@ -9,14 +9,21 @@ import {
...
 
 export const createCourierClient = (
   opts: ConnectTransportOptions,
 ): CourierClient => {
-  const transport = createConnectTransport(opts)
+  const transport = createConnectTransport({
+    ...opts,
+    interceptors: [
+      ...(opts.interceptors ?? []),
+      tracingInterceptor({ peerService: 'courier' }),
```
estrattonbailey, line 24: Evidently we don't actually set `courier` anywhere in the `tango` repo 🤔

services/bsky/package.json
```diff
@@ -6,6 +6,6 @@
   "dependencies": {
     "@atproto/bsky": "workspace:^",
     "@atproto/crypto": "workspace:^",
-    "dd-trace": "^5.103.0"
+    "@opentelemetry/instrumentation": "^0.219.0"
```
matthieusieben: You don't need this here ;-)

packages/bsky/src/telemetry.ts
```diff
@@ -0,0 +1,198 @@
...
+// startNodeSDK and new NodeSDK. We determine which one to use based on the
+// presence of the OTEL_CONFIG_FILE environment variable. If it is set, we use
+// startNodeSDK, which will load the configuration from a YAML file.
+// Otherwise, we use new NodeSDK, which will load the configuration from
+// environment variables (and supports creating an HTTP prometheus exporter).
+// @NOTE Inverted from the OTEL_SDK_DISABLED spec semantics (matching the pds
+// telemetry script): the SDK stays off unless explicitly opted in with
+// OTEL_SDK_DISABLED=false.
+const enabled = process.env.OTEL_SDK_DISABLED?.toLowerCase() === 'false'
+if (enabled) {
```
matthieusieben: I've updated the detection logic to make it more convenient to configure (ie. not have to specify `OTEL_SDK_DISABLED=false` to opt-in otel)


https://github.com/bluesky-social/atproto/blob/47ca5e341b277246a69f7331970dbbec6720f622/packages/pds/src/telemetry.ts#L54-L61

packages/bsky/src/rpc-tracing.ts
```diff
@@ -0,0 +1,145 @@
+import { Code, ConnectError, type Interceptor } from '@connectrpc/connect'
```
matthieusieben, line 1: The [recommended](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/GUIDELINES.md) way of adding your own traces is to "monkey patch" the modules. Example here (express instrumentation):

https://github.com/open-telemetry/opentelemetry-js-contrib/blob/b3ca00387b00beb1ccd688fe46b8f938fbb09d7e/packages/instrumentation-express/src/instrumentation.ts#L52-L96

This allows to:

1) Force to write generic (re-usable) tracing (eg. avoid copy paste in `bsync`)
2) Completely split the tracing logic from the business logic
3) Avoid importing tracing utilities when not needed/used (not critical here, just a good practice)

It does make writing the instrumentation a little bit more involved & brittle though, but I believe it makes sense to adhere to best practices here.

services/bsky/package.json
estrattonbailey: b7b47c1af

packages/bsky/src/rpc-tracing.ts
estrattonbailey: Took me a sec to grok this, but def makes sense for the codebases that are reused by people downstream. For the appview, the approach this PR takes seems a little more clear and doesn't have the same downstream compat requirements, so maybe we leave it as is? Doesn't mean we can't move to this in the future if your work elsewhere makes that seamless.

Thoughts?

packages/bsky/src/telemetry.ts
estrattonbailey: 6efdf9572

packages/bsky/src/rpc-tracing.ts
matthieusieben: I have basically the same dilema as you do. I lean more towards doing it in the "recommended" way but I understand why you are taking this approach.

My take is; the approach here makes business logic intertwined with infra logic, which is something I personally want to avoid at all cost. Since our code is open-source, it can be run by different actors (this is esp. true for self-hoster, but also true for the appview), which may not have the same infra as we do. This approach also makes the code in this repo inconsistent (`pds` does not intertwine tracing with business logic; `bsky` does), which is undesirable.

That being said, I do understand and respect the pragmatism of the approach you took here.

PR #5202, opened 2026-07-08
Require createGroup members

### LINKED ISSUE #5085: The lexicon chat.bsky.group.createGroup has invalid members config
**Describe the bug**

It seems like a members field is missing the minimum members number. It definitely should be at least one

**Expected behavior**

The members field should have `minLength` equal to `1`. 

**Details**

Lexion:
https://github.com/bluesky-social/atproto/blob/cc329bf55c658752675d561dba792fedbe9c2626/lexicons/chat/bsky/group/createGroup.json#L23-L29

arjxnt:
## Summary
Adds `minLength: 1` to `chat.bsky.group.createGroup`'s `members` array so group creation requires at least one member.

## Root Cause
The lexicon required the `members` field and capped it at 49 entries, but did not reject an empty array.

Fixes #5085.

## Validation
- `pnpm exec prettier --check lexicons/chat/bsky/group/createGroup.json`
- `git diff --check`
- Parsed `lexicons/chat/bsky/group/createGroup.json` with `parseLexiconDoc`
- Validated `chat.bsky.group.createGroup` input accepts one member and rejects an empty `members` array

### DIFF
--- a/lexicons/chat/bsky/group/createGroup.json
+++ b/lexicons/chat/bsky/group/createGroup.json
@@ -22,6 +22,7 @@
           "properties": {
             "members": {
               "type": "array",
+              "minLength": 1,
               "maxLength": 49,
               "items": {
                 "type": "string",


matthieusieben:
cc @estrattonbailey not sure what is the expected behavior here.

rafaeleyng:
The group creator is always part of the group, automatically.

The API allows for creating a group with only the owner, but the app doesn't.

This was done for flexibility, in case we want to allow that case in the future.

Creating a group with only yourself is not unheard of, and can be useful. People used to create a group with someone on WhatsApp then remove the other member to have a group with themselves. Later, WhatsApp allowed support for doing that on group creation.

So, IMO, this is desired behavior (on the API).

PR #5204, opened 2026-07-08
Fix lexiconDoc safeParse errors

### LINKED ISSUE #4388: `lexUserType` zod validation throwing errors even when using `.safeParse`
Summary
---

I came across a weird bug when trying to validate lexicons using the `safeParse` method of the [lexiconDoc](https://github.com/bluesky-social/atproto/blob/9f87ff3aa60090c8c38b6ce400cba6ceff5cd2e9/packages/lexicon/src/types.ts#L426-L456) zod schema.
```ts
const result = lexiconDoc.safeParse(lexiconRecord);
```

This method should always return an object with the shape (simplified):
```
{
  sucess: bool,
  error: ZodError,
  data: object
}
```

However, sometimes this `.safeParse` will throw a `ZodError`. I thought that was weird and upon investigation I found that the `lexiconDoc` schema uses the `lexUserType` custom schema. Which the `lexUserType` schema makes additional zod validations, but does so with `.parse` which ends up throwing the error.

You can view the offending code here:

https://github.com/bluesky-social/atproto/blob/9f87ff3aa60090c8c38b6ce400cba6ceff5cd2e9/packages/lexicon/src/types.ts#L360-L393

It seems that this bug was introduced during this PR https://github.com/bluesky-social/atproto/pull/1040

Proposed Solution
---

It seems that we could refactor the `.parse` calls with `.safeParse` and that should fix the issue. However I'm uncertain of how to properly work with zod in such a way that it would understand those objects returned from a safe parse.

I would like someone with more knowledge and context around this part of the code to take this issue from me, as I feel I wouldn't have the knowledge to account for other potential issues.

arjxnt:
## Summary
Fixes `lexiconDoc.safeParse()` so invalid lexicon user type definitions return a failed parse result instead of throwing from nested `.parse()` calls.

## Root Cause
`lexUserType` used `z.custom()` with nested `.parse()` calls to dispatch by the `type` field. When a nested schema failed, the thrown `ZodError` escaped even through callers using `lexiconDoc.safeParse()`.

This keeps the fast custom dispatch, switches nested validation to `safeParse()`, and forwards nested issues into the outer refinement context.

Fixes #4388.

## Validation
- `NODE_OPTIONS='--experimental-vm-modules --localstorage-file=./work-jest-localstorage.db' pnpm --dir packages/lexicon exec jest --runInBand`
- `pnpm --filter @atproto/lexicon build`
- `pnpm exec prettier --check .changeset/quiet-lexicons-exist.md packages/lexicon/src/types.ts packages/lexicon/tests/general.test.ts`
- `pnpm changeset status --since=origin/main`
- `git diff --check`


matthieusieben (approved):

matthieusieben:
Hey @arjxnt thanks for you contributions! You seem to be using `lex-cli`, which is becoming deprecated. Any reason why you don't use `@aptroto/lex` ?

PR #5207, opened 2026-07-08
Add email-based authentication factor to reference PDS

ThisIsMissEm:
This PR supersedes #4221, which wouldn't have been an easy update due to the code drift between when it was opened and now.

Whilst building this, we made a deliberate call to make `updateEmail` a very simple compatibility pathway for enabling & disabling email-based 2FA, it's an open-call as to if we want explicit test coverage over the updateEmail pathway for toggling 2FA?

The localisation strings are extracted, and Claude has performed the localisations, as I don't speak french. Claude used feminine ("la 2FA"), stating "Grammatically mine is arguably more defensible — "2FA" abbreviates l'authentification à deux facteurs, which is feminine" — but social-app uses "le 2FA", it's an easy enough change depending on which is correct.

The pathway for disabling email based 2FA, via `disableEmailAuthFactor` in AccountManager **must** use `update_email` as the token type with `sendUpdateEmail`, since this is what the social-app expects, as it's using the `requestEmailUpdate` and then passing that token to `updateEmail` when toggling 2FA, rather than trying `updateEmail` once without a token, then once with a token (2-step).

It is currently expected that email-based 2FA is bypassed for app passwords, which now has test coverage.

I did note that `mock-api.ts` isn't actually asserted against the oauth-provider-api contract, which may be something we wish to improve in the future.

Currently the routes are called `/enable-email-otp` and `/disable-email-otp` in the oauth-provider-api, arguably we could change to `/enable-email-auth-factor` and `/disable-email-auth-factor` which would be more congruent with the naming through out the code.

## Screenshots
when email is unverified
when email is verified
image
image
image

TODO:
- delete the `ThisIsMissEm:feat/email-based-otp` branch once this is reviewed & ready for merge. I don't think we need anything from that branch still.


packages/pds/tests/_mailcatcher.ts
ThisIsMissEm, line 1: @matthieusieben / @devinivy I did end up doing this change, because the tests were getting messy repeating the same mail-catching code everywhere. It's not leveraged as much as it was when I decided to make that refactor though, as I since moved to using the internal APIs (`network.pds.ctx.accountManager.*`) for test setup, which is where I originally thought we'd need it a bit.

I think it still makes sense to do this change, since it centralises this logic and removes the types override in `_types.d.ts` for the tests.

packages/oauth/oauth-provider-ui/src/lib/api.ts
```diff
@@ -202,6 +204,39 @@ export class Api extends JsonClient<ApiEndpoints> {
...
 
+  async enableEmailAuthFactor(
+    {
+      did,
+      email,
+      locale = this.locale,
+    }: WithOptionalLocale<EnableEmailAuthFactorInput>,
+    options?: Options,
+  ) {
+    await this.fetch(
```
ThisIsMissEm: Claude did flag these as not returning the value; I'm not sure what the pattern is here as I couldn't quite discern if there even was a pattern in this file?

packages/oauth/oauth-provider-ui/src/components/enable-email-auth-factor-dialog.tsx
```diff
@@ -0,0 +1,49 @@
...
+      open={open}
+      onOpenChange={setOpen}
+      dismissable={!submitting}
+    >
+      <SmartForm
+        submitColor="primary"
+        submitLabel={
+          <>
+            <Trans>Enable</Trans>
+            <ShieldCheckIcon aria-hidden className="ml-1.5 size-5" />
```
ThisIsMissEm: The icon within the submit label is a little anachronistic within the code base, I'm not sure if we should keep it or not?

packages/oauth/oauth-provider-ui/src/pages/account/(authenticated)/manage/page.tsx
```diff
@@ -170,6 +177,53 @@ function PasswordUpdateRow(props: Omit<RowProps, 'icon' | 'value'>) {
...
+  const { did, email, emailVerified, emailAuthFactor } = account
+
+  const enableEmailAuthFactor = useEnableEmailAuthFactor()
+  const disableEmailAuthFactor = useDisableEmailAuthFactor()
+
+  // These endpoints requires an email, so if the user doesn't have one, we can't
+  // let them update their email auth factor. These users should not exist in
+  // normal conditions (may have been created manually by an admin), and are
+  // expected to contact support to update their password.
+  if (!email || !emailVerified) return null
```
ThisIsMissEm: It's debatable as to if we want to always show email-based auth factor status here, regardless of there either not being an email or not having a verified email? (and yeah, I did just catch the typo in the comment, it should refer to 2FA not "password")

packages/oauth/oauth-provider/src/account/account-manager.ts
```diff
@@ -182,6 +185,10 @@ export class AccountManager {
       const account = await callAsync(() =>
         this.store.authenticateAccount(data),
       ).catch(async (err) => {
+        if (err instanceof SecondAuthenticationFactorRequiredError) {
+          throw err
+        }
+
```
ThisIsMissEm: I don't think this is actually needed; my original thinking was that it was a condition we could fail fast on, but that'd assume an overlap between `SecondAuthenticationFactorRequiredError` and `InvalidRequestError`, which on second look there doesn't appear to be.

packages/pds/src/account-manager/oauth-store.ts
```diff
@@ -688,6 +688,36 @@ export class OAuthStore
     }
   }
 
+  async enableEmailAuthFactor({
+    did,
+    email,
+  }: EnableEmailAuthFactorInput): Promise<Account | null> {
+    const account = await this.accountManager.enableEmailAuthFactor({
+      did,
+      email,
```
ThisIsMissEm, line 718: I could add in a `locale` parameter here, but it's really not used anywhere downstream. The disable pathway has it though, as the `sendUpdateEmailMock` in `account-manager.test.ts` had an existing pattern of requiring it. It's not too much work to add it in if we want it for completeness.

packages/pds/src/api/com/atproto/server/createSession.ts
```diff
@@ -58,8 +59,17 @@ export default function (server: Server, ctx: AppContext) {
...
+          .login(body)
+          .catch((err) => {
+            if (err instanceof SecondAuthenticationFactorRequiredError) {
+              throw new AuthRequiredError(
+                'A sign in code has been sent to your email address',
+                'AuthFactorTokenRequired',
+              )
+            }
+            throw err
+          })
```
ThisIsMissEm: This change felt a little odd, should this be being handled somewhere else?

packages/pds/src/mailer/index.ts
```diff
@@ -20,11 +21,13 @@ export class ServerMailer {
     public readonly transporter: Transporter,
     private readonly email: EmailConfig | null,
     branding: BrandingConfig,
+    publicUrl: string,
   ) {
     transporter.use('compile', htmlToText())
 
     this.config = {
       serviceName: branding.name ?? 'Bluesky',
+      publicUrl: publicUrl,
```
ThisIsMissEm: I've made this "publicUrl" either the entryway URL or the publicUrl as configured, since it's maybe poorly named here, it's the URL for the `/accounts` page's host.

packages/pds/tests/oauth.test.ts
```diff
@@ -340,4 +342,143 @@ describe('oauth', () => {
...
+  describe('with 2FA', () => {
+    let mailCatcher: MailCatcher
+    let jane
+
+    beforeAll(async () => {
+      mailCatcher = new MailCatcher(network.pds.ctx.mailer)
+
+      const sc = network.getSeedClient()
+
+      jane = await sc.createAccount('jane', {
```
ThisIsMissEm, line 416: I did debate actually adding this `jane` account to the dev-env, but as we don't currently interact with the pds via the seed client in the manner below, I decided against it, and instead to just duplicate that code for now (I think it's three places)

ThisIsMissEm (commented):
Self-review flagging a few things I noted whilst working on this feature.

ThisIsMissEm:
Is `FAIL Ozone tests/blob-divert.test.ts` a flaky test? The failure doesn't appear to be related to my changes here.

packages/oauth/oauth-provider-ui/src/components/enable-email-auth-factor-dialog.tsx
matthieusieben: I'm fine either way

packages/oauth/oauth-provider-ui/src/lib/api.ts
matthieusieben: No pattern: Solely based on whether the return value is needed somewhere or not. Either way is fine.

packages/oauth/oauth-provider-ui/src/pages/account/(authenticated)/manage/page.tsx
matthieusieben: It might be better to show a disabled line, or a "verify email" CTA.

```ts
if (!email) return null

if (!emailVerified) {
  return (
    <Row {...props} disabled icon={ShieldWarningIcon} value={t`Verify email to enable`}>
      <Trans>Two-factor authentication (2FA)</Trans>
    </Row>
  )
}

```

packages/oauth/oauth-provider/src/account/account-manager.ts
matthieusieben: I wouldn't change this code path if not needed. And if we do need to change it, a comment would be nice.

packages/oauth/oauth-provider/src/account/account-store.ts
```diff
@@ -213,6 +217,19 @@ export interface AccountStore {
...
 
+  enableEmailAuthFactor(
+    data: EnableEmailAuthFactorInput,
+  ): Awaitable<Account | null>
+  /**
+   * Must trigger a verification email to be sent to the current email address
+   * when the `token` is undefined. An email-based OTP code will be sent and
+   * used to confirm that the account's email auth factor should indeed be
+   * disabled.
+   */
```
matthieusieben, line 256: This should also explain that `null` is to be returned in that case.

packages/oauth/oauth-provider/src/account/account-manager.ts
```diff
@@ -459,6 +466,70 @@ export class AccountManager {
...
+    // Disabling is a two-phase flow. When no token is supplied, the store
+    // dispatches an OTP email and returns `null` to signal the change is
+    // pending confirmation: nothing has changed yet, so return the account
+    // as-is and do not fire the "confirmed" hook. Once a valid token is
+    // supplied, the store actually disables the factor and returns the
+    // updated account, at which point the change is confirmed.
+    if (!updatedAccount) {
+      return account
+    }
```
matthieusieben: What happens if the email 2Fa is already disabled here (could happen if the user plays with two browser tabs) ? What is the store expected to return in that case ?

packages/oauth/oauth-provider/src/oauth-hooks.ts
```diff
@@ -178,6 +180,31 @@ export type OAuthHooks = {
     account: Account
   }) => Awaitable<void>
 
+  /**
+   * This hook is called when a user request to modify their email auth factor
+   * (OTP) settings, before the change is actually saved.
+   */
+  onUpdateEmailAuthFactor?: (data: {
+    input: EnableEmailAuthFactorInput | DisableEmailAuthFactorInput
```
matthieusieben: Analytics need to be able to distinguish enabling from disabling but the interfaces are too similar here to do so from a consumer perspective:

```ts
export type EnableEmailAuthFactorInput = {
  did: DidString
  email: string
  locale?: string
}

export type DisableEmailAuthFactorInput = {
  did: DidString
  email: string
  token?: string
  locale?: string
}
```

I would prefer we created a dedicated hook for each action.

packages/pds/src/account-manager/db/migrations/008-account-email-auth-factor.ts
```diff
@@ -0,0 +1,15 @@
...
+    .addColumn('emailAuthFactorAt', 'varchar')
+    .execute()
+}
+
+export async function down(db: Kysely<unknown>): Promise<void> {
+  await db.schema
+    .alterTable('account')
+    .dropColumn('emailAuthFactorAt')
+    .execute()
+}
```
matthieusieben, line 18: I know altering an sqlite table can be tedious. @devinivy is this fine ?

packages/pds/src/account-manager/helpers/account.ts
```diff
@@ -259,6 +261,34 @@ export const setEmailConfirmedAt = async (
...
+        .where('email', '=', email.toLowerCase()),
+    )
+  } else {
+    await db.executeWithRetry(
+      db.db
+        .updateTable('account')
+        .set({ emailAuthFactorAt })
+        .where('did', '=', did)
+        .where('email', '=', email.toLowerCase())
+        .where('emailConfirmedAt', 'is not', null),
```
matthieusieben: Should we return a boolean to indicate that the change was successful, allowing the caller to throw when trying to enable with non verified email ?

packages/pds/src/account-manager/account-manager.ts
```diff
@@ -896,6 +923,124 @@ export class AccountManager {
...
+    // Phase one: no token yet, send a one-time code and signal "pending". MUST
+    // use `update_email` due to social-app using requestEmailUpdate, not
+    // two-step updateEmail:
+    if (!token) {
+      const otp = await this.createEmailToken(did, 'update_email')
+      await this.mailer.sendUpdateEmail(
+        { token: otp, locale: locale },
+        { to: account.email },
+      )
+      return null
```
matthieusieben: Would it make sense to throw a "2fa token required error" here, to match the sign-in behavior?

packages/pds/src/account-manager/oauth-store.ts
matthieusieben: I've started adding it here, and adding TODOs in the mailer so that the reason why it is wired all the way to there is obvious when re-visiting the code.

packages/pds/src/api/com/atproto/server/createSession.ts
matthieusieben: Yeah. The way I've been going about this is to throw `AuthRequiredError` from the account manager, then translate those expected errors into OAuth specific errors (i.e. `SecondAuthenticationFactorRequiredError`) in the oauth-store.

packages/pds/src/mailer/index.ts
matthieusieben: The value to use is `cfg.oauth.issuer`, so maybe `oauthIssuer` ?

packages/pds/tests/_mailcatcher.ts
matthieusieben: You are right that the way things are done is not very clean:

```ts
    _origSendMail = mailer.transporter.sendMail
    mailer.transporter.sendMail = async (opts) => {
      const result = await _origSendMail.call(mailer.transporter, opts)
      mailCatcher.emit('mail', opts)
      return result
    }
```

The new recommended way of doing this is to use proper mocking:

```ts
using sendMail = jest.spyOn(mailer.transporter, 'sendMail')
  .mockImplementation(async (...args) => {
    expect(args).toBeDefined() // ensures the data is passed through
  })
```

See how things are done in the `oauth.test.ts`, which allows for much more "native" test writing:

https://github.com/bluesky-social/atproto/blob/55ef3aff486be060e4fcab6881fd8ef46e2a7cf6/packages/pds/tests/oauth.test.ts#L167-L176

matthieusieben (commented):
Looking pretty good! A few minor things to address but we should be able merge this pretty soon 🥳

packages/pds/src/api/com/atproto/server/createSession.ts
ThisIsMissEm: Though the account manager's contract is defined by the oauth provider, so I'd think more specific errors transformed to less specific errors would probably be the way to go?

packages/pds/src/account-manager/account-manager.ts
ThisIsMissEm: I'm a bit concerned this could bubble throw to the OAuth UI incorrectly, but I'm making this change to be more like `requestEmailUpdate` which returns a `{ tokenRequired: boolean }` shape. So I'll have `{ account: Account, tokenRequired: boolean }` — I've also pre-emptively added it on both enable and disable, keeping them aligned, even though currently enable would always be false.

packages/oauth/oauth-provider-ui/src/components/enable-email-auth-factor-dialog.tsx
ThisIsMissEm: I don't hate it, but it's probably more a product design consistency matter, we don't normally use icons within buttons as far as I know.

packages/oauth/oauth-provider-ui/src/lib/api.ts
ThisIsMissEm: Yeah, here we passively get the return value via `session.tsx` which is hooked into the data loading.

packages/pds/src/api/com/atproto/server/createSession.ts
matthieusieben: > Though the account manager's contract is defined by the oauth provider

The account manager's contract is actually historically defined by the xprc endpoints. It has changed since the introduction of the oauth-provider to be able to serve both, but it currently does not depend on any oauth stuff, and we should keep it that way.

When it comes to errors, I've typically tried to keep the exact same current xrpc errors, and move them inside the account manager (when appropriate), and them translate the XrpcError into OAuthErrors, parsing error messages if need be (I know, it's not ideal but it allows to keep the xrpc paths as little impacted as possible).

There are some instances where we use a locally defined error class (`UserAlreadyExistsError`) which are then translated into XrpcError and OAuthError, and that is fine though in this instance, and if possible, I would recommend making the AccountManager throw the `AuthRequiredError` and translate that in the oauth-store into the appropriate behavior (return null or throw a wrapped error, whichever the interface requires).

packages/oauth/oauth-provider/src/oauth-hooks.ts
ThisIsMissEm: Okay, I'll do that! I didn't realise these were used for analytics, I thought they were perhaps security related.

packages/pds/src/account-manager/db/migrations/008-account-email-auth-factor.ts
ThisIsMissEm: The other path that we can go here is up-fronting the work of introducing an account_auth_factor table, which would support email / otp / passkeys. I think it'd be doable with a `id / type / data / enabledAt` type setup, where `data` is a blob of json relevant for the required auth factor method, but I haven't done anything to confirm this gut feeling yet.

ThisIsMissEm: I did also pay really close attention to this with claude and had it analyse things here to try to ensure this migration was safe, as I know we've historically shied away from database migrations.

packages/pds/src/account-manager/helpers/account.ts
ThisIsMissEm: I think that's be determined by whether or not the operation threw, no? We don't generally consistently use `numUpdatedRows`, but if we want we can do that. I don't know if it changes anything that much since this is in a retry and the likely hood of it doing nothing doesn't really change the state returned (which is based on the account)

packages/pds/src/account-manager/oauth-store.ts
ThisIsMissEm: I'll do this is a follow-up commit (not PR), just wasn't sure which stance to take so went with less code.

packages/pds/src/api/com/atproto/server/createSession.ts
ThisIsMissEm: I feel like more-specific errors classes is probably better long term, and then we can automatically translate them to less specific — going the other way is typically much harder.

packages/pds/tests/_mailcatcher.ts
ThisIsMissEm: Is using spy's the ideal method going forwards? I don't mind spinning a refactor towards that into a separate PR, I just found myself starting to copy/paste this code around a bunch, and didn't really want to be adding tech debt.

Would a migration towards vitest be on the books at some point in the future? If so, it may be worth checking which approach is more long-term maintainable if you've not already done so.

ThisIsMissEm (commented):
I'm currently working on several changes to respond to feedback more fully.

packages/oauth/oauth-provider/src/account/account-manager.ts
ThisIsMissEm: This ends up being a no-op now, with the confirmed hook not firing.

ThisIsMissEm: Actually, the no-op here would still fire both hooks based on all the other code in this file, so a no-op if changed from another browser tab will still fire hooks, same for other methods.

packages/pds/src/account-manager/db/migrations/008-account-email-auth-factor.ts
fatfingers23: I think doing it as a separate table sounds good and I am in favor of it. For the 2 reasons listed. Would be nice to setup for future MFA work/types and letting you shy away from doing a migration on this table

ThisIsMissEm: I don't think I can upfront that work though, because I'm not sure there's clear enough insight into what different MFA methods would actually need, e.g., for passkeys there's a counter that's needed.

packages/pds/tests/_mailcatcher.ts
ThisIsMissEm: @matthieusieben I've taken the mail catching changes out of this PR, but one thing I noticed whilst doing that is that we explicitly should avoid:
```
.mockImplementation(async (...args) => {
    expect(args).toBeDefined() // ensures the data is passed through
  })
```

And instead assert on the `lastCall` since the mocked implementation's expectation throwing gets translated into an Internal Server Error via XRPCError, which effectively swallows the tests failure.

ThisIsMissEm: Extracted to: https://github.com/bluesky-social/atproto/pull/5246

packages/pds/src/account-manager/oauth-store.ts
ThisIsMissEm: Okay, so I added to the interfaces here, but the enableEmailAuthFactor never sends any emails, so it's unused in practice. It is passed through for disableEmailAuthFactor which does send emails.

packages/pds/src/api/com/atproto/server/createSession.ts
ThisIsMissEm: Have fixed this.

ThisIsMissEm:
@matthieusieben @devinivy okay, so I've refactored to use a new table for the email auth factor being enabled, per our previous discussion, but I don't like the result — it makes the `getAccount` hot-path need to do a join on `account_email_auth_factor` and I honestly don't think it's as readable as a column. It also makes some things we had for free with the extra column based approach much more complicated.

I've pushed the change as a single commit so we can reason about the approach.

packages/pds/src/account-manager/db/migrations/008-account-email-auth-factor.ts
ThisIsMissEm: See: https://github.com/bluesky-social/atproto/pull/5207/changes/43f216de87f15342857aae8aaa31c5499341d283

matthieusieben (approved):
Looks good to me. I like the way this landed.

packages/pds/src/api/com/atproto/server/updateEmail.ts
```diff
@@ -31,12 +31,58 @@ export default function (server: Server, ctx: AppContext) {
...
+            // `update_email` OTP: the first call (no token) emails a code and
+            // makes no change; the second (with token) verifies and disables.
+            const account = await ctx.accountManager.disableEmailAuthFactor({
+              did,
+              email: user.email,
+              token,
+              locale,
+            })
+
+            if (account === null) {
```
matthieusieben: This is a dead `if`. Did you mean to de-structure `{ account }` ?

packages/pds/src/api/com/atproto/server/updateEmail.ts
```diff
@@ -31,12 +31,58 @@ export default function (server: Server, ctx: AppContext) {
...
         // @TODO get the locale somehow (either by adding a field in the request
         // body, or by using the `Accept-Language` header).
         const locale = undefined
 
+        // Pure auth-factor toggle: the email isn't changing, the caller is only
+        // flipping the OTP factor on/off. Handle it and return; falling
+        // through to updateEmail() would re-set the (unchanged) email and null
+        // out emailConfirmedAt, silently un-confirming a confirmed address.
+        if (
+          user.email === email &&
```
matthieusieben: Do we need to account for different casing here ?

matthieusieben (changes requested):
Small issue;

packages/pds/src/api/com/atproto/server/updateEmail.ts
ThisIsMissEm: Yeah, the destructuring was what was intended, pushing a fix. Mildly concerned the tests didn't catch this.

ThisIsMissEm: Turns out I missed test coverage on the XRPC endpoint.

packages/pds/src/api/com/atproto/server/updateEmail.ts
ThisIsMissEm: Fixed with tests

packages/pds/src/api/com/atproto/server/updateEmail.ts
```diff
@@ -31,12 +31,65 @@ export default function (server: Server, ctx: AppContext) {
...
+        // Pure auth-factor toggle: the email isn't changing, the caller is only
+        // flipping the OTP factor on/off. Handle it and return; falling
+        // through to updateEmail() would re-set the (unchanged) email and null
+        // out emailConfirmedAt, silently un-confirming a confirmed address.
+        // Emails are normalized to lowercase:
+        if (
+          user.email === email.toLowerCase() &&
+          user.emailConfirmedAt &&
+          emailAuthFactor !== undefined
+        ) {
```
matthieusieben: Would it make sense to error when attempting to change `emailAuthFactor` when the email does not match (instead of silently ignoring):

```ts

const hasEmailAuthFactor = user.emailConfirmedAt != null

if (emailAuthFactor != null && emailAuthFactor !== hasEmailAuthFactor) {
  // User is trying to change email otp status

  if (user.email !== email.toLowerCase()) {
    throw new InvalidRequestError("Please change and validate your email bebore enabling OTP")
  }

  if (emailAuthFactor) {
    // Enabling
  } else {
    // Disabling
  }

  // No need to continue to change email (since we checked above that it didn't change)
  return
}

```

packages/pds/src/api/com/atproto/server/updateEmail.ts
```diff
@@ -41,53 +41,59 @@ export default function (server: Server, ctx: AppContext) {
...
-        // flipping the OTP factor on/off. Handle it and return; falling
-        // through to updateEmail() would re-set the (unchanged) email and null
-        // out emailConfirmedAt, silently un-confirming a confirmed address.
-        // Emails are normalized to lowercase:
-        if (
-          user.email === email.toLowerCase() &&
-          user.emailConfirmedAt &&
-          emailAuthFactor !== undefined
-        ) {
+        const hasEmailAuthFactor = user.emailConfirmedAt != null
```
ThisIsMissEm, line 50: This doesn't do the same logic, this checks if the email was confirmed, not if they'd enabled 2FA

PR #5215, opened 2026-07-09
Fix PDS's describe server in entryway setup

matthieusieben:
In a setup where account management is performed by an Entryway (like Bluesky's deployment), account creation logic should not be allowed on PDSs.

Similarly, some configuration should be dictated by the Entryway, and not individual PDSs.

This PR addresses that by delegating part of the `describeServer` response to entryway values.

Most notably, it removes the `inviteCodeRequired` property from the output since this value does not makes sense on a server that should not handle `createAccount`.

This PR does not remove the ability to call `createAccount` though, which might be part of a larger rework.

### DIFF
--- a/.changeset/common-cases-travel.md
+++ b/.changeset/common-cases-travel.md
@@ -1,0 +1,5 @@
+---
+'@atproto/pds': patch
+---
+
+Read part of the `describeServer` config from entryway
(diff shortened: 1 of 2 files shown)


mozzius (approved):
Looks about right to me

devinivy:
Instead going down a different path for the time being— allowing the local service domain config to safely be set on PDS instances behind an entryway.

Replaced by #5236.

PR #5217, opened 2026-07-09
examples: add post-with-facets script

ralyodio:
Demonstrates that clickable #hashtags, links, and @mentions come from a client-generated `facets` array (via RichText.detectFacets), not from any server-side parsing. A raw createRecord with only { text, createdAt } renders tags/links as plain text; this example shows the missing facet-detection step.

### DIFF
--- a/examples/post-with-facets/README.md
+++ b/examples/post-with-facets/README.md
@@ -1,0 +1,54 @@
+# Posting with facets (clickable #tags, links, and @mentions)
+
+Bluesky does **not** parse your post text. A post record stores `text` literally —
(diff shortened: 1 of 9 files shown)


matthieusieben:
Hey! Thanks for trying to contribute here!

While I do see the value of this feature, this is not one we are willing to merge. The main reason is that we want to keep the reference PDS implementation app logic agnostic. I know there is still some `app.bsky.*` stuff in there. It is a goal to eventually get rid of it.

The proper way to implement a feature like this one is from the client (either front-end, or back-end for BFF clients).

PR #5222, opened 2026-07-10
Sanitize bidi control characters in display names

### LINKED ISSUE #1480: Directional Unicode characters in display names break notifications for other users
I have come across multiple instances of people complaining that their notification text changed directions when users with directional control characters in their usernames engaged with their posts.

The bug appears to be confusing users and reducing the legibility of notifications when it arises.

Could Bluesky either sanitize directional control characters in display names or else check for them and reset direction to the default for a user’s locale upon referencing them?

**examples:**
https://bsky.app/profile/birdapp.rip/post/3k527wpl3jx2d
https://bsky.app/profile/pangolinfox.bsky.social/post/3k2kinjtfj32x

a-y-ibrahim:
Fixes #1480

## Problem

`displayName` records are validated only for length (`maxGraphemes`,
`maxLength`), there is no character-level validation. A display name
containing a bidi embedding/override character (for example U+202E
RIGHT-TO-LEFT OVERRIDE) is accepted as-is and served unchanged, so
notification text and verification-badge text built from that display
name can render in a confusing or reversed direction for other users.

I confirmed this is still reproducible today: I set a display name
containing U+202E on a real account and fetched it back through the
public `getProfile` endpoint, the character round-trips unchanged.

## Fix

Adds `hasBidiControls` / `stripBidiControls` to `@atproto/syntax` and
applies `stripBidiControls` everywhere I found `displayName` independently
built into a `ProfileViewBasic`-shaped response:

- `profileBasic` in bsky views, the value that flows into notification
  author views
- `verification` in bsky views, the issuer's display name shown on
  verification badges
- `LocalViewer.getProfileBasic` and `LocalViewer.updateProfileViewBasic`
  in the PDS's read-after-write path, which independently builds/patches
  the same view shape and could otherwise reintroduce the unsanitized
  value right after a profile edit

Legitimate single-character direction marks (LRM, RLM, ALM) and script
joiners (ZWJ, ZWNJ, used by real RTL text and emoji) are intentionally
left untouched, only embedding/override/isolate control characters are
stripped.

## Scope

This PR sanitizes `displayName` at every read/view-construction site I
could find for `app.bsky.actor.profile`. Two things I deliberately left
out, happy to follow up separately if maintainers want them here instead:

- Write-time validation/rejection in the PDS record-validation path
  (`packages/pds/src/repo/prepare.ts`). I chose read-side sanitization
  because it protects every consumer regardless of which PDS wrote the
  record (this is a federated protocol), but it means a malicious
  `displayName` is still accepted and stored, just no longer served
  unsanitized through the paths this PR touches.
- `app.bsky.feed.generator#displayName` (a feed generator's own name) is
  a different lexicon field with a different threat model, not covered
  here since issue #1480 is specifically about actor identity.

## Testing

- Added unit tests in `packages/syntax/tests/bidi.test.ts` covering the
  attack character class, every embedding/override/isolate character, and
  that legitimate direction marks and emoji ZWJ sequences are left intact.
- Verified against the compiled `Views` class directly (`profileBasic`
  and `verification`), not just the standalone utility, with a malicious
  display name, a clean display name, a real Arabic display name with a
  legitimate direction mark, a missing profile, and a stacked multi-character
  attack.

### DIFF
--- a/packages/syntax/src/bidi.ts
+++ b/packages/syntax/src/bidi.ts
@@ -1,0 +1,20 @@
+// Directional embedding/override/isolate characters. These can make text
+// render in an order different from its logical character order, the
+// mechanism behind the "Trojan Source" disclosure (CVE-2021-42574) and the
+// class of bug reported in bluesky-social/atproto#1480, where a display
+// name containing one of these characters made notification text render in
+// a confusing or reversed direction for other users.
+//
+// Single-character directional *marks* (U+200E LRM, U+200F RLM, U+061C ALM)
+// are intentionally excluded: they have no reordering power on their own,
+// and real Arabic/Hebrew text can legitimately contain them.
+const BIDI_CONTROL_CHARS = /[\u202A-\u202E\u2066-\u2069]/
+const BIDI_CONTROL_CHARS_G = /[\u202A-\u202E\u2066-\u2069]/g
+
+export function hasBidiControls(input: string): boolean {
+  return BIDI_CONTROL_CHARS.test(input)
+}
+
+export function stripBidiControls(input: string): string {
+  return input.replace(BIDI_CONTROL_CHARS_G, '')
+}
--- a/packages/bsky/src/views/index.ts
+++ b/packages/bsky/src/views/index.ts
@@ -14,6 +14,7 @@ import {
   type DidString,
   INVALID_HANDLE,
   normalizeDatetimeAlways,
+  stripBidiControls,
 } from '@atproto/syntax'
 import type { Actor, ProfileViewerState } from '../hydration/actor.js'
 import {
@@ -389,7 +390,9 @@ export class Views {
     return {
       did,
       handle: actor.handle ?? INVALID_HANDLE,
-      displayName: actor.profile?.displayName,
+      displayName:
+        actor.profile?.displayName &&
+        stripBidiControls(actor.profile.displayName),
       pronouns: actor.profile?.pronouns,
       avatar: actor.profile?.avatar
         ? this.imgUriBuilder.getPresetUri(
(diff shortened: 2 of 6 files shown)


a-y-ibrahim:
Addressed Copilot's feedback:

- The two "will fail to typecheck under strict settings" comments: I checked this empirically with an isolated repro against this repo's own `tsgo` in strict mode, both patterns typecheck cleanly (0 errors). The underlying suggestion (avoid re-dereferencing the optional chain) is still good practice for readability, so I applied it in 86c95a2.
- Missing bsky-level regression test: added one to `profile.test.ts` in 86c95a2, asserting `getProfile` strips a bidi override character from `displayName`. I did not add an equivalent to `verification.test.ts`, that file's coverage is built around a large shared `verificationsSeed` fixture and a data-driven `it.each` table; bolting a one-off case onto it without fully understanding how the fixture is reused elsewhere seemed more likely to introduce a fragile test than a good one. That path is still covered by the `hasBidiControls`/`stripBidiControls` unit tests and by directly exercising the compiled `Views.verification()` method during development (not committed, just how I verified it locally).

matthieusieben (approved):

matthieusieben:
@claude Read .github/claude-review-prompt.md in the checked-out repo and review this pull request following its guidance.

packages/syntax/src/bidi.ts
```diff
@@ -0,0 +1,20 @@
...
+const BIDI_CONTROL_CHARS = /[\u202A-\u202E\u2066-\u2069]/
+const BIDI_CONTROL_CHARS_G = /[\u202A-\u202E\u2066-\u2069]/g
+
+export function hasBidiControls(input: string): boolean {
+  return BIDI_CONTROL_CHARS.test(input)
+}
+
+export function stripBidiControls(input: string): string {
+  return input.replace(BIDI_CONTROL_CHARS_G, '')
+}
```
matthieusieben, line 20: I don't think we want this as part of the syntax package. Syntax is a low level package, while this is clearly business logic.

matthieusieben (commented):
This should be tackled in the app itself. Please open a PR in https://github.com/bluesky-social/social-app/

matthieusieben:
Thank you for you interest. I don't think that this is the right fix for that. bidi chars as part of any atproto text and should be handled whenever rendered to the user.

matthieusieben:
Feel free to open a discussion if you want to tackle this further.

PR #5226, opened 2026-07-12
oauth: keep refreshable sessions in connected apps

### LINKED ISSUE #3998: "Connected apps" only show recent apps
### Steps to Reproduce

1. Authorize some application via oauth2;
2. Wait some time (I'd assume at least an hour, maybe more);
3. Go to https://bsky.social/account;
4. The app you authorized shouldn't be there;
5. Use[/interact with] the app you authorized;
6. It should apppear again in https://bsky.social/account

### Attachments

https://github.com/user-attachments/assets/2c9fc5f2-5fad-4ffd-ac17-506f286de8a7

### What platform(s) does this occur on?

Web (Desktop)

### Device Info

_No response_

### What version of the app are you using?

what ever version bsky.social used at May 11th, 11:16 AM UTC+2

### Additional Information

The "Connected apps" section should not just list the applications that have recently used my account (i.e.: like creating a `app.bsky.feed.like`-record as shown in the video), but rather should list *all* oauth2 applications that have access to my account.

The wording/description on the page (https://bsky.social/account) doesn't clarify this either.

AG0708:
## Summary

- keep refreshable OAuth sessions visible after the current access token expires
- apply the existing public vs. confidential/first-party refresh and session lifetimes before listing sessions
- add focused Vitest coverage for token lifecycle boundaries and client classification

## Root cause

`TokenManager.listAccountTokens()` treated the current access-token expiry as the OAuth-session expiry. Access tokens expire after an hour, while a client with a refresh token can retain access for the applicable refresh and session lifetime. The account UI therefore hid an app until its next token refresh updated `expiresAt`.

The route now cheaply drops expired access-only rows before metadata loading, then classifies the remaining clients and applies the same lifetime calculation used by refresh-grant validation. If current client metadata is unavailable, classification falls back to the stored authentication method so the existing partial-session UI remains useful.

## User impact

Connected Apps continues to show apps that can refresh their credentials, while excluding access-only, refresh-expired, and session-expired entries.

Fixes #3998

## Validation

- `pnpm test` in `packages/oauth/oauth-provider` (16 tests)
- `pnpm build --force`
- `pnpm verify`

### DIFF
--- a/packages/oauth/oauth-provider/src/oauth-provider.ts
+++ b/packages/oauth/oauth-provider/src/oauth-provider.ts
@@ -74,14 +74,7 @@ import {
   type CustomMetadata,
   buildMetadata,
 } from './metadata/build-metadata.js'
-import {
-  AUTHENTICATION_MAX_AGE,
-  CONFIDENTIAL_CLIENT_REFRESH_LIFETIME,
-  CONFIDENTIAL_CLIENT_SESSION_LIFETIME,
-  PUBLIC_CLIENT_REFRESH_LIFETIME,
-  PUBLIC_CLIENT_SESSION_LIFETIME,
-  TOKEN_MAX_AGE,
-} from './oauth-constants.js'
+import { AUTHENTICATION_MAX_AGE, TOKEN_MAX_AGE } from './oauth-constants.js'
 import type { OAuthHooks } from './oauth-hooks.js'
 import {
   type DpopProof,
@@ -102,6 +95,7 @@ import type { AccessTokenPayload } from './signer/access-token-payload.js'
 import { refreshTokenSchema } from './token/refresh-token.js'
 import type { TokenData } from './token/token-data.js'
 import { TokenManager } from './token/token-manager.js'
+import { getOAuthSessionLifetimes } from './token/token-session.js'
 import { type TokenStore, asTokenStore } from './token/token-store.js'
 import { isPARResponseError } from './types/par-response-error.js'
 
@@ -1037,13 +1031,9 @@ export class OAuthProvider extends OAuthVerifier {
     clientAuth: ClientAuth,
     data: TokenData,
   ): Promise<void> {
-    const [sessionLifetime, refreshLifetime] =
-      clientAuth.method !== 'none' || client.info.isFirstParty
-        ? [
-            CONFIDENTIAL_CLIENT_SESSION_LIFETIME,
-            CONFIDENTIAL_CLIENT_REFRESH_LIFETIME,
-          ]
-        : [PUBLIC_CLIENT_SESSION_LIFETIME, PUBLIC_CLIENT_REFRESH_LIFETIME]
+    const [sessionLifetime, refreshLifetime] = getOAuthSessionLifetimes(
+      clientAuth.method !== 'none' || client.info.isFirstParty,
+    )
 
     const sessionAge = Date.now() - data.createdAt.getTime()
     if (sessionAge > sessionLifetime) {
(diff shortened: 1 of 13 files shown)


packages/oauth/oauth-provider/src/router/create-api-middleware.ts
```diff
@@ -505,23 +517,39 @@ export function createApiMiddleware<
...
-        // TokenInfo are stored (see TokenManager#isTokenExpired and
-        // TokenManager#isTokenInactive).
-        const json = tokenInfos.map(({ id, data }): ActiveOAuthSession => {
-          return {
-            tokenId: id,
+        const activeTokenInfos = tokenCandidates.filter((tokenInfo) => {
+          const client = clients.get(tokenInfo.data.clientId)
+          const usesExtendedSessionLifetime = usesExtendedOAuthSessionLifetime(
+            tokenInfo,
+            client && {
```
matthieusieben, line 524: Because `onError` above will swallow errors, a missing `client` is not a sign that the session is "short lived". This is what the previous `@TODO` was trying to capture, and the reason this was not done previously.

The solution should either:
- propagate client load errors (failed API call, not ideal)
- show the session in a "client broken" state
- use the data model to determine session lifetime

But the solution *MUST NOT* show the session as short lived only because metadata were not loaded.

packages/oauth/oauth-provider/src/oauth-provider.ts
```diff
@@ -1037,13 +1031,9 @@ export class OAuthProvider extends OAuthVerifier {
...
-    const [sessionLifetime, refreshLifetime] =
-      clientAuth.method !== 'none' || client.info.isFirstParty
-        ? [
-            CONFIDENTIAL_CLIENT_SESSION_LIFETIME,
-            CONFIDENTIAL_CLIENT_REFRESH_LIFETIME,
-          ]
-        : [PUBLIC_CLIENT_SESSION_LIFETIME, PUBLIC_CLIENT_REFRESH_LIFETIME]
+    const [sessionLifetime, refreshLifetime] = getOAuthSessionLifetimes(
+      clientAuth.method !== 'none' || client.info.isFirstParty,
+    )
```
matthieusieben, line 1036: We can probably replace this with a getter on the `Client` class. Something like:

```suggestion
    const { sessionLifetime, refreshLifetime } = client
```


```ts
export class Client {

  //...


  get isFirstParty(): boolean {
    return this.info.isFirstParty
  }

  get isConfidential(): boolean {
    return this.metadata.token_endpoint_auth_method !== 'none'
  }

  get refreshLifetime() {
    return this.isConfidential || this.isFirstParty
      ? CONFIDENTIAL_CLIENT_REFRESH_LIFETIME
      : PUBLIC_CLIENT_REFRESH_LIFETIME
  }

  get sessionLifetime() {
    return this.isConfidential || this.isFirstParty
      ? CONFIDENTIAL_CLIENT_SESSION_LIFETIME
      : PUBLIC_CLIENT_SESSION_LIFETIME
  }
}
```

Note that we don't really need the `clientAuth` to compute it (the same info is available in the client metadata, as you already did elsewhere).

matthieusieben (changes requested):
Hey, Thank you for contributing here. This is a welcome fix for a long standing issue.

There are still a few things I'd like to see be done differently (see comments).

Also, we should probably rename `CONFIDENTIAL_CLIENT_*_LIFETIME` into `EXTENDED_CLIENT_*_LIFETIME`.

matthieusieben:
Thank you for tracking this. #5273 should fix this issue.

PR #5228, opened 2026-07-13
test(lex-client): lock in that a 404 with an XRPC payload surfaces the server error

### LINKED ISSUE #5058: @atproto/lex: xrpcSafe always returns XRPCNotSupported when response is 404
**Describe the bug**

When using `@atproto/lex`'s xrpcSafe, the library will always override the responses `error` field with `XRPCNotSupported` when the http status code is 404, regardless of what the body indicates.

According to the specification, this behaviour seems wrong:

> 404 Not Found: Can indicate a missing resource. This can also indicate that the server does not support atproto, or does not support this endpoint. See response error message (or lack thereof) to clarify.

I am trying to match on `error` to see what type of 404 it is, but when the library always shadows it with XRPCNotSupported I cannot clearly determine it for myself.

**To Reproduce**

```typescript
const result = await auth.client.xrpcSafe(myEndpoint, {
  params: { actor: params.actor },
});
if (result.success) {
  //
}

if (result.error ==="ActorNotFound" || result.error == "ProfileNotFound") {
  // Never reached, result.error has been overwritten by library.
}

```

**Expected behavior**

The library surfaces the origin error message when a response body is available, falls back to this behaviour if one is not.

ryanda9910:
Closes #5058 (as already-fixed — see below).

### What's going on with #5058

The reported bug is **already fixed on `main`**. `XrpcResponseError` gates the `XRPCNotSupported` status fallback behind `isXrpcErrorPayload`, so a 404 whose body is a valid XRPC error is surfaced as-is:

| 404 response | behaviour on `main` |
|---|---|
| `{"error":"ProfileNotFound"}` | `ProfileNotFound` ✅ |
| `{"error":"AccountDeactivated"}` | `AccountDeactivated` ✅ |
| plain-text / empty body | `XRPCNotSupported` (fallback) ✅ |

That matches the expected behaviour in the issue, so there is no production fix to make here.

### Why this PR then

**Nothing tests it.** `errors.test.ts` covers the fallback path (404 with no payload → `XRPCNotSupported`) and the payload-wins case for **400**, but never for **404** — which is the one status where the shadowing actually bites, because `XRPCNotSupported` reads like "this endpoint doesn't exist" and hides domain errors like a missing actor.

So a future refactor of the error path could silently reintroduce #5058 and every test would stay green.

This PR adds three cases to close that gap:
- a 404 carrying `ProfileNotFound` surfaces the server's code (and explicitly *not* `XRPCNotSupported`)
- same for `ActorNotFound`
- a payload-less 404 still falls back to `XRPCNotSupported`

### Verification

- `pnpm vitest run src/errors.test.ts` → **60 passed**
- Mutation-checked: flipping the `isXrpcErrorPayload` gate off (i.e. reintroducing the bug) makes the new tests **fail**, so they genuinely lock the behaviour rather than passing vacuously.
- Prettier clean. **No production code touched** — the diff is test-only (+29 lines).

Happy to close this and just close the issue instead if you'd rather not carry the extra tests.


devinivy (approved):
Looks good, thanks!

PR #5230, opened 2026-07-14
Fix .gitignore entry for lexicons directory

DenysVuika:
Update .gitignore instruction to use absolute path.


matthieusieben:
Thank you

PR #5231, opened 2026-07-14
feature: Adds rate limit headers to Access-Control-Expose-Headers on xrpc servers (PDS)

fatfingers23:
Not sure if this is the best path, but while adding rate limiting in [pdsmoover.com](https://pdsmoover.com) I found that I could not access the rate limit headers since it is all client side. After some research I found that `Access-Control-Expose-Headers` needed to have them set. This fix resolved my issues for me and allowed me to access the rate limit headers on the browser so I can do things like "Limit blob uploads to leave 5 so that they can still upload pics if limit is hit during migrations" which is not possible since it goes till rate limit and the user is out of luck for 24 hours. 

Happy too if there is another path to be able to add this. But would really appreciate some kind of resolution to give users this.

### DIFF
--- a/.changeset/pink-trams-reply.md
+++ b/.changeset/pink-trams-reply.md
@@ -1,0 +1,5 @@
+---
+'@atproto/xrpc-server': patch
+---
+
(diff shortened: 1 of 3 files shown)


matthieusieben (approved):
This looks good to me.

matthieusieben:
It looks like a changeset is required though.

fatfingers23:
Thank you! Just rebased against main and added one, first time using changeset so not 100% sure if there's anything I'm missing or should do different.

matthieusieben (approved):

PR #5236, opened 2026-07-14
Support entryway-matching `PDS_SERVICE_HANDLE_DOMAINS` on PDS

devinivy:
When a PDS runs behind an entryway, `PDS_SERVICE_HANDLE_DOMAINS` could not be configured to match the entryway's handle domains: any handle under those domains that wasn't hosted locally would fail resolution, even though it may belong to an account on a sibling PDS behind the same entryway.

- `com.atproto.identity.resolveHandle` no longer consults the PDS's local account store when an entryway is configured — the entryway is authoritative for handles, so resolution always goes through the appview or network. This also prevents serving local handle state that has diverged from the entryway.
- `/.well-known/atproto-did` no longer answers authoritatively for service domains when an entryway is configured, deferring handle resolution to the entryway.
- dev-env's `mockResolvers()` accepts an optional URL controlling where handle resolution is served from, and the mock entryway now implements `/.well-known/atproto-did` — used by new entryway test cases covering resolution of hosted accounts, accounts on sibling PDSes, diverged local handle state, and well-known deferral.

### DIFF
--- a/.changeset/violet-poems-attack.md
+++ b/.changeset/violet-poems-attack.md
@@ -1,0 +1,6 @@
+---
+'@atproto/pds': patch
+'@atproto/dev-env': patch
+---
+
+defer service handle resolution to entryway when configured, allowing PDS_SERVICE_HANDLE_DOMAINS to match the entryway's domains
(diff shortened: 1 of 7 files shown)


ThisIsMissEm (commented):
@devinivy question on implementation: usually when behind an entryway requests are just proxied wholesale to the entryway, is there a reason not to do that here?

PR #5241, opened 2026-07-16
:sparkles: Add bulk close report endpoint and enhance queryReport

foysalit:


packages/ozone/src/mod-service/report.ts
gcwill70, line 266: This PR had me looking through this logic and I forgot to add convo handling here when adding support in these endpoints. The current function will include chat and convo reports when targeting an account.


So something like:
```
if (params.subjectUri) {
    const uri = new AtUri(params.subjectUri)
    if (uri.collection === CHAT_MESSAGE_COLLECTION) {
      builder = builder.where('r.subjectMessageId', '=', uri.rkey)
    } else if (uri.collection === CHAT_CONVO_COLLECTION) {
      builder = builder
        .where('r.subjectConvoId', '=', uri.rkey)
        .where('r.subjectMessageId', 'is', null)
    } else {
      builder = builder.where(
        'r.recordPath',
        '=',
        `${uri.collection}/${uri.rkey}`,
      )
    }
  } else {
    builder = builder
      .where('r.recordPath', '=', '')
      .where('r.subjectMessageId', 'is', null)
      .where('r.subjectConvoId', 'is', null)
  }
```

gcwill70 (changes requested):
Looks good but probably should add in the missing convo handling before merging

packages/ozone/src/mod-service/report.ts
foysalit: nice catch! fixed those up.

gcwill70 (approved):

PR #5242, opened 2026-07-16
Overhaul @atproto/ws-client: isomorphic WebSocketClient and WebSocketConnection

devinivy:
Rebuilds `@atproto/ws-client` around two new isomorphic classes, replacing `WebSocketKeepAlive`. There is a single public entrypoint whose API works identically in Node.js (via `ws`) and the browser (via native `WebSocket`) — platform routing happens internally through the package `imports` field.

`WebSocketClient` is a robust client that prioritizes liveness, flow control, and transparent reconnects:
- Messages are read with `for await`; the stream spans reconnects. `dataMode` (`'auto' | 'text' | 'binary'`) types what the iterator yields and is enforced at runtime.
- Constructing a client doesn't open it.  Instead the connection opens lazily once it's iterated, and stops via `close()` (idempotent), an aborted `signal`, or breaking out of the loop. Lifecycle is observable through the `onOpen`/`onReconnect`/`onError`/`onClose` hook options; `onClose` fires exactly once per started client.
- Reconnects use exponential backoff and are classified by a typed error taxonomy plus WebSocket close codes (no more errno string-matching) — each error classifies itself via `shouldRetry()`; streams now survive graceful server restarts. `shouldReconnect: boolean | ((error, attempt) => boolean)` controls the policy, with the default exported for composition.
- Liveness: protocol heartbeat on Node.js (any inbound message counts, so busy connections are never falsely killed) and an optional message-based idle timeout everywhere — the browser's dead-connection detector.
- Flow control: real socket backpressure on Node.js past `highWaterMark`; `maxBufferedBytes` as a hard cap that fails the connection rather than growing memory unbounded (useful for browser environments that don't have reader-side backpressure).
- Node.js-only request `headers` (`Record | Headers`); the browser throws at construction rather than silently dropping what is usually auth. Browser-targeted code can opt into compile-time checking via the exported `BrowserWebSocketClientOptions` / `BrowserWebSocketConnectionOptions` types, which omit `headers`.

`WebSocketConnection` is the underlying single-connection primitive with the same lifecycle and options (hooks: `onOpen`/`onError`/`onClose`) — it never reconnects, for consumers that handle failures themselves. Platform differences surface as data via `capabilities` rather than API shape.

permessage-deflate remains offered by default on both platforms (identical negotiation headers, no configuration surface).

Consumers migrated: ozone `Jetstream`, xrpc-server `Subscription`, and tap `TapChannel`. Breaking changes for those packages: `Subscription` and `TapWebsocketOptions` no longer accept arbitrary `ws` `ClientOptions` (explicit option sets now, including `headers`), and `DisconnectError` moves from `@atproto/ws-client` to `@atproto/xrpc-server`. `CloseCode` expands to the full WebSocket code list.

Here's an example:
```ts
import { WebSocketClient } from '@atproto/ws-client'

const ws = new WebSocketClient('wss://jetstream.example.com/subscribe', {
  dataMode: 'text',
  onReconnect: () => console.log('reconnected'),
})

for await (const message of ws) {
  const event = JSON.parse(message)
  // ... handle event, spanning any number of reconnects
}
```


packages/ws-client/tsconfig.build.json
```diff
@@ -5,5 +5,6 @@
   "compilerOptions": {
     "rootDir": "./src",
     "outDir": "./dist",
+    "lib": ["ES2023", "DOM", "DOM.Iterable"],
```
matthieusieben: I don't like this pattern (which, sadly, is the way things are often done in TS) because it can give a false sense of type-safety: NodeJS code can use Browser only APIs, Browser code can use NodeJS only APIs, and TS won't complain in either cases.


In order to avoid this, we could do the following:

- this file extends `base.tsconfig.json`
- this file excludes `./src/transport/**/*`
- `tsconfig.build-node.json` extends `node.tsconfig.json` and includes `./src/transport/**/node-*`
- Same as ^ for browser

packages/ws-client/package.json
```diff
@@ -21,25 +21,31 @@
...
+      "node": {
+        "types": "./dist/node.d.ts",
+        "default": "./dist/node.js"
+      },
+      "default": {
+        "types": "./dist/browser.d.ts",
+        "default": "./dist/browser.js"
+      }
     }
   },
```
matthieusieben, line 33: I believe that a more conventional way of achieving this is to use a single `exports` (with "types" and "default"), and use `imports` https://nodejs.org/api/packages.html#imports for nodejs specific imports (e.g. override `import/export Foo from "./transport.js"` with `import/export Foo "./transport-node.js"`)

matthieusieben (commented):
I don't really like that there are two `"exports"` fields. This can lead to confusing setup: a NodeJS consumer has to configure the [customConditions](https://www.typescriptlang.org/tsconfig/#customConditions) to `node` for it to properly work. This is not ideal as it prevents this packages from appearing are truly isomorphic. It also results in a lot of duplication in the entrypoints.

Instead, I recommend defining a single entrypoint, and use `imports` conditions so that the "routing" happens 100% internally (see comment)

packages/ws-client/src/client.ts
```diff
@@ -0,0 +1,386 @@
...
+  }
+
+  get connected(): boolean {
+    return this.connection?.readyState === 'open'
+  }
+
+  send(data: MessageOf<M>): Promise<void> {
+    // With no connection at hand (never iterated, or stopped) the client
+    // speaks for itself; otherwise the connection does — it rejects with a
+    // WebSocketConnectionError if it can't send (e.g. mid-reconnect).
```
matthieusieben: This should be JSDoc

packages/ws-client/src/client.ts
```diff
@@ -0,0 +1,386 @@
...
+  // every started lifecycle end with exactly one 'close' — carrying the real
+  // close code when a close frame provided one, or the WHATWG-conventional
+  // 1005 (no status) when none did (e.g. a stop mid-backoff).
+  private lastCloseDetail?: CloseEventDetail
+  private closeDispatched = false
+  // Internal wake signal for close(): lets a parked backoff `sleep` resolve
+  // (not reject) promptly instead of waiting out the full backoff window.
+  // Distinct from `options.signal`, which is the caller's abort — that path
+  // must still reject the iterator with the abort reason.
+  private readonly stopController = new AbortController()
```
matthieusieben: Is there any reason why not use actual private fields (`#prefixed`) ?

packages/ws-client/package.json
devinivy: 👍 Done in 3d03062bc687c245bc72886efe63ceed322c2fe1 — single `exports` entrypoint, with `imports` conditioning `#transport` to the node/browser transport internally.

packages/ws-client/tsconfig.build.json
devinivy: 👍 Done in 3d03062bc687c245bc72886efe63ceed322c2fe1.

Two small differences: the shared code extends `isomorphic.tsconfig.json` rather than `base.tsconfig.json` (base has empty `lib`/`types`, so it cannot compile the engines alone), and `src/transport/transport.ts` stays in the shared project. The shared engines do still see DOM+node via the isomorphic preset.

packages/ws-client/src/client.ts
devinivy: 👍 Done in 97577f706e11b544e9ce90dae76147a684297b6a.

packages/ws-client/src/client.ts
devinivy: 👍 No reason — switched to native `#private` fields throughout in 97577f706e11b544e9ce90dae76147a684297b6a.

packages/ws-client/src/connection.ts
```diff
@@ -0,0 +1,492 @@
...
+    this.dispatchEvent(new CustomEvent('error', { detail: { error } }))
+    this.#dispatchClose(closeDetailForError(error))
+  }
+
+  #detachSignal(): void {
+    if (this.#signal && this.#onAbort) {
+      this.#signal.removeEventListener('abort', this.#onAbort)
+      this.#onAbort = undefined
+    }
+  }
```
matthieusieben: A pattern I've been using is to create a private signal, and bind it two ways with the external signal. This has several advantages:
- The signal is always present (no optional `?.` chaining) and can safely be used with internal dependencies (e.g. the WS client) without risking to clutter the external signal with leaking internals
- The signal can be used with tools called internally without having to potentially manage multiple signals (e.g. `sleep(delay, internalSignal, externalSignal)` is not an ideal pattern to use)
- The binding and cleanup happens in one place (no need to have a `#detachSignal`, with setup & cleanup code "far" apart from each other, no need for extraneous `this.#onAbort`)
- Bindings to other stop signals (e.g. `SIGINT`, or an external promise, or whatever) can be done in the same way and requires no additional cleanup logic (ie. no need to make #detachSignal more complex)
- The internal signal can be used as `destroyed` proxy (get destroyed() { return this.signal.aborted }`)
- The internal signal can be used as guard against "use after destroyed" (`this.signal.throwIfAborted()`)

something like this:

```ts
class Foo {
  this.#abortController = new AbortController()

  constructor ({ signal: externalSignal }: { signal?: AbortSignal }) {
  
    // two-way binding that always cleans itself up whenever internal controller is aborted:
    externalSignal?.addEventListener("abort", () => this.destroy(), { signal: this.signal })
  }
  
  get signal() {
    return this.#abortController.signal
  }
  
  get destroyed() {
    return this.signal.aborted
  }
  
  protected assertNotDestroyed() {
    return this.signal.throwIfAborted()
  }
  
  destroy() {
    // Prevent invalid use
    this.signal.throwIfAborted()
  
    // Always abort the internal controller when we are done, allowing cleaning up resources
    this.#abortController.abort()
  }
}
```

packages/ws-client/src/connection.ts
```diff
@@ -0,0 +1,492 @@
...
+  implements AsyncIterable<MessageOf<M>>
+{
+  readonly capabilities: TransportCapabilities
+
+  readonly #transport: Transport
+  readonly #dataMode: DataMode
+  readonly #signal?: AbortSignal
+  #onAbort?: () => void
+
+  #state: ReadyState = 'initialized'
```
matthieusieben, line 109: Would an enum be better suited ?

packages/tap/src/channel.ts
```diff
@@ -42,25 +65,41 @@ export class TapChannel implements AsyncDisposable {
...
+      heartbeat: rest.heartbeatIntervalMs
+        ? { intervalMs: rest.heartbeatIntervalMs }
+        : undefined,
+      signal: this.abortController.signal,
+      shouldReconnect,
     })
+    // Flush buffered acks on each reconnect (not the initial open).
+    this.ws.addEventListener('reconnect', () => {
+      this.flushBufferedAcks()
+    })
```
matthieusieben: It feels like events here could benefit from being able to support async operations (e.g. to wait for the flushing to complete before actually attempting to reconnect), to better support backpressure, esp. in case the "reconnect" task takes a long time, and the connection is unstable (connects -> drops fast -> reconnects -> drops fast -> etc.). Such cases could result in an unbound growth of concurrent tasks.

Also, `EventTarget` is not well supported across platforms, which is why we [removed it](https://github.com/bluesky-social/atproto/blob/7e6eb9d48138f60d1a9d511549e75d74623f8c28/packages/oauth/oauth-client/CHANGELOG.md#060) form OAuthClient.

I would recommend using "hooks" instead of extending `EventTarget`

packages/ws-client/src/lib/typed-event-target.ts
```diff
@@ -0,0 +1,51 @@
+/**
+ * EventTarget with a typed event map. The runtime behavior is the native
+ * EventTarget; the `declare` overrides only refine the types so
+ * `addEventListener('open', ev => ...)` infers the right event/detail.
+ */
+export class TypedEventTarget<
```
matthieusieben: I've struggled with this pattern in the past because it provides a fake sense of type safety. Consider the following:

```ts
declare const ws: WebSocketClientBase

function emitEvent(eventTarget: EventTarget) {
  eventTarget.dispatch(
    new CustomEvent('error', {
      detail: "not the valid shape",
    })
  )
}

emitEvent(ws)
```

There is no type cast, or unsafe use of `any`, yet this causes unsafe behaviors because there is no way to properly prevent emitting invalid event shapes.

Since the Client class is public, the user of the lib could trigger errors from withing our code without realizing, and in a way that is very hard to debug.

packages/ws-client/src/transport/browser-transport.ts
```diff
@@ -0,0 +1,150 @@
...
+  close(code?: number, reason?: string): void
+  addEventListener(type: 'open', listener: () => void): void
+  addEventListener(
+    type: 'message',
+    listener: (ev: { data: unknown }) => void,
+  ): void
+  addEventListener(
+    type: 'close',
+    listener: (ev: { code: number; reason: string; wasClean: boolean }) => void,
+  ): void
```
matthieusieben, line 53: Wouldn't these be better typed as:

```ts
interface WHATWGCloseEvent extends Event {
  code: number
  reason: string
  wasClean: boolean
}
```

```suggestion
  addEventListener(
    type: 'close',
    listener: (ev: WHATWGCloseEvent) => void,
  ): void
```

packages/ws-client/src/transport/browser-transport.ts
```diff
@@ -0,0 +1,150 @@
...
+    // `headers`, but a consumer type-checking against the Node.js entrypoint
+    // can still reach here at runtime.)
+    if (hasHeaders(options?.headers)) {
+      throw new TypeError(
+        'WebSocket headers are not supported in the browser; use URL or subprotocol-based auth instead',
+      )
+    }
+    this.#url = url
+    this.#options = options
+    this.#WebSocketImpl = WebSocketImpl
```
matthieusieben, line 92: I would add a guard here, just for good measure:

```ts
    if (!WebSocketImpl) {
      throw new TypeError('WebSocket are not availalbe in this environment')
    }
```

We only explicitly support nodejs and browser but this might be useful when used from Expo/Bun/Deno.

packages/ws-client/src/transport/browser-transport.ts
```diff
@@ -0,0 +1,150 @@
...
+        )
+      } else {
+        this.handlers.onError(
+          new Error('Unsupported WebSocket message data type'),
+        )
+      }
+    })
+    ws.addEventListener(
+      'close',
+      (ev: { code: number; reason: string; wasClean: boolean }) => {
```
matthieusieben: Explicit typing should not be needed here (it should be inferred from the `WHATWGWebSocket` interface.


```suggestion
      (ev) => {
```

packages/ws-client/src/transport/browser-transport.ts
```diff
@@ -0,0 +1,150 @@
...
+        this.handlers.onClose(ev.code, ev.reason, ev.wasClean)
+      },
+    )
+    ws.addEventListener('error', () => {
+      this.handlers.onError(new Error('WebSocket error'))
```
matthieusieben: Would it be useful to wire the `cause` here ?

packages/ws-client/src/transport/transport.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export interface Transport {
+  readonly capabilities: TransportCapabilities
+  /** Negotiated subprotocol; '' until open (and if none negotiated). */
+  readonly protocol: string
+  /** Set once by the engine before events begin. */
+  handlers: TransportHandlers
+  /** Instantiate and connect the underlying socket, wiring it to `handlers`. */
+  open(): void
+  send(data: string | Uint8Array, onFlush: (err?: Error) => void): void
```
matthieusieben: Is there any particular reason for using a callback style instead of a promise? If so, I would add a comment here.

packages/ws-client/src/transport/node-transport.ts
```diff
@@ -0,0 +1,105 @@
...
+        code,
+        reason.toString('utf8'),
+        code === CloseCode.Normal,
+      )
+    })
+    ws.on('error', (err: Error) => this.handlers.onError(err))
+  }
+
+  get protocol(): string {
+    return this.#ws?.protocol ?? ''
```
matthieusieben: I dislike using `""` as marker for "unknown". Is there a way we can avoid this (e.g. by making the return type optional, or by throwing "protocol can only be read when connection is open") ?

packages/ws-client/src/client.ts
```diff
@@ -0,0 +1,396 @@
...
+    if (!this.#connection) {
+      return Promise.reject(
+        new WebSocketClientError('WebSocketClient is not connected'),
+      )
+    }
+    return this.#connection.send(data)
+  }
+
+  async close(code: number = CloseCode.Normal, reason?: string): Promise<void> {
+    this.#stopped = true
```
matthieusieben: You can avoid the extra property by using a getter:

```ts
get closed() {
  return this.#stopController.signal.aborted
}
```

matthieusieben: Also, should this protect against multiple close?

packages/ws-client/src/client.ts
```diff
@@ -0,0 +1,396 @@
...
+          reason: '',
+          wasClean: false,
+        },
+      }),
+    )
+  }
+
+  #resolveUrl(): Awaitable<string | URL> {
+    return typeof this.#url === 'function' ? this.#url() : this.#url
+  }
```
matthieusieben: We get better stack traces when using explicit `async` (or maybe we used to and it no longer matters)

```suggestion
  async #resolveUrl(): Promise<string | URL> {
    return typeof this.#url === 'function' ? this.#url() : this.#url
  }
```

packages/ws-client/src/client.ts
```diff
@@ -0,0 +1,396 @@
...
+        if (!firstAttempt) {
+          // Exponential backoff: ~1s (with jitter) on the first reconnect after
+          // an open, escalating across consecutive failures.
+          // Rejects with signal.reason on external abort; resolves promptly
+          // (not rejects) on close()'s internal stop signal — the loop below
+          // observes `this.#stopped` and exits the clean way.
+          await sleep(
+            backoffMs(retries, maxMs),
+            signal,
+            this.#stopController.signal,
```
matthieusieben: I bi-directionnaly linking signals (see my other comment) also also allows to simplify logic like this one by using a signal instead of passing around multiple signals.

packages/ws-client/src/client.ts
```diff
@@ -0,0 +1,396 @@
...
+
+        const url = await this.#resolveUrl()
+        // resolveUrl() is an async gap: close() or signal abort landing while
+        // it's in flight must not fall through to creating (and leaking) a
+        // connection that nothing will ever terminate. Re-check both before
+        // createConnection — an abort here would otherwise be missed entirely,
+        // since an already-aborted signal never fires a listener added after
+        // the fact.
+        if (this.#stopped) break
+        if (signal?.aborted) throw signal.reason
```
matthieusieben:
```suggestion
        signal?.throwIfAborted()
```

(also see my recomendation about using a signle signal and avoid using an extra `#stopped` property)

packages/ws-client/src/client.ts
```diff
@@ -0,0 +1,396 @@
...
+      reject(signal!.reason)
+    }
+    const onStop = () => {
+      cleanup()
+      resolve()
+    }
+    signal?.addEventListener('abort', onAbort, { once: true })
+    stopSignal?.addEventListener('abort', onStop, { once: true })
+    // Node: don't keep the process alive on the backoff timer.
+    ;(timer as { unref?: () => void }).unref?.()
```
matthieusieben: Is this right ? If a node process is only reading from a WS, and there is a reconnect happening, this `unref` will cause the node process to exit instead of reconnecting.

packages/ws-client/src/connection.ts
```diff
@@ -0,0 +1,492 @@
...
+    return this.#state === 'open'
+  }
+
+  get protocol(): string {
+    return this.#negotiatedProtocol
+  }
+
+  // ---- transport handlers ----
+
+  #buildHandlers(): TransportHandlers {
```
matthieusieben, line 184: It is too bad we have to keep this function around for the entire lifetime of the object; when it is only used during the init...

packages/ws-client/src/connection.ts
matthieusieben: See other comments (e.g. in the transport classes) where the use of this pattern would also be useful.

packages/ws-client/package.json
```diff
@@ -25,21 +25,33 @@
...
+  "imports": {
+    "#transport": {
+      "node": {
+        "types": "./dist/transport/node-transport.d.ts",
+        "default": "./dist/transport/node-transport.js"
+      },
+      "default": {
+        "types": "./dist/transport/browser-transport.d.ts",
+        "default": "./dist/transport/browser-transport.js"
+      }
```
matthieusieben: I'm a little surprised we need `types` here?

packages/ws-client/src/lib/reconnect-policy.ts
```diff
@@ -0,0 +1,55 @@
...
+export function defaultShouldReconnect(error: unknown): boolean {
+  if (error instanceof CloseError) {
+    return isReconnectableClose(error.code)
+  }
+  return (
+    error instanceof SocketError ||
+    error instanceof HeartbeatTimeoutError ||
+    error instanceof IdleTimeoutError
+  )
+}
```
matthieusieben, line 13: Would it make sense to move this logic as abstract property of the `WebSocketConnectionError` class ?

```ts
export abstract class WebSocketConnectionError extends Error {
  abstract shouldRetry(): boolean

  // ...
}

export class CloseError extends WebSocketConnectionError {
  shouldRetry() {
    return this.code === CloseCode.Abnormal
  }

  // ...
}

export class SocketError extends WebSocketConnectionError {
  shouldRetry () {
    return true
   }

  // ...
}
```

I've done something like this in lex-sdk errors:

https://github.com/bluesky-social/atproto/blob/e133a65cba992e6429d4526dc3e7cfe8bf87426f/packages/lex/lex-client/src/errors.ts#L149

This would simplify usage implementation to:

```ts
if (err instanceof WebSocketConnectionError && err.shouldRetry()) {
  // 
}
```

packages/ws-client/src/connection.ts
devinivy: I am pretty ambivalent about this— do you feel it makes a practical difference anywhere?  I would usually tie-break by going with whatever is best established in the codebase, but I am not sure either pattern is more common than the other.  I do kind of like sticking with erasable syntax, though also not an explicit goal.

Side note— I see that `ReadyState` is not exported and I believe it should be in the public interface.

packages/tap/src/channel.ts
devinivy: I didn't realize it wasn't well supported across environments!  I reached for it because I assumed it was solid common ground.  Which environments don't have it?

I can't think of a use case for async work in reconnect—I anticipate the main use case to be application logging/metrics.  But at the same time I am not opposed to switching to hooks.

packages/ws-client/src/transport/browser-transport.ts
devinivy: Yes, I think we could wire up `ev.error`.

packages/ws-client/src/transport/node-transport.ts
devinivy: This behavior is inherited from WHATWG `WebSocket` and the `ws` module, which both use `''` in this way.  We could interpret `''` into `'unknown'` or `null`.  I think we'd need to go with `null` since `'unknown'` is a valid subprotocol name (whereas `''` is not, for example).  I would rather not throw when accessing this.

packages/ws-client/src/client.ts
devinivy: Both WHATWG `WebSocket` and `ws` are idempotent on closes, and I can't think of any significant reason to deviate from their behavior.

packages/ws-client/src/client.ts
devinivy: Yeah, I'd prefer to switch to this too.

packages/ws-client/src/client.ts
devinivy: Good call, will get a failing test in there for this behavior.

packages/ws-client/src/connection.ts
devinivy: What is the issue?

packages/ws-client/package.json
devinivy: We very well may not, but as far as I can tell by convention we usually include both in our exports.

packages/ws-client/src/lib/reconnect-policy.ts
devinivy: Sure, I like that and it parallels lex.

packages/ws-client/src/transport/transport.ts
devinivy: I was in the mindset of making transport as thin as possible over `WebSocket` and `ws`, but I don't think we really need to stick to that here.

packages/ws-client/src/connection.ts
devinivy: Yeah, I like that too for the reasons you describe, and it is more my style anyway—not sure why I did not do it that way to begin with.

packages/ws-client/src/connection.ts
matthieusieben: I don't have a strong opinion here. The main reason I like them is that they generally offer better dx. Consider, for example, refactoring a value. Trivial with enum, much less so with string literals.

packages/ws-client/src/connection.ts
matthieusieben: Nevermind, I thought that `#fn(){}` methods where created once per instance (similar to `fn: () => {}`) but that is not the case.

packages/ws-client/package.json
matthieusieben: But imports and exports are quite different things, and I don't _think_ that `imports` types have any use.

packages/tap/src/channel.ts
```diff
@@ -15,10 +37,11 @@ export interface TapHandler {
   onError: (err: Error) => void
 }
 
-export type TapWebsocketOptions = ClientOptions & {
+export type TapWebsocketOptions = {
   adminPassword?: string
   maxReconnectSeconds?: number
   heartbeatIntervalMs?: number
+  headers?: Record<string, string> | Headers
```
matthieusieben: Can you use `HeadersInit` here ? If not a comment might be welcome.

packages/ws-client/src/client.ts
```diff
@@ -0,0 +1,400 @@
...
+    createConnection: ConnectionFactory,
+    url: string | URL | (() => Awaitable<string | URL>),
+    options: WebSocketClientOptions<M> = {},
+  ) {
+    this.#createConnection = createConnection
+    this.#url = url
+    this.#options = options
+    const { signal } = options
+    if (signal?.aborted) {
+      this.#stopController.abort(signal.reason)
```
matthieusieben: It feels odd to support creating  a destroyed instance. I like doing:

```ts
signal?.throwIfAborted()
signal?.addEventListener(...)
```

packages/ws-client/src/client.ts
```diff
@@ -0,0 +1,400 @@
...
+   * otherwise the connection answers for itself — rejecting with a
+   * {@link WebSocketConnectionError} if it can't send (e.g. mid-reconnect).
+   */
+  send(data: MessageOf<M>): Promise<void> {
+    if (!this.#connection) {
+      return Promise.reject(
+        new WebSocketClientError('WebSocketClient is not connected'),
+      )
+    }
+    return this.#connection.send(data)
```
matthieusieben, line 212: If this fails synchronously (for any reason), the return value won't be a rejected promise but a thrown error.

There are two ways to fix this:
- make the method `async` (adds a "tick")
- `try/catch` with `return Promise.reject(err)` (more verbose)

packages/ws-client/src/client.ts
```diff
@@ -0,0 +1,400 @@
...
+      await this.#connection.close(code, reason).catch(() => {})
+    } else {
+      // No live connection to hand-shake with (never opened, or stopped
+      // mid-backoff after a failure): no status code from the server applies
+      // to this stop, so drop any stale detail — the final onClose hook
+      // synthesizes 1005 (no status) per the WHATWG convention.
+      this.#lastCloseDetail = undefined
+    }
+    this.#state = 'closed'
+  }
```
matthieusieben:
```ts
await client.close()
client.close()
```

Might cause the `#state` to (briefly) go from `closed` to `closing`, which feels odd. It would also clear the `#lastCloseDetail`, which might cause `onClose` to be called with the wrong value.

It might be worth being explicit about the fact that we don't defend against this (or actually prevent any misuse).

packages/ws-client/src/client.ts
```diff
@@ -0,0 +1,400 @@
...
+    if (this.#iterated) {
+      throw new WebSocketClientError(
+        'WebSocketClient is already being iterated',
+      )
+    }
+    this.#iterated = true
+    return this.#iterate()
+  }
+
+  async *#iterate(): AsyncGenerator<MessageOf<M>> {
```
matthieusieben: `AsyncGenerator` has unsafe defaults (`any`). This makes it safer:

```suggestion
  async *#iterate(): AsyncGenerator<MessageOf<M>, void, unknown> {
```

matthieusieben: (we should have a lint rule for this)

packages/ws-client/src/connection.ts
```diff
@@ -0,0 +1,516 @@
...
+
+// The two terminal outcomes. `done` drains the buffer first; `error` discards it.
+type Terminal = { type: 'done' } | { type: 'error'; error: unknown }
+
+const CLEAN_CLOSE_CODES = new Set([CloseCode.Normal, CloseCode.GoingAway])
+
+export class WebSocketConnectionEngine<M extends DataMode = 'auto'>
+  implements AsyncIterable<MessageOf<M>>
+{
+  readonly capabilities: TransportCapabilities
```
matthieusieben: Would this better be exposed as a getter ?

packages/ws-client/src/connection.ts
```diff
@@ -0,0 +1,516 @@
...
+    const hb = options.heartbeat
+    if (hb !== false && this.#transport.capabilities.heartbeat) {
+      this.#heartbeatIntervalMs = hb?.intervalMs ?? 10_000
+    }
+    this.#idleTimeoutMs = options.idleTimeoutMs ?? null
+
+    const { signal } = options
+    if (signal) {
+      if (signal.aborted) {
+        // Defer so the caller can attach consumers first.
```
matthieusieben: This is why I prefer failing:

```ts
signal?.throwIfAborted()
signal?.addEventListener(...)
```

packages/ws-client/tests/browser-transport.test.ts
```diff
@@ -0,0 +1,123 @@
...
+      /* drain */
+    }
+  })
+
+  it('yields text and ends on clean close', async () => {
+    const { url, terminate } = await startServer((ws) => {
+      ws.send('hi')
+      ws.close(1000)
+    })
+    await using _ = { [Symbol.asyncDispose]: async () => terminate() }
```
matthieusieben: https://github.com/bluesky-social/atproto/blob/main/packages/pds/tests/_util.ts#L150-L173
https://github.com/bluesky-social/atproto/blob/main/packages/ozone/tests/_util.ts#L219-L242


We already have `startServer` functions in `pds`/`ozone`/`bsky` that return an `AsyncDisposable`. I recommend doing the same here:

```ts
await using server = await startServer((ws) => { ... })

server.url
```

packages/ws-client/src/transport/transport.ts
```diff
@@ -0,0 +1,41 @@
+export interface TransportCapabilities {
+  /** Can send protocol pings and observe pongs. */
+  heartbeat: boolean
+  /** Can exert real read-side backpressure. */
+  pauseResume: boolean
+}
+
+export interface TransportHandlers {
+  onOpen(): void
+  onMessage(data: string | Uint8Array, isBinary: boolean): void
```
matthieusieben: `isBinary` can be inferred from the `data`. It might be useful to explain why we use a redundant argument.

```suggestion
  onMessage(data: string | Uint8Array): void
```

packages/ws-client/src/transport/browser-transport.ts
```diff
@@ -0,0 +1,168 @@
...
+        this.handlers.onMessage(new Uint8Array(data), true)
+      } else if (ArrayBuffer.isView(data)) {
+        // Unreachable for a spec-compliant WebSocket: with binaryType set to
+        // 'arraybuffer', received binary data is an ArrayBuffer, never a view.
+        // Kept as a defensive fallback for non-conforming implementations.
+        const view = data as ArrayBufferView
+        this.handlers.onMessage(
+          new Uint8Array(view.buffer, view.byteOffset, view.byteLength),
+          true,
+        )
```
matthieusieben: Can this really happen ?

```suggestion
```

packages/ws-client/src/connection.ts
```diff
@@ -0,0 +1,516 @@
...
+
+  // ---- transport handlers ----
+
+  #buildHandlers(): TransportHandlers {
+    return {
+      onOpen: () => {
+        if (this.#terminal || this.#state !== 'connecting') return
+        this.#state = 'open'
+        this.#negotiatedProtocol = this.#transport.protocol
+        this.#onOpenHook?.()
```
matthieusieben: When external hook functions are passed like this, their `this` argument is typically not typed. For this reason, I generally like to avoid "leaking" the `this` value when calling an external function:

```suggestion
        this.#onOpenHook?.call(null)
```

Alternatively, the `this` arg could be specified on the options, making it explicit that the user can access it:

```ts
export interface WebSocketConnectionOptions<M extends DataMode = 'auto'> {
  onOpen?: (this: WebSocketConnectionEngine) => void
  onError?: (this: WebSocketConnectionEngine, error: unknown) => void
  onClose?: (this: WebSocketConnectionEngine, detail: CloseEventDetail) => void
}
```

```suggestion
        this.#onOpenHook?.call(this)
```

packages/ws-client/src/connection.ts
```diff
@@ -0,0 +1,516 @@
...
+  idleTimeoutMs?: number
+  highWaterMark?: number
+  maxBufferedBytes?: number
+  signal?: AbortSignal
+  /** Called once when the socket opens. */
+  onOpen?: () => void
+  /** Called once when the connection ends with an error, before onClose. */
+  onError?: (error: unknown) => void
+  /** Called once when the connection ends, whichever way. */
+  onClose?: (detail: CloseEventDetail) => void
```
matthieusieben: If any of these hook throws an error, the connection will endup in a broken state. We should either protect against that or state the these *must not* throw.

packages/ws-client/src/client.ts
```diff
@@ -0,0 +1,400 @@
...
+        // code/reason/wasClean.
+        let closeDetail: CloseEventDetail | undefined
+        const connection = this.#createConnection<M>(url, {
+          ...this.#connectionOptions(),
+          onOpen: () => {
+            this.#state = 'open'
+            retries = 0 // stable open: backoff starts over
+            if (firstOpen) {
+              firstOpen = false
+              this.#options.onOpen?.()
```
matthieusieben: `this` in the `onOpen` function would be the private `#options`


```ts
new WebSocketClient('wss://a', {
  onOpen() {
    console.log(this) // the private #option object from the client
  }
}) 
```

```suggestion
              this.#options.onOpen?.call(null)
```

or

```suggestion
              this.#options.onOpen?.call(this)
```

packages/ws-client/src/connection.ts
```diff
@@ -0,0 +1,516 @@
...
+    this.#onCloseHook = options.onClose
+    this.#highWaterMark = options.highWaterMark ?? 1_048_576
+    this.#maxBufferedBytes = options.maxBufferedBytes ?? Infinity
+
+    this.#transport = createTransport(url, {
+      protocols: options.protocols,
+      headers: options.headers,
+    })
+    this.capabilities = this.#transport.capabilities
+    this.#transport.handlers = this.#buildHandlers()
```
matthieusieben: Why is this set on the instance instead of being added through the `createTransport` options ?

packages/ws-client/src/connection.ts
devinivy: I feel like the refactoring a value is more or less the same: you change the value, and then correct the type errors that result.  I think I'll leave this as-is for now.

packages/ws-client/package.json
devinivy: Ah, maybe!  Let me take a closer look.  Didn't realize this was `exports` from the diff here.

matthieusieben (approved):
looking good. A few more nits

packages/ws-client/src/transport/transport.ts
devinivy: It is an artifact of `ws` but I think we can get rid of it.

packages/ws-client/src/transport/node-transport.ts
```diff
@@ -0,0 +1,109 @@
...
+function toHeaderRecord(
+  headers?: HeadersInit,
+): Record<string, string> | undefined {
+  if (!headers) return undefined
+  // Headers normalizes every HeadersInit form (record, entry pairs, Headers).
+  const record: Record<string, string> = {}
+  new Headers(headers).forEach((value, key) => {
+    record[key] = value
+  })
+  return record
```
matthieusieben: This can be simplified quite a bit

```suggestion
  return Object.fromEntries(new Headers(headers))
```

packages/tap/src/channel.ts
```diff
@@ -1,10 +1,33 @@
...
+// the wire — ends the session, matching prior behavior exactly.
+function shouldReconnect(error: unknown): boolean {
+  if (error instanceof CloseError) {
+    return error.code === CloseCode.Abnormal
+  }
+  return (
+    error instanceof SocketError ||
+    error instanceof HeartbeatTimeoutError ||
+    error instanceof IdleTimeoutError
+  )
```
matthieusieben: Since these define their own rule for retrying (which may evolve over time), I recommend relying on `shouldRetry` to avoid returning true here when `shouldRetry` would return false.

```suggestion
  if (
    error instanceof SocketError ||
    error instanceof HeartbeatTimeoutError ||
    error instanceof IdleTimeoutError
  ) {
    return error.shouldRetry()
  }
  
  return false
```

packages/ws-client/src/transport/transport.ts
```diff
@@ -0,0 +1,46 @@
...
+  onClose(code: number, reason: string, wasClean: boolean): void
+  onError(err: Error): void
+}
+
+export interface Transport {
+  readonly capabilities: TransportCapabilities
+  /** Negotiated subprotocol; `null` until open (and if none negotiated). */
+  readonly protocol: string | null
+  /** Set once by the engine before events begin. */
+  handlers: TransportHandlers
```
matthieusieben: This should be `readonly` and provided by the constructor.

```suggestion
  readonly handlers: TransportHandlers
```

packages/ws-client/src/connection.ts
```diff
@@ -0,0 +1,518 @@
...
+  }
+
+  // ---- public methods ----
+
+  async send(data: MessageOf<M>): Promise<void> {
+    if (this.#state !== 'open') {
+      throw new WebSocketConnectionError('WebSocketConnection is not open')
+    }
+    return this.#transport.send(data)
+  }
```
matthieusieben, line 392: Am i right to assume that `send()` is not reliable here ? e.g. if the network drops, and `send()` causes an error, the reconnect behavior will kick in, but the sent item will never have reached it's destination, and the caller a no (easy) way to determine that the item was delivered (TCP ack); or even to have an idea of the list of sent items that might have been missed ?

With the bare `WebSocket` implementation, this is left as responsibility to the user of the `WebSocket` object (ie. using sime application level acks), but since this lib wraps those, what is our story around failed send ?

matthieusieben: One mitigation would be to have a `onPingPong` hook, that would allow the user of this lib to know that a full ping/ping cycle occurred after their message was `send()`, allowing them to assume that the received got the message. But this would not work in browsers...

devinivy: > With the bare WebSocket implementation, this is left as responsibility to the user of the WebSocket object (ie. using sime application level acks), but since this lib wraps those, what is our story around failed send ?

It would be the same story as with the bare WebSocket implementation.  This is just a property of WebSockets, it's typically dealt with in the application.

It is an interesting idea to proactively ping/pong, but even then you adopt the complexity while also opening more new questions.  For example, if you don't get a pong back it doesn't mean the message _wasn't_ delivered, so throwing an error in that case doesn't necessarily bring the app the clarity that it may expect/assume.  And then as you mention, it's not compatible with the browser.  Lastly, we don't have specific needs around this for atproto usage, so I think there's a strong case not to try to do anything too clever here.

> Am i right to assume that send() is not reliable here ? e.g. if the network drops, and send() causes an error, the reconnect behavior will kick in

Just a note that only `WebSocketClient` has the reconnect behavior, whereas this class `WebSocketConnection` deals with the lifecycle of a single connection.

packages/ws-client/src/connection.ts
devinivy: Just a quirk of various changes that happened during the course of implementation.  It's pretty weird and I don't think there's any particular use for it, I will clean that up.

packages/ws-client/src/transport/browser-transport.ts
devinivy: Yeah, overly paranoid— will nix.

packages/ws-client/src/connection.ts
devinivy: Idea being that we dynamically defer to `this.#transport`?  I don't think it will make any practical difference but I suppose it's kind of a nice way to pass through `#transport`'s behavior.  If we make this change we'll want to do the same with `this.#transport.protocol`.

packages/ws-client/src/client.ts
```diff
@@ -0,0 +1,555 @@
...
+// external signal abort, or a fatal error recorded by the loop) rejects the
+// iterator with that reason.
+const STOPPED_CLEANLY = Symbol('WebSocketClient stopped cleanly')
+
+export class WebSocketClientBase<M extends DataMode = 'auto'>
+  implements AsyncIterable<MessageOf<M>>
+{
+  #connection: WebSocketConnectionEngine<M> | null = null
+  #state: ReadyState = 'initialized'
+  #iterated = false
```
matthieusieben, line 109: Nit: can be removed and replaced with `this.#iterator != null`

packages/ws-client/src/client.ts
```diff
@@ -0,0 +1,555 @@
...
+    } finally {
+      // The generator body is the lifecycle: entering it started this client,
+      // and this finally is its single terminal transition — whichever way
+      // the loop exits (fatal error, non-reconnectable clean close, close(),
+      // abort, or a consumer break/return). Record the stop cause if none
+      // was recorded (clean exits — server-sent fatal close, or consumer
+      // break — never abort the stop signal themselves), settle state, fire
+      // the one onClose, and release close() waiters. A never-iterated
+      // client never runs this: no lifecycle, no onClose.
+      if (!stopSignal.aborted) this.#stopController.abort(STOPPED_CLEANLY)
```
matthieusieben, line 530: Nit:

```suggestion
      this.#stopController.abort(STOPPED_CLEANLY)
```

packages/ws-client/src/connection.ts
```diff
@@ -0,0 +1,538 @@
...
+        'WebSocketConnection is already being iterated',
+      )
+    }
+    this.#iterated = true
+    return {
+      next: () => this.#next(),
+      return: async () => {
+        // Consumer abandoned iteration: polite close.
+        if (this.#state === 'open' || this.#state === 'connecting') {
+          void this.close(CloseCode.Normal).catch(() => {})
```
matthieusieben, line 473: Why don't we ?
1) await
2) propagate errors


```suggestion
          await this.close(CloseCode.Normal)
```

packages/ws-client/src/connection.ts
```diff
@@ -0,0 +1,538 @@
...
+    const { signal } = options
+    // Constructing an already-aborted connection is a programmer error: fail
+    // now rather than produce an instance that can never open. The listener
+    // below self-detaches when the connection settles its terminal
+    // (#doneController aborts in #fail/#finishDone).
+    signal?.throwIfAborted()
+    signal?.addEventListener('abort', () => this.#fail(signal.reason, true), {
+      once: true,
+      signal: this.#doneController.signal,
+    })
```
matthieusieben, line 162: From reading the code here, it seems that it might be possible to do:

```ts
const connection = new WebSocketConnectionEngine({
  // ...
  signal: ac.signal
})

ac.abort()

console.log(connection.readyState) // "initialized"

// will connect
for await (const _ of connection) {
  // works
}
```

matthieusieben (commented):
it seems to me that this PR has grown organically up to a state where we might want to take a step back and evaluate the level of complexity.

The 4 layers design (Client -> Connection -> Transport -> WebSocket), in particular, is pretty hard to reason about, especially when it comes to the various way one can terminate the iteration (signal, close(), break, throw). I would probably try to drop the `Connection` layer.

I wonder if an API that exposes a generator function instead of a class woudn't lead to a simpler implementation. The instance implementation really makes sense when we need to `send()` messages, which is not the case for us.

```ts
export async function* websocket(
  url: string | URL | (() => string | URL),
  transport: ...,
  options?: {
    // Used to abort while awaiting for a message
    signal,
  
    // Do we really need all the hooks, besides "shouldReconnect" ?
  }
) {
  //
}
```


This is especially true since `ws` already exports utilities that make a NodeJS only implementation trivially simple when compared to the one in this PR (see "working"(?) example below). If we reworked the "Transport" abstraction to be an AsyncIterable WebSocket abstraction (that internally handles ping/pongs), and adopted a schema such as ^ then it seems to me that the whole implementation would be much simpler.

```ts
// transport-interface.ts
interface Sender<TMode> {
  send(message: MessageOf<TMode>): Promise<void>
}

interface Transport<TMode> extends AsyncIterable<MessageOf<TMode>>, Sender<TMode> {
}

type TransportFactoryOptions<TMode> = {
  url: string,
  mode: TMode,
  signal?: AbortSignal,
  /** The TCP connection was established (send() is safe to use) */
  onOpen: () => void,
  /** The socket was open, then closed (send() was safe to use but not anymore) */
  onClose: (reason: unknown) => void,
}

type TransportFactory = <TMode>(options: TransportFactoryOptions<TMode>) => Transport<TMode>

// transport-factory.node.ts
import { createWebSocketStream, WebSocket } from 'ws'
import { TransportFactory } from './transport-interface.ts'

export const transportFactory: TransportFactory = ({ url, signal, onOpen, onClose }) => {
  const ws = new WebSocket(url)
  
  // @NOTE readable back pressure automatically propagated
  const stream = createWebSocketStream(ws, {
    readableObjectMode: true,
    signal
    // supports wartermark & co.
  })
  
  // @TODO wire onOpen/onClose hooks
  
  // @TODO handle ping/pong internally here:
  // stream.destroy(new Error('ping timeout'))

  return stream
}

// transport-factory.whatwg.ts
import { TransportFactory } from './transport-interface.ts'

export const transportFactory: TransportFactory = ({ url, signal }) => {
  const ws = new WebSocket(url)
  
  // @NOTE ping/pong not supported by whatwg WebSockets

  // @TODO wire onOpen/onClose hooks
  
  // @TODO support readableHightWaterMark (triggering a special, retryable, error) so
  // that the memory consumption of the client stays in check
  
  // ...
  // (Actual implementation could be provided by a dependency package.)
  // Seems like this is pretty simple to implement:
  // https://github.com/vadzim/websocket-iterator/blob/master/src/websocket-iterator.ts
  
  const iterator: Iterator<any> = {
    async next() {
      // TODO
    },
    async return() {
      // TODO
    },
    async throw(err) {
      await this.return()
      throw err
    },
  }
  
  return {
    send: async (message) => ws.send(message),
    [Symbol.asyncIterator]: () => iterator,
  }
}


// index.ts
import { transportFactory } from '#transport-factory'

export function websocket<TMode = 'auto'>(...args) {
  // @NOTE Because of the way async generators work in JS, 
  // we need to start the iteration so that the websocket
  // is open and ready for sending asap.
  return startedIterator(websocketLazy(...args))
}

export async function* websocketLazy<TMode = 'auto'>(
  url: string | URL | (() => string | URL),
  {
    signal,
    shouldRetry = err instanceof WebsocketError && err.shouldRetry(),
    onOpen,
    onClose,
    onEnded,
  }: {
    mode?: TMode,
    signal?: AbortSignal
    shouldRetry?: (err: unknown, attempt: number) => boolean
    /** A connection was established and messages can be sent using the sender */
    onOpen?: (sender: Sender<TMode>) => void
    /** The transport can no longer be used to send messages */
    onClose?: (reason: unknown) => void
    /** The websocket is terminally closed */
    onEnded?: () => void,
  } = {}
) {
  let attempt = 0
  try {
    for(;;) {
      signal?.throwIfAborted()
  
      const transport = transportFactory({
        url: String(typeof url === 'string' ? url : url()),
        signal,
        onOpen: () => onOpen?.(transport),
        onClose: (reason) => onClose?.(reason),
      })
  
      try {
        for await (const message of transport) {
          attempt = 0 // reset on success
          yield message
        }
        
        // @NOTE this should never be reached (every close event will
        // be turned into an error by the transport).
      } catch (err) {
        if (signal?.aborted) throw err
        if (!shouldRetry?.(err, attempt)) throw err
        attempt++
        await expBackoff(attempt, signal)
      }
    }
  } finally {
    onEnded?.()
  }
}
```

If we **do** need the `send()` ability, here is how we can built on top of ^.

```ts
class WebSocketClient {
  #destroyed = false
  #iterator = null

  #sender: null | Sender = null
  #senderQueue: Array<{ message: any; resolve: (thenable?: PromiseLike<void>) => void; reject: (err: Error) => void  }> = []
  #senderQueueLimit = 10
  
  get state() {
    if (this.#destroyed) return 'destroyed'
    if (!this.#iterator) return 'pending' // not "open" yet (neither send() nor [Symbol.asyncIterator]() were called)
    if (!this.#sender) return 'connecting'
    return 'conntected'
  }
  
  async send(message): Promise<void> {
    this.#open()
    if (this.#sender) return this.#sender.send(message)
    if (this.#senderQueue.length > this.#senderQueueLimit) throw new Error('full')
      this.#senderQueue.push({ resolve, reject, message })
    })
  }
  
  /**
   * Lazily open the connection when either send() or [Symbol.asyncIterator]()
   * is called.
   *
   *  @note this could also be called from the constructor based on some condition if that makes sense.
   */
  #open() {
    if (this.#destroyed) {
      throw new Error('destroyed')
    }

    if (!this.#iterator) {
      const iterable = websocket(this.url, {
        // ...
    
        onOpen: (sender) => {
          this.#sender = sender
          for (const { message, resolve } of this.#senderQueue.splice(0, Infinity)) {
            resolve(sender.send(message))
          }
        },
        onClose: () => {
          this.#sender = null
        },
        onEnded: () => {
          this.#destroyed = true
          this.#iterable = null
          for (const { reject } of this.#senderQueue.splice(0, Infinity)) {
            reject(new Error('destroyed'))
          }
        }
      })

      this.#iterator = startedIterator(iterable[Symbol.asyncIterator]())
    }
    
    return this.#iterator
  }

  [Symbol.asyncIterator]() {
    // @NOTE there is no need to protect against multiple calls to this
    return this.#open()
  }

  async [Symbol.asyncDispose]() {
    if (this.#destroyed) return
    this.#destroyed = true

    const iterator = this.#iterator
    this.#iterator = null

    await iterator?.return()
  }
}
```

I believe this architecture allows for simpler state management. It's biggest drawback is that we need to start iterating before we can start sending messages, but that can easily be worked around by starting the iterator (`next()`) and storing the promise for later.

```ts
function startedIterator(iterator) {
  let first = iterator.next()
  first.catch(() => { /* noop */ })
  
  return {
    next() {
      return getFirst() || iterator.next()
    },
    async return() {
      await getFirst()
      return iterator.return()
    },
    async throw(...args) {
      await getFirst()
      return iterator.throw(...args)
    }
  }
  
  function getFirst() {
    const p = first
    first = null
    return p
  }
}
```

matthieusieben:
It seems that chromuim based browser (and Deno) support https://developer.mozilla.org/en-US/docs/Web/API/WebSocketStream . It might be worth supporting that, if available.

packages/ws-client/src/client.ts
devinivy: Weird, I remember doing just that.  I must have lost it!

packages/ws-client/src/connection.ts
devinivy: I think we the await is good, but I am not sure about throwing here.  This could happen on a `break`, right?  I don't think a stream `cancel()` (whatwg) or `destroy()` (node) would analogously result in an error, you know what I mean?

packages/ws-client/src/connection.ts
devinivy: Urgh.  Yeah this supports your general point about simplifying the lifecycle and layering.

devinivy:
@matthieusieben I think you have a good point: this work organically grew into a bit of a beast, and I haven't done an amazing job managing that.  I am going to investigate your proposal to remove a layer, remove the `close()` terminal, and generally lean more on iteration+abort semantics to drive the lifecycle.

I may not pare down the hooks so much—I strongly associate `open` and `close` as occurring at most once in the lifecycle on an object, and I am not sure I love the story of leaning so hard on exposing the lifecycle of the underlying resource.  I'd like it to be simple to instrument reconnects and errors for applications that rely on a persistent websocket.

We do have a `send()` use case with tap to preserve, so I will keep that in the design.  I assessed `WebSocketStream` as part of this, and I determined that it's too unstable to be worth the additional code branches that we can always add later.  Deno flags it as unstable, it's experimental in the web platform, and in undici the implementation doesn't provide real backpressure yet—I think implementations are still shaking out.

matthieusieben:
The only reason I've been exposing onOpen and onClose (whatever we name them) is for the caller to know when "send()" can be used. The latest version of my comment (I've edited it) no longer needs that though.

devinivy:
Closing in favor of #5285

PR #5246, opened 2026-07-17
Refactor pds email tests to use jest spies

ThisIsMissEm:
Whilst working on #5207 I didn't realise we were moving to jest spies for the email testing, so I'd done a bunch of work in a different direction as I would've ended up creating more mess with the previous mail catching setup.

This PR completely removes that from our pds tests. I also took this chance to correctly type the accounts from seed client as we normally do.

I've kept a top-level `sendMail` transporter spy/mock to ensure mails are never sent from the tests, but moved to `using jest.spy` inline for the mailer methods. We **can't** use what @matthieusieben proposed of:
```
using sendMail = jest.spyOn(mailer.transporter, 'sendMail')
  .mockImplementation(async (...args) => {
    expect(args).toBeDefined() // ensures the data is passed through
  })
```
Since the `expect` throws which inside `mockImplementation` results in the error bubbling up not to the test framework, but to XRPCError which translates it to a Internal Server Error, effectively eating your test assertion failure. So instead we expect things on the last call of the `sendMailMock` if needed (we need to do this when testing that the right email template was used with the right subject line).

For tests that expected previous test cases state to be already setup, I've started doing an `expect(token).toBeDefined()` such that if the test that sets up that depended on state fails, then the subsequent tests fail fast.


.changeset/old-dingos-give.md
ThisIsMissEm, line 1: This empty changeset requirement feels a bit odd. Do y'all know about the changesets app that automatically comments on PRs to remind you to add a changeset?

matthieusieben (approved):

PR #5247, opened 2026-07-17
Fix S3/R2 socket leak on aborted getBlob downloads (pin @smithy/core >=3.29.5)

jcalabro:
## What

Floor `@smithy/core` to `>=3.29.5` (via a `pnpm-workspace.yaml` override) to fix an outbound S3/R2 socket leak in `S3BlobStore`.

## Why

On PDS blob hosts we observed containers climbing to 100% system CPU with load >100 over ~2 days. The root cause is a socket/fd leak on the blob **download** path:

- When a `com.atproto.sync.getBlob` client aborts a download mid-stream (`ERR_STREAM_PREMATURE_CLOSE` / `Error: aborted`), the underlying outbound `TLSSocket` to S3/R2 is never destroyed and its fd is never released.
- A heap snapshot from an affected worker showed ~720 leaked `TLSSocket` + `Agent` + `NodeHttpHandler` + `ChecksumStream` objects (retained via bound listeners / `IncomingMessage.socket`), ~210 MB of native `JSArrayBufferData`, with a flat JS heap — a classic native/socket leak, one stranded socket per aborted `getBlob`.
- ~720 leaked sockets/worker × 16 workers drove `tcp_memory_allocated` to the `net.ipv4.tcp_mem` ceiling, putting the whole box into permanent TCP memory-pressure mode → stalled teardowns → 100% system CPU.

The defect is in `@smithy/core`'s `ChecksumStream`: pre-3.29.5 it listened for `data`/`end`/`error` on the source socket but **not `close`**, so a mid-stream abort (which arrives as a bare `close` with no `end`/`error`) was swallowed — the source was never destroyed and its listeners never removed. `@smithy/core` 3.29.5 ([smithy-typescript#2152](https://github.com/smithy-lang/smithy-typescript/pull/2152), for [aws-sdk-js-v3#8098](https://github.com/aws/aws-sdk-js-v3/issues/8098)) adds the `close` listener that destroys the source and removes its listeners.

The repo's lockfile resolved `@smithy/core@3.26.0`. This became active after the SDK upgrade in #5233 swapped `requestTimeout` (a total-request timer that had incidentally been reaping abandoned download sockets) for `socketTimeout` (idle-only, which never fires on a socket pinned by a retained stream reference).

## The fix

No published `@aws-sdk/*` requires `@smithy/core >=3.29.5` yet (latest `@aws-sdk/core` pins `^3.29.4`), so `pnpm update` alone cannot reach the fix — a declared floor is required. Added a `pnpm-workspace.yaml` override pinning `@smithy/core` to `^3.29.5`, plus the `minimumReleaseAgeExclude` entry pnpm needs to accept it. Remove the override once `@aws-sdk/core` requires `>=3.29.5` transitively.

## Reproducer

Standalone repro that measures leaked sockets from the origin side via `server.getConnections()` (OS-independent; works on macOS). It leaks 40/40 aborted downloads on `@smithy/core@3.26.0` and 0/40 on `3.29.5`, and shows that no consumer-side workaround (destroying the `Body`, passing an `AbortController`) helps on 3.26.0 — only the version bump fixes it.

<details>
<summary><code>package.json</code></summary>

```json
{
  "name": "s3leak-repro",
  "private": true,
  "type": "module",
  "dependencies": {
    "@aws-crypto/crc32": "^5.2.0",
    "@aws-sdk/client-s3": "3.1073.0"
  }
}
```
</details>

<details>
<summary><code>repro.mjs</code></summary>

```js
// Reproduces the PDS getBlob socket leak: an S3 GetObject download whose
// downstream consumer aborts mid-stream. Pre-@smithy/core 3.29.5, ChecksumStream
// did not destroy its source / remove listeners, so the outbound socket leaks.
// Measured from the ORIGIN side via server.getConnections() (works on macOS).
//
// Fix matrix (env toggles), to show consumer-side workarounds do NOT help on 3.26.0:
//   FIX_DESTROY=1 -> destroy the SDK Body stream on consumer abort
//   FIX_ABORT=1   -> pass an AbortController to getObject and abort it
// Dependency fix = whichever @smithy/core is installed (3.26.0 leaky vs 3.29.5 fixed).

import http from 'node:http'
import { PassThrough, pipeline } from 'node:stream'
import { once } from 'node:events'
import { createRequire } from 'node:module'
import { Crc32 } from '@aws-crypto/crc32'
import { S3 } from '@aws-sdk/client-s3'

const require_ = createRequire(import.meta.url)
let coreVersion = 'unknown'
for (const p of ['@smithy/core/package.json', '@smithy/node-http-handler/node_modules/@smithy/core/package.json']) {
  try { coreVersion = require_(p).version; break } catch {}
}

const FIX_DESTROY = process.env.FIX_DESTROY === '1'
const FIX_ABORT = process.env.FIX_ABORT === '1'
const ITERS = Number(process.env.ITERS || 40)
const BLOB_SIZE = 1024 * 1024 // 1MB

const BODY = Buffer.alloc(BLOB_SIZE, 0x61)
const crc = new Crc32(); crc.update(BODY)
const crcBuf = Buffer.alloc(4); crcBuf.writeUInt32BE(crc.digest() >>> 0)
const crcHeader = crcBuf.toString('base64')

// mock R2/S3 origin: streams a large body slowly, emits a valid crc32 so the
// SDK may wrap the response in a ChecksumStream. keep-alive on.
const origin = http.createServer((_req, res) => {
  res.writeHead(200, {
    'content-type': 'application/octet-stream',
    'content-length': String(BLOB_SIZE),
    'x-amz-checksum-crc32': crcHeader,
    etag: '"deadbeef"',
  })
  let sent = 0
  const tick = () => {
    if (res.writableEnded) return
    if (sent >= BLOB_SIZE) return res.end()
    res.write(BODY.subarray(sent, sent + 64 * 1024))
    sent += 64 * 1024
    setTimeout(tick, 5)
  }
  tick()
})
origin.keepAliveTimeout = 60_000
await once(origin.listen(0), 'listening')
const port = origin.address().port
const originConns = () => new Promise((r) => origin.getConnections((_e, n) => r(n)))

const s3 = new S3({
  region: 'auto',
  endpoint: `http://127.0.0.1:${port}`,
  forcePathStyle: true,
  credentials: { accessKeyId: 'x', secretAccessKey: 'y' },
  responseChecksumValidation: 'WHEN_SUPPORTED',
  requestHandler: { socketTimeout: 15_000, connectionTimeout: 5_000 },
})

async function oneAbortedDownload(i) {
  const controller = new AbortController()
  const res = await s3.getObject(
    { Bucket: 'b', Key: `blob-${i}`, ChecksumMode: 'ENABLED' },
    FIX_ABORT ? { abortSignal: controller.signal } : {},
  )
  const body = res.Body

  const sink = new PassThrough()
  const p = new Promise((resolve) => pipeline(body, sink, () => resolve()))
  await once(sink, 'data')       // read one chunk...
  sink.destroy()                 // ...then the downstream client goes away

  if (FIX_DESTROY && !body.destroyed) body.destroy()
  if (FIX_ABORT) controller.abort()

  await p.catch(() => {})
  await new Promise((r) => setTimeout(r, 5))
}

console.log(`@smithy/core=${coreVersion}  FIX_DESTROY=${FIX_DESTROY}  FIX_ABORT=${FIX_ABORT}  iters=${ITERS}`)
const before = await originConns()
for (let i = 0; i < ITERS; i++) {
  await oneAbortedDownload(i).catch((e) => {
    if (!/aborted|premature|ECONN|socket hang|checksum/i.test(String(e))) throw e
  })
}
await new Promise((r) => setTimeout(r, 500))
const after = await originConns()
console.log(`origin live connections: before=${before}  after ${ITERS} aborted downloads=${after}`)
console.log(after > 3 ? `LEAK: ${after} sockets still open at origin` : `OK: sockets released (${after})`)
origin.close()
process.exit(0)
```
</details>

### How to run

```bash
# fixed version (default resolves to @smithy/core 3.29.5):
npm install
node repro.mjs
# -> OK: sockets released (0)

# reproduce the leak on the version the lockfile had:
npm pkg set overrides.@smithy/core=3.26.0 && npm install
node repro.mjs
# -> LEAK: 40 sockets still open at origin

# consumer-side workarounds do NOT help on 3.26.0:
FIX_DESTROY=1 FIX_ABORT=1 node repro.mjs
# -> LEAK: 40 sockets still open at origin

# confirm the fix:
npm pkg delete overrides.@smithy/core && npm install
node repro.mjs
# -> OK: sockets released (0)
```

## Notes / scope

- This fixes the client-**abort** case. A pure silent-stall (origin stops sending with no FIN/RST) is a separate gap that #2152 does not cover (no body-read inactivity timeout); a defensive `AbortController` on `S3BlobStore` reads would be a reasonable follow-up but is out of scope here.
- No tests added.

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -21,6 +21,11 @@ allowBuilds:
   '@datadog/pprof': false
   core-js: false
 
+minimumReleaseAgeExclude:
+  - '@smithy/core@3.29.5'
+
 overrides:
   cookie: ^0.7.2
   opentelemetry-plugin-better-sqlite3>@opentelemetry/core: ^1.30.0
+  # Remove once @aws-sdk/core requires >=3.29.5 transitively.
+  '@smithy/core': ^3.29.5
--- a/.changeset/smithy-core-socket-leak.md
+++ b/.changeset/smithy-core-socket-leak.md
@@ -1,0 +1,5 @@
+---
+'@atproto/aws': patch
+---
+
+Pin `@smithy/core` to `>=3.29.5` to fix an outbound S3/R2 socket leak. Earlier versions did not destroy the underlying stream's socket when a `getObject` download was aborted mid-transfer (`ChecksumStream` swallowed the premature close), so `S3BlobStore` blob downloads that clients cancelled leaked sockets and file descriptors until the host exhausted TCP memory. `@smithy/core` 3.29.5 (smithy-typescript#2152) destroys the source and removes its listeners on premature closure.
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -206,6 +206,7 @@ settings:
 overrides:
   cookie: ^0.7.2
   opentelemetry-plugin-better-sqlite3>@opentelemetry/core: ^1.30.0
+  '@smithy/core': ^3.29.5
 
 importers:
 
@@ -352,16 +353,16 @@ importers:
         version: link:../repo
       '@aws-sdk/client-cloudfront':
         specifier: ^3.1073.0
-        version: 3.1075.0
+        version: 3.1088.0
       '@aws-sdk/client-kms':
         specifier: ^3.1073.0
-        version: 3.1075.0
+        version: 3.1088.0
       '@aws-sdk/client-s3':
         specifier: ^3.1073.0
-        version: 3.1075.0
+        version: 3.1088.0
       '@aws-sdk/lib-storage':
         specifier: ^3.1073.0
-        version: 3.1075.0(@aws-sdk/client-s3@3.1075.0)
+        version: 3.1088.0(@aws-sdk/client-s3@3.1088.0)
       '@noble/curves':
         specifier: ^1.7.0
         version: 1.8.0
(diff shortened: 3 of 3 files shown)


pnpm-workspace.yaml
```diff
@@ -21,6 +21,11 @@ allowBuilds:
...
   core-js: false
 
+minimumReleaseAgeExclude:
+  - '@smithy/core@3.29.5'
+
 overrides:
   cookie: ^0.7.2
   opentelemetry-plugin-better-sqlite3>@opentelemetry/core: ^1.30.0
+  # Remove once @aws-sdk/core requires >=3.29.5 transitively.
+  '@smithy/core': ^3.29.5
```
jcalabro, line 31: override required because it's less than our specified min dependency age policy

PR #5249, opened 2026-07-17
Add canonical OP thread numbering

estrattonbailey:
An OP thread is the oldest contiguous chain of replies starting from the root, all authored by OP.


lexicons/app/bsky/unspecced/defs.json
```diff
@@ -109,7 +109,15 @@
         },
         "opThread": {
           "type": "boolean",
-          "description": "This post is part of a contiguous thread by the OP from the thread root. Many different OP threads can happen in the same thread."
+          "description": "This post is part of a contiguous thread by the OP from the thread root. Sub-threads by OP deeper in the tree are not considered an OP thread."
```
estrattonbailey, line 112: This description was wrong previously

packages/bsky/src/data-plane/server/db/migrations/20260729T223157563Z-add-op-thread-reply.ts
```diff
@@ -0,0 +1,16 @@
+import type { Kysely } from 'kysely'
+
+export async function up(db: Kysely<unknown>): Promise<void> {
+  await db.schema
+    .createTable('op_thread_reply')
```
estrattonbailey, line 5: Fully mimics our prod table design instead of trying to recreate it using relational queries

packages/bsky/src/views/threads-v2.ts
```diff
@@ -207,6 +213,9 @@ function applyBumping(
     /*
       General bumps.
     */
+    // OP thread replies: the chain continuation always appears first beneath
+    // its parent, regardless of the selected sort.
+    ['up', (i) => !!opThreadUris?.has(i.item.uri)],
```
estrattonbailey, line 218: Just noting that I did include this as discussed

rafaeleyng (approved):

PR #5250, opened 2026-07-17
Add recIdStr field to getTrends endpoint

ds-boyce:
Need this for engagement metrics.

### DIFF
--- a/packages/bsky/src/api/app/bsky/unspecced/getTrends.ts
+++ b/packages/bsky/src/api/app/bsky/unspecced/getTrends.ts
@@ -115,6 +115,7 @@ const noBlocks: RulesFn<Context, Params, SkeletonState> = (input) => {
   const blocks = hydration.bidirectionalBlocks?.get(viewer)
 
   return {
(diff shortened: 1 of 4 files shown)


packages/bsky/src/api/app/bsky/unspecced/getTrends.ts
```diff
@@ -115,6 +115,7 @@ const noBlocks: RulesFn<Context, Params, SkeletonState> = (input) => {
   const blocks = hydration.bidirectionalBlocks?.get(viewer)
 
   return {
+    ...skeleton,
```
matthieusieben: `skeleton` could be updated in the future to contain internal data. That is for example the case in some endpoints where the skeleton contains headers, or params. This data would leak because of the spread operator used here, and TypeScript would not warn us.

For this reason, I recommend using exact fields enumeration here.


```suggestion
    recIdStr: skeleton.recIdStr,
```

estrattonbailey (approved):

PR #5259, opened 2026-07-22
Add recId to getTrendingTopics

ds-boyce:
https://github.com/bluesky-social/tango/blob/main/iris/iris/routes/topics.py#L81


ds-boyce:
This is legacy, we’re just going to cut over to `getTrends`.

PR #5263, opened 2026-07-23
Hydrate starter pack info for follow notifications

ds-boyce:


packages/bsky/src/hydration/graph.ts
```diff
@@ -44,6 +44,20 @@ export type ListMembershipStates = HydrationMap<
 export type Follow = RecordInfo<FollowRecord>
 export type Follows = HydrationMap<AtUriString, Follow>
 
+export const getStarterPackUriFromFollow = (
```
estrattonbailey: Nit: pry better suited to the `util.ts` file in this same directory

lexicons/app/bsky/notification/listNotifications.json
```diff
@@ -80,6 +80,11 @@
         },
         "reasonSubject": { "type": "string", "format": "at-uri" },
         "record": { "type": "unknown" },
```
estrattonbailey, line 82: Note for other reviewers: for v2, we'll pry want to combine all related records/views into a single field. We talked about that, but for now we think it's fine to just add a one-off field.

estrattonbailey (approved):
LGTM! One small nit

PR #5266, opened 2026-07-24
oauth: propagate store errors in CachedGetter and SessionGetter

matthieusieben:
The goal of this PR is to make the OAuth session getter a little bit simpler and to surface `store.get()` errors instead of silently swallowing them, which can cause confusing errors like in https://bsky.app/profile/kandake.africa/post/3mrevwuwols22

When an `"The session was deleted by another process"` error occurs, it **could** mean that the store failed, and that the error was swallowed by this try catch:

https://github.com/bluesky-social/atproto/blob/5c5df26bdeed22a06aa08e0463b76ad19be59ed8/packages/internal/simple-store/src/cached-getter.ts#L182-L186

This PR makes sure those errors get propagated instead of transformed into seemingly unrelated errors messages.


packages/oauth/oauth-client/src/session-getter.ts
```diff
@@ -185,64 +189,64 @@ export class SessionGetter extends CachedGetter<AtprotoDid, Session> {
...
+        await server.revoke(tokenSet.refresh_token ?? tokenSet.access_token)
+      } catch (err) {
+        errors.push(err)
+      }
+
+      try {
+        await this.delStored(sub, err)
+      } catch (err) {
+        errors.push(err)
+      }
```
matthieusieben, line 232: Moved from the (now removed) `onStoreError` hook

packages/internal/simple-store/src/cached-getter.ts
```diff
@@ -155,47 +168,45 @@ export class CachedGetter<
...
 
   async setStored(key: K, value: V): Promise<void> {
-    try {
-      await this.store.set(key, value)
-    } catch (err) {
-      const onStoreError = this.options?.onStoreError
-      await onStoreError?.(err, key, value)
-    }
+    await this.#store.set(key, value)
   }
```
matthieusieben, line 217: Main difference of this PR. Error handling now has to be done either in the store or in a sub-class overriding these methods

packages/internal/simple-store/src/cached-getter.ts
```diff
@@ -124,6 +130,9 @@ export class CachedGetter<
 
     const currentExecutionFlow: PendingItem<V> = Promise.resolve()
       .then(async () => {
+        // @NOTE that the signal should not be aborted at this point, because
+        // the loop above would have thrown an error if it was.
+
         const storedValue = await this.getStored(key, { signal })
```
ThisIsMissEm, line 136: do we need to handle the signal being aborted after the getStored? Not just on line 148; I'm not sure how long `allowStored` is meant to take to complete?

packages/internal/simple-store/src/cached-getter.ts
```diff
@@ -155,47 +168,45 @@ export class CachedGetter<
...
 
-    this.pending.set(key, currentExecutionFlow)
+    this.#pending.set(key, currentExecutionFlow)
 
     const { value } = await currentExecutionFlow
     return value
   }
 
+  // @NOTE We propagate errors from the store. If the use-case prefers to ignore
+  // errors from the state, it should be handled in the store implementation
```
ThisIsMissEm:
```suggestion
  // errors from the store, it should be handled in the store implementation
```
`state` is a typo here isn't it? I'm not 100% sure?

packages/internal/simple-store/src/cached-getter.ts
ThisIsMissEm: or in the caller right?

ThisIsMissEm (commented):
This makes sense to me, just a few small comments

packages/internal/simple-store/src/cached-getter.ts
matthieusieben: These methods get called as part of the `get` method. If they throw, the get method will fail (so yes), but what I meant is that if we want to prevent these methods from throwing (e.g. ignore a cache fetch and return undefined instead), it has to be done from either the store or an override of the methods here.

packages/internal/simple-store/src/cached-getter.ts
matthieusieben: The reasoning here is that there is no point in aborting instead of returning the value, if we have the value. In this particular instance, it allows concurrent requests to obtain the same stored value without having to call `getStored`.

ThisIsMissEm: Okay, cool, wasn't 100% sure

packages/internal/simple-store/src/cached-getter.ts
ThisIsMissEm: Right, that makes sense; I was thinking from the perspective of public API since that's how I've most interacted with stores

PR #5267, opened 2026-07-24
Add default `retry` option to `putRecord`

matthieusieben:
`InvalidSwap` errors on `client.put()` should not be retried by default. this PR adds a default `retry` method that implements that behavior.


matthieusieben:
Looks like this is not necessary because of 400 not being retryable by default. The pattern should be to provide `retry` overrides when we want to force a retry rather than prevent it.

PR #5270, opened 2026-07-24
Resolve DID login_hint to the account handle on the authorization page

### LINKED ISSUE #4900: login_hint handling should allow an app to specify a did and have the identifier rendered as a handle to the user
**Describe the bug**

If we pass a did as a login_hint, the user is presented with the identifier field pre-filled with the did instead of resolved to the user-facing handle

**To Reproduce**

1. Start an OAuth user authentication flow with the `https://bsky.social` authorization server, with a header specifying a `login_hint` of the did.

**Expected behavior**

In a situation where the user did not enter a did, but we know the did we want to authorize, the user should see the current handle for that did. Instead, the user is presented with a did they are unlikely to recognize.

**Details**

- Operating system: iOS 26.2
- Node version:

**Additional context**

I understand the guidance from the [documentation](https://atproto.com/specs/oauth):

> The resolved DID should be bound to the overall auth session and should be used as the primary account identifier within client app code. Handles (when verified) are acceptable to display in user interfaces, but may change over time and need to be re-verified periodically. When passing an account identifier through to the Authorization Server as part of the Authorization Request in the login_hint, it is recommended to use the exact account identifier supplied by the user (handle or DID) to ensure any sign-in flow is consistent (users might not recognize their own account DID).

This presumes that the identifier was directly entered by the user and should be passed through as-is to the authorization server. There are two issues with this:
- we should treat the case of users entering their did as exceptional
- there are many cases where the app's intent is to authenticate for a particular did, and apps should be able to communicate the did to the authorization server and not use the handle as an intermediate representation between two systems.

For example:
* The (link underneath) a Germ button signifies an intent to message as a sender did to a recipient did. The user didn't enter their did, but it was conveyed across applications using the durable identifier and not the mutable user-facing handle representation. 
* When users' refresh token expires, we re-initiate an OAuth authorization flow to regain access to protected resources. The user previously entered a handle, we resolved it to and authenticated it for the did, which corresponds to some local state indexed by their did. We're going to re-authenticate them for that did, and the expected behavior is that the user lands on the authorization page and see a familiar user-facing identifier, ideally their current handle (not email, would be a privacy leakage). 

So yes, of course, on our side we can (and are) resolving the did back to handle and supplying the freshly resolved handle as the login_hint. I'd characterize this as a workaround and not the ideal. When the app's intent is to authorize a did, I think the correct behavior is for the relying party to specify the did (not the handle) and expect the authorization server to render to the user the non-exceptional path of showing the user their handle.

bigmoves:
## Motivation

Fixes #4900 — when an OAuth client passes a DID as `login_hint` (e.g. re-authenticating an account it already knows by its durable identifier), the sign-in form pre-fills the raw `did:plc:…` string, which users are unlikely to recognize as their own account.

## Changes

- New `resolveLoginHint()` helper in `packages/oauth/oauth-provider/src/account/login-hint.ts`: when the hint is a DID belonging to an account on this server, it returns the account's current handle; otherwise (unknown account, or account without a handle) it returns the hint unchanged.
- The authorize flow computes this value and passes it to the authorization page via a new optional `loginHint` field on `AuthorizationResultAuthorizePage`. The client's stored request parameters are untouched — only the display value sent to the UI changes, so SSO session matching and error redirects behave exactly as before. This also means the `@TODO` about verifying account existence in `request-manager.ts` is left as-is: an unknown DID still renders (as itself) rather than being rejected, to avoid changing behavior for existing clients.
- No UI changes needed: the page renders whatever identifier it receives, and its session-matching logic already accepts either DID or handle.
- First tests in `@atproto/oauth-provider`: adopted vitest (package config + `tsconfig.test.json`, registered in the root vitest projects list) with 5 unit tests covering resolution, both fallbacks, the no-lookup path for handle hints, and the undefined case. Also added the `**/*.test.ts` exclusion to `tsconfig.build.json` that this package never needed before.
- Changeset: patch bump for `@atproto/oauth-provider`.

## Screenshots

Client sends `login_hint=did:plc:272ug475putgca2kk7qqf4ux` (a known account) — the Identifier field now shows the handle. This is where the raw DID appeared before this change:

Sign-in form showing the resolved handle alice.test in the read-only Identifier field

Client sends a DID unknown to this server — graceful fallback, the DID renders unchanged (this is also exactly how every DID hint rendered before this change):

Sign-in form showing an unresolved DID in the Identifier field

> Note: screenshots are from a fork with a redesigned authorization UI; the affected Identifier field behaves identically in the stock UI.

## Verification

- `pnpm test` in `packages/oauth/oauth-provider` — 5/5 green (watched the DID-resolution case fail first, TDD).
- `tsgo --build` clean for `oauth-provider` and `pds`; eslint + prettier clean.
- Live end-to-end against dev-env: PAR with a known account's DID as `login_hint` → authorize page embeds the handle; PAR with an unknown DID → page shows the DID unchanged (see screenshots).

### DIFF
--- a/packages/oauth/oauth-provider/src/account/login-hint.ts
+++ b/packages/oauth/oauth-provider/src/account/login-hint.ts
@@ -1,0 +1,25 @@
+import { isAtprotoDid } from '@atproto/did'
+import type { AccountManager } from './account-manager.js'
+
+/**
+ * Computes the user-facing value for a `login_hint` request parameter. Clients
+ * may provide a DID (typically when re-authenticating an account they already
+ * know), which users are unlikely to recognize as their own. When the DID
+ * belongs to an account on this server, return that account's current handle
+ * for display instead.
+ */
+export async function resolveLoginHint(
+  hint: string | undefined,
+  accountManager: Pick<AccountManager, 'getAccount'>,
+): Promise<string | undefined> {
+  if (hint && isAtprotoDid(hint)) {
+    const account = await accountManager.getAccount(hint).then(
+      (result) => result.account,
+      // Unknown account: fall back to displaying the hint as-is
+      () => null,
+    )
+    if (account?.handle) return account.handle
+  }
+
+  return hint
+}
--- a/.changeset/olive-pans-yell.md
+++ b/.changeset/olive-pans-yell.md
@@ -1,0 +1,5 @@
+---
+'@atproto/oauth-provider': patch
+---
+
+Display the account's current handle on the authorization page when a client provides a DID as `login_hint`
--- a/packages/oauth/oauth-provider/package.json
+++ b/packages/oauth/oauth-provider/package.json
@@ -93,9 +93,11 @@
     "@types/cookie": "^0.6.0",
     "@types/forwarded": "0.1.3",
     "@types/http-errors": "^2.0.4",
-    "@types/send": "^0.17.4"
+    "@types/send": "^0.17.4",
+    "vitest": "^4.0.16"
   },
   "scripts": {
-    "build": "tsgo --build tsconfig.build.json"
+    "build": "tsgo --build tsconfig.build.json",
+    "test": "vitest run"
   }
 }
--- a/packages/oauth/oauth-provider/src/account/login-hint.test.ts
+++ b/packages/oauth/oauth-provider/src/account/login-hint.test.ts
@@ -1,0 +1,55 @@
+import { describe, expect, it, vi } from 'vitest'
+import type { Account } from '@atproto/oauth-provider-api'
+import type { AccountManager } from './account-manager.js'
+import { resolveLoginHint } from './login-hint.js'
+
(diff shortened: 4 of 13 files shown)


packages/oauth/oauth-provider/src/account/login-hint.ts
```diff
@@ -0,0 +1,25 @@
...
+import type { AccountManager } from './account-manager.js'
+
+/**
+ * Computes the user-facing value for a `login_hint` request parameter. Clients
+ * may provide a DID (typically when re-authenticating an account they already
+ * know), which users are unlikely to recognize as their own. When the DID
+ * belongs to an account on this server, return that account's current handle
+ * for display instead.
+ */
+export async function resolveLoginHint(
```
matthieusieben, line 11: I would typically have added a method to the AccountManager but this is fine too

matthieusieben (approved):
Looking great thanks !

PR #5276, opened 2026-07-27
ozone: add materialized view refresher timeout

gcwill70:
Add timeouts in ozone daemon processes for better durability

The daemon's materialized view refresher previously ran with no timeouts and no locking. The refresh cycle now runs on a single dedicated connection with:

- a pg_try_advisory_lock guard so only one cycle is guaranteed to run at once
- statement_timeout so that crashed refreshes release the lock

## Tests

- skipping the cycle while another cycle holds the advisory lock
- a full cycle logging per-view durations and releasing the lock between runs
- a failing view not preventing the remaining views from refreshing

### DIFF
--- a/packages/ozone/src/daemon/materialized-view-refresher.ts
+++ b/packages/ozone/src/daemon/materialized-view-refresher.ts
@@ -1,32 +1,62 @@
-import { sql } from 'kysely'
 import { MINUTE } from '@atproto/common'
 import { type BackgroundQueue, PeriodicBackgroundTask } from '../background.js'
 import { dbLogger } from '../logger.js'
+import { MATERIALIZED_VIEW_REFRESH_LOCK_ID } from '../db/index.js'
+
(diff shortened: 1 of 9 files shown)


packages/ozone/src/daemon/materialized-view-refresher.ts
```diff
@@ -1,32 +1,62 @@
...
-        if (signal.aborted) break
+  constructor(
+    backgroundQueue: BackgroundQueue,
+    interval = 30 * MINUTE,
+    statementTimeoutMs = 10 * MINUTE,
+  ) {
+    super(backgroundQueue, interval, async (db, signal) => {
+      let locked = false
+      // Create single client for the whole refresh cycle
+      const client = await db.pool.connect()
```
foysalit, line 17: interesting. dont think we have used this approach anywhere before!

foysalit (approved):
looks good!

foysalit (approved):
nice! like that it's smaller in scope.

PR #5278, opened 2026-07-29
Check plc.directory before trying to update it, unbreaking updates for accounts with self-custodied rotation keys

edmundedgar:
`updateHandle` for a did:plc account currently assumes it always needs to update plc.directory and unconditionally tries to sign an update and send it. But in some circumstances this assumption does not hold: If the user is holding their own keys, they should update the directory themselves, then call `updateHandle` against the PDS to get it to update its local database and send the relevant sync event. This is also the flow a did:web user would use. This currently fails with an invalid signature error because the PDS doesn't have a rotation key they can sign the update with, so the unnecessary update attempt fails and the failure blocks the other steps in the update.

There was a TODO in the code to check the state of the plc.directory before trying to update it, probably because even if the PDS has the necessary keys, the lack of a check will still result in the PDS creating an unnecessary duplicate PLC operation if a previous handle update failed partway through.

This PR adds the missing check, and also adds some tests to cover the expected handle update flow of a user who doesn't trust the PDS with their rotation keys.

### DIFF
--- a/.changeset/tame-hounds-relax.md
+++ b/.changeset/tame-hounds-relax.md
@@ -1,0 +1,5 @@
+---
+'@atproto/pds': patch
+---
+
+Don't require the PDS's own rotation key to update a handle that's already been changed directly on the DID document (e.g. by a self-custodied account)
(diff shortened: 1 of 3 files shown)


edmundedgar:
Closing, will resubmit as a new PR that combines this and https://github.com/bluesky-social/atproto/pull/5279 and other feedback from that PR

PR #5279, opened 2026-07-29
Don't require the PDS's own rotation key in the DID doc when activating

### LINKED ISSUE #4819: activateAccount for non-server-custodied keys fails with "Server rotation key not included in PLC DID data"
**Describe the bug**

If you create an account with a did:plc ID and delegate it to a PDS without delegating control of the ID itself to the PDS, `activateAccount` fails with "Server rotation key not included in PLC DID data".

1. Generate a single self-custodied rotation key
2. Create a new did:plc ID controlled by that rotation key 
3. Call `reserveSigningKey` to get the signing key the PDS wants you to use
4. Call `createAccount` with handle, password, invite code etc to create the account and get the resulting `accessJwt`
5. Update the did:plc ID to use the handle, PDS and signing key expected by the PDS, but don't change the rotation key
5. Call `activateAccount` passing it the supplied `accessJwt`.

**Expected behavior**

Call returns a success code, account is activated

**Details**

You get an error 400 with the error message "Server rotation key not included in PLC DID data".

**Additional context**

This seems to be caused by a sanity check in [atproto/server/util.ts](https://github.com/bluesky-social/atproto/blob/main/packages/pds/src/api/com/atproto/server/util.ts#L99) which seems to assume that you would always want your PDS to have permission to rewrite your account, albeit potentially with a lower-priority key. The check only seems to happen on account activation so it could be bypassed by setting both a self-custodied key and the server's key initially, then using the self-custodied key to remove the server's key once activation is done.

It would be useful if we could either remove the check entirely or allow it to be bypassed with a flag stating that self-custody really is the user's intention.

edmundedgar:
Fixes #4819. This PR allows activation even if the user is holding their own rotation keys instead of entrusting them with the PDS.

Without this fix a user could still remove the PDS's rotation key, which may break assumptions that some front-ends make about handle changes etc, but they had to first create the account with it included, and then immediately remove it, which is more error-prone and increases the user's trust assumptions on the PLC directory.

### DIFF
--- a/.changeset/silly-otters-jump.md
+++ b/.changeset/silly-otters-jump.md
@@ -1,0 +1,5 @@
+---
+'@atproto/pds': patch
+---
+
+Don't require the PDS's own rotation key to be in the DID document when activating an account (e.g. for a self-custodied account)
--- a/packages/pds/src/api/com/atproto/server/util.ts
+++ b/packages/pds/src/api/com/atproto/server/util.ts
@@ -84,7 +84,6 @@ export const assertValidDidDocumentForService = async (
     await assertValidDocContents(ctx, did, {
       pdsEndpoint: resolved.services['atproto_pds']?.endpoint,
       signingKey: resolved.verificationMethods['atproto'],
-      rotationKeys: resolved.rotationKeys,
     })
   } else {
     const resolved = await ctx.idResolver.did.resolve(did, true)
@@ -108,18 +107,9 @@ const assertValidDocContents = async (
   contents: {
     signingKey?: string
     pdsEndpoint?: string
-    rotationKeys?: string[]
   },
 ) => {
-  const { signingKey, pdsEndpoint, rotationKeys } = contents
-
-  const plcRotationKey =
-    ctx.cfg.entryway?.plcRotationKey ?? ctx.plcRotationKey.did()
-  if (rotationKeys !== undefined && !rotationKeys.includes(plcRotationKey)) {
-    throw new InvalidRequestError(
-      'Server rotation key not included in PLC DID data',
-    )
-  }
+  const { signingKey, pdsEndpoint } = contents
 
   if (!pdsEndpoint || pdsEndpoint !== ctx.cfg.service.publicUrl) {
     throw new InvalidRequestError(


matthieusieben:
Hey,

Thank you for your contribution !

While we **do** want to support this use case, there are two things that worry me:

- This change is not covered by any tests
- Other places in the code assume that the rotation key is the one of the PDS. Those other places should be adapted accordingly. The UX when an operation cannot be completed should never be a 5xx.

matthieusieben:
@claude review this

edmundedgar:
@matthieusieben Thanks for looking at this.

> This change is not covered by any tests

Will add tests.

> Other places in the code assume that the rotation key is the one of the PDS. Those other places should be adapted accordingly. The UX when an operation cannot be completed should never be a 5xx.

I have a separate PR for `updateHandle` ( https://github.com/bluesky-social/atproto/pull/5278 ) . I'm happy to replace this one and that one with a broader PR to cover that and `submitPlcOperation` and possibly also `signPlcOperation`  if that's helpful?

matthieusieben:
Yeah, merging both PRs might be a good idea. These are both part of a same feature.

edmundedgar:
@matthieusieben ok, will merge those and also take a look at `submitPlcOperation` and `signPlcOperation` to see if they need anything doing to them.

I think currently `signPlcOperation` will just sign whatever you send it regardless of whether its key is actually useful for your account, and also will fail in a slightly mysterious way for a did:web. I suppose it would be more helpful it it checked and gave you an error.

edmundedgar:
Closing this, will include it in a new PR which also does https://github.com/bluesky-social/atproto/pull/5278 and incorporates other feedback.

edmundedgar:
@matthieusieben OK, this PR fixes this and`updateHandle`. For consistency I also fixed `submitPlcOperation` although I don't have a strong opinion about whether it should be loosened or not. It also does a general clean-up before removing the rotation key requirement to make sure these checks are done in a consistent way anywhere the PDS wants to sign stuff for the user. 

https://github.com/bluesky-social/atproto/pull/5349

PR #5285, opened 2026-07-30
Rework `@atproto/ws-client` around a reconnecting async iterable

devinivy:
Replaces `WebSocketKeepAlive` with `websocket(url, options?)`, an async iterable that connects, reads messages, reconnects on failure, and yields one continuous stream. Same API on Node.js and in the browser, with the platform transport selected at import time via a conditional `imports` entry. Follows the direction proposed in review of #5242: iterator-only consumption, no connection class, no `close()`.

**Consuming**

- A stream ends by breaking out of the loop or aborting a `signal` — there is no `close()`. How you stop decides how the socket ends: `break` or a bare `abort()` closes with 1000, `abort(new CloseError(…))` closes with that code, any other abort reason destroys the connection.
- A non-reconnectable clean close ends the loop normally; any other end rejects the iterator.
- `dataMode: 'auto' | 'text' | 'binary'` types what's yielded and is enforced at runtime.
- Hooks at two levels: `onOpen`/`onClose` fire once per stream, `onConnect(sender)`/`onDisconnect` per connection.

**Liveness and flow control**

- Heartbeat on by default on Node.js (10s, `heartbeat: false` to disable); `idleTimeoutMs` on both platforms and the browser's only dead-connection detector.
- Real socket backpressure on Node.js past `highWaterMark`; `maxBufferedBytes` as a hard cap on both platforms.

**Sending**

Sending goes through the per-connection `sender` handed to `onConnect`, and `send()` resolves on hand-off, not delivery. There is deliberately no send queue: a queued send could only be flushed by a later connection, and reconnects only happen as the consumer pulls, so awaiting a queued send from inside the loop deadlocks — both tap indexers did exactly that (the deadlock reproduces against the previous client on `main`).

**Reconnects**

Jittered exponential backoff, resetting after each successful open. Retry decisions are driven by close codes (per RFC 6455) and typed errors extending `WebSocketClientError`, overridable via `shouldReconnect`.

**Consumers migrated**

- `xrpc-server` `Subscription` — now owns `DisconnectError`; explicit options instead of arbitrary `ws` `ClientOptions`; reconnects on transient close codes (e.g. 1001) where it previously gave up.
- `sync` `Firehose` — explicit connection options.
- `tap` `TapChannel` — `ackEvent()` resolves once an ack is sent or recorded for the next connection; pending acks flush from `onConnect`.
- `ozone` `Jetstream` — `close()` now resolves once the stream has fully stopped.

`@atproto/ws-client` no longer depends on `@atproto/common`. See the changesets for per-package breaking-change details and the rewritten README for full usage.


packages/ws-client/src/transport/transport.ts
```diff
@@ -0,0 +1,54 @@
...
+   * {@link DEFAULT_HEARTBEAT_INTERVAL_MS}; pass `false` to disable, or an object
+   * to set the interval. `intervalMs` is optional, so `{}` means "on, default
+   * interval".
+   *
+   * Node only. The WHATWG API has no ping/pong, so the browser transport ignores
+   * this and relies on `idleTimeoutMs` instead.
+   */
+  heartbeat?: { intervalMs?: number } | false
+  idleTimeoutMs?: number
+  highWaterMark?: number
```
matthieusieben: We should distinguish `readableHighWaterMark` from `writeableHighWaterMark` (which one is this supposed to be?)

packages/ws-client/src/transport/transport.ts
```diff
@@ -0,0 +1,54 @@
...
+  maxBufferedBytes?: number
```
matthieusieben: We should comment that these are readable bytes.

And also have a specific mechanism (error + should retry hook) that allows users of the lib to retry on buffer overflow. This would allow a slow consuming browser to *optionally* rely on the server's ability to restart from a point in the past (if it does, so this should be optional)

packages/ws-client/src/transport/transport.ts
```diff
@@ -0,0 +1,54 @@
...
+  headers?: HeadersInit
+  protocols?: string | string[]
+  /** The socket opened; `sender` is valid until this connection ends. */
+  onOpen(sender: Sender<M>): void
+  onClose(detail: CloseEventDetail): void
```
matthieusieben, line 92: We should document wether this always triggers, or only if the transport was ever open.

packages/ws-client/src/transport/browser-transport.ts
```diff
@@ -0,0 +1,231 @@
...
+  const ws = new WebSocketImpl(options.url, options.protocols)
+  ws.binaryType = 'arraybuffer'
+
+  let open = false
+  let closeReported = false
+
+  function reportClose(detail: CloseEventDetail): void {
+    if (closeReported) return
+    closeReported = true
+    options.onClose(detail)
```
matthieusieben, line 161: hooks should not leak their context

```suggestion
    options.onClose.call(null, detail)
```

packages/ws-client/src/transport/browser-transport.ts
```diff
@@ -0,0 +1,231 @@
...
+        throw new WebSocketClientError('WebSocket is not open')
+      }
+      // The WHATWG API gives no flush notification, only indicates that the browser accepted the write
+      ws.send(data)
+    },
+  }
+
+  ws.addEventListener('open', () => {
+    open = true
+    options.onOpen(sender)
```
matthieusieben, line 198:
```suggestion
    options.onOpen.call(null, sender)
```

packages/ws-client/src/transport/browser-transport.ts
```diff
@@ -0,0 +1,231 @@
...
+
+  ws.addEventListener('close', (ev) => {
+    open = false
+    const detail: CloseEventDetail = {
+      code: ev.code,
+      reason: ev.reason,
+      wasClean: ev.wasClean,
+    }
+    channel.finish()
+    reportClose(detail)
```
matthieusieben: Nit: Any particular reason for not relying on TS's type inference here? 

```suggestion
    channel.finish()
    reportClose({
      code: ev.code,
      reason: ev.reason,
      wasClean: ev.wasClean,
    })
```

packages/ws-client/src/transport/node-transport.ts
```diff
@@ -0,0 +1,225 @@
...
+} from './transport.js'
+
+/**
+ * Node transport, built on `ws`. Created already connecting; torn down by
+ * `options.signal`. Every receive-side concern (buffering, watermarks,
+ * `dataMode` enforcement, idle timeout) belongs to `createMessageChannel`; this
+ * wires the raw socket to it.
+ *
+ * Deliberately not `createWebSocketStream`, which never reads the close event's
+ * code — the reconnect policy classifies by exactly that.
```
matthieusieben, line 37: Damn that's too bad.

packages/ws-client/src/transport/node-transport.ts
```diff
@@ -0,0 +1,225 @@
...
+
+  const sender: Sender<M> = {
+    async send(data) {
+      if (!open) {
+        throw new WebSocketClientError('WebSocket is not open')
+      }
+      return new Promise<void>((resolve, reject) => {
+        ws.send(data, (err) => (err ? reject(new SocketError(err)) : resolve()))
+      })
+    },
```
matthieusieben, line 185: Nit: we can save a "tick" by doing:
Also Nit: `Promise` should not need to specify the generics (should be inferred from the `Sender` interface).

```suggestion
    send(data) {
      return new Promise((resolve, reject) => {
        if (!open) {
          reject(new WebSocketClientError('WebSocket is not open'))
        } else {
          ws.send(data, (err) => (err ? reject(new SocketError(err)) : resolve()))
        }
      })
    },
```

packages/ws-client/src/transport/node-transport.ts
```diff
@@ -0,0 +1,225 @@
...
+  }
+
+  ws.on('open', () => {
+    open = true
+    startHeartbeat()
+    options.onOpen(sender)
```
matthieusieben, line 191:
```suggestion
    options.onOpen.call(null, sender)
```

packages/ws-client/src/transport/node-transport.ts
```diff
@@ -0,0 +1,225 @@
...
+      // closes politely at 1000, a CloseError closes with its code, anything else
+      // is a failure that destroys the connection. On the polite paths the real
+      // close event carries the detail, which is why nothing is reported here.
+      const code = closeCodeForStop(options.signal.reason)
+      if (code !== undefined) {
+        ws.close(code)
+        return
+      }
+      ws.terminate()
+      reportClose(ABNORMAL_CLOSE_DETAIL)
```
matthieusieben: I am not too sure about this. It seems reasonable to me that a user of the lib could provide their own `signal` that would be aborted with a custom error. This should trigger the "polite" path IMO. What is the use case for abruptly terminating here ?

packages/ws-client/src/websocket.ts
```diff
@@ -0,0 +1,379 @@
...
+
+export type Awaitable<T> = T | Promise<T>
+
+/**
+ * The receive, liveness, and flow-control options a transport accepts. The
+ * per-connection wiring (`url`, `signal`, `onOpen`, `onClose`) is omitted
+ * because the reconnect loop supplies it for itself, so a caller can't collide
+ * with it.
+ */
+export type ConnectionOptions<M extends DataMode = 'auto'> = Omit<
```
matthieusieben: Does this need to be part of the public interface ?

packages/ws-client/src/websocket.ts
```diff
@@ -0,0 +1,379 @@
...
+  maxReconnectSeconds?: number
+  /** Abort to end the stream permanently; the iterator rejects with the reason. */
+  signal?: AbortSignal
+  /**
+   * Controls reconnection. `true` (the default) uses the built-in policy
+   * ({@link defaultShouldReconnect}: typed errors self-classify, close codes
+   * by RFC 6455). `false` never reconnects — the first end is terminal. A
+   * function fully replaces the default classification.
+   */
+  shouldReconnect?: boolean | ((error: unknown, attempt: number) => boolean)
```
matthieusieben, line 50: If true is the default, a user who wants `false` can very easily provide `() => false`. I don't think the extra complexity is worth it in this case.

```suggestion
  shouldReconnect?: ((error: unknown, attempt: number) => boolean)
```

packages/ws-client/src/websocket.ts
```diff
@@ -0,0 +1,379 @@
...
+ * `signal`.
+ *
+ * An `AsyncGenerator` rather than a bare `AsyncIterable`, so a consumer can use
+ * `next()` or `return()` if it wants to. The runtime object is a generator
+ * either way, and a quieter type would only hide that.
+ */
+export type WebSocketIterable<M extends DataMode = 'auto'> = AsyncGenerator<
+  MessageOf<M>,
+  void,
+  undefined
```
matthieusieben: `unknown` is safer here. We can't prevent the caller from calling `it.next("blah")`, which would make our code inaccurate (if we ever used the result of `yield`).
```suggestion
  unknown
```

matthieusieben: The "best" solution is probable to use `AsyncGenerator` instead of `WebSocketIterable` below:

- `WebSocketIterable` is defined simply as `AsyncGenerator<MessageOf<M>>` ( we don't care what the user passes to the `next()` function)
- The generator bellow explicitly uses `AsyncGenerator<MessageOf<M>, void, unknown>` to ensure no invalid use of `yield` output is made

packages/ws-client/src/websocket.ts
```diff
@@ -0,0 +1,379 @@
...
+) => WebSocketIterable<M>
+
+// Reported when a stream ends with no close frame ever having applied — e.g. a
+// stop while parked in backoff between attempts. 1005 is the WHATWG convention
+// for "no status received".
+const NO_STATUS_DETAIL: CloseEventDetail = {
+  code: CloseCode.NoStatus,
+  reason: '',
+  wasClean: false,
+}
```
matthieusieben: This is exposed in userland so it should be protected (or not use a singleton at all).

```suggestion
const NO_STATUS_DETAIL: CloseEventDetail = Object.freeze({
  code: CloseCode.NoStatus,
  reason: '',
  wasClean: false,
})
```

packages/ws-client/src/websocket.ts
```diff
@@ -0,0 +1,379 @@
...
+    let awaitClose: Promise<void> | undefined
+
+    try {
+      // The generator body doesn't run until the first pull, so an
+      // already-aborted signal rejects that first `next()` rather than the call
+      // that created the generator.
+      signal?.throwIfAborted()
+
+      for (;;) {
+        const resolved = typeof url === 'function' ? await url() : url
```
matthieusieben: It seems to me that it's simpler (and more future proof) to throw a single time at the start of the loop instead of once before the loop, and once on "retry"

```suggestion
      for (;;) {
        // The generator body doesn't run until the first pull, so an
        // already-aborted signal rejects that first `next()` rather than the call
        // that created the generator.
        signal?.throwIfAborted()
  
        const resolved = typeof url === 'function' ? await url() : url
```

^ and remove the `signal?.throwIfAborted()` in the catch block below.

packages/ws-client/src/websocket.ts
```diff
@@ -0,0 +1,379 @@
...
+// listener. There's no separate cleanup to remember, and a per-attempt
+// controller that's always aborted in a `finally` can never leak one.
+function forwardAbort(
+  source: AbortSignal | undefined,
+  target: AbortController,
+): void {
+  source?.addEventListener('abort', () => target.abort(source.reason), {
+    once: true,
+    signal: target.signal,
+  })
```
matthieusieben, line 316: See my other comment. I think we should probably not propagate custom errors here, since it is a legit thing a client might do:

```ts
const ac = new AbortContoller()
ac.signal.addEventListener("abort", () => {
  console.error('shutting down... Reason:', ac.signal.reason)
})

process.once('SIGINT', () => ac.abort(new Error('SIGINT')))
process.once('SIGTERM', () => ac.abort(new Error('SIGTERM')))

websocket({
 // ...
 signal: ac.signal
})


```


Either we always use `Normal` as fallback (see other comment), or we do something like this here:


```suggestion
  source?.addEventListener(
    'abort',
    () => {
      const { reason } = source
      
      // Allow the consumer to control the WS close code by using a specific reason
      target.abort(reason instanceof CloseError ? reason : undefined)
    }, {
      once: true,
      signal: target.signal,
    })
```

packages/ws-client/src/websocket.ts
```diff
@@ -0,0 +1,379 @@
...
+        let reportClosed!: () => void
+        const closeReported = new Promise<void>((resolve) => {
+          reportClosed = resolve
+        })
+        const invalidateSender = () => {
+          live = false
+        }
+        try {
+          const transport = createTransport<M>({
+            ...connectionOptions(options),
```
matthieusieben: Nit: In terms of readability, I don't like that the protperties are not inlined here

packages/ws-client/src/transport/transport.ts
```diff
@@ -0,0 +1,54 @@
...
+  /**
+   * Protocol ping/pong liveness. On by default at
+   * {@link DEFAULT_HEARTBEAT_INTERVAL_MS}; pass `false` to disable, or an object
+   * to set the interval. `intervalMs` is optional, so `{}` means "on, default
+   * interval".
+   *
+   * Node only. The WHATWG API has no ping/pong, so the browser transport ignores
+   * this and relies on `idleTimeoutMs` instead.
+   */
+  heartbeat?: { intervalMs?: number } | false
```
matthieusieben: Note: since this is an internal API, I recommend *not* marking these as optional, but instead by marking them as explicitly potentially `undefined`

```suggestion
  heartbeat: undefined | false | { intervalMs?: number }
```

This makes a difference at the call site where a missing property or a typo could go undetected:

```ts

createTransport({
  //...

  // Typo here goes undetected because `heartbeat` is optional
  hartBeat: {}
})

```

packages/ws-client/src/transport/browser-transport.ts
matthieusieben: nevermind I see that these are internal hooks.

packages/ws-client/src/websocket.ts
```diff
@@ -0,0 +1,379 @@
...
+                },
+              }
+              // onOpen bookends the stream and fires once, before the first
+              // onConnect; onConnect fires for every connection including this
+              // one.
+              if (firstOpen) {
+                firstOpen = false
+                invokeHook(options.onOpen)
+              }
+              invokeHook(options.onConnect, scoped)
```
matthieusieben: Nit: Any reason not to use TS's type inference here ?


```suggestion
              // onOpen bookends the stream and fires once, before the first
              // onConnect; onConnect fires for every connection including this
              // one.
              if (firstOpen) {
                firstOpen = false
                invokeHook(options.onOpen)
              }
              invokeHook(options.onConnect, {
                async send(data) {
                  if (!live) {
                    throw new WebSocketClientError(
                      'WebSocket is not open: this connection has ended',
                    )
                  }
                  return sender.send(data)
                },
              })
```

packages/ws-client/src/websocket.ts
```diff
@@ -0,0 +1,379 @@
...
+              opened = true
+              const scoped: Sender<M> = {
+                async send(data) {
+                  if (!live) {
+                    throw new WebSocketClientError(
+                      'WebSocket is not open: this connection has ended',
+                    )
+                  }
+                  return sender.send(data)
+                },
```
matthieusieben: This check is already performed inside the transport itself:


```suggestion
                async send(data) {
                  return sender.send(data)
                },
```

Thanks to this we can actually avoid double async function wrapping:

```suggestion
                // @NOTE we avoid creating an additional
                // function here to avoid nested async wrappers
                send: sender.send.bind(sender),
```

We can actually go one step further and just use the `sender` here as hook argument. I am not worried about someone fiddling with non documented properties (async iterable method).

packages/ws-client/src/websocket.ts
```diff
@@ -0,0 +1,379 @@
...
+            },
+            onClose: (detail) => {
+              closeDetail = detail
+              reportClosed()
+              // Stop accepting writes on the scoped sender *now*. Waiting for
+              // the loop to notice would leave a window — the loop only advances
+              // when the consumer pulls — in which a hook holding the sender
+              // could hand data to a dead socket and see it silently dropped.
+              const wasLive = live
+              invalidateSender()
```
matthieusieben: Looks like this can be inlined:

```suggestion
              live = false
```

packages/ws-client/src/websocket.ts
```diff
@@ -0,0 +1,379 @@
...
+              invokeHook(options.onConnect, scoped)
+            },
+            onClose: (detail) => {
+              closeDetail = detail
+              reportClosed()
+              // Stop accepting writes on the scoped sender *now*. Waiting for
+              // the loop to notice would leave a window — the loop only advances
+              // when the consumer pulls — in which a hook holding the sender
+              // could hand data to a dead socket and see it silently dropped.
+              const wasLive = live
```
matthieusieben: Looks like this is always true, unless `onClose` is called more than once (which should never be the case) ?

Looks like the only effect this has is to prevent `onDisconnect` from being called more than once, which could be achieved using better named variables;


```ts
let onDisconnectInvoked = false
let onOpenInvoked = false

// ...
  
  onOpen: () => {
    onOpenInvoked = true
    invokeHook(options.onOpen)
  },
  onClose: () => {
    // ...
  
    if (onOpenInvoked && !onDisconnectInvoked) {
      onDisconnectInvoked = true
      invokeHook(options.onDisconnect)
    }
  }
```

Though more verbose, this makes all the comments here (about calling in "pair", etc.) un-necessary.

packages/ws-client/src/websocket.ts
```diff
@@ -0,0 +1,379 @@
...
+          // an unclean end rejects.
+          const wasClean = error instanceof CloseError && error.wasClean
+          if (!willReconnect && wasClean) return
+
+          invokeHook(
+            options.onError,
+            error,
+            willReconnect ? { attempt: retries } : undefined,
+          )
+          if (!willReconnect) throw error
```
matthieusieben: would you see any objection to making this two distinct hooks?

```suggestion
          if (!willReconnect) {
            invokeHook(options.onError, error)
            throw error
          } else {
            invokeHook(
              options.onReconnect,
              error,
              { attempt: retries },
            )
          }
```

One of the reason is that the `onError` hook's error will also be thrown by the async iterator, so it can be implemented in two different ways by the caller, while the `onReconnect` has different semantics:

```ts
try {
  for await (const _ of websocket({
    onReconnect: () => {
      // Will only ever trigger this
    },
    onError: () => {
      // Will trigger this
    },
  }) {
    
  }
} catch (err) {
  // And also this
}


```

packages/ws-client/src/websocket.ts
```diff
@@ -0,0 +1,379 @@
...
+    } finally {
+      // The single terminal transition: runs however the stream ends, exactly
+      // once. A generator that was never pulled never gets here.
+      //
+      // On a polite close the transport's close event lands just after the
+      // generator unwinds, so wait for it (bounded) rather than synthesizing a
+      // second answer. That means "our end is closed", not "the peer
+      // acknowledged" — waiting on the peer could hang teardown forever.
+      if (closeDetail === undefined && awaitClose !== undefined) {
+        await Promise.race([awaitClose, sleep(CLOSE_GRACE_MS)])
```
matthieusieben: I think we should not use `Promise.race` here.

`Promise.race` is a dangerous thing. It can lead to resource exhaustion (I think we may have encountered this behavior with aws's S3 client):

If a client maintains a pool of connection (to keep resource usage in check), this `race` hides the fact that, under the hood, many connections may still be open, and the client would create new once without knowing it.

Instead we should either find a way to forcefully cleanup the entire WebSocket related resources (at the "transport" abstraction level), or provide a hook that the user can use to detect such cases and determine on their own if it's safe to keep this dangling resource. In any case, there should be a hook that only gets called when the work is actually done.

I think we should add a `closeTimeout` at the transport level, that triggers a `terminate()` if the close callback is not triggered fast enough.

matthieusieben (commented):
This is definitely simpler to read ! Left a few comments but looks good overall.

packages/ws-client/src/transport/node-transport.ts
matthieusieben: See my other comments that touch on this;

1) While convenient to be able to control the code by either throwing or aborting the signal using a `CloseError`, other abort reason should not be considered errors
2) `ws.terminate()` should be used to implement a `closeTimeout` and forcefully free resources (instead of the `Promise.race`).

packages/ws-client/src/websocket.ts
matthieusieben: (nit, sorry)

packages/ws-client/src/transport/transport.ts
devinivy: There is just one buffer, and that is for received messages that are being pulled via async iteration.  So this is a readable high water mark.  The main purpose of the library is to be robust for long-lived streams, and this is maybe the key part of the backpressure / flow control story.

I'd propose we ensure it has a good tsdoc but not rename it, particularly given that there is just one option and not two.  I think that node uses `readableHighWaterMark` and `writeableHighWaterMark` in two ways: a. for `Duplex` stream options because they have two buffers that need to be distinguished, b. on the stream properties because they are avoiding clashes between `Readable`, `Writable`, and `Duplex`.

packages/ws-client/src/transport/transport.ts
devinivy: I do think that it's fairly intuitive that these are for readable bytes, but we should have a proper tsdoc for it regardless given that it is in the public API 👍

packages/ws-client/src/transport/node-transport.ts
devinivy: I kind of prefer how it's written now, do you think it's worth the microtick?

packages/ws-client/src/websocket.ts
devinivy: Ah yeah I had this change but hadn't pushed it!

packages/ws-client/src/websocket.ts
devinivy: I've gotten rid of `live` entirely, will push.  I did a little cleanup pass I forgot to push 🙃 sorry!

packages/ws-client/src/websocket.ts
devinivy: I'm good with that.

packages/ws-client/src/websocket.ts
devinivy: I am already working this morning on getting rid of it.  It turns out the latest version of `ws` has a `closeTimeout`, so I am just going to handle it in the transport.  It's not an issue on browser since all the APIs are "fire and forget" anyway, and the browser manages the underlying resources.

packages/ws-client/src/websocket.ts
devinivy: I was able to get rid of `live` entirely.  I will take this into account in another pass.

packages/ws-client/src/websocket.ts
devinivy: I don't think so.

packages/ws-client/src/websocket.ts
devinivy: I figured this may be a common/natural use case, to essentially just want to use the library for a single connection (i.e. like `WebSocket`).

packages/ws-client/src/websocket.ts
devinivy: Not really, but I also don't have a particular goal of using type inference as much as possible.

packages/ws-client/src/transport/node-transport.ts
matthieusieben: No strong opinion here!

packages/ws-client/src/websocket.ts
matthieusieben: Smaller code footprint and fewer ai tokens spent 😋

packages/ws-client/src/transport/transport.ts
devinivy: I think the typo in that case would cause a type error and jump out at you, but in any case point taken 👍 The missing type possibility is real, especially here where it is wired up through the public API.

matthieusieben: I've recently come across this in OAuth provider's hooks where no error was returned. I can reproduce it if you want.

matthieusieben: But I don't think it was a function call so maybe you're right

packages/ws-client/src/transport/node-transport.ts
devinivy: It kind of is, but I also appreciate being able to control the exact semantics of this library rather than map them onto `createWebSocketStream()` so I am not too sad about it.  I think it could help the nodejs vs. browser behaviors be more consistent because they have the same machinery and the transport implementations are extremely similar.

packages/ws-client/src/websocket.ts
matthieusieben: Fewer brain tokens too

packages/ws-client/src/websocket.ts
devinivy: I like moving to `AsyncGenerator<MessageOf<M>, void, unknown>` but I don't really want to split the types like that.  We don't want the user to pass anything to `next()` because it is not meaningful usage of the library, but there is not a great way to limit it, and there is very little chance that we will read those values by mistake.  So let's go to `unknown` and keep a single type for now 👍

packages/ws-client/src/transport/transport.ts
devinivy: Added 👍

packages/ws-client/src/websocket.ts
devinivy: Done 👍

packages/ws-client/src/transport/browser-transport.ts
devinivy: Done 👍

packages/ws-client/src/transport/node-transport.ts
devinivy: The impetus was to provide _some way_ of triggering an immediate termination (e.g. `ws.terminate()`).  I think I agree that it may be surprising to abort with a custom error and have an unclean teardown, so I updated the behavior.

Now there is no way to forcibly teardown the socket, which isn't ideal to me but I am not too hung up on it and maybe we can find a path later.

packages/ws-client/src/websocket.ts
devinivy: Done 👍

packages/ws-client/src/websocket.ts
devinivy: Done 👍

packages/ws-client/src/websocket.ts
devinivy: Done 👍 The future is now.

packages/ws-client/src/websocket.ts
devinivy: Handled, see also comment above 👍

packages/ws-client/src/websocket.ts
devinivy: Sure, I've inlined them.

packages/ws-client/src/websocket.ts
devinivy: This is done.  `awaitClose` is also gone, and socket teardown has been pushed up into the transport.  So by the time `yield* transport` completes, the socket is closed and the transport's `onClose` has fired.  And the timeout has been pushed up into transport's socket teardown.  Right now the timeout is hardcoded, and we can expose it in the future if desired.

devinivy:
Thanks @matthieusieben, I addressed this feedback 👍 

Found one weird thing along the way, which is that in node v22 undici's `WebSocket` has a bug where it sometimes it emits `'error'` without a `'close'`, generally when closing the socket before it has fully opened.  Sometimes it even emits two `'error'`!  So I included a small workaround in the browser transport.  Node v24 does not have the same issue.  I also tried to document more explicitly each transport's close semantics.

packages/ws-client/src/transport/transport.ts
devinivy: Done 👍

packages/ws-client/src/websocket.ts
devinivy: This whole area looks a little different now, but I think you'll find this to be resolved.

packages/ws-client/src/websocket.ts
devinivy: Done 👍

packages/ws-client/src/transport/node-transport.ts
```diff
@@ -0,0 +1,287 @@
...
+      // event. The 'close' handler above is what settles the channel, so a
+      // consumer awaiting the end of iteration is awaiting the socket.
+      ws.close(closeCodeForStop(options.signal.reason))
+    },
+    { once: true },
+  )
+
+  return {
+    send: sender.send,
+    [Symbol.asyncIterator]: () => closeGuard(channel.iterable, closed),
```
matthieusieben: This is an internal API so it might not matter too much but I think that this should be internal to the channel implementation:

```ts
const closeGuard = createCloseGuard()
const channel = createMessageChannel({
  // ...

  // Optional
  closeGuard,
})

const markClosed = () => closeGuard.markClosed()
```

Or, alternatively, make the channel need to be explicitly closed:

```ts
  const channel = createMessageChannel({
    // ...
  })

  const markClosed = () => channel.notifyClosed()
```

matthieusieben (commented):
I've created a PR https://github.com/bluesky-social/atproto/pull/5289 that serves as review suggestions (I've reviews this from VSCode).

The main issue in the current implementation are:

- The iterator has no `throw` member, which can lead to resource leak (edit: looking a clause's review I'm wrong here. TIL)
- The `closeGuard` utility would not await the promise if the inner iterator's `return`/`throw` throws.
- The `closeGuard` utility has a `throw` function, but if the inner iterator doesn't, the inner iterator is never `return()`ed, causing a leak.
- It is possible to get invalid types in user-land by doing `websocket<'text'>('url')`

packages/ws-client/src/transport/node-transport.ts
matthieusieben: Note that this was changed in my PR.

packages/ws-client/src/transport/node-transport.ts
matthieusieben: I'd be fine having a custom error class for that purpose ! I just didn't like the "unknown reason = termination". I have nothing against being able to forcefully terminate:


```ts
import {WebSocketTerminationReason} from '@atproto/ws-client'

ac.abort(new WebSocketTerminationReason())



```

PR #5289, opened 2026-07-31
#5285 review

matthieusieben:
Serves as review suggestion for #5285 

Here are the motivations for some of the changes:

## Improvements

**`websocket` function interface**

Removed `WebSocketFn` in favor of a self defined export function

Before:
Capture d’écran 2026-07-31 à 10 16 27

After:
Capture d’écran 2026-07-31 à 10 23 17

**protection against invalid call**

Before:
Specifying a generic would invalidly cast the return type

After:
Capture d’écran 2026-07-31 à 10 48 00

**Exposes `HEADERS_SUPPORTED_PLATFORM`**

Allows consumers to dynamically detect support for headers.

**`Uint8Array<ArrayBuffer>` instead of `Uint8Array<ArrayBufferLike>`**

## Fixes

- ~Iterators should be cleaned up when `throw()` is called~ (not accurate, `return()` should be called when `throw()` is not defined.)
- `guarded` would not always await the promise (e.g. if the inner iterator's `return()` throws)
- `WHATWGWebSocket` is now compatible with the `dom.lib.ts` definition (no more type cast needed)
- `timer.unref?.()` should not be needed since the lifetime of timers is tied to the channel, and both should be cleaned up together (there is no reason why the timers would still run, but not the channel)

## Aestetical changes

- Removed a few un-necessary function nesting (`() => iterable[Symbol.asyncIterator]()`, `(error) => defaultShouldReconnect(error)`, `() => pull()`)
- `closeGuard` now uses an `AbortSignal` instead of a `Promise`. This was mostly the result of intermediary work and can be reverted.

### DIFF
--- a/packages/ws-client/src/index.ts
+++ b/packages/ws-client/src/index.ts
@@ -1,6 +1,11 @@
-import { createTransport } from '#transport'
-import { createWebSocket } from './websocket.js'
-import type { WebSocketFn } from './websocket.js'
+import { HEADERS_SUPPORTED, createTransport } from '#transport'
+import type { DataMode } from './message-channel.js'
+import {
+  type Awaitable,
+  type WebSocketIterable,
+  type WebSocketOptions,
+  websocketFactory,
+} from './websocket.js'
 
 export type {
   CloseEventDetail,
@@ -16,9 +21,6 @@ export type {
 export type { HeadersInit, Sender } from './transport/transport.js'
 export type {
   Awaitable,
-  BrowserWebSocketOptions,
-  NodeWebSocketOptions,
-  WebSocketFn,
   WebSocketIterable,
   WebSocketOptions,
 } from './websocket.js'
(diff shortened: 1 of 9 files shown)


devinivy (approved):
Thanks, these changes look good to me.  Worth a quick peek at Claude's review above re: the `@NOTE`s on the new work around `throw()`.

While I would accept this PR as-is, I am a little torn on the `throw()` machinery— it is not code that will really get used, and I think it's honest to end up with a cleanup+`TypeError` if a consumer tries to use the generator that way.  I am ambivalent about the new `AbortSignal`, all good either way to me.  And the other changes I am +1 on, clear Ws!

Since these are effectively just review suggestions, I would rather not land these as two commits w/ stacked PRs, I think it makes more sense to just land this work in the feature branch if possible.  But also not a major deal.

If you are happy with the state of the work in #5285 with these changes feel free to merge all of it!

PR #5293, opened 2026-07-31
:sparkles: Add recommended policy config in queue

foysalit:


packages/ozone/src/queue/service.ts
```diff
@@ -67,6 +69,29 @@ export class QueueService {
     return (db: Database) => new QueueService(db)
   }
 
+  async assertRecommendedPolicies(
```
gcwill70, line 72: One potential issue is if a policy is deleted. This check would block the queue from being updated.

Probably should update `tools.ozone.setting.removeOptions` to delete policy keys here or make this check more flexible?

gcwill70 (approved):

packages/ozone/src/queue/service.ts
foysalit: this is intended. policies are always loosely associated so we allow deleting a policy without affecting other parts of the app (for the most part). this forces admins to take notice when it happens and consciously disassociate the policy from queue on next update.

PR #5296, opened 2026-08-01
i18n: Add plural formatting to relevant strings + fix string concatenation issues

nilaallj:
## What & why

After translating the OAuth provider UI to Swedish, I noticed the time indicators for apps and devices were concatenated strings I could not translate in a way that made the complete strings grammatically correct in all contexts. I also noticed they did not have proper plural formatting, which seemed to be one of the reasons for the string concatenation. 

This PR suggests making complete time indicator strings for `devices/page.tsx` and `apps/page.tsx` with proper plural formatting. It also suggests fixes for other strings I found with similar issues.

**I used Claude Sonnet 4.6 for assistance.**

### `Apps` and `Devices` pages: Interpolated “time ago” strings with plural formatting

- Moved all i18n responsibilities from `use-date-ago.tsx` to the call sites `devices/page.tsx` and `apps/page.tsx`.
- Deleted `date-ago.tsx` as it was left without any call sites.
- Lingui will extract complete strings such as `Last accessed {0, plural, one {# minute} other {# minutes}} ago`, instead of strings such as `1 minute` and `{0} minutes` separated from the current context.

### Plural formatting for languages with multiple plural forms

For some strings adding plural formatting doesn’t make sense in English since the singular form would never be used. But in some languages the plural form varies depending on the number. Therefore I added plural formatting to three strings which will look like this:

- `Between {minLength} and {maxLength, plural, other {# characters}}` (from `input-handle-default.tsx`)
- `Password with at least {MIN_PASSWORD_LENGTH, plural, other {# characters}}` (from `input-new-password.tsx`)
- `Please wait {remainingSeconds, plural, one {# second} other {# seconds}} before trying again.` (from `button-cooldown.tsx`, it also seemed reasonable to add the singular form here)

Lingui allows adding plural formatting without a singular form, since the `other` form is used as a fallback when no other form fits. With this, translators can add plural forms such as `few` and `many` if needed.

### Fix concatenation of sign-up disclaimer string

Currently, the string `By creating an account you agree to the Terms of Service and the Privacy Policy of this service.` provided by `sign-up-disclaimer.tsx` is concatenated with `Terms of Service` and `Privacy Policy` cut out. This is problematic because the cut out strings are also used in other parts of the UI where they are not a part of a sentence.

The concatenation happens because the cut out strings are optionally wrapped in an anchor. The suggested solution adds a helper component (`ConditionalLink`) which makes it possible to inline the link texts in a complete message.

## Checklist

- [x] `pnpm build:ui && pnpm test` passes
- [x] Extracted strings look as intended
- [x] Strings work as intended in the UI running dev env


packages/oauth/oauth-provider-ui/src/pages/account/(authenticated)/devices/page.tsx
```diff
@@ -79,7 +80,31 @@ function AccountSessionCard({
...
+        })} ago`
+      case 'yesterday':
+        return t`Last seen yesterday`
+      case 'days':
+        return t`Last seen ${plural(dateAgo.count, {
+          one: '# day',
+          other: '# days',
+        })} ago`
+    }
+  })()
```
matthieusieben: Would you mind refactoring this into a new hook:

```ts
  const lastSeenText = useLastSeenText(lastSeenAt)
```

Make sure to add a comment in that hook explaining that it's structure is needed for proper transatability in some languages (SW).

packages/oauth/oauth-provider-ui/src/pages/account/(authenticated)/apps/page.tsx
```diff
@@ -99,6 +100,30 @@ function ApplicationSessionCard({
...
+        })} ago`
+      case 'yesterday':
+        return t`Last accessed yesterday`
+      case 'days':
+        return t`Last accessed ${plural(dateAgo.count, {
+          one: '# day',
+          other: '# days',
+        })} ago`
+    }
+  })()
```
matthieusieben: Would you mind refactoring this into a new hook:

```ts
  const lastAccessedText = useLastAccessedText(lastSeenAt)
```

Make sure to add a comment in that hook explaining that it's structure is needed for proper transatability in some languages (SW).

matthieusieben (changes requested):
This looks good! In addition to the comments bellow, and fixing the `pnpm verify` output, could you add fr translations (use google translate or an LLM) ?

matthieusieben (approved):
Thank you 🙏

matthieusieben:
Looks like there is a merge conflict still

matthieusieben (approved):
Thank you

PR #5298, opened 2026-08-02
chore: upgrade to ESLint 9.x

43081j:
This upgrades ESLint to 9.x and switches to flat config (`eslint.config.mjs`). As part of that, unused ignore comments are flagged by default now and that is why a bunch of sources slightly changed.

I was digging around this repo for a fork of mine and did this switch while i was in there. let me know if you want any changes or have reason to not do it yet.

## Checklist


package.json
```diff
@@ -23,7 +23,7 @@
   },
   "scripts": {
     "lint:fix": "pnpm lint --fix",
-    "lint": "NODE_OPTIONS=--max_old_space_size=4096 eslint . --ext .ts,.js,.tsx,.jsx",
+    "lint": "NODE_OPTIONS=--max_old_space_size=4096 eslint .",
```
matthieusieben: I'd like to see if this allows to drop this (was added due to failing on CI)


```suggestion
    "lint": "eslint .",
```

package.json
```diff
@@ -89,5 +91,6 @@
       "packages/oauth/*",
       "packages/internal/*"
     ]
-  }
+  },
+  "packageManager": "pnpm@8.15.9+sha512.499434c9d8fdd1a2794ebf4552b3b25c0a633abcee5bb15e7b5de90f32f47b513aca98cd5cfd001c31f0db454bc3804edccd578501e4ca293a6816166bbd9f81"
```
matthieusieben: This is not right.

`package.json` already defined this under `devEngines`:

```json
    "packageManager": {
      "name": "pnpm",
      "version": "11.11.0",
      "onFail": "error"
    }
````

matthieusieben (commented):
This is great, and something we really appreciate! However, if we are going to update eslint, we should adopt the latest version (10.x). Any reason why this only updates to 9.x ?

matthieusieben:
Note you can create an empty changeset with `pnpm changeset add --empty`

43081j:
> This is great, and something we really appreciate! However, if we are going to update eslint, we should adopt the latest version (10.x). Any reason why this only updates to 9.x ?

not all plugins support 10.x yet (e.g. the `import` plugin doesn't).

so my plan was going to be this:

1. update to ESLint 9.x (which all the existing plugins support)
2. switch to [eslint-plugin-import-x](https://github.com/un-ts/eslint-plugin-import-x/) (faster/lighter than `import` and handles resolution a bit better)
3. update to ESLint 10.x

package.json
43081j: reverted 👍  i use corepack locally and it added this (corepack and pnpm don't agree)

matthieusieben:
Sounds like a good plan 👍

PR #5301, opened 2026-08-03
chore: upgrade to ESLint 10.x

43081j:
This updates to ESLint 10.x, and switches to `eslint-plugin-import-x` as
part of that.

The new import plugin has better module resolution than the old one, so
now picks up where we were using default imports for modules which
declare having no such export.

NOTE: i can't run all the tests locally, so we should see if node handles those
CJS export assignments correctly after switching to named imports. I imagine
they were originally using default imports because of that.

## Checklist


matthieusieben (commented):
> NOTE: i can't run all the tests locally, so we should see if node handles those
CJS export assignments correctly after switching to named imports. I imagine
they were originally using default imports because of that.

These were indeed introduced as part of the ESM migration because for some reason the nodejs interop was not working properly. I've tested a few locally (`http-terminator`, `mime-types`) and it seemed fine. The fact CI passes reassures me too.

Since these *do* introduce a new runtime behavior, we should issue a (patch) `changeset` for every affected package.

Could you please remove `.changeset/red-cameras-draw.md` and run `pnpm changeset` (it should pick up all the updated packages automatically)

matthieusieben (approved):

PR #5306, opened 2026-08-04
Adopt TypeScript 7 stable and tune editor/lint config

devinivy:
TypeScript 7 now ships under the standard `typescript` package, with the compiler binary back to `tsc` ([announcement](https://devblogs.microsoft.com/typescript/announcing-typescript-7-0/)). This replaces `@typescript/native-preview` with `typescript` 7.0.2 and updates every `tsgo --build` script to `tsc --build`. Compiler behavior is unchanged — it's the same native compiler under its permanent name. Note that TS7 has no stable programmatic API until 7.1; nothing in the repo uses one, and the linter brings its own parser.

After pulling this, a plain `pnpm run build` from the root regenerates any stale codegen; `build:ts` alone may fail on stale generated files until then.

Also included, all editor/tooling-only:
- Zed and ESLint now exclude `.claude/worktrees`, keeping agent worktrees out of file scanning, search, and repo-wide lint runs.
- Zed uses the tsgo extension's language server exclusively (now registered as `typescript-ls`, per the [extension README](https://github.com/zed-extensions/tsgo)), disabling `vtsls` and `typescript-language-server` so only one server runs.
- TSX gets the same language-server and prettier-via-eslint formatting config that JavaScript and TypeScript already had.

No runtime or published-package changes, so no changeset.


package.json
```diff
@@ -77,6 +77,7 @@
     "prettier": "^3.9.6",
     "prettier-config-standard": "^7.0.0",
     "prettier-plugin-tailwindcss": "^0.6.11",
+    "typescript": "npm:@typescript/typescript6@^6.0.2",
```
matthieusieben, line 80: Edit: I now see that typescript-eslint needs access to TypeScript’s programmatic API, and this is the [recommended workaround](https://devblogs.microsoft.com/typescript/announcing-typescript-7-0/#running-side-by-side-with-typescript-6.0).

matthieusieben (approved):
Nice.

PR #5314, opened 2026-08-05
fix(pds): receiving `InvalidRequest` instead of `HandleNotFound`

### LINKED ISSUE #5313: `resolveHandle` gives `InvalidRequest` instead of `HandleNotFound`
**Describe the bug**

When a handle cannot be resolved by `com.atproto.identity.resolveHandle` (there is no account with has the handle) it seems to give `InvalidRequest` value in `error` property of response body instead of `HandleNotFound`. [It also seems that `HandleNotFound` is never used at all in code](https://github.com/search?q=repo%3Abluesky-social%2Fatproto+HandleNotFound&type=code), despite [lexicon suggesting it may be given as error code in `resolveHandle`](https://github.com/bluesky-social/atproto/blob/main/lexicons/com/atproto/identity/resolveHandle.json#L32) and the same being in [API reference docs](https://endpoints.bsky.app/#bluesky-app/tag/comatprotoidentity/GET/xrpc/com.atproto.identity.resolveHandle).

The current response when the handle cannot be resolved is:

```json
{
    "error": "InvalidRequest",
    "message": "Unable to resolve handle"
}
```

**To Reproduce**

Through Bluesky:

1. Open https://bsky.app/profile/example.com
2. Open devtools (Firefox, Chrome, whatever)
3. Find `com.atproto.identity.resolveHandle` request
4. Check response body of the request

With local PDS:

1. Open /xrpc/com.atproto.identity.resolveHandle?handle=example.com route of the PDS in browser or any other handle that is unused (e.g., `abc.test` if there is no account with `abc.test` handle).

**Expected behavior**

Based on the lexicon, the following response body JSON should be given when handle cannot be resolved:

```json
{
    "error": "HandleNotFound",
    "message": "Unable to resolve handle"
}
```

**Details**

- Operating system: CachyOS x86_64, Linux 7.1.5.1
- Node version: v26.4.0

**Additional context**

It is unclear whether that is something that was not removed from lexicon or not properly implemented?

IdkGoodName:
## What & why

Fixes #5313, where `InvalidRequest` error is thrown instead of `HandleNotFound` when using `resolveHandle` method/route.

## Checklist


packages/pds/tests/handles.test.ts
```diff
@@ -75,6 +75,13 @@ describe('handles', () => {
     })
     expect(res.data.did).toBe(alice)
   })
+  it('does not resolve unknown handles', async () => {
+    const rejection = agent.api.com.atproto.identity.resolveHandle({
+      handle: 'john.test',
+    })
+    await expect(rejection).rejects.toThrow('Unable to resolve handle')
+    await expect(rejection).rejects.toHaveProperty('error', 'HandleNotFound')
```
ThisIsMissEm: We'd normally do:

```suggestion
    const promise = agent.api.com.atproto.identity.resolveHandle({
      handle: 'john.test',
    })
    await expect(promise).rejects.toMatchObject({
      error: 'HandleNotFound',
      message: 'Unable to resolve handle',
    })
```

ThisIsMissEm (commented):
I think this looks correct, though question for @matthieusieben: should this actually be proxying up to the entryway rather than doing appview/HandleResolver?

matthieusieben (approved):
This looks correct to me. We should probably also tackle the behavior of the `if (ctx.cfg.entryway)` branch but that can be done as a separate PR.

Could you please add a changeset so that CI can pass ? Thank you.

ThisIsMissEm:
@matthieusieben changeset added 👍🏻

matthieusieben (approved):

PR #5340, opened 2026-08-06
chore: remove typed-emitter

43081j:
In modern Node types, `EventEmitter` is already generic. This means we
can drop `typed-emitter` and use `EventEmitter<T>` directly.

For example:

```ts
class MyEmitter extends EventEmitter<{ myEvent: [Arg1Type, Arg2Type] }> {
  // ...
}
```

## Checklist

### DIFF
--- a/.changeset/perky-mice-visit.md
+++ b/.changeset/perky-mice-visit.md
@@ -1,0 +1,9 @@
+---
+'@atproto/lex-cli': patch
+'@atproto/bsync': patch
+'@atproto/ozone': patch
+'@atproto/bsky': patch
+'@atproto/pds': patch
+---
+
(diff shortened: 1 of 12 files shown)


matthieusieben (approved):
💙

43081j:
@matthieusieben there's quite a lot of these easy wins. It could be a good way for the e18e community to collaborate with bsky if I make an issue that lists them all, so people can pick them up individually. 

Would that make sense to you? I'm happy to drive by clean up every now and then but seems a good opportunity to get the community involved 😄

matthieusieben:
Let's do that! Thanks for helping us get the repo in a better state!

PR #5347, opened 2026-08-07
Update oauth provider ui routing

bigmoves:
## What

Move the account manager's authentication and selected-account state into the
router instead of in-component state, and put the routing itself on TanStack
Router's file-based conventions.

- Auth becomes real routes: `/account/sign-in`, `/account/sign-up`.
- The active account is a path segment: `/account/u/<handle-or-did>/…`.
- Routing is file-based. `src/routes/` is the tree and `@tanstack/router-plugin`
  generates `src/routeTree.gen.ts` from it, with `autoCodeSplitting` giving each
  page its own chunk. The generated file is committed and excluded from
  formatting.
- Access checks run in `beforeLoad` and `throw redirect()` before a page
  renders. `/account` decides where the device's sessions belong, and
  `/account/u/$accountId` resolves the account the URL names, returning it so
  descendants read it from route context instead of resolving it again.
- Router context carries what the guards and loaders need but can't reach
  through React: the session store, the API client and the query client.
- The devices and apps pages load through the route's `loader`
  (`queryClient.ensureQueryData`) and read the same query with
  `useSuspenseQuery`, so they carry no in-component loading state.

## Why

Previously the account manager ran a hand-rolled `#step=` view machine *alongside*
TanStack Router (effectively two routers). On reload, a device with several
remembered accounts dropped back to the account picker. Now the URL is the single
source of truth, so a reload restores both the sub-page and the selected account.

This builds on @matthieusieben's #5357, which broke the hand-rolled route module
into one file per route. Going the rest of the way to file-based routing removes
the hand-assembled tree — the place where a subtree could register its children
but not itself and silently 404 every path beneath it — and moving the guards
into `beforeLoad` removes a render-then-redirect window where a page mounted and
briefly fetched as the wrong account.

Two things that are easy to get wrong, both documented in the package's
`CLAUDE.md`:

- The session state reaches the router as an object of **getters**, not a render
  snapshot. A guard runs during a navigation that is typically dispatched from
  the same event handler that just signed the user in, so a captured value would
  be one render stale and the guard would reject the account it was just sent to.
- Changing the router context does **not** re-run the guards. The entry
  invalidates the router when the sessions change, which is what lets signing out
  move the user with no navigation wired into the action.

The third-party **consent** flow keeps its existing linear wizard (unchanged).

Dependencies: `@tanstack/react-router` `1.115 → 1.170.27`, plus
`@tanstack/router-plugin` (codegen) and `@tanstack/react-router-devtools`
(dev-only; `import.meta.env.DEV`-gated, so it is stripped from the production
bundle).

Also: the dev mock now serves the account page for every path under `/account`,
as the PDS does, so deep links and refreshes reach the router — without that,
the thing this PR is about could not be exercised locally.

## Reviewer note

This drops `buildRoutes(path, customPages)`, the (unused) hook for a deployment
to add or replace account-manager pages without forking the package. It went in
#5357; flagging it so it's a decision rather than an oversight.

## Tests

`account-manager` e2e 13/13, `oauth` (consent) e2e 7/7, unit 94/94, `tsc` clean
repo-wide, eslint + prettier clean.

> The devices and apps pages have no e2e coverage and are the pages that changed
> most (loader + `useSuspenseQuery`), so they were checked by hand; their error
> path could not be exercised, because the dev mock has no way to fail a request.

> The experimental popup/webview embedding mode is preserved but has no e2e
> coverage in-tree.

## Checklist

- [x] Written with the help of an LLM or coding agent? Say so here, and name the
      tooling. — **Claude Code (Opus 5)**


matthieusieben (changes requested):
My main issue with this is that the type checking is being disabled in many places (`as any`, `as never`, ...).

I think the part of the cause is that some new routes (`sign-in`, `sign-up`) are not being built from the right place: they should be build in a similar fashion to the `reset-password` route. This definitely should be addressed.

It **should** also be possible to keep strong typings in every dynamic routes (as is already the case right now). Inlining the `/account` base path might make things easier.

I initially created the dynamic routing system to be able to override routes from consumer packages (ie. to offer an easy way to provide a custom `oauth-provider-ui`). But this might have been over-engineered and a simpler/better solution might be to actually get rid of that altogether (which I tried to achieve [here](https://github.com/bluesky-social/atproto/pull/5357), but didn't finish).

packages/oauth/oauth-provider-ui/src/routeTree.gen.ts
```diff
@@ -0,0 +1,263 @@
+/* eslint-disable */
```
matthieusieben, line 1: should this be referenced as `linguist-generated=true` from the root `.gitattributes` ?

matthieusieben (approved):
Looks great

matthieusieben:
@claude do your thing

matthieusieben:
@claude Read .github/claude-review-prompt.md in the checked-out repo and review this pull request (5347) following its guidance.

Issue #5341, opened 2026-08-06
Code modernisation (umbrella issue)


43081j:
This is just to track opportunities where we can either:

- Modernise the code
- Remove dependencies which have native equivalents
- Use newer capabilities of the existing dependencies
- Improve supply chain security (e.g. deprecated dependencies, CVEs, etc)

| Package | What? | Why? | PR |
| -- | -- | -- | -- |
| N/A | Remove `typed-emitter` | `@types/node` has `EventEmitter<T>` | #5340 |
| pds | Upgrade `express` | Allows us to drop `express-async-errors` as 5.x does this natively | N/A  |
| pds | Remove `glob` | This is only used at dev-time, so we can use `fs.glob` instead (built-in) | #5343 |
| multiple | Remove `key-encoder` | We already pull in [@noble/curves](https://github.com/paulmillr/noble-curves) which can do the same thing | #5348  |
| oauth-provider | Upgrade `cookie` | We can then drop `@types/cookie` as it ships its own types now | N/A |
| multiple | Remove `get-port` | Rely on OS assigning a default | N/A |
| multiple | De-dupe instances of `zod` | Multiple versions exist in the repo | N/A |

will add more as i come across them

matthieusieben:
Regarding `get-port`, I want to get rid of it completely and have the OS assign a port (`listen(0)`), wherever possible.

dreyfus92:
hey everybody ✌🏻

i prototyped the express row locally. sharing what i measured and what the change actually involves, since it's a bit bigger than the table suggests.

**what it buys**

comparing the published dependency closures (`express@4.18.2` + `express-async-errors` vs `express@5.2.1`):

|  | before | after |
| -- | -- | -- |
| transitive packages | 69 | 66 |
| install size | 2.22 MB | 2.16 MB |
| `npm audit` | 7 vulns (3 high) | 0 |

dropped from the tree: `express-async-errors`, `array-flatten`, `destroy`, `methods`, `mime@1`, `safe-buffer`, `utils-merge`. added: `router`, `is-promise`, `once`, `wrappy`.

the real wins here are: 
- `express-async-errors` leaving the published graph, it's unmaintained (last publish 2019) and works by monkeypatching express/lib/router/layer internals at import time 
- moving off a maintenance-mode major, so transitives land on their current lines (`body-parser 2.x`, `path-to-regexp 8.x`, `send 1.x`). 

[...]

PR #5348, opened 2026-08-08
chore: replace key-encoder with native JWK import and `@noble/curves`

dreyfus92:
## What & why

drops the `key-encoder` dependency from `pds`, `bsky`, `aws` and `xrpc-server`, which also removes its whole transitive tree (`elliptic`, `asn1.js`, `bn.js`). 

`key-encoder` had exactly one job here: converting raw `secp256k1` key bytes into PEM so `node:crypto` could import them. but node can import keys directly from a JWK, which is just a JSON object with the key coordinates base64url-encoded, so the auth verifiers in pds/bsky now build KeyObjects that way, slicing the x/y coordinates out of the uncompressed point via `@noble/curves` (already in the tree via `@atproto/crypto`). 

the aws kms case gets its public key SPKI DER-encoded, and for secp256k1 the raw 65-byte point is always the trailing bytes of that structure, so we slice those and validate with noble's `ProjectivePoint.fromHex` instead of parsing the DER. tests build private keys as JWKs too, deriving the public coordinates explicitly with `secp256k1.getPublicKey()` since a private EC JWK requires them alongside `d` (the old PEM path did the same derivation, just hidden inside openssl). 

>[!NOTE]
behavior is unchanged for all valid inputs: same KeyObjects, same signatures, compressed and uncompressed keys both still accepted. the one difference is that garbage keys now fail at parse time with an on-curve check, which key-encoder never did.

related: #5341 

## Checklist


packages/aws/src/kms.ts
```diff
@@ -33,13 +26,12 @@ export class KmsKeypair implements crypto.Keypair {
...
-      Buffer.from(res.PublicKey),
-      'der',
-      'raw',
-    )
-    const publicKey = ui8.fromString(rawPublicKeyHex, 'hex')
+    // public key comes back SPKI DER-encoded, whose BIT STRING payload is the
+    // trailing 65-byte uncompressed point; fromHex validates the 0x04 prefix
+    // and that the point is on the curve
+    const publicKey = noble.ProjectivePoint.fromHex(
+      new Uint8Array(res.PublicKey).slice(-65),
```
43081j, line 36: What is the significance of-6 and 5 here?

May be worth a brief comment in the code on why

dreyfus92: it's a single argument, `.slice(-65)` takes the last 65 bytes. an uncompressed `secp256k1` point is always 65 bytes (0x04 prefix + 32-byte `x` + 32-byte `y`), and in the SPKI DER structure kms returns it's always the trailing bytes of the BIT STRING. i will expand the comment to spell that out.

43081j: weird i must be hallucinating 😂  i swear when i read this i saw two args. very strange

packages/bsky/src/auth-verifier.ts
```diff
@@ -472,25 +469,32 @@ export const buildBasicAuth = (username: string, password: string): string => {
   )
 }
 
-const keyEncoder = new KeyEncoder('secp256k1')
+const createSecp256k1KeyObject = (
+  publicKey: Uint8Array | string,
+): KeyObject => {
+  // accepts compressed or uncompressed keys; validates the point is on-curve
+  const point = secp256k1.ProjectivePoint.fromHex(publicKey).toRawBytes(false)
```
43081j, line 476:
```suggestion
  const point = secp256k1.Point.fromHex(publicKey).toRawBytes(false)
```

the readme suggests `ProjectivePoint` is the "old way"

dreyfus92: ah yeah 😄, the readme on main is for v2 where the rename to `Point` happened. this repo currently resolves `@noble/curves to 1.8.0` (same ^1.7.0 range `@atproto/crypto` already uses), and Point doesn't exist there yet, so `ProjectivePoint` is still the right name for the installed version. 

definitely worth doing the `v2` bump at some point though, i'd just rather keep this pr scoped to dropping key-encoder since the upgrade touches five packages. happy to follow up with it if that sounds good!

packages/pds/src/auth-verifier.ts
```diff
@@ -680,10 +677,19 @@ export const createSecretKeyObject = (secret: string): KeyObject => {
...
 }
 
-const keyEncoder = new KeyEncoder('secp256k1')
 export const createPublicKeyObject = (publicKeyHex: string): KeyObject => {
-  const key = keyEncoder.encodePublic(publicKeyHex, 'raw', 'pem')
-  return createPublicKey({ format: 'pem', key })
+  // accepts compressed or uncompressed keys; validates the point is on-curve
+  const point =
+    secp256k1.ProjectivePoint.fromHex(publicKeyHex).toRawBytes(false)
+  return createPublicKey({
```
43081j, line 684: given we do this in many places, i wonder if we should extract it to a shared function somewhere 🤔

43081j:
looks good to me!

nit: it may be worth giving the various magic numbers constants so they're easier to understand

matthieusieben (approved):
Looks good ! Thank you 💙

matthieusieben:
@claude do your thing

matthieusieben:
@claude Read .github/claude-review-prompt.md in the checked-out repo and review this pull request (5348) following its guidance.

PR #5366, opened 2026-08-11
Simplify the sign-up username step

bigmoves:
## Summary

The sign-up step that asks for a username puts the domain in a listbox inside
the text input. This layout has three problems:

- A listbox inside a text box is not a control that users know.
- The listbox covers the right end of the input. A click there does not put
  the cursor in the text.
- The width of the listbox changes with the length of the domain. A server
  with long domains gets a very small input, and a server with short domains
  gets a large one. The same field thus looks different on each deployment.

This PR puts the domain below the input. If the server has more than one
domain, the UI shows one radio row for each domain. If the server has only one
domain, the UI shows a preview of the resulting username. Radio rows have the
same width for all domains, and they show all of the options at the same time.

The step also had two rows of validation rules above the input. This PR
replaces the two rows with one hint. The hint keeps its position and changes
only its colour, thus the layout does not move while the user clicks.

The terms-of-service text moves to the last step of the wizard, because the
last step is the step that creates the account.

### More than one domain

| Before | After |
| --- | --- |
|  |  |

With a username:

### Only one domain

There are no radio rows. The preview gets a surface of its own, because it is
the only part of the step that shows the domain.

| Empty | With a username |
| --- | --- |
|  |  |

### Account manager

The dialog that changes the username of an account uses the same field.

## Notes for the reviewer

- The value of each radio is the domain. Base UI thus submits the correct
  value under the name `domain`, and no code maps an index to a domain. The
  layout with one domain keeps the hidden input.
- The names of the controls do not change. They are `handle` and `domain`, as
  before.
- The preview keeps the message id that the catalogs already contain, thus the
  five translations of that message stay correct.
- `atLast` is new in the render props of the wizard. The configuration of
  hCaptcha controls which step is the last one, thus the wizard supplies this
  value and the list of steps does not calculate it.

## Test plan

- `packages/pds`: `oauth.test.ts` gives 7 passed of 7.
- `packages/pds`: `account-manager.test.ts` gives 13 passed of 13.
- `packages/oauth/oauth-provider-ui`: `pnpm test` gives 94 passed of 94.
- `tsc --build`, `pnpm run build:ui`, ESLint and Prettier are clean.
- `pnpm i18n` adds one message and removes four. French is complete.

### DIFF
--- a/packages/oauth/oauth-provider-ui/src/components/forms/fields/handle-field.tsx
+++ b/packages/oauth/oauth-provider-ui/src/components/forms/fields/handle-field.tsx
@@ -1,3 +1,4 @@
+import { plural } from '@lingui/core/macro'
 import { Trans, useLingui } from '@lingui/react/macro'
 import { AtSignIcon } from 'lucide-react'
 import { type ReactNode, useMemo, useState } from 'react'
@@ -150,7 +151,13 @@ export function HandleField({
           !segment || valid ? 'text-muted-foreground' : 'text-destructive',
         )}
       >
-        {t`Use ${minLength}–${maxLength} letters, numbers or hyphens`}
+        {/* @NOTE The noun agrees with the end of the range, so the plural is
+          driven by `maxLength`. Locales with more than two plural categories
+          need the form even though the count is never one here. */}
+        {t`Use ${minLength}–${plural(maxLength, {
+          one: '# letter, number or hyphen',
+          other: '# letters, numbers or hyphens',
+        })}`}
       </p>
 
       {domains.length > 1 ? (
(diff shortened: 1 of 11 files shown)


matthieusieben (approved):
Nice !

packages/oauth/oauth-provider-ui/src/components/forms/fields/handle-field.tsx
```diff
@@ -125,54 +134,69 @@ export function HandleField({
...
+        never changes under the cursor mid-click — only its colour does. */}
+      <p
+        id="handle-hint"
+        className={cn(
+          'text-xs',
+          !segment || valid ? 'text-muted-foreground' : 'text-destructive',
         )}
-      </div>
+      >
+        {t`Use ${minLength}–${maxLength} letters, numbers or hyphens`}
```
nilaallj: The plural formatting is gone here.

bigmoves: fixed!

bigmoves: https://github.com/bluesky-social/atproto/pull/5366/commits/7e35515f119594620af0ff880f2cb9823f114234

PR #5372, opened 2026-08-11
Add content visibility lexicon

vineyardbovines:
## Summary

Enable account-level visibility discovery opt-in.

- add the singleton `app.bsky.actor.contentVisibilityDeclaration/self` record
- define the required `hideFromAlgorithmicRecommendations` boolean
- document that an absent record implies `false`
- add an `@atproto/api` patch changeset

### DIFF
--- a/.changeset/green-dogs-hide.md
+++ b/.changeset/green-dogs-hide.md
@@ -1,0 +1,5 @@
+---
+"@atproto/api": patch
+---
+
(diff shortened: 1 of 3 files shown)


rafaeleyng (approved):
Looks good. We usually call these things a "declaration" (we have a few `declaration.json` lexicons for record definitions). Maybe we could call this `contentVisibilityDeclaration`? Just to make sure it is in that same family of things.

But not a strong feeling about it.

PR #5376, opened 2026-08-11
Implement account preferences route in Ozone

foysalit:
## What & why

Implements app.bsky.actor.getPreferences in Ozone so authorized moderators can retrieve private preferences for a hosted account through the same NSID exposed by the PDS. This lets internal consumers use the existing Ozone service-signing authority rather than receiving the moderation service private key.

Ozone uses the modern schema route API so it can retain the internal, undocumented did query parameter while keeping that parameter out of the public Lexicon. It calls the PDS endpoint with a short-lived, method-bound service JWT. Redirects are disabled so that credential cannot be forwarded to another origin.

Depends on #5318.

## Testing

- Ozone integration test covering moderator -> Ozone -> PDS -> preferences through app.bsky.actor.getPreferences
- Ozone TypeScript package build
- targeted ESLint
- adversarial cross-model review: no findings

## Changeset

Adds a minor changeset for @atproto/ozone. No public Lexicon or @atproto/api change is introduced.

Written with the help of OpenAI Codex.


packages/ozone/src/api/app/bsky/actor/getPreferences.ts
```diff
@@ -0,0 +1,52 @@
+import { type LexValue, isDidString, l } from '@atproto/lex'
+import { XRPCError } from '@atproto/xrpc'
+import { type AuthResult, InvalidRequestError } from '@atproto/xrpc-server'
+import type { AppContext } from '../../context.js'
+import type { Server } from '../../lexicon/index.js'
+import { ids } from '../../lexicon/lexicons.js'
+
+const getPreferences = l.query(
+  ids.AppBskyActorGetPreferences,
+  l.params(),
```
matthieusieben: Since you are not using the "real" lexicon derived schema here, you can define your own schema:

```suggestion
  l.params({
    did: l.string({ format: 'did' })
  }),
```

packages/ozone/src/api/app/bsky/actor/getPreferences.ts
```diff
@@ -0,0 +1,52 @@
...
+      }
+
+      // @NOTE `did` is an internal moderator parameter omitted from the public
+      // Lexicon, so the generated client would reject it before sending.
+      const url = new URL(
+        '/xrpc/app.bsky.actor.getPreferences',
+        ctx.cfg.pds.url,
+      )
+      url.searchParams.set('did', did)
+      const auth = await ctx.pdsAuth(ids.AppBskyActorGetPreferences)
```
matthieusieben, line 30: You should still be able to use `agent.api.app.bsky.actor.getPreferences`, and add a `@ts-expect-error` where TS tells you that no `did` should be present.

gcwill70 (approved):

packages/ozone/src/api/app/bsky/actor/getPreferences.ts
```diff
@@ -0,0 +1,52 @@
...
+)
+
+export default function (server: Server, ctx: AppContext) {
+  server.xrpc.add(getPreferences, {
+    auth: async (authCtx): Promise<AuthResult> =>
+      ctx.authVerifier.modOrAdminToken(authCtx),
+    handler: async ({ params }) => {
+      if (!ctx.pdsAgent || !ctx.cfg.pds) {
+        throw new Error('PDS not configured')
+      }
```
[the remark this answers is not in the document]
foysalit: Addressed in 58e1dee5f: this now throws InvalidRequestError so the moderator client receives the actionable configuration message.

packages/ozone/src/api/app/bsky/actor/getPreferences.ts
foysalit: Addressed in 58e1dee5f: the private runtime schema now declares and validates the required did parameter, and the handler uses the typed value directly.

packages/ozone/src/api/app/bsky/actor/getPreferences.ts
foysalit: I tried this, but the current generated client cannot carry this undocumented parameter. QueryParams is {}, so TypeScript accepts { did } and an @ts-expect-error is unused; at runtime packages/xrpc/src/util.ts looks up schema.parameters.properties.did and throws Invalid query parameter: did before sending. I retained the raw request for this one call, with method-bound auth, redirects disabled, and a timeout. The inbound private schema now validates did per your other comment.

PR #5377, opened 2026-08-11
Show current account values on narrow viewports

bigmoves:
On the account management screen (`/account/manage`), each row's current value —
the email address, the username — sat in the trailing slot with
`hidden … sm:inline`, so below 640px it was not rendered at all. On a phone the
rows read `Email address ›` with no indication of what the setting was currently
set to.

The value now shares the label's line at every width. The label is `shrink-0`,
the value takes the space it leaves and ellipsizes.

| Before | After |
| --- | --- |
| Account rows at 390px with no values shown | Account rows at 390px showing the email address and username |

Desktop is unchanged:

| Before | After |
| --- | --- |
| Account rows at 900px | Account rows at 900px |

### `min-w-0`

`min-w-0` on both the content and the value is what makes the ellipsis work. An
email address has no break opportunity, so its automatic minimum size is the
whole string — without it a long address overflows the row and, because `Item`
is `flex-wrap`, drops the chevron onto a line of its own.

### Reading a truncated address

A very long address still truncates on a narrow viewport, and there was nowhere
to read it in full: the update-email dialog only carried the current address in
a hidden input for password managers. It now names it in the description.

Update your email dialog naming the current address

That surfaced the same overflow one level down — an address with no hyphens
produced 643px of content in a 288px dialog at 320px wide — so addresses quoted
in dialog copy now wrap. That was already latent in the verify-email and
delete-account dialogs, which use the same `<strong>{email}</strong>` shape, so
they are fixed too.

### i18n

One new msgid (`Your account currently uses <0>{emailCurrent}</0>. …`), French
filled in, none removed: 349 strings before and after.

### Testing

- `packages/pds/tests/account-manager.test.ts` — 13/13
- `packages/pds/tests/oauth.test.ts` — 7/7
- `packages/oauth/oauth-provider-ui` unit tests — 94/94
- Rendered at 320 / 390 / 641 / 900 with both a short address and a 61-character
  one, in English and with the longer French labels

### DIFF
--- a/.changeset/show-account-values-on-mobile.md
+++ b/.changeset/show-account-values-on-mobile.md
@@ -1,0 +1,5 @@
+---
+'@atproto/oauth-provider-ui': patch
+---
+
+Show the current email address and username on the account management screen at every viewport width. They were previously hidden below the `sm` breakpoint, so on mobile the rows gave no indication of what the setting was currently set to. The update-email dialog now names the address it is replacing, and email addresses quoted in dialog copy wrap instead of overflowing the dialog.
--- a/packages/oauth/oauth-provider-ui/src/components/delete-account-dialog.tsx
+++ b/packages/oauth/oauth-provider-ui/src/components/delete-account-dialog.tsx
@@ -136,8 +136,8 @@ export function DeleteAccountDialog({
         description={
           email ? (
             <Trans>
-              Check <strong>{email}</strong> for an email with the confirmation
-              code to enter below:
+              Check <strong className="break-words">{email}</strong> for an
+              email with the confirmation code to enter below:
             </Trans>
           ) : (
             <Trans>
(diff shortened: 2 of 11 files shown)


ThisIsMissEm:
I think this was also truncating values. It would be worth testing with longer strings and emails (e.g., french locale)

matthieusieben (approved):
nice !

bigmoves:
> I think this was also truncating values. It would be worth testing with longer strings and emails (e.g., french locale)

Yep, should account for that. Here's some screenshots:

account-home-320-fr
email-dialog-320-fr
handle-dialog-320-fr
manage-320-en-long
manage-320-fr-long
manage-900-en-long

ThisIsMissEm:
@bigmoves could the change handle get the same treatment with inlining the current value?

PR #5379, opened 2026-08-11
Expose canonical OP thread numbering on feed items

ds-boyce:
## What & why

Part two of two. Requires https://github.com/bluesky-social/tango/pull/1329.

## Checklist

- [x] Written with the help of an LLM or coding agent? Yes, `gpt-5.6-sol` and review by Claude Opus 5.

### DIFF
--- a/packages/bsky/src/views/types.ts
+++ b/packages/bsky/src/views/types.ts
@@ -94,7 +94,12 @@ export type PostgateRecord = app.bsky.feed.postgate.Main
 export const isPostgateDisableRuleType =
   app.bsky.feed.postgate.disableRule.$isTypeOf
 
-export type FeedViewPost = app.bsky.feed.defs.FeedViewPost
+// Temporary off-Lexicon fields used to expose canonical OP thread numbering
+// on feed items while the public response schema remains unchanged.
+export type FeedViewPost = app.bsky.feed.defs.FeedViewPost & {
+  opThreadPostIndex?: number
+  opThreadPostCount?: number
+}
 export type ReasonPin = app.bsky.feed.defs.ReasonPin
 export type ReasonRepost = app.bsky.feed.defs.ReasonRepost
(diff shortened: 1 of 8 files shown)


packages/dev-env/src/network.ts
```diff
@@ -69,6 +69,7 @@ export class TestNetwork extends TestNetworkNoAppView {
     const bsync = await TestBsync.create({
       apiKeys: [bsyncApiKey],
       dbUrl: dbPostgresUrl,
+      dbSchema: `bsync_${dbPostgresSchema}`,
```
rafaeleyng, line 72: Curious, why this? Not even saying it is wrong, but did something happen that you added that?

packages/bsky/src/views/types.ts
```diff
@@ -94,7 +94,12 @@ export type PostgateRecord = app.bsky.feed.postgate.Main
 export const isPostgateDisableRuleType =
   app.bsky.feed.postgate.disableRule.$isTypeOf
 
-export type FeedViewPost = app.bsky.feed.defs.FeedViewPost
+// Temporary off-Lexicon fields used to expose canonical OP thread numbering
```
rafaeleyng: And also why are we doing this instead of changing the lexicon? IIRC we had a lexicon change yesterday

rafaeleyng (approved):
Looks good

packages/dev-env/src/network.ts
ds-boyce, line 72: Added this to avoid a race condition when the test from this PR and the test from #5365 are run in parallel:

```
duplicate key violates pg_namespace_nspname_index
```

packages/bsky/src/views/types.ts
ds-boyce: It was suggested to keep it off-lexicon while we beta test the feature.

PR #5381, opened 2026-08-11
xrpc-server: accept empty chunked request bodies

### LINKED ISSUE #3267: "A request body was provided when none was expected" with proxies that set transfer-encoding
**Describe the bug**

When running a PDS behind some proxies (in this instance, a Cloudflare tunnel) any POST without a body fails with, "A request body was provided when none was expected". This is because the proxy sets a "Transfer-Encoding: Chunked" header even when no body is sent.

Arguably this is a bug in the proxy, but the `getBodyPresence` distinguishes between "no body" and "empty body" when the only thing that actually matters to the validator function is whether a body is present and non-zero length or absent/empty. The `validateInput` reads the body anyway, so it might be more robust to wait until after it reads the body and then just asserts that the body is present and non-zero length if one is expected and vice-versa.

I may be able to make a PR for this if welcome.

**To Reproduce**

1. Run a PDS instance behind a Cloudflare tunnel
2. Log in with a client and make a request with no body, eg. request an email confirmation token.
3. Observe that the request is rejected with the above error.

**Expected behavior**

The request succeeds as normal.

**Details**

- Operating system: Ubuntu 24.04.1 LTS
- Node version: 20.11.1 (official PDS docker image)

**Additional context**

A workaround to this particular failure is to disable chunked transfer encoding in the cloudflare tunnel options.

AG0708:
## What & why

Some proxies add `Transfer-Encoding: chunked` to requests even when they send no body. The header describes framing, so treating it alone as proof of a non-empty body rejects valid no-input XRPC calls.

For methods without input, this waits for either body data or the end of the request stream. Empty chunked requests are accepted, while non-empty requests remain rejected. Tests cover both the legacy and schema-based server paths.

Fixes #3267.

## Checklist

OpenAI Codex assisted with the implementation and tests. I reviewed the full diff and validation results.

### DIFF
--- a/.changeset/empty-bodies-forward.md
+++ b/.changeset/empty-bodies-forward.md
@@ -1,0 +1,5 @@
+---
+'@atproto/xrpc-server': patch
+---
+
+Allow requests with empty chunked transfer encoding when a method expects no input.
--- a/packages/xrpc-server/src/util.ts
+++ b/packages/xrpc-server/src/util.ts
@@ -1,5 +1,4 @@
 import assert from 'node:assert'
-import { once } from 'node:events'
 import type { IncomingMessage, OutgoingMessage } from 'node:http'
 import { type Duplex, type Readable, pipeline } from 'node:stream'
 import {
@@ -248,11 +247,7 @@ export function createLexiconInputVerifier<I extends Input = Input>(
   if (def.type === 'query' || !def.input) {
     return async (req) => {
       // @NOTE We allow (and ignore) "empty" bodies
-      if (await hasNonEmptyBody(req)) {
-        throw new InvalidRequestError(
-          `A request body was provided when none was expected`,
-        )
-      }
+      await assertEmptyBody(req)
 
       return undefined as I
     }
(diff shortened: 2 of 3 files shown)


matthieusieben (changes requested):
Looks mostly good !

One comment I have is esthetic: `hasNonEmptyBody` does not make it clear enough the the response will be consumed and that it should not be used anymore. A dev, in the future could be tempted to read the `req` after using it.

I recommend renaming it, and changing its usage to always throw for non empty payloads:

```ts
// Something like (don't hesitate to use a better name)
await processEmptyBody(req)
```

Also, we should account for any error that occurs while reading the body (eg. an invalid encoding), and only return from the method when the body was drained (instead of a background `res.resume()`):

```ts
try {
  for await (const _ of req) {
    // drain the body, propagating any processing error
  }
} catch (cause) {
  throw new InvalidRequestError('Failed to process (unexpected) body payload', { cause })
}
```

Note there migth exist a `drain` utility somewhere accessible ?

Or maybe there is a smart way of doing this with `once` (which is very likely more efficient than using an async iterable)?

```ts
res.resume()
await once(res, 'end').catch(cause => {
  throw new InvalidRequestError('Failed to process (unexpected) body payload', { cause })
})
```

AG0708:
Addressed in c9dfb0327. I renamed the helper to assertEmptyBody, made it drain the chunked stream completely before returning or rejecting, and translate stream read failures into InvalidRequestError with the original cause. Regressions cover waiting for the full drain and error propagation. The branch is rebased onto current main; xrpc-server build, all 168 package tests, Prettier, ESLint, and diff checks pass.

matthieusieben (approved):
Thank you very much 💙

PR #5389, opened 2026-08-13
Allows the $type property of the LexiconDoc type

### LINKED ISSUE #5388: The `LexiconDoc` type does not allow the `$type` property
**Describe the bug**

When you retrieve a Lexicon using `@atproto/lex` and generate a server API with `@atproto/lex-cli`,
```
$type: 'com.atproto.lexicon.schema',
```
a property such as this is included in the object that expects a `LexiconDoc`, resulting in a type error.

**To Reproduce**

1. `ts-lex install app.bsky.feed.post`
2. `pnpm lex gen-server ./src/lexicons/ ./lexicons/**`
3. `src/lexicons/lexicon.ts` has type error.

**Expected behavior**

It is expected that a type error will not occur. 
I believe that `LexiconDoc` should allow the `$type` property.

**Details**

- Operating system: Linux 7.1.4-1-cachyos
- Node version:v24.11.1

**Additional context**

tomo-x7:
## What & why

After installing the lexicon using `@atproto/lex` and generating code using `@atproto/lex-cli`, a type error occurs due to the `$type` property. To resolve this, we will modify the `LexiconDoc` type so that it accepts the `$type` property.
fix #5388 

## Checklist

### DIFF
--- a/packages/lexicon/src/types.ts
+++ b/packages/lexicon/src/types.ts
@@ -430,7 +430,7 @@ export const lexiconDoc = z
     revision: z.number().optional(),
     description: z.string().optional(),
     defs: z.record(z.string(), lexUserType),
-    $type: z.string().optional(),
+    $type: z.literal('com.atproto.lexicon.schema').optional(),
   })
   .refine(
     (doc) => {
--- a/packages/lexicon/src/types.ts
+++ b/packages/lexicon/src/types.ts
@@ -430,6 +430,7 @@ export const lexiconDoc = z
     revision: z.number().optional(),
     description: z.string().optional(),
     defs: z.record(z.string(), lexUserType),
+    $type: z.string().optional(),
   })
   .refine(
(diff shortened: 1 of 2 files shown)


packages/lexicon/src/types.ts
```diff
@@ -430,6 +430,7 @@ export const lexiconDoc = z
     revision: z.number().optional(),
     description: z.string().optional(),
     defs: z.record(z.string(), lexUserType),
+    $type: z.string().optional(),
```
matthieusieben: This should be defined as a literal with the actual lexicon document `$type` value.

matthieusieben (changes requested):
Looks mostly good, thanks.

packages/lexicon/src/types.ts
tomo-x7: OK. I fixed!

PR #5393, opened 2026-08-13
Fix ascending report pagination tie-breaker

gcwill70:
## Summary

There is a risk that mismatched sort direction in keyset columns will lead to skipped or missed reports. This will make the sort direction in both columns the same to avoid that.

This was added so that [ozone#561](https://github.com/bluesky-social/ozone/pull/561) can show oldest reports reliably.

## Testing

- `pnpm exec eslint packages/ozone/src/mod-service/report.ts packages/ozone/tests/query-reports.test.ts`
- integration test could not run locally because Docker and `DB_TEST_POSTGRES_URL` are unavailable
- package build is blocked by unrelated `isolatedDeclarations` errors in `packages/lex/lex-schema`

### DIFF
--- a/.changeset/tidy-report-pages.md
+++ b/.changeset/tidy-report-pages.md
@@ -1,0 +1,5 @@
+---
+'@atproto/ozone': patch
+---
+
+Fix ascending report pagination when multiple reports share a timestamp.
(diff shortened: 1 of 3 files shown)


gcwill70:
This keeps the `createdAt` and `id` sort directions aligned so ascending queries can scan the existing descending index backward and the cursor predicate matches the result ordering. In addition to avoiding an extra sort, it prevents duplicate or skipped reports when multiple reports share the same timestamp.

foysalit (approved):
nice catch!

PR #5394, opened 2026-08-13
Add Romanian (ro) localization

Laszlo19:
Adds Romanian (`ro`) as an available locale for the OAuth provider's
sign-in / sign-up / account-management UI (`@atproto/oauth-provider-ui`).

Changes:
- New `src/locales/ro/messages.po` with the Romanian translations
- Registered `ro` (Română 🇷🇴) in `src/locales/locales.ts` so it appears in
  the locale selector and in locale detection
- Added `ro` to the `locales` array in `lingui.config.ts` so extraction and
  compilation include the catalog

### DIFF
--- a/.changeset/romanian-locale.md
+++ b/.changeset/romanian-locale.md
@@ -1,0 +1,5 @@
+---
+'@atproto/oauth-provider-ui': patch
+---
+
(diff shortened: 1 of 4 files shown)


matthieusieben (approved):
thank you

PR #5408, opened 2026-08-14
Persist the OAuth Provider UI's selected interface language

bigmoves:
## What & why

Selecting an interface language in the OAuth provider UI did not survive a page
reload: the locale lived only in `LocaleProvider`'s React state, initialized from
`detectLocale(userLocales)`, so every load fell back to the client's `ui_locales`
hint or `navigator.languages` and discarded the user's choice.

An explicit selection is now persisted in `localStorage` (new
`src/locales/locale-storage.ts`) and takes precedence over both — otherwise a
reload of the authorization page, which re-supplies `ui_locales`, would undo it.
A merely *detected* locale is not stored, so it keeps following the browser.

Details:

- The key is `@@atproto/oauth-provider-ui(locale)`, following the
  `@@<package>(<key>)` namespacing `@atproto/oauth-client-browser` already uses
  for its `localStorage` entries — the origin is shared with whatever else the
  PDS serves.
- Every storage access is guarded: reading `localStorage` throws (rather than
  returning `null`) when a browser is configured to block storage, and an
  unrecognized stored value is ignored.
- Storage is origin-scoped, so a choice made on the authorization page also
  applies to the account page.

Verified against the mock UI (`pnpm dev:ui`): picking Français on the
authorization page and reloading comes back in French, and `/account` follows.

## Checklist

- [x] `pnpm verify` passes; built the touched package (`tsc --build` +
      `vite build`) rather than a full `pnpm build --force` of the repo
      (`src/locales/locale-storage.test.ts`; full package suite green)
- [x] Written with the help of an LLM or coding agent? Say so here, and name the
      tooling. — Yes: Claude Code (Opus 5).

### DIFF
--- a/.changeset/persist-locale-selection.md
+++ b/.changeset/persist-locale-selection.md
@@ -1,0 +1,5 @@
+---
+'@atproto/oauth-provider-ui': patch
+---
+
+Persist the interface language selected in the OAuth UI, so that it survives a page reload.
--- a/packages/oauth/oauth-provider-ui/CLAUDE.md
+++ b/packages/oauth/oauth-provider-ui/CLAUDE.md
@@ -208,6 +208,12 @@ form field is `code` but the API takes `token`.
 - A msgid that disappears and returns loses its translation. After extraction,
   check the _count_ of untranslated entries, not just the msgid diff.
 - Fill in French only; other locales are translated externally.
+- The locale is React state, so anything not persisted is lost on reload.
+  `LocaleProvider` restores an explicit selection from `localStorage`
+  (`locale-storage.ts`) and gives it precedence over the client's `ui_locales`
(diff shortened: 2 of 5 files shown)


matthieusieben (approved):
This is definitely a good improvement. Eventually we will want to save this to the backend.

matthieusieben:
Small nit: do we want to validate the saved locale before using it? This will allow falling back to the detected locale instead of en

bigmoves:
> Small nit: do we want to validate the saved locale before using it? This will allow falling back to the detected locale instead of en

Yeah it's validated via `isLocale`. There's a test for it here https://github.com/bluesky-social/atproto/pull/5408/changes#diff-7a1eea2a652b9405ad9616a165fd63401721ce4cc7531840a9bc823f7ae7f67aR46

PR #5421, opened 2026-08-17
Use dedicated Ozone account preferences endpoint

foysalit:
## Summary

- replace the Ozone implementation of `app.bsky.actor.getPreferences` with `tools.ozone.moderation.getAccountPreferences`
- preserve moderator/admin authentication and the existing downstream PDS service-auth request
- add a Lexicon for the Ozone-specific method and update the integration test
- remove the no-longer-needed direct `@atproto/lex` dependency from Ozone

This avoids colliding with the public PDS NSID, which caused AT Protocol proxy routing to treat browser requests differently from direct Ozone requests.

## Verification

- `CI=true pnpm --dir packages/ozone run build`
- focused ESLint on changed TypeScript files
- Prettier check on changed files
- `git diff --check`
- adversarial cross-model review: no findings

The focused integration test could not run locally because Docker is unavailable and no host `DB_TEST_POSTGRES_URL` is configured. Full `pnpm verify` also stopped at the existing PDS import check because `key-encoder` was unavailable from the local built dependency state.

## AI assistance

Implemented and reviewed with Codex/LLM assistance.


matthieusieben (approved):
Looks good now.

PR #5422, opened 2026-08-17
Drop @atproto/api runtime dependency from bsky

bigmoves:
## What & why

`packages/bsky` carried `@atproto/api` as a **runtime dependency** because of exactly two production files. Both imported the age assurance region/rule helpers — `getAgeAssuranceRegionConfig`, `computeAgeAssuranceRegionAccess`, `AgeAssuranceRuleID` — from `@atproto/api`:

- `src/api/age-assurance/util.ts`
- `src/api/app/bsky/ageassurance/begin.ts`

Those helpers live in `packages/api/src/age-assurance.ts`, which is hand-written logic that happens to sit alongside the legacy generated client. So two imports were pinning the whole legacy client stack into bsky's shipped artifact and into the `services/bsky` image.

Per review, this sources them from **`@bsky/sdk`** (`@bsky/sdk/utils`), which already ships `getAgeAssuranceRegionConfig`, `computeAgeAssuranceRegionAccess`, and `AgeAssuranceRuleID` typed against `@atproto/lex` — so no logic is duplicated into this repo.

`src/api/age-assurance/util.ts` keeps only the two bsky-specific wrappers (`computeAgeAssuranceAccessOrThrow`, `createLocationString`); `begin.ts` imports `getAgeAssuranceRegionConfig` straight from the SDK.

This works because the SDK declares `@atproto/lex` as a **peer** dependency, so it resolves to `link:../lex/lex` in this workspace. The branded types (`l.UnknownString` inside `Access`) therefore come from the same instance, and the SDK's `Config` / `ConfigRegion` are structurally compatible with this package's own generated `app.bsky.ageassurance.defs` tree — including `AGE_ASSURANCE_CONFIG`, which is still built from the local tree.

`@atproto/api` moves to `devDependencies`, and the `../api` project reference moves from `tsconfig.build.json` to `tsconfig.test.json` — mirroring how `packages/pds` is already set up. `pnpm why @atproto/api --filter @atproto/bsky --prod` now returns nothing.

Also converts the `age-assurance-v2` mocked config fixture to `$build`, so it is checked against the same schema the production code consumes rather than the structurally-compatible legacy type.

### What this does not do

`@atproto/api` stays in `devDependencies`. All 58 bsky test files drive the services through `AtpAgent`, which `dev-env` hands them (`packages/dev-env/src/bsky.ts:148`, `pds.ts:75`), and pass legacy `ids.*` NSID constants to `network.serviceHeaders()`. That can't go away until `dev-env` itself migrates — a separate, much larger change that would also unblock pds and ozone shedding their test-side dep.

`packages/api` keeps its own copy of these helpers, since it is a published package and they are public API for external clients. That copy is now the redundant one and can go with the rest of `@atproto/api`.

## Checklist

Written with Claude Code (Claude Opus 5).

### Test evidence

- `tests/views/age-assurance.test.ts` + `tests/views/age-assurance-v2.test.ts` — 29 passed, 1 skipped


packages/bsky/src/api/age-assurance/util.ts
```diff
@@ -1,10 +1,130 @@
...
+  reason: AgeAssuranceRuleID
+}
+
+/**
+ * Returns the first matched region configuration based on the provided
+ * filters. Region configurations that declare `platforms` only match when the
+ * provided platform is included in that list. If no platform filter is
+ * provided, platform restrictions are ignored.
+ */
+export function getAgeAssuranceRegionConfig(
```
devinivy: We should be able to use `@bsky/sdk` for this and other age assurance helpers.

packages/bsky/src/api/age-assurance/util.test.ts
```diff
@@ -0,0 +1,267 @@
+import { describe, expect, it } from 'vitest'
```
devinivy: I think we can nix this once we move to `@bsky/sdk` for the age assurance utils.

packages/bsky/src/api/age-assurance/util.ts
bigmoves: https://github.com/bluesky-social/atproto/pull/5422/commits/b6ba87e02aebd4f173e5b98a042c883a80381814

packages/bsky/src/api/age-assurance/util.test.ts
bigmoves: https://github.com/bluesky-social/atproto/pull/5422/commits/b6ba87e02aebd4f173e5b98a042c883a80381814

bigmoves:
@devinivy I updated to use bsky/sdk

devinivy (approved):

packages/bsky/package.json
```diff
@@ -47,6 +46,7 @@
     "@atproto/sync": "workspace:^",
     "@atproto/syntax": "workspace:^",
     "@atproto/xrpc-server": "workspace:^",
+    "@bsky/sdk": "^1.0.1",
```
matthieusieben, line 49: This introduces a cyclic dependency on published version of packages from this repo.

Ideally, the stuff we need here would be available as an internal standalone package in the `@bsky/` ns.

matthieusieben (approved):

PR #5425, opened 2026-08-18
Migrate dev-env off @atproto/api onto @atproto/lex

bigmoves:
## What & why

`@atproto/dev-env` is the fixture builder behind almost every integration test in `pds`, `bsky`, and `ozone`, and it was still built on `AtpAgent`. Because it hands that agent to its consumers, it was also the thing pinning ~140 test files to the legacy stack. This migrates the package itself onto `@atproto/lex`.

The blocker turned out to be upstream of the migration: **dev-env had no `lex build` codegen at all**, so it had no generated schemas to call with. That's wired up first (`codegen:lex` + `prebuild`, `.gitignore`, tsconfig reference), then the internals move over:

- **`ServiceProfile`** and its Ozone / Lexicon-authority subclasses hold a `PasswordSession` plus a `Client` instead of an `AtpAgent`. This lets a hack go: `migrateTo` used to reach into `newAgent.sessionManager.pdsUrl = undefined` to stop a mid-migration session from calling the old PDS, and now simply constructs the session without the stale `didDoc`.
- **`ModeratorClient`** exposes `client` instead of `agent`, and its methods return the response body directly rather than an `{ data }` wrapper.
- **`SeedClient`** builds every record through `Client`. `uploadFile` and `createProfile` now return a `BlobRef` from `@atproto/lex-data` — read its CID with `getBlobCid()` / `getBlobCidString()` rather than `.ref`.
- **`mock/`** and **`seed/thread-v2.ts`** follow the same path. `TestOzone` gains a `getClient()` alongside its existing `getAgent()`.

`@atproto/lexicon` is dropped from the package entirely — its last use was a single type, now `LexiconDocument` from `@atproto/lex-document`.

### What's deliberately left

`getAgent()` on `TestPds` / `TestBsky` / `TestOzone` and `SeedClient.agent` stay, as the compatibility surface for the test suites still driving `AtpAgent`. That's **four intentional `@atproto/api` imports, down from nine files**.

Removing them means migrating 176 call sites across 140 test files, and a third of that is hard-blocked: `ozone` still runs `lex gen-server` into `src/lexicon`, and 48 of its test files import from that legacy tree, so they have nothing to import yet. `bsky` (80 sites) and `pds` (40 sites) are unblocked and can follow in separate PRs; `ozone`'s tests are gated on lexifying its server first.

### Why the test churn

Migrating a shared dev-env helper retypes its signature against the generated schemas, so every caller inherits the branded scalars at once — that's what the `ozone` / `pds` / `bsky` test diffs are. Branding at the source rather than the call site collapsed whole clusters (typing `DidAndKey.did` as `DidString` fixed every test passing `network.ozone.adminAccnt.did`). Elsewhere: `new Date().toISOString()` → `currentDatetimeString()`, and local test types derived from the helper (`Parameters<ModeratorClient['emitEvent']>[0]`) so they track it automatically.

### One hazard worth flagging

The legacy `BlobRef` is a **class** that carries `$type: 'blob'` only through `toJSON()`; the lex `BlobRef` is a plain object carrying it as an own property. A blob read through one stack and written through the other therefore serializes wrong, and the PDS rejects it at runtime:

```
XrpcResponseError: Invalid app.bsky.actor.profile record:
Expected blob value type (got {"ref": ..., "mimeType": ...}) at $.record.avatar
```

`author-feed.test.ts` hit exactly this — it read a profile via the legacy agent and spread it into the now-lex `SeedClient.updateProfile`. Fixed by reading and writing through the same client. It typechecked because `updateProfile` takes `Record<string, unknown>`, so all 11 call sites were audited by hand; the rest already source their blobs from `sc.uploadFile` / `res.body.blob`. The hazard is now documented in the `lexification-client` skill.

Also adds `packages/dev-env/src/lexicons/**` to the eslint ignore list, matching every other package with a generated lex tree.

## Test plan

| | Result |
|---|---|
| `pnpm build --force && pnpm verify` | passes |
| ozone | 65/65 files, 493 passed + 1 skipped |
| bsky | 84/84 files, 946 tests (102 skipped) |
| pds | 44/45 — the miss is `oauth.test.ts` failing to resolve `@atproto/oauth-client-browser-example`, an unbuilt example package unrelated to this change |

No new tests: this is a refactor of test infrastructure, and the existing suites are the regression check — the `author-feed` bug above is a case of that working.

Note for reviewers reproducing locally: both suites boot a full constellation per test file, so at default worker counts a number of them blow their `beforeAll` timeout on smaller machines. `--maxWorkers=3` was needed here. Overlapping runs also share Postgres schemas and produce convincing false failures (duplicate keys, extra rows in snapshots) — reset with `cd packages/dev-infra && docker compose down --volumes`.

## Checklist

Written with Claude Code (Claude Opus 5). The migration, the test fallout, and the `author-feed` blob fix were agent-authored; all of it was verified against the suites above.


packages/bsky/tests/views/author-feed.test.ts
```diff
@@ -602,19 +599,15 @@ describe('pds author feed views', () => {
...
         repo: alice,
-        collection: 'app.bsky.actor.profile',
-        rkey: 'self',
       })
 
-      assert(isValidProfile(profile.data.value))
-
-      const newProfile: AppBskyActorProfile.Record = {
-        ...profile.data.value,
+      const newProfile = {
```
matthieusieben: We should probably explicit types when possible:

```suggestion
      const newProfile: app.bsky.actor.profile.Main = {
```

or:

```ts
      const newProfile = app.bsky.actor.profile.$build({
        ...profile.value,
        pinnedPost: {
          uri: post.ref.uriStr,
          cid: post.ref.cidStr,
        },
      })
```

packages/dev-env/src/mock/index.ts
```diff
@@ -1,24 +1,32 @@
...
+} from '@atproto/lex'
+import { PasswordSession } from '@atproto/lex-password-session'
 import { AtUri, type AtUriString } from '@atproto/syntax'
 import { EXAMPLE_LABELER, RecordRef, type TestNetwork } from '../index.js'
+import { app, chat, com, tools } from '../lexicons/index.js'
 import { postTexts, replyTexts } from './data.js'
 import blurHashB64 from './img/blur-hash-avatar-b64.js'
 import labeledImgB64 from './img/labeled-img-b64.js'
 
+const REASON_SPAM = 'com.atproto.moderation.defs#reasonSpam'
```
matthieusieben: see https://github.com/bluesky-social/atproto/pull/5439

packages/dev-env/src/service-profile.ts
```diff
@@ -14,82 +15,94 @@ export type ServiceMigrationOptions = {
...
-      lxm: 'com.atproto.server.createAccount',
-    })
+    const newPdsDesc = await newPdsClient.call(
+      com.atproto.server.describeServer,
+    )
+    const serviceAuth = await this.client.call(
+      com.atproto.server.getServiceAuth,
+      {
+        aud: newPdsDesc.did,
+        lxm: 'com.atproto.server.createAccount',
```
matthieusieben: We should remove string literals with references whenever possible.

```suggestion
        lxm: com.atproto.server.createAccount.$lxm,
```

packages/ozone/tests/get-profiles.test.ts
```diff
@@ -31,15 +31,17 @@ describe('get profiles through ozone', () => {
...
-        { actors },
-        {
-          headers: await network.ozone.modHeaders(
-            'app.bsky.actor.getProfiles',
-            'admin',
-          ),
-        },
-      )
+      const { data } = await network.ozone
+        .getAgent()
```
matthieusieben, line 35: Should we add a `.getClient()` on the `TestOzone` class (so that those code paths can be migrated) ?

packages/ozone/tests/moderation-statuses.test.ts
```diff
@@ -137,7 +135,7 @@ describe('moderation-statuses', () => {
     it('returns paginated statuses', async () => {
       // We know there will be exactly 4 statuses in db
       const getPaginatedStatuses = async (
-        params: ToolsOzoneModerationQueryStatuses.QueryParams,
+        params: Parameters<ModeratorClient['queryStatuses']>[0],
```
matthieusieben: Could this use the `QueryStatusesParams` type defined in the `moderator-client.ts` file ?

matthieusieben (approved):
Looks good. Left a few minor comments that you may choose to ignore.

Feel free to merge whenever.

packages/ozone/tests/get-profiles.test.ts
bigmoves: yeah it's added but we can't use here until #5426 lands then we'll circle back

packages/ozone/tests/moderation-statuses.test.ts
bigmoves: yeah we can clean this up

packages/bsky/tests/views/author-feed.test.ts
bigmoves: [bfb0b3a](https://github.com/bluesky-social/atproto/pull/5425/commits/bfb0b3a23a70e5d7e8ad5a9ff90f70d0b11b96c6)

packages/ozone/tests/moderation-statuses.test.ts
bigmoves: d1ea31e

packages/dev-env/src/service-profile.ts
bigmoves: [1808f4c](https://github.com/bluesky-social/atproto/pull/5425/commits/1808f4cf6436e7d620099268b4c727ee81df053d)

packages/dev-env/src/mock/index.ts
bigmoves: [ff1c87d](https://github.com/bluesky-social/atproto/pull/5425/commits/ff1c87d1f09e17b454aeead66066a73d82cb0496)

bigmoves:
@matthieusieben addressed your feedback, should be ready to merge now

PR #5426, opened 2026-08-19
Migrate ozone off @atproto/api onto @atproto/lex

bigmoves:
## What & why

Ozone was the last package still generating its server stack with `lex gen-server`. It now uses `lex build` like `pds` and `bsky`, the legacy `src/lexicon` tree is gone, and `@atproto/api` is out of `src/` entirely.

The route migration and the branded database types turn out to depend on each other, which is why they land together. With `server.add()`, handler params arrive already carrying `DidString` / `DatetimeString`, so they flow into a branded Kysely schema and back out into lex views with no cast at either seam. Doing only one half leaves a pile of casts at whichever boundary was left behind — I tried the type-only half first and had ~150 scattered `as DidString` casts in the view formatters to show for it. Migrating the routes deleted them, along with 23 inbound `$Params` casts that became redundant.

- **95 route registrations** across 73 files move from `server.<ns>.<method>()` to `server.add(schema, …)` from `@atproto/xrpc-server`. `createServer` comes from the same package; `server.router` replaces `server.xrpc.router`.
- `AtpAgent` → `Client` at all 43 call sites. `CredentialSession` → `PasswordSession` from `@atproto/lex-password-session` — ozone is its first consumer in this repo. `fetchRecord` moves to `xrpcSafe` with `res.error === 'RecordNotFound'`.
- `lexicons.validate()` → `$safeParse`, `asPredicate(validateX)` → `X.$matches`, `ids.*` → `$lxm` / `$nsid`, `new Date().toISOString()` → `currentDatetimeString()`.
- Kysely schema types, config and auth credentials carry the branded scalars.
- `@atproto/api` and `@atproto/lexicon` move to `devDependencies` (the suites still drive the legacy client). `@atproto/xrpc` and `@atproto/lex-cli` are dropped outright, which also retires the `noUnusedLocals: false` escape hatch and its "needed because of lex-cli" comment.

Agent files are updated in the same change: `CLAUDE.md` no longer claims ozone's prebuild skips codegen, and the `xrpc-server` / `lexification-server` skills no longer describe ozone as un-migrated.

## Behaviour changes

Both fall out of validating against the Lexicon schemas rather than the legacy runtime, and both are in the changeset:

- The AT Proto data model has no floating-point numbers; the legacy stack accepted them anyway. A float in an `unknown` field — e.g. a `tools.ozone.setting` value — is now rejected at input validation. Reads are unaffected (`validateResponse: false`).
- A value whose shape contradicts the Lexicon is rejected by schema validation before a route's own validators run, so the error surfaced is the schema's rather than the application's.

`BlobDiverter` and the account-preferences proxy previously threw `XRPCError` from `@atproto/xrpc` purely to signal retryability to `retryHttp`. They now throw `UpstreamHttpError`, which extends `XRPCError` from `@atproto/xrpc-server` so an error escaping a handler keeps its upstream status and message instead of collapsing into a generic 500.

## Test plan

| | Result |
|---|---|
| `pnpm build` | passes |
| `pnpm verify` | passes |
| `tsc` for `src` + `tests` | clean |
| ozone suite | see below |

**The only snapshot that changed in this branch is `10.5` → `10`.** All 43 snapshots across 65 suites are otherwise byte-identical — that single diff is the float behaviour change above. I refreshed only that one and ran everything else with no `-u`, so nothing is masked.

Local suite runs on this machine hit `beforeAll` timeouts booting a full constellation per file (`--maxWorkers=2` was needed, and some suites still time out); overlapping runs against a shared Postgres schema also produce convincing false failures — sequence-ID drift and extra snapshot rows — so reset with `cd packages/dev-infra && docker compose down --volumes` between runs. CI is the real check here.

## Notes for reviewers

- **Touches `packages/dev-env/src/ozone.ts`** (branding the `add*Did` params), which #5425 also edits. Expect a conflict depending on merge order.
- **The branding rules were name-based heuristics, not proof.** I caught two false positives — `communication_template.subject` is an email subject line, not a DID, and `member.lastUpdatedBy` stores an `'admin_token'` sentinel — so `src/db/schema/` is worth a skim in case a similar one slipped through.
- Two of the three test failures I chased were bugs introduced during the migration, both fixed: swapping `tests/_util.ts` to lex's `lexToJson` leaked legacy `BlobRef` instances' internal `original` field into 10 snapshots (it reads exactly like a wire-format regression, and is now documented in the `lexification-client` skill); and `UpstreamHttpError` first broke status propagation, then broke the retry predicate when it keyed on `instanceof` against an ozone-local class — it now keys off `statusCode` from the shared `XRPCError`.

## Checklist

Written with Claude Code (Claude Opus 5). The migration, the branded-type fallout, and the fixes above were agent-authored; verification is per the test plan.


bigmoves:
Local suite is fully green: **488 passed + 1 skipped** in the full run, with the remaining 5 re-run in isolation and passing (`report-action` ×4, `get-account-preferences` ×1). Those 5 were `beforeAll` timeouts from this machine booting two full constellations under contention — 5 failures, 5 hook-timeout messages, zero assertion failures — not code failures.

**Snapshots: 44/44 passed**, run with no `-u`. The only snapshot changed anywhere in this branch is the intentional `10.5` → `10`, which is the float-rejection behaviour change described above. Everything else is byte-identical, which is the strongest evidence I have that the migration is wire-compatible.

`pnpm build` and `pnpm verify` pass, and `tsc` is clean for both `src` and `tests`.

packages/ozone/src/api/label/fetchLabels.ts
```diff
@@ -1,13 +1,18 @@
...
 export default function (server: Server, ctx: AppContext) {
-  server.com.atproto.temp.fetchLabels({
+  server.add(com.atproto.temp.fetchLabels, {
     auth: ctx.authVerifier.standardOptionalOrAdminToken,
     handler: async ({ params }) => {
       const { limit } = params
       const since =
-        params.since !== undefined ? new Date(params.since).toISOString() : ''
+        params.since !== undefined
+          ? toDatetimeString(new Date(params.since))
```
matthieusieben: https://github.com/bluesky-social/atproto/pull/5441

packages/ozone/src/api/label/subscribeLabels.ts
```diff
@@ -1,25 +1,28 @@
...
+        }
       }
-    }
 
-    for await (const evt of outbox.events(cursor, signal)) {
-      yield { $type: '#labels', ...evt }
-    }
-  })
+      for await (const evt of outbox.events(cursor, signal)) {
+        yield {
```
matthieusieben: Can we use either `com.atproto.label.subscribeLabels.$build(evt)` or `$type: com.atproto.label.subscribeLabels.labels.$type` here ?

packages/ozone/src/api/moderation/cancelScheduledActions.ts
```diff
@@ -36,10 +37,10 @@ export default function (server: Server, ctx: AppContext) {
               comment,
             },
             subject: subjectFromInput({
-              did: subject,
+              did: subject as DidString,
               $type: 'com.atproto.admin.defs#repoRef',
             }),
-            createdBy,
+            createdBy: createdBy as DidString,
```
matthieusieben: would it be possible to propagate the typing into the services instead of type casting everywhere ?

packages/ozone/src/api/moderation/cancelScheduledActions.ts
```diff
@@ -36,10 +37,10 @@ export default function (server: Server, ctx: AppContext) {
               comment,
             },
```
matthieusieben: Can we use `tools.ozone.moderation.defs.cancelScheduledTakedownEvent.$build` for these.

packages/ozone/src/api/moderation/getSubjects.ts
```diff
@@ -63,10 +66,12 @@ export default function (server: Server, ctx: AppContext) {
...
-          profile: profile && {
-            $type: 'app.bsky.actor.defs#profileViewDetailed',
-            ...profile,
-          },
+          profile:
+            profile &&
+            ({
+              $type: 'app.bsky.actor.defs#profileViewDetailed',
+              ...profile,
+            } as unknown as Unknown$TypedObject),
```
matthieusieben:
```suggestion
          // @ts-expect-error (documented as "unknown")
          profile: profile
            ? app.bsky.actor.defs.profileViewDetailed.$build(profile)
            : undefined,
```

packages/ozone/src/api/setting/upsertOption.ts
```diff
@@ -1,15 +1,20 @@
...
 import type { Member } from '../../db/schema/member.js'
-import type { Server } from '../../lexicon/index.js'
+import { tools } from '../../lexicons/index.js'
 import type { SettingService } from '../../setting/service.js'
 import { settingValidators } from '../../setting/validators.js'
 
+const ROLEADMIN = tools.ozone.team.defs.roleAdmin.value
+const ROLEMODERATOR = tools.ozone.team.defs.roleModerator.value
+const ROLETRIAGE = tools.ozone.team.defs.roleTriage.value
+const ROLEVERIFIER = tools.ozone.team.defs.roleVerifier.value
```
matthieusieben, line 14: FYI (no _need_ to change): https://github.com/bluesky-social/atproto/pull/5439

packages/ozone/src/api/setting/upsertOption.ts
```diff
@@ -27,7 +32,7 @@ export default function (server: Server, ctx: AppContext) {
       // if the caller is using moderator auth and storing personal setting
       // use the caller's DID as the owner
       if (scope === 'personal' && access.type === 'moderator') {
-        ownerDid = access.iss
+        ownerDid = access.iss as DidString
```
matthieusieben: `ctx.authVerifier` should be typed to return `iss` as `DidString`

packages/ozone/src/report/stats.ts
```diff
@@ -376,8 +378,8 @@ export class ReportStatsService {
    * Returns 5 result sets covering all group types.
    */
   private async computeBatchedStats(date: string): Promise<BatchedStats> {
-    const dayStart = `${date}T00:00:00.000Z`
-    const dayEnd = `${nextDate(date)}T00:00:00.000Z`
+    const dayStart = `${date}T00:00:00.000Z` as DatetimeString
+    const dayEnd = `${nextDate(date)}T00:00:00.000Z` as DatetimeString
```
matthieusieben: Can we avoid type casting and introduce a `DateString` type in this file instead ?

```ts
type DateString = `${string}-${string}-${string}`
function toDateString(d: Date): DateString {
  // ...
}

const date = toDateString(new Date())
const datetime: DatetimeString = `${date}T00:00:00.000Z`
```

packages/ozone/src/verification/issuer.ts
```diff
@@ -17,47 +30,46 @@ export type VerificationIssuerCreator = (
...
+    // is only needed once the refresh token itself stops working. Probe the
+    // session and re-authenticate on 401 only — a network blip should not
+    // discard a session that is still valid.
+    if (!this.client) return this.login()
+
+    const res = await this.client.xrpcSafe(com.atproto.server.getSession, {})
+    if (res.success) return this.client
+    if (res.reason instanceof XrpcAuthenticationError) return this.login()
+    throw res.reason
   }
```
matthieusieben, line 61: This is not needed. The `PasswordSession` will automatically re-auth internally on 401 responses.


```suggestion
  async getClient() {
    return this.client || this.login()
  }
```

packages/ozone/src/verification/issuer.ts
```diff
@@ -17,47 +30,46 @@ export type VerificationIssuerCreator = (
...
-    }
-
-    return this.agent
+  private async login() {
+    const session = await PasswordSession.login({
+      service: this.verifierConfig.url,
+      identifier: this.verifierConfig.did,
+      password: this.verifierConfig.password,
+    })
+    this.client = new Client(session)
```
matthieusieben:
```suggestion
    if (this.client) throw new Error('concurrent login')
    this.client = new Client(session)
```

packages/ozone/src/verification/issuer.ts
```diff
@@ -17,47 +30,46 @@ export type VerificationIssuerCreator = (
...
-        await this.session.login({
-          identifier: this.verifierConfig.did,
-          password: this.verifierConfig.password,
-        })
-      }
-    }
-
-    return this.agent
+  private async login() {
+    const session = await PasswordSession.login({
```
matthieusieben, line 38: Related to this PR, but could be part of a followup PR; the sessions should ideally be destroyed.

`VerificationIssuer` should implement `AsyncDisposable` and call `await this.client?.logout()` upon disposal.

packages/ozone/tests/db.test.ts
```diff
@@ -49,7 +50,7 @@ describe('db', () => {
           .insertInto('repo_push_event')
           .values({
             eventType: 'pds_takedown',
-            subjectDid: 'x',
+            subjectDid: 'x' as DidString,
```
matthieusieben: we should use an actual did value instead of casting a non did into a `DidString`

matthieusieben (commented):
This is a good first pass, and we could merge as-is. I do recommend that we go a little bit deeper on the migrations though:

- A lot of places still use `$type: 'string literal'`. all these should be replaced with either `ns.$build()` or `ns.$type`
- I'm a bit surprised by the amount of type casting (`as DidString`, ...). The strong types should be propagated into the services (all the way to the database schema definitions).

packages/ozone/src/api/moderation/emitEvent.ts
```diff
@@ -345,7 +347,7 @@ const handleModerationEvent = async ({
 
     if (isLabelEvent) {
       await moderationTxn.formatAndCreateLabels(
-        result.event.subjectUri ?? result.event.subjectDid,
+        (result.event.subjectUri ?? result.event.subjectDid) as UriString,
```
matthieusieben: This type cast is not right. `subjectDid` is a `DidString` and should not be casted into `UriString`. Not sure what the right call is but it looks like `formatAndCreateLabels` should accept `DidString` as input

packages/ozone/src/api/moderation/emitEvent.ts
```diff
@@ -259,7 +259,7 @@ const handleModerationEvent = async ({
     const result = await moderationTxn.logEvent({
       event,
       subject,
-      createdBy,
+      createdBy: createdBy as DidString,
```
matthieusieben: can these `as DidString` be removed?

packages/ozone/src/api/moderation/getRecords.ts
```diff
@@ -17,7 +19,7 @@ export default function (server: Server, ctx: AppContext) {
         ),
         getPdsAccountInfos(
           ctx,
-          params.uris.map((uri) => new AtUri(uri).hostname),
+          params.uris.map((uri) => new AtUri(uri).hostname as DidString),
```
matthieusieben: There is a `did` accessor (throws if not a did)

```suggestion
          params.uris.map((uri) => new AtUri(uri).did),
```

matthieusieben (approved):
I've done a pass (merging main, removing more `as X` type cast, and replacing more `'tools.ozone....'` string literals with the generated counterpart). I think it's good to go now.

foysalit (approved):
Looking great! thank you @bigmoves 🙌🏽

bigmoves:
@matthieusieben ci was failing and i fixed the migrations

matthieusieben:
@claude review this PR according to the guidance in .github/claude-review-prompt.md

PR #5448, opened 2026-08-26
Forward request headers through PDS service proxies.

estrattonbailey:
Forwards all `x-atproto-*` headers through the PDS to connecting services. During the switchover, this PR dual-writes the old `x-bsky-topics` header and new `x-atproto-bsky-topics` header. We'll remove this later.

General note: this is not intended as a load bearing interface, but rather as a convenience for off-protocol and prototyping needs. We're keeping this scoped to `x-aptroto` for now as opposed to `x-*` to avoid functional overlap and unexpected header confusion.

Companion PR https://github.com/bluesky-social/social-app/pull/11571

### DIFF
--- a/packages/pds/src/pipethrough.ts
+++ b/packages/pds/src/pipethrough.ts
@@ -258,11 +258,16 @@ export async function pipethrough(
 function getAtprotoPassthroughHeaders(
   headers: IncomingHttpHeaders,
 ): IncomingHttpHeaders {
-  return Object.fromEntries(
-    Object.entries(headers).filter(([name]) =>
-      name.toLowerCase().startsWith('x-atproto-'),
-    ),
-  )
+  // @NOTE node lower-cases all incoming header names, so a case-sensitive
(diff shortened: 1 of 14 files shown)


packages/pds/src/pipethrough.ts
```diff
@@ -251,6 +255,16 @@ export async function pipethrough(
 // Request setup/formatting
 // -------------------
 
+function getAtprotoPassthroughHeaders(
+  headers: IncomingHttpHeaders,
+): IncomingHttpHeaders {
+  return Object.fromEntries(
+    Object.entries(headers).filter(([name]) =>
+      name.toLowerCase().startsWith('x-atproto-'),
```
devinivy: I believe the header names in `IncomingHttpHeaders` should already be lowercased.  Isn't that right @claude?

I am also a bit inclined to make the implementation more similar to the one in the bsky package (imperative, no intermediate array), only because we're in the hot path here.  But also not bad for the two related implementations to be more similar to each other.

devinivy (approved):

PR #5449, opened 2026-08-26
Support AbortSignal implementations without throwIfAborted

mozzius:
## Summary

- avoid requiring the optional AbortSignal.throwIfAborted method in lex-client
- preserve native abort reasons with a cross-platform fallback
- use the compatibility helper for requests, retries, waits, and paginated record iteration

React Native uses abort-controller@3, whose signals do not implement throwIfAborted. Passing one to lex-client currently throws before the request begins.

Related to APP-2938 and bluesky-social/social-app#11564.

## Test plan

- Use a React Native AbortSignal for a client request and confirm the request begins normally and still aborts when its controller is aborted.

### DIFF
--- a/packages/lex/lex-client/src/util.ts
+++ b/packages/lex/lex-client/src/util.ts
@@ -286,20 +286,23 @@ export function mergeHeaders(
 }
 
 function getAbortReason(signal: AbortSignal): unknown {
-  return (
-    signal.reason ??
-    (typeof DOMException !== 'undefined'
-      ? new DOMException('Aborted', 'AbortError')
-      : new Error('Aborted'))
-  )
+  if ('reason' in signal) return signal.reason
+  return typeof DOMException !== 'undefined'
+    ? new DOMException('Aborted', 'AbortError')
+    : new Error('Aborted')
 }
 
 /**
- * Throws when a signal is aborted without requiring the optional
- * `AbortSignal.throwIfAborted()` API.
+ * Uses the native `AbortSignal.throwIfAborted()` when available, falling back
+ * for older implementations such as React Native's `abort-controller`.
  */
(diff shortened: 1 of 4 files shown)


packages/lex/lex-client/src/util.ts
```diff
@@ -285,12 +285,29 @@ export function mergeHeaders(
...
+      : new Error('Aborted'))
+  )
+}
+
+/**
+ * Throws when a signal is aborted without requiring the optional
+ * `AbortSignal.throwIfAborted()` API.
+ */
+export function throwIfAborted(signal?: AbortSignal): void {
+  if (signal?.aborted) throw getAbortReason(signal)
```
matthieusieben: I would just want to make sure (and document) that `throwIfAborted` does the same thing. 

A safer alternative might be to add

```
if (signal && 'throwIfAborted' in signal) signal.throwIfAborted()
```

Right above

packages/lex/lex-client/src/util.ts
```diff
@@ -285,12 +285,29 @@ export function mergeHeaders(
...
 }
 
+function getAbortReason(signal: AbortSignal): unknown {
+  return (
+    signal.reason ??
+    (typeof DOMException !== 'undefined'
+      ? new DOMException('Aborted', 'AbortError')
+      : new Error('Aborted'))
+  )
+}
```
estrattonbailey, line 293: Is this comprehensive?

mozzius: This is what the existing code had

packages/lex/lex-client/src/util.ts
mozzius: Done

matthieusieben (approved):

PR #5453, opened 2026-08-27
bsky: count fail-open hydration failures

rafaeleyng:
Three hydration steps run with `AbortSignal.timeout(100)` and fail open. Two of them swallow the error silently, and the third only logs it. Either way, we have no way to see how often that happens in production.

This adds a `hydration_failed` counter, with attributes:

1. `source`: `known_likers`, `known_followers`, `activity_subscriptions`.
2. `reason`: `timeout` or `error`. The catch blocks swallow every error, not just timeouts, so a counter named after timeouts alone would be lying.

Six timeseries in total. Lands in VictoriaMetrics as `hydration_failed_total`.

The call sites:

1. `Hydrator.hydrateKnownLikers` in `packages/bsky/src/hydration/hydrator.ts`.
2. `ActorHydrator.getKnownFollowers` in `packages/bsky/src/hydration/actor.ts`.
3. `ActorHydrator.getActivitySubscriptions` in `packages/bsky/src/hydration/actor.ts`.

Also adds `packages/bsky/src/events.ts`, which bsky didn't have. It follows `packages/pds/src/events.ts` and the template in STYLE_GUIDE.md, so it writes the event to pino and to the OTEL Logs API on top of the counter. That log half is arguable here - there is no high-cardinality detail to carry, `source` and `reason` are the whole event - so it can come out along with the `@opentelemetry/api-logs` dependency if we'd rather keep this to just the counter.

### DIFF
--- a/.changeset/hip-donkeys-shave.md
+++ b/.changeset/hip-donkeys-shave.md
@@ -1,0 +1,5 @@
+---
+'@atproto/bsky': patch
+---
+
+Add a `hydration_failed` counter for the fail-open hydration steps (known likers, known followers, activity subscriptions)
--- a/packages/bsky/package.json
+++ b/packages/bsky/package.json
@@ -56,6 +56,7 @@
     "@growthbook/growthbook": "^1.6.2",
     "@hapi/address": "^5.1.1",
     "@opentelemetry/api": "^1.9.1",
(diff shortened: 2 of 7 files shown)


rafaeleyng:
Moving these changes to another branch.

PR #5459, opened 2026-08-28
ozone/mod-inbox: `tools.ozone.inbox.appealActionedSubject`

gcwill70:
## What & why

New `tools.ozone.inbox.appealActionedSubject` for mod inbox!

Updates how appeals are handled by attaching the decision being appealed and properly routing the appeal to the specialized queue that the actioned report was in.

The existing appeal path is still available (`com.atproto.moderation.createReport` with `tools.ozone.report.defs#reasonAppeal`)

## Checklist

- [ ] `pnpm build --force && pnpm verify` passes
- [ ] Tests pass, and new behavior is covered by tests

### DIFF
--- a/lexicons/tools/ozone/inbox/appealActionedSubject.json
+++ b/lexicons/tools/ozone/inbox/appealActionedSubject.json
@@ -1,0 +1,66 @@
+{
+  "lexicon": 1,
+  "id": "tools.ozone.inbox.appealActionedSubject",
+  "defs": {
+    "main": {
+      "type": "procedure",
+      "description": "Appeal a moderation action affecting the user's account or content. Either actionId or subject is required.",
+      "input": {
+        "encoding": "application/json",
+        "schema": {
+          "type": "object",
+          "properties": {
+            "actionId": {
+              "type": "integer",
+              "minimum": 1,
+              "description": "Moderation action being appealed. Either this or subject is required."
+            },
+            "subject": {
+              "type": "union",
+              "refs": [
+                "com.atproto.admin.defs#repoRef",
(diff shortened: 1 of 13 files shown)


lexicons/tools/ozone/inbox/appealActionedSubject.json
```diff
@@ -0,0 +1,66 @@
...
+  "defs": {
+    "main": {
+      "type": "procedure",
+      "description": "Appeal a moderation action affecting the user's account or content. Either actionId or subject is required.",
+      "input": {
+        "encoding": "application/json",
+        "schema": {
+          "type": "object",
+          "properties": {
+            "actionId": {
```
foysalit, line 13: we'd probably want this a bit more flexible and make it a union. i can see a couple of ways a client may want to appeal an action and actionId is the most direct way but not every api surface would have actionId available. so for instance, when an account is suspended, user may send appeal without having an actionId available or a label that was placed on their post. so a schema like 

```
action: {
  type: "union",
  refs: [
    "#actionRef",
    "#takedownRef",
    "#labelRef"
  ]
}
...
actionRef: {
  type: 'object',
  required: ['id'],
  properties: {
    id: {type: 'integer'} 
  }
},
labelRef: {
  type: 'object',
  required: ['val'],
  properties: {
    val: {type: 'string'} 
  }
},
takedownRef: {
  type: 'object',
  properties: {}
}
```

this way, i think we can eventually expand the lexicon without causing too much issue but also give users a way to appeal actions without having an ozone actionId available to them.

foysalit (commented):
just some initial thoughts on the lexicon

PR #5468, opened 2026-08-31
Harden Ozone PDS requests against SSRF

foysalit:
## Summary

- route DID-derived Ozone PDS requests through connection-time SSRF protection
- protect `did:web` resolution in both the API and daemon identity resolvers
- reject private/non-unicast DNS results, IP hosts, non-HTTPS production endpoints, and redirects
- preserve nonstandard HTTPS ports and the blob path’s existing 10-second header and 30-second body-inactivity timeouts
- add regression coverage for loopback, metadata, HTTP, redirects, custom ports, and `did:web` resolution

## Context

Ozone previously validated only the URL scheme, literal IP hostnames, and `localhost` before constructing an `AtpAgent`. DNS names resolving to private addresses and DNS rebinding remained possible. Blob diversion passed the PDS endpoint from a DID document directly to `undici.request()`, and the default `did:web` resolver fetched identity-controlled hosts without private-address filtering.

This change deliberately does not introduce response-size policy or a new global PDS timeout.

## Validation

- Ozone TypeScript build
- focused ESLint checks
- focused SSRF tests: 8 passed
- `git diff --check`
- two adversarial cross-model reviews; final review has no findings

The existing blob-diversion integration test could not run locally because Docker was unavailable and `DB_TEST_POSTGRES_URL` was not configured.

### DIFF
--- a/packages/ozone/src/daemon/blob-diverter.ts
+++ b/packages/ozone/src/daemon/blob-diverter.ts
@@ -14,8 +14,7 @@ import type { Database } from '../db/index.js'
 import { createSafeFetch } from '../safe-fetch.js'
 import { retryHttp } from '../util.js'
 
-const MAX_BLOB_SIZE = 100 * 1024 * 1024
-const safeBlobFetch = createSafeFetch(MAX_BLOB_SIZE)
+const safeBlobFetch = createSafeFetch()
 
 export class BlobDiverter {
   serviceConfig: BlobDivertConfig
--- a/packages/ozone/src/daemon/blob-diverter.ts
+++ b/packages/ozone/src/daemon/blob-diverter.ts
@@ -85,7 +85,6 @@ export class BlobDiverter {
         stream: verifier,
       }
     } catch (err) {
-      // Typically un-supported content encoding
       await blobResponse.body?.cancel()
       throw err
     }
--- a/packages/ozone/src/daemon/blob-diverter.ts
+++ b/packages/ozone/src/daemon/blob-diverter.ts
@@ -1,4 +1,4 @@
-import { Readable, Transform, type TransformCallback } from 'node:stream'
+import { Readable } from 'node:stream'
 import { finished, pipeline } from 'node:stream/promises'
 import { CID } from 'multiformats/cid'
 import * as undici from 'undici'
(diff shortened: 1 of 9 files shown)


packages/ozone/src/daemon/blob-diverter.ts
```diff
@@ -35,43 +42,50 @@ export class BlobDiverter {
...
+    const headersController = new AbortController()
+    const headersTimer = setTimeout(
+      () => headersController.abort(),
+      BLOB_HEADERS_TIMEOUT,
+    )
+    headersTimer.unref()
+
+    const blobResponse = await this.fetch(blobUrl, {
+      signal: headersController.signal,
+    })
```
devinivy, line 57: May be able to use `AbortSignal.timeout()` to remove some boilerplate.

devinivy: But also, `safeFetchWrap()` has a timeout.

packages/ozone/src/daemon/blob-diverter.ts
```diff
@@ -145,6 +159,51 @@ export class BlobDiverter {
   }
 }
 
+class BodyTimeoutTransform extends Transform {
```
devinivy: May be worth moving to `../safe-fetch.js`.

Also, I think `pipeline()` may accept a `signal` that we could use with `AbortSignal.timeout()` to get a similar effect with a bit less code/ceremony.

packages/ozone/src/safe-fetch.ts
```diff
@@ -0,0 +1,53 @@
...
+import { type Fetch, safeFetchWrap } from '@atproto-labs/fetch-node'
+
+const DID_DOC_PATH = '/.well-known/did.json'
+
+export const createSafeFetch = (timeout = Infinity) => {
+  const safeFetch = safeFetchWrap({
+    allowCustomPort: true,
+    allowImplicitRedirect: true,
+    allowIpHost: false,
+    responseMaxSize: Infinity,
```
devinivy: Would recommend _some_ max response size.

packages/ozone/src/safe-fetch.ts
```diff
@@ -0,0 +1,53 @@
...
+    }
+    if (parts.length !== 1) {
+      throw new UnsupportedDidWebPathError(did)
+    }
+
+    const url = new URL(`https://${parts[0]}${DID_DOC_PATH}`)
+    const res = await this.safeFetch.call(globalThis, url, {
+      headers: { accept: 'application/did+ld+json,application/json' },
+    })
+    if (!res.ok) return null
```
devinivy: In this case I think we generally need to `await res.body?.cancel()` to clean up resources.  Node-specific thing, this isn't an issue in browser.

devinivy (approved):

packages/ozone/src/safe-fetch.ts
foysalit: yep great catch!

packages/ozone/src/daemon/blob-diverter.ts
foysalit: the main difference here is waiting until response header arrives and a body inactvity timeout that resets whenever data flows again which isn't done in the timeout in safeFetchWrap

PR #5479, opened 2026-09-02
ozone: only hook instrumented modules

gcwill70:
## What & why

Using otel tracer in ozone can cause runtime errors for star-imported modules.

This PR scopes otel down to only trace modules that are instrumented to avoid valid star-import ESM syntax from causing runtime errors.

## Reproduction Steps

On this branch, revert the tracer change and rebuild:
```
git checkout main -- packages/ozone/src/tracer.ts
cd packages/ozone && pnpm run build
```

Create a static import of an affected name (from `packages/ozone`, so `@atproto/lex` resolves):
```
cat > repro.mjs <<'EOF'
import { isBlobRef } from '@atproto/lex'
console.log('ok', typeof isBlobRef)
EOF
```

Run it with the tracer active (any OTLP endpoint value; no collector needed):
```
OTEL_EXPORTER_OTLP_TRACES_ENDPOINT=http://127.0.0.1:4318 \
node --enable-source-maps --import ./dist/tracer.js ./repro.mjs
```

Output:
```
SyntaxError: The requested module '@atproto/lex' does not provide an export named 'isBlobRef'
```

Restore this branch's tracer, rebuild, and re-run:
```
git checkout HEAD -- src/tracer.ts && pnpm run build
OTEL_EXPORTER_OTLP_TRACES_ENDPOINT=http://127.0.0.1:4318 \
node --enable-source-maps --import ./dist/tracer.js --input-type=module -e "
const { isBlobRef } = await import('@atproto/lex')
const pg = await import('pg')
console.log(typeof isBlobRef)                                        // function
console.log(pg.default.Client.prototype.query.__wrapped === true)    // true (pg still instrumented)
process.exit(0)
"
rm repro.mjs
cd ../..
```

Output:
```
function
true
```

## Checklist

Written with fable5

### DIFF
--- a/packages/ozone/src/tracer.ts
+++ b/packages/ozone/src/tracer.ts
@@ -18,7 +18,21 @@ const otlpEndpoint =
 if (otlpEndpoint) {
   // ESM loader hook, required for instrumentation of ES modules. Registered
   // before the SDK imports below so it is in place for everything after us.
-  register('@opentelemetry/instrumentation/hook.mjs', import.meta.url)
+  //
+  // The message channel restricts the hook to the modules that the
+  // instrumentations below actually target. Without it, import-in-the-middle
+  // wraps every ES module, and its export-name analysis drops names reachable
+  // through more than one `export *` path (a legal pattern that resolves to a
+  // single binding under the ESM spec), breaking imports from packages it was
+  // never meant to touch.
+  const { createAddHookMessageChannel } = await import('import-in-the-middle')
+  const { registerOptions, waitForAllMessagesAcknowledged } =
+    createAddHookMessageChannel()
+  register(
+    '@opentelemetry/instrumentation/hook.mjs',
+    import.meta.url,
+    registerOptions,
+  )
 
   // Imported lazily so that when tracing is disabled the OTel SDK is never
   // loaded at all.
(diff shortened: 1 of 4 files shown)


foysalit (approved):

devinivy (approved):

packages/ozone/src/tracer.ts
```diff
@@ -18,7 +18,21 @@ const otlpEndpoint =
...
   // before the SDK imports below so it is in place for everything after us.
-  register('@opentelemetry/instrumentation/hook.mjs', import.meta.url)
+  //
+  // The message channel restricts the hook to the modules that the
+  // instrumentations below actually target. Without it, import-in-the-middle
+  // wraps every ES module, and its export-name analysis drops names reachable
+  // through more than one `export *` path (a legal pattern that resolves to a
+  // single binding under the ESM spec), breaking imports from packages it was
+  // never meant to touch.
+  const { createAddHookMessageChannel } = await import('import-in-the-middle')
```
matthieusieben, line 28: This should be imported at the top.

packages/ozone/src/tracer.ts
```diff
@@ -51,6 +65,11 @@ if (otlpEndpoint) {
 
   sdk.start()
 
+  // Wait for the hook thread to acknowledge the modules the instrumentations
+  // registered above, so application imports resolving after us are correctly
+  // wrapped (and only those).
+  await waitForAllMessagesAcknowledged()
```
matthieusieben, line 71: The top level await is unfortunate but it looks like there is no other option.

matthieusieben (commented):
I would move the import from `import-in-the-middle` to the top (as a regular import), but except from that, look good.

PR #5305, opened 2026-08-03
Redesign oauth-provider-ui

bigmoves:
## Summary

Rebuilds `oauth-provider-ui` on shadcn/ui with Base UI primitives. Same flows, same steps, same copy. Forms keep the browser's native constraint validation, so field errors render in the user's browser locale. Branding colors configured on the PDS apply at runtime. The pds e2e suites are unchanged and finish at their baseline (`oauth.test.ts` 7/7, `account-manager.test.ts` 13/13); the added tests are unit tests for the theme-token contract and the customization CSS.

## Key Changes

- **shadcn/ui on Base UI** (`@base-ui/react` 1.6.0, style `base-nova`). `components/ui/*` is registry output, kept byte-faithful apart from documented rewrites so a future `shadcn add` stays mergeable.
- **Forms** are `FormShell` on `@base-ui/react/form` plus one wrapper per field type. Validation stays on the controls as HTML constraint attributes. No form library.
- **Runtime branding.** The provider-injected `--branding-color-*` variables map onto the shadcn tokens for `primary` and `error`, with neutral defaults when a deployment configures nothing. A PDS without curated branding does not look like Bluesky.
- **Account manager** on a collapsible sidebar shell. Mobile sub-pages navigate with a back button, and a sticky header keeps the logo and name in view.
- **Auth screens** are a single card carrying the brand mark, content, locale selector and footer links.
- **Toasts** are built on `@base-ui/react/toast` via the registry component, in place of the Radix toast. **Lucide** icons replace Phosphor.

## Notable Implementation Details

- Base UI's `onFormSubmit` only reports controls registered through `Field.Root`. `FormShell` reads values off the element with `FormData` instead, the same pattern Base UI's own docs use.
- Base UI's `Form` defaults to `noValidate`. `FormShell` re-enables the browser gate so controls outside a `Field.Root` still validate.
- A `Field.Root` binds every descendant control to its own name. `HandleField` composes one value from two controls, so it opts out of `Field` entirely.
- A `disabled` control is omitted from form values. The read-only username on the second sign-in step is `readOnly` only.
- The e2e suites drive the UI by visible text and element tag, in French, so markup is load-bearing: rows the tests click stay real `<button>`s, identifiers stay in `<span>`s, notice titles stay `<h3>`.
- Message catalogs are unchanged apart from one retired msgid (`Close account selector`, gone with the icon button it labelled).
- Deviations inside `components/ui/*` are limited to lint-forced changes and a localized close label, each noted inline.
[...]

PR #5482, opened 2026-09-03
oauth-provider-ui: tighter auth and account screens

abcbrookie:
## Summary

Follow-up to #5305. Same flows and the same field names, just tightened up. Buttons and inputs are one size everywhere, the error pages look like the other screens, and copy that repeated the heading or the button is gone. The reasoning is in `DesignPrinciples.md` so the next change does not have to re-decide any of it.

## Key Changes

- **Buttons and inputs.** 40px buttons, 44px inputs with 16px text (which also stops iOS zooming in on a field). Two constants set the size everywhere: `actionButton` in `form-shell.tsx` and `inputSize` in `text-field.tsx`.
- **Card.** 26rem, or 28rem on consent so the scope rows fit. The logo stands on its own and the service name is its alt text. The background is painted on `<html>` so overscroll on iOS matches.
- **Less copy.** No subtitles or paragraphs that repeat the heading. The phishing warning is one line instead of a box. The consent legal line is just the two links. Apps are named by domain, with the full client id in the details dialog.
- **Error pages** are regular cards now, with Try again and Go back.
- **Account manager** uses the same row style as the picker. An unverified email is a row with a Verify button. Deactivate is a moon, reactivate a sun, and the deactivated badge is grey.
- **Footer** is the language pill, then Home · Terms · Privacy · Support on one line.
- **Docs and mock.** A README with screenshots, `DesignPrinciples.md` with the reasoning, and the mock handles `/consent` and `/reject` so you can reach the redirect screen.
- **i18n.** 25 msgids added or reworded, 23 removed. French is filled in as CLAUDE.md asks. The other five locales are filled in too; drop them if you would rather route those externally.

## Notable Implementation Details

- **Field names, heading tags and the real-button rule did not change.** Six e2e assertions matched French copy that is gone now, so they match the new strings. `oauth.test.ts` 7/7, `account-manager.test.ts` 13/13.
- **`components/ui/*` is untouched.** Sizes are classes at the call site, so `shadcn add` still merges cleanly.
- **Line breaks are in the messages.** Where a heading has to break in a specific spot, the `\n` is in the msgid and each locale puts its own. Adding an element inside `<Trans>` would have renumbered the placeholders and orphaned every translation.
- **The background moved from a div to `<html>`.** `AuthShell` toggles a class in an effect and `body` goes transparent under it. The branding variables are read the same way as before.
- **`ErrorNotice` is its own component now** instead of wrapping `Notice`. `ConsentForm` lost its `account` prop because the account is only named once, under the heading.
- **Footer labels come from each link's `rel`,** not its title, so they stay one word in every locale.

## Screenshots

From the mock at 390×844. The illustration and the Bluesky logo are placeholders; a real PDS supplies its own.

### Sign in

| Sign in | Confirm password | 2FA code | Wrong credentials |
|---|---|---|---|
|  |  |  |  |

### Consent and redirect

| Account picker | Consent | Client details | Redirecting |
|---|---|---|---|
|  |  |  |  |

### Errors

| Error | Cookies blocked | Reactivate |
|---|---|---|
|  |  |  |

### Account manager

| Home | Settings | Devices | Apps |
|---|---|---|---|
|  |  |  |  |

## Checklist

- [x] Written with the help of an LLM or coding agent? Yes, Claude Code. I directed and reviewed every change against the mock and read the whole diff.


devinivy:
This is awesome stuff, will definitely be digging into this soon!

packages/oauth/oauth-provider-ui/src/components/identity/account-avatar.tsx
```diff
@@ -1,5 +1,5 @@
 import { useLingui } from '@lingui/react/macro'
-import { SnowflakeIcon, UserIcon } from 'lucide-react'
+import { MoonIcon, UserIcon } from 'lucide-react'
```
bigmoves, line 2: praise: moon icon is better than snowflake icon

packages/oauth/oauth-provider-ui/src/components/forms/form-shell.tsx
```diff
@@ -18,6 +18,13 @@ export const actionRow = [
   'sm:flex-row-reverse sm:flex-wrap sm:items-center sm:justify-start',
 ].join(' ')
 
+/**
+ * Size for the buttons in an action row — a touch taller than the stock
+ * button so the primary action is an easy tap on a phone. Shared with the
+ * sign-in picker so every screen's actions read as one set.
+ */
+export const actionButton = 'h-10 text-[15px]'
```
bigmoves, line 26: suggestion: instead of exporting class names, could turn this into a component

```tsx
export function ActionButton({ className, ...props }: ComponentProps<typeof Button>) {
  return <Button {...props} className={cn('text-action h-10', className)} />
}
```

bigmoves: Also could move arbitrary tokens like 15px to a theme token in styles.css

```css
--text-action: 0.9375rem;
```

packages/oauth/oauth-provider-ui/src/components/layouts/auth-shell.tsx
```diff
@@ -60,36 +96,42 @@ export function AuthShell({
...
 
   return (
-    <div className="auth-background flex min-h-svh flex-col items-center justify-center gap-6 p-6 md:p-10">
+    <div className="flex min-h-svh flex-col items-center justify-center gap-6 p-4 sm:p-6 md:p-10">
       {documentTitleString && <title>{documentTitleString}</title>}
 
       <div
         {...props}
-        className={cn('flex w-full max-w-sm flex-col', className)}
+        className={cn('flex w-full max-w-[26rem] flex-col', className)}
```
bigmoves, line 104: suggestion: could move this to a token as well 

```css
--container-auth-card: 26rem;
```

packages/oauth/oauth-provider-ui/src/components/sign-in-form.tsx
```diff
@@ -131,7 +130,12 @@ export function SignInForm({
         name="username"
         defaultValue={usernameDefault}
         pattern={SIGN_IN_IDENTIFIER_PATTERN}
-        label={<Trans>Identifier</Trans>}
+        label={<Trans>Username</Trans>}
```
bigmoves, line 133: suggestion: it does still support an email address so we need to deal with that somehow, maybe "Username or email address" is right the label

packages/oauth/oauth-provider-ui/src/components/layouts/auth-shell.tsx
```diff
@@ -98,16 +140,21 @@ export function AuthShell({
 
           <CardContent>{children}</CardContent>
 
-          <CardFooter className="flex-col justify-center gap-3">
-            <LocaleSelector />
+          <CardFooter className="flex-col justify-center gap-4 border-t-0 bg-transparent pt-2">
+            {/* @NOTE Same height as the action buttons; the trigger sizes
+              itself through a data attribute, so the override carries the
+              same variant. */}
+            <LocaleSelector className="px-3 text-[15px] data-[size=sm]:h-10" />
```
bigmoves, line 147: same here, could use the token

packages/oauth/oauth-provider-ui/src/components/utils/account-card.tsx
```diff
@@ -23,6 +23,21 @@ export type AccountCardProps = Override<
   }
 >
 
+/**
+ * Row utilities shared with the "Another account" row in `SignInPicker`, so
+ * the two stay the same height, padding and surface.
+ */
+export const accountRowClassName =
```
bigmoves, line 30: suggestion: this could be an AccountRow component rather than exporting classnames

packages/oauth/oauth-provider-ui/src/components/authenticate-welcome-view.tsx
```diff
@@ -54,7 +56,11 @@ export function AuthenticateWelcomeView({
           <>
             <Separator className="my-2" />
 
-            <Button className="w-full" variant="secondary" onClick={onCancel}>
+            <Button
```
bigmoves, line 59: see above: this could be ActionButton instead of importing the classname

packages/oauth/oauth-provider-ui/src/components/forms/fields/text-field.tsx
```diff
@@ -22,6 +22,13 @@ export type FieldBaseProps = {
   below?: ReactNode
 }
 
+/**
+ * Size for every text control in the auth forms: 44px tall for a comfortable
+ * tap target, and 16px text at every breakpoint — `Input` drops to 14px from
+ * `md` up, and 16px is also what stops iOS zooming into a focused field.
+ */
+export const inputSize = 'h-11 text-base md:text-base'
```
bigmoves, line 30: suggestion: this could be a prop rather than a css export

packages/oauth/oauth-provider-ui/src/components/layouts/auth-shell.tsx
```diff
@@ -29,6 +30,33 @@ export type AuthShellProps = Override<
   }
 >
 
+/**
+ * The footer names the standard links by one word each — "Terms" rather
+ * than "Terms of Service" — keyed on the link's `rel`, so all four fit on one
+ * line at phone width. A link with any other `rel` keeps its own title.
+ */
+function shortLinkLabel(rel: string | undefined): ReactNode {
```
bigmoves, line 38: note: this makes the labels different on the auth screens ie Terms vs Terms of Service on the dashboard pages. not a huge deal i don't think just pointing it out. maybe we switch to the shortened versions in both cases?

packages/oauth/oauth-provider-ui/src/components/identity/client-name.tsx
```diff
@@ -49,18 +46,16 @@ export function ClientName({
   }
 
   if (url?.protocol === 'https:') {
-    // Only display the url details if the client id does not follow our
-    // convention.
-    const simplifiedView = isConventionalOAuthClientId(clientId)
```
bigmoves, line 54: issue (blocking): i don't think we want to remove this? two different apps on the same domain would look identical, and the only thing telling them apart sits behind "Technical details."

bigmoves (commented):
@abcbrookie these changes look great! added a few comments. there's one issue around the client name i pointed out but @matthieusieben is the better judge on that (see https://github.com/bluesky-social/atproto/pull/5482/changes#diff-a2f102a3cbbc9a6bc180164c35aa2f956e6b858ef0340dfa5c0dc990d625a222L52)

packages/oauth/oauth-provider-ui/src/components/forms/form-shell.tsx
abcbrookie: Yeah definitely.

packages/oauth/oauth-provider-ui/src/components/sign-in-form.tsx
abcbrookie: "Username or Email" sounds good to me !

packages/oauth/oauth-provider-ui/src/components/layouts/auth-shell.tsx
abcbrookie: Oh yeah, definitely I was thinking the shortened version in both cases. Oversight on my end.

packages/oauth/oauth-provider-ui/src/components/identity/client-name.tsx
matthieusieben: Agree with @bigmoves here. I know this is a controversial topic but there are security concerns with this. E.g. consider these two client ids:

- https://raw.githubusercontent.com/legit.json
- https://raw.githubusercontent.com/evil.json

We do *not* want both to be simplified into "`raw.githubusercontent.com`".

Only clients using the "blessed" convention should be displayed in a user friendly way.

Note that an alternative might be to only show the domain name **along with** a scary warning "this client is using a non-conventional client identifier".

packages/oauth/oauth-provider-ui/src/components/layouts/auth-shell.tsx
```diff
@@ -29,6 +30,33 @@ export type AuthShellProps = Override<
...
+      return undefined
+  }
+}
+
+/**
+ * The card's inner spacing. `Card` sets `--card-spacing` to 1rem through an
+ * arbitrary-property utility; a competing utility would race it in the
+ * stylesheet, so the wider auth spacing is set inline instead.
+ */
+const cardSpacing = { '--card-spacing': '1.5rem' } as CSSProperties
```
matthieusieben: Any reason for casting here? If so, then a comment would be welcome. If not, let's just type it as:

```suggestion
const cardSpacing: CSSProperties = { '--card-spacing': '1.5rem' }
```

matthieusieben (commented):
Nice tidying! 1 or 2 minor things to address but otherwise great.

packages/oauth/oauth-provider-ui/src/components/identity/client-name.tsx
abcbrookie: It would only simplify into "raw.githubusercontent.com" when the the client metadata is conventional. 

Like: "https://raw.githubusercontent.com/oauth-client-metadata.json"

If it was "https://raw.githubusercontent.com/evil.json" the full URL would be displayed.

abcbrookie: Oops! Yeah my bad. It definitely should continue to work this way, only simplifying to the host name if it matches the  "https://.../oauth-client-metadata.json" format.

packages/oauth/oauth-provider-ui/src/components/layouts/auth-shell.tsx
abcbrookie: This looks good to me. It shouldn't change anything. 

edit: actually I checked and it doesn't compile with this change. I found a better way to do it using `<Card className="[--card-spacing:--spacing(6)]">` 38d79957cfcc0b648ddf1b043edf922d86fc2e43

packages/oauth/oauth-provider-ui/src/components/sign-in-form.tsx
abcbrookie: 943d5b316c7ee9b616838dfbcf5e49644ab28a67 updated :)

packages/oauth/oauth-provider-ui/src/components/identity/client-name.tsx
abcbrookie: updated 82aebd87b9b51b1cd52a803fd0be143102af3462

packages/oauth/oauth-provider-ui/src/components/authenticate-welcome-view.tsx
abcbrookie: updated 3bd22bfa4d8e7202bfc0ecf27adfeb373bf1427a

packages/oauth/oauth-provider-ui/src/components/forms/form-shell.tsx
abcbrookie: Both done in 3bd22bfa4d8e7202bfc0ecf27adfeb373bf1427a.

packages/oauth/oauth-provider-ui/src/components/layouts/auth-shell.tsx
abcbrookie: Updated 3bd22bfa4d8e7202bfc0ecf27adfeb373bf1427a

packages/oauth/oauth-provider-ui/src/components/layouts/auth-shell.tsx
abcbrookie: 3bd22bfa4d8e7202bfc0ecf27adfeb373bf1427a the locale selector uses text-action now.

packages/oauth/oauth-provider-ui/src/components/utils/account-card.tsx
abcbrookie: Updated 3bd22bfa4d8e7202bfc0ecf27adfeb373bf1427a

packages/oauth/oauth-provider-ui/src/components/forms/fields/text-field.tsx
abcbrookie: 3bd22bfa4d8e7202bfc0ecf27adfeb373bf1427a done as a component.

packages/oauth/oauth-provider-ui/src/components/layouts/auth-shell.tsx
abcbrookie: 3bd22bfa4d8e7202bfc0ecf27adfeb373bf1427a dashboard version will use it too now :)

PR #5497, opened 2026-09-09
fix(pds): resolve local lexicons without public self-fetch

devxan:
## What & why

The PDS fetched its own Lexicons through its public hostname, which can fail under local networking or SSRF protection. The fix reads local signed proofs directly while preserving validation and remote protections.

The reference PDS not resolving lexicon schema if the repo owner of the schema is on the same PDS has been an issue for [developers before](https://marvins-guide.leaflet.pub/3mbfvey7sok26/l-quote/119_0-119_516#119_0); I just came across it when trying to sign in to a local build of https://witchsky.app from @witchsky.app (using oauth with `app.witchsky.theme.authFull`), which at the time is hosted on a ref pds. 

## Checklist

- [x] Written with the help of an LLM or coding agent? **GPT-6**

### DIFF
--- a/.changeset/quiet-clubs-mix.md
+++ b/.changeset/quiet-clubs-mix.md
@@ -1,0 +1,5 @@
+---
+"@atproto/pds": patch
+---
+
+Resolve locally hosted OAuth permission-set Lexicons using local repository proofs instead of fetching the PDS public hostname. This avoids failures when self-fetches are blocked by local networking or SSRF protection, while retaining DID authority, signature and schema validation, and account availability checks.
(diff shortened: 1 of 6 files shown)


devxan:
#5499 was opened to resolve the same issue, but it may cause failures if a DID has migrated while an old local repository remains; however, it may be a bit faster due to the onFetch running before a DID doc lookup.

matthieusieben:
Thank you for engaging with this. https://github.com/bluesky-social/atproto/pull/5499 fixes this issue. Feel free to open another PR/issue if you think that didn't fix the issue.

PR #5498, opened 2026-09-10
deps: override form-data, fast-xml-parser and protobufjs past their critical advisories

Portll:
## What

Three `overrides` in `pnpm-workspace.yaml`, scoped to the vulnerable ranges so they stop applying once the parents move on their own, and the lockfile re-resolved with `pnpm install --lockfile-only`:

| package | was | now | advisory |
|---|---|---|---|
| form-data | 4.0.0 | 4.0.6 | GHSA-fjxv-7rqg-78g4 / CVE-2025-7783, unsafe random boundary, critical |
| fast-xml-parser | 5.2.5 | 5.11.1 | GHSA-m7jm-9gc2-mpf2 / CVE-2026-25896, entity encoding bypass, critical |
| protobufjs | 7.2.5 | 7.6.5 | GHSA-xq3m-2v4x-88gg / CVE-2026-41242, arbitrary code execution, critical |

`protobufjs@7.6.5` was already in the tree, so that one only removes the duplicate older copy.

## Why

All three are transitive pins that no workspace package requests directly, so Renovate/Dependabot do not raise them, and the lockfile keeps resolving the old versions until a parent bumps its range. The override makes the lockfile carry the fixed versions now.

## Verification

Lockfile-only resolution with pnpm 11.11.0 (`Lockfile passes supply-chain policies`). No install or test run was done on the machine that prepared this; CI is the first full run. The peer-dependency warning pnpm prints is pre-existing.

Found by commitwork.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -206,6 +206,9 @@ settings:
 overrides:
   cookie: ^0.7.2
   opentelemetry-plugin-better-sqlite3>@opentelemetry/core: ^1.30.0
+  form-data@>=4.0.0 <4.0.4: '>=4.0.4'
+  fast-xml-parser@>=5.0.0 <5.3.5: '>=5.3.5'
+  protobufjs@<7.5.5: '>=7.5.5'
 
 importers:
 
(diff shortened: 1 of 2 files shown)


matthieusieben:
Rebased on `main` in https://github.com/bluesky-social/atproto/pull/5532

PR #5499, opened 2026-09-10
bugfix: resolves lexicon permission schemas on local PDS

fatfingers23:
## What & why

This fixes a bug with looking up permission schemas if they are hosted on the same PDS as the account logging in, does not look like it's able to currently resolve them and does not allow a login. One example is the Witchsky's permission sets as show in this [post](https://bsky.app/profile/xan.lol/post/3mv33fku6b22e).

image

Can recreate by doing this.
1. Identity A has an atproto app with a permission set in their lexicon schema repo.
2. Identity B is also on the same PDS as Identity A, when they try to login they get the above error. 

Can also see in prod by making an account on [selfhosted.social](https://selfhosted.social) and try to login to [canary.witchsky.app](https://canary.witchsky.app/)

May be a bit overboard to be getting the record from the actor store directly, the direct cause looked to be the PDS url through a "SSRF-protected fetch." Very open to any other solutions and I can keep digging if needed

## Checklist

- [Claude] Written with the help of an LLM or coding agent? Say so here, and name the tooling.


matthieusieben (approved):
Looks good, thanks !

packages/pds/src/context.ts
```diff
@@ -344,6 +349,64 @@ export class AppContext implements AsyncDisposable {
...
+      })
+
+      // Not hosted here: fall back to the network resolution
+      if (!account) return undefined
+
+      // Mirror what com.atproto.sync.getRecord would answer to an
+      // unauthenticated requester
+      if (account.takedownRef || account.deactivatedAt) {
+        throw new LexResolverError(nsid, `Repo is not available: ${did}`)
+      }
```
matthieusieben: If the account is deactivated because it was migrated, this should not result in an error.

```suggestion
      if (!account) return undefined

      // Account deactivated, maybe it was moved somewhere else
      if (account.deactivatedAt) return undefined

      // Mirror what com.atproto.sync.getRecord would answer to an
      // unauthenticated requester
      if (account.takedownRef) {
        throw new LexResolverError(nsid, `Repo is not available: ${did}`)
      }
```

matthieusieben (changes requested):
Looks good. I think we should handle deactivated accounts differently though

matthieusieben (approved):

PR #5506, opened 2026-09-11
fix: add missing Bluesky full-app OAuth permissions

devxan:
## What & why

Adds missing draft, suggested-user, content-visibility, and verification permissions to `app.bsky.authFullApp` so OAuth clients requesting full Bluesky access don't have to request more scopes.

I felt the need for these when trying to use permissions sets in my forked bluesky client https://witchsky.app/ and suddenly finding that I couldn't save to drafts anymore...

## Checklist


matthieusieben (commented):
This requires a changeset in the affected packages.

devxan:
I looked at other changes that only altered lexicons and they didn't have changesets. Which package would this affect?

PR #5516, opened 2026-09-15
fix(lex-client): preserve service URL path prefix

danielroe:
## What & why

whilst building a demo for [airspace](https://github.com/danielroe/airspace), I noticed that the prefix I wanted was being stripped out

```ts
const client = new Client({ service: 'https://example.com/proxy' })
// requests go to https://example.com/xrpc/...
```

that's because `new URL(path, service)` always resolves the path relative to the root. now, I _think_ that service is meant to support a path, so I'm taking this as a bug rather than intended

... so we now join the base pathname first.

however, if I've got this wrong and `service` is meant to be origin-only, feel free to close! 🙏

## Checklist

   > pi + opus 5 for the code


matthieusieben:
Hey @danielroe,

This is an interesting proposal we definitely don't want to dismiss.

The [Lexicon XRPC spec](https://atproto.com/specs/xrpc#lexicon-http-endpoints) specifically states that XRPC endpoints should start with `/xrpc/`, which explains why things are the way they are.

The `@atproto/xrpc-server` and `@atproto/lex-server` both hard wire `/xrpc/` prefixes too, relying on "fallthrough" and headers related logic to route requests to other services.

Could you elaborate a little bit why you need this, and in particular why you can't use other existing paradigms (like the `atproto-proxy` header) for this?

Also note that `service` is only used to build a `FetchHandler` function that can be built manually like so:

```ts
new Client(async (path, init) => {
  return globalThis.fetch(`https://example.com/proxy${path}`, init)
})
```

matthieusieben:
This is such an edge case, and with an available workaround, that I am going to close this issue in favor of recommending the custom fetch handler function instead.

danielroe:
sorry for the delay in replying!

I am totally fine with that... I think maybe we can clarify (in the JSdoc?) so that someone else doesn't fall into my pit/pattern

PR #5517, opened 2026-09-15
Ozone: purge video content on blob takedown

foysalit:
## What & why

Adds a dedicated video invalidator to Ozone blob takedowns so transcoded video content can be purged independently of image CDN invalidation. The invalidator calls a configured purge endpoint with the subject DID and blob CID using a configured helper key.

The endpoint URL and helper key are configurable through `OZONE_VIDEO_INVALIDATION_URL` and `OZONE_VIDEO_INVALIDATION_KEY`. Failures are logged without failing the moderation event.

## Checklist

- [ ] `pnpm build --force && pnpm verify` passes
- [x] Written with the help of an LLM or coding agent: Codex

## Testing

- `NODE_OPTIONS=--experimental-vm-modules ./node_modules/.bin/jest --config packages/ozone/jest.config.cjs tests/video-invalidator.test.ts --runInBand`


packages/ozone/src/video-invalidator.ts
```diff
@@ -0,0 +1,26 @@
...
+export class VideoInvalidator {
+  readonly invalidate: (did: string, cid: string) => Promise<void>
+
+  constructor(cfg: VideoInvalidatorConfig) {
+    this.invalidate = async (did: string, cid: string) => {
+      const url = new URL(cfg.url)
+      url.searchParams.set('did', did)
+      url.searchParams.set('cid', cid)
+
+      const res = await fetch(url.toString(), {
```
gcwill70, line 17: May want to add a timeout here to prevent stalling requests

gcwill70 (approved):
Looks good! May want to add the timeout I mentioned before merging

PR #5527, opened 2026-09-17
PDS: optimize listRecords pagination using the URI index

devinivy:
Large `com.atproto.repo.listRecords` collections were slow to paginate because sqlite filtered by `collection`, sorted by `rkey` in a temp B-tree, and joined every qualifying record before applying the limit.

Paginate over `record.uri` instead. Since record URIs encode the DID, collection, and rkey, sqlite can use the existing URI primary-key index for collection bounds, ordering, and cursors. This avoids adding a new index or migrating actor databases.


packages/pds/src/actor-store/record/reader.ts
```diff
@@ -324,6 +340,16 @@ export class RecordReader {
   }
 }
 
+function formatRecordUriKey(
+  did: string,
```
matthieusieben, line 344: Is it possible/easy to use the "tagged" string types here ? (`DidString`, et.c)

matthieusieben (approved):

packages/pds/src/actor-store/record/reader.ts
devinivy: Yeah, not too bad— I did a brief pass to get those in.

PR #5539, opened 2026-09-21
bsky: gate discovery skeleton routing to Iris

ds-boyce:
## What & why

Routes v2 suggested content traffic to Iris while preserving Topics for control groups.

## Checklist

- [x] Written with the help of an LLM or coding agent? Yes, `gpt-5.6-sol`.


ocseckin (commented):
Hey, this looks good to me, though we probably need one more thing before we can run an experiment. I asked Claude and it says that the starter packs won't be measured at all:

`The logging in social-app [#11701](https://github.com/bluesky-social/social-app/pull/11701) skips starterPack:suggestion:seen and press when there's no recId (if (!recId) return). With the AppView dropping it, neither arm logs starter-pack events. Can we pass it through (and add it to the skeleton lexicons) so the client logging gets it?`

packages/bsky/src/config.ts
```diff
@@ -178,6 +179,7 @@ export class ServerConfig {
     const topicsUrl = process.env.BSKY_TOPICS_URL || undefined
     const topicsApiKey = process.env.BSKY_TOPICS_API_KEY
     const irisUrl = process.env.BSKY_IRIS_URL || undefined
+    const irisApiKey = process.env.BSKY_IRIS_API_KEY || undefined
```
ds-boyce, line 182: https://github.com/bluesky-social/bsky-infra/commit/e8758203974dbd2951c7bb23646a5a2a5f945573

PR #5544, opened 2026-09-23
ozone: add max buffer size and overflow checks

gcwill70:
## What & why

* add max buffer size for label subscriptions
* disconnect when buffer size is reached
* disconnect if event processing is too slow

## Checklist

- [x] Written with codex


foysalit (approved):

dholms (approved):
looks good 👍

PR #5545, opened 2026-09-23
feature: adds the ability to disable accounts via oauth sessions on com.atproto.server.deactivateAccount

fatfingers23:
## What & why

I have been working on a new version of PDS MOOver at https://oauth.pdsmoover.com. The main thing holding back a full OAuth migration is deactivation of the source account via the `com.atproto.server.deactivateAccount` endpoint. Which is not always a deal breaker, but we have seen bugs in the pass about previous accounts staying active and the Bluesky AppView would say the post was deleted if both accounts were on the same PDS before the migration. 

I also think that at the level of OAuth access needed it's the same as a full password, six of one and half a dozen of the other. But it is a feature request I get a lot, and I think for most people it makes them feel a bit more comfortable. 

### Some points on the PR
- mimic a bit what `updateHandle` does with the entryway for auth. Not sure if the entryway supports `entrywayPassthruHeaders` on the `deactivateAccount` endpoint, but if it does we can probably just use that here?
- I am clearing credentials on deactivate since you cannot login with OAuth if an account is deactivated. The hope there was to clear up the sessions since there would be a small window for authenticated requests. Mimics what the oauth-store does for it's deactivateAccount
- Updated the error message on `activate` a bit, I think that that endpoint would only most likely be hit during development, but thought a bit more info would not be a bad thing there if we support deactivate via OAuth sessions

## Checklist

- [Claude Code] Written with the help of an LLM or coding agent? Say so here, and name the tooling.


matthieusieben (approved):

packages/pds/src/api/com/atproto/server/deactivateAccount.ts
```diff
@@ -9,19 +9,27 @@ export default function (server: Server, ctx: AppContext) {
...
+        // are authenticated towards the entryway using service auth instead.
+        // Mimics what updateHandle does
+        const { headers } =
+          auth.credentials.type === 'oauth'
+            ? await ctx.entrywayAuthHeaders(
+                req,
+                auth.credentials.did,
+                com.atproto.server.deactivateAccount.$lxm,
+              )
+            : ctx.entrywayPassthruHeaders(req)
```
matthieusieben: Note: entryway accepts either so we could remove the ternary and always use new `entrywayAuthHeaders`.

packages/pds/src/api/com/atproto/server/deactivateAccount.ts
```diff
@@ -33,6 +41,9 @@ export default function (server: Server, ctx: AppContext) {
       auth,
       handler: async ({ input: { body }, auth }) => {
         await ctx.accountManager.deactivateAccount(auth.credentials.did, {
+          // Revoke credentials for OAuth deactivations since you cannot log into
+          // the account via OAuth if it has been deactivated
```
matthieusieben: I would change this comment to state that for legacy reasons, calling this endpoint with non-oauth creds keeps all those creds alive. We should definitely investigage why it has been like that and address that (in a separate line of work)

packages/pds/tests/oauth-deactivation.test.ts
```diff
@@ -0,0 +1,177 @@
...
+
+    await using page = await PageHelper.from(browser, { languages })
+
+    await page.goto(authorizeUrl)
+
+    // A fresh device has to sign in first. A device that already holds a
+    // session jumps straight to the consent screen.
+    if ((await page.title()) === 'Connexion') {
+      await page.typeInInput('password', 'alice-pass')
+      await page.clickOnText('Se connecter')
```
matthieusieben: I would prefer if the tests where fully predictable, and use an extra argument here to dictate the branching. Somehting like:

```ts
if (options?.performSignIn) {
  await page.expectTitle('Connexion')
  await page.typeInInput('password', 'alice-pass')
  await page.clickOnText('Se connecter')
}
```

matthieusieben (commented):
Looks good overall and confident that we could merge as-is. Would you mind addressing the small comments I left ?

fatfingers23:
> Looks good overall and confident that we could merge as-is. Would you mind addressing the small comments I left ?

Awesome! Yes I can do that in a bit, thank you for the feedback!

packages/pds/tests/oauth-deactivation.test.ts
fatfingers23: Ended up just removing the check. Each test expected a login session for the tests.

matthieusieben (approved):
lgtm thanks

